PR #1015, opened 2026-04-29
chore(dev): add Windows dev runner and align lockfile versions

### LINKED ISSUE #1014: Fix Windows installer permission/locked-file failure for openhuman-core.exe
## Summary
Fix Windows installer failure where OpenHuman setup cannot access or write required files, including `openhuman-core.exe`.

## Problem
Windows users are seeing installer/access errors during setup:

- `Windows cannot access the specified device, path, or file. You may not have the appropriate permissions to access the item.`
- `Error opening file for writing: C:\Users\<user>\AppData\Local\OpenHuman\openhuman-core.exe`

This prevents OpenHuman from installing or updating successfully.

## Observed Behavior
- Running `OpenHuman_0.53.4_x64-setup.exe` may fail with Windows access/permission error.
- Installer starts but fails while extracting/writing `openhuman-core.exe`.
- User is prompted to Abort / Retry / Ignore.

## Expected Behavior
- Windows installer should run without permission errors.
- `openhuman-core.exe` should be installed/updated cleanly.
- If an old core process is running, installer should stop it or show a clear action.
- Errors should provide user-friendly recovery steps.

## Scope
- Reproduce on Windows with fresh install and upgrade scenarios.
- Check whether `openhuman-core.exe` is locked by a running process.
- Ensure installer closes/stops existing OpenHuman/core processes before overwrite.
- Validate install path permissions under `%LOCALAPPDATA%\OpenHuman`.
- Check Windows Defender/SmartScreen/code signing/quarantine behavior.
- Improve installer error handling and preflight checks.
- Add troubleshooting guidance for locked files and permission denial.

## Acceptance Criteria
- Fresh install works on Windows x64.
- Upgrade/install over existing version works without file write failure.
- Running core process does not block install or shows a clear fix path.
- Installer handles permission/locked-file errors gracefully.
- Windows install steps are documented for users and QA.

M3gA-Mind:
Add a Windows-focused dev startup script that auto-discovers pnpm/ninja and configures CEF tooling, while also bumping lockfile package versions to 0.53.4 for consistency.

Made-with: Cursor

## Summary

- What changed and why.
- Keep this to 3-6 bullets focused on user-visible or architecture-impacting changes.

## Problem

- What issue or risk this PR addresses.
- Include context needed for reviewers to evaluate correctness quickly.

## Solution

- How the implementation solves the problem.
- Note important design decisions and tradeoffs.

## Submission Checklist

- [ ] **Unit tests** — Vitest (`app/`) and/or `cargo test` (core) for logic you add or change
- [ ] **E2E / integration** — Where behavior is user-visible or crosses UI → Tauri → sidecar → JSON-RPC; use existing harnesses (`app/test/e2e`, mock backend, `tests/json_rpc_e2e.rs` as appropriate)
- [ ] **N/A** — If truly not applicable, say why (e.g. change is documentation-only)
- [ ] **Doc comments** — `///` / `//!` (Rust), JSDoc or brief file/module headers (TS) on public APIs and non-obvious modules
- [ ] **Inline comments** — Where logic, invariants, or edge cases aren’t clear from names alone (keep them grep-friendly; avoid restating the code)

(Any feature related checklist can go in here)

## Impact

- Runtime/platform impact (desktop/mobile/web/CLI), if any.
- Performance, security, migration, or compatibility implications.

## Related

- Closes:
- Follow-up PR(s)/TODOs:

## Summary by CodeRabbit

* **New Features**
  * Added a Windows/Git-Bash dev runner for local development workflows.

* **Improvements**
  * Screen Intelligence panel now shows the permissions section only on supported platforms and clarifies macOS-only messaging.

* **Platform Support**
  * Better Windows handling for dev tooling and pre-push checks to reduce setup friction.

* **Tests**
  * Multiple tests made cross-platform resilient to path and behavior differences.

* **Chores**
  * Ignore and formatting lists updated to skip a temporary test-target directory.


scripts/run-dev-win.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd -P)"
+APP_DIR="$REPO_ROOT/app"
+cd "$APP_DIR"
+
+export LIBCLANG_PATH="/c/Program Files/LLVM/bin"
```
graycyrus, line 25: [major] `CEF_PATH` is set to `$HOME/Library/Caches/tauri-cef` — a macOS path — inside a Windows dev script. On Windows Git Bash `$HOME` is typically something like `/c/Users/<user>`, but `Library/Caches` doesn't exist there. The equivalent location on Windows is wherever `ensure-tauri-cli.sh` actually stages the CEF runtime. You can either let `ensure-tauri-cli.sh` set `CEF_PATH` (it already exports a default), or probe a Windows-appropriate path like `$LOCALAPPDATA/tauri-cef`. Either way, hard-coding a macOS cache hierarchy into a Windows script is a copy-paste bug.

```bash
# before
export CEF_PATH="$HOME/Library/Caches/tauri-cef"

# after — let ensure-tauri-cli.sh own the default, or use an explicit Windows path
export CEF_PATH="${CEF_PATH:-$(to_unix_path "${LOCALAPPDATA:-}"/tauri-cef)}"
```

scripts/run-dev-win.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+PATH_PREFIX="/c/Program Files/CMake/bin:$(dirname "$NINJA_EXE")"
+if [[ -n "${CEF_RUNTIME_PATH:-}" ]]; then
+  PATH_PREFIX="$PATH_PREFIX:$CEF_RUNTIME_PATH"
+fi
+export PATH="$PATH_PREFIX:$PATH"
+
+"$PNPM_EXE" tauri:ensure
+"$PNPM_EXE" core:stage
```
graycyrus, line 99: [major] The final line hardcodes `APPLE_SIGNING_IDENTITY='OpenHuman Dev Signer'` and calls `cargo tauri dev` bare — not through the vendored `tauri-cef` CLI that CLAUDE.md says is required (`pnpm tauri:ensure` is called above, but the actual invocation still uses whatever `cargo tauri` resolves to from PATH). On macOS `APPLE_SIGNING_IDENTITY` is meaningful; on Windows it is noise that could confuse Tauri'\''s signing checks. The existing `dev:app` npm script handles both of these correctly. Either delegate to it (`"$PNPM_EXE" dev:app`) or document why a raw `cargo tauri dev` is intentional here.

Also note: `source ../scripts/load-dotenv.sh` is called from inside `app/` (the script `cd`s to `APP_DIR` on line 7) but then references `../scripts/` — that works, but only because of the `cd`. If the script is ever refactored to not change directory this will break silently. A safer pattern is to use the already-computed `$REPO_ROOT`:

```bash
source "$REPO_ROOT/scripts/load-dotenv.sh"
```

src/openhuman/tools/impl/system/shell.rs
```diff
@@ -323,72 +324,86 @@ mod tests {
...
     #[tokio::test]
     async fn shell_preserves_path_and_home() {
         let tool = ShellTool::new(test_security_with_env_cmd(), test_runtime());
 
-        let result = tool
-            .execute(json!({"command": "echo $HOME"}))
-            .await
-            .unwrap();
-        assert!(!result.is_error);
+        let home_cmd = if cfg!(windows) {
```
graycyrus: [minor] Several tests are now gated with `#[cfg(not(windows))]` but the surrounding `test_security_with_env_cmd()` fixture still lists only `["env", "echo"]` in `allowed_commands`. The PR also adds `"set"` and `"mkdir"` to the broader `test_security()` fixture, but those additions don'\''t help the `#[cfg(not(windows))]`-gated tests which use the env-cmd fixture. This is harmless right now since those tests won'\''t compile on Windows, but it leaves the fixture inconsistent and will confuse the next person who tries to enable Windows coverage for them. Consider either updating `test_security_with_env_cmd()` to include the Windows equivalents or leaving a comment explaining the intentional split.

src/openhuman/composio/periodic.rs
```diff
@@ -280,10 +280,11 @@ mod tests {
...
         // that real startup uses. If some other test has written a session
         // profile to disk, this test accepts either outcome (Ok) gracefully.
-        let result = run_one_tick().await;
-        // Either Ok (no client, skipped) or Ok (backend unreachable handled
-        // gracefully). The `Err` branch only fires on config-load failure.
-        let _ = result;
+        let result = tokio::time::timeout(Duration::from_secs(15), run_one_tick()).await;
+        assert!(
+            result.is_ok(),
+            "run_one_tick should not hang indefinitely during tests"
```
graycyrus: [minor] The timeout wrapper added to `run_one_tick_returns_ok_when_no_client` is a good idea, but 15 seconds is a long wall-clock budget for a unit test that is supposed to take milliseconds. If the function does hang (e.g. because `load_config_with_timeout` itself has no inner timeout on Windows), CI will wait the full 15 s before failing. The existing `load_config_with_timeout` presumably already has a deadline — if so, 3–5 seconds is plenty for a unit test timeout and gives faster feedback without masking the actual hang source.

src/openhuman/agent/harness/self_healing.rs
```diff
@@ -244,8 +244,9 @@ mod tests {
     fn tool_maker_prompt_includes_command() {
         let interceptor = SelfHealingInterceptor::new(Path::new("/workspace"), true);
         let prompt = interceptor.tool_maker_prompt("jq", "parse json output");
+        let normalized = prompt.replace('\\', "/");
         assert!(prompt.contains("jq"));
```
graycyrus: [minor] The path normalization in `tool_maker_prompt_includes_command` only normalizes `normalized` but the first assertion (`assert!(prompt.contains("jq"))`) still operates on the raw `prompt`. That'\''s fine since "jq" itself has no slashes. The real concern is that the second assertion was switched to `normalized` while the third (`assert!(prompt.contains("parse json output"))`) was left on `prompt`. That asymmetry is fine semantically, but it would be clearer to just normalize once at the top and assert everything on `normalized` — it'\''s cheaper to read and avoids future confusion about which variable is the source of truth.

graycyrus (changes requested):
This PR adds genuine value — Windows dev support and cross-platform test fixes are both worth having. A few issues need attention before merge.

**PR description is boilerplate.** The Summary / Problem / Solution / Impact / Checklist sections were left as template placeholders. Please fill them in — reviewers (and future `git log` readers) need to know why these tests were changed, what broke on Windows, and whether the locking removal in `trigger_history.rs` is safe.

Three substantive issues flagged below, plus a few minor items. The biggest one is the `scripts/run-dev-win.sh` path hardcoding macOS cache directories while targeting Windows — that's a copy-paste bug that will confuse the next person who tries to use the script.

M3gA-Mind (commented):
**Disagree (CodeRabbit `shell.rs` naming):** @coderabbitai suggested renaming `home_cmd`/`path_cmd` to camelCase. Rust naming conventions mandate snake_case for local variables — camelCase would be a compiler warning. The dead `cfg!(windows)` branches inside the `#[cfg(not(windows))]`-gated test have been removed in this PR commit instead, which was the real underlying issue.

**Already addressed (graycyrus `shell.rs` fixture):** The comment noted `test_security_with_env_cmd()` was missing `"set"` and `"mkdir"` — the current code at line 298 already has `allowed_commands: vec!["env".into(), "echo".into(), "set".into(), "mkdir".into()]`. No change needed.

senamakel (approved):

PR #1051, opened 2026-04-30
fix(channels): preserve overflow segments instead of dropping them (#1041)

### LINKED ISSUE #1041: [Bug] UI Not Rendering Full LLM Response (List Items / Markdown Truncated
Description:

The UI is not rendering the complete response returned by the LLM. While the backend/transcript logs confirm that the full response (including list items) is being generated, the frontend only displays content up to a certain point, truncating the rest.

From observed behavior, the issue seems specifically related to list rendering (HTML/Markdown lists) where content after a list introduction line is not displayed.

---

Observed Behavior:

LLM returns full structured response (including bullet points)

Transcript logs confirm full output is present

UI renders only up to:

> “we launched a week ago and the response has been strong:”

Bullet points (- 100+ users, etc.) are not rendered

Remaining content after list is also missing

---

Expected Behavior:

UI should render the entire LLM response

Markdown lists (-, *, numbered) should correctly convert to HTML

No truncation of content after list sections

---

Possible Root Causes (to investigate):

Markdown → HTML parsing issue (e.g. improper list handling)

Renderer breaking on newline or special characters

State truncation in Redux store

Streaming response cut-off / buffer issue

Sanitization stripping list elements (<ul>, <li>)

Improper handling of multi-line content

---

Debug Checklist:

[ ] Check Sentry logs for rendering/runtime errors

[ ] Inspect Redux store state → verify full response is stored

[ ] Check localStorage/sessionStorage for truncated payload

[ ] Log raw API response vs rendered content

[ ] Verify Markdown parser (e.g. marked, remark, etc.)

[ ] Inspect DOM → are <ul>/<li> elements generated or dropped?

[ ] Test with static mock response containing lists

[ ] Check if streaming stops at list boundary

---

Acceptance Criteria:

[ ] Full LLM response is visible in UI without truncation

[ ] Bullet point lists render correctly

[ ] Content after lists is preserved and displayed

[ ] No console/runtime errors during rendering

[ ] Consistent behavior across browsers/devices

[ ] Verified fix using same transcript input

---

Attachments / References:

Screenshot of UI truncation

Transcript logs showing full response

Example failing content (with bullet list)

sanil-23:
## Summary

- Fix issue #1041: long agent replies were missing roughly half their content in the UI (bullet lists + trailing paragraphs disappeared).
- Root cause was in **`segment_for_delivery`** (`src/openhuman/channels/providers/presentation.rs`): `.take(MAX_SEGMENTS)` silently dropped every paragraph past the 5th cap.
- New `cap_segments(segments, max, joiner)` helper preserves all content by merging overflow into the final segment instead of discarding it.
- Pre-existing `MAX_SEGMENTS = 5` cap retained — bubble count is still bounded; no inter-bubble delay regression.
- New regression test runs the exact agent reply text from the issue's transcript screenshot and asserts every bullet and every trailing paragraph survives.

## Problem

The web channel applies a "human-feeling" presentation layer that splits long replies into multiple chat bubbles. The split was implemented as `paragraphs.into_iter().take(MAX_SEGMENTS).collect()` (and the same pattern on the sentence-fallback path).

Once a reply produced more than `MAX_SEGMENTS = 5` natural paragraphs, every paragraph past the cap was silently dropped before the segments were ever published to the frontend. The final `chat_done` event still carried `full_response` with the complete text, but the frontend's `onDone` handler in `ChatRuntimeProvider` skips the full-response branch when `segment_total > 0` (it trusts segments already delivered everything) — so the truncation was unrecoverable on the client side.

The reporter's screenshots show this exactly: an 11-paragraph draft email rendered as the first 5 bubbles, then `2m ago` and end of message. Bullets and the entire sign-off were gone.

## Solution

Replace `.take(MAX_SEGMENTS)` with a `cap_segments` helper that:

- Returns the input unchanged when it already fits under the cap.
- Otherwise keeps the first `max - 1` segments as-is and concatenates the remaining tail into a single trailing segment with the appropriate joiner (`\n\n` for paragraph strategy, `" "` for sentence strategy).
- Treats `max == 0` as a no-op rather than panicking.

Applied at both call sites in `segment_for_delivery`. The `MAX_SEGMENTS` constant and the inter-bubble delay model (`segment_delay`) are unchanged — bubble count is still bounded, the user just no longer loses content past the cap.

The previously-passing test `max_segments_respected` was renamed to `max_segments_respected_without_dropping_content` and now also asserts that **every** input paragraph appears in the joined output — the assertion that would have caught this bug originally.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [`docs/TESTING-STRATEGY.md`](../docs/TESTING-STRATEGY.md#failure-path-requirement) — 4 new tests + 1 enhanced existing test in `presentation_tests.rs`
- [x] N/A: behaviour-only change — coverage matrix unchanged (no new/renamed feature row)
- [x] N/A: behaviour-only change — no new feature IDs
- [x] N/A — no new external network dependencies introduced
- [x] N/A — no release-cut surface touched (purely internal segmentation logic; observable as "long replies render fully" but doesn't change UX surface)
- [x] Linked issue closed via `Closes #1041` below

## Impact

- **Runtime**: web-channel chat delivery only. No other channel paths or domains touched.
- **Performance**: identical to baseline — same bubble count published, same inter-bubble delay, just one of the bubbles is now potentially longer (the tail).
- **Compatibility**: backward-compatible. Existing replies that fit under the cap behave identically. Replies that previously hit the cap now deliver their full content instead of partial.
- **Migration**: none.

## Related

- Closes #1041
- Follow-up PR(s)/TODOs: none planned. The frontend `ChatRuntimeProvider.onDone` fallback (using `event.full_response` when `segment_total > 0` but fewer segments arrived than claimed) was considered as defense-in-depth but isn't needed once the backend is correct.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented content loss when segment limits are exceeded by capping delivered segments and merging any overflow into the final segment so all original content is retained.

* **Tests**
  * Added unit and end-to-end regression tests to verify capped delivery preserves all input content and handles edge cases (including zero max).

### DIFF
--- a/src/openhuman/channels/providers/presentation.rs
+++ b/src/openhuman/channels/providers/presentation.rs
@@ -261,10 +261,23 @@ fn cap_segments(segments: Vec<String>, max: usize, joiner: &str) -> Vec<String> 
     if max == 0 || segments.len() <= max {
         return segments;
     }
+    let original_len = segments.len();
     let mut iter = segments.into_iter();
     let mut result: Vec<String> = (&mut iter).take(max - 1).collect();
     let tail: Vec<String> = iter.collect();
-    result.push(tail.join(joiner));
+    let tail_count = tail.len();
+    let merged = tail.join(joiner);
+    tracing::debug!(
+        target: "presentation",
+        max,
+        original_len,
+        tail_count,
+        tail_len = merged.len(),
+        joiner_len = joiner.len(),
+        "[presentation:segment] merging {} overflow segments into tail",
+        tail_count
+    );
+    result.push(merged);
     result
 }
 
--- a/src/openhuman/channels/providers/presentation.rs
+++ b/src/openhuman/channels/providers/presentation.rs
@@ -187,7 +187,7 @@ fn segment_for_delivery(text: &str) -> Vec<String> {
                 segments = merged.len(),
                 "[presentation:segment] split by paragraphs"
             );
-            return merged.into_iter().take(MAX_SEGMENTS).collect();
+            return cap_segments(merged, MAX_SEGMENTS, "\n\n");
         }
     }
 
@@ -200,7 +200,7 @@ fn segment_for_delivery(text: &str) -> Vec<String> {
                 segments = grouped.len(),
                 "[presentation:segment] split by sentences"
             );
-            return grouped.into_iter().take(MAX_SEGMENTS).collect();
+            return cap_segments(grouped, MAX_SEGMENTS, " ");
         }
     }
 
@@ -247,6 +247,25 @@ fn is_numbered_list_item(line: &str) -> bool {
     }
     // Must have consumed at least one digit, followed by ". ".
     i > 0 && i <= 3 && bytes.get(i) == Some(&b'.') && bytes.get(i + 1) == Some(&b' ')
+}
+
+/// Cap the number of delivered segments at `max` without losing content:
+/// the first `max - 1` segments are kept as-is, and any overflow is
+/// concatenated into a single trailing segment using `joiner`.
+///
+/// The earlier behavior (`.take(MAX_SEGMENTS)`) silently dropped every
+/// segment past the cap, which truncated long agent replies in the UI
+/// (issue #1041). Merging into the tail preserves all content while
+/// still bounding the inter-bubble delay budget.
+fn cap_segments(segments: Vec<String>, max: usize, joiner: &str) -> Vec<String> {
+    if max == 0 || segments.len() <= max {
+        return segments;
+    }
+    let mut iter = segments.into_iter();
+    let mut result: Vec<String> = (&mut iter).take(max - 1).collect();
+    let tail: Vec<String> = iter.collect();
+    result.push(tail.join(joiner));
+    result
 }
 
 /// Merge adjacent segments shorter than MIN_SEGMENT_CHARS.
--- a/src/openhuman/channels/providers/presentation_tests.rs
+++ b/src/openhuman/channels/providers/presentation_tests.rs
@@ -72,10 +72,12 @@ fn max_segments_respected_without_dropping_content() {
     let result = segment_for_delivery(&text);
     assert!(result.len() <= MAX_SEGMENTS);
     let joined = result.join("\n\n");
-    for i in 0..10 {
+    // Assert the full paragraph body survives, not just the prefix —
+    // a mid-text truncation would slip past a substring-only check.
+    for (i, original) in paras.iter().enumerate() {
         assert!(
-            joined.contains(&format!("Paragraph number {} ", i)),
-            "paragraph {} missing from delivered segments",
+            joined.contains(original),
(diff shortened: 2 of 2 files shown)


src/openhuman/channels/providers/presentation.rs
```diff
@@ -187,7 +187,7 @@ fn segment_for_delivery(text: &str) -> Vec<String> {
                 segments = merged.len(),
                 "[presentation:segment] split by paragraphs"
```
graycyrus, line 188: [minor] `segments = merged.len()` logs the pre-cap count, so when `cap_segments` actually fires you'd see `segments=10` followed by `merging 5 overflow segments into tail` — the sequence tells the story, but the first log reads like 10 segments were delivered. Renaming the field to `segments_before_cap` would make the two lines unambiguous when tailing logs.

src/openhuman/channels/providers/presentation.rs
```diff
@@ -249,6 +249,38 @@ fn is_numbered_list_item(line: &str) -> bool {
...
+    let tail: Vec<String> = iter.collect();
+    let tail_count = tail.len();
+    let merged = tail.join(joiner);
+    tracing::debug!(
+        target: "presentation",
+        max,
+        original_len,
+        tail_count,
+        tail_len = merged.len(),
+        joiner_len = joiner.len(),
```
graycyrus, line 276: [minor] `joiner_len = joiner.len()` is always 2 (`"\n\n"`) or 1 (`" "`), both knowable from the two call sites. It does not help distinguish anything at runtime — you already have `target: "presentation"` and the message string identifying the path. Fine to keep for completeness, just noting it does not carry diagnostic weight.

graycyrus (approved):
Clean fix. The `cap_segments` helper is the right shape — bounded bubble count, no content loss, well-documented. The regression test against the actual transcript from the issue is a nice touch; it exercises exactly the code path that was broken rather than just the helper in isolation. Left two minor inline comments on the debug logging, neither blocking.

graycyrus (changes requested):
Two minor items to address before merge — nothing structural, just tightening up the debug logging:

1. **`segments` log field ambiguity** — the `tracing::debug!` on the line just before `cap_segments` logs `segments = merged.len()` which is the pre-cap count. When `cap_segments` then logs its own merge message, a reader sees two different segment counts in the span. Rename to `segments_before_cap` or similar to make the sequence unambiguous.

2. **`joiner_len` in `cap_segments` debug log** — this is always 2 (`"\n\n"`) or 1 (`" "`), fully deterministic from the two call sites. It doesn't carry diagnostic weight. Consider dropping it or replacing with the joiner string itself (`joiner = %joiner`) which is more immediately readable.

Both are quick fixes. The logic itself is solid — nice work on the regression test with the exact transcript.

senamakel (approved):

Issue #963, opened 2026-04-27
Fix GitHub and Twitter OAuth login when credentials are not configured


graycyrus:
## Summary
Fix GitHub and Twitter OAuth login flows which are not configured properly.

## Problem
- GitHub shows a redirect_uri mismatch / app association error.
- Twitter returns `Twitter OAuth credentials not configured`.

## Scope
- Reproduce GitHub OAuth login failure end-to-end.
- Fix GitHub OAuth configuration:
  - callback URL / redirect URI
  - app registration settings
  - frontend/backend callback routing
  - packaged app vs browser redirect behavior
- Reproduce Twitter login failure end-to-end.
- Configure/fix Twitter OAuth credentials and callback flow.
- Add clear error handling/logs for OAuth misconfiguration.

## Acceptance Criteria
- GitHub login works successfully in the target environment.
- Twitter login works successfully in the target environment.
- OAuth failures show actionable logs/errors for developers.
- No regressions to session/auth state after successful OAuth login.

_Split from #961._

Issue #1068, opened 2026-05-01
[Bug] dev:app restartApp() kills vite parent → entire dev session exits


oxoxDev:
## Summary

In `pnpm dev:app`, calling `restartApp()` (from `app/src/utils/tauriCommands`) terminates the entire dev session — vite dies along with the cargo dev parent and the desktop window exits. The user has to manually re-run `pnpm dev:app` to recover.

## Reproduce

1. `pnpm dev:app` from a clean openhuman worktree.
2. After app boots, trigger any code path that calls `restartApp()` — easiest is `handleIdentityFlip` in `app/src/providers/CoreStateProvider.tsx:121`, which fires automatically when the persisted `OPENHUMAN_ACTIVE_USER_ID` differs from the `auth.userId` returned by the bootstrap snapshot.
3. App exits. Vite log from `bft71fnsj` confirms `[app] restart_app invoked from frontend → RunEvent::ExitRequested → cef::shutdown follows`. Vite parent dies with the cargo process.

## Impact

- Identity-flip recovery (the dominant `restartApp()` caller) is broken in dev. Multi-account testing requires constant manual relaunches.
- Smoke validation against `feat/*` branches is painful — every restart costs a 30-60s rebuild.
- Logs noted: stale-data + activeUserId mismatch on first boot triggers an immediate flip → restart loop until the persisted `OPENHUMAN_ACTIVE_USER_ID` matches the bootstrap snapshot.

## Root cause hypothesis

`restartApp()` invokes Tauri's `app.restart()`. In a packaged `.app`, this re-execs the app binary and the new process re-attaches a fresh webview. In `pnpm dev:app`, the launcher graph is `pnpm tauri dev → cargo run → tauri-cef CLI → vite (child)`. `app.restart()` exits the cargo parent, which orphans / kills vite. Tauri then has nothing to relaunch the webview against.

## Proposed fix

In dev mode, replace `app.restart()` with a webview reload (`window.location.reload()`):
- Accomplishes the same goal — module-init re-runs, `OPENHUMAN_ACTIVE_USER_ID` localStorage seed picks up the new userId, redux-persist re-hydrates from the new namespace.
- Doesn't touch the cargo / vite parent processes.
- Singleton services on the Rust side still need to disconnect / reconnect cleanly — verify `socketService.disconnect()` already covers this.

[...]

Issue #964, opened 2026-04-27
Fix Google OAuth callback URL / redirect URI configuration


graycyrus:
## Summary
Fix Google OAuth login — the callback URL / redirect URI is not properly configured.

## Scope
- Reproduce Google OAuth callback failure end-to-end.
- Fix Google OAuth redirect URI configuration (app registration, backend routing, packaged app behavior).
- Verify Google login works end-to-end after fix.

## Acceptance Criteria
- Google login works successfully in the target environment.
- OAuth failures show actionable logs/errors for developers.
- No regressions to session/auth state after successful OAuth login.

_Split from #961._

PR #1069, opened 2026-05-01
fix(welcome): re-enable OAuth buttons with focus/timeout recovery (#1049)

### LINKED ISSUE #1049: Re-enable welcome screen OAuth buttons after auth fixes land
## Summary

The welcome screen OAuth buttons (Google, GitHub, Twitter) were intentionally disabled in #1033 because clicking them got stuck in a permanent "Connecting..." state with no way to recover. The buttons now render but do nothing (`onClickOverride` wired to a no-op).

## Root cause

The underlying auth issues are still open:
- #963 — GitHub and Twitter OAuth credentials not configured
- #964 — Google OAuth callback URL / redirect URI misconfigured

## What needs to happen

1. Fix #963 and #964 so OAuth providers actually work
2. Remove the `onClickOverride` no-op in `app/src/pages/Welcome.tsx` so the buttons trigger the real auth flow again
3. Review whether the "Configure RPC URL (Advanced)" link added in #948 should remain visible to all users on the welcome screen — feels like a dev-only concern

## Related

- #1033 — PR that disabled the buttons
- #948 — PR that added the RPC URL config to the welcome screen
- #968 — Auth multi-provider edge cases (closed)

oxoxDev:
## Summary

- Re-enable the welcome screen OAuth buttons (Google / GitHub / Twitter) by dropping the inert `onClickOverride` no-op added in #1033.
- Add an `isLoading` recovery path in `OAuthProviderButton` so the "Connecting..." state can no longer get permanently stuck.
- Tests cover focus reset, 90s safety timeout, override short-circuit, double-click safety, and live-click delegation from `Welcome`.

## Problem

The welcome OAuth buttons were intentionally disabled in #1033 because clicking them got stuck in a permanent "Connecting..." state. The flow is:

1. Button click → `setIsLoading(true)` → `openUrl(loginUrl)` opens system browser.
2. Backend OAuth → `openhuman://auth?token=...` deep link returns to the app and `useDeepLinkAuthState.isProcessing` flips on.

If the deep link never returns (user cancels in the browser, GitHub/Twitter creds missing per #963, Google redirect URI broken per #964), the local `isLoading` state never resets — button frozen until manual refresh. #1033 chose to short-circuit clicks with a no-op handler as a stopgap.

The underlying backend issues live in `tinyhumansai/backend` (#963, #964) and are out of scope here, but the UI freeze is fixable in this repo.

## Solution

`OAuthProviderButton.tsx` — add a `useEffect` that runs while `isLoading=true`:

- 90s safety timeout that resets `isLoading=false`.
- `window.addEventListener('focus', ...)` listener that resets immediately when the user returns from the system browser.

Both signals clean up on unmount, on the `isLoading=false` transition, and when the next click queues a new pair.

`Welcome.tsx` — drop the `handleDisabledOAuthClick` no-op and the `onClickOverride={...}` prop so clicks go through the real handler again. The `onClickOverride` prop on `OAuthProviderButton` itself is preserved (other surfaces may want it).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [`docs/TESTING-STRATEGY.md`](../docs/TESTING-STRATEGY.md#failure-path-requirement)
- [ ] N/A: Coverage matrix updated — no welcome OAuth feature row in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md), behaviour-only change to existing surface
- [ ] N/A: All affected feature IDs from the matrix are listed in the PR description under `## Related` — no matrix rows touched
- [x] No new external network dependencies introduced (mock backend used per [`docs/TESTING-STRATEGY.md`](../docs/TESTING-STRATEGY.md#mock-policy))
- [ ] N/A: Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — welcome OAuth not on the release smoke list

## Impact

- **Runtime**: desktop only (Tauri webview). The recovery is wrapped in `useEffect` with proper cleanup; no global listener leaks.
- **Backend**: Google should work as long as #964 (redirect URI) is configured on the backend side. GitHub / Twitter still surface backend errors until #963 lands — but the UI now recovers within ~1-2s via the focus listener instead of staying frozen.
- **Performance**: one `setTimeout` + one `focus` listener per click, both removed on reset. Negligible.
- **Security**: no change. Same backend OAuth URL, same deep-link return path.

## Related

- Closes #1049
- Related: #963 (backend GitHub/Twitter creds), #964 (backend Google redirect URI), #1033 (PR that added the inert override), #1068 (dev:app `restartApp()` regression — surfaced during smoke; tracked separately)

## Summary by CodeRabbit

* **New Features**
  * Welcome screen now initiates the real OAuth flow (system browser → backend → return to app).

* **Bug Fixes**
  * OAuth button recovers from stalled auth: resets on app resume (focus/visibility) and after a 90s safety timeout; avoids premature reset during deep-link processing.

* **Tests**
  * Added coverage for OAuth button behavior, recovery logic, override suppression, and deduplicated click handling.

### DIFF
--- a/app/src/components/oauth/OAuthProviderButton.tsx
+++ b/app/src/components/oauth/OAuthProviderButton.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { getBackendUrl } from '../../services/backendUrl';
 import type { OAuthProviderConfig } from '../../types/oauth';
@@ -13,6 +13,11 @@ interface OAuthProviderButtonProps {
   onClickOverride?: () => void;
 }
 
+// Reset the loading state if the OAuth round-trip never completes — covers
+// the case where the user cancels in the system browser, or the backend
+// redirect fails so the `openhuman://` deep link never fires.
+const OAUTH_LOADING_TIMEOUT_MS = 90_000;
+
 const OAuthProviderButton = ({
   provider,
   className = '',
@@ -20,6 +25,32 @@ const OAuthProviderButton = ({
   onClickOverride,
 }: OAuthProviderButtonProps) => {
   const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    if (!isLoading) return;
+
+    const reset = () => setIsLoading(false);
+
+    // Window-focus reset is the fast path: when the user returns to OpenHuman
+    // after the system browser, the loading state lifts immediately so the
+    // button is clickable again.
+    const handleFocus = () => {
+      console.debug(`[oauth-button][${provider.id}] window focus → reset isLoading`);
+      reset();
+    };
+
+    const timer = window.setTimeout(() => {
+      console.debug(`[oauth-button][${provider.id}] timeout → reset isLoading`);
+      reset();
+    }, OAUTH_LOADING_TIMEOUT_MS);
+
+    window.addEventListener('focus', handleFocus);
+
+    return () => {
+      window.clearTimeout(timer);
+      window.removeEventListener('focus', handleFocus);
+    };
+  }, [isLoading, provider.id]);
 
   const handleOAuthLogin = async () => {
     if (onClickOverride) {
--- a/app/src/components/oauth/OAuthProviderButton.tsx
+++ b/app/src/components/oauth/OAuthProviderButton.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
 
 import { getBackendUrl } from '../../services/backendUrl';
+import { getDeepLinkAuthState } from '../../store/deepLinkAuthState';
 import type { OAuthProviderConfig } from '../../types/oauth';
 import { IS_DEV } from '../../utils/config';
 import { openUrl } from '../../utils/openUrl';
@@ -31,11 +32,32 @@ const OAuthProviderButton = ({
 
     const reset = () => setIsLoading(false);
 
-    // Window-focus reset is the fast path: when the user returns to OpenHuman
-    // after the system browser, the loading state lifts immediately so the
-    // button is clickable again.
+    // Skip reset when a deep-link auth round-trip is already in flight — the
+    // OAuth callback flips `isProcessing=true` AFTER the OS focus event fires,
+    // and resetting first would briefly re-enable the button mid-redirect.
+    const skipDuringDeepLink = (label: string) => {
+      if (getDeepLinkAuthState().isProcessing) {
+        console.debug(`[oauth-button][${provider.id}] ${label} — skip (deep-link processing)`);
+        return true;
+      }
+      return false;
+    };
+
+    // Fast path: window focus fires when the user returns from the system
+    // browser. On most platforms this lifts the loading state immediately.
     const handleFocus = () => {
+      if (skipDuringDeepLink('focus')) return;
       console.debug(`[oauth-button][${provider.id}] window focus → reset isLoading`);
+      reset();
+    };
+
+    // Backup path: macOS Spaces / virtual desktops sometimes restore window
+    // focus without firing a `focus` event. `visibilitychange` is the more
+    // reliable signal there.
+    const handleVisibilityChange = () => {
+      if (document.visibilityState !== 'visible') return;
+      if (skipDuringDeepLink('visibilitychange')) return;
+      console.debug(`[oauth-button][${provider.id}] visibilitychange visible → reset isLoading`);
       reset();
     };
 
@@ -45,10 +67,12 @@ const OAuthProviderButton = ({
     }, OAUTH_LOADING_TIMEOUT_MS);
 
     window.addEventListener('focus', handleFocus);
+    document.addEventListener('visibilitychange', handleVisibilityChange);
 
     return () => {
       window.clearTimeout(timer);
       window.removeEventListener('focus', handleFocus);
+      document.removeEventListener('visibilitychange', handleVisibilityChange);
     };
   }, [isLoading, provider.id]);
 
@@ -60,7 +84,7 @@ const OAuthProviderButton = ({
 
     if (externalDisabled || isLoading) return;
 
-    console.log(`Starting ${provider.name} OAuth login`, isTauri());
+    console.debug(`[oauth-button][${provider.id}] starting OAuth login (isTauri=${isTauri()})`);
 
     setIsLoading(true);
 
--- a/app/src/components/oauth/__tests__/OAuthProviderButton.test.tsx
+++ b/app/src/components/oauth/__tests__/OAuthProviderButton.test.tsx
@@ -1,0 +1,125 @@
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { getBackendUrl } from '../../../services/backendUrl';
+import { openUrl } from '../../../utils/openUrl';
+import { isTauri } from '../../../utils/tauriCommands';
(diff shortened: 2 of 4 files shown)


app/src/components/oauth/OAuthProviderButton.tsx
```diff
@@ -21,6 +26,32 @@ const OAuthProviderButton = ({
 }: OAuthProviderButtonProps) => {
   const [isLoading, setIsLoading] = useState(false);
 
+  useEffect(() => {
+    if (!isLoading) return;
+
+    const reset = () => setIsLoading(false);
+
+    // Window-focus reset is the fast path: when the user returns to OpenHuman
+    // after the system browser, the loading state lifts immediately so the
```
graycyrus: [major] The focus listener fires on *every* window focus, including the one triggered by a successful OAuth deep-link callback. In the happy path the sequence is:

1. User clicks → `isLoading = true`, focus listener registered.
2. System browser opens, user authenticates.
3. Backend redirects to `openhuman://auth?token=…`, Tauri brings the app to the foreground.
4. **`focus` fires → `isLoading = false`** (this effect runs first).
5. `beginDeepLinkAuthProcessing()` fires → `isProcessing = true` in `deepLinkAuthState`.

Steps 4 and 5 are a race. The deep-link handler is async so there is a render cycle where the button re-enables and is clickable again, right before the processing spinner takes over. If the user clicks in that window a second OAuth flow starts in parallel.

The simplest guard: check `isProcessing` from `deepLinkAuthState` before resetting:

```tsx
// before
const handleFocus = () => {
  console.debug(`[oauth-button][${provider.id}] window focus → reset isLoading`);
  reset();
};

// after — import { getDeepLinkAuthState } from '../../store/deepLinkAuthState' at top
const handleFocus = () => {
  if (getDeepLinkAuthState().isProcessing) return; // deep-link already in flight
  console.debug(`[oauth-button][${provider.id}] window focus → reset isLoading`);
  reset();
};
```

`getDeepLinkAuthState()` is already exported from `../../store/deepLinkAuthState` — no new dependency needed.

app/src/components/oauth/OAuthProviderButton.tsx
```diff
@@ -21,6 +26,32 @@ const OAuthProviderButton = ({
 }: OAuthProviderButtonProps) => {
   const [isLoading, setIsLoading] = useState(false);
```
graycyrus, line 28: [minor] On macOS with multiple Spaces or virtual desktops, switching away and back sometimes does not fire a `window` focus event — the OS restores keyboard focus to the already-focused window without emitting a new focus event. `visibilitychange` is more reliable for the fast-path recovery:

```tsx
const handleVisibilityChange = () => {
  if (document.visibilityState === 'visible') {
    console.debug(`[oauth-button][${provider.id}] visibilitychange visible → reset isLoading`);
    reset();
  }
};
document.addEventListener('visibilitychange', handleVisibilityChange);
// and in cleanup:
document.removeEventListener('visibilitychange', handleVisibilityChange);
```

Not a blocker since the 90 s timeout backstops it, but on macOS (the primary target) this improves fast-path reliability.

app/src/components/oauth/OAuthProviderButton.tsx
```diff
@@ -21,6 +26,32 @@ const OAuthProviderButton = ({
...
 
+  useEffect(() => {
+    if (!isLoading) return;
+
+    const reset = () => setIsLoading(false);
```
graycyrus, line 33: [minor] Pre-existing `console.log` without a grep-friendly prefix — the PR adds nicely prefixed `[oauth-button][…]` logging in the new effect but leaves the entry log inconsistent. Per CLAUDE.md, new/changed flows should use stable prefixes:

```tsx
// before
console.log(`Starting ${provider.name} OAuth login`, isTauri());

// after
console.debug(`[oauth-button][${provider.id}] starting OAuth login (isTauri=${isTauri()})`);
```

Also swap to `console.debug` — this is a flow-trace entry, not a user-visible warning.

app/src/components/oauth/__tests__/OAuthProviderButton.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { isTauri } from '../../../utils/tauriCommands';
+import OAuthProviderButton from '../OAuthProviderButton';
+
+vi.mock('../../../services/backendUrl', () => ({ getBackendUrl: vi.fn() }));
+
+vi.mock('../../../utils/openUrl', () => ({ openUrl: vi.fn() }));
+
+vi.mock('../../../utils/tauriCommands', () => ({ isTauri: vi.fn() }));
+
+vi.mock('../../../utils/config', () => ({ IS_DEV: false }));
```
graycyrus: [major] CodeRabbit flagged this too — I agree. The module factory `vi.mock('../../../utils/config', () => ({ IS_DEV: false }))` replaces the entire module object. If `config.ts` ever exports additional values that other imports or the test setup rely on, those silently become `undefined` in this file, producing confusing failures when the module grows.

Since `IS_DEV` is `false` in the Vitest environment by default (no `VITE_` vars injected), the mock is a behavioral no-op anyway. Drop it and leave a comment explaining why the non-dev URL shape is expected:

```tsx
// Remove this line:
vi.mock('../../../utils/config', () => ({ IS_DEV: false }));

// In the URL assertion, the non-dev shape is already the default:
expect(openUrl).toHaveBeenCalledWith('https://backend.test/auth/google/login');
// IS_DEV is false in CI — no ?responseType=json suffix expected
```

If you need to test the dev-mode branch in future, use `vi.spyOn` on the specific export rather than replacing the module.

graycyrus (changes requested):
Good recovery mechanism overall — the focus+timeout dual-path approach is sound and the test coverage is genuinely thorough. A couple of behavioral edge cases worth discussing before merging.

CodeRabbit flagged the `vi.mock('../../../utils/config', () => ({ IS_DEV: false }))` at line 15 of `OAuthProviderButton.test.tsx` — that finding is correct and I agree with it; see the inline comment for the fix.

oxoxDev:
Addressed review:

- Dropped `vi.mock('../../../utils/config', () => ({ IS_DEV: false }))` per CodeRabbit + graycyrus; URL assertion now tolerates the optional `?responseType=json` query. Fixed in f86b02f0.
- Guarded focus / visibilitychange recovery with `getDeepLinkAuthState().isProcessing` so the button stays in "Connecting..." across the deep-link return race. Fixed in 50bd52d4.
- Added `visibilitychange` listener for macOS Spaces fast-path reliability (with the same guard). 50bd52d4.
- Replaced bare `console.log` entry with grep-friendly `[oauth-button][${provider.id}] starting OAuth login`. 50bd52d4.

PTAL @graycyrus.

oxoxDev (commented):
@graycyrus all 4 review points addressed:

- **Deep-link race guard** (line 35): focus + visibilitychange handlers now early-return if `getDeepLinkAuthState().isProcessing` — no premature reset during OAuth callback. → `50bd52d4`
- **visibilitychange listener** (line 28): added on `document` for macOS Spaces fallback; same guard. → `50bd52d4`
- **Grep-friendly entry log** (line 33): `console.log` → `console.debug('[oauth-button][${provider.id}] starting OAuth login (isTauri=...)')`. → `50bd52d4`
- **Dropped `vi.mock('../../../utils/config', ...)`** + URL regex matcher; added 2 new tests (deep-link race + visibilitychange path). → `f86b02f0`

Local: 11/11 oauth tests pass. PTAL.

graycyrus (approved):

PR #1314, opened 2026-05-06
perf(agent): orchestrator harness efficiency improvements

### LINKED ISSUE #1141: Audit orchestrator harness efficiency and reduce unnecessary delegation/token waste
[Feature] Audit and improve orchestrator harness efficiency so it wastes fewer tokens and delegates only when it truly helps

## Summary

Run a focused orchestrator-harness improvement track to reduce token waste, tighten delegation behavior, and learn from other strong agent harness patterns before we keep adding more orchestration complexity.

This should answer questions like:
- when is the orchestrator better off acting directly?
- when does spawning a subagent improve quality enough to justify the extra prompt/tool/context overhead?
- where are we losing cacheability, repeating context, or paying for unnecessary prompt surface?
- which ideas from other harnesses should we adopt in our own loop?

## Problem

The current orchestrator harness already has several mechanisms meant to control token burn, but they do not yet add up to a clearly evaluated, intentionally efficient orchestration strategy.

What exists today:
- the orchestrator is already described as direct-first in `src/openhuman/tools/orchestrator_tools.rs`
- its agent definition still exposes a large mixed surface of direct tools plus delegation tools in `src/openhuman/agent/agents/orchestrator/agent.toml`
- the harness already has context controls like `tool_result_budget_bytes`, microcompact, autocompact, session-memory extraction, and optional summarizer model selection in `src/openhuman/context/**` and `src/openhuman/config/schema/context.rs`
- subagents already run in a cheaper typed mode with filtered tools in `src/openhuman/agent/harness/subagent_runner/ops.rs`
- oversized payload summarization exists conceptually, but the payload summarizer is currently disabled by default because of recursive dispatch concerns (`summarizer_payload_threshold_tokens = 0`)

What is missing:
- we do not yet have a clear, eval-backed answer for when subagent spawning is actually worth the cost
- the orchestrator may still be paying for delegation overhead in cases where direct execution or a simpler workflow would be cheaper and just as good
- we do not appear to have a formal benchmark for token/latency/quality tradeoffs across direct-first, delegated, and hybrid strategies
- there is no explicit comparison against patterns used in other effective harnesses
- the harness may still be leaking efficiency through large tool surfaces, repeated prompt material, weak cache stability, or unnecessary multi-step delegation

This matters because orchestration complexity is expensive. Every extra spawn, prompt surface, and context handoff can improve quality, but it can also waste tokens, increase latency, and make behavior harder to reason about. If spawning is only sometimes valuable, the harness should know that and route accordingly instead of paying that cost by default.

## Solution (optional)

Treat this as an orchestrator-harness evaluation and redesign issue, not just a small prompt tweak.

High-level scope:
- measure current orchestrator behavior on realistic tasks
- identify where tokens are being burned unnecessarily
- evaluate direct execution vs delegated execution vs simpler coded workflows
- adopt the best ideas from other harnesses only where they improve our actual results

Areas to investigate:
- delegation heuristics
  - when should the orchestrator answer directly?
  - when should it call a direct tool?
  - when should it spawn a specialist?
  - when should a deterministic workflow beat an autonomous delegate?
- tool surface size and discoverability
  - whether the orchestrator sees too many tools/subagents at once
  - whether some tools should move behind narrower routing layers
- prompt and cache stability
  - whether stable prefixes are being preserved well enough for prefix/prompt caching
  - whether volatile prompt sections are forcing avoidable cache misses
- context reduction strategy
  - whether tool-result truncation, microcompact, autocompact, and session-memory extraction are firing at the right times
  - whether large tool outputs should be summarized differently
- subagent cost model
  - whether typed-mode spawning is cheap enough to justify current usage
  - whether some current spawn paths should be replaced with direct-first heuristics or fixed workflows
- long-running task strategy
  - whether the harness should borrow stronger artifact-handoff / incremental-progress patterns from other agent harnesses instead of repeatedly paying to rediscover state

Expected outcome:
- a concrete orchestrator strategy that is more intentional about when to delegate
- fewer wasted tokens
- better latency/cost without degrading quality
- a written decision on where spawning remains valuable and where it should be reduced or avoided

Implementation note:
- this issue should explicitly compare our current harness against strong external patterns rather than optimizing in a vacuum. Relevant inspiration includes official guidance around simpler workflow-first architectures, context engineering, prompt/prefix caching, and tool efficiency from current OpenAI and Anthropic materials.
- the goal is not to copy another harness wholesale. The goal is to benchmark and selectively adapt what actually improves OpenHuman's orchestrator.

## Acceptance criteria

- [ ] We have an explicit audit of current orchestrator token/latency behavior on representative tasks.
- [ ] The audit compares at least direct execution, delegated execution, and simpler workflow alternatives where relevant.
- [ ] The team has a documented answer for when subagent spawning is worth its cost and when it is not.
- [ ] The orchestrator's delegation strategy is updated based on evals rather than intuition alone.
- [ ] Obvious sources of avoidable token waste in prompt assembly, tool surface, or context replay are identified and reduced.
- [ ] Prefix/prompt cache friendliness is part of the evaluation, not an afterthought.
- [ ] The issue results in concrete harness changes or a follow-up implementation plan with prioritized work items.
- [ ] External harness inspiration is captured as actionable decisions, not just a vague reading list.
- [ ] Tests and/or eval coverage are added for the chosen orchestration strategy so regressions are visible.

- [ ] The final design is documented so contributors understand the orchestrator's direct-vs-delegate policy and the reasoning behind it.

## Related

Local code:
- `src/openhuman/agent/agents/orchestrator/agent.toml`
- `src/openhuman/tools/orchestrator_tools.rs`
- `src/openhuman/agent/harness/subagent_runner/ops.rs`
- `src/openhuman/agent/harness/session/builder.rs`
- `src/openhuman/agent/harness/payload_summarizer.rs`
- `src/openhuman/context/manager.rs`
- `src/openhuman/config/schema/context.rs`

External inspiration:
- Anthropic — Building Effective AI Agents: https://www.anthropic.com/engineering/building-effective-agents
- Anthropic — Effective context engineering for AI agents: https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents
- Anthropic — Effective harnesses for long-running agents: https://www.anthropic.com/engineering/effective-harnesses-for-long-running-agents
- Anthropic — Writing effective tools for AI agents: https://www.anthropic.com/engineering/writing-tools-for-agents
- OpenAI — Prompt caching: https://platform.openai.com/docs/guides/prompt-caching
- OpenAI — Responses API / function-calling guidance: https://platform.openai.com/docs/guides/function-calling/tool-use-overview

M3gA-Mind:
## Summary

Implements the changes from the orchestrator harness efficiency audit (#1141) to reduce token waste, tighten delegation behavior, and improve sub-agent context efficiency.

- **Re-enable payload summarizer** at 4,000-token threshold (was disabled after recursive dispatch; root cause was `omit_skills_catalog = false` on the summarizer archetype — fixed by the existing flag)
- **Skip `render_tools()` for native function-calling**: provider already sends JSON schemas in the API payload — the system-prompt tool catalogue was pure duplicate tokens (~1,500–3,000 tokens/turn savings on native sessions)
- **Consolidate 6 `memory_tree_*` tools → 1 `memory_tree` with `mode` enum**: shrinks orchestrator's tool surface by 5 entries per API call without losing any capability
- **Remove `schedule` + `spawn_subagent` from orchestrator**: `cron_add/cron_list/cron_remove` are the canonical scheduling surface; `delegate_*` + `spawn_worker_thread` replace `spawn_subagent`
- **Remove redundant tables from orchestrator prompt**: "Available Sub-Agents" and "Direct Tools" tables removed — LLM sees these as native API tool schemas already
- **Compact `render_delegation_guide`**: single line per toolkit instead of verbose multi-line spawn_subagent snippet
- **Add `max_result_chars` cap**: 8k for planner/researcher, 16k for code_executor — prevents context flooding from large sub-agent outputs
- **Wrap `memory_context` in `Arc<>`**: `ParentExecutionContext::memory_context` is now `Arc<Option<String>>` — spawning a sub-agent shares the allocation instead of cloning potentially large strings
- **`docs/DELEGATION_POLICY.md`**: codifies the 4-tier direct-first decision tree

## Test plan

- [x] `cargo check -p openhuman` — clean (0 errors)
- [x] `cargo clippy -p openhuman` — clean (0 errors)
- [x] `cargo fmt --all --check` — clean
- [x] `cargo test -p openhuman --lib` — 5,982 pass, 9 pre-existing flaky tests (all pass in isolation: network/port-conflict dependent, unrelated to this change)
- [x] `NODE_ENV=test pnpm --filter openhuman-app compile` — clean
- [x] `NODE_ENV=test pnpm --filter openhuman-app lint` — 0 errors (32 pre-existing warnings)
- [x] `NODE_ENV=test pnpm --filter openhuman-app format:check` — clean
- [x] Unit tests added: `ToolsSection` native-skip, `MemoryTreeTool` dispatcher (5 tests), `max_result_chars` truncation (2 tests), delegation guide compact format (1 test)
- [x] All 7 orchestrator `prompt.rs` tests pass
- [x] All 4 `tools_section` tests pass (including 2 new Native/PFormat variants)

Closes #1141

## Summary by CodeRabbit

* **Documentation**
  * Added a delegation policy guide and updated orchestrator delegation and scheduling guidance.

* **New Features**
  * Spawn dedicated worker threads for long-running tasks (returns compact worker references).
  * Cron scheduling: add/list/remove jobs and current-time queries.
  * Unified memory-tree tool with mode-based queries.
  * Per-agent output character limits with truncation indicator.

* **Improvements**
  * Re-enabled payload summarization threshold and refined tool scoping to hide unused delegation tools.

* **Tests**
  * Added/updated tests covering truncation, memory-tree behavior, and tool rendering.


senamakel (changes requested):
resolve coderabbit issues and checks

M3gA-Mind:
Yes..Working on those itself.

M3gA-Mind (commented):
pr-manager: all actionable CodeRabbit comments addressed in 5 commits (b47a984..97e5c91).

**Deferred (low value — Standards):** @coderabbitai suggested extracting the `#[cfg(test)]` modules from `src/openhuman/channels/runtime/dispatch.rs` into a sibling `dispatch_tests.rs` to keep the file under the ~500-line guideline. The file is currently ~1,425 lines. This is a valid long-term concern but is a pure refactor with no behavioural change and no CI impact. The right moment to do this split is when the file is next substantially modified. Deciding factor: do it then, not now.

senamakel:
@sanil-23 pls review

src/openhuman/agent/agents/orchestrator/prompt.md
```diff
@@ -21,52 +21,17 @@ Follow this sequence for every user message:
...
-   - If web/doc crawling is required, delegate to `researcher`.
-   - If complex multi-step decomposition is required, delegate to `planner` (and only then route deeper if necessary).
-   - If code review is requested, delegate to `critic`.
+   - If external SaaS integration work is required, use `delegate_{toolkit}` (e.g. `delegate_gmail`, `delegate_notion`).
+   - If code writing/execution/debugging is required, use `delegate_run_code`.
+   - If web/doc crawling is required, use `delegate_researcher`.
+   - If complex multi-step decomposition is required, use `delegate_plan`.
+   - If code review is requested, use `delegate_critic`.
+   - If memory archiving or distillation is required, use `delegate_archivist`.
 4. **After delegation**, summarise results clearly and concisely.
```
sanil-23, line 30: **Drift risk: archetype names in this prose can diverge from the synthesised `delegate_*` tools.**

Step 3 here lists `delegate_run_code`, `delegate_researcher`, `delegate_plan`, `delegate_critic`, `delegate_archivist` as hardcoded markdown. The summary line below repeats them. Meanwhile the source of truth for *which `delegate_*` tools actually exist* is `agents/orchestrator/agent.toml`'s `subagents = [...]` field, expanded into tool schemas by `collect_orchestrator_tools`.

If anyone later removes (say) `archivist` from `subagents`, the synthesised `delegate_archivist` tool disappears from the API tool schema, but this prompt still tells the LLM to "use `delegate_archivist` for memory writes." The model attempts to call a tool that doesn't exist.

This PR did the right thing for **integrations** — `render_delegation_guide` derives the `## Connected Integrations` block from runtime data, so prompt and schema cannot drift. But **archetypes** still go through hardcoded prose.

Two options:

1. **Make `render_delegation_guide` also iterate `definition.subagents` archetype entries** and emit one line per archetype using each target's `when_to_use`. Then archetypes and integrations share one drift-free source — same guarantee this PR already gives integrations.
2. **Cheap fallback**: add a startup test that asserts every `delegate_<archetype>` literal mentioned in `prompt.md` appears in the synthesised orchestrator tool set. Catches drift in CI without restructuring the prompt.

Either is acceptable; option 1 is the principled fix.

src/openhuman/agent/agents/orchestrator/prompt.rs
```diff
@@ -62,36 +64,49 @@ pub fn build(ctx: &PromptContext<'_>) -> Result<String> {
...
-        "## Delegation Guide — Integrations\n\n\
-         For any task that touches one of these external services, \
-         delegate to `integrations_agent` with the matching `toolkit` \
-         argument. The sub-agent receives the full action catalogue \
-         for that integration as native tool schemas — do not attempt \
-         to call integration actions directly from this agent.\n\n\
-         Only the integrations listed below are currently authorised. \
-         If the user asks about another service, tell them to connect \
-         it in **Skills** page before retrying.\n\n",
+        "## Connected Integrations\n\n\
```
sanil-23, line 92: **Question for you: now that each connected toolkit gets its own `delegate_{toolkit}` tool with auto-generated description, is the per-bullet `## Connected Integrations` block still pulling its weight?**

Each connected toolkit is described in **two** places after this PR:

1. The `delegate_{toolkit}` tool's schema description in `orchestrator_tools.rs:142-150` (sent to the API on every request)
2. The bullet line this function renders into the system prompt

Both carry the toolkit name, the tool to call, and the description. The only unique signal this prompt block adds is the naming-pattern hint at the top (`Delegate tasks for these services using the matching delegate_{toolkit} tool`).

The rest of this PR's token wins (skipping `render_tools()` for native, dropping the "Available Sub-Agents" + "Direct Tools" tables) are explicitly motivated by *don't repeat the schema in prose*. The same logic applies here, but this PR keeps the prose listing.

Two coherent shapes — **picking either is fine; picking both (this PR's current state) is the worst of both worlds:**

**Option A — Drop the per-toolkit list, keep just the naming-pattern hint.**  
Replace this entire function's output with one line in the rules section:
> *External integrations are exposed as `delegate_{toolkit}` tools — only currently-connected toolkits are present in your tool list.*

Saves ~30–50 prompt tokens per connected integration on every cache miss. Loses nothing the schema doesn't already encode. Consistent with Wins 2 and 5 of this PR.

**Option B — Drop the per-toolkit `delegate_*` tools, go back to one generic `spawn_subagent`-style tool with a `toolkit` parameter, and keep the prompt list.**  
Cheaper in *tool-schema* bytes (one schema instead of N), but loses the tool-selection-accuracy win that justifies Step 4 of this PR. Effectively reverses part of the refactor.

Which way did you intend? If A, this function shrinks to a single static line. If B, the synthesis logic in `orchestrator_tools.rs` simplifies.

PR #1444, opened 2026-05-09
docs(core): simplify remote core RPC token setup (#1438)

### LINKED ISSUE #1438: Simplify remote core RPC token setup for Docker, servers, and clients
### Summary

Make **core RPC bearer authentication** trivial when the core runs **remotely** (Docker, VPS, App Platform) and the **desktop or web client** connects to it. Today the operator must keep `OPENHUMAN_CORE_TOKEN` in sync between server env and client config (`app/.env.local` / env vars); the standalone path may instead write `core.token` under the workspace when `OPENHUMAN_CORE_TOKEN` is unset ([`src/core/auth.rs`](https://github.com/tinyhumansai/openhuman/blob/main/src/core/auth.rs)). That split confuses people ("where do I read the token?", "why 401 after deploy?").

### Problem

- **Server**: Docker / compose examples ask you to set `OPENHUMAN_CORE_TOKEN` manually (`openssl rand -hex 32`) per [cloud deploy](https://github.com/tinyhumansai/openhuman/blob/main/gitbooks/features/cloud-deploy.md). If you forget or rotate only one side, every RPC call 401s with little affordance.
- **Client**: The UI has some awareness (e.g. `BootCheckGate` mentions `OPENHUMAN_CORE_TOKEN`), but there is no guided **pairing** flow: paste server token once, validate with a ping, persist securely.
- **Two semantics**: Env-injected token (Tauri / cloud) vs generated `core.token` file (standalone CLI) are both correct but not surfaced as a single mental model for "remote core + local app".

Who it hurts: anyone self-hosting the core, internal testers using a shared hosted core, and automation that shells into a container expecting `cat core.token` while the process was started with a different env token.

### Solution (optional)

Pick a **default story** and implement it end-to-end (docs + UX + ops):

1. **Single source of truth** — Document and enforce one recommended path for remote deploys, e.g. "always set `OPENHUMAN_CORE_TOKEN` in the container; never rely on `core.token` for dockerized cores" **or** "first boot prints/logs a one-time pairing URL / writes only `core.token`" with compose volume docs — but avoid ambiguous overlap with [`init_rpc_token`](https://github.com/tinyhumansai/openhuman/blob/main/src/core/auth.rs).

2. **Client simplicity** — In-app or first-run settings: field for **Remote core URL** + **RPC token**, test connection (`core.ping` or health + rpc), save to the same config surface the app already uses for cloud mode. Optional: deep link or QR from server admin page (future).

3. **Operator simplicity** — `docker compose` / `.env.example` with one variable block; optional script `scripts/print-core-token.sh` that reads the agreed location; document token rotation (bounce container + update client).

4. **Security** — Never log the token; redact in debug; document TLS + token storage on the client (OS keychain later if out of scope).

5. **Tests** — Vitest for client pairing UX; Rust or integration test that env token and file token paths cannot diverge unexpectedly for a given deploy mode if behavior changes.

### Acceptance criteria

- [ ] A new contributor can follow **one** short doc path: start server (Docker or binary), configure client, and complete a successful authenticated `/rpc` call without reading multiple conflicting token stories.
- [ ] Cloud deploy doc ([`gitbooks/features/cloud-deploy.md`](https://github.com/tinyhumansai/openhuman/blob/main/gitbooks/features/cloud-deploy.md)) and `.env.example` / compose align with the chosen semantics.
- [ ] Client configuration for remote core + token is **obvious** (settings UI and/or env) with a **connection test** where feasible.
- [ ] **Diff coverage ≥ 80%** on any new/changed code (docs + tests per repo norms).

### Related

- [`src/core/auth.rs`](https://github.com/tinyhumansai/openhuman/blob/main/src/core/auth.rs) — `OPENHUMAN_CORE_TOKEN`, `core.token`, `init_rpc_token`
- [`gitbooks/features/cloud-deploy.md`](https://github.com/tinyhumansai/openhuman/blob/main/gitbooks/features/cloud-deploy.md)
- [`app/src/components/BootCheckGate/BootCheckGate.tsx`](https://github.com/tinyhumansai/openhuman/blob/main/app/src/components/BootCheckGate/BootCheckGate.tsx) — cloud token UX hint

senamakel:
## Summary

- Add a **Test connection** button to the cloud picker in `BootCheckGate` that calls `core.ping` with the typed URL+token and surfaces `Connected ✓` / `Auth failed (401/403)` / `Unreachable` inline before the mode is committed.
- Add `scripts/print-core-token.sh` so operators can resolve the active token (env first, then `${OPENHUMAN_WORKSPACE:-~/.openhuman}/core.token`) with `--redact` / `--where` modes safe for logs.
- Document `OPENHUMAN_CORE_TOKEN` as the **single source of truth** for any remote / dockerized core in [`gitbooks/features/cloud-deploy.md`](../blob/main/gitbooks/features/cloud-deploy.md), and add a token-rotation runbook.
- Rewrite the `OPENHUMAN_CORE_TOKEN` block in `.env.example` so the three deploy modes (Tauri / Docker-cloud / standalone CLI) are obvious.
- 8 new Vitest cases cover the picker test-connection flow.

## Problem

Per #1438, the bearer-token story today is split between two paths in [`src/core/auth.rs`](../blob/main/src/core/auth.rs):

1. `OPENHUMAN_CORE_TOKEN` env (Tauri / Docker / cloud), and
2. `{workspace}/core.token` file (standalone CLI),

and operators routinely confuse the two — "where do I read the token?", "why 401 after deploy?". The desktop picker already accepts a URL + token but had no way to prove either of them was correct before navigating into the boot-check, leaving users on a generic `unreachable` screen when they typo'd anything.

## Solution

Three thin layers — UX, operator tooling, docs — each independently useful, none of them changing how `init_rpc_token` decides which token to use:

- **UX**: the picker now exposes `testCoreRpcConnection` (already exported from `coreRpcClient`) behind a button. Status sub-states (`idle`/`testing`/`ok`/`auth`/`unreachable`) reset whenever the user edits either field, so the indicator never goes stale.
- **Operator tool**: `scripts/print-core-token.sh` mirrors the resolution order in `init_rpc_token` exactly — env first, then file — so `cat core.token` (which silently misses env-mode containers) is no longer the recommended way to find the active token. `--redact` makes it safe to paste into a runbook or CI log.
- **Docs**: a "Single source of truth" section makes the rule explicit ("for any remote / dockerized deploy, always set `OPENHUMAN_CORE_TOKEN`"), plus a copy-pasteable rotation runbook.

No `core/auth.rs` behavior change; the dual-path design there is fine, the gap was that we never told operators which path their deploy used.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../blob/main/gitbooks/developing/testing-strategy.md#failure-path-requirement)
- [x] **Diff coverage ≥ 80%** — new TS in `BootCheckGate.tsx` is exercised by 8 new Vitest cases (200, 401, 403, 500, network reject, missing URL, missing token, status reset). Docs / `.env.example` / shell script are not measured by `diff-cover`.
- [ ] Coverage matrix updated — `N/A: docs + UX affordance, no new feature row`.
- [x] All affected feature IDs from the matrix are listed under `## Related` (`N/A` — none).
- [x] No new external network dependencies introduced (test uses mocked `testCoreRpcConnection`).
- [ ] Manual smoke checklist updated — `N/A: not on a release-cut surface`.
- [x] Linked issue closed via `Closes #NNN` below.

## Impact

- Desktop only — picker UX is gated by cloud mode, no change to local-mode users.
- Docs + `.env.example` are advisory; existing deployments keep working unchanged.
- Security: `print-core-token.sh` defaults to printing the full value (so it composes with `curl … -H "Authorization: Bearer $(scripts/print-core-token.sh)"`); CI / shared-log paths should use `--redact` or `--where`. Tokens are never written to debug logs by the picker — only `tokenLen` is logged.

## Related

- Closes #1438
- Follow-up PR(s)/TODOs:

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: issue/1438-simplify-remote-core-rpc-token-setup-for
- Commit SHA: $(git rev-parse HEAD)


senamakel:
@coderabbitai review

PR #1487, opened 2026-05-11
feat(memory): add tool-scoped memory with durable rule capture, RPC APIs, and prompt pinning

### LINKED ISSUE #1400: Add tool-scoped memory for durable learnings and high-priority rules
## Summary

Add tool-scoped durable memory so OpenHuman can record learnings, corrections, and high-priority operational rules against specific tools and reliably surface them when those tools are considered or invoked.

## Problem

Today the memory system has the pieces for broad persistence, but not the right substrate for tool-specific learnings:

- `memory_store` / `memory_recall` already persist facts by namespace, but the common path is still broad namespaces like `global` and `skill-{id}`.
- `src/openhuman/learning/tool_tracker.rs` records tool effectiveness statistics in the `tool_effectiveness` namespace, but only as aggregate call/failure metrics.
- `src/openhuman/agent/harness/memory_context.rs` and session-memory flows can surface durable context, but they do not treat tool-specific rules as a first-class, high-priority retrieval surface.
- `gitbooks/features/token-compression.md` and `src/openhuman/context/session_memory.rs` make it clear that in-flight context is intentionally compressed and synthesized over time.

That creates a failure mode for important operational instructions and learnings. If the user says “never email Sarah”, storing that only in global memory makes it vulnerable to being missed, deprioritized, or compressed away in future turns. The same applies to tool learnings like:

- “This shell command failed because the repo uses pnpm, not npm.”
- “For this contact, never use the email tool without explicit confirmation.”
- “This API needs a dry-run first.”

These are not just generic user memories. They are tool-shaped guardrails and learned heuristics that should follow the relevant tool more closely than normal conversational memory.

Constraints:

- The implementation must stay local-first and avoid logging secrets or sensitive contact rules carelessly.
- Tool-specific memory should improve recall quality without flooding every prompt with all past tool notes.
- The design should fit existing memory namespaces, post-turn learning hooks, tool registry/harness flows, and token-compression behavior rather than introducing a parallel memory stack.

## Solution (optional)

Introduce a first-class tool-memory layer for durable, high-priority learnings tied to one or more tools.

Expected shape:

- A dedicated storage convention for tool learnings, such as `tool-{name}` or similar domain-owned namespaces, separate from generic `global` memory.
- Support for explicit priority/criticality so some rules are surfaced more aggressively than ordinary memories.
- A way for post-turn learning hooks or explicit tool actions to record “what went wrong”, “what to avoid”, and “what to do differently next time”.
- Retrieval in the agent/tool-selection path so the relevant tool gets its high-priority notes before or during execution.
- Clear UX/documentation describing the difference between global memory and tool-based memory.

The “never email Sarah” case is a good product test: that instruction should be storable in a way that preferentially follows email-sending tools instead of relying on generic global recall alone.

## Acceptance criteria

- [ ] **Tool-scoped memory model** — there is a first-class way to store and retrieve learnings tied to specific tools or tool families, rather than only generic `global` memory.
- [ ] **High-priority rules supported** — the system can mark some tool memories as high priority / safety-relevant so they are less likely to be lost behind ordinary recall/compression behavior.
- [ ] **Learning capture path** — the agent can record mistakes, corrections, and “next time do X instead” insights after tool usage, either via post-turn learning hooks or an explicit tool-memory write path.
- [ ] **Retrieval in execution path** — relevant tool memories are surfaced during tool selection and/or just before tool execution so they can materially change behavior.
- [ ] **Safety example covered** — cases like “never email Sarah” can be represented as tool-scoped guidance for communication tools instead of only living in broad global memory.
- [ ] **Stats + learnings separation** — aggregate tool effectiveness metrics (`tool_effectiveness`) remain distinct from actionable tool guidance, or are intentionally unified under a documented model.
- [ ] **GitBook docs updated** — `gitbooks/` documentation explains how tool-based memory works, how it differs from global memory, how priority affects retrieval, and where it fits relative to token compression and session memory.

## Related

- #1399
- `src/openhuman/tools/impl/memory/store.rs`
- `src/openhuman/tools/impl/memory/recall.rs`
- `src/openhuman/learning/tool_tracker.rs`
- `src/openhuman/tools/impl/system/tool_stats.rs`
- `src/openhuman/agent/harness/memory_context.rs`
- `src/openhuman/context/session_memory.rs`
- `docs/memory-sync-functions.md`
- `gitbooks/features/token-compression.md`

YellowSnnowmann:
# Summary

- Added a new `tool_memory` domain under:
  - `src/openhuman/memory/`
  - Persists per-tool rules in dedicated `tool-{name}` namespaces

- Introduced tool-memory RPC methods:
  - `tool_rule_put`
  - `tool_rule_get`
  - `tool_rule_list`
  - `tool_rule_delete`
  - `tool_rules_for_prompt`
  - `tool_rules_json`

- Registered schemas/controllers in memory registry

- Updated session prompt construction:
  - Prefetches Critical/High tool rules
  - Pins them into system prompt
  - Ensures safety-important constraints survive context compression

- Added `ToolMemoryCaptureHook`:
  - Automatically captures durable tool guidance from learning signals
  - Controlled via config

- Added config support:
  - `learning.tool_memory_capture_enabled`
  - `OPENHUMAN_LEARNING_TOOL_MEMORY_CAPTURE_ENABLED`

- Updated capability/docs surfaces:
  - `about_app/catalog.rs`
  - GitBook feature docs
  - Summary/docs pages

- Added focused tests:
  - Store behavior
  - Schema integration

---

# Problem

- Tool guidance was fragmented across generic memory systems
- Safety-critical tool constraints could be lost during:
  - context compression
  - prompt trimming

- No dedicated API for:
  - storing tool-specific rules
  - retrieving durable operational constraints

- No centralized architecture for:
  - capturing durable learnings
  - injecting them reliably into prompts

---

# Solution

## Tool Memory Storage

Implemented dedicated storage layer:

- `ToolMemoryStore`
- Stores structured `ToolMemoryRule` entries
- Uses:
  - `tool-{tool}` namespaces
  - priority metadata
  - source metadata

---

## RPC + Controller Integration

Added tool-memory RPC handlers and schema registrations:

- Exposed through standard controller routing
- Avoids adapter-specific implementations

---

## Prompt Integration

Added deterministic prompt injection:

- Prefetch Critical/High rules
- Inject into system prompt at session start
- Improves durability across compression boundaries

---

## Learning Hook

Added `ToolMemoryCaptureHook`:

- Captures durable guidance post-turn
- Integrated into learning flow
- Rollout controlled via config

---

# Tradeoffs

- Prefetch is:
  - best-effort
  - intentionally non-fatal

- Unsupported runtimes/errors:
  - return empty results
  - preserve stability in sync/single-thread harnesses

---

# Submission Checklist

- **Tests**
  - Added/updated per Testing Strategy
  - Includes:
    - happy paths
    - edge/failure cases

- **Coverage**
  - Diff coverage ≥ 80%
  - Enforced via:
    - `.github/workflows/coverage.yml`

- **Commands**
  - `pnpm test:coverage`
  - `pnpm test:rust`

- **Coverage Matrix**
  - Updated in:
    - `docs/TEST-COVERAGE-MATRIX.md`
  - Or marked N/A where appropriate

- **Feature IDs**
  - Listed under `## Related`

- **Network Dependencies**
  - No new external dependencies
  - Uses mock backend per Testing Strategy

- **Manual Smoke Checklist**
  - Updated if release-cut surfaces affected:
    - `docs/RELEASE-MANUAL-SMOKE.md`

- **Linked Issues**
  - Added via:
    - `Closes #NNN`

---

# Impact

## Runtime / Platform

- Affects:
  - Rust core memory system
  - agent session prompt pipeline

- No platform-specific shell behavior added

---

## Performance

- Small startup overhead from best-effort prefetch
- Bounded by:
  - prompt cap
  - scoped retrieval

---

## Security / Safety

- Improves durability of:
  - high-priority tool constraints
  - safety-critical rules

- Reduces unsafe behavior caused by:
  - prompt compression loss

---

## Compatibility / Migration

- Additive API/config changes
- Default behavior remains enabled
- Supports opt-out via:
  - config
  - environment variable

---

## Related

- Closes #1400 

## Summary by CodeRabbit

* **New Features**
  * Tool-Scoped Memory to manage durable, tool-specific rules with priority handling
  * Rules can be injected into sessions (critical/high pinned; normal retrieved on demand)
  * RPC endpoints for rule CRUD and prompt-resolution
  * Optional automatic capture of user edicts and repeated tool failures

* **Configuration**
  * New toggle to enable/disable automatic tool memory capture

* **Documentation**
  * Added docs and TOC entry for Tool-Scoped Memory

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1487)


senamakel:
wow what a fucking goated PR

YellowSnnowmann:
@coderabbitai review

src/openhuman/agent/harness/session/builder.rs
```diff
@@ -787,6 +787,24 @@ impl Agent {
...
         }
 
+        // (#1400) Pre-fetch Critical + High priority tool-scoped memory
+        // rules so they pin into the (compression-resistant) system
+        // prompt for the whole session. Skipped silently when the
+        // current runtime cannot host a synchronous bridge — typically
+        // a single-threaded test harness — so this stays safe for every
+        // call site of `from_config_*`. The capture hook still runs in
+        // every session via [`ToolMemoryCaptureHook`] above.
+        if config.learning.tool_memory_capture_enabled {
```
graycyrus: [blocker] **Prefetch fires even when `learning.enabled = false`**

This is gated on `tool_memory_capture_enabled` but not on `learning.enabled`. The capture hook (line 862) is correctly nested inside `if config.learning.enabled { ... }`, so no new rules are written when learning is off. But previously-stored rules still get pinned into the system prompt every session.

Users who disable learning expect deterministic, stored-rule-free agent behavior.

**Suggested fix:**
```rust
// before
if config.learning.tool_memory_capture_enabled {

// after
if config.learning.enabled && config.learning.tool_memory_capture_enabled {
```

src/openhuman/memory/tool_memory/capture.rs
```diff
@@ -0,0 +1,573 @@
...
+        user_message: &str,
+        tool_calls: &[ToolCallRecord],
+    ) -> Vec<(String, String)> {
+        let trimmed = user_message.trim();
+        if trimmed.is_empty() {
+            return Vec::new();
+        }
+        let lower = trimmed.to_lowercase();
+        if !(lower.contains("never ") || lower.contains("don't ") || lower.contains("do not "))
+            && !lower.contains("stop ")
```
graycyrus: [major] **`"stop "` edict trigger is too broad**

`lower.contains("stop ")` matches routine utterances: "I want to stop working on this", "stop the server for maintenance", "you should stop after three retries". All of these fall through to the per-line scan, which also matches `contains(" stop ")` (line 109), creating false-positive edicts.

The doc comment says "stop \<verb\>ing" but the code doesn't enforce that form.

**Suggested fix:** Narrow to sentence-boundary forms only:
```rust
// before (line 80-83)
if !(lower.contains("never ") || lower.contains("don't ") || lower.contains("do not "))
    && !lower.contains("stop ")
{

// after
if !lower.contains("never ")
    && !lower.contains("don't ")
    && !lower.contains("do not ")
    && !lower.starts_with("stop ")
    && !lower.contains("\nstop ")
    && !lower.contains(". stop ")
{
```

Also add a negative test:
```rust
#[test]
fn extract_user_edicts_ignores_incidental_stop_usage() {
    let edicts = ToolMemoryCaptureHook::extract_user_edicts(
        "I want to stop the server for maintenance.",
        &[call("shell", true)],
    );
    assert!(edicts.is_empty());
}
```

src/openhuman/agent/harness/session/builder.rs
```diff
@@ -1152,3 +1179,51 @@ impl Agent {
...
+    let Ok(handle) = tokio::runtime::Handle::try_current() else {
+        return Vec::new();
+    };
+    if handle.runtime_flavor() != tokio::runtime::RuntimeFlavor::MultiThread {
+        return Vec::new();
+    }
+    tokio::task::block_in_place(|| {
+        handle.block_on(async move {
+            let store = crate::openhuman::memory::ToolMemoryStore::new(memory);
+            match store.rules_for_prompt(&[]).await {
```
graycyrus: [major] **Prefetch scans ALL tool namespaces regardless of agent's tool set**

`rules_for_prompt(&[])` with an empty slice falls back to `list_tool_names`, scanning every `tool-*` namespace. An agent with a restricted tool set (e.g. a read-only sub-agent that can't use `send_email`) will still have those tools' critical rules injected into its system prompt.

Beyond prompt-budget waste, this is a correctness issue for restricted agents — rules about tools the agent shouldn't know about get injected.

The session builder already resolves `visible_tools` later in the same function. The minimal fix is to thread tool names through, or defer the inject until after tool resolution.

**At minimum:** add a `// TODO: pass resolved tool names to scope prefetch` comment so the debt is visible, and document the scope gap in the function's rustdoc.

src/openhuman/memory/tool_memory/types.rs
```diff
@@ -0,0 +1,250 @@
...
+
+impl Default for ToolMemoryPriority {
+    fn default() -> Self {
+        Self::Normal
+    }
+}
+
+impl ToolMemoryPriority {
+    /// True for priorities that must be eagerly surfaced to the agent.
+    pub fn is_eager(self) -> bool {
```
graycyrus, line 48: [minor] **`is_pinned` and `is_eager` are identical — remove one**

Both return `matches!(self, Self::Critical | Self::High)`. Two names for the same predicate creates confusion about whether they might diverge. The PR uses "eager" for prefetch and "pinned" for compression resistance, but the code makes no distinction.

Either remove `is_pinned` (and use `is_eager` everywhere), or give `is_pinned` a genuinely different definition (e.g. only `Critical` is truly compression-pinned).

src/openhuman/memory/tool_memory/prompt.rs
```diff
@@ -0,0 +1,227 @@
...
+        let rules = vec![
+            rule("email", "never email Sarah", ToolMemoryPriority::Critical),
+            rule("shell", "avoid sudo", ToolMemoryPriority::High),
+        ];
+        let first = render_tool_memory_rules(&rules);
+        let again = render_tool_memory_rules(&rules);
+        assert_eq!(first, again);
+    }
+
+    #[test]
```
graycyrus, line 218: [minor] **Test named `section_renders_via_prompt_section_trait` doesn't call `build()`**

This test only calls `is_empty()`, which reads the `rendered` field directly. The `build(&ctx)` method — the actual `PromptSection` trait contract — is never exercised. If `build` were accidentally changed to return an error, no test would catch it.

**Suggested fix:** Actually invoke the trait method:
```rust
let ctx = PromptContext::default();
let output = section.build(&ctx).expect("build should succeed");
assert!(output.contains("never email Sarah"));
```

src/openhuman/memory/ops/tool_memory.rs
```diff
@@ -0,0 +1,154 @@
+//! RPC handlers for the tool-scoped memory layer (see
```
graycyrus, line 1: [nitpick] **No debug logging in any RPC handler**

Every other ops file logs `[rpc]` or `[memory]` prefixed lines at handler entry. These 6 handlers have no `log::debug!` at all. Even a single line per handler (`log::debug!("[tool-memory] {fn_name} tool={tool_name}")`) would satisfy the project convention and make traces grep-able.

graycyrus (commented):
## Independent review (non-CodeRabbit)

Overall: well-structured PR that follows project conventions closely. The `tool_memory/` domain is cleanly separated, store tests are thorough (22 async tests), and the CodeRabbit feedback was addressed. A few things stand out — see inline comments.

### Summary
- **1 blocker** — config gate asymmetry (prefetch fires when `learning.enabled = false`)
- **2 major** — broad "stop" edict trigger false positives; prefetch scans all namespaces for restricted agents
- **2 suggestions** — duplicate `is_pinned`/`is_eager`; test doesn't exercise `build()` trait method
- **4 nitpicks** — duplicate `MockMemory`, missing ops logging, `__unscoped__` sentinel leak, stale issue ref in log

### Looks good
- Module layout (`tool_memory/` subdirectory with light `mod.rs`)
- Controller registration wired correctly through `schemas/mod.rs`
- No bare `.unwrap()` in production code
- No PII in logs (capture hook truncates rule bodies)
- `block_in_place` runtime check is correct
- `put_rule` preserves `created_at` on upsert
- Config env override follows `parse_env_bool` pattern
- Capability catalog entry present with `LOCAL_RAW` privacy tag

src/openhuman/memory/tool_memory/capture.rs
```diff
@@ -0,0 +1,573 @@
...
+        }
+
+        // Default tool: the first tool that ran in the turn. When there
+        // were no tool calls we still want to capture user edicts so
+        // they survive into the next turn — those land under the
+        // `__unscoped__` tool name and the agent can refile them.
+        let default_tool = tool_calls
+            .first()
+            .map(|tc| tc.name.clone())
+            .unwrap_or_else(|| "__unscoped__".to_string());
```
graycyrus, line 98: [nitpick] **`__unscoped__` sentinel leaks into namespace scanning**

When no tool calls ran, edicts land under `tool-__unscoped__`. This namespace will then be scanned by `list_tool_names` / `rules_for_prompt`, potentially surfacing unscoped rules into future prompts unexpectedly.

Consider filtering this sentinel from `list_tool_names`, or using a reserved prefix that the scanner explicitly skips.

src/openhuman/agent/harness/session/builder.rs
```diff
@@ -840,6 +858,15 @@ impl Agent {
                 )));
                 log::info!("[learning] tool_tracker hook registered");
             }
+
+            if config.learning.tool_memory_capture_enabled {
+                post_turn_hooks.push(Arc::new(
+                    crate::openhuman::memory::ToolMemoryCaptureHook::new(memory.clone(), true),
+                ));
+                log::info!(
```
graycyrus: [nitpick] **Raw issue number `#1400` in runtime log message**

Log messages with issue references become stale. Prefer a stable description:
```rust
// before
"[learning] tool_memory_capture hook registered (#1400 — durable tool rules)"

// after
"[learning] tool_memory_capture hook registered"
```

src/openhuman/memory/tool_memory/capture.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+    if lower.contains("slack") {
+        out.push("slack");
+        out.push("dm");
+    }
+    out
+}
+
+#[cfg(test)]
+mod tests {
```
graycyrus, line 289: [nitpick] **Duplicate `MockMemory` impl**

This `MockMemory` is defined identically in `store_tests.rs`. Extract it into a shared `#[cfg(test)]` helper in `tool_memory/mod.rs` or a `test_helpers` module to avoid silent drift between the two copies.

YellowSnnowmann:
Addressed all blockers and majors from the review. Changes pushed to the branch ([`cd86784a`](https://github.com/YellowSnnowmann/openhuman/commit/cd86784a) + [`492d8e3d`](https://github.com/YellowSnnowmann/openhuman/commit/492d8e3d)):

**Blockers / Majors fixed:**
- **`learning.enabled` gate** — prefetch now checks `config.learning.enabled && config.learning.tool_memory_capture_enabled`; users who disable learning no longer get stored rules pinned into the system prompt
- **Prefetch scoped to agent's tool set** — moved the prefetch after the tool list is finalised and passes actual tool names to `rules_for_prompt`; unrelated namespaces are never scanned
- **"stop " false positives** — outer check now requires "stop " at a sentence boundary (`starts_with` / `. stop ` / `\nstop `); removed `contains(" stop ")` from the per-line scanner
- **PII in logs** — repeated-failure log now emits only `body_len`, matching the edict capture path
- **`tool_rules_for_prompt` error masking** — the schema handler already uses `parse_params` (which returns an error on bad input); the `unwrap_or_default()` calls are in optional fields (`priority`, `source`) that default correctly per spec — no change needed there

**Majors / Minors cleaned up:**
- Removed duplicate `is_pinned` predicate (`is_eager` already covers Critical+High); updated doc-comment
- Filtered `__unscoped__` sentinel from `list_tool_names` so it never surfaces in prompt-injection scans
- Added `log::debug!` entry to all six RPC handlers (`[tool-memory]` prefix)
- Removed raw issue reference from the `tool_memory_capture` registration log line
- `section_renders_via_prompt_section_trait` now actually calls `build(&ctx)` and asserts on the output
- Extracted shared `MockMemory` to `tool_memory/test_helpers.rs`; both `store_tests.rs` and `capture.rs` use it — eliminates silent drift

All 41 tool-memory tests pass; `cargo check` clean.

PR #1517, opened 2026-05-12
perf(conversations): track message stats in threads log so list_threads is O(threads.jsonl)

senamakel:
## Summary

- `ConversationStore::list_threads` no longer reads every per-thread messages JSONL on every call. It folds `message_count` and `last_message_at` straight out of `threads.jsonl`.
- Adds two new `ThreadLogEntry` variants: `MessageAppended` (incremental, emitted from `append_message`) and `Stats` (absolute snapshot, used to backfill legacy data).
- Legacy threads with no stat history get backfilled on the first `list_threads` call after upgrade — one slow per-file read each, then a persisted `Stats` snapshot so every subsequent call is fast.
- Old-format `threads.jsonl` still parses (`#[serde(default)]` on new fields, `Upsert`/`Delete` shape unchanged).

## Problem

Sentry OPENHUMAN-REACT-X surfaced as a 30 s timeout on `openhuman.threads_list` (unhandled-rejection band-aid landed in #1513). Tracing the call into the core revealed the actual root cause:

```
list_threads_unlocked
 └─ for each thread:
     thread_summary_unlocked(id)
       ├─ thread_index_unlocked()           ← re-reads threads.jsonl per thread
       └─ read_jsonl::<ConversationMessage> ← reads the ENTIRE per-thread messages file
```

That's **O(threads² + total messages bytes)** on every `list_threads`, all inside the global `CONVERSATION_STORE_LOCK` — so any other conversation RPC queues behind it. A user with 100 threads + a year of chat history can easily push a single call past 30 s on slow disk.

## Solution

Push stats into `threads.jsonl` so folding the log gives us everything:

| Variant | Purpose |
| --- | --- |
| `Upsert` (unchanged) | thread metadata (title, labels, created_at, parent) |
| `Delete` (unchanged) | remove from index |
| `MessageAppended { thread_id, last_message_at }` | bump count by 1, overwrite timestamp — emitted by `append_message` under the existing lock |
| `Stats { thread_id, message_count, last_message_at }` | absolute snapshot — used to backfill legacy threads once, after which the fast path takes over |

The fold in `thread_index_unlocked` carries `(message_count: Option<usize>, last_message_at: Option<String>)` per thread. `None` means \"no stat history yet\" → `list_threads_unlocked` does the one-time slow read and writes a `Stats` entry. `thread_summary_unlocked` (single-thread queries) reads the index first and only falls back to a per-thread file read when stats are unknown.

The hot path is now one sequential read of `threads.jsonl` + a sort. No per-thread file opens.

### Trade-off / follow-up

The log now grows by one entry per message append. A compaction pass (rewriting `threads.jsonl` as a folded snapshot when it crosses a size threshold) is a sane future improvement but not needed for correctness. Not in scope for this PR.

## Submission Checklist

- [x] Tests added — 4 new tests covering: fast path after warm-up, legacy backfill writes `Stats` snapshot, old-format log still parses, delete clears index stats. Existing 15 tests in `store_tests.rs` all still pass.
- [x] Diff coverage ≥ 80% — every new branch in `store.rs` is exercised by the new tests (`MessageAppended` fold, `Stats` fold, backfill loop, `measure_messages_unlocked`, `thread_summary_unlocked` index-first path).
- [x] N/A: behaviour-only change to an existing storage path, no feature matrix rows affected
- [x] N/A: same — no feature IDs to list
- [x] N/A: no release-cut smoke surface touched
- [x] N/A: no linked issue (Sentry OPENHUMAN-REACT-X; this PR refs but does not close it — #1513 is the user-facing fix)

## Impact

- **Runtime/platform**: desktop. Same on-disk layout, no migration needed.
- **Performance**: `list_threads` drops from O(threads² + total message bytes) to O(threads.jsonl). For a user with a year of chat history, expect a multi-second → sub-millisecond improvement on warm cache, plus the global `CONVERSATION_STORE_LOCK` is released sooner so concurrent conversation RPCs stop queuing.
- **Compatibility**: forward-compatible read of legacy logs (verified by test `legacy_log_without_stats_still_parses`). Backward-compatible writes are not relevant — once `MessageAppended` / `Stats` entries are in the log, an older binary opening the same workspace would skip them (`serde` tagged-enum default behaviour is to fail; this would surface as warnings via the existing `read_jsonl` line-skip path). In practice this is fine because we don't expect downgrades.
- **Storage**: threads.jsonl grows by one short entry per message append (`~80 bytes`). For a user with 10 k messages that's ~800 KB. Acceptable until we add compaction.

## Related

- Refs Sentry OPENHUMAN-REACT-X (root cause behind the timeout that #1513 swallowed at the UI layer)
- Follow-up PR(s)/TODOs: compaction for `threads.jsonl` (rewrite as folded snapshot when log size > threshold)

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: perf/threads-list-stats-in-log
- Commit SHA: e3ff27d0cdc57c86311b734c2fa19c560b67fc76

### Validation Run
- [x] N/A: format check is run by CI; cargo fmt was applied locally before commit
- [x] N/A: this PR touches only the Rust core; no app-side TypeScript changes
- [x] Focused tests: `cargo test --lib conversations::store` — 19 / 19 pass (15 pre-existing + 4 new)
- [x] Rust fmt/check: `cargo fmt -- --check src/openhuman/memory/conversations/store{.rs,_tests.rs}` clean; `cargo check --manifest-path Cargo.toml` clean
- [x] N/A: no Tauri shell changes

### Validation Blocked
- command: N/A
- error: N/A
- impact: N/A

### Behavior Changes
- Intended behavior change: `list_threads` returns the same data shape and ordering, computed from the threads-log index instead of from per-thread file reads. `append_message` emits a `MessageAppended` entry to `threads.jsonl`.
- User-visible effect: faster thread list loads (especially for users with long chat history); reduced lock-hold time on `CONVERSATION_STORE_LOCK` improves concurrent-RPC behaviour. Same UX otherwise.

### Parity Contract
- Legacy behavior preserved: ordering by (last_message_at desc, created_at desc); message_count fallback to 0 and last_message_at fallback to created_at when no messages exist; existing tests unchanged.
- Guard/fallback/dispatch parity checks: `thread_summary_unlocked` falls back to the old per-file read path when index stats are missing; `list_threads` triggers a one-shot backfill that persists a `Stats` snapshot, restoring the fast path on the next call.

- Duplicate PR(s): N/A
- Canonical PR: N/A — complements #1513
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Improved error message formatting for conversation loading failures.

* **Bug Fixes**
  * Enhanced compatibility with legacy conversation data through automatic backfill.
  * Better handling of missing thread statistics.

* **Performance**
  * Optimized conversation list loading by reducing unnecessary file access.

* **Tests**
  * Expanded test coverage for error formatting and conversation indexing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1517)

### DIFF
--- a/app/src/pages/Conversations.tsx
+++ b/app/src/pages/Conversations.tsx
@@ -110,6 +110,24 @@ export function isComposerInteractionBlocked(args: {
   return !args.rustChat || Boolean(args.activeThreadId) || args.welcomePending;
 }
 
+/**
+ * Normalise the value thrown out of `dispatch(loadThreads()).unwrap()` into a
+ * displayable string. `createAsyncThunk` re-throws Redux's `SerializedError`
+ * (a plain object, not an `Error` instance) when the thunk rejects — which is
+ * why the original Sentry report (OPENHUMAN-REACT-X) showed up as
+ * "Non-Error promise rejection captured with value: …" rather than a stack.
+ * Exported so the mount-effect's `.catch` stays a one-liner and the message
+ * shape can be unit-tested without mounting the full page.
+ */
+export function formatThreadLoadError(err: unknown): string {
+  if (err instanceof Error) return err.message;
+  if (err && typeof err === 'object' && 'message' in err) {
+    const message = (err as { message?: unknown }).message;
+    if (typeof message === 'string') return message;
+  }
+  return String(err);
+}
+
 // [#1123] Commented out — welcome-agent onboarding replaced by Joyride walkthrough
 // function WelcomeThinkingTypewriter() {
 //   const text = 'Your agent is thinking...';
@@ -291,8 +309,7 @@ const Conversations = ({ variant = 'page', composer = 'text' }: ConversationsPro
       })
       .catch(err => {
         if (cancelled) return;
-        const message = err instanceof Error ? err.message : String(err?.message ?? err);
-        console.warn('[conversations] loadThreads failed on mount:', message);
+        console.warn('[conversations] loadThreads failed on mount:', formatThreadLoadError(err));
       });
 
     return () => {
--- a/app/src/pages/__tests__/Conversations.test.tsx
+++ b/app/src/pages/__tests__/Conversations.test.tsx
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { isComposerInteractionBlocked } from '../Conversations';
+import { formatThreadLoadError, isComposerInteractionBlocked } from '../Conversations';
 
 describe('isComposerInteractionBlocked', () => {
   it('blocks composer interaction while the welcome agent loader is visible', () => {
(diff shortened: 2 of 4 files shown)


senamakel:
@coderabbitai review

Issue #1472, opened 2026-05-11
Track and fix active Sentry issues


Al629176:
## Summary

Track, triage, and fix the current Sentry issues affecting OpenHuman so the app has a clean, monitored error baseline.

## Problem / Context

Sentry is reporting issues that need systematic review instead of one-off fixes. Without a tracking pass, recurring frontend, Tauri shell, and Rust core failures can stay noisy, hide new regressions, and make it harder to know whether staging or production builds are healthy.

This work should identify the active Sentry issue set, group duplicates/root causes, prioritize user-impacting crashes and high-volume errors, fix actionable issues, and document any intentional or external-noise items.

Constraints:
- Preserve privacy: do not copy sensitive user data, tokens, or raw event payloads into issues/PRs.
- Separate real product errors from environment noise, extension/browser noise, and expected cancellation/timeout paths.
- Keep fixes scoped and covered so new Sentry regressions are easier to detect.

## Scope (optional)

In scope:
- Review active Sentry issues for OpenHuman staging and production.
- Group related Sentry issues by root cause.
- Fix actionable frontend, Tauri, and core errors.
- Add breadcrumbs/logging or error classification where needed to make future events diagnosable.
- Update Sentry issue status after fixes are released or verified.

Out of scope:
- Broad observability redesign unless required to resolve the current issues.
- Logging sensitive payloads or user data.

## Acceptance criteria

- [ ] **Issue inventory** — Current active Sentry issues are reviewed and grouped by root cause, severity, environment, and affected app surface.
- [ ] **Priority fixes landed** — User-impacting crashes, high-volume errors, and staging/production blockers have fixes or linked follow-up issues with owners.
- [ ] **Noise classified** — Non-actionable, expected, external, or duplicate Sentry issues are documented and resolved/ignored in Sentry with rationale.
- [ ] **Regression safety** — Tests are added or updated for fixed code paths where practical, especially for crashes or repeated high-volume errors.
- [ ] **Release verification** — Fixed issues are verified in the relevant staging or production build and marked resolved in Sentry after release.
- [ ] **Future diagnosability** — Missing context is improved with safe breadcrumbs, structured logs, or error metadata that does not expose secrets or sensitive user data.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)) when code changes are involved.

[...]

PR #1529, opened 2026-05-12
fix(observability): drop transient upstream HTTP from Sentry (429/408/502/503/504)

oxoxDev:
## Summary

- Treat 429/408/502/503/504 from upstream providers as transient — no per-attempt Sentry report (already retried + failed-over by `reliable.rs`).
- Demote transient `agent.provider_chat` failures to `warn!`; the aggregate `all_exhausted` event still surfaces real outages.
- Defense-in-depth `before_send` filter in `main.rs` drops residual transient events that bypass the call-site classifier.

## Problem

Three Sentry issues bury actionable signal under per-attempt transient-upstream noise:

| Sentry ID | Symptom | Events |
|---|---|---|
| [OPENHUMAN-TAURI-2E](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-2E) | `llm_provider.api_error … 429 Too Many Requests` | ~1393 / day |
| [OPENHUMAN-TAURI-84](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-84) | `… 503 Service Unavailable` | ~1050 |
| [OPENHUMAN-TAURI-T](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-T) | `… 504 Gateway Timeout` | ~871 |
| [OPENHUMAN-TAURI-3Y](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-3Y) | `llm_provider.reliable_chat … All providers/models failed` (cascade) | 23 |
| [OPENHUMAN-TAURI-3Z](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-3Z) | `agent.provider_chat … All providers/models failed` (cascade) | 23 |

`providers/ops.rs::should_report_provider_http_failure` only excluded 429; 503/504 still went per-attempt. `agent/harness/tool_loop.rs` reported every provider_chat error, duplicating reliable.rs's aggregate.

## Solution

1. **`providers/ops.rs`** — expand the no-report set to `{429, 408, 502, 503, 504}`. Doc comment updated to name the Sentry issues; existing test extended to cover all five transient codes plus auth/4xx/5xx-permanent codes that still report.
2. **`agent/harness/tool_loop.rs`** — classify the chat error via existing `is_rate_limited` / `is_upstream_unhealthy` helpers before calling `report_error`. Transient → `tracing::warn!` only; permanent → unchanged Sentry path.
3. **`main.rs`** — `before_send` filter inspects Sentry event tags (`failure=non_2xx` + `status` in transient set) and short-circuits. Four unit tests cover dropped vs kept variants (transient drop, permanent keep, aggregate `all_exhausted` keep, missing tag keep).

The aggregate `report_error` in `reliable.rs::reliable_chat_*` (`failure=all_exhausted`) intentionally still fires — that's the cascade signal for genuine outages, one event per request.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — extended `skips_sentry_report_for_transient_upstream_statuses` to cover transient set + permanent set; added four `transient_filter_tests` for the `before_send` filter
- [x] **Diff coverage ≥ 80%** — every changed line in the classifier, tool_loop guard, and `before_send` filter is exercised by the targeted tests. `cargo test --lib openhuman::providers::ops::tests::skips_sentry_report_for_transient_upstream_statuses` + `cargo test --bin openhuman-core transient_filter` both pass; full lib suite not run locally per workspace guidance (let CI matrix run it).
- [x] Coverage matrix updated — `N/A: observability-only fix, no new user-visible feature row`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: behaviour-only change`

## Impact

- **Runtime/platform**: desktop (macOS, Windows, Linux) — Rust core only. No frontend / Tauri-shell changes.
- **Observability**: Sentry event volume drops ~3300+/day on the top-three transient noise issues. Real outage signal (`reliable_chat … all_exhausted`) untouched. No log lost — every transient drop still emits a structured `warn!` trace with `provider`, `model`, `iteration`, `error` fields.
- **Compatibility**: no API/RPC surface change; no behavioral change for retry/fallback (the error still propagates through `reliable.rs` exactly as before).
- **Security**: secret scrubbing path in `before_send` unchanged; new filter runs strictly before existing logic.
- **Performance**: tag lookup is a `BTreeMap` get + `matches!` on a five-arm pattern — negligible per-event overhead.

## Related

- Closes OPENHUMAN-TAURI-2E
- Closes OPENHUMAN-TAURI-84
- Closes OPENHUMAN-TAURI-T
- Closes OPENHUMAN-TAURI-3Y
- Closes OPENHUMAN-TAURI-3Z
- Refs #1472 (Sentry triage meta-issue)
- Follow-up PR(s)/TODOs: insufficient-budget (-3M, ~4 events) handled by same `should_report_provider_http_failure` change since it returns 400 — wait, 400 is reportable; deliberately left alone (legitimate misconfig surface)

---

### Linear Issue
- Key: N/A — Sentry-driven, not Linear
- URL: N/A

### Commit & Branch
- Branch: `fix/obs-rate-limit`
- Commit SHA: see git log on PR

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — N/A: no app/ changes
- [x] `pnpm typecheck` — N/A: no TS changes
- [x] Focused tests:
  - `cargo test --lib openhuman::providers::ops::tests::skips_sentry_report_for_transient_upstream_statuses` (1 passed)
  - `cargo test --bin openhuman-core transient_filter` (4 passed)
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo check --manifest-path Cargo.toml --bins` clean
- [x] Tauri fmt/check (if changed): N/A — no `app/src-tauri/` changes

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: per-attempt 429/408/502/503/504 LLM provider failures no longer produce Sentry events; agent.provider_chat transient errors no longer produce Sentry events; new `before_send` defense-in-depth filter for any future bypass
- User-visible effect: none direct — observability change. Indirect benefit: real Sentry issues become legible because transient noise drops.

### Parity Contract
- Legacy behavior preserved: error still propagates through reliable.rs retry/fallback exactly as before; aggregate `all_exhausted` Sentry event still fires for genuine outages
- Guard/fallback/dispatch parity checks: tests assert both "drop transient" and "keep permanent" / "keep aggregate" branches

- Duplicate PR(s): N/A
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduce monitoring noise by suppressing per-attempt reporting for a broader set of transient upstream/provider HTTP failures (rate limits, timeouts, temporary outages) while preserving permanent and aggregate failure reports.

* **Tests**
  * Add runtime smoke tests and unit tests to validate transient-event filtering and ensure permanent/aggregate events still report.

* **Documentation**
  * Expand observability docs describing reporting/filtering behavior.

* **Chores**
  * Add test-only Sentry dependency for observability smoke tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1529)


oxoxDev:
Hey @graycyrus — ready for review when you have a minute.

Addressed all three items from the CodeRabbit pass in three micro-commits (one logical change each, GPG-signed):

| Commit | What | CR item |
|---|---|---|
| [`4b5ede24`](https://github.com/oxoxDev/openhuman/commit/4b5ede24) | Extract `TRANSIENT_PROVIDER_HTTP_STATUSES: &[u16]` in `core/observability.rs`; both `ops::should_report_provider_http_failure` and `is_transient_provider_http_failure` now reference the same slice — no more drift between the two layers. Pure refactor, no behavior change. | ⚡ Nitpick — duplicated status set |
| [`1388d790`](https://github.com/oxoxDev/openhuman/commit/1388d790) | **The real bug.** `is_upstream_unhealthy` was only text-pattern matching ("no healthy upstream" / "service unavailable") and missed the `"<provider> API error (408 Request Timeout): …"` format that `ops::api_error` actually emits. Added a `reqwest::Error` downcast for typed 408/502/503/504 plus the three matching text fallbacks. `is_rate_limited` (429) stays unchanged. Three new tests in `reliable_tests.rs` covering each new pattern. | 🔴 Critical — 408 leaked per-iteration |
| [`0ebe64fa`](https://github.com/oxoxDev/openhuman/commit/0ebe64fa) | Added `domain == "llm_provider"` precondition to `is_transient_provider_http_failure`. Every current emit site already passes that domain to `report_error`, so behavior is preserved for target events — the guard only narrows what *future* call sites can accidentally match. Updated fixtures + added two regression tests (no-domain-tag and `scheduler` domain reusing the same shape both surface). | 🟠 Major — filter scope too broad |

All targeted tests pass locally:
- `cargo test --lib core::observability::tests::transient_filter` — 6 passed
- `cargo test --lib openhuman::providers::reliable::tests::upstream_unhealthy` — 7 passed (3 new + 4 existing)
- `cargo test --test observability_smoke` — 4 passed (integration via `sentry::test::TestTransport`)
- `cargo fmt --check` clean, `cargo check --lib --tests` clean (only pre-existing warnings in unrelated `webview_accounts`).

No changes to the headline transient-set semantics or the aggregate `all_exhausted` cascade signal — the per-attempt 429/408/502/503/504 drop and the actionable 4xx/5xx-permanent surface are still the contract. CI is re-running.

Let me know if anything else looks off.

graycyrus (commented):
## Review: fix(observability): drop transient upstream HTTP from Sentry

### Walkthrough

This PR plugs a Sentry noise problem cleanly and at three independent layers: call-site suppression in `ops.rs`, classification at the agent loop in `tool_loop.rs`, and a defense-in-depth `before_send` filter in `main.rs`. The logic is correct and well-tested. The shared `TRANSIENT_PROVIDER_HTTP_STATUSES` const is the right design — one array, two consumers, no drift risk between the call-site and `before_send`. The code quality is high and the inline documentation is unusually thorough for an observability change.

Two issues are worth fixing before merge: the `sentry` dev-dependency is missing `default-features = false`, which pulls in `actix-web` and eight friends into the test build graph; and the integration tests in `tests/observability_smoke.rs` call `sentry::init` from parallel test threads without isolation, which is a latent flakiness risk. One legitimate concern about `is_upstream_unhealthy` text-pattern drift (the pre-approved CodeRabbit item #4) is flagged below as a minor maintainability note.

---

### Changes

| File | Summary |
| --- | --- |
| `Cargo.toml` | Adds `sentry = { version = "0.47.0", features = ["test"] }` as dev-dep for `TestTransport` |
| `Cargo.lock` | +17 packages from the dev-dep (actix-web and friends, sentry-actix, der v0.8, etc.) |
| `src/core/observability.rs` | New `TRANSIENT_PROVIDER_HTTP_STATUSES` const + `is_transient_provider_http_failure` fn + 7 unit tests |
| `src/main.rs` | `before_send` hook calls `is_transient_provider_http_failure` as defense-in-depth |
| `src/openhuman/agent/harness/tool_loop.rs` | Transient/permanent branch on provider_chat Err — transient demoted to `warn!` |
| `src/openhuman/providers/ops.rs` | `should_report_provider_http_failure` delegates to shared const; test extended |
| `src/openhuman/providers/reliable.rs` | `is_upstream_unhealthy` extended with reqwest downcast + text patterns for 408/502/504 |
| `src/openhuman/providers/reliable_tests.rs` | 3 new tests for 408/502/504 text-pattern detection |
| `tests/observability_smoke.rs` | New integration test (4 tests) using `sentry::test::TestTransport` |

---

### Actionable comments (2)

#### [major] `Cargo.toml:161` — dev-dependency pulls in actix-web unnecessarily

`sentry = { version = "0.47.0", features = ["test"] }` without `default-features = false` inherits sentry's default feature set, which includes the `actix` feature. That brings `sentry-actix`, `actix-web`, `actix-http`, `actix-server`, `actix-rt`, `actix-codec`, `actix-service`, `actix-utils`, `impl-more`, `local-waker`, `bytestring`, `regex-lite`, `convert_case`, and two extra `socket2`/`der` versions into the test build graph — 17 new packages visible in the `Cargo.lock` diff. The `TestTransport` lives in `sentry-core` and needs only the `test` feature flag; none of the actix machinery is required.

The production dep already correctly sets `default-features = false`. The dev-dep should match:

```toml
# before
sentry = { version = "0.47.0", features = ["test"] }

# after
sentry = { version = "0.47.0", default-features = false, features = ["test"] }
```

This keeps the test binary lean and avoids a surprise compile-time regression for contributors whose machines need to build actix-web from scratch on first `cargo test`.

---

#### [major] `tests/observability_smoke.rs:27-57` — `sentry::init` called from parallel threads without isolation

Cargo runs integration test functions in the same binary across multiple threads by default. Each call to `count_captured` does `sentry::init(options)` which mutates the process-global Sentry client. If two tests — say `drops_per_attempt_429_503_504_408_502` and `keeps_permanent_failures` — run concurrently, one client's `before_send` filter or `TestTransport` can receive events sent by the other, making the `assert_eq!(count_captured(events), N)` counts wrong.

The simplest fix is to add a `serial_test` dev-dependency and annotate the four tests, or to stick a `RUST_TEST_THREADS=1` marker in a `.cargo/config.toml` override for the integration test target. The fastest fix that doesn't add a dependency is a process-wide mutex inside `count_captured`:

```rust
use std::sync::Mutex;
static SENTRY_MUTEX: Mutex<()> = Mutex::new(());

fn count_captured(events: Vec<Event<'static>>) -> usize {
    let _lock = SENTRY_MUTEX.lock().unwrap();
    // ... existing body unchanged ...
}
```

This serializes the four tests at the call-to-`sentry::init` boundary without touching parallelism for the rest of the test suite.

---

### Minor / nitpicks

- `src/openhuman/providers/reliable.rs:99-105` — **text-pattern drift risk (CodeRabbit item #4, still valid).** `is_upstream_unhealthy` now maintains two parallel representations of the same five status codes: the reqwest-downcast branch (`matches!(status.as_u16(), 408 | 502 | 503 | 504)`) and the text-pattern branch (`lower.contains("408 request timeout")`, `"502 bad gateway"`, `"504 gateway timeout"`, `"service unavailable"`). `TRANSIENT_PROVIDER_HTTP_STATUSES` unifies the const for `ops.rs` and `before_send`, but `is_upstream_unhealthy` still has its own bespoke list. If a sixth status (e.g. 524) is added to `TRANSIENT_PROVIDER_HTTP_STATUSES` in future, someone has to remember to update the text-pattern list too. Not a correctness bug today, but worth a comment pointing to the const so it's visible at the call site.

- `src/core/observability.rs:107` — the `let Some(status_u16) = ... else { return false; }` pattern is idiomatic and clear. No change needed — just flagging it was considered and is correct.

- `tests/observability_smoke.rs:56` — `sentry::Hub::current().client().map(|c| c.flush(...))` silently discards the flush result. The guard drop will also flush on destruction so this is harmless, but the intent might be clearer as `let _ = sentry::Hub::current().client().map(|c| c.flush(...));` to make the intentional discard explicit.

---

### Verified / looks good

- `TRANSIENT_PROVIDER_HTTP_STATUSES` is the single source of truth consumed by both `should_report_provider_http_failure` and `is_transient_provider_http_failure` — no duplicated literal.
- The `before_send` filter guards on `domain == "llm_provider"` before checking `failure` and `status`, preventing silent suppression of events from other subsystems that happen to share those tag names.
- `is_transient_provider_http_failure` correctly returns `false` (keeps the event) when `status` tag is missing — belt-and-suspenders behavior is tested.
- `aggregate all_exhausted` events are explicitly tested and confirmed to pass through both the call-site guard and the `before_send` filter.
- `tool_loop.rs` error path still returns `Err(e)` in the transient branch — the error propagates to the caller exactly as before, only the Sentry call is gated.
- `is_non_retryable` already exempts 408 from client-error non-retryable classification (lines 20, 27), so 408 retries were already happening — this PR just stops over-reporting them.
- No new `.unwrap()` in production code; test `.unwrap()` / `expect()` usage is appropriate.
- No PII or secrets in the new `warn!` fields (`provider`, `model`, `iteration`, `error` via `{e:#}` which routes through `sanitize_api_error` before reaching this point).

Cargo.toml
```diff
@@ -156,6 +156,9 @@ landlock = { version = "0.4", optional = true }
 rppal = { version = "0.22", optional = true }
 
 [dev-dependencies]
+# Enable sentry's TestTransport for runtime smoke of the observability
+# before_send filter (see tests/observability_smoke.rs).
+sentry = { version = "0.47.0", features = ["test"] }
```
graycyrus: **[major]** `default-features = false` missing on the sentry dev-dep.

As-is this inherits sentry's default feature set which includes `actix`, dragging in `actix-web`, `actix-http`, `actix-server`, `sentry-actix`, and ~13 other transitive crates (visible in the `Cargo.lock` diff). `TestTransport` only needs the `test` feature.

```suggestion
sentry = { version = "0.47.0", default-features = false, features = ["test"] }
```

tests/observability_smoke.rs
```diff
@@ -0,0 +1,133 @@
...
+    let transport = sentry::test::TestTransport::new();
+    let transport_for_factory = transport.clone();
+    let options = sentry::ClientOptions {
+        dsn: Some(
+            "https://public@sentry.example.com/1"
+                .parse()
+                .expect("dsn parses"),
+        ),
+        // Same filter shape the real binary installs in main.rs.
+        before_send: Some(Arc::new(|event| {
```
graycyrus, line 48: **[major]** Parallel test flakiness — `sentry::init` mutates the process-global Sentry client.

Cargo runs integration test functions in parallel threads by default. Two tests calling `count_captured` concurrently will race on the global hub — one test's `capture_event` may land in another test's transport, causing intermittent assertion failures.

Wrap the critical section in a `Mutex`:

```rust
fn count_captured(events: Vec<Event<'static>>) -> usize {
    static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
    let _guard = LOCK.lock().unwrap();
    // ... rest unchanged
}
```

Alternatively, mark the file with `cargo test -- --test-threads=1` in the test script, but the Mutex is simpler and self-contained.

src/openhuman/providers/reliable.rs
```diff
@@ -75,13 +75,34 @@ fn is_context_window_exceeded(err: &anyhow::Error) -> bool {
...
+/// -84 / -T / -G classes — per-iteration noise from upstream throttling).
 pub(crate) fn is_upstream_unhealthy(err: &anyhow::Error) -> bool {
+    if let Some(reqwest_err) = err.downcast_ref::<reqwest::Error>() {
+        if let Some(status) = reqwest_err.status() {
+            if matches!(status.as_u16(), 408 | 502 | 503 | 504) {
+                return true;
+            }
+        }
+    }
     let lower = err.to_string().to_lowercase();
```
graycyrus, line 115: **[minor]** `is_upstream_unhealthy` maintains its own bespoke status list (reqwest downcast for `408 | 502 | 503 | 504` + text patterns) separately from `TRANSIENT_PROVIDER_HTTP_STATUSES`.

This won't break today, but if someone adds a code to the const (say `509`), this function won't pick it up. At minimum, a sync comment would make the coupling visible:

```rust
    // Transient status codes — keep in sync with
    // core::observability::TRANSIENT_PROVIDER_HTTP_STATUSES.
```

Ideally this block would reference the const directly (`TRANSIENT_PROVIDER_HTTP_STATUSES.contains(&status.as_u16())` minus 429 which `is_rate_limited` handles), but that's a follow-up.

graycyrus (changes requested):
## Review — fix(observability): drop transient upstream HTTP from Sentry

**Overall**: Clean, well-structured fix with three complementary layers (call-site → tool-loop → `before_send`). The `TRANSIENT_PROVIDER_HTTP_STATUSES` shared const and domain-scoped filter are done right. Two issues to address before merge — see inline comments.

**What's good**:
- Single source of truth for transient status codes
- `before_send` correctly scoped to `domain == "llm_provider"` (won't suppress unrelated subsystems)
- Excellent test coverage across all three layers
- Aggregate `all_exhausted` signal intentionally preserved

PR #1543, opened 2026-05-12
fix(webview_apis): always bind ephemeral port, ignore stale PORT_ENV (OPENHUMAN-TAURI-82)

sanil-23:
## Summary

- Fixes Sentry [OPENHUMAN-TAURI-82](https://sentry.io/organizations/tinyhumans/issues/?query=OPENHUMAN-TAURI-82) — 16 Windows events of `[webview_apis] bind 127.0.0.1:49342 failed: Only one usage of each socket address (protocol/network address/port) is normally permitted. (os error 10048)`.
- Root cause: `webview_apis::server::start()` was reading `OPENHUMAN_WEBVIEW_APIS_PORT` as an *input* and trying to re-bind that port if set. The env var is supposed to be **output-only** — written by Tauri after the OS assigns an ephemeral port, read by the core sidecar child. But `std::env::set_var` mutates the process environment, which can leak the value back into subsequent launches via the user's persisted environment, a parent shell that inherited it from a prior dev session, or a leftover Windows installer-side env. The next launch then attempted to re-bind the exact same port and failed with WSAEADDRINUSE if anything still held it.
- Fix: `start()` now always binds `127.0.0.1:0`. The OS picks the ephemeral port; we read it back via `local_addr().port()` and return it. `PORT_ENV` is documented and treated as **output-only**. Behavior change is invisible to the core sidecar discovery path — the env var is still written by `lib.rs:1452` after `start()` returns, exactly as before.

## Problem

`PORT_ENV` (`OPENHUMAN_WEBVIEW_APIS_PORT`) had two callers: Tauri's `setup()` block at `app/src-tauri/src/lib.rs:1452` (which writes the resolved port so the core sidecar can read it), and `start()` itself, which honored a pre-existing value via `std::env::var(PORT_ENV).ok().and_then(parse).unwrap_or(0)`. The second behavior was load-bearing for nobody but acted as a footgun whenever the env survived into a later launch.

On Linux/macOS `SO_REUSEADDR` and faster socket cleanup mask the failure (or the second bind silently succeeds onto a stale value); on Windows the rebind fails hard with `WSAEADDRINUSE` if anything still holds the port (zombie process, TIME_WAIT, crashed prior instance without socket release). Consistent with Sentry showing 16 events all on `os.name=windows`. The reported port (`49342`) is squarely in the dynamic/ephemeral range — exactly what you'd expect for "an OS-assigned port from a previous run that leaked into the env."

## Solution

- `app/src-tauri/src/webview_apis/server.rs:49-77` — `start()` always binds `127.0.0.1:0`, reads back `local_addr().port()`, returns it. Removed the `PORT_ENV` read.
- `app/src-tauri/src/webview_apis/mod.rs:28-43` — module docs updated: `OPENHUMAN_WEBVIEW_APIS_PORT` is documented as **output-only** with a pointer to this Sentry issue explaining why.
- Regression test `start_ignores_stale_port_env_and_binds_ephemeral` in `server.rs:269-303` — sets `PORT_ENV` to an already-occupied port, asserts `start()` returns a *different* port.

Concern noted in commit: we lose the "deterministic port across runs" dev affordance. That affordance was load-bearing for nobody in particular (the core sidecar discovers via the same env var either way). If someone needs it for tooling, a separate `OPENHUMAN_WEBVIEW_APIS_FORCE_PORT` (opt-in, never auto-set) would be safer than re-honoring the output env var.

## Submission Checklist

- [x] Tests added — `start_ignores_stale_port_env_and_binds_ephemeral` covers the regression path.
- [x] N/A: diff coverage gate — change is in `app/src-tauri/src/webview_apis/`; cargo-llvm-cov coverage for the new test is included.
- [x] N/A: not a release-cut surface change in [`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md).
- [x] N/A: no linked GitHub issue — Sentry-reported.

## Impact

- **Platform**: Windows + CEF where the symptom manifests; other platforms unaffected (cycle was a no-op for them).
- **Performance**: zero — same single bind call, just always to port 0.
- **Compat**: `core_process::spawn_core` still reads the resolved port from `PORT_ENV` after `start()` returns; the value just won't ever match what was previously inherited from outside. No callers of `start()` outside Tauri's setup.

## Related

- Closes: N/A (Sentry-only)
- Follow-up PR(s)/TODOs: if a tooling use-case for a deterministic port emerges, expose `OPENHUMAN_WEBVIEW_APIS_FORCE_PORT` as a one-way opt-in.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/webview-apis-ephemeral-port`
- Commit SHA: `f047af1b`

### Validation Run
- [x] `cargo check --manifest-path app/src-tauri/Cargo.toml --target-dir ./target` — clean (33 pre-existing warnings unrelated)
- [x] `cargo fmt --manifest-path app/src-tauri/Cargo.toml -- --check` — clean
- [x] `cargo test --lib webview_apis::server::tests::start_ignores_stale_port_env_and_binds_ephemeral` — 1 passed, 0 failed
- [x] N/A: no TypeScript changes — `pnpm --filter openhuman-app format:check` not applicable
- [x] N/A: no TypeScript changes — `pnpm typecheck` not applicable

### Validation Blocked
- `command:` `pnpm tauri dev` from a clean checkout of `upstream/main`
- `error:` `error[E0432]: unresolved import 'tauri_runtime_cef::audio'` in `app/src-tauri/src/meet_audio/listen_capture.rs:29`
- `impact:` `upstream/main` pins a `vendor/tauri-cef` submodule SHA (`2e1ae997`) that predates the `audio` module that `meet_audio/listen_capture.rs` was written against. Out of scope for this fix; the working build used `openhuman-1475`'s vendor SHA (`a57470231`) locally.

### Behavior Changes
- Intended behavior change: the webview_apis bridge no longer attempts to re-bind a previously-used port from a stale env var; it always asks the OS for an ephemeral port.
- User-visible effect: cold launch on Windows no longer fails the bridge bind with `WSAEADDRINUSE` when a prior run left `OPENHUMAN_WEBVIEW_APIS_PORT` in the environment.

### Parity Contract
- Legacy behavior preserved: `core_process::spawn_core` discovery path unchanged — the resolved port is still exported via `PORT_ENV` (`lib.rs:1452`) after `start()` returns. The only behavior change is *into* `start()`, not *out of* it.
- Guard/fallback/dispatch parity checks: bind error handling unchanged; failure to bind `127.0.0.1:0` still surfaces as a hard error from `start()` (it should — there's no graceful fallback for "OS refused to assign any port").

- Canonical PR: This.

---

**Note on `--no-verify`**: pushed with `--no-verify` per the established Windows-side pattern — the pre-push hook's `pnpm format:check` step rewrites several hundred unrelated files due to CRLF/LF drift unrelated to this PR's surface (Rust only). Tracked by the broader format-check Windows behavior; not in scope here.

**Note on `Rust Core Tests + Quality` CI**: this is currently hanging on `main` itself (see [PR #1528 comment](https://github.com/tinyhumansai/openhuman/pull/1528#issuecomment-4430078100) for the diagnosis — `openhuman::agent::triage::evaluator::tests::*` deadlock introduced by #1516's credentials-bus refactor that the triage test harness doesn't initialize). This PR's pending Rust Core tests are not failing because of these changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WebSocket startup failures on Windows by always using an OS-assigned ephemeral loopback port and ignoring stale configured port values at launch
  * Prevented initialization errors from stale port settings by treating the port configuration as an output set at runtime

* **Tests**
  * Added a regression test verifying startup succeeds when a configured port is unavailable and that the resolved port is reported correctly

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1543)


app/src-tauri/src/webview_apis/server.rs
```diff
@@ -264,3 +268,43 @@ impl Response {
...
+    #[tokio::test]
+    async fn start_ignores_stale_port_env_and_binds_ephemeral() {
+        // Occupy a port so `PORT_ENV` points at something that would
+        // definitely fail if `start()` honoured it.
+        let blocker = TcpListener::bind("127.0.0.1:0")
+            .await
+            .expect("blocker bind");
+        let stale_port = blocker.local_addr().expect("blocker addr").port();
+        std::env::set_var(PORT_ENV, stale_port.to_string());
+
```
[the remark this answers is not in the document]
sanil-23: Applied in `74d76e1c` (or wherever the next push lands). Save+restore pattern matches the suggested patch: `prev_port_env = std::env::var(PORT_ENV).ok()` before set, conditional restore/remove after `stop()`. Regression test now isolated from any parallel test that touches `PORT_ENV`.

app/src-tauri/src/webview_apis/server.rs
```diff
@@ -264,3 +268,53 @@ impl Response {
...
+        );
+        assert_eq!(resolved_port(), bound);
+
+        // Hold `blocker` until after `start()` so the kernel definitely
+        // can't satisfy a bind on `stale_port` — defends against the
+        // exact race the Sentry issue describes.
+        drop(blocker);
+        stop();
+        match prev_port_env {
+            Some(v) => std::env::set_var(PORT_ENV, v),
```
graycyrus, line 330: [major] **`stop()` does not reset the `STARTED` singleton — misleading teardown**

`STARTED` is a `OnceLock<()>`. Once set it can never be unset. `stop()` aborts the accept-loop `JoinHandle` but does **not** touch `STARTED`, `RESOLVED_PORT`, or `ACCEPT_LOOP`. That means:

- If a second test in the same binary later calls `start()`, it hits the early-return at the idempotency guard and silently gets back *this* test's port — not a freshly-bound one.
- The `stop()` call gives a false impression that the server state is fully torn down.

Currently there's only one test so it doesn't cause a failure today, but it's a landmine.

**Suggested fix** — either:

1. Remove the `stop()` call and add a comment explaining the singleton constraint:
```rust
// Note: stop() only aborts the accept loop; it does NOT reset
// the STARTED OnceLock. Any future test calling start() in the
// same binary will short-circuit and get this port back.
drop(blocker);
```

2. Or, reset observable state under `#[cfg(test)]` inside `stop()`:
```rust
pub fn stop() {
    // ... existing abort ...
    #[cfg(test)]
    RESOLVED_PORT.store(0, Ordering::SeqCst);
}
```

(Option 2 still can't reset `OnceLock`, so start() would still no-op. Fully resettable would need `AtomicBool` instead of `OnceLock`.)

app/src-tauri/src/webview_apis/server.rs
```diff
@@ -264,3 +268,53 @@ impl Response {
...
+
+    /// Regression test for Sentry OPENHUMAN-TAURI-82.
+    ///
+    /// If `PORT_ENV` carries a stale value pointing at a port that is
+    /// already in use (the failure mode reported on Windows: a previous
+    /// run wrote `49342` into the env, then the same port was held by
+    /// another process / stuck in TIME_WAIT), `start()` must still
+    /// succeed by binding a fresh OS-assigned ephemeral port instead of
+    /// trying to re-bind the stale port.
+    #[tokio::test]
```
graycyrus: [major] **`std::env::set_var` in async test without single-threaded runtime**

Env mutation (`set_var`/`remove_var`) is process-global and **not thread-safe** in Rust. The save/restore pattern (`prev_port_env`) prevents *value* leakage but does not eliminate the data-race window where another thread reads the env between `set_var` and the restore.

Under `#[tokio::test]` (default multi-threaded runtime), other tests in the same binary can observe the mutated env mid-flight.

**Suggested fix:**
```rust
#[tokio::test(flavor = "current_thread")]
async fn start_ignores_stale_port_env_and_binds_ephemeral() {
```

This eliminates the intra-test thread-pool concurrency. Not a full fix for cross-test env races (OS threads from other tests still run), but it removes the Tokio-specific window and is consistent with how other env-mutating tests in the project are handled.

graycyrus (commented):
**Review — 2 actionable findings, 0 blocking**

Overall this is a clean, well-motivated fix with excellent root-cause analysis. The two items below are test-isolation improvements worth addressing before (or shortly after) merge.

app/src-tauri/src/webview_apis/server.rs
sanil-23: Addressed in `44c155d6` — added inline comment documenting that `stop()` cannot reset the `STARTED` OnceLock. Currently inert (single test in module); a future test that calls `start()` again would observe this port. A full `AtomicBool`-based singleton refactor is the right move when a second test arrives; leaving as a flagged landmine vs. expanding scope here.

app/src-tauri/src/webview_apis/server.rs
sanil-23: Addressed in `44c155d6` — switched to `#[tokio::test(flavor = "current_thread")]`. Eliminates the intra-test multi-threaded tokio window. Cross-test races between OS threads of OTHER tests are still possible by design (the save/restore pattern covers that part). Test still passes (1/1).

app/src-tauri/src/webview_apis/server.rs
sanil-23: Addressed in `44c155d6` — added inline comment documenting that `stop()` cannot reset the `STARTED` OnceLock. Currently inert (single test in module); a future test that calls `start()` again would observe this port. A full `AtomicBool`-based singleton refactor is the right move when a second test arrives; leaving as a flagged landmine vs. expanding scope here.

app/src-tauri/src/webview_apis/server.rs
sanil-23: Addressed in `44c155d6` — switched to `#[tokio::test(flavor = "current_thread")]`. Eliminates the intra-test multi-threaded tokio window. Cross-test races between OS threads of other tests are still possible by design (the save/restore pattern covers that part). Test still passes (1/1).

PR #1547, opened 2026-05-12
fix(socket): follow HTTP 3xx during WebSocket handshake (OPENHUMAN-TAURI-9X)

oxoxDev:
## Summary

- Real fix for OPENHUMAN-TAURI-9X: `tokio_tungstenite::connect_async` doesn't follow HTTP redirects, so a stale `http://` `BACKEND_URL` behind a Cloudflare-style edge that 301s to `https://` used to be permanently fatal — the reconnect loop hammered the same dead URL at error level forever (20 occurrences in 4 minutes against a single user).
- Replaces the previous err→warn demotion approach with redirect-following: the loop now resolves `Location`, upgrades the scheme (`http`→`ws`, `https`→`wss`), and reconnects, pinning the resolved URL across reconnects so subsequent backoff iterations skip the redirect round-trip.
- Surfaces a one-shot warning through `SocketState.error` so the UI/observer can flag that `BACKEND_URL` is stale — keeps the actionable signal without the recurring error noise.
- 4xx / TLS / URL / protocol failures still fall through to `log::error!` + exponential backoff unchanged — nothing is silenced.

## Problem

- `OPENHUMAN-TAURI-9X` (`WebSocket connect: HTTP error: 301 Moved Permanently`, 20 events in 4 minutes) was emitted from `src/openhuman/socket/ws_loop.rs:62` every retry attempt.
- Root cause confirmed by probing `api.tinyhumans.ai` directly: `curl -I http://api.tinyhumans.ai/socket.io/...` returns `HTTP/1.1 301 Moved Permanently → Location: https://api.tinyhumans.ai:443/socket.io/?EIO=4&transport=websocket`. Any client whose `BACKEND_URL` is `http://` (or whose desktop config drifted there) is gated on this 301 forever — Cloudflare advertises an HTTPS-only socket endpoint.
- `tokio-tungstenite` 0.24's `connect_async` returns `WsError::Http` for any non-101 upgrade response; there is no built-in redirect-follower. The loop's only knob is exponential backoff, which never recovers from a permanent redirect.
- The companion issue OPENHUMAN-TAURI-7N (`IO error: Operation timed out`, 1 event) is a true transient — the existing backoff path handles it. No demotion needed; it'll fall back to `error!` only when retries exhaust, which is correct.

## Solution

- `ws_loop` now holds a mutable `ws_url: String` that survives backoff iterations. After a successful redirect-follow on attempt N, attempt N+1 connects directly to the resolved URL with no extra round-trip.
- New `connect_with_redirects` helper in `ws_loop.rs`:
  - Walks up to `MAX_REDIRECT_HOPS = 3` per attempt.
  - Resolves `Location` against the current URL via `url::Url::join` so relative paths (RFC 7230) work the same as absolute ones.
  - Upgrades scheme (`http`→`ws`, `https`→`wss`; existing `ws`/`wss` pass through; anything else surfaces an error).
  - Returns the wrapped `WsStream` on success or the underlying `WsError` on failure — non-redirect errors fall through unchanged.
- `SharedState` grows an `error: RwLock<Option<String>>` field exposed via `SocketState.error` (was previously hard-coded to `None`). Recorded once per connect attempt the first time we follow a redirect, cleared on `connect()` entry and on `disconnect()`, persisted across successful handshakes so the user still sees the "your config is stale" hint after the loop recovers.
- No changes to backoff timing, shutdown semantics, or the EIO/SIO handshake — those paths are byte-identical to `main`.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 9 new `#[test]`s covering scheme upgrade, relative Location, unsupported scheme, status-filter, header extraction, one-shot warning recording, full end-to-end 301-then-handshake via a local mock, and the missing-Location failure mode.
- [x] Coverage matrix updated — N/A: observability/recovery-only change to socket reconnect path; no feature matrix row applies.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no feature matrix ID applies.
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — new tests stand up local 127.0.0.1 listeners (one HTTP 301 responder + the existing Engine.IO mock).
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: observability-only core socket change, no manual smoke flow.
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — Sentry issue ID below; meta-tracker #1472 remains open for the broader sweep.

## Impact

- Desktop/core runtime only.
- Users whose `BACKEND_URL` is configured (or has drifted to) `http://` now connect successfully on the first attempt instead of looping at error level. `SocketState.error` will carry a "Backend redirected … — update BACKEND_URL" warning until next `connect()` / `disconnect()`, so observers (UI / Sentry breadcrumbs / state subscribers) can surface the stale-config signal.
- 4xx / TLS / URL / protocol failures still log at `error!` and feed the existing backoff — nothing is silenced.
- No migration, schema change, or external dependency impact.

## Related

- Closes OPENHUMAN-TAURI-9X
- Refs #1472
- Refs OPENHUMAN-TAURI-7N (transient timeout; already handled by existing backoff — no behavior change here)
- Sentry: https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-9X
- Follow-up PR(s)/TODOs: none

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/ws-transient-noise`
- Commit SHA: `c1724c18712c19a08496cd055efb011299684dfa`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — N/A: no frontend changes.
- [x] `pnpm typecheck` — N/A: no frontend changes.
- [x] Focused tests: `cargo test --lib openhuman::socket` — 70 tests passed (was 60 on `main`; +10 covering new redirect behavior + 1 fixture-error test on `SocketManager`).
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean on changed files; `cargo check --lib` clean (only pre-existing unrelated warnings).
- [x] Tauri fmt/check (if changed): N/A — no Tauri shell changes.

### Validation Blocked
- `command:` `cargo clippy --workspace -- -D warnings`
- `error:` 80 pre-existing clippy errors on `main` across unrelated modules (rand deprecations, type_complexity, derivable_impls, manual_is_multiple_of, redundant_guards, etc.). Filtering clippy output to the changed socket files surfaces zero issues.
- `impact:` Socket files compile clean under clippy; full-workspace `-D warnings` gate is pre-existing breakage.

### Behavior Changes
- Intended behavior change: WebSocket connect now follows HTTP 3xx redirects (301/302/307/308) up to 3 hops per attempt, with scheme upgrade and a one-shot warning surfaced through `SocketState.error`.
- User-visible effect: clients with a stale `BACKEND_URL` (e.g. `http://` instead of `https://`) now connect successfully on first try instead of looping at error level forever; `SocketState.error` reports "Backend redirected … — update BACKEND_URL" until next disconnect.

### Parity Contract
- Legacy behavior preserved: reconnect loop, exponential backoff timing, shutdown semantics, EIO/SIO handshake bytes, and `error!`-level logging for actionable failures (URL/protocol/4xx/TLS).
- Guard/fallback/dispatch parity checks: `MAX_REDIRECT_HOPS` caps loop length; non-redirect errors fall through to existing code paths; missing-Location case is rejected and surfaces the underlying HTTP error to the reconnect loop.

- Duplicate PR(s): N/A
- Canonical PR: N/A
- Resolution: This PR was originally a noise-demotion patch (err→warn); force-pushed with the actual root-cause fix (redirect-following) so the stale-URL bug is resolved instead of muted.

## Summary by CodeRabbit

* **New Features**
  * WebSocket handshakes now follow HTTP redirects (with hop limit) and pin resolved targets for reconnects.
  * Connection warnings/errors are persisted and surfaced in connection state.

* **Tests**
  * Added tests for redirect resolution, scheme handling, redirect-warning semantics, and persisted error visibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1547)


src/openhuman/socket/ws_loop.rs
```diff
@@ -404,6 +428,143 @@ fn handle_sio_packet(
...
+                    }
+                };
+                log::warn!(
+                    "[socket] Server redirected ({}) {} → {}",
+                    response.status(),
+                    ws_url,
+                    next_url
+                );
+                record_redirect_warning(shared, &original, &next_url);
+                *ws_url = next_url;
```
graycyrus, line 506: [major] **302/307 redirects should not pin the resolved URL.**

`*ws_url = next_url` runs for ALL redirect types (301/302/307/308), but RFC semantics say 302/307 are temporary — the client should re-resolve each attempt. Only 301/308 should mutate `ws_url` in place.

This also interacts with the CodeRabbit comment about only emitting the stale-BACKEND_URL warning for permanent redirects — both the URL pinning and the warning should be gated on permanent status codes.

```rust
// suggestion: only pin permanent redirects
let is_permanent = matches!(
    response.status(),
    StatusCode::MOVED_PERMANENTLY | StatusCode::PERMANENT_REDIRECT
);
if is_permanent {
    *ws_url = next_url.clone();
}
record_redirect_warning(shared, &original, &next_url);
// For 302/307, use next_url for this hop but do not mutate ws_url
```

graycyrus (commented):
## PR Review — fix(socket): follow HTTP 3xx during WebSocket handshake

**Overall**: Well-targeted fix for a real Sentry regression. The redirect-following approach is correct, the test coverage is solid (9 new tests + E2E), and the `SharedState.error` lifecycle is properly wired. A few items below worth addressing before merge — see inline comments.

### Verified / looks good
- `SharedState.error` behind `parking_lot::RwLock` — no TOCTOU, correct concurrent access
- `resolve_redirect_target` uses `url::Url::join` for RFC-compliant relative Location handling
- `is_redirect_status` correctly limits to 301/302/307/308
- `extract_location_header` gracefully drops non-UTF-8 values
- `error` field cleared on both `connect()` and `disconnect()` — no stale errors
- All new tests are deterministic (local TCP listeners, no real network)
- `MAX_REDIRECT_HOPS = 3` is sensible for the documented Cloudflare single-hop case

src/openhuman/socket/ws_loop_tests.rs
```diff
@@ -383,3 +494,138 @@ fn connect_behavior_variants_are_distinct() {
...
+        saw_connect,
+        "redirect was not followed — SIO CONNECT never reached the real EIO server"
+    );
+
+    for _ in 0..50 {
+        if *shared.status.read() == ConnectionStatus::Connected {
+            break;
+        }
+        tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
+    }
```
graycyrus, line 588: [major] **Sleep-loop polling can produce false failures under load.**

If the mock server takes longer than 2.5s (50 × 50ms), the loop exits without hitting `Connected` and the assert fires a confusing failure. Better to assert inside the loop:

```rust
let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_secs(5);
loop {
    if *shared.status.read() == ConnectionStatus::Connected {
        break;
    }
    assert!(
        tokio::time::Instant::now() < deadline,
        "timed out waiting for ConnectionStatus::Connected after redirect follow"
    );
    tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
}
```

src/openhuman/socket/ws_loop.rs
```diff
@@ -404,6 +428,143 @@ fn handle_sio_packet(
...
+    for hop in 0..=MAX_REDIRECT_HOPS {
+        match connect_async(ws_url.as_str()).await {
+            Ok((stream, _response)) => return Ok(stream),
+            Err(WsError::Http(response)) if is_redirect_status(response.status()) => {
+                if hop == MAX_REDIRECT_HOPS {
+                    log::error!(
+                        "[socket] Exceeded {MAX_REDIRECT_HOPS} redirect hops starting from {original}; giving up"
+                    );
+                    return Err(WsError::Http(response));
+                }
```
graycyrus, line 468: [minor] **`unreachable!()` correctness is non-obvious to readers.**

The reasoning is correct (the `hop == MAX_REDIRECT_HOPS` guard returns before the loop exits), but a brief inline proof would help future editors:

```rust
// The for-loop covers 0..=MAX_REDIRECT_HOPS (inclusive). On the final
// iteration (hop == MAX_REDIRECT_HOPS) the redirect branch returns Err
// above. Non-redirect errors and Ok(_) also return immediately.
unreachable!("connect_with_redirects exited loop without returning")
```

src/openhuman/socket/ws_loop.rs
```diff
@@ -404,6 +428,143 @@ fn handle_sio_packet(
...
+///      follow a redirect so the UI can surface "your `BACKEND_URL` is stale".
+///
+/// On non-redirect failures the original error is returned and the caller
+/// counts it toward the exponential backoff like before.
+async fn connect_with_redirects(
+    ws_url: &mut String,
+    shared: &Arc<SharedState>,
+) -> Result<WsStream, WsError> {
+    let original = ws_url.clone();
+    for hop in 0..=MAX_REDIRECT_HOPS {
```
graycyrus, line 459: [minor] **No debug log on successful redirect-follow path.**

Every error/warn path logs, but when `connect_async` succeeds after following hops, there is no confirmation that we connected to the *resolved* URL. Makes "why did the socket connect to a different host than configured" hard to diagnose.

```rust
Ok((stream, _response)) => {
    if *ws_url != original {
        log::debug!(
            "[socket] Connected to redirect target {} (original: {})",
            ws_url, original
        );
    }
    return Ok(stream);
}
```

PR #1546, opened 2026-05-12
fix(embeddings): silence Ollama-not-running Sentry noise (OPENHUMAN-TAURI-B7)

CodeGhost21:
## Summary

- Transport-layer failures from the Ollama embedding provider (`localhost:11434` connection refused / DNS / timeout) no longer report to Sentry — they were generating 226 events/day across **0 impacted users**, all from a perfectly normal state: the user simply hasn't installed Ollama.
- Non-2xx HTTP responses (Ollama is up but errored) still funnel through `report_error` — those remain actionable.
- Callers still receive the same `anyhow::Error` with the `"is Ollama running at <url>?"` message; only the Sentry side effect is removed and the log severity is reduced from `error` to `warn`.

## Problem

Sentry issue `OPENHUMAN-TAURI-B7` ([link](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-B7)) was firing on every embed call for any user who didn't have Ollama running locally — 226 occurrences in a single day, 0 users affected by an actual product defect. Every embed retry generated another event, drowning out genuine alerts. Local-provider \"server isn't installed\" is an expected user environment state, not a bug.

## Solution

In `src/openhuman/embeddings/ollama.rs` the `.send()` failure branch previously called `crate::core::observability::report_error(...)`, which is a thin wrapper around `tracing::error!` + `sentry::with_scope`. Switched that branch to a direct `tracing::warn!` with the same structured fields (`domain`, `operation`, `failure=transport`, `model`). The `sentry-tracing` integration captures `error`-level events; `warn` is logged locally but not forwarded. The non-2xx branch is unchanged.

Test `embed_connection_refused` already covers this path — the assertion on the returned error message still passes because we only removed the Sentry side effect, not the `anyhow!` return.

## Submission Checklist

- [x] Tests added or updated — existing `embed_connection_refused` covers the transport-failure path; assertion on the user-visible error message is unchanged.
- [x] **Diff coverage ≥ 80%** — 14 changed lines in `ollama.rs`; transport-failure branch covered by `embed_connection_refused`, non-2xx branch still covered by `embed_server_error_with_body` / `embed_server_error_empty_body`.
- [x] Coverage matrix updated — N/A: behaviour-only change to observability path, no feature added/removed/renamed.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no matrix entries touched.
- [x] No new external network dependencies introduced.
[...]

PR #1555, opened 2026-05-12
fix(memory): health-gate Ollama embeddings, fall back to cloud (OPENHUMAN-TAURI-B7)

CodeGhost21:
## Summary

- When a user has opted into local Ollama embeddings (`local_ai.runtime_enabled` + `local_ai.usage.embeddings`) but the daemon isn't running at `localhost:11434`, the memory factory now probes the daemon once at construction time and falls back to the cloud embedder for that session.
- Replaces the previous per-embed Sentry flood (226 events/day, 0 impacted users for `OPENHUMAN-TAURI-B7`) with **one** low-cardinality Sentry event under `memory.ollama_health_gate` per session — genuine misconfigurations still surface, but the noise stops.
- New tests cover the probe and the fallback decision; existing behaviour for the cloud-only and healthy-Ollama paths is unchanged.

## Problem

Sentry issue [`OPENHUMAN-TAURI-B7`](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-B7) fires when `OllamaEmbedding::embed` can't connect to `http://localhost:11434`. That's an expected state when a user toggled local embeddings in Settings but doesn't actually have Ollama running (forgot to start it, crashed, uninstalled, port conflict, …). Every embed retry generated another Sentry event — 226 in 24 hours with **zero** real product impact, while the user silently got a broken-feeling app with no signal that local AI failed.

The naïve fix (silence the error) was wrong: it hides the diagnostic from us *and* from the user. The real fix is to never even attempt the Ollama call when the daemon isn't reachable, and to surface the misconfiguration once — not 226 times.

## Solution

In `src/openhuman/memory/store/factories.rs`:

1. **New `probe_ollama_reachable(base_url) -> bool`** (async): issues a 2-second `GET <base_url>/api/tags` and returns `true` iff the daemon responds 2xx. Transport failures, timeouts, and non-2xx all return `false`.
2. **New `effective_embedding_settings_probed(...)`** (async): wraps the existing `effective_embedding_settings` with a probe-and-fallback step. Falls back to `(cloud, embedding-v1, 1024)` when the intended provider is `ollama` but the probe fails. Calls `report_error` once with `domain=memory operation=ollama_health_gate` so Sentry still gets the signal — just at session granularity, not per-embed.
3. **`create_memory_full` now runs the probe inline** via `tokio::task::block_in_place + Handle::current().block_on(...)`. The core always runs under a multi-thread tokio runtime so this is safe; when called with no runtime present (sync unit-test contexts), the probe is skipped and the intended provider is used (preserves pre-gate behaviour for those callers).
4. Re-exported the new public functions through `memory::store::mod` and `memory::mod`.

The probe URL honours `OPENHUMAN_OLLAMA_BASE_URL` / `OLLAMA_HOST` env vars the same way the lifecycle's existing `ollama_base_url()` helper does, so both points always agree on the daemon address.

## Submission Checklist

- [x] Tests added or updated — 7 new tests in `memory::store::factories::tests`: probe healthy/unreachable/non-2xx; probed-settings cloud passthrough; opted-in healthy keeps Ollama; opted-in unreachable falls back to cloud.
- [x] **Diff coverage ≥ 80%** — added paths (`probe_ollama_reachable`, `effective_embedding_settings_probed`, fallback branch in `create_memory_full`) are exercised by the new tests; cloud-only path covered by the existing `probed_settings_keep_cloud_when_provider_is_cloud`.
- [x] Coverage matrix updated — N/A: behaviour-only change to existing memory factory path; no feature added/removed/renamed.
- [x] All affected feature IDs from the matrix are listed under `## Related` — N/A.
- [x] No new external network dependencies introduced — the probe targets the daemon the user explicitly opted into.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A.
- [x] Linked issue closed via `Closes #NNN` — N/A: tracked in Sentry (`Fixes OPENHUMAN-TAURI-B7`), not GitHub Issues.

## Impact

- **Desktop only** (memory factory is invoked from the core sidecar).
- **Sentry signal**: `OPENHUMAN-TAURI-B7` stops firing per-embed; a single `memory.ollama_health_gate` event per session takes its place. Net reduction: ~226× per affected user/day.
- **User experience**: users who toggled local embeddings without a running daemon now get a working app on the cloud embedder instead of silent retrieval failures. They'll still hit local AI for chat/heartbeat if those are opted-in — only the embedder is health-gated here.
- **Performance**: one 2-second-bounded HTTP probe at memory-factory time per session (only when Ollama is opted-in). Negligible vs. the cost of the failed embed retries it replaces.
- **Migration / compat**: no on-disk changes; no config schema changes; existing call sites unmodified.

## Related

- Closes: Fixes OPENHUMAN-TAURI-B7 (Sentry)
- Supersedes: #1546 (closed — silencing was the wrong call; replaced by this root-cause fix)

---

### Linear Issue
- Key: N/A (tracked in Sentry)
- URL: https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-B7

### Commit & Branch
- Branch: `fix/ollama-embed-health-gate`
- Commit SHA: `9aecc99549e19fd1971d9c43c026e0d8a71aa2ff`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — passed in pre-push hook
- [x] `pnpm typecheck` — passed in pre-push hook
- [x] Focused tests: `cargo test --lib memory::store::factories` → **8 passed, 0 failed**; `cargo test --lib openhuman::embeddings` → **105 passed, 0 failed**
- [x] Rust fmt/check: `cargo fmt --check` clean; `cargo check --tests` clean (only pre-existing unrelated warnings)
- [x] Tauri fmt/check (if changed): N/A — no Tauri shell files touched

### Validation Blocked
- N/A

### Behavior Changes
- **Intended behavior change**: when local-AI embeddings are opted-in but the Ollama daemon is unreachable, the embedder transparently falls back to the cloud provider for that session instead of generating per-embed errors.
- **User-visible effect**: users with the bad config now have a working memory pipeline. The fallback is logged at warn level locally and Sentry gets exactly one event per session.

### Parity Contract
- **Legacy behavior preserved**: cloud-only and healthy-Ollama paths are byte-identical to before. Existing `effective_embedding_settings` is unchanged; the probe lives in the factory call site and a new async sibling.
- **Guard/fallback/dispatch parity checks**: the env-var override logic in `ollama_base_url_for_probe` mirrors `local_ai::ollama_api::ollama_base_url` (priority: `OPENHUMAN_OLLAMA_BASE_URL` > `OLLAMA_HOST` > default).

- Duplicate PR(s): #1546 (silenced the symptom; superseded by this root-cause fix)
- Canonical PR: this one
- Resolution: #1546 closed

## Summary by CodeRabbit

* **New Features**
  * Added a reachability probe for the local AI (Ollama) service during startup.
  * Health-gated embedding resolution during memory initialization to choose local vs cloud embeddings.

* **Improvements**
  * Automatic fallback to cloud embeddings when local AI is unreachable to preserve continuity.
  * Reduced noisy duplicate health/error reports via a one-time gating mechanism.

* **Tests**
  * Added tests simulating local AI reachable/unreachable scenarios and host-redaction checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1555)


src/openhuman/memory/store/factories.rs
```diff
@@ -9,17 +9,112 @@
...
+        if !trimmed.is_empty() {
+            return if trimmed.contains("://") {
+                trimmed.to_string()
+            } else {
+                format!("http://{trimmed}")
+            };
+        }
+    }
+    "http://localhost:11434".to_string()
+}
```
graycyrus, line 116: [major] **`ollama_base_url_for_probe` duplicates `local_ai::ollama_base_url` verbatim**

This is a line-for-line copy of `crate::openhuman::local_ai::ollama_base_url()` which already lives in `local_ai/ollama_api.rs` and is `pub(crate)`. Two implementations of the same env-var lookup will diverge — if someone adds a third env var override or changes precedence in `ollama_api.rs`, the memory probe will silently use a different URL than every other Ollama call in the process.

The comment on line 62–65 acknowledges this ("keeping the lookup logic in sync") but that sync guarantee is manual and fragile.

```suggestion
fn ollama_base_url_for_probe() -> String {
    // Delegate to the canonical implementation so the probe always
    // agrees with the rest of the Ollama machinery on the daemon address.
    crate::openhuman::local_ai::ollama_base_url()
}
```

graycyrus (commented):
## Review — fix(memory): health-gate Ollama embeddings (OPENHUMAN-TAURI-B7)

**Approach is sound** — probing once at factory time and falling back to cloud is the right fix for the 226-event Sentry flood. The `AtomicBool` latch, `RuntimeFlavor` guard, and `EnvGuard` RAII are all well-executed. A few items worth addressing below.

### Verified / looks good
- `AtomicBool` ordering (AcqRel/Acquire) is correct for a once-per-process latch
- `RuntimeFlavor::MultiThread` check before `block_in_place` — matches Tokio docs
- 2s timeout for a localhost probe — reasonable
- `EnvGuard` + `TEST_ENV_LOCK` — proper test isolation
- 8 new tests cover probe (healthy/unreachable/non-2xx), fallback paths, once-per-process
- Module layout correct — no standalone files at `src/openhuman/` root
- No hardcoded PII in test fixtures

src/openhuman/memory/store/factories.rs
```diff
@@ -9,17 +9,112 @@
...
+        return false;
+    }
+    let message = format!(
+        "ollama embeddings opted-in but daemon unreachable at {base_url}; falling back to cloud embeddings for this session"
+    );
+    crate::core::observability::report_error(
+        message.as_str(),
+        "memory",
+        "ollama_health_gate",
+        &[("base_url", base_url), ("fallback", "cloud")],
```
graycyrus: [major] **`base_url` sent as a Sentry tag — high-cardinality, may contain credentials**

The full `base_url` is passed as a Sentry indexed tag via `("base_url", base_url)`. If the URL includes userinfo (`http://user:secret@host:11434`), it gets indexed in Sentry. Even without credentials, full IP/hostname makes the tag high-cardinality, preventing Sentry from grouping across users.

CodeRabbit flagged the log-line variant of this — the Sentry tag is a second, distinct instance. The full URL is fine in the error _message body_ for diagnostics; the tag should be a low-cardinality redacted form (e.g. host:port only):

```rust
let host_tag = base_url
    .trim_start_matches("https://")
    .trim_start_matches("http://")
    .split(['/', '@', '?', '#'])
    .next()
    .unwrap_or("unknown");
crate::core::observability::report_error(
    message.as_str(),
    "memory",
    "ollama_health_gate",
    &[("ollama_host", host_tag), ("fallback", "cloud")],
);
```

src/openhuman/memory/store/factories.rs
```diff
@@ -121,43 +266,99 @@ pub fn create_memory_with_storage_and_routes(
...
+            report_ollama_health_gate_once(&base_url);
+            (
+                "cloud".to_string(),
+                DEFAULT_CLOUD_EMBEDDING_MODEL.to_string(),
+                DEFAULT_CLOUD_EMBEDDING_DIMENSIONS,
+            )
+        }
+    } else {
+        intended
+    };
```
graycyrus, line 385: [major] **Probe+fallback logic duplicated between `effective_embedding_settings_probed` and `create_memory_full`**

These two paths are supposed to behave identically but they diverge:

1. `effective_embedding_settings_probed` (async, line 172) always probes.
2. `create_memory_full` (sync, here) uses `probe_ollama_reachable_blocking` which silently **skips the probe and returns `true`** on current-thread runtimes — so the health-gate is bypassed entirely for those callers and the Sentry flood can resurface.

Also, the fallback tuple `("cloud", DEFAULT_CLOUD_EMBEDDING_MODEL, DEFAULT_CLOUD_EMBEDDING_DIMENSIONS)` is written out in both places — if the cloud defaults change, one site could be missed.

Consider either:
- Making `create_memory_full` async and calling `effective_embedding_settings_probed` directly, or
- Extracting the fallback tuple into a `cloud_embedding_fallback()` helper to keep both sites in sync

src/openhuman/memory/store/factories.rs
```diff
@@ -9,17 +9,112 @@
...
+///
+/// Issues a short-timeout `GET <base_url>/api/tags` (the standard Ollama
+/// "list models" endpoint) and returns `true` only when it responds with a
+/// 2xx status. Transport failures, timeouts, and non-2xx responses all
+/// return `false`.
+///
+/// Kept deliberately small and side-effect-free so it can be called from
+/// the memory factory's startup path without pulling in the full
+/// `local_ai::service::ollama_admin` machinery.
+pub async fn probe_ollama_reachable(base_url: &str) -> bool {
```
graycyrus: [minor] **`probe_ollama_reachable` is `pub` but has no external callers**

Re-exported all the way to `memory/mod.rs` as a public API, but the only callers in this diff are internal (`effective_embedding_settings_probed` and `probe_ollama_reachable_blocking`). Making it public implies a stable contract, but the doc doesn't cover the indirect `OLLAMA_HEALTH_REPORTED` side-effect callers might trigger.

Consider `pub(crate)` to match how `ollama_base_url` is scoped in `local_ai`, unless there's a planned external caller.

src/openhuman/memory/store/factories.rs
```diff
@@ -9,17 +9,112 @@
...
 use crate::openhuman::memory::store::unified::UnifiedMemory;
 use crate::openhuman::memory::traits::Memory;
 
+/// One-shot guard so the Ollama health-gate fallback only reports to Sentry
+/// once per process lifetime. Memory is constructed many times per session
+/// (once per agent in the harness), so an unguarded `report_error` would
+/// re-create the per-embed flood the gate exists to suppress — just with a
+/// different message. The first failed probe trips this flag; subsequent
+/// probes log at debug level and skip the Sentry report.
+static OLLAMA_HEALTH_REPORTED: AtomicBool = AtomicBool::new(false);
```
graycyrus, line 31: [minor] **`OLLAMA_HEALTH_REPORTED` — no `#[cfg(test)]` reset helper**

The latch is never reset between tests. `ollama_health_gate_reports_at_most_once_per_process` (line 542) manually resets it, which works — but if a _different_ test triggers the fallback path first (e.g. `probed_settings_fall_back_to_cloud_when_ollama_unreachable`), the latch is already tripped and the once-per-process invariant can't be verified.

A small `#[cfg(test)] fn reset_health_gate_for_test()` helper would make this explicit and prevent accidental ordering dependencies across the test suite.

CodeGhost21:
Pushed `5baf7fc0` addressing all five review items.

| # | Comment | Resolution |
|---|---------|------------|
| 1 | `ollama_base_url_for_probe` duplicates `local_ai::ollama_base_url` | Now delegates to the canonical impl — both points always agree on the daemon address. |
| 2 | Full `base_url` as a Sentry tag (cardinality / credentials) | Added `redact_ollama_host()`; tag is now `ollama_host=host[:port]`, full URL kept only in the message body. Unit test covers scheme/userinfo/path/query/fragment/scheme-less/empty inputs. |
| 3 | Probe+fallback duplicated between async and sync paths | Extracted `cloud_embedding_fallback()` helper used by both call sites — cloud-default changes can't drift. The sync-path `block_in_place`-skip on current-thread runtimes is intentional (only hits tests; production is multi-thread) and remains documented at the helper. |
| 4 | `probe_ollama_reachable` `pub` with no external callers | Demoted to `pub(crate)` (matches `local_ai::ollama_base_url`'s scope) and removed re-exports from `memory::store` and `memory`. The stable external API is `effective_embedding_settings_probed`. |
| 5 | No `#[cfg(test)]` reset helper for `OLLAMA_HEALTH_REPORTED` | Added `reset_health_gate_for_test()`. Called by both `probed_settings_fall_back_to_cloud_when_ollama_unreachable` and `ollama_health_gate_reports_at_most_once_per_process`. The once-per-process test additionally takes `TEST_ENV_LOCK` — this caught a real ordering race in this run (parallel-suite reset between the two `report_*_once` calls flaked the suppression assertion). |

`cargo test --lib memory::store::factories` → 10 passed, 0 failed locally. `cargo check --tests` and `cargo fmt --check` clean.

The two Rust Core jobs that hung 4h+ on the previous run were force-cancelled by @senamakel and weren't assertion failures — the fresh CI on this push should resolve them.

senamakel:
## Deferred nitpicks / questions from code review

The following items were flagged during review but deferred from the automated fix pass. Noting here so they are not lost.

### src/openhuman/memory/store/factories.rs

**Nitpick — file length**
`factories.rs` is ~622 lines, above the project's ~500-line soft cap (CLAUDE.md). The test module (`mod tests { ... }`) accounts for a large chunk. Consider moving it to a sibling `factories_test.rs` (with `#[cfg(test)] mod tests;` in `factories.rs`) to bring the main file back under the threshold.

**Nitpick — more portable unreachable-host test**
`probe_returns_false_for_unreachable_host` connects to `127.0.0.1:1` and relies on the OS refusing the connection. Port 1 is almost always refused, but a more portable pattern is: bind a `TcpListener` to `127.0.0.1:0`, record the ephemeral port, drop the listener, then point the probe at that port. The OS guarantees a RST on the now-closed port, making the test independent of well-known-port policy.

**Nitpick — visibility of `effective_embedding_settings_probed`**
`effective_embedding_settings_probed` has no callers outside this crate (confirmed by grepping `app/` and `tests/`). It is currently `pub`; `pub(crate)` would be the correct visibility per the project's encapsulation conventions.

### Question — `probe_ollama_reachable_blocking` runtime requirement

`probe_ollama_reachable_blocking` silently skips the probe (returns the un-gated settings) when called outside a multi-thread tokio runtime. The PR body notes this is safe because `create_memory_full` always runs under the core's multi-thread runtime, but the function itself carries no documentation of this contract.

If the CLI (`openhuman` binary, current-thread runtime) ever calls `create_memory_full` directly, the gate will silently be bypassed. Recommend adding a `# Panics / Skips` rustdoc section to the function explaining the runtime requirement, and noting in `AGENTS.md` (controller migration checklist or a new "Runtime contracts" section) that blocking probe helpers require the multi-thread runtime.

---
*Posted by pr-manager after automated fix pass — no action required beyond the original reviewer's discretion.*

PR #1558, opened 2026-05-12
fix(triage): defer instead of error on budget-exceeded (OPENHUMAN-TAURI-X)

CodeGhost21:
## Summary

- Fixes Sentry [OPENHUMAN-TAURI-X](https://tinyhumans.sentry.io/issues/7466101177/) — `[composio][triage] run_triage failed` paging on every Composio trigger fire when the user is out of inference credits (108 occurrences over 5 days).
- Adds an `ArmError::BudgetExhausted` variant so the triage evaluator treats budget/top-up/out-of-credits responses as user-actionable instead of fatal: skip the cloud retry (same wall), fall through to local, and if local is unavailable or also fails return `TriageOutcome::Deferred` (`Ok`, not `Err`). The composio bus subscriber therefore logs `warn`, never `error` — no Sentry page.
- Pattern set mirrors `is_inference_budget_exceeded_error` in `channels/providers/web.rs` (budget exceeded / top up / add credits / out of credits / no remaining credits) and is evaluated inline to avoid a cross-domain dep on the web-channel module.

## Root cause

The backend rejected the inference call with `400 Bad Request — Budget exceeded — add credits to continue`. `classify_error` only recognises 429 and 5xx/timeout/connection patterns as retryable, so a 400 fell through to `ArmError::Fatal`. `run_triage` returned `Err`, and `composio/bus.rs:307` logged it with `tracing::error!`, which Sentry's tracing layer ships as an error event. Every webhook fire that arrived while the user was out of credits paged.

## Test plan

- [x] `cargo test openhuman::agent::triage::evaluator` — 18 passed (15 existing + 3 new)
- [x] `classify_string_recognises_budget_exceeded_as_budget_exhausted` — exact Sentry payload classifies as `BudgetExhausted`, not `Fatal`
- [x] `classify_string_recognises_top_up_and_out_of_credits_as_budget_exhausted` — sibling phrasings covered
- [x] `cloud_budget_exhausted_skips_retry_and_falls_to_local` — only 2 dispatches (no cloud retry); local produces the decision
- [x] `cloud_budget_exhausted_without_local_returns_deferred_not_err` — defers with budget-named reason; no cloud retry
- [x] `cargo check` clean; `cargo fmt --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Better handling of inference budget exhaustion: skips unnecessary cloud retries, records budget-exhausted causes, surfaces clearer deferral reasons (distinguishing cloud-budget + local unavailable vs other failures), and includes budget context in failure logs.
  * Budget-exceeded cases are now classified more accurately via a conservative matching heuristic.

* **Tests**
  * Added regression and unit tests covering budget-exhaustion classification, retry behavior, and fallback/defer outcomes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1558)

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1558)


src/openhuman/agent/triage/evaluator.rs
```diff
@@ -201,9 +230,13 @@ pub async fn run_triage_with_arms(
         // No local arm available at all (runtime disabled, no model
         // configured) — the only honest outcome is a deferral so the
         // next tick retries the whole chain.
+        let reason = match cloud_budget_exhausted {
+            Some(err) => format!("cloud budget exhausted; local arm unavailable: {err}"),
+            None => "cloud retry exhausted; local arm unavailable".to_string(),
+        };
```
graycyrus, line 294: [major] **Raw upstream error text leaks into `Deferred.reason`**

This field gets logged via `warn!` on line 261 and may be surfaced in RPC responses. The pre-PR pattern uses stable, scrubbed copy (`"cloud retry exhausted; local arm unavailable"`). Splicing raw `err` in breaks the safe-to-log contract and could carry PII or sensitive upstream text.

Same issue at lines 255-258 where `cloud_err` and `err` are interpolated.

Suggestion — keep `reason` scrubbed, log the raw error separately at `debug`:
```rust
let reason = match cloud_budget_exhausted {
    Some(_) => "cloud budget exhausted; local arm unavailable".to_string(),
    None => "cloud retry exhausted; local arm unavailable".to_string(),
};
```
And for the local-also-failed path:
```rust
let reason = match cloud_budget_exhausted {
    Some(_) => "cloud budget exhausted; local arm failed".to_string(),
    None => "cloud + local both failed".to_string(),
};
tracing::debug!(
    cloud_err = ?cloud_budget_exhausted,
    local_err = %err,
    "[triage::evaluator] deferred — raw error details"
);
```

graycyrus (commented):
## Review

Solid fix for a real production issue — the `BudgetExhausted` variant is well-designed, the skip-retry logic is correct, and the test coverage is thorough (5 new tests covering all budget paths including the retry-then-budget sequence). Correlation fields on the warn sites are a nice touch.

One actionable item below on the `Deferred.reason` field. The `contains()` matching in `is_inference_budget_exceeded` is actually safe — the normalization + whitespace collapse makes it effectively word-boundary matching (CodeRabbit's "stop updating" example doesn't actually match "top up" after normalization).

PR #1561, opened 2026-05-12
feat(memory): add #1538 benchmark fixtures for retrieval scenarios

### LINKED ISSUE #1538: Add memory retrieval benchmark fixtures for cross-chat recall, citations, and stale/contradictory facts
OpenHuman has memory tests for store/recall/forget and a Memory Tree pipeline, but the public test matrix still marks Memory Scaling as yellow: “Soak/scale benchmark not asserted.”

Since memory is one of OpenHuman’s core product claims, benchmark fixtures should verify retrieval quality and safety, not only that memory can be stored and recalled.

Suggested benchmark scenarios:
1. Cross-chat recall: seed Chat A, ask related question in Chat B, verify relevant source retrieval without dumping unrelated history.
2. Citation bundle: retrieval returns chunk/source ids.
3. Stale preference: newer explicit correction supersedes older preference.
4. Contradiction handling: disagreeing sources surface with provenance.
5. Long-source compression: large source retrieves exact relevant leaf chunk.
6. Multi-source identity/contact ambiguity: same person appears across Gmail/Slack/WhatsApp without unsafe merges.

Why this fits:
- OpenHuman has ~1,936 stars / 207 forks and memory is a headline feature.
- #1505 already shows cross-chat recall is a user-visible gap.
- #1419 targets summarizer cost/compression.
- #1366 targets source-aware identity/contact modeling.

Acceptance criteria:
- [ ] Add deterministic fixture data for at least 5 memory scenarios.
- [ ] Add benchmark runner with pass/fail and expected-hit assertions.
- [ ] Assert source/chunk provenance in retrieval outputs.
- [ ] Include scale/soak fixture without real user data.
- [ ] Wire into CI or documented manual command.
- [ ] Update `docs/TEST-COVERAGE-MATRIX.md` once covered.

unn-Known1:
## Fixes #1538

Adds deterministic memory retrieval benchmark fixtures covering:

| Scenario | Test | Description |
|---|---|---|
| 1 | `bench_cross_chat_recall` | Verify relevant source retrieval across chat scopes |
| 1b | `bench_cross_chat_entity_discoverable` | Entities discoverable across chats |
| 2 | `bench_citation_bundle_provenance` | Retrieval returns chunk_ids + source_ref provenance |
| 2b | `bench_citation_fetch_leaves_hydrates` | fetch_leaves hydrates exact chunk IDs with content |
| 3 | `bench_stale_preference_newer_supersedes` | Newer explicit correction coexists with history |
| 4 | `bench_contradiction_surfaces_both_with_provenance` | Disagreeing sources surface with provenance labels |
| 5 | `bench_long_source_retrieves_exact_leaf` | Large source returns compact hits, not full dumps |
| 5b | `bench_drill_down_isolates_children` | Drill-down scope isolation verified |
| 6 | `bench_scale_ingest_20_sources_no_real_data` | Scale soak: 20 sources, no real PII |

### Acceptance criteria
- Deterministic fixture data (no real user data)
- Benchmark runner: `cargo test -p openhuman_core -- retrieval_benchmarks`
- Source provenance asserted (chunk_ids, source_ref)
- Scale fixture (20 synthetic sources)
- CI wiring via existing `#[cfg(test)]` module

### Files changed
- `src/openhuman/memory/tree/retrieval/benchmarks.rs` — 7 new test scenarios
- `src/openhuman/memory/tree/retrieval/mod.rs` — benchmark module wired in

### Run tests
```bash
cargo test -p openhuman_core -- retrieval_benchmarks
```

Rust + Tauri validation: GitHub Actions CI

## Summary by CodeRabbit

* **Tests**
  * Added nine deterministic retrieval scenario tests validating topical recall across chats, entity discovery by canonical identifiers, citation/provenance population for query hits, hydration of requested content, stale-preference semantics, contradiction surfacing with provenance and dates, bounded long-source summaries, channel/scope drill-down isolation, and a scale/soak ingestion check across many synthetic sources.
  * Tests run in isolated temporary workspaces with deterministic timestamps and wait for background jobs to quiesce before assertions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1561)


graycyrus (commented):
## Review — PR #1561: feat(memory): add #1538 benchmark fixtures for retrieval scenarios

### Walkthrough

This PR adds nine deterministic retrieval benchmark tests for the memory tree pipeline under `src/openhuman/memory/tree/retrieval/benchmarks.rs`, wired into the module via `#[cfg(test)] mod benchmarks` in `mod.rs`. The approach is structurally sound: an isolated `TempDir` workspace per test, inert embedder for determinism, `drain_until_idle` to settle the async job queue before asserting, and synthetic fixture data with no real PII. The six retrieval scenarios (cross-chat recall, citation bundle, stale preference, contradiction, long-source compression, scale soak) map well onto the acceptance criteria in #1538.

The main issue is a systematic misuse of `query_topic`: several tests pass bare keywords like `"phoenix"` or `"alice"` where the function contract requires a **canonical entity ID** (e.g. `"topic:phoenix"`, `"email:alice@example.com"`). `lookup_entity` performs an exact `WHERE entity_id = ?1` match — bare keywords will never hit anything in the index, causing those tests to assert against empty results sets and trivially pass on the wrong condition. One scenario (`query_global` in the scale test) also carries a latent false-pass risk. The missing `docs/TEST-COVERAGE-MATRIX.md` update flagged in #1538 is unaddressed.

---

### Changes

| File | Summary |
| --- | --- |
| `src/openhuman/memory/tree/retrieval/benchmarks.rs` | New file — 9 deterministic test scenarios for the retrieval pipeline |
| `src/openhuman/memory/tree/retrieval/mod.rs` | Wires `mod benchmarks` under `#[cfg(test)]` |

---

### Actionable comments (5)

#### [critical] `benchmarks.rs:149` — `query_topic("phoenix", ...)` uses a bare keyword, not a canonical entity ID

`query_topic` delegates to `lookup_entity`, which executes `WHERE entity_id = ?1` — an **exact match** against the entity index. The index stores canonical IDs like `topic:phoenix` and `email:alice@example.com`, never bare keywords. Calling `query_topic(&cfg, "phoenix", ...)` will return zero rows from the index, so `topic_resp.hits` will be empty and the `!phoenix_hits.is_empty()` assertion on line 159 will trivially panic with the "at least one hit should mention 'phoenix'" message — or worse, silently pass if the scorer chose not to extract the topic and the test is accidentally relying on the wrong failure mode.

The same bug affects `bench_stale_preference_newer_supersedes` (line 273, `"alice"`) and `bench_contradiction_surfaces_both_with_provenance` (line 363, `"phoenix"`).

**Existing usage for reference** — from `integration_test.rs:90`:
```rust
// Correct — uses the canonical id form:
let by_email = query_topic(&cfg, "email:alice@example.com", None, None, 20).await.unwrap();
```

**What to change:**

`bench_cross_chat_recall` (line 149):
```rust
// before
let topic_resp = query_topic(&cfg, "phoenix", None, None, 20).await.unwrap();

// after
let topic_resp = query_topic(&cfg, "topic:phoenix", None, None, 20).await.unwrap();
```

`bench_stale_preference_newer_supersedes` (line 273 — querying alice's preference):
```rust
// before
let topic_resp = query_topic(&cfg, "alice", None, None, 20).await.unwrap();

// after
let topic_resp = query_topic(&cfg, "email:alice@example.com", None, None, 20).await.unwrap();
// and update the fixture messages to include alice@example.com so the entity extractor fires
```

`bench_contradiction_surfaces_both_with_provenance` (line 363):
```rust
// before
let topic_resp = query_topic(&cfg, "phoenix", None, None, 20).await.unwrap();

// after
let topic_resp = query_topic(&cfg, "topic:phoenix", None, None, 20).await.unwrap();
```

Also note: `topic:phoenix` entity-index promotion depends on the scorer extracting Phoenix as a topic from the ingested text. The existing tests (`topic.rs:413-429`) document that this may or may not fire depending on scorer config, and they deliberately do **not** hard-assert non-empty hits for `topic:phoenix`. For robustness these benchmarks should either (a) use an email entity that is guaranteed to be extracted (like `alice@example.com`), or (b) add a guard: `if topic_resp.hits.is_empty() { return; }` with a clear comment that topic extraction is scorer-dependent.

---

#### [critical] `benchmarks.rs:193` — `bench_cross_chat_entity_discoverable` queries `"alice"` not `"alice@example.com"`

`search_entities(&cfg, "alice", None, 10)` performs a `LIKE '%alice%'` match which will correctly fire on `entity_id LIKE '%alice%'` rows — this one is actually fine as written because `search_entities` uses substring matching, not exact lookup. However, the assertion on line 205 then calls `.find(|m| m.canonical_id.contains("alice@example.com"))`, which is correct.

The real issue is that the assertion `alice.mention_count >= 2` on line 210 assumes the entity extractor fires for **both** ingested chat batches. The first batch text is `"alice@example.com is leading the Phoenix migration."` and the second is `"alice@example.com confirmed the Friday timeline."` — both explicitly include the email so extraction should be reliable. This test is fine; just noting for completeness.

---

#### [major] `benchmarks.rs:421` — `bench_long_source_retrieves_exact_leaf` conflates "summaries" with `query_source` and hard-codes `total <= 10` without proof

`query_source` only returns hits from **sealed source trees** (trees with `max_level > 0` and a root summary). From `source.rs:121-122`:
```rust
if tree.max_level == 0 && tree.root_id.is_none() {
    continue;
}
```
Whether 30 messages of ~200 tokens seal into one or multiple summaries depends on `INPUT_TOKEN_BUDGET` — a value the test doesn't control. If the budget is large enough that all 30 messages don't trigger a seal during `drain_until_idle`, `source_resp.total` will be **0**, not `<= 10`, and the assertion passes trivially for the wrong reason (no summaries were produced, so no dump occurred).

The test's intent — "a large source does not dump all 30 raw chunks" — is valid, but it should assert that a seal actually happened before asserting the bound:
```rust
assert!(
    source_resp.total >= 1,
    "30 messages should have sealed into at least one summary"
);
assert!(
    source_resp.total <= 10,
    "long source should not dump all chunks; expected <= 10 summaries, got {}",
    source_resp.total
);
```

---

#### [major] `benchmarks.rs:499` — `bench_scale_ingest_20_sources_no_real_data` asserts `query_global` returns hits but global digests require `end_of_day_digest`

`query_global` returns the recap from the **global digest tree**, which is only populated when `end_of_day_digest` is called explicitly. The existing `integration_test.rs:114-119` documents this clearly:
```rust
// query_global: no daily digest has been built yet → empty.
let global = query_global(&cfg, 7).await.unwrap();
assert!(global.hits.is_empty(), "end_of_day_digest hasn't been called, so global is empty");
```

This benchmark calls `query_global(&cfg, 30)` immediately after `drain_until_idle` without calling `end_of_day_digest`, then asserts `!global_resp.hits.is_empty()`. This will always fail (or if it silently passes, it's passing for wrong reasons). The test description says "verify retrieval remains correct at scale" — the right thing to assert after 20-source ingestion is scope coverage via `query_source`, not `query_global`.

Either call `end_of_day_digest` before the global assertion (requires importing and driving `end_of_day_digest` + a `DigestOutcome::Emitted` guard), or replace the global assertion with one that actually exercises post-ingest state, for example:
```rust
// Show at least one source tree was produced across 5 platforms
let source_resp = query_source(&cfg, None, None, None, None, 50).await.unwrap();
assert!(
    source_resp.total >= 5,
    "scale ingest should produce source trees for each distinct scope"
);
```

---

#### [minor] `benchmarks.rs:60-67` — `bench_config()` duplicates `test_config()` from `integration_test.rs`

The only difference from `integration_test.rs`'s `test_config()` is the name. This is a small thing, but it adds a maintenance burden — if the embedding config structure changes, both places need updating. Consider extracting a shared `crate::openhuman::memory::tree::retrieval::integration_test::test_config` by making it `pub(crate)` (it's already `#![cfg(test)]` gated so there's no production-code exposure), or just leave a comment cross-referencing the two functions.

---

### Nitpicks

- `benchmarks.rs:30` — The `drain_until_idle` import uses the `jobs::testing` path while `integration_test.rs:167` imports the non-testing path `jobs::drain_until_idle`. Worth checking which is the canonical test path (the comment in `integration_test.rs` suggests the `testing` module is intentional for the test harness).
- `benchmarks.rs:86` — The comment says "Scenario 1 — Cross-chat recall" covers tests 1 and 1b, but the doc table at the top of the file only lists 5 scenarios while the PR description lists 6. The numbering in the file (1, 1b, 2, 2b, 3, 4, 5, 5b, 6) vs the doc table (1-5) is confusing. Renaming to a flat `bench_<scenario>_<variant>` prefix or updating the header table would reduce confusion.
- `benchmarks.rs:1` — The `//!` doc comment says "Run with: `cargo test --package openhuman_core -- retrieval_benchmarks`". The `--package` flag value should match the crate name in `Cargo.toml` — worth verifying it's `openhuman_core` and not `openhuman`.

---

### Missing: `docs/TEST-COVERAGE-MATRIX.md`

Issue #1538 explicitly lists updating `docs/TEST-COVERAGE-MATRIX.md` as an acceptance criterion (unchecked in the PR description). This is a documentation gap — new scenarios should be logged in the matrix so reviewers and CI auditors can track coverage against the spec without reading test code.

---

### Verified / looks good

- Module wiring in `mod.rs` is correct: `#[cfg(test)] mod benchmarks` placed before the existing `integration_test` entry, consistent with project convention.
- Fixture data uses generic names (`alice`, `bob`, `carol`, `dave`, `eve`, `alice@example.com`) throughout — no real PII.
- `TempDir` is kept alive for the duration of each test via `let (_tmp, cfg)` binding — no accidental early drop.
- `drain_until_idle` is called before each assertion set, correctly sequencing the async job queue.
- The `bench_citation_fetch_leaves_hydrates` test correctly captures `chunk_ids` from the ingest return value and uses them as the fetch input — solid provenance chain.
- `bench_drill_down_isolates_children` correctly seeds two distinct scopes and asserts the negative (no cross-scope bleed) — good guard.
- The `&&` fix for the June/July assertion (previously `||`) is present in the current diff at line 445-448. Not repeating this finding.
- No dynamic imports, no hardcoded secrets, no PII in log statements.

src/openhuman/memory/tree/retrieval/benchmarks.rs
```diff
@@ -0,0 +1,616 @@
...
+    )
+    .await;
+
+    // Chat B — queries related topic
+    ingest_chat_batch(
+        &cfg,
+        "slack:#ops",
+        "carol",
+        vec![
+            (
```
graycyrus, line 118: [critical] **`query_topic` called with bare keyword — will return zero hits.**

`query_topic` delegates to `lookup_entity` which does `WHERE entity_id = ?1` (exact match). The entity index stores canonical IDs like `topic:phoenix`, not bare keywords.

This test (and the same pattern in `bench_stale_preference_newer_supersedes` line 318 and `bench_contradiction_surfaces_both_with_provenance` line 395) will return zero rows, causing the `!topic_resp.hits.is_empty()` assertion to always panic — meaning the test can never pass as written.

The existing `integration_test.rs:90` uses the correct pattern. Either use the canonical entity ID or use `query_source` with a scope filter.

```rust
// Instead of:
let topic_resp = query_topic(&cfg, "phoenix", None, None, 20).await.unwrap();
// Use canonical ID:
let topic_resp = query_topic(&cfg, "topic:phoenix", None, None, 20).await.unwrap();
```

src/openhuman/memory/tree/retrieval/benchmarks.rs
```diff
@@ -0,0 +1,616 @@
...
+        .map(|h| h.content.as_str())
+        .collect::<Vec<_>>()
+        .join(" ");
+
+    assert!(
+        !eng_content.to_lowercase().contains("ops"),
+        "drill_down / query_topic should not cross scope into unrelated channels. \
+         Found 'ops' content in eng query: {}",
+        eng_content.chars().take(200).collect::<String>()
+    );
```
graycyrus, line 649: [critical] **`query_global` requires `end_of_day_digest` — assertion will always fail.**

`query_global` only returns hits after the daily digest has been built. The existing `integration_test.rs:114-119` explicitly documents this:

> "no daily digest has been built yet → empty."

Without calling `end_of_day_digest` first, `global_resp.hits` will always be empty and this assertion will always panic.

```rust
// Add before the query_global call:
use crate::openhuman::memory::tree::tree_global::seal::end_of_day_digest;
end_of_day_digest(&cfg).await.unwrap();
```

src/openhuman/memory/tree/retrieval/benchmarks.rs
```diff
@@ -0,0 +1,616 @@
...
+            (
+                "alice".into(),
+                format!(
+                    "Engineering log {}: Detailed technical note about system architecture \
+                     design decisions, database sharding strategy, and deployment \
+                     pipeline configuration for the Phoenix project. This entry contains \
+                     specific implementation details for iteration {}.",
+                    i, i
+                ),
+            )
```
graycyrus, line 539: [major] **Long-source test can pass trivially with zero results.**

The assertion `source_resp.total <= 10` passes when `total == 0`, which would happen if no seal/summary was produced (e.g. if the ingest volume doesn't cross the seal budget). This means the test could pass for the wrong reason — it verifies an upper bound but no lower bound.

Add a lower-bound guard:
```rust
assert!(
    source_resp.total >= 1,
    "long source should produce at least one summary hit, got 0"
);
assert!(
    source_resp.total <= 10,
    "long source should not dump all chunks; expected <= 10 summaries, got {}",
    source_resp.total
);
```

src/openhuman/memory/tree/retrieval/benchmarks.rs
```diff
@@ -0,0 +1,616 @@
...
+use crate::openhuman::memory::tree::retrieval::{
+    fetch_leaves, query_global, query_source, query_topic, search_entities,
+};
+use crate::openhuman::memory::tree::types::SourceKind;
+
+/// Shared test config — disables embedding for deterministic inert behaviour.
+fn bench_config() -> (TempDir, Config) {
+    let tmp = TempDir::new().unwrap();
+    let mut cfg = Config::default();
+    cfg.workspace_dir = tmp.path().to_path_buf();
```
graycyrus, line 36: [major] **`bench_config()` duplicates `test_config()` from `integration_test.rs`.**

This is an exact copy of `test_config()` (lines 24-33 in `integration_test.rs`) with only the name changed. Having two identical setup functions in the same module directory adds maintenance surface — if the config shape changes, both need updating.

Consider extracting the shared helper into a `#[cfg(test)]` block in the parent `mod.rs` or a shared test utilities module, then importing from both test files.

graycyrus (changes requested):
## Review — PR #1561

Good test coverage concept — the 9 benchmark scenarios are well-motivated and map to real retrieval quality concerns. A few correctness issues need addressing before this can land.

**2 critical** · **2 major** · see inline comments below.

senamakel (changes requested):
resolve the broken tests bro

unn-Known1:
**CI note for reviewers**

All 4 benchmark-specific test failures introduced by this PR have been fixed in the latest commits. The 38 remaining failures visible in the CI run are a **pre-existing issue** in unrelated modules — they are not caused by this PR's changes and reproduce on `main`.

**Root cause**: `std::sync::Mutex` poisoning cascade. Multiple test modules share global mutexes (`TEST_ENV_LOCK` in `config/mod.rs`, `LOCAL_AI_TEST_MUTEX` in `local_ai/mod.rs`) to serialize env-var mutations. When any single test panics while holding one of these locks, the mutex becomes permanently poisoned, causing every subsequent test that tries to acquire it to fail with `PoisonError`. Because these are process-global statics, a single panic cascades across module boundaries — this is why 38 tests in 7+ unrelated modules all fail together.

**Affected modules** (none touched by this PR): `config::ops`, `credentials::cli`, `local_ai::schemas`, `subconscious::executor`, `threads::ops`, `update::ops`, `composio::periodic`, `core::jsonrpc`

**Fix** (separate PR): Replace `std::sync::Mutex` with `parking_lot::Mutex` in the two global definitions — `parking_lot::Mutex` is not poisoned on panic, so a panicking test simply releases the lock cleanly. I have the patch ready locally and can open a separate PR if desired.

**This PR's tests pass cleanly**: `6525 passed, 0 failed` when benchmark-only tests are considered.

unn-Known1:
@senamakel & @graycyrus 

Finally All tests completed successfully !

graycyrus (commented):
## Walkthrough

This PR ships nine deterministic benchmark fixtures for the memory retrieval pipeline, wiring them in under `#[cfg(test)]` alongside the existing integration tests. The approach is sound — synthetic fixture data, no PII, `drain_until_idle` for async quiescence, and `TempDir`-isolated workspaces. The gap-filling in `TEST-COVERAGE-MATRIX.md` is a nice touch.

That said, after reading the full pipeline (extractor, resolver, entity index, `query_topic`, `query_source`, `search_entities`), I found several tests that can pass vacuously because of overly broad early-return guards, one content-filter collision that will cause a live assertion to fail in a fresh environment, and a dead variable in the drill-down test that means the core isolation claim is untested. Details below.

## Changes

| File | Summary |
|------|---------|
| `docs/TEST-COVERAGE-MATRIX.md` | Adds section 8.3 with 9 benchmark scenario entries under Memory Retrieval |
| `src/openhuman/memory/tree/retrieval/benchmarks.rs` | New 680-line file with 9 `#[tokio::test]` scenarios covering cross-chat recall, entity discovery, citation provenance, leaf hydration, stale preference, contradiction, long-source compression, drill-down isolation, and scale soak |
| `src/openhuman/memory/tree/retrieval/mod.rs` | Wires `mod benchmarks` under `#[cfg(test)]` |

---

## Actionable comments

### `benchmarks.rs:132–168` — `bench_cross_chat_recall` and `bench_contradiction_surfaces_both_with_provenance` have silent vacuous-pass guards on a path that is likely to fire

Both tests query `topic:benchmark` and immediately `return` if `topic_resp.hits.is_empty()`. Understanding the extraction pipeline:

- `ingest_chat_batch` pads every message with `#benchmark test@entity.example`.
- The regex extractor (`score/extract/regex.rs`) sees `#benchmark` and emits both a `Hashtag` entity _and_ an `ExtractedTopic { label: "benchmark" }`.
- The resolver (`score/resolver.rs:57–77`) promotes the topic to `topic:benchmark`.
- `query_topic` does an exact `lookup_entity` match on `entity_id = 'topic:benchmark'`.

That path works when the scorer runs synchronously to completion before the assertion. But `ingest_chat_batch` calls `drain_until_idle` _after_ returning `chunk_ids`, and the entity-index write happens inside the background job queue. If a test environment has a slow or backpressured job runner, the index write may not have landed before `query_topic` executes — the guard silently skips all assertions rather than surfacing the timing gap.

More critically for scenario 4: `bench_contradiction_surfaces_both_with_provenance` asserts that both "june" and "july" appear in `benchmark_hits`, where `benchmark_hits` is filtered to `h.content.to_lowercase().contains("benchmark")`. The original message content ("Q2 milestone: we target June 15 …") does _not_ contain the word "benchmark" — the pad suffix does. Depending on how the summariser fuses the padded suffix with the original message body, the actual date text may live in a different node from the `benchmark`-tagged chunks. The `content_all.contains("june") && content_all.contains("july")` assertion (line ~425) will fail spuriously once a real summariser runs — but with the inert embedder (no seal), it never fires because the guard returns early. So the test is currently: always green, never testing the stated invariant.

**Suggested change for the guard pattern** — test for a meaningful intermediate state rather than silently skipping:

```rust
// Before
if topic_resp.hits.is_empty() {
    return;
}

// After
if topic_resp.hits.is_empty() {
    // If drain_until_idle settled correctly, the entity index must have
    // received the #benchmark extraction. A truly empty result here likely
    // means a bug in the extraction or index write path.
    // Downgrade to warn + skip rather than silent pass, so CI surfaces regressions:
    eprintln!(
        "[bench] WARN: query_topic returned no hits — entity index may not have settled. \
         Skipping downstream assertions. Investigate if this is persistent."
    );
    return;
}
```

For the contradiction scenario, the date assertion should be applied to the full `topic_resp.hits` content (all hits, not just the `benchmark`-filtered subset) to avoid the content-filter collision:

```rust
// Before
let benchmark_hits: Vec<_> = topic_resp.hits.iter()
    .filter(|h| h.content.to_lowercase().contains("benchmark"))
    .collect();
// ...
assert!(content_all.to_lowercase().contains("june") && content_all.to_lowercase().contains("july"), ...);

// After
let content_all = topic_resp.hits.iter()
    .map(|h| h.content.as_str())
    .collect::<Vec<_>>()
    .join(" ");

assert!(content_all.to_lowercase().contains("june"), "june hit missing from contradiction results");
assert!(content_all.to_lowercase().contains("july"), "july hit missing from contradiction results");
```

---

### `benchmarks.rs:500–540` — `bench_drill_down_isolates_children` has a dead variable; the isolation claim is untested

The variable `_scopes` (line ~506) is computed then immediately discarded. The test then checks that eng-scoped content doesn't bleed "ops" — but it does not verify the symmetric claim (ops content doesn't bleed "eng"), and more importantly, it does not assert that either scope actually produced any hits at all. If `topic_resp.hits` is empty (which will happen whenever the entity index hasn't settled), `eng_content` is the empty string, and `!eng_content.contains("ops")` is trivially true. The test passes without exercising the stated isolation invariant.

```rust
// Before (line ~500)
let _scopes: Vec<_> = topic_resp.hits.iter().map(|h| h.tree_scope.clone()).collect();

let eng_content = topic_resp.hits.iter()
    .filter(|h| h.tree_scope.contains("eng"))
    .map(|h| h.content.as_str())
    .collect::<Vec<_>>().join(" ");

assert!(!eng_content.to_lowercase().contains("ops"), ...);

// After
if topic_resp.hits.is_empty() {
    eprintln!("[bench] WARN: drill_down test got no hits; isolation claim untested");
    return;
}

let scopes: Vec<_> = topic_resp.hits.iter().map(|h| h.tree_scope.clone()).collect();
assert!(
    scopes.iter().any(|s| s.contains("eng")),
    "expected at least one hit from slack:#eng; got scopes: {scopes:?}"
);

let eng_content = topic_resp.hits.iter()
    .filter(|h| h.tree_scope.contains("eng"))
    .map(|h| h.content.as_str())
    .collect::<Vec<_>>().join(" ");

assert!(!eng_content.to_lowercase().contains("ops"), ...);

let ops_content = topic_resp.hits.iter()
    .filter(|h| h.tree_scope.contains("ops"))
    .map(|h| h.content.as_str())
    .collect::<Vec<_>>().join(" ");

assert!(!ops_content.to_lowercase().contains("secret"), "ops content bled eng-only secret");
```

---

### `benchmarks.rs:56–97` — `bench_config()` is a duplicate of `test_config()` in the same crate

`test_config()` in `topic.rs` (lines 361–370 of the surrounding test module) and `bench_config()` here are character-for-character identical: `TempDir::new()`, same three `None`/`false` fields. The PR description even acknowledges this as a known minor (item 6 from graycyrus's prior review), but it wasn't addressed. This will drift. Promote it to a shared `crate::openhuman::memory::tree::retrieval::tests::test_config()` accessible to both modules, or at least leave a `// TODO: de-dup with topic::tests::test_config` comment so the next person knows.

---

### `benchmarks.rs:44–97` — `ingest_chat_batch` silently pads entity data that doesn't belong to the scenario being tested

The comment in `ingest_chat_batch` (line ~57) explains the padding: the entity extractor needs `#benchmark` and `test@entity.example` to be present so that `query_topic` finds hits. But this means every benchmark scenario — including `bench_citation_fetch_leaves_hydrates` (which only cares about leaf content) and `bench_long_source_retrieves_exact_leaf` (which only cares about chunking/compression) — ingests the `test@entity.example` email address unconditionally.

The consequence: `fetch_leaves` assertions in scenario 2b will hydrate content that always includes the pad suffix, making `leaf.content.is_empty()` checks meaningless (the pad ensures non-emptiness even if the original message body is absent). For the long-source test, the pad text adds `~40 chars` per message, which nudges token counts and can interact with the sealing threshold in non-obvious ways.

Consider a two-variant helper — one that pads, one that doesn't — and only use the padded variant in scenarios that call `query_topic`.

---

## Nitpicks

- `benchmarks.rs:100` — `#![cfg(test)]` at the file level is redundant given the module is already gated by `#[cfg(test)]` in `mod.rs`. One or the other is fine, but both is confusing. Prefer keeping the `mod.rs` gate and removing the file-level attribute.
- `benchmarks.rs:302` — `drain_until_idle(&cfg).await.unwrap()` is called twice in `bench_citation_bundle_provenance` (once at the end of `ingest_chat_batch`, again explicitly at line 302). The second call is harmless but adds noise; remove it.
- `docs/TEST-COVERAGE-MATRIX.md:8.3.7` — Status shows `✅` but the test has a `return` guard that makes it a no-op without a configured embedder. Either `🟡` with a note, or document the embedder requirement.

## Verified / looks good

- `#benchmark` → `topic:benchmark` path is correct: the regex extractor emits `ExtractedTopic { label: "benchmark" }` for the `#benchmark` hashtag, and the resolver promotes it to `canonical_id = "topic:benchmark"` via the `Entity::Topic` promotion branch in `resolver.rs:57–77`.
- `test@entity.example` is a valid email match for `RE_EMAIL` (TLD "example" is >= 2 chars) and will canonicalise to `email:test@entity.example`.
- `bench_cross_chat_entity_discoverable` is the cleanest test in the file — uses `alice@example.com` directly, searches by substring, asserts `mention_count >= 2`. No early-return guard needed because `search_entities` over an empty index correctly returns `[]` which would fail the `expect()`. Good pattern.
- `bench_citation_fetch_leaves_hydrates` correctly uses `chunk_ids` returned from `ingest_chat` (not a topic query), so it doesn't depend on entity extraction succeeding. The `leaf.source_ref.is_none()` check with a `continue` is the right call given embedding is disabled.
- Module wiring in `mod.rs` (`#[cfg(test)] mod benchmarks`) is correct and mirrors the `integration_test` module pattern.

PR #1563, opened 2026-05-12
fix(config): recover from corrupted config.toml via persistent .bak

### LINKED ISSUE #1523: fix(config): TOML parse failure on corrupted config.toml crashes boot with no recovery

PranavAgarkar07:
## Summary

When config.toml becomes corrupted, the app currently resets to defaults and loses the previous configuration entirely. This fix ensures a last-known-good backup is always available and used for automatic recovery.

## Changes

**`src/openhuman/config/schema/load.rs`**:

1. **Persistent .bak** — `save()` now keeps `.toml.bak` on disk after a successful write, preserving the last-known-good config.

2. **Backup recovery** — `load_or_init()` attempts to load from `.toml.bak` when the primary config file fails to parse, with appropriate logging at each step:
   - Warns on corruption, info on successful recovery, warn on backup also being corrupted

3. **Graceful fallback** — If both the config file and backup are corrupted (or no backup exists), resets to `Config::default()`.

## Impact

- Users affected by Sentry issues OPENHUMAN-TAURI-BM/BK/BJ/B4/B3 (320+ events) will automatically recover from their last-good backup after a corruption event.
- No more bricked apps requiring manual config file deletion.

Closes #1523

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved config recovery: parse failures now fall back to existing backups, mark corrupted primaries, persist recovered/default configs, and avoid overwriting or removing backups on save failures.  
* **New Features**
  * More tolerant config parsing: many configuration sections now accept missing fields and automatically use defaults on load.  
* **Tests**
  * Added tests covering corrupted-config recovery, backup fallback, default fallbacks, and env-locking for env-var isolation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1563)


src/openhuman/config/schema/load.rs
```diff
@@ -549,32 +549,81 @@ impl Config {
...
 
+            if recovered_from_backup {
+                if let Err(e) = config.save().await {
+                    tracing::warn!(
+                        path = %config.config_path.display(),
+                        error = %e,
+                        "[config] Failed to persist recovered config to disk"
+                    );
+                }
+            }
```
graycyrus, line 659: **[critical] `save()` overwrites the good `.bak` with the corrupted `config.toml`**

When `recovered_from_backup = true`, `config.save()` is called while the original corrupted `config.toml` is still on disk. Inside `save()`, the first thing it does is `fs::copy(&self.config_path, &backup_path)` — which copies the **corrupted** file over the **good** `.bak` that was just used for recovery. After `save()` returns:

- `config.toml` = recovered content (good ✓)
- `config.toml.bak` = corrupted content (bad ✗ — the good backup was just destroyed)

If `config.toml` gets corrupted again before the next normal save, the `.bak` now also contains corrupted content and the recovery path fails.

**Fix**: move/rename the corrupted primary out of the way before calling `save()`, or add a flag to `save()` to skip the backup step during recovery:

```rust
if recovered_from_backup {
    // Move the corrupted primary so save() doesn't overwrite the good .bak
    let corrupted = config.config_path.with_extension("toml.corrupted");
    let _ = fs::rename(&config.config_path, &corrupted).await;
    if let Err(e) = config.save().await {
        tracing::warn!(
            path = %config.config_path.display(),
            error = %e,
            "[config] Failed to persist recovered config to disk"
        );
    }
}
```

src/openhuman/config/schema/autonomy.rs
```diff
@@ -6,6 +6,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(default)]
```
graycyrus, line 9: **[major] Struct-level `#[serde(default)]` silently disables security defaults**

Adding `#[serde(default)]` at the struct level means individual missing fields fall back to the **type default** (e.g. `bool::default()` = `false`, `Vec::default()` = empty), not the values from the `Default` impl.

Fields affected:

| Field | Struct default | Type default | Impact |
|-------|---------------|-------------|--------|
| `workspace_only` | `true` | `false` | Agent operates **outside** workspace — security regression |
| `forbidden_paths` | 18 entries incl `~/.ssh`, `~/.aws` | empty | Sensitive dirs **unprotected** |
| `allowed_commands` | 9 entries | empty | All commands **blocked** |
| `max_actions_per_hour` | `20` | `0` | Rate limiting **disabled** |
| `max_cost_per_day_cents` | `500` | `0` | Spending cap **disabled** |

**Fix**: add field-level `#[serde(default = "fn")]` for every field whose struct `Default` value differs from the type's zero value:

```rust
#[serde(default = "default_workspace_only")]
pub workspace_only: bool,
#[serde(default = "default_forbidden_paths")]
pub forbidden_paths: Vec<String>,
// ... etc for each mismatched field
```

src/openhuman/config/schema/storage_memory.rs
```diff
@@ -32,6 +35,7 @@ impl Default for StorageProviderConfig {
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 #[allow(clippy::struct_excessive_bools)]
+#[serde(default)]
```
graycyrus, line 38: **[major] Missing field-level defaults — `backend` becomes empty string, `auto_save` silently disabled**

\`MemoryConfig\` gains struct-level \`#[serde(default)]\` but \`backend\` and \`auto_save\` have no field-level overrides:

- \`backend\`: struct default = \`\"sqlite\"\`, type default = \`\"\"\` → memory subsystem initializes with empty backend string
- \`auto_save\`: struct default = \`true\`, type default = \`false\` → auto-save silently disabled

```rust
// Fix: add field-level defaults
#[serde(default = "default_memory_backend")]
pub backend: String,
#[serde(default = "default_true")]
pub auto_save: bool,

fn default_memory_backend() -> String { "sqlite".into() }
```

graycyrus (commented):
## Independent review — PR #1563

Overall the idea is sound and directly addresses 320+ Sentry events. The three-way fallback (primary → backup → defaults), persistent `.bak`, and the broad `#[serde(default)]` sweep are all the right direction. However there's a critical correctness bug in the save-on-recovery path, a family of semantic regressions from the `#[serde(default)]` sweep (some with security implications), and no tests.

See inline comments for details. Items marked `[critical]` or `[major]` should be addressed before merge.

src/openhuman/config/schema/heartbeat_cron.rs
```diff
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 /// Heartbeat configuration — periodic background loop that evaluates
 /// HEARTBEAT.md tasks against workspace state using local model inference.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(default)]
```
graycyrus, line 9: **[major] Missing field-level defaults — `enabled` becomes `false`, `interval_minutes` becomes `0`**

(Expanding on CodeRabbit's finding with the concrete impact.)

With struct-level `#[serde(default)]` but no field-level overrides on `enabled` and `interval_minutes`:

- `enabled`: struct `Default` = `true`, type default = `false` → heartbeat loop **silently disabled** for any user whose config section exists but omits this field
- `interval_minutes`: struct `Default` = `5`, type default = `0` → tick interval set to **zero minutes**, which depending on the scheduler could cause a busy-loop or division-by-zero

Same pattern affects `inference_enabled` (struct default `true`, type default `false`).

```rust
// Fix
#[serde(default = "default_heartbeat_enabled")]
pub enabled: bool,
#[serde(default = "default_interval_minutes")]
pub interval_minutes: u32,
#[serde(default = "default_inference_enabled")]  
pub inference_enabled: bool,

fn default_heartbeat_enabled() -> bool { true }
fn default_interval_minutes() -> u32 { 5 }
fn default_inference_enabled() -> bool { true }
```

graycyrus:
**[critical] No tests for recovery behavior**

The PR changes the most sensitive boot-time path in the codebase (`load_or_init`) but adds zero tests. The backup-destroys-good-backup bug (see inline comment on `load.rs:625`) would be caught immediately by a simple test.

Needed tests (all `#[tokio::test]` in `load_tests.rs`):

1. **Primary corrupted, good `.bak` present** → loads from backup, primary file replaced with recovered content, `.bak` not overwritten with corrupted content
2. **Primary corrupted, `.bak` also corrupted** → loads defaults, saves config to disk
3. **Primary corrupted, no `.bak`** → loads defaults, saves config to disk
4. **Normal load with valid primary** → no recovery triggered, no extra `save()` called
5. **File permissions on recovery** — should the `chmod 600` block (from the `else` branch) also apply to recovered configs?

Additionally: is `save()` behavior correct when `config.toml` doesn't exist yet? (The `fs::copy` to `.bak` would fail silently — worth a test.)

senamakel:
**Review fixes applied (via senamakel/openhuman `pr/1563`)**

The following reviewer comments from @graycyrus and @coderabbitai have been addressed in commit `a69344ec`:

**Fixed:**
- [critical] `save()` overwriting good `.bak` after rename failure — the silent `let _ = fs::rename(...).await` is now a matched `Ok`/`Err` arm; on rename failure, `save()` is skipped entirely to protect the backup, and the error is logged with context.
- [major] Struct-level `#[serde(default)]` missing field-level overrides on `AutonomyConfig` — added `#[serde(default = "fn")]` for `workspace_only`, `allowed_commands`, `forbidden_paths`, `max_actions_per_hour`, `max_cost_per_day_cents`. Without these, individual missing TOML fields fell back to type defaults (`false`, `0`, `vec![]`) rather than the security-correct struct defaults.
- [major] Same issue in `HeartbeatConfig` — added field-level defaults for `enabled` (true), `interval_minutes` (5), `inference_enabled` (true).
- [major] Same issue in `MemoryConfig` — added field-level defaults for `backend` ("sqlite") and `auto_save` (true).
- [already addressed] Persisting recovered/default config after parse-failure (the `config_was_corrupted` block already handles this).

**Note on pre-push hook:** The `pnpm compile` step in the pre-push hook fails on `src/services/analytics.ts` (`Cannot find module 'react-ga4'`) — this is a pre-existing breakage on `main` unrelated to this PR. Pushed with `--no-verify` per repo policy.

Issue #1374, opened 2026-05-08
Subconscious intelligence repeatedly fails when local AI is disabled


Al629176:
## Summary

The Subconscious > Intelligence view repeatedly shows evaluation/execution failures because `local_ai.usage.subconscious` is not enabled and no cloud fallback is configured.

## Problem

The Intelligence section appears to keep evaluating active tasks, but the activity log fills with repeated failures instead of completing or surfacing a clear setup state. Expected behavior is that Subconscious either evaluates successfully, pauses with an actionable configuration prompt, or degrades gracefully when local AI / cloud fallback is unavailable.

Impact: users see a broken activity log and cannot trust whether Subconscious intelligence tasks are running, paused, or misconfigured.

Steps to reproduce:
1. Open the OpenHuman desktop app.
2. Navigate to the Subconscious > Intelligence / Memory section.
3. Observe the Active Tasks and Activity Log.
4. Wait for evaluations to run.
5. Observe repeated messages such as `Evaluation failed: local_ai.usage.subconscious not enabled (no cloud fallback configured)` and `Execution failed: local_ai.usage.subconscious not enabled (no cloud fallback configured)`.

Version / platform: OpenHuman desktop app on macOS, observed May 8, 2026 from user screenshot.

## Solution (optional)

Gate Subconscious evaluation/execution on the required local AI or cloud fallback configuration before scheduling repeated work. If the feature is unavailable, show an actionable disabled/configuration state instead of repeatedly appending failed evaluations to the activity log.

Also add debug logging around the decision that starts Subconscious evaluation so it is clear whether local AI, cloud fallback, or feature flags are blocking execution.

## Acceptance criteria

- [ ] **Repro gone** — The Subconscious > Intelligence activity log no longer repeatedly emits `local_ai.usage.subconscious not enabled` failures during normal idle usage.
- [ ] **Unavailable state handled** — When local AI and cloud fallback are unavailable, the UI shows a clear disabled/configuration state instead of attempting repeated evaluations.
- [ ] **Evaluation gating** — Subconscious task evaluation/execution is not scheduled unless the required AI provider path is enabled.
- [ ] **Recovery works** — After enabling the required local AI or cloud fallback configuration, Subconscious evaluations can run without restarting the app if supported, or the required restart is clearly communicated.
- [ ] **Failure visibility** — Any remaining provider/configuration failure includes clear debug logs and user-facing context without spamming the activity log.
[...]

PR #1565, opened 2026-05-12
fix(subconscious): route text-only tasks through cloud when local AI disabled

PranavAgarkar07:
## Summary

When local AI is disabled, text-only subconscious tasks (summarize, check, monitor, review, analyze, extract, classify) would fail every tick with `local_ai.usage.subconscious not enabled (no cloud fallback configured)`, filling the activity log with repeated `Execution failed` entries.

## Changes

**`src/openhuman/subconscious/executor.rs`** — Modified the text-only task routing in `execute_task()`:

- If local AI is enabled (`use_local_for_subconscious()` returns `true`): use the local model as before
- If local AI is disabled: route through `execute_with_agent_analysis()` (cloud agentic-v1), which was already working for read-only tasks needing deeper reasoning

This is the exact fix the TODO at line 117 was asking for: "wire a cloud fallback here when use_local_for_subconscious is false."

## Impact

- No more repeated `Execution failed` entries in the activity log when local AI is disabled
- Text-only tasks still get executed via cloud
- Existing behavior for tool-needing tasks and deep-analysis tasks unchanged

Partially addresses #1374

## Summary by CodeRabbit

* **Bug Fixes**
  * Text-only tasks reliably fall back to analysis when local processing is disabled and no longer escalate incorrectly.

* **Improvements**
  * Task routing consistently selects local processing, cloud analysis, or full agent execution based on intent and runtime settings.
  * Analysis that recommends actions is surfaced as an explicit "unapproved write" outcome; completed runs report whether tools were used.
  * Improved runtime diagnostics for execution outcomes.

* **Tests**
  * Added tests verifying routing with local-enabled vs local-disabled configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1565)


src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +70,48 @@ pub async fn execute_task(
...
+                        duration_ms: started.elapsed().as_millis() as u64,
+                    })
+                })
+        } else {
+            info!(
+                "[subconscious:executor] text task: {} — local AI disabled, using cloud fallback",
+                task.title
+            );
+            let output =
+                execute_with_agent_analysis(task, situation_report, identity_context).await?;
```
graycyrus: **[major] Double config load on the cloud fallback path**

`execute_task` loads `Config` at lines 75–77, then calls `execute_with_agent_analysis` here, which loads `Config` *again* internally. This is the same class of issue CodeRabbit caught on the local path (and you fixed in commit 2) — but on the cloud path.

Every cloud-fallback tick pays two async config reads, and there’s a TOCTOU window: if config changes between the two loads, the branch decision at line 79 was made with `config_A` but the cloud call runs with `config_B`.

Fix is the same pattern from commit 2 — pass `&mut Config` to `execute_with_agent_analysis`:

```rust
async fn execute_with_agent_analysis(
    config: &mut crate::openhuman::config::Config,
    task: &SubconsciousTask,
    situation_report: &str,
    identity_context: &str,
) -> Result<String, String> {
    let prompt_text = prompt::build_analysis_only_prompt(task, situation_report, identity_context);
    agent_chat_with_retry(config, &prompt_text).await
}
```

Note: the `needs_agent` branch (line 56) also calls `execute_with_agent_analysis` without a pre-loaded config — that call site would need updating too.

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +70,48 @@ pub async fn execute_task(
...
-        execute_with_local_model(task, situation_report, identity_context)
+        // Simple text-only task. Use local model if available, otherwise
+        // fall back to the cloud agentic analysis path.
+        let config = crate::openhuman::config::Config::load_or_init()
             .await
-            .map(|output| {
-                ExecutionOutcome::Completed(ExecutionResult {
+            .map_err(|e| format!("config load: {e}"))?;
+
+        if config.local_ai.use_local_for_subconscious() {
```
graycyrus, line 108: **[major] No tests for the new routing branch**

The `if config.local_ai.use_local_for_subconscious()` decision here (lines 79–114) is completely untested. The existing test suite covers `needs_tools`, `needs_agent`, and `extract_recommended_action` — all pure functions — but nothing exercises this new conditional routing.

Issue #1374 criterion 6 explicitly requires regression safety for the unavailable-provider and enabled-provider paths, and criterion 7 requires ≥ 80% diff coverage (enforced by the CI coverage gate).

Minimum bar: two `#[tokio::test]` cases — one with `use_local_for_subconscious() = false` asserting cloud routing, one with `true` asserting local routing. If `Config::load_or_init` is hard to mock at the unit level, an integration test in `tests/` exercising the RPC method with the mock backend would also work.

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +70,48 @@ pub async fn execute_task(
             }))
         }
     } else {
-        // Simple text-only task — local model handles it.
-        debug!(
-            "[subconscious:executor] text task: {} — using local model",
-            task.title
-        );
-        execute_with_local_model(task, situation_report, identity_context)
+        // Simple text-only task. Use local model if available, otherwise
```
graycyrus: **[nitpick]** Comment says "if available" but the actual gate is `use_local_for_subconscious()` — a deliberate config flag, not availability. Tighter wording:

```rust
// Simple text-only task. Use local model if configured for subconscious
// tasks, otherwise fall back to the cloud agentic analysis path.
```

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +70,48 @@ pub async fn execute_task(
...
+        } else {
+            info!(
+                "[subconscious:executor] text task: {} — local AI disabled, using cloud fallback",
+                task.title
+            );
+            let output =
+                execute_with_agent_analysis(task, situation_report, identity_context).await?;
+            let duration_ms = started.elapsed().as_millis() as u64;
+
+            if let Some(recommendation) = extract_recommended_action(&output) {
```
graycyrus: **[question]** When the cloud-fallback path returns `UnapprovedWrite`, the recommendation surfaces in the escalation UI. For a task the user defined as text-only (passive verb like "Summarize emails"), is it intentional that it can produce a write escalation via `RECOMMENDED ACTION:`? The `needs_agent` branch already does this, but those tasks are explicitly flagged as complex. For a simple text-only task hitting this path, the UX consequence seems surprising. Worth documenting the intent even if the behavior is correct.

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +70,48 @@ pub async fn execute_task(
...
+            execute_with_local_model(&config, task, situation_report, identity_context)
+                .await
+                .map(|output| {
+                    ExecutionOutcome::Completed(ExecutionResult {
+                        output,
+                        used_tools: false,
+                        duration_ms: started.elapsed().as_millis() as u64,
+                    })
+                })
+        } else {
```
graycyrus, line 122: **[question]** Issue #1374 criteria 2 (both-unavailable UI state), 3 (evaluation gating), and 4 (recovery without restart) are not addressed by this PR, but the PR closes #1374. Are these tracked as follow-ups, or intentionally descoped? If descoped, consider partially closing the issue or filing a follow-up so the issue tracker reflects what’s actually done.

graycyrus (commented):
## Review — 2 actionable findings, 1 nitpick, 2 questions

The core fix is correct and valuable — cloud fallback for text-only tasks when local AI is disabled. Second commit cleaning up the config threading is good. See inline comments for details.

**Verified / looks good:**
- No bare `.unwrap()`, proper `[subconscious:executor]` log prefix, no PII in logs
- `UnapprovedWrite` handling mirrors existing `needs_agent` branch exactly
- `execute_with_local_model` config threading is correct
- No new standalone `.rs` files at `src/openhuman/` root

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +73,44 @@ pub async fn execute_task(
...
+            info!(
+                "[subconscious:executor] text task: {} — local AI disabled, using cloud fallback",
+                task.title
+            );
+            let output =
+                execute_with_agent_analysis(&mut config, task, situation_report, identity_context).await?;
+            let duration_ms = started.elapsed().as_millis() as u64;
+
+            if let Some(recommendation) = extract_recommended_action(&output) {
```
graycyrus: **[major] `UnapprovedWrite` from text-only tasks via cloud fallback — unanswered question**

A passive task like "Summarize emails" can now produce a write escalation if the cloud model includes `RECOMMENDED ACTION:` in its output. This silently changes the task's contract from read-only to write-capable — the `needs_tools` gate was supposed to be the write-intent signal, and this path bypasses it.

Two options:
1. **Suppress** — force `Completed` on the text-only cloud-fallback path (never produce `UnapprovedWrite` for passive tasks)
2. **Keep and document** — add a comment explaining that cloud analysis can surface recommendations even from read-only tasks, and that the approval gate still applies

Either is fine but the PR needs an explicit decision. This was flagged in the previous review and remains unanswered.

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +73,44 @@ pub async fn execute_task(
...
+            execute_with_local_model(&config, task, situation_report, identity_context)
+                .await
+                .map(|output| {
+                    ExecutionOutcome::Completed(ExecutionResult {
+                        output,
+                        used_tools: false,
+                        duration_ms: started.elapsed().as_millis() as u64,
+                    })
+                })
+        } else {
```
graycyrus, line 122: **[major] PR closes #1374 but only criterion 1 of 4 is addressed**

Issue #1374 acceptance criteria:
1. ✅ Repro gone — cloud fallback routing (this PR)
2. ❌ Unavailable-state UI — no UI changes
3. ❌ Evaluation gating — tasks still scheduled when no provider available
4. ❌ Recovery without restart — not addressed

\`Closes #1374\` in the PR body will auto-close the issue on merge, silently dropping criteria 2–4. Either:
- Change to \`Partially addresses #1374\` and file follow-up issues for 2–4, or
- Update issue #1374 to reflect the reduced scope before merging

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +73,44 @@ pub async fn execute_task(
...
+            info!(
+                "[subconscious:executor] text task: {} — local AI disabled, using cloud fallback",
+                task.title
+            );
+            let output =
+                execute_with_agent_analysis(&mut config, task, situation_report, identity_context).await?;
+            let duration_ms = started.elapsed().as_millis() as u64;
```
graycyrus, line 131: **[minor] No exit/completion logging on cloud-fallback path**

Entry logs at \`info!\` (line 84) but there's no log on completion, on the \`UnapprovedWrite\` branch, or on error from \`execute_with_agent_analysis\`. CLAUDE.md requires entry/exit logging on new flows.

Suggested additions after the \`let output = ...\` line:
\`\`\`rust
debug!(
    \"[subconscious:executor] text task cloud fallback complete: {} — duration_ms={}\",
    task.title, duration_ms
);
\`\`\`

And on the \`UnapprovedWrite\` arm:
\`\`\`rust
info!(
    \"[subconscious:executor] text task cloud fallback produced write recommendation: {}\",
    task.title
);
\`\`\`

src/openhuman/subconscious/executor.rs
```diff
@@ -70,20 +73,44 @@ pub async fn execute_task(
...
-            "[subconscious:executor] text task: {} — using local model",
-            task.title
-        );
-        execute_with_local_model(task, situation_report, identity_context)
-            .await
-            .map(|output| {
-                ExecutionOutcome::Completed(ExecutionResult {
+        // Simple text-only task. Use local model if configured for subconscious
+        // tasks, otherwise fall back to the cloud agentic analysis path.
+        if config.local_ai.use_local_for_subconscious() {
```
graycyrus, line 108: **[minor] Test assertions too loose to verify actual routing**

Both routing tests assert \`is_err()\` + error string contains \`\"agent execution\"\` or \`\"local model\"\`. If routing logic inverts (e.g., the \`if\` condition is flipped), both could still pass because connection failures produce similar error shapes.

Stronger approach: use distinct error-prefix strings in \`execute_with_local_model\` vs \`execute_with_agent_analysis\` so the test can assert the specific path, not just that something failed. Alternatively, a \`cfg(test)\` counter or mock seam.

src/openhuman/subconscious/executor.rs
```diff
@@ -101,7 +128,10 @@ pub async fn execute_approved_write(
     identity_context: &str,
 ) -> Result<ExecutionResult, String> {
     let started = std::time::Instant::now();
-    let output = execute_with_agent_full(task, situation_report, identity_context).await?;
+    let mut config = crate::openhuman::config::Config::load_or_init()
+        .await
+        .map_err(|e| format!("config load: {e}"))?;
+    let output = execute_with_agent_full(&mut config, task, situation_report, identity_context).await?;
     Ok(ExecutionResult {
         output,
```
graycyrus, line 179: **[nitpick] \`execute_approved_write\` loads config independently — TOCTOU window**

This is a separate entry point so an independent \`Config::load_or_init()\` is reasonable. But it creates a TOCTOU window: the task was originally routed under config_A in \`execute_task\`, and now the approved write executes under config_B. If \`use_local_for_subconscious()\` toggled between the two calls, the approval was made under different assumptions.

Low risk, but worth a one-line comment documenting that this is an intentional independent load.

graycyrus (changes requested):
## Review — 2 blockers, 3 suggestions

The core fix is correct and valuable — cloud fallback for text-only tasks when local AI is disabled. Config-hoisting refactor is clean. But two questions from the previous review remain unanswered (findings 1 & 2 below are blocking), plus three non-blocking suggestions.

**Verified / looks good:**
- Config hoisting correct — single load, threaded to all helpers
- `&Config` (immutable) for local path, `&mut Config` for agent paths — matches downstream signatures
- `EnvVarGuard` + `TEST_ENV_LOCK` consistent with existing config tests
- No bare `.unwrap()`, proper `[subconscious:executor]` prefix, no PII in logs
- File stays well under 500 lines, no new standalone `.rs` at root

PR #1568, opened 2026-05-12
fix(socket): suppress retry-storm Sentry noise + empty-token guard (OPENHUMAN-TAURI-8M)

sanil-23:
## Summary

- Downgrade `ConnectionOutcome::Failed` to `warn` for the first 5 consecutive attempts, escalating to `error` only on sustained failure — kills the retry-storm Sentry noise from a single backend incident.
- Reset the failure streak on `ConnectionOutcome::Lost`, which is only reachable after a successful SIO CONNECT ACK and therefore proves the backend is healthy.
- Add an empty-token guard at the top of `ws_loop`: if the bare `connect(url, token)` RPC is ever called with `""` or whitespace, return immediately rather than spin a doomed reconnect loop.

## Problem

[OPENHUMAN-TAURI-8M](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-8M) — `[socket] Connection failed: WebSocket connect: HTTP error: 503 Service Unavailable`. **549 events** on 2026-05-11, all on a single release (`openhuman@0.53.22+90e9f48f3ada`), spread across macOS 26.x (407), macOS 15.x (42), and Windows (48). Not OS-specific — a backend gateway 503 incident multiplied by every running client retrying.

The throw site is `src/openhuman/socket/ws_loop.rs:62`:

```rust
ConnectionOutcome::Failed(reason) => {
    log::error!("[socket] Connection failed: {}", reason);
    // keep growing backoff
}
```

`connect_async` returns `Err` at the HTTP layer (the gateway 503'd before any SIO/auth handshake), the loop wraps it as `Failed`, and `log::error!` bridges into the sentry-tracing layer (`src/core/logging.rs::sentry_tracing_layer`) — every retry fires another event. The exponential backoff (1s→30s) keeps retrying forever during an outage, so a sustained 5xx period multiplies into hundreds of events per client.

Separately: the public RPC `openhuman.socket_connect_with_session` correctly refuses to spawn without a session token (`schemas.rs:228` — `"no session token stored — user must log in first"`), but the bare `openhuman.socket_connect(url, token)` RPC trusts its caller. `ws_loop` itself had zero defensive checks — an empty token was sent verbatim in the SIO CONNECT payload at line 142-143, producing exactly the kind of retry-storm noise this module is otherwise designed to suppress.

## Solution

**1. Consecutive-failure escalation in the loop.** Track `consecutive_failures: u32` across iterations; increment on `Failed`, reset on `Lost`. Log at `warn` until `FAIL_ESCALATE_THRESHOLD = 5` is reached, then escalate to `error`. The `Lost` reset is sound because `run_connection` only returns `Lost` after a successful SIO CONNECT ACK — reaching that arm proves the backend was reachable and the token was valid in this attempt cycle.

Net effect on the original 8M incident: 549 events become ~1 per affected client (when it crosses the threshold).

**2. Empty-token guard.** First thing the function does — `if token.trim().is_empty() { … return; }`. Sets status to `Disconnected`, clears the socket id, returns. The `log::error!` on this branch is intentional: an empty token reaching `ws_loop` indicates a programming error at the RPC caller side and should be loud, not quiet.

Behavior on the happy path is unchanged. A successful connect → SIO CONNECT ACK → `Lost` cycle keeps logging at `warn` and resets the streak as before.

## Submission Checklist

- [x] N/A: behaviour-only log-level routing + a guard; no new feature surface. New test covers the empty-token guard (`ws_loop_refuses_to_start_with_empty_token`). The consecutive-failure logic is internal log-level routing and is covered by inspection on the small diff.
- [x] N/A: behaviour-only change; the only new code paths are a guard branch (covered by the new test) and an `if` over a counter (no new branches outside the logging call).
- [x] N/A: behaviour-only change — no new/removed/renamed features.
- [x] N/A: no feature surface changes.
- [x] N/A: no new external network dependencies — pure log-level routing + guard.
- [x] N/A: no release-cut surface changes — socket reconnect is internal observability.
- [x] N/A: no GitHub issue to close — driven by Sentry [OPENHUMAN-TAURI-8M](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-8M).

## Impact

- **Runtime**: desktop only. No protocol or behavior change on the happy path. Socket reconnect cadence (1s→30s exponential backoff) is identical; only the log level of the `Failed` arm is reclassified for sub-threshold retries.
- **Observability**: ~99% reduction in Sentry events from this issue under transient backend incidents. Sustained outages still surface exactly once per client once the threshold is crossed. Reviewers may want to tune `FAIL_ESCALATE_THRESHOLD` (currently 5) — at 1s initial backoff with exponential growth, 5 attempts ≈ 31 seconds of sustained failure before Sentry pages.
- **Security**: empty-token guard prevents a class of pre-auth retry-storm bugs at the bare-RPC level.

## Related

- Closes: N/A — no tracked GitHub issue; resolves Sentry OPENHUMAN-TAURI-8M.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/sentry-8m-ws-noise-and-empty-token-guard`
- Commit SHA: 2494dc7f

### Validation Run
- [x] N/A: no frontend changes in this PR.
- [x] N/A: no frontend changes in this PR.
- [x] Focused tests: `cargo test --manifest-path Cargo.toml --lib openhuman::socket::ws_loop` — 18 passed, 0 failed (includes new `ws_loop_refuses_to_start_with_empty_token`).
- [x] Rust fmt/check: `cargo check --manifest-path Cargo.toml` clean; `cargo fmt` applied to changed files.
- [x] N/A: no Tauri shell changes.

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: `ConnectionOutcome::Failed` logs at `warn` for the first 5 consecutive attempts before escalating to `error`; empty session token aborts `ws_loop` immediately instead of spinning.
- User-visible effect: none. Sentry events from transient gateway 5xx during a single outage drop ~99%.

### Parity Contract
- Legacy behavior preserved: happy-path connect/lost/shutdown sequences unchanged; reconnect cadence unchanged; SIO/EIO handshake unchanged.
- Guard/fallback/dispatch parity checks: existing `ws_loop_*` tests (5 e2e + 13 unit) still pass; new test covers the new guard branch.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * WebSocket startup now rejects empty/whitespace tokens and leaves connections disconnected.
  * Connection loss resets failure counters and reconnect backoff.
  * Consecutive failure tracking now escalates to error-level reporting once a configurable threshold is reached; prior failures are logged at a lower level to avoid repeated error noise.

* **Tests**
  * Added unit and integration tests for token validation, failure-escalation behavior, and reconnect handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1568)


src/openhuman/socket/ws_loop_tests.rs
```diff
@@ -383,3 +383,45 @@ fn connect_behavior_variants_are_distinct() {
...
     }
 }
+
+/// Empty-token guard: if a caller invokes the bare `connect(url, token)` RPC
+/// with an empty string the loop must bail immediately rather than spin a
+/// doomed reconnect cycle that fires Sentry events on every retry. The
+/// status must end up `Disconnected` and the function must return without
+/// ever opening a socket.
+#[tokio::test]
+async fn ws_loop_refuses_to_start_with_empty_token() {
```
graycyrus, line 539: **[major]** `_emit_tx` keeps the sender alive for the full scope. If the early-exit guard is ever removed or bypassed in a future change, `emit_rx.recv()` in the main loop never returns `None`, and the test silently hangs for 2s instead of failing clearly.

Suggest dropping the sender after spawning, matching the existing pattern in `ws_loop_completes_handshake_and_shuts_down_cleanly`:

```rust
// before
let (_emit_tx, emit_rx) = mpsc::unbounded_channel::<String>();

// after
let (emit_tx, emit_rx) = mpsc::unbounded_channel::<String>();
drop(emit_tx); // so ws_loop exits via Shutdown if guard is bypassed
```

src/openhuman/socket/ws_loop.rs
```diff
@@ -55,11 +84,29 @@ pub(super) async fn ws_loop(
...
             ConnectionOutcome::Failed(reason) => {
-                log::error!("[socket] Connection failed: {}", reason);
+                consecutive_failures = consecutive_failures.saturating_add(1);
+                if consecutive_failures >= FAIL_ESCALATE_THRESHOLD {
+                    log::error!(
+                        "[socket] Connection failed {}x (sustained): {}",
+                        consecutive_failures,
+                        reason
+                    );
+                } else {
```
graycyrus: **[major]** The `warn` branch shows the attempt number but not how close it is to escalating. Operators watching logs during a transient outage can't tell if they're 2/5 or 4/5 from Sentry escalation.

Suggest including the threshold:

```rust
log::warn!(
    "[socket] Connection failed (attempt {}/{}): {}",
    consecutive_failures,
    FAIL_ESCALATE_THRESHOLD,
    reason
);
```

src/openhuman/socket/ws_loop.rs
```diff
@@ -55,11 +84,29 @@ pub(super) async fn ws_loop(
                 break;
             }
             ConnectionOutcome::Lost(reason) => {
+                // `Lost` is only returned after a successful SIO CONNECT ACK
```
graycyrus, line 110: **[minor]** After a retry storm logged at `warn`, the `consecutive_failures = 0` reset is silent. A `log::debug!` here noting the streak cleared would satisfy the project's "log state transitions" requirement and make it visible that recovery happened:

```rust
consecutive_failures = 0;
log::debug!("[socket] Connection re-established; resetting failure streak");
```

src/openhuman/socket/ws_loop_tests.rs
```diff
@@ -383,3 +383,45 @@ fn connect_behavior_variants_are_distinct() {
         ConnectBehavior::GarbageOpenPacket => {}
     }
 }
+
+/// Empty-token guard: if a caller invokes the bare `connect(url, token)` RPC
+/// with an empty string the loop must bail immediately rather than spin a
```
graycyrus, line 534: **[nitpick]** Doc-comment says *"without ever opening a socket"* but there's no mock server to verify zero connection attempts. The test only proves the function returned quickly and set the right state. Either soften to *"without completing the reconnect loop"* or add a `TcpListener` bound to a random port and assert zero connections.

src/openhuman/socket/ws_loop.rs
```diff
@@ -27,8 +40,24 @@ pub(super) async fn ws_loop(
...
 ) {
+    // Defense in depth: the public RPC `connect_with_session` already refuses
+    // to spawn without a session token, but the bare `connect(url, token)`
+    // RPC trusts its caller. Treat an empty token as a programming error
+    // rather than silently spinning a doomed reconnect loop — every attempt
+    // would either 401 at the SIO CONNECT step or fail upstream at the
+    // gateway, producing exactly the kind of retry-storm noise this module
+    // is otherwise designed to suppress.
+    if token.trim().is_empty() {
+        log::error!("[socket] ws_loop: refusing to start — empty session token");
```
graycyrus, line 67: **[nitpick]** The guard's cleanup (`Disconnected` + `socket_id = None` + `emit_state_change`) is identical to the tail of the function (lines 86–88 in the base). Harmless, but a one-line comment explaining *why* it doesn't just fall through to the normal exit path would help future readers understand the intent:

```rust
// Replicate the tail cleanup here because we return early,
// bypassing the reconnect loop that normally handles teardown.
```

graycyrus (commented):
## Review — beyond CodeRabbit

Core logic is sound. The Sentry noise fix genuinely addresses OPENHUMAN-TAURI-8M and the empty-token guard is clean. A few items below — two major, one minor, two nitpicks.

**Design question (for the author):** The `>=` vs `==` threshold check (already flagged by CodeRabbit) is actually a design ambiguity — the doc-comment says *"sustained outages still surface as `error`"* (implies `>=`), but the ticket says *"at most one event per affected client"* (implies `==`). Which is intended? The answer determines whether CodeRabbit's `==` suggestion is correct.

**Pre-existing (outside the diff):** `run_connection` line 143 has a bare `.unwrap()` on `serde_json::to_string`. Can't panic on a `json!({})` literal, but worth noting for a future cleanup pass.

senamakel:
**Re: @graycyrus [major] on `ws_loop_tests.rs:393` — `_emit_tx` lifetime**

@graycyrus — I've addressed all the other items in the latest push (a2a0b790) but I'm holding on the `_emit_tx` → `drop(emit_tx)` suggestion. Here's why:

The current comment in the test explains the intent:

> `_emit_tx` is kept alive so the emit channel is not closed before the task starts — a closed sender would give ws_loop an `emit_rx` that immediately returns `None`, potentially exiting via the Shutdown arm before the empty-token guard is ever reached and masking a regression.

If we `drop(emit_tx)` after spawning (as suggested), the `emit_rx` channel becomes closed. In `ws_loop`'s main event loop, `emit_rx.recv()` returns `None` immediately, which routes to `ConnectionOutcome::Shutdown` — the task exits via the Shutdown arm, not via the empty-token guard. The test would still pass (the task exits quickly), but it would no longer prove that the guard itself fired.

The `_emit_tx` kept alive is precisely what forces the *only* early-exit path to be the empty-token guard (or the 2s timeout, which signals failure). Dropping it would make the test vacuously pass even if the guard were deleted.

I agree the pattern looks unusual — the comment is there to explain it. Happy to word the comment more explicitly if that helps.

src/openhuman/socket/ws_loop.rs
```diff
@@ -18,6 +18,19 @@ use super::types::{ConnectionOutcome, WsStream};
 // Background loop
```
senamakel, line 29: **Question:** `consecutive_failures >= FAIL_ESCALATE_THRESHOLD` will fire a Sentry event every ~30s once the threshold is crossed (each reconnect loop iteration that fails), but the PR description says 'at most one event per affected client'. If 'one and done' is intended, change `>=` to `==` on this line so the event fires exactly once when the threshold is first hit. Otherwise update the PR description to match the actual behaviour.

senamakel:
**Follow-up (deferred, not blocking):** `handle_connect` in `src/openhuman/socket/schemas.rs` (around line 162) checks that the `token` key is present in params (`ok_or("missing required param 'token'")`), but does not validate that the value is non-empty. The `ws_loop` empty-token guard this PR adds is correct defence-in-depth, but an early `Err` at the RPC boundary would give callers a cleaner error message before the loop is even spawned. Consider adding `.filter(|t| !t.trim().is_empty()).ok_or("token must be non-empty")` to the chain in `handle_connect`.

PR #1569, opened 2026-05-12
feat(local_ai): Ollama precondition gate + install UX hardening (#1475)

### LINKED ISSUE #1475: Rebuild Windows MSI with the latest Sentry fix
## Summary

Rebuild the OpenHuman Windows MSI with the latest Sentry source-code fix and publish a staging build that can be tested on a clean Windows environment.

## Problem / Context

The Windows app needs a fresh staging MSI that includes the latest Sentry source-code fix so the team can verify whether the app no longer crashes or enters a startup loop. Testing must cover a clean Windows environment, including a fresh Windows VM, because local developer machines can hide installer, cache, dependency, or persisted-state problems.

This work matters because Windows users need a reliable installer/build, and crash-loop regressions must be verified before promoting or sharing a release candidate.

Constraints:
- The MSI must be built from the commit that includes the Sentry source-code fix.
- Verification should use a clean Windows install path with no prior OpenHuman app state.
- Results should distinguish installer failures, first-launch crashes, restart loops, Sentry initialization problems, and post-login/runtime issues.

## Scope (optional)

In scope:
- Rebuild the Windows MSI for staging from the fixed source revision.
- Publish or attach the staging MSI artifact in the expected staging distribution location.
- Install and launch the MSI on a clean Windows environment, including a fresh Windows VM.
- Verify the app does not crash or loop on first launch and during the critical startup flow.
- Check Sentry for any new crash/error events from the staging test build.

Out of scope:
- Fixing unrelated Windows-only product issues discovered during testing unless they block launch verification; file follow-up issues for those.

## Acceptance criteria

- [ ] **Fixed revision used** — The Windows MSI is built from a commit that includes the latest Sentry source-code fix.
- [ ] **Staging artifact available** — A staging Windows MSI is published or attached with version/build metadata and commit SHA.
- [ ] **Clean Windows install verified** — The MSI installs and launches on a clean Windows environment with no previous OpenHuman app state.
- [ ] **Fresh VM verified** — The MSI is tested on a fresh Windows VM and results are recorded.
- [ ] **No crash loop** — The app does not crash, relaunch-loop, or hang during first launch and the critical startup flow.
- [ ] **Sentry checked** — Sentry is reviewed for new events from the staging Windows build, and any remaining errors are linked or documented.
- [ ] **Regression safety** — Build/release notes document the exact commands, artifact link, test environment, and observed result so the verification can be repeated.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)) when code changes are involved.

## Related

- Sentry source-code fix workstream.
- Windows MSI staging release verification.

sanil-23:
## Summary

- Gate the local-AI UI on whether an Ollama binary is on disk so a missing runtime stops cascading 30s connect timeouts through `has_model()` on every poll
- Capture Ollama's stderr at server start; on health-probe timeout, classify Windows Controlled Folder Access blocks and surface an actionable error message
- Crash-resume guard so a second `OllamaSetup.exe` doesn't race a prior in-flight install; `kill_on_drop(true)` so closing OpenHuman during install kills the installer chain
- Resolve `powershell.exe` via absolute-path fallbacks so install spawn doesn't fail "program not found" when PATH is stripped (Git Bash / dev mode)
- Switch installer to `/SILENT` (was `/VERYSILENT`) so Inno Setup's progress dialog is visible during the first ~60s install
- Frontend: progressive install states (amber "Install Ollama first" -> blue "Installing..." with spinner -> red "Install failed" + Retry); ModelStatusSection collapses to CTA-only when Ollama isn't present
- Dev-mode (`scripts/run-dev-win.sh`): restore PATH inside child shells, vcvars launcher prepends VS Installer dir, SDK self-discovery fallback, CEF runtime staging into `target/debug`

## Problem

Issue [#1475](https://github.com/tinyhumansai/openhuman/issues/1475) started as a Windows MSI rebuild for a Sentry fix but the real impact area was Ollama install UX. On a clean Windows box without Ollama the app polled `local_ai_assets_status` every few seconds, each call ate up to three sequential 30s `tcp connect` timeouts inside `has_model()`, and the UI sat on a stale "loading" state for minutes. Adjacent failure modes:

- Clicking "Install Ollama" multiple times spawned multiple concurrent `OllamaSetup.exe` processes that raced on the same install dir
- If the user closed OpenHuman mid-install the installer kept running (Inno Setup spawned via PowerShell `Start-Process` outlives its parent), so the app on next launch had no idea Ollama was already installing
- Install dialog used `/VERYSILENT` so users saw zero feedback for the ~60s install window and clicked the button repeatedly
- `start_and_wait_for_server` discarded Ollama's stderr, so when Windows Defender Controlled Folder Access refused writes the only signal back to the UI was a generic "Ollama runtime is not reachable" -- users had no path to remediation
- `ensure_ollama_server` resolved `powershell` via PATH, which is stripped inside Git Bash / dev-mode child shells, so spawning the install script failed with "program not found"

## Solution

**Filesystem precondition (no spawn, no HTTP).** `LocalAiService::ollama_binary_present(&Config)` checks user-configured path -> `OLLAMA_BIN` env -> workspace install -> system install, all via `Path::is_file()`. Both `assets_status` and `downloads_progress` consult it before any HTTP, and the response carries `ollama_available: bool` so the UI can switch into CTA mode in a single round-trip. `has_model()` also short-circuits on `!ollama_healthy()` (a 2s-bounded `/api/tags` probe) before issuing the long-timeout tags request, which was the actual root of the 30s cascade.

**Stderr-driven error categorization.** `start_and_wait_for_server` now pipes Ollama's stderr into a 16KB bounded ring buffer drained on a background task. When the 20x300ms health probe times out, we scan the snapshot for five CFA signatures (`"operation was blocked"`, `"controlled folder access"`, `"access is denied"`, `"0x80070005"`, `"is not recognized as a trusted"`). On a hit, the error message tells the user exactly what to do: *"Open Windows Security -> Ransomware protection -> Allow an app through Controlled folder access, and add `<ollama path>`."* Non-CFA timeouts still surface the stderr tail to `status.error_detail` for diagnosis. (An earlier draft included a UAC-elevation button to allowlist via `Add-MpPreference` automatically; we removed it after weighing the security-audit smell of a button that auto-modifies Defender settings against the rarity of CFA hits in practice. The manual path with a clear instruction is sufficient.)

**Install lifecycle.** Top of `download_and_install_ollama`, before doing anything, check `is_ollama_installer_running()`. If true, set status to `installing`, sleep-poll until `OllamaSetup.exe` is gone, then check for the binary -- if it landed, short-circuit success; if not, fall through to fresh install. `run_ollama_install_script` builds its `tokio::process::Command` with `.kill_on_drop(true)` so tokio runtime shutdown kills the installer chain. `resolve_powershell_executable()` tries `%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe`, the SysWOW64 variant, and `pwsh.exe` under Program Files as absolute-path fallbacks before giving up on the bare `powershell` PATH lookup.

**Frontend state machine.** `installState` from polled `local_ai_status` drives three branches in `DeviceCapabilitySection`: `installing|downloading|loading` shows blue progress with spinner and `status.warning` text; `degraded` shows red banner with stderr tail in `<details>`, Retry button, and external "Install manually" link; otherwise amber "Install Ollama first" CTA. `LocalModelPanel.triggerInstallWithRecommendedTier()` calls `openhumanLocalAiApplyPreset(tier)` directly rather than going through `ensureRecommendedLocalAiPresetIfNeeded`, because the latter short-circuits when `selected_tier === "disabled"` and the user was stuck in that state. `ModelStatusSection` collapses to CTA-only mode when `ollama_available === false` instead of rendering an empty model table that polls forever.

**Dev mode.** `scripts/run-dev-win.sh` had grown a sequence-of-failures during this work: Git Bash `/etc/profile` strips PATH on every shell spawn; vcvars wipes PATH and degrades silently when no SDK is pinned; `pnpm` shims are `.cmd` files that need `cmd.exe` which itself needs `node` which needs PATH. Fix layers: restore Windows PATH inside child shells via a `cmd` subprocess; vcvars launcher prepends the VS Installer dir so `vswhere.exe` resolves; SDK self-discovery scans the disk and patches `LIB`/`INCLUDE`/`PATH` when vcvars doesn't pin a SDK; CEF runtime staged into `target/debug` so `libcef.dll` resolves under `pnpm dev:app:win`.

## Submission Checklist

- [ ] N/A: behaviour change with no new pure-logic surface; the new Rust paths (`ollama_binary_present`, CFA detection, crash-resume guard) are tightly coupled to filesystem and child-process state and the existing local_ai test suite doesn't mock either. UI states are conditional renders over polled status fields and don't carry independent logic worth unit-testing in isolation. Will be exercised by the local-AI E2E spec on Linux CI.
- [ ] N/A: diff coverage gate -- changed lines are filesystem/child-process integration paths and conditional renders; the codebase doesn't have the harness for either, and adding mock layers solely to lift coverage would defeat the point of the change.
- [ ] N/A: coverage matrix -- no new feature row; this hardens the existing local-AI install flow.
- [ ] N/A: matrix feature IDs -- no new IDs introduced.
- [ ] N/A: manual smoke checklist -- not a release-cut surface change; install path verified ad hoc on Windows during development.
- [x] Linked issue closed via `Closes #1475` below

## Impact

- **Platform**: Windows-primary (Controlled Folder Access detection, PowerShell fallbacks, vcvars/SDK rescue) but the precondition gate + `has_model` short-circuit benefit macOS/Linux too -- they no longer get a 30s connect timeout cascade on the first poll of a missing Ollama.
- **Performance**: removes up to ~90s of cumulative blocking on `assets_status` polls when Ollama is missing; UI surfaces install state within one poll interval (~2s) instead of after the connect-timeout chain.
- **Security**: no new attack surface. The earlier CFA UAC-elevation button was dropped -- we do not silently modify Defender state; the user gets a clear manual instruction instead.
- **Migration**: none -- new struct fields (`ollama_available`) are additive on existing RPC responses; old UI builds ignoring the field still render correctly.

## Related

- Closes #1475
- Follow-up PR(s)/TODOs: TAURI-5A (.lnk resolution in path validator), TAURI-82 (webview_apis port collision), TAURI-7K (workspace dir cleanup race) -- all surfaced during this work, all out of scope here.

---

### Linear Issue
- Key: N/A: GitHub issue, not Linear
- URL: https://github.com/tinyhumansai/openhuman/issues/1475

### Commit & Branch
- Branch: fix/ollama-precondition
- Commit SHA: bb733ab74d1e612c05d34f703d3c7c07bee7b9c1

### Validation Run
- [ ] N/A: `pnpm --filter openhuman-app format:check` -- workspace-wide Prettier check fails on ~687 unrelated files due to pre-existing CRLF/LF drift on Windows; failures are outside this PR's scope.
- [x] `pnpm typecheck` -- exit 0
- [ ] N/A: focused tests -- see Submission Checklist rationale above; no test harness for filesystem/child-process integration paths.
- [x] Rust fmt/check -- `cargo fmt --manifest-path Cargo.toml --check` clean; `cargo check --manifest-path Cargo.toml` exit 0
- [ ] N/A: Tauri fmt/check -- `app/src-tauri/` not modified by this PR.

### Validation Blocked
- `command:` pnpm --filter openhuman-app format:check
- `error:` Code style issues found in 687 files (CRLF/LF drift, all outside this PR's diff)
- `impact:` none on PR correctness; established workspace-wide issue, push uses `--no-verify` per repo norms

### Behavior Changes
- Intended behavior change: pre-Ollama state is now actionable instead of looking like an indefinite "loading"; mid-install lifecycle is observable; CFA blocks produce a remediable error message.
- User-visible effect: faster failure surface for missing Ollama; Inno Setup install dialog is visible; CFA-blocked users get specific instructions instead of "not reachable".

### Parity Contract
- Legacy behavior preserved: when Ollama IS present and healthy, the entire precondition path is a single `Path::is_file()` + 2s health probe before the existing code path runs unchanged.
- Guard/fallback/dispatch parity checks: `assets_status` still returns full asset detail when `ollama_available=true`; the new field is additive on the response.

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * App shows an “Install Ollama” CTA when Ollama is absent, hides model status, locks local tier buttons with “Needs Ollama” badges, and offers inline install path set/clear, retry, manual-install, and error-detail UI.
  * Presets and local AI status refresh after install actions.

* **Bug Fixes**
  * Resume-safe installer flow with running-installer detection, improved installer handling, CFA-aware diagnostics, and faster local server connect timeouts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1569)


src/openhuman/local_ai/service/ollama_admin.rs
```diff
@@ -77,14 +77,20 @@ impl LocalAiService {
         serve_cmd
             .arg("serve")
             .stdout(std::process::Stdio::null())
-            .stderr(std::process::Stdio::null());
+            // Pipe stderr so we can detect specific failure modes — most
+            // importantly Windows Controlled Folder Access blocks, which
+            // surface as "Access is denied" / "operation was blocked" /
+            // 0x80070005 in Ollama's own stderr when CFA refuses writes
+            // to the model cache or even prevents the binary from running.
+            .stderr(std::process::Stdio::piped());
```
graycyrus, line 85: `[critical]` **Verify this is the only `stderr(...)` call in `start_and_wait_for_server`.**

The existing function body (pre-PR) has `.stderr(Stdio::null())` at line ~80. The merge conflict likely stems from both the old and new versions coexisting. After resolving conflicts, confirm the final function has exactly **one** `.stderr(std::process::Stdio::piped())` call — not the old `.null()` version.

If the old `.null()` survives the merge, the entire stderr-capture pipeline (bounded buffer, CFA classification, `error_detail` surface) is dead code in the common `ensure_ollama_server` path.

src/openhuman/local_ai/service/ollama_admin.rs
```diff
@@ -186,6 +250,43 @@ impl LocalAiService {
...
+        // second one — two concurrent installers race on the same dir
+        // and corrupt the install.
+        if crate::openhuman::local_ai::install::is_ollama_installer_running() {
+            log::info!(
+                "[local_ai] detected in-flight OllamaSetup.exe — \
+                 waiting for it to finish before deciding whether to install"
+            );
+            {
+                let mut status = self.status.lock();
+                status.state = "installing".to_string();
```
graycyrus, line 272: `[critical]` **Crash-resume poll loop has no upper bound — can block forever.**

If `OllamaSetup.exe` hangs (network stall, hidden dialog, stuck Inno Setup), `download_and_install_ollama` — and the entire bootstrap task — blocks indefinitely. The user sees "Resuming Ollama install…" with no escape.

```rust
// suggestion: cap at 10 minutes
const RESUME_WAIT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10 * 60);
let deadline = tokio::time::Instant::now() + RESUME_WAIT_TIMEOUT;
while crate::openhuman::local_ai::install::is_ollama_installer_running() {
    if tokio::time::Instant::now() >= deadline {
        log::warn!(
            "[local_ai] crash-resume: OllamaSetup.exe still running after {}s — \
             aborting wait and running fresh install",
            RESUME_WAIT_TIMEOUT.as_secs()
        );
        break;
    }
    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
}
```

app/src/components/settings/panels/LocalModelPanel.tsx
```diff
@@ -188,14 +238,30 @@ const LocalModelPanel = () => {
           presetError={presetError}
           presetSuccess={presetSuccess}
           formatRamGb={formatRamGb}
+          ollamaAvailable={downloads?.ollama_available ?? true}
```
graycyrus, line 246: `[major]` **`?? true` default causes the Model Status section to flash visible on first render.**

Before the first `downloads` poll resolves, this evaluates to `true`, so the full model-status card renders momentarily, then blinks away if Ollama is absent.

Suggestion: default to `undefined` so the install/model sections stay hidden until the first poll resolves:

```tsx
ollamaAvailable={downloads?.ollama_available}
```

And in `DeviceCapabilitySection.tsx`, change the destructured default from `ollamaAvailable = true` to just `ollamaAvailable` — the `locked` check (`ollamaAvailable === false`) already handles `undefined` correctly (treats it as not-locked).

Same for the conditional wrapper around the Model Status `<section>` further down:
```tsx
{downloads != null && downloads.ollama_available && (
```

src/openhuman/local_ai/service/ollama_admin.rs
```diff
@@ -105,6 +140,35 @@ impl LocalAiService {
...
         }
 
+        // Health probe timed out. Classify the failure from captured stderr.
+        let stderr_snapshot = stderr_buffer.lock().clone();
+        let lowered = stderr_snapshot.to_ascii_lowercase();
+        let cfa_signatures = [
+            "operation was blocked",
+            "controlled folder access",
+            "access is denied",
+            "0x80070005",
```
graycyrus: `[major]` **CFA detection will false-positive on generic permission errors.**

`"access is denied"` and `"is not recognized as a trusted"` are extremely common Windows error strings that appear in contexts completely unrelated to Controlled Folder Access (antivirus policy, UAC, missing registry keys). Mis-classifying a generic permission error as CFA and telling the user to go to "Ransomware protection" will be confusing.

Suggestion — require explicit CFA markers:

```rust
let cfa_hit = lowered.contains("controlled folder access")
    || lowered.contains("operation was blocked by ransomware")
    || (lowered.contains("0x80070005")
        && (lowered.contains("controlled") || lowered.contains("folder")));
```

For plain `"access is denied"` without CFA context, fall through to the generic stderr-tail display.

app/src/components/settings/panels/local-model/ModelStatusSection.tsx
```diff
@@ -76,8 +76,112 @@ const ModelStatusSection = ({
...
+            <div className="text-xs text-amber-800">
+              Local AI features (chat, vision, embedding) need the Ollama runtime. Install it
+              below — the installer runs silently and lands in your workspace; no console window
+              will appear.
+            </div>
+          </div>
+        </div>
+        <div className="flex items-center gap-2 pt-1">
+          <button
+            onClick={() => onTriggerDownload(true)}
```
graycyrus, line 121: `[major]` **Install button missing `type="button"`** — defaults to `type="submit"` which will submit any enclosing `<form>` if one is ever added. Also lacks an accessible label for screen readers.

```tsx
<button
  type="button"
  aria-label={isTriggeringDownload ? 'Installing Ollama' : 'Install Ollama runtime'}
  onClick={() => onTriggerDownload(true)}
  ...
```

Same applies to the `Set Path` and `Clear` buttons further down in this block.

src/openhuman/local_ai/service/ollama_admin.rs
```diff
@@ -909,6 +1040,13 @@ impl LocalAiService {
         &self,
         model: &str,
     ) -> Result<bool, String> {
+        // Short-circuit: if the server isn't reachable, no model can be installed.
+        // `ollama_healthy()` runs a 2s-bounded /api/tags probe — without this guard
+        // a missing server cascades through up to three sequential 30s connect
+        // timeouts on every `assets_status` call.
+        if !self.ollama_healthy().await {
+            return Ok(false);
+        }
```
graycyrus: `[minor]` **`has_model` short-circuit doubles the round-trips to `/api/tags`.**

`ollama_healthy()` sends a 2s-bounded GET to `/api/tags`, then `has_model` sends the same GET again. The intent (avoid cascading 30s timeouts) is correct, but the 500ms connect timeout in `bootstrap.rs` already handles the fast-fail case.

Consider merging into one request — treat connection/timeout errors as `Ok(false)`:

```rust
let response = match self.http
    .get(format!("{}/api/tags", ollama_base_url()))
    .send().await
{
    Ok(r) => r,
    Err(e) if e.is_connect() || e.is_timeout() => return Ok(false),
    Err(e) => return Err(format!("ollama tags request failed: {e}")),
};
```

app/src/components/settings/panels/local-model/DeviceCapabilitySection.tsx
```diff
@@ -24,7 +50,15 @@ const DeviceCapabilitySection = ({
...
   formatRamGb,
   onPresetApplied,
+  ollamaAvailable = true,
+  onTriggerOllamaInstall,
+  isTriggeringInstall = false,
+  installState,
+  installWarning,
+  installError,
 }: DeviceCapabilitySectionProps) => {
+  const installInProgress = installState === 'installing' || installState === 'downloading' || installState === 'loading';
```
graycyrus: `[minor]` **`installInProgress` doesn't include `'starting'` state.**

`LocalAiStatus.state` can emit `'starting'` during server startup (e.g. inside `start_and_wait_for_server`). A user who triggers install and then sees the server spinning up will see the amber "Install Ollama first" banner instead of the blue progress banner.

```tsx
const installInProgress =
  installState === 'installing' ||
  installState === 'downloading' ||
  installState === 'loading' ||
  installState === 'starting';
```

graycyrus (commented):
## Review Summary

7 findings (2 blockers, 3 major, 2 minor/refactor). Skipping the 4 items CodeRabbit already flagged.

**Blocking gate:** PR has merge conflicts (`CONFLICTING/DIRTY`) — must rebase on `main` before merge.

**Scope note:** Issue #1475 is "Rebuild Windows MSI with Sentry fix" but the PR pivoted to Ollama install UX hardening. Author acknowledges this in the PR body.

**Test coverage:** Zero new tests. The diff-cover gate (≥80% changed lines) will likely fail. At minimum needs tests for `ollama_binary_present`, `resolve_powershell_executable`, CFA classification logic, and the frontend install-state banner transitions.

**Looks good:** `ollama_binary_present` is clean filesystem-only; `kill_on_drop(true)` on all 3 platform branches; 500ms connect timeout placed at client construction; `ollama_available` propagated consistently to both DTOs; "Disabled (cloud fallback)" tile stays enabled; `/SILENT` change is well-motivated; Windows SDK self-discovery guards on `kernel32.lib`.

senamakel:
**Deferred reviewer item — stderr drain task lifetime**

During review, a question was raised about the lifetime of the background task that drains Ollama's stderr ring buffer. For the record: the drain task is bounded by the 16 KB ring buffer (a fixed-capacity circular buffer), so even if the task outlives the health probe it cannot grow unboundedly. The task exits naturally when the stderr pipe closes (i.e., when the Ollama process exits). No action needed — noting here so it's not lost.

senamakel:
**Pre-existing issue noted during review (outside this PR's diff) — `ollama_runner_ok` uses POST on a GET endpoint**

`ollama_runner_ok` was observed to issue a POST request to `/api/tags`, which is a GET endpoint. This is pre-existing code not touched by this PR. Flagging for a follow-up fix in a separate PR.

senamakel:
**Nitpick (outside this PR's primary diff) — `scripts/run-dev-win.sh` comment inaccuracy**

A comment in `run-dev-win.sh` describes the PATH restoration as restoring "full machine + user PATH", but the actual implementation only restores the machine PATH. The comment is misleading. Low priority — suggest fixing the comment wording in a cleanup pass.

PR #1620, opened 2026-05-13
Fix UTF-8 body_preview slicing in memory ingest

### LINKED ISSUE #1595: Panic in memory ingestion: byte index slicing on multi-byte UTF-8 char boundary
## Summary

Rust panic in `openhuman_core::openhuman::memory::tree::ingest::persist` when slicing markdown content for `body_preview`. The code slices at a fixed byte offset (`len - 2048`) which can land inside a multi-byte UTF-8 character, causing a fatal panic.

## Sentry Issues

- [OPENHUMAN-TAURI-CP](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-CP) — 4 events
- [OPENHUMAN-TAURI-CK](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-CK) — 1 event

**Error:** `panic: byte index 1323 is not a char boundary; it is inside '\u{200c}' (bytes 1321..1324)`
**Severity:** fatal (unhandled panic)
**Environment:** staging
**Affected platform:** macOS (MacBookAir7,2, x86_64)

## Root Cause

In `src/openhuman/memory/tree/ingest.rs` lines 169-173:

```rust
Some(if len <= 2048 {
    md.clone()
} else {
    md[len - 2048..].to_string()  // <-- panics here
})
```

The slice `md[len - 2048..]` uses a byte offset, but `String::len()` returns byte length. When the byte at position `len - 2048` falls inside a multi-byte UTF-8 sequence (like `\u{200c}` zero-width non-joiner, which is 3 bytes), Rust panics.

## Suggested Fix

Use `floor_char_boundary` (stable since Rust 1.82) or manually walk backward to find a valid char boundary:

```rust
let start = md.floor_char_boundary(len - 2048);
md[start..].to_string()
```

## Impact

- Crashes the memory ingestion pipeline for any document/email containing multi-byte Unicode near the 2048-byte-from-end boundary
- Fatal panic — kills the task, content is not ingested
- Likely underreported: only triggers when a multi-byte char happens to straddle the exact cut point

**Priority:** Critical

honor2030:
## Summary
- avoid panics when `body_preview` starts inside a multi-byte UTF-8 character
- centralize preview truncation in `markdown_body_preview` with a 2048-byte cap
- add regression coverage for the UTF-8 boundary case

Fixes #1595.

## Verification
- `cargo test -p openhuman markdown_body_preview_respects_utf8_boundary_and_byte_cap --lib -- --nocapture`
- `cargo test -p openhuman ingest_document_handles_utf8_at_body_preview_boundary --lib -- --nocapture`
- `cargo test -p openhuman openhuman::memory::tree::ingest::tests --lib`
- `cargo check -p openhuman --lib`
- `cargo fmt --all --check`
- pre-push hook: `format:check`, `lint`, `compile`, `rust:check`, `lint:commands-tokens`

Notes: `cargo check` / pre-push report existing warnings only; no new failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Email and document previews now enforce a strict byte-length cap and never cut multibyte UTF-8 characters, preventing corrupted or malformed preview text.
* **Tests**
  * Added unit tests to verify preview truncation respects the byte cap and UTF-8 boundaries and that ingestion succeeds when previews would otherwise split multibyte characters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1620)


src/openhuman/memory/tree/ingest.rs
```diff
@@ -485,6 +492,46 @@ mod tests {
...
+    #[tokio::test]
+    async fn ingest_document_handles_utf8_at_body_preview_boundary() {
+        let (_tmp, cfg) = test_config();
+        let body = format!("{}{}{}", "a".repeat(17), '\u{200c}', "b".repeat(2045));
+        let preview_start = body.len() + 1 - BODY_PREVIEW_MAX_BYTES;
+        assert!(
+            !body.is_char_boundary(preview_start),
+            "test fixture must put the preview boundary inside a multi-byte character"
+        );
+
```
graycyrus, line 519: **[major]** Integration test bakes in a canonicalizer implementation detail

`body.len() + 1` assumes `document::canonicalise` adds exactly 1 byte of overhead (a trailing `\n`). If the canonicalizer ever adjusts formatting, the `+ 1` silently becomes wrong and the `assert!` guard either passes for the wrong reason or breaks without making the regression obvious.

The unit test already proves the helper in isolation against the exact boundary condition. This integration test's job is just to show `ingest_document` survives — consider removing the byte-arithmetic precondition entirely:

```rust
// The unit test verifies exact boundary arithmetic.
// Here we just confirm the full pipeline doesn't panic on this input.
```

If a guard is desired, measure after canonicalization, not before.

src/openhuman/memory/tree/ingest.rs
```diff
@@ -361,6 +358,16 @@ async fn persist(
     })
 }
 
+fn markdown_body_preview(md: &str) -> String {
+    let len = md.len();
+    if len <= BODY_PREVIEW_MAX_BYTES {
```
graycyrus, line 368: **[minor]** Consistency note — the codebase has `crate::openhuman::util::floor_char_boundary` used at ~9 other call sites. This PR uses the stdlib `str::ceil_char_boundary` directly, which is semantically correct (ceil for a start index, floor for an end index) and works on the pinned Rust 1.93.0 toolchain.

Not a bug, but for discoverability a one-line comment would help a future reader:

```rust
// ceil_char_boundary (stable since Rust 1.93) advances the index
// to the next char boundary, keeping the slice <= BODY_PREVIEW_MAX_BYTES bytes.
```

src/openhuman/memory/tree/ingest.rs
```diff
@@ -361,6 +358,16 @@ async fn persist(
     })
```
graycyrus, line 357: **[nitpick]** Missing doc comment — every other function in this file (public and private) has at least a one-liner. Suggestion:

```rust
/// Returns the trailing at-most `BODY_PREVIEW_MAX_BYTES` bytes of `md`,
/// aligned to a UTF-8 character boundary.
```

This would also close the docstring coverage gap CodeRabbit flagged (66.67% -> 80%+).

graycyrus (commented):
## Review — Fix UTF-8 body_preview slicing in memory ingest

### Walkthrough

This PR fixes a real panic: `persist` in the memory ingest pipeline was slicing `canonical.markdown` at a fixed byte offset without checking whether it landed on a UTF-8 character boundary. For documents/emails whose canonical markdown exceeded 2048 bytes, any multi-byte character at the slice point (e.g. zero-width non-joiner, CJK codepoint, emoji) would panic. The fix extracts the logic into a `markdown_body_preview` helper using `str::ceil_char_boundary` to advance past any partial character before slicing, keeping the result within the byte cap. Two regression tests added.

### Changes

| File | Summary |
|---|---|
| `src/openhuman/memory/tree/ingest.rs` | Extracts `markdown_body_preview` helper; replaces unsafe byte-slice with `ceil_char_boundary`; adds `BODY_PREVIEW_MAX_BYTES` constant; adds 2 regression tests |

### Verified / looks good

- `ceil_char_boundary` (stable since Rust 1.93.0) is the correct choice for a *start* index — rounding up keeps the result ≤ 2048 bytes
- Unit test fixture correctly triggers the boundary condition (offset 18 inside 3-byte `\u{200c}`)
- `BODY_PREVIEW_MAX_BYTES` is module-private, matching the helper's scope
- Email/Document match arms are the only callers; Chat correctly excluded
- No external API surface changed
- Pre-existing `tracing::debug!` on `DocumentCanonicalized` is untouched

senamakel (changes requested):
pf above

PR #1542, opened 2026-05-12
fix(si_server): skip embedded server autostart on non-macOS (OPENHUMAN-TAURI-50)

oxoxDev:
## Summary

- Skip embedded `si_server` autostart on non-macOS at the spawn site so cores no longer crash with `accessibility automation is macOS-only in V1`. Sentry issue [OPENHUMAN-TAURI-50](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-50) (~15 events).
- Demote residual macOS-only failure logs from `error!` to `info!` for defense-in-depth.
- macOS path unchanged.

## Problem

`screen_intelligence::server::start_if_enabled` spawns the embedded screen-intelligence server on every platform during core boot. The underlying accessibility engine is macOS-only in V1 (per the existing check at `engine.rs:95-96`), so on Windows/Linux the spawned task immediately errors and surfaces as Sentry [OPENHUMAN-TAURI-50](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-50) (~15 events). The error log is the noise — the actual feature gap is expected and documented.

## Solution

- `src/openhuman/screen_intelligence/server.rs::start_if_enabled`: runtime `if !cfg!(target_os = "macos")` guard returns early with an `info!` log explaining the V1 platform constraint. Avoids `#[cfg]`-gating the function symbol so non-mac callers still compile.
- Same file: when the inner `start_session` or spawned `run()` does fail with a macOS-only message (defense-in-depth — covers any future caller that bypasses `start_if_enabled`), demote `error!` → `info!` so any residual non-mac entrypoint doesn't pollute Sentry.
- Existing engine-side check at `engine.rs:95-96` left intact.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — non-macOS cfg-gated test asserts `start_if_enabled` returns without panic and server is `Stopped`; macOS cfg-gated test asserts global server initializes
- [x] **Diff coverage ≥ 80%** — both platform arms covered plus both demoted error sites; total diff is +47/-2 in a single file
- [x] Coverage matrix updated — `N/A: platform-gating fix, no new feature row`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: behaviour-only change`
- [x] No new external network dependencies introduced (mock backend used per Testing Strategy)
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface touched`
[...]

PR #1625, opened 2026-05-13
fix(si_server): skip Sentry for benign "session already active" start_session race (OPENHUMAN-TAURI-5H)

CodeGhost21:
## Summary

- Extends the SI server's `start_session` error classifier with a new sibling, `is_expected_session_already_active_failure`, plus a union helper `is_expected_benign_start_session_failure`.
- Both `SiServer::run` (start_session error arm) and `start_if_enabled`'s spawned future now consume the union helper, so duplicate `start_session` calls log at `info!` instead of `error!` — sentry-tracing records at most a breadcrumb.

Fixes OPENHUMAN-TAURI-5H (16 macOS occurrences in 4 days on `openhuman@0.53.22`).

## Why this is benign, not a failure

`engine::start_session` returns `"session already active"` ([`engine.rs:107`](src/openhuman/screen_intelligence/engine.rs#L107)) when the engine already has a live session. This happens at the embedded server entry point when something else got there first:

- `screen_intelligence.enable` invoked over RPC (UI/onboarding flow) before the embedded `run()` reached its own `start_session`, or
- A prior `run()` cycle left the engine session in place across a server restart.

The session is *running*. The embedded server's `start_session` is a duplicate, not a real failure — there's no remediable Sentry signal.

## Why introduce a union helper instead of inlining

The classifier sits at two call sites today (`SiServer::run`, `start_if_enabled`). A union helper keeps the log-level decision in lockstep across both — same pattern as PR #1542 (`is_expected_macos_only_failure`).

## Test plan

- [x] `cargo test --lib openhuman::screen_intelligence::server` — 9 tests pass.
- [x] `is_expected_session_already_active_failure_classifies_known_signal` covers the bare wire shape, a wrapped/prefixed form, and pins unrelated session-domain failures (permission errors, IO errors) so they still surface as Sentry events.
- [x] `is_expected_benign_start_session_failure_is_union_of_classifiers` pins the union helper's contract against both known-benign shapes.
- [x] `cargo check --lib` — no new warnings.
- [x] `cargo fmt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Screen intelligence server now treats duplicate session start attempts (a known benign condition) as non-fatal and logs them at info level instead of failing.
  * Error logging improved to distinguish benign start-session failures from critical errors.

* **Tests**
  * Added unit tests covering duplicate session detection and the combined error-classification behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1625)

### DIFF
--- a/src/openhuman/screen_intelligence/server.rs
+++ b/src/openhuman/screen_intelligence/server.rs
@@ -156,7 +156,7 @@ impl SiServer {
                 );
             }
             Err(e) => {
-                if is_expected_macos_only_failure(&e) {
+                if is_expected_benign_start_session_failure(&e) {
                     info!("{LOG_PREFIX} failed to start session: {e}");
                 } else {
                     error!("{LOG_PREFIX} failed to start session: {e}");
@@ -363,7 +363,7 @@ pub async fn start_if_enabled(app_config: &Config) {
 
     tokio::spawn(async move {
         if let Err(e) = server.run(&config_for_run).await {
-            if is_expected_macos_only_failure(&e) {
+            if is_expected_benign_start_session_failure(&e) {
                 info!("{LOG_PREFIX} embedded server autostart skipped: {e}");
             } else {
                 error!("{LOG_PREFIX} embedded server exited with error: {e}");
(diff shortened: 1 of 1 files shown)


src/openhuman/screen_intelligence/server.rs
```diff
@@ -382,6 +382,32 @@ fn is_expected_macos_only_failure(err: &str) -> bool {
...
 
+/// True when an engine-side `start_session` error is a benign idempotency
+/// signal — the engine already has a live session, so the embedded server's
+/// own `start_session` call is a duplicate, not a failure. Canonical wire
+/// shape from `engine::start_session`:
+///
+///     "session already active"
+///
+/// Reaches the server when the engine session was created via a different
+/// entry point before the embedded `run()` got there — typically
```
graycyrus, line 395: **[critical]** Doctest compilation failure — CI is red.

The indented `"session already active"` in the doc comment (4-space indent) makes rustdoc treat it as a fenced code block / doctest. The bare string expression returns `&'static str` inside an implicit `fn main() -> ()`, causing `error[E0308]: mismatched types`.

Fix: mark the block as `text` so rustdoc skips it:

```rust
/// ```text
/// "session already active"
/// ```
```

Or use a list item instead of indented code:

```rust
/// - `"session already active"`
```

graycyrus (commented):
## Walkthrough

Adds a new `is_expected_session_already_active_failure` classifier for the benign "session already active" error from `engine::start_session`, unions it with the existing `is_expected_macos_only_failure` via `is_expected_benign_start_session_failure`, and swaps both call sites to the union helper. Net effect: duplicate `start_session` calls log at `info!` instead of `error!`, silencing the OPENHUMAN-TAURI-5H noise in Sentry.

## Change Summary

| File | Change | Description |
|------|--------|-------------|
| `src/openhuman/screen_intelligence/server.rs` | Modified | Added two new classifier fns, replaced call sites, added 2 unit tests |

## Analysis

Clean, focused observability fix. Follows the established pattern from `is_expected_macos_only_failure` exactly — private classifier, union helper, both call sites updated in lockstep. Tests cover canonical wire shape, substring matching (wrapped/prefixed), negative cases for unrelated failures, and the union contract. Good negative-case discipline (permission errors, IO errors pinned as non-benign).

One CI-blocking issue found (see inline comment).

Issue #80, opened 2026-03-30
Local AI: guided model selection by device capability (tiered presets + user override)


senamakel:
## Summary

Today local inference is driven by **static config** (`model_id` / `chat_model_id`, etc.) with defaults such as a **4B-class** chat model in code (`src/openhuman/local_ai/model_ids.rs`). There is no first-class **selection flow** that helps users pick a model that fits **their machine** (RAM, GPU/CPU, thermal headroom). We want a **clear onboarding or settings path** so users can choose **presets** (e.g. lightweight vs balanced vs large) informed by **device profile**, with room to **override** for power users.

## Problem

- **One size does not fit all:** A default aimed at "typical" hardware can **OOM, run slowly, or drain battery** on low-end devices; conversely, high-end users may want **larger context or quality** without hand-editing config.
- **Opaque IDs:** Model names (Ollama tags, quant variants) are not self-explanatory for non-technical users.
- **No structured link** between **reported or inferred capabilities** and **recommended artifacts** (download size, VRAM/RAM hints).

## Goals

1. **Tiered recommendations** — Define a small set of **presets** (names are illustrative, not final), e.g.:
   - **Low / entry:** ~**1B** (or smallest supported instruct model) — minimal RAM, CPU-friendly.
   - **Medium:** ~**4B** — default for "balanced" laptops.
   - **High:** ~**8B–16B** (or project-supported upper bound) — where RAM/GPU allows.

   Exact model IDs, quantizations, and caps should come from **product + engineering** review (availability in Ollama, licensing, bundle size).

2. **Selection flow (UX)** — A dedicated step or settings surface that:
   - Explains **tradeoffs** (speed vs quality vs disk).
   - Shows **recommended preset** based on **device signals** (see below).
   - Allows **manual choice** and **advanced** path (custom model ID / existing config fields).

3. **Device-aware hints** — Use **best-effort signals** to pick a default recommendation (not a hard gate), e.g.:
   - **Total RAM**, **CPU core count**, **GPU presence** (platform-specific via Tauri/OS APIs where feasible).
   - Optional: user **self-report** ("This machine is slow / I'm on battery") if hardware APIs are insufficient.

4. **Persistence** — Selected preset maps to existing **`local_ai`** fields (e.g. `chat_model_id`, related vision/embed if we define bundles per tier) and stays in user config.

5. **Graceful degradation** — If a chosen model **fails to load** or **OOMs**, surface a **clear error** and suggest **stepping down** a tier.

[...]

PR #1630, opened 2026-05-13
fix(integrations): fall back to default backend when api_url points at local AI (#51, #80, #7Z)

oxoxDev:
## Summary
- Users who set `config.api_url` to a local-AI endpoint (Ollama `:11434/v1`, vLLM `:8080/...`) had every backend integration request concatenated onto that URL, flooding Sentry with 404s from Ollama/vLLM and silently breaking Composio for them.
- Guard at `IntegrationClient::build_client`: if the override looks like a local-AI endpoint (loopback host or `/v1/chat/completions` path), fall through to env/default for the integrations base. Chat completions still use the override.

## Problem
`effective_api_url` returns whatever the user set in `config.api_url`. Local-AI users set it to their Ollama / vLLM URL so chat completions hit the local model. But `IntegrationClient::build_client` reused the same URL as the base for ALL backend-proxied paths (composio, channels, teams, billing, …). Result: requests like `/agent-integrations/composio/toolkits` got concatenated onto the local-AI URL → 404 from the local LLM → `report_error` to Sentry → flood, plus silent UX breakage where Composio just didn't work for these users.

Sentry IDs in scope (~13 events total in 3d):
- `OPENHUMAN-TAURI-51` — 11 events: `GET http://127.0.0.1:11434/v1/agent-integrations/composio/toolkits 404`
- `OPENHUMAN-TAURI-80` — 1 event: `GET http://127.0.0.1:8080/v1/chat/completions/agent-integrations/composio/connections 404` (double-concat onto vLLM)
- `OPENHUMAN-TAURI-7Z` — 1 event: same shape via `llm_provider.api_error: OpenHuman API error (404 Not Found)` with the Ollama 404 page body

## Solution
- New `looks_like_local_ai_endpoint(url) -> bool` in `src/api/config.rs`. Tight heuristic: loopback host (`127.0.0.1` / `localhost` / `::1` / `0.0.0.0`) OR path explicitly names the OpenAI chat-completions endpoint (`/v1/chat/completions` / `/v1/completions`). Bare `/v1` is **not** matched — many self-hosted backends use it as a legit version prefix and over-matching would silently break real users.
- New `effective_integrations_api_url(api_url) -> String` — same resolution chain as `effective_api_url`, but the user override is skipped when it matches `looks_like_local_ai_endpoint`, falling through to env / default backend. One-shot `warn!` (via `std::sync::Once`) when the fallback fires so users see the diagnostic in core logs.
- `IntegrationClient::build_client` now uses the new helper. `effective_api_url` semantics are unchanged — local-AI chat needs the override to keep working.

Considered (and deferred): splitting `config.api_url` into a separate `backend_api_url` field. That's the "correct" architectural fix but it's a schema migration touching the config UI, dotfile reader, and migration story. The current guard is narrow, reversible, and ships the user-facing fix today; the split can come later if anyone wants explicit control of the integrations base independently.

## Submission Checklist
- [x] Unit tests added for `looks_like_local_ai_endpoint` (loopback v4/v6, chat-completions path on non-loopback, real backends incl. `api.openai.com/v1`, garbage input)
- [x] Unit tests added for `effective_integrations_api_url` (local override → default, local override + env → env, real override respected, no-override agrees with `effective_api_url`)
- [x] `cargo test --lib openhuman::integrations` passes (67/67)
- [x] `cargo test --lib api` passes (124/124, includes new helper tests)
- N/A: no UI change in this PR
- N/A: no schema migration (deliberately narrow; separate `backend_api_url` field deferred — see Solution)

`cargo clippy --workspace -- -D warnings` is currently failing on `upstream/main` itself (78 errors in `src/openhuman/wallet/execution.rs`, `webview_apis/client.rs`, `webview_notifications/types.rs`, `tools/impl/cron/add.rs` — all `manual_is_multiple_of` / `type_complexity` / `derivable_impls` from a clippy version bump). None of the errors touch the files in this PR. Confirmed via `grep -E "api/config\.rs|integrations/client\.rs"` over the clippy output — zero hits.

## Impact
- Stops the OPENHUMAN-TAURI-51 / -80 / -7Z Sentry cluster (~13 events / 3d) at the source.
- Restores Composio + channels + teams functionality for users who route chat to a local LLM — they were silently broken before.
- Zero change to chat-completions routing: `effective_api_url` is untouched.

## Related
- Closes OPENHUMAN-TAURI-51, OPENHUMAN-TAURI-80, OPENHUMAN-TAURI-7Z

## Summary by CodeRabbit

* **New Features**
  * Detects and classifies local AI endpoints (loopback/localhost/0.0.0.0/IPv6, private LAN IPs, and OpenAI-style chat paths) to handle them differently.
* **Bug Fixes**
  * Integration client routing now avoids sending integration paths to local AI services, preventing 404s and noisy error reports.
* **Tests**
  * Added unit tests for endpoint detection heuristics and integrations URL resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1630)


graycyrus (commented):
## Review — PR #1630

### Walkthrough

This PR fixes a real and actively-reported user pain point: anyone who sets `config.api_url` to a local Ollama or vLLM endpoint for chat completions had every backend-proxied integration request (Composio toolkits, channels, billing, teams, credentials, ...) concatenated onto that URL, producing 404s from the local LLM and a Sentry flood. The fix introduces `looks_like_local_ai_endpoint`, a carefully-scoped heuristic that detects loopback hosts and explicit chat-completions paths, then `effective_integrations_api_url`, which skips the user override when it looks like a local AI endpoint and falls through to env/default instead. The implementation is clean, well-documented, and comes with solid test coverage.

The core mechanic is correct and the Sentry cluster will stop firing for `IntegrationClient` callers once this ships.

### Changes

| File | Summary |
| --- | --- |
| `src/api/config.rs` | New `looks_like_local_ai_endpoint` heuristic, `effective_integrations_api_url` resolver, `warn_integrations_url_fallback_once` one-shot logger, and 8 unit tests |
| `src/openhuman/integrations/client.rs` | `build_client` switched from `effective_api_url` to `effective_integrations_api_url` |

---

### Actionable comments

#### [major] `src/api/config.rs:74-75` — `contains` vs `ends_with` inconsistency in path check

The check for `/v1/chat/completions` uses `path.contains(...)` but `/v1/completions` uses `path.ends_with(...)`. A URL like `https://real-backend.example/audit/v1/chat/completions-logs` would match `contains` and be misclassified. Both arms should use `ends_with` for consistency:

```rust
// before
path.contains("/v1/chat/completions") || path.ends_with("/v1/completions")
// after
path.ends_with("/v1/chat/completions") || path.ends_with("/v1/completions")
```

#### [major] `src/openhuman/integrations/client.rs:241-246` — doc comment still references old resolver

The rustdoc says `effective_api_url` but the implementation now uses `effective_integrations_api_url`. Should be updated to match.

#### [minor] Incomplete fix scope — 38+ other `effective_api_url` callers remain broken for local-AI users

The same class of bug exists across most non-integrations domains: `billing/ops.rs`, `team/ops.rs`, `referral/ops.rs`, `webhooks/ops.rs`, `credentials/ops.rs` (9 call sites), `channels/controllers/ops.rs` (9 call sites), `voice/`, `socket/`, `app_state/`, `core/jsonrpc.rs`, etc. The cleanest near-term fix would be to rename `effective_integrations_api_url` to `effective_backend_api_url` and make it the default for all non-chat callers. This PR makes things no worse (these were already broken), but a follow-up issue should be filed explicitly tracking these callers.

### Nitpick

- `src/api/config.rs:51` — `trimmed` is computed but `url::Url::parse` receives `trimmed` correctly. However, the `url::Url::parse(trimmed)` line is fine as-is — ignore if no concern.

### Verified / looks good

- `url::Host` typed matching handles IPv4-mapped IPv6 and `::1` correctly
- `std::sync::Once` one-shot warning prevents log spam
- No panic on malformed URLs (garbage input test confirms)
- `ENV_LOCK` mutex used consistently with existing test suite
- CI failures are all in `composio::ops::tests` — pre-existing, unrelated to this PR
- No secrets or PII in new log output

oxoxDev:
graycyrus review Round 1 addressed in `39d60eb3`:

- [Major] `looks_like_local_ai_endpoint` path check now uses `ends_with` on both arms; new unit test `looks_like_local_ai_rejects_substring_path_false_positives` locks the false-positive case (`/audit/v1/chat/completions-logs` → false, plus a couple of other substring-shaped real-backend paths).
- [Major] `build_client` rustdoc updated to reference `effective_integrations_api_url` and call out *why* it differs from the plain `effective_api_url` resolver (local-AI fallback).
- [Minor / scope] Filed follow-up #1663 tracking the 38+ other `effective_api_url` callers across billing / team / referral / webhooks / credentials / channels / voice / socket / app_state / core/jsonrpc. The TODO note above `effective_integrations_api_url` references it. Renaming `effective_integrations_api_url` → `effective_backend_api_url` and migrating those call sites is a much larger surface, so I'd rather ship it separately under that issue than expand this PR.

Open question: do you want the typed-`Url` host check kept as-is, or would you prefer a tighter allowlist (e.g. explicit `.is_loopback()` only without the `*.localhost` domain match)? Happy to tighten further if so.

senamakel:
**Deferred items (context for reviewers — no action needed on this PR)**

Two minor items were identified during the review pass but are intentionally out of scope:

1. **File size** — `src/api/config.rs` is now ~600 lines, slightly over the project's preferred ~500-line ceiling. The overage is test-only (`#[cfg(test)]` block). Splitting the tests into a sibling `src/api/config/tests.rs` would be the right long-term move, but is a pure refactor with no behaviour change. Deferred to a follow-up.

2. **`EnvGuard` test helper** — several tests in this file set/unset env vars by hand (`std::env::set_var` / `remove_var`) with manual teardown. A small `EnvGuard` RAII helper would eliminate the boilerplate and make tests panic-safe. Deferred as a refactor-only item.

Neither item affects correctness or CI for this PR.

PR #1639, opened 2026-05-13
fix(whatsapp_data): preserve anyhow cause chain in RPC error envelopes (OPENHUMAN-TAURI-6B)

CodeGhost21:
## Summary

- Switches the four `whatsapp_data` RPC handlers (`whatsapp_data_ingest`, `…_list_chats`, `…_list_messages`, `…_search_messages`) from plain `{e}` to alternate `{e:#}` Display rendering in their `map_err` envelopes.
- Same anyhow-chain-rendering fix called out in the `report_error` doc-string in `core/observability.rs` (referencing OPENHUMAN-TAURI-B2 — _\"Without this, anyhow's default `to_string()` only emits the outermost context and the underlying cause is dropped — making the captured Sentry event undiagnosable.\"_).

This is **not** a Sentry-skip — it's the inverse. The current Sentry events are too vague to act on; the fix makes them actionable so we can diagnose and fix the actual root cause.

Fixes OPENHUMAN-TAURI-6B (12 occurrences in 3 days, Linux user on `openhuman@0.53.22`).

## Why the chain matters here

[`store::upsert_chats`](src/openhuman/whatsapp_data/store.rs#L115) wraps every rusqlite error with `.with_context(|| format!(\"upsert wa_chat {chat_id}\"))`. The anyhow chain at the failure site is:

- **outer**: `\"upsert wa_chat 919xxxxxxxxxx@s.whatsapp.net\"`
- **cause**: `rusqlite::Error` — could be `SQLITE_BUSY`, schema mismatch, FK violation, disk-full, …

`format!(\"... {e}\")` renders only the outer label. The 12 Sentry events all read `\"[whatsapp_data] ingest failed: upsert wa_chat <jid>\"` with **no underlying cause** — we can't tell which SQLite condition is firing on Linux ingest.

`{e:#}` (anyhow's alternate format) joins the outer context plus every wrapped cause with `\": \"`, so the next Sentry event surfaces the rusqlite detail we need to actually fix the underlying issue.

## Why update all four handlers, not just `ingest`

Only `ingest` has Sentry occurrences today, but the other three use the identical pattern and would mask the same class of bug on any future failure. Single sibling pass keeps the file consistent and forecloses the next \"vague \\$method failed: \\$context\" Sentry issue from `whatsapp_data`.

## Test plan

- [x] `cargo check --lib` — no new warnings.
- [x] `cargo fmt`.
- [x] No new tests: the `{e:#}` rendering contract is already covered by `core::observability::tests::anyhow_chain_is_rendered_in_full`. A behavioural test here would require a running global store + synthetic SQLite failure injection — disproportionate for a four-site format-specifier fix.

## Follow-up

Once a fresh Sentry event lands on `main` post-merge, the actual rusqlite cause will be visible and a targeted follow-up PR can fix the underlying SQLite condition (likely lock contention given the 23-second `elapsed_ms` on the reported event, but the chain will tell us for sure).

## Summary by CodeRabbit

* **Chores**
  * Enhanced error message formatting and logging for WhatsApp data operations (ingest, chat listing, message listing, and search). Improved error reporting provides better visibility into underlying system issues for troubleshooting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1639)

### DIFF
--- a/src/openhuman/whatsapp_data/rpc.rs
+++ b/src/openhuman/whatsapp_data/rpc.rs
@@ -40,8 +40,15 @@ pub async fn whatsapp_data_ingest(req: IngestRequest) -> Result<RpcOutcome<Inges
     );
     let store = require_store()?;
     let result = ops::ingest(&store, req).map_err(|e| {
-        log::warn!("[whatsapp_data][rpc] ingest error: {e}");
-        format!("[whatsapp_data] ingest failed: {e}")
+        // {e:#} renders the full anyhow chain — store::upsert_chats wraps
+        // the rusqlite error with `.with_context("upsert wa_chat <jid>")`,
+        // so plain Display ({e}) only emits the outer context and drops
+        // the underlying SQLite cause (database locked, schema mismatch,
+        // FK violation, …) before it reaches Sentry. See OPENHUMAN-TAURI-6B
+        // / TAURI-B2 for the canonical instance where this masked the
+        // real failure.
+        log::warn!("[whatsapp_data][rpc] ingest error: {e:#}");
+        format!("[whatsapp_data] ingest failed: {e:#}")
     })?;
     log::debug!(
         "[whatsapp_data][rpc] ingest ok chats={} messages={} pruned={}",
@@ -67,8 +74,8 @@ pub async fn whatsapp_data_list_chats(
     );
     let store = require_store()?;
     let chats = ops::list_chats(&store, req).map_err(|e| {
-        log::warn!("[whatsapp_data][rpc] list_chats error: {e}");
-        format!("[whatsapp_data] list_chats failed: {e}")
+        log::warn!("[whatsapp_data][rpc] list_chats error: {e:#}");
+        format!("[whatsapp_data] list_chats failed: {e:#}")
     })?;
     log::debug!("[whatsapp_data][rpc] list_chats ok count={}", chats.len());
     Ok(RpcOutcome::single_log(
@@ -87,8 +94,8 @@ pub async fn whatsapp_data_list_messages(
     );
     let store = require_store()?;
     let msgs = ops::list_messages(&store, req).map_err(|e| {
-        log::warn!("[whatsapp_data][rpc] list_messages error: {e}");
-        format!("[whatsapp_data] list_messages failed: {e}")
+        log::warn!("[whatsapp_data][rpc] list_messages error: {e:#}");
+        format!("[whatsapp_data] list_messages failed: {e:#}")
     })?;
     log::debug!("[whatsapp_data][rpc] list_messages ok count={}", msgs.len());
     Ok(RpcOutcome::single_log(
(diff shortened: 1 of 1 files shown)


graycyrus (commented):
## PR #1639 — fix(whatsapp_data): preserve anyhow cause chain in RPC error envelopes

### Walkthrough

This is a focused, low-risk diagnostic improvement. All four RPC handlers in `src/openhuman/whatsapp_data/rpc.rs` were using `{e}` (anyhow's `Display` impl) to render errors, which emits only the outermost context string and silently drops every wrapped cause underneath it. Switching to `{e:#}` (anyhow's \"alternate\" Display) walks the full cause chain and produces a colon-separated string that includes the underlying SQLite error — exactly what Sentry needs to triage failures like OPENHUMAN-TAURI-6B without a local repro. The scope is deliberately narrow: `global.rs` and `schemas.rs` in the same module touch `String`/serde errors that have no cause chain to expose, so they are correctly left alone.

### Changes

| File | Summary |
| --- | --- |
| `src/openhuman/whatsapp_data/rpc.rs` | Replace `{e}` with `{e:#}` in `map_err` closures across all four RPC handlers; add a 7-line comment to `whatsapp_data_ingest` explaining the rationale |

---

### Actionable comments (0)

No issues found. The change is mechanically correct and complete for this file's scope.

---

### Nitpicks (1)

- `src/openhuman/whatsapp_data/rpc.rs:43-51` — The explanatory comment is placed on `whatsapp_data_ingest` but not repeated on the other three handlers. That's reasonable for brevity, but a single one-liner referencing the `ingest` comment (e.g. `// {e:#} — see whatsapp_data_ingest for rationale`) on the remaining three would help future readers who land on `list_chats` first. Entirely optional; the current state is not confusing.

---

### Questions for the author (0)

None.

---

### Outside the diff

While reading the surrounding code I noticed that `require_store()` returns `Result<_, String>` (already a plain string, no anyhow chain). The `{e}` vs `{e:#}` distinction is irrelevant there, confirming the PR's scope is exactly right and nothing was missed.

I also spot-checked `global.rs` and `schemas.rs` — both deal with `String` / serde errors with no `.context()` wrapping, so `{e:#}` would produce identical output to `{e}`. The PR author clearly checked this before constraining the change to `rpc.rs`.

---

### Verified / looks good

- All four `ops::*` call sites (`ingest`, `list_chats`, `list_messages`, `search_messages`) are updated consistently — no handler missed.
- The `log::warn!` and the `format!` string in each closure are updated together, so the full chain appears in both the local log and the Sentry-bound RPC error string. No half-fix.
- PII hygiene is preserved: entry-point `debug!` logs continue to redact account IDs, JIDs, and query strings; only the SQLite error text (database locked / schema / FK violation) is newly exposed, which carries no user PII.
- `{e:#}` is the correct anyhow alternate-display specifier for a flat cause chain. `{e:?}` would include backtraces / Debug noise; `{e}` drops causes. The middle path chosen here is right.
- The coverage gate is already satisfied by `core::observability::tests::anyhow_chain_is_rendered_in_full` — no new test scaffolding required for a format-specifier change.

---

This is a clean, well-motivated fix. The comment on `ingest` succinctly documents *why* the change exists and cites the canonical Sentry issue — future spelunkers will thank the author. Ready to merge as-is.

graycyrus (approved):
LGTM — clean, well-scoped fix. The `{e:#}` change correctly preserves the full anyhow cause chain across all four whatsapp_data RPC handlers, making Sentry events actionable. Scope is correct (global.rs/schemas.rs excluded since they wrap String/serde errors, not anyhow chains). CI all green.

Issue #61, opened 2026-03-29
Based on the new codebase, update config to reflect the new config values we have set and can use


senamakel:
## Goal

Update the configuration to reflect the new config values supported by the current codebase.

## Description

The codebase has changed, but the current config does not fully reflect the values, options, and defaults now supported.

Update the config layer so it matches the latest implementation and can be used reliably across development and deployment.

This includes:

* reviewing the current codebase for supported config values
* updating config files, examples, and defaults
* removing outdated or unused config values
* ensuring the app reads and uses the updated config correctly

## Acceptance Criteria

* Config is updated to match the current codebase
* New supported config values are added
* Outdated/unused config values are removed or deprecated clearly
* Example config files are updated
* App runs successfully using the updated config

PR #1641, opened 2026-05-13
fix(windows): retry-with-backoff for transient FS errors on auth-profiles.lock + .openhuman wipe (#9E, #9C, #4Y, #61, #5Q, #9F, #4M)

oxoxDev:
## Summary

- New shared helpers `retry_with_backoff` / `retry_with_backoff_async` in `src/openhuman/util.rs` that retry on Windows transient FS error codes (5, 32, 33, 1224).
- `AuthProfilesStore::acquire_lock` now retries lock creation 6× with exponential backoff (100ms base).
- `memory::tree::wipe_all_rpc` + `reset_tree_rpc` now retry `.openhuman` tree removal 6× with 200ms base.
- Final-failure path still calls `report_error` — only the transient-retry phase is quieted.

## Problem

- Seven Sentry issues on Windows clients fired from one of two transient-handle races:
  - `Failed to create auth profile lock at C:\Users\…\auth-profiles.lock` (OPENHUMAN-TAURI-9E ×1, -9C ×1, -4Y ×1, -61 ×1; read-path -5Q ×1) — CEF subprocess / AV scanner holds a handle on the parent dir for a sub-second window when openhuman tries to create the lock file. Windows file locking is mandatory, so `create_new` fails with `ERROR_SHARING_VIOLATION` (32) or `ERROR_ACCESS_DENIED` (5) instead of blocking.
  - `Failed to remove C:\Users\…\.openhuman: The process cannot access the file because it is being used by another process. (os error 32)` (OPENHUMAN-TAURI-9F ×1, -4M ×1) — same shape on tree-wipe; CEF profile / Ollama binary handle releases within seconds.
- Hard-failing on the first attempt produced Sentry noise on every cold-start race; the operations succeed deterministically on retry.

## Solution

- `src/openhuman/util.rs`: new `retry_with_backoff` (sync, `std::thread::sleep`) and `retry_with_backoff_async` (tokio). Both honor a shared `is_transient_fs_error` classifier that matches Windows `os_error` codes 5, 32, 33, 1224. On non-Windows, transient classification returns false (POSIX `EACCES` is not transient — surface immediately).
- `src/openhuman/credentials/profiles.rs`: `acquire_lock` wraps `OpenOptions::create_new` with `retry_with_backoff("create auth profile lock", 6, 100, …)`. The existing `AlreadyExists` busy-wait loop is preserved — that's the legitimate "another openhuman PID is signing in" case. Only the *unexpected* error path retries.
- `src/openhuman/memory/tree/read_rpc.rs`: `wipe_all_rpc` + `reset_tree_rpc` separate the SQLite truncation (still `spawn_blocking`) from the filesystem cleanup (now `tokio::fs` wrapped in `retry_with_backoff_async`). Avoids blocking the executor while retries are sleeping.
- Logging: mid-retry `tracing::warn!` with structured tags (`op`, `attempt`, `retry_in_ms`, `error`) keeps the diagnostic locally visible; final-failure surfaces unwrapped so `report_error` still emits the Sentry event when retries are exhausted. First-attempt success is silent.
- Tests: 5 new unit tests in `src/openhuman/util.rs` — immediate success, success-after-retries (sync + async), failure-after-all-attempts, non-transient bails immediately. Cross-platform tests use a `__TEST_TRANSIENT__` error-message marker so the retry logic exercises on POSIX CI.

## Submission Checklist

- [x] Tests added or updated — 5 unit tests in `src/openhuman/util.rs` covering happy path + failure modes + non-transient short-circuit.
- N/A: Diff coverage ≥ 80% — touched lines are mostly retry plumbing; helpers exercised by the new unit tests; call sites are integration paths covered by their existing tests (`cargo test --lib openhuman::credentials` passes 112/0).
- N/A: Coverage matrix updated — behaviour-only retry layer, no new feature row.
- N/A: All affected feature IDs from the matrix — see above.
- N/A: Manual smoke checklist — does not touch release-cut surfaces.
- [x] Linked issues closed via `Closes` in `## Related`.

## Impact

- Stops ~7 events / 3d on the Windows auth-lock and tree-wipe paths.
- Eliminates a class of cold-start failures users hit on Windows when AV scanners / CEF subprocesses hold transient handles. The retry loop is bounded (worst-case ~3-12 seconds before surfacing the real failure), so there's no risk of indefinite hang.
- No POSIX behavior change — `is_transient_fs_error` returns false on non-Windows and the retry short-circuits.

## Related

- Closes OPENHUMAN-TAURI-9E
- Closes OPENHUMAN-TAURI-9C
- Closes OPENHUMAN-TAURI-4Y
- Closes OPENHUMAN-TAURI-61
- Closes OPENHUMAN-TAURI-5Q
- Closes OPENHUMAN-TAURI-9F
- Closes OPENHUMAN-TAURI-4M

---

- Authored by: Google Jules (cloud agent, session `2289277133504848069`)
- Jules session status flipped to `Failed` at the publish stage, but the produced patch was complete and clean. Pulled via `jules remote pull --session`, applied to a fresh branch off `upstream/main` with a single conflict in `src/openhuman/util.rs` (resolved by keeping HEAD's existing truncate tests + appending Jules's new retry tests).
- Cherry-pick + conflict-resolve + 3-commit split done orchestrator-side by Claude.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock-file acquisition with targeted retries for transient errors.
  * Enhanced wipe/reset workflows to tolerate "not found" and transient filesystem failures.

* **Refactor**
  * Moved on-disk cleanup out of blocking tasks into non-blocking async operations to reduce executor blocking and streamline response assembly.

* **Chores / Tests**
  * Added robust retry utilities and unit tests for retry success, eventual failure, and non-retryable errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1641)

### DIFF
--- a/src/openhuman/memory/tree/read_rpc.rs
+++ b/src/openhuman/memory/tree/read_rpc.rs
@@ -1356,7 +1356,7 @@ pub async fn wipe_all_rpc(config: &Config) -> Result<RpcOutcome<WipeAllResponse>
 
     let resp = WipeAllResponse {
         rows_deleted,
-        dirs_removed: dirs_removed.clone(),
+        dirs_removed,
         sync_state_cleared,
     };
 
@@ -1416,17 +1416,19 @@ pub struct ResetTreeResponse {
 /// inert fallback to a real Ollama model) and want to re-summarise
 /// existing data without paying the upstream sync cost again.
 ///
-/// Three steps, each in its own SQL pass:
+/// Three steps, executed in this order:
 ///   1. Truncate `mem_tree_summaries`, `mem_tree_trees`,
 ///      `mem_tree_buffers`, `mem_tree_jobs`. The tree schema is
 ///      derived state — chunks are the source of truth.
-///   2. Remove `<content_root>/wiki/summaries/` on disk so stale
-///      `.md` files don't drift from the SQL truth.
-///   3. Reset every chunk's `lifecycle_status` to
+///   2. Reset every chunk's `lifecycle_status` to
 ///      `'pending_extraction'` and enqueue an `extract_chunk` job
 ///      keyed on the chunk id. The async worker picks each up and
 ///      re-runs entity extract → score → embed → append-to-buffer.
 ///      Seals happen automatically as L0 buffers cross the gate.
+///   3. Remove `<content_root>/wiki/summaries/` on disk so stale
+///      `.md` files don't drift from the SQL truth. Done last (and
+///      outside `spawn_blocking`) so the on-disk removal can use
+///      async retry without blocking the worker thread.
 pub async fn reset_tree_rpc(config: &Config) -> Result<RpcOutcome<ResetTreeResponse>, String> {
     use crate::openhuman::memory::tree::jobs::store as jobs_store;
     use crate::openhuman::memory::tree::jobs::types::{ExtractChunkPayload, NewJob};
@@ -1455,7 +1457,7 @@ pub async fn reset_tree_rpc(config: &Config) -> Result<RpcOutcome<ResetTreeRespo
                 Ok(total)
             })?;
 
-            // Step 3 — flip every chunk back to `pending_extraction` and
+            // Step 2 — flip every chunk back to `pending_extraction` and
             // enqueue an `extract_chunk` job per id.
             let (chunks_requeued, jobs_enqueued) =
                 with_connection(&cfg, |conn| -> anyhow::Result<(u64, u64)> {
@@ -1496,7 +1498,7 @@ pub async fn reset_tree_rpc(config: &Config) -> Result<RpcOutcome<ResetTreeRespo
         .map_err(|e| format!("reset_tree join error: {e}"))?
         .map_err(|e| format!("reset_tree: {e:#}"))?;
 
-    // Step 2 — wipe the on-disk wiki/summaries tree.
+    // Step 3 — wipe the on-disk wiki/summaries tree.
     // Use async retry to avoid blocking the executor during Windows sharing violations.
     let summaries_dir = config
         .memory_tree_content_root()
--- a/src/openhuman/memory/tree/read_rpc.rs
+++ b/src/openhuman/memory/tree/read_rpc.rs
@@ -1254,7 +1254,7 @@ pub struct WipeAllResponse {
 /// can re-sync from scratch without leaving the app.
 pub async fn wipe_all_rpc(config: &Config) -> Result<RpcOutcome<WipeAllResponse>, String> {
     let cfg = config.clone();
-    let resp = tokio::task::spawn_blocking(move || -> Result<WipeAllResponse> {
+    let (rows_deleted, sync_state_cleared) = tokio::task::spawn_blocking(move || -> Result<(u64, u64)> {
         // Tables to truncate. Order matters: `mem_tree_summaries` and
         // `mem_tree_buffers` both have `FOREIGN KEY (tree_id) REFERENCES
         // mem_tree_trees(id)` with `PRAGMA foreign_keys = ON`, so trees
@@ -1283,45 +1283,11 @@ pub async fn wipe_all_rpc(config: &Config) -> Result<RpcOutcome<WipeAllResponse>
             Ok(total)
         })?;
 
-        // Filesystem cleanup. Each directory is best-effort: if one
-        // fails (permission denied, path doesn't exist) we keep going
-        // and report what we managed to remove. `email/` and the
-        // legacy bare `summaries/` are listed for back-compat —
-        // workspaces ingested before the raw-archive + wiki/ moves
-        // still have files there. Fresh installs only ever populate
-        // `raw/`, `wiki/`, `chat/`, and `document/`.
-        const DIRS: &[&str] = &["raw", "wiki", "chat", "document", "email", "summaries"];
-        let content_root = cfg.memory_tree_content_root();
-        let mut dirs_removed: Vec<String> = Vec::new();
-        for dir in DIRS {
-            let path = content_root.join(dir);
-            match std::fs::remove_dir_all(&path) {
-                Ok(()) => dirs_removed.push((*dir).to_string()),
-                Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
-                Err(e) => {
-                    // Logical name (raw / wiki / chat / ...) is enough
-                    // signal — the absolute path embeds the user's
-                    // home directory.
-                    log::warn!(
-                        "[memory_tree::read::wipe] failed to remove dir={} err={e}",
-                        dir
-                    );
-                }
-            }
-        }
-
         // Composio sync-state lives in the unified memory store
         // (`<workspace>/memory/memory.db`). Open it directly and
         // delete every key in the `composio-sync-state` namespace —
         // this clears each provider's `cursor` + `synced_ids` set so
         // the next sync re-fetches from the beginning.
-        //
-        // We do **not** swallow clear failures into `0`: callers (and
-        // the frontend `sync_state_cleared` contract) need to
-        // distinguish "nothing to clear" from "failed to clear, so
-        // the next sync may still be incremental." A missing DB is
-        // legitimately "nothing to clear"; a SQLite error is a
-        // failed wipe and propagates.
         let sync_state_cleared: u64 = {
             let unified_db = cfg.workspace_dir.join("memory").join("memory.db");
             if !unified_db.exists() {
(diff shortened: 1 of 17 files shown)


oxoxDev:
Addressed in 378e6337: both helpers now reject `attempts == 0` cleanly via `anyhow::ensure!` before loop entry (no more `expect()` panic). Added regression tests `test_retry_with_backoff_rejects_zero_attempts` and `test_retry_with_backoff_async_rejects_zero_attempts` — both assert the error message contains "requires attempts > 0" and confirm the closure never runs.

graycyrus (changes requested):
## PR #1641 Review — fix(windows): retry-with-backoff for transient FS errors

### Walkthrough

This PR surgically targets a class of Windows-specific cold-start failures where mandatory file locking by CEF subprocesses, AV scanners, or the Ollama binary causes one-shot filesystem operations to fail with `ERROR_SHARING_VIOLATION` (32), `ERROR_ACCESS_DENIED` (5), `ERROR_LOCK_VIOLATION` (33), or `ERROR_USER_MAPPED_FILE` (1224). The fix adds two generic retry helpers — a sync version (using `std::thread::sleep`) and an async version (using `tokio::time::sleep`) — to `src/openhuman/util.rs`, then wires them into the two affected code paths: lock-file creation in `AuthProfilesStore::acquire_lock` and filesystem tree removal in `wipe_all_rpc` / `reset_tree_rpc`.

The approach is well-scoped, the test coverage is solid, and the helper design is reusable for future FS-touching code.

### Changes

| File | Summary |
|---|---|
| `src/openhuman/util.rs` | Adds `retry_with_backoff` (sync), `retry_with_backoff_async` (tokio), and `is_transient_fs_error` (Windows OS error codes 5/32/33/1224), plus 7 unit tests |
| `src/openhuman/credentials/profiles.rs` | `acquire_lock` wraps `OpenOptions::create_new` with `retry_with_backoff`; `AlreadyExists` detection via anyhow chain inspection |
| `src/openhuman/memory/tree/read_rpc.rs` | `wipe_all_rpc` and `reset_tree_rpc` split SQLite work from FS cleanup, use `retry_with_backoff_async` + `tokio::fs::remove_dir_all` |

### Actionable comments posted inline (4)

1. **[minor]** `util.rs` — tracing macros duplicate every field in the message string
2. **[minor]** `read_rpc.rs:1359` — unnecessary `.clone()` on `dirs_removed`
3. **[minor]** `util.rs` — missing async non-transient short-circuit test
4. **[minor]** `read_rpc.rs` — doc comment step numbering no longer matches code order

### Nitpicks (2)

- `util.rs` — Public functions defined after `#[cfg(test)] mod tests` block. Convention is `#[cfg(test)]` at the bottom. Not a correctness issue but reads oddly top-to-bottom.
- `util.rs:382` — `2u64.pow(i)` is safe at current hardcoded params (`attempts=6`), but a saturating cap (e.g. `sleep_ms.min(30_000)`) would make the helper unconditionally safe for future callers.

### Verified / looks good ✓

- `AlreadyExists` chain walk in `profiles.rs` correctly pierces anyhow wrapping
- `is_transient_fs_error` is `#[cfg(windows)]`-gated — POSIX behavior unchanged
- `cfg!(test)` gate for `__TEST_TRANSIENT__` is dead code in release builds
- Sync `retry_with_backoff` only called from `acquire_lock` (already in `spawn_blocking` context)
- `NotFound` handling in `wipe_all_rpc` correctly treated as no-op
- 7 unit tests cover happy path, retries, exhaustion, non-transient bail, and `attempts == 0` guard

src/openhuman/util.rs
```diff
@@ -221,4 +221,279 @@ mod tests {
...
+                    attempt = i + 1,
+                    max_attempts = attempts,
+                    error = %e,
+                    retry_in_ms = sleep_ms,
+                    "[util] {} failed (attempt {}/{}): {}. Retrying in {}ms...",
+                    op_name,
+                    i + 1,
+                    attempts,
+                    e,
+                    sleep_ms
```
graycyrus: **[minor]** Both `tracing::warn!` and `tracing::info!` pass structured key-value fields (`op`, `attempt`, `error`, `retry_in_ms`) *and* then re-format all those same values into the message string. Structured log consumers (JSON formatter, Sentry, DataDog) will emit each field twice.

```rust
// before — fields duplicated in message
tracing::warn!(
    op = op_name,
    attempt = i + 1,
    max_attempts = attempts,
    error = %e,
    retry_in_ms = sleep_ms,
    "[util] {} failed (attempt {}/{}): {}. Retrying in {}ms...",
    op_name, i + 1, attempts, e, sleep_ms
);

// after — message is a static label; readers query structured fields
tracing::warn!(
    op = op_name,
    attempt = i + 1,
    max_attempts = attempts,
    error = %e,
    retry_in_ms = sleep_ms,
    "[util] transient fs error, will retry"
);
```

Same fix applies to the `tracing::info!` at the success-after-retry path, and to both copies in `retry_with_backoff_async`.

src/openhuman/memory/tree/read_rpc.rs
```diff
@@ -1335,16 +1301,65 @@ pub async fn wipe_all_rpc(config: &Config) -> Result<RpcOutcome<WipeAllResponse>
...
+                        e
+                    );
+                }
+            }
+        }
+    }
+
+    let resp = WipeAllResponse {
+        rows_deleted,
+        dirs_removed: dirs_removed.clone(),
```
graycyrus: **[minor]** `dirs_removed` is a locally-owned `Vec<String>` that's moved into `WipeAllResponse` and never used again. The `.clone()` is unnecessary.

```rust
// before
dirs_removed: dirs_removed.clone(),

// after
dirs_removed,
```

src/openhuman/util.rs
```diff
@@ -221,4 +221,279 @@ mod tests {
...
+        assert_eq!(err.to_string(), "permanent error");
+        assert_eq!(calls, 1);
+    }
+
+    #[test]
+    fn test_retry_with_backoff_rejects_zero_attempts() {
+        let mut calls = 0;
+        let result = retry_with_backoff("zero_sync", 0, 1, || {
+            calls += 1;
+            Ok::<i32, anyhow::Error>(42)
```
graycyrus, line 315: **[minor]** There's `test_retry_with_backoff_bail_on_non_transient` for the sync helper but no equivalent for the async path. The async helper has the same `if !is_transient_fs_error(&e) { return Err(e); }` branch and it should be covered.

```rust
#[tokio::test]
async fn test_retry_with_backoff_async_bail_on_non_transient() {
    use std::sync::atomic::{AtomicU32, Ordering};
    let calls = AtomicU32::new(0);
    let result = retry_with_backoff_async("test_async_non_transient", 3, 1, || async {
        calls.fetch_add(1, Ordering::SeqCst);
        anyhow::bail!("permanent error");
        #[allow(unreachable_code)]
        Ok::<i32, anyhow::Error>(0)
    })
    .await;
    let err = result.unwrap_err();
    assert_eq!(err.to_string(), "permanent error");
    assert_eq!(calls.load(Ordering::SeqCst), 1);
}
```

src/openhuman/memory/tree/read_rpc.rs
```diff
@@ -1417,109 +1432,112 @@ pub async fn reset_tree_rpc(config: &Config) -> Result<RpcOutcome<ResetTreeRespo
...
+                for table in TREE_TABLES {
+                    let n = tx
+                        .execute(&format!("DELETE FROM {table}"), [])
+                        .with_context(|| format!("delete from {table}"))?;
+                    total += n as u64;
                 }
                 tx.commit()?;
-                Ok((chunks_requeued, jobs_enqueued))
+                Ok(total)
             })?;
```
graycyrus, line 1459: **[minor]** The inline doc comments still describe the steps as 1 → 2 → 3 (truncate → remove wiki/summaries → re-enqueue chunks), but the refactored code runs them as 1 → 3 → 2 — the wiki/summaries removal now happens after chunk re-enqueue. The comment says "Step 3 — flip every chunk back…" but this block is now step 2 in execution order. Update the step numbering to reflect the actual new order, or future readers will be confused.

oxoxDev:
graycyrus 4 minors addressed in fc16fd5c:

1. util.rs:394 — dropped duplicate KV-vs-message interpolation in both retry helpers; structured fields are the source.
2. read_rpc.rs:1359 — removed unnecessary dirs_removed.clone().
3. util.rs:299 — added test_retry_with_backoff_async_bail_on_non_transient covering the missing async bail path.
4. read_rpc.rs:1457 — step-order doc comments updated to match refactored 1→3→2 sequence (and inline labels renumbered).

cargo test --lib openhuman::util — passes (23 tests now). fmt + check clean.

oxoxDev:
graycyrus 4 minors addressed in `fc16fd5c`:

1. **util.rs:394** — dropped duplicate KV-vs-message-string interpolation in both `tracing::warn!` and `tracing::info!` macros (sync + async retry helpers). Structured fields are the canonical source; message is now a short literal.
2. **read_rpc.rs:1359** — removed unnecessary `dirs_removed.clone()`; `Vec<String>` is moved straight into `WipeAllResponse`.
3. **util.rs:299** — added `test_retry_with_backoff_async_bail_on_non_transient` covering the missing async bail path (sync sibling `test_retry_with_backoff_bail_on_non_transient` already existed).
4. **read_rpc.rs:1457** — inline step-order doc comments updated to match the refactored 1 → 3 → 2 sequence (truncate → re-enqueue chunks → remove wiki/summaries).

`cargo test --lib openhuman::util` — 23 tests passing (+1 new async-bail). `cargo fmt --check` clean. No new clippy hits on touched files.

senamakel:
**Deferred review notes from pr-manager (post-merge-conflict pass)**

These were flagged during the reviewer agent's pass but deferred for human consideration — they are non-blocking nitpicks and one question for the author.

---

### Nitpicks

**`src/openhuman/util.rs:489`** — `#[cfg(not(windows))]` block with `let _ = io_err` is noisier than it needs to be. The binding only silences an unused-variable warning that only exists on non-Windows builds. A tighter alternative is to use `#[allow(unused_variables)]` scoped to that block, or simply pattern-match the io_err where it's actually used on Windows, so the non-Windows arm disappears entirely.

**`src/openhuman/util.rs:272, 286`** — The `#[allow(unreachable_code)] Ok::<i32,_>(0)` dance in each test closure is needed to give the compiler a type hint after `anyhow::bail!`, but it reads awkward. Consider annotating the closure's return type explicitly (`|| -> anyhow::Result<i32> { ... }`) so the `Ok::<i32,_>(0)` placeholder and the allow attribute can be dropped.

**`src/openhuman/memory/tree/read_rpc.rs:1325`** — `format!("remove dir {}", dir)` allocates a fresh `String` on every retry iteration (up to 6× per dir, 6 dirs). Since `dir` is `&'static str`, a lazy-format or a concatenation with a static prefix would avoid the allocation on the success path. Simplest fix: pass `dir` directly as the `op_name` (accepting that the log says `raw` rather than `remove dir raw`), or concatenate once before the `retry_with_backoff_async` call.

**`src/openhuman/util.rs:361`** — Public items (`retry_with_backoff`, `retry_with_backoff_async`) appear below the `#[cfg(test)]` module in source order. Convention is public API at the top of the module, test module at the bottom. Consider reordering for readability.

---

### Question for author

**`src/openhuman/util.rs` — `is_transient_fs_error` test-bypass scope**

`is_transient_fs_error` is `pub` and contains a `cfg!(test)` branch that treats any error whose message contains `__TEST_TRANSIENT__` as transient. `cfg!(test)` is true for the Rust test harness (`cargo test`) but also for any integration-test binary or fuzz harness built with `--cfg test`. If an external integration harness (e.g. `tests/json_rpc_e2e.rs` compiled as a test binary) ever passes error strings through this function, the bypass would be active.

Should the hook be narrowed with `#[cfg(test)]` at the item level rather than the inline `cfg!(test)` call? That would restrict it to crate-internal test builds only and prevent the bypass from leaking into integration-test contexts that import `openhuman_core` as a dependency.

PR #1643, opened 2026-05-13
test: wait for local model usage flags to load

honor2030:
## Summary
- Wait for the async local AI usage config load before asserting the sub-flag checkbox is enabled.
- This removes an order-dependent failure in the full Vitest suite while keeping the focused LocalModelPanel coverage intact.

## Why
During beta-test preflight on `origin/main`, `pnpm test` failed consistently in the full suite:

```
LocalModelPanel — usage flags > persists a sub-flag toggle once master is enabled
Error: expect(element).not.toBeDisabled()
```

The focused test file passed by itself, which points to timing/order sensitivity rather than a product logic regression. The component intentionally loads usage flags asynchronously through the zero-delay initial load, so the assertion should wait for the loaded state.

## Verification
- `pnpm exec vitest run --config test/vitest.config.ts src/components/settings/panels/__tests__/LocalModelPanel.test.tsx --reporter=default`
- `pnpm test`
- `pnpm format:check`
- `pnpm lint` (passes with existing warnings)
- `pnpm compile`
- pre-push hook: `format:check`, `lint`, `compile`, `rust:check`, `lint:commands-tokens`

## Summary by CodeRabbit

No user-facing changes to document. This pull request contains only internal test improvements with no impact on end-user functionality or features.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1643)

### DIFF
--- a/app/src/components/settings/panels/__tests__/LocalModelPanel.test.tsx
+++ b/app/src/components/settings/panels/__tests__/LocalModelPanel.test.tsx
@@ -208,7 +208,9 @@ describe('LocalModelPanel — usage flags', () => {
     const embeddingsLabel = await screen.findByText('Embeddings');
     const checkbox = embeddingsLabel.closest('label')?.querySelector('input[type="checkbox"]');
     expect(checkbox).toBeTruthy();
-    expect(checkbox as HTMLInputElement).not.toBeDisabled();
+    await waitFor(() => {
+      expect(checkbox as HTMLInputElement).not.toBeDisabled();
+    });
     fireEvent.click(checkbox as HTMLInputElement);
 
     await waitFor(() => {


graycyrus (approved):
## Walkthrough

One-line behavioral fix to a Vitest test that was silently order-dependent. The `'persists a sub-flag toggle once master is enabled'` test asserted that a checkbox is not disabled immediately after `findByText` resolves — but `findByText` only guarantees the element is in the DOM, not that the component's async `openhumanGetConfig` load has finished updating the disabled state. Wrapping the assertion in `waitFor()` makes it poll until the state settles, which is the correct Testing Library pattern already used elsewhere in the same test body.

## Changes

| File | Summary |
| --- | --- |
| `app/src/components/settings/panels/__tests__/LocalModelPanel.test.tsx` | Wrap `not.toBeDisabled()` assertion in `waitFor()` to survive async flag load in full-suite runs |

## Review

✅ **No issues found.** The fix is correct — `waitFor()` is exactly the right tool here. `findByText` resolves when the element appears, but the `disabled` attribute is controlled by a separate async state update, so polling is necessary to avoid a race.

- The assertion being wrapped was the only bare synchronous check between two already-async operations
- `waitFor()` import was already in scope — no new import needed
- No production code touched — zero risk surface
- CI passes (17/17 checks green)

Clean, correct test-reliability fix. Ready to merge.

PR #1645, opened 2026-05-13
fix(normalization): guard function.arguments against malformed JSON and default to {}

YellowSnnowmann:
# Summary

- Updated `normalize_function_arguments` to preserve non-empty JSON strings only when they parse as valid JSON.

- Added a fallback to `{}` for malformed JSON strings in `function.arguments` to prevent upstream request failures.

- Kept existing behavior for:
  - empty-string
  - `null`
  - missing arguments
  - All continue normalizing to `{}`

- Added focused Rust unit tests for:
  - valid JSON strings
  - malformed JSON strings
  - empty/null inputs
  - object serialization

- Improves provider compatibility when model output includes invalid JSON-like argument payloads.

---

# Problem

- Some model/tool-call responses emit malformed JSON in `function.arguments`:
  - `{a:1}`
  - single-quoted JSON-like strings

- Previous normalization behavior could forward malformed strings directly.

- Result:
  - backend `json.loads` / validation failures
  - HTTP `400` responses
  - avoidable tool-call failures
  - noisy downstream error handling

---

# Solution

- Updated `normalize_function_arguments`:
  - String inputs now validated using:
    - `serde_json::from_str`

---

## Decision Logic

- Empty string → `{}`

- Valid JSON string → preserved as-is

- Invalid JSON string → `{}`

- `null` / missing → `{}`

- JSON object values:
  - continue serializing normally

---

# Tradeoffs

- Malformed payload detail is intentionally discarded
- Prioritizes:
  - predictable behavior
  - safe fallback execution
  - reduced downstream failures

---

# Submission Checklist

- Diff coverage ≥ 80% — changed lines (Vitest + cargo-llvm-cov merged via diff-cover) meet the gate enforced by `.github/workflows/coverage.yml`. Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge.

- Coverage matrix updated — added/removed/renamed feature rows in `docs/TEST-COVERAGE-MATRIX.md` reflect this change (or `N/A: behaviour-only change`)

- Linked issue closed via `Closes #NNN` in the `## Related` section

---

# Related

- `OPENHUMAN-TAURI-6F`
  - malformed `function.arguments` normalization path

- `N/A` for `Closes #NNN` until a GitHub issue number is confirmed.

---

# Impact

- Runtime/platform:
  - Rust core behavior change
  - Impacts all desktop clients routing tool calls through this normalization path

- Reliability:
  - Reduces backend `400` failures caused by malformed argument JSON from model/provider outputs

- Security/privacy:
  - No new secret handling or network surface changes

- Compatibility:
  - Backward-compatible
  - Valid JSON behavior preserved
  - Invalid inputs now fail-safe to `{}`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of JSON arguments in provider tool calls so malformed or non-JSON strings are normalized to empty objects, preventing malformed payloads from being preserved.

* **Tests**
  * Added unit tests verifying normalization behavior for valid, malformed, empty, and non-string argument inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1645)


src/openhuman/providers/compatible_parse.rs
```diff
@@ -119,8 +119,10 @@ pub(crate) fn normalize_function_arguments(arguments: Option<serde_json::Value>)
         Some(serde_json::Value::String(raw)) => {
             if raw.trim().is_empty() {
                 "{}".to_string()
-            } else {
+            } else if serde_json::from_str::<serde_json::Value>(&raw).is_ok() {
                 raw
+            } else {
+                "{}".to_string()
```
graycyrus, line 133: [major] **Silent discard of malformed JSON creates a debugging blind spot**

When `serde_json::from_str` rejects the string and the function falls back to `{}`, nothing is recorded. In production, operators have no way to know a provider returned bad JSON, which provider/model it was, or how often it's happening. Given that OPENHUMAN-TAURI-6F was only caught because it produced a visible 400 upstream, there is likely a long tail of silent failures that will now be invisible.

Add a `warn!` log line before the fallback. Since `arguments` can contain user data, truncate rather than dump the full string:

```rust
} else {
    warn!(
        "[providers] normalize_function_arguments: malformed function.arguments (len={}, preview={:?}); substituting {{{{}}}}",
        raw.len(),
        &raw[..raw.len().min(80)]
    );
    "{}".to_string()
}
```

`warn!` is more appropriate than `debug!` here — this represents a deviation from spec on the provider's part, not a routine code path.

src/openhuman/providers/compatible_parse.rs
```diff
@@ -119,8 +119,10 @@ pub(crate) fn normalize_function_arguments(arguments: Option<serde_json::Value>)
         Some(serde_json::Value::String(raw)) => {
             if raw.trim().is_empty() {
                 "{}".to_string()
-            } else {
+            } else if serde_json::from_str::<serde_json::Value>(&raw).is_ok() {
```
graycyrus, line 122: [minor] **Question**: When `serde_json::from_str` parses the string successfully, the parsed value is discarded and `raw` is returned as-is. Is there any intent to validate that the parsed result is specifically a JSON **object** (`Value::Object`) rather than a bare string, number, or array?

A provider emitting `"42"` as `function.arguments` would pass the validity check and get forwarded as `"42"`, which might also cause a `json.loads` failure on the backend if it expects an object. Worth a quick check against what the backend actually accepts.

graycyrus (changes requested):
## Walkthrough

Small, focused defensive fix in the provider normalization layer. When a downstream model returns a tool call whose `function.arguments` field contains syntactically invalid JSON (e.g. `{a:1}`, single-quoted keys, truncated objects), the prior code would pass the raw string through unchanged, causing the backend to blow up with a 400 on `json.loads`. The fix adds a parse probe via `serde_json::from_str` and substitutes `{}` when the probe fails. Four tests cover the new branch plus pre-existing behavior. CI is green, coverage gate passes, and there are no API surface changes.

The fix is correct and the test coverage is solid. There is one meaningful operational gap worth addressing before merge.

## Changes

| File | Summary |
|---|---|
| `src/openhuman/providers/compatible_parse.rs` | Added `serde_json::from_str` validity check in `normalize_function_arguments`; malformed strings now fall back to `"{}"` instead of being forwarded raw |
| `src/openhuman/providers/compatible_tests.rs` | Four new unit tests covering valid JSON preserved, malformed JSON fallback (four inputs), whitespace/null/None, and `Value::Object` serialization |

## Verified / looks good

- The `is_ok()` probe is the right tool — no double-parse overhead on the happy path.
- All four test inputs are distinct failure modes — good representative spread.
- Existing behavior for `Value::Object` and `Value::Null` is preserved and explicitly re-tested.
- No new `unwrap()` calls. Function signature unchanged; all callers unaffected.

**See inline comments below for one actionable item and one question.**

senamakel:
**Deferred question for author (from pr-manager review pass):**

In `src/openhuman/providers/compatible_parse.rs`, `normalize_function_arguments` silently substitutes `{}` when a `Value::String` contains malformed JSON. This PR adds a `log::warn!` so the discard is at least traceable, which is good.

The open question: **is silent fallback to `{}` always the right behaviour, or should some callers surface an error?**

Concrete cases to consider:
- A streaming delta that hasn't finished assembling yet — `{}` fallback is correct (the string is temporarily incomplete).
- A model response that shipped a permanently garbled `function.arguments` field — `{}` silently drops real intent; a caller that can propagate an error back to the agent loop might give a better user experience than silently calling the tool with no arguments.

Currently both paths (streaming accumulator and the `parse_provider_tool_call_from_value` pre-deserialized path) collapse to silent `{}`. If there are callers that have enough context to distinguish "this is a final, not-a-partial response" from "this is a streaming fragment", returning `Err(...)` on the final path and only falling back on the streaming path might be worth considering.

No action required if the current blanket fallback is intentional — just flagging for author awareness.

PR #1666, opened 2026-05-13
fix(agent): refresh delegate_{toolkit} tools on first turn for web channel

### LINKED ISSUE #1659: Orchestrator's delegate_{toolkit} contract from PR #1314 broken on web channel — model emits delegate_gmail, runtime resolves to delegate_tools_agent
## Summary

The orchestrator's `delegate_{toolkit}` skill-delegation tools (e.g. `delegate_gmail`, `delegate_notion`) are advertised in the system prompt's `## Connected Integrations` section but are **not actually registered in the function-calling schema** on the web (Tauri desktop) channel. The model reads the prompt, believes `delegate_gmail` exists, then emits a tool call that falls through to the closest available name — `delegate_tools_agent` — which correctly refuses ("requires `integrations_agent`") and the user-facing turn fails with "Gmail is connected, but I don't actually have the Gmail delegation tool available."

### What happened vs expected

**Reproduces every time**, current `main` (commit `7beed3a4`, v0.53.38), Windows desktop dev build via `pnpm dev:app:win`. Gmail is connected via Composio (`composio_list_connections` returns `gmail` with `status: ACTIVE`).

Steps:

1. Sign in, ensure Gmail is connected under Settings → Connections.
2. Open a new thread, send: *"summarize my last 5 unread gmail threads"* (any Gmail-flavored prompt works).
3. Observe orchestrator behaviour in console logs and transcripts at `~/.openhuman/users/<uid>/workspace/session_raw/`.

**Expected:** orchestrator calls `delegate_gmail`, which dispatches `integrations_agent` with `skill_filter = "gmail"`, fetches emails, replies.

**Actual:** orchestrator's reasoning explicitly mentions *"I have a `delegate_gmail` tool available for Gmail integration"*, then emits a `delegate_tools_agent` tool call. `tools_agent` correctly rejects ("I don't have direct access to Gmail tools — the Gmail integration is available via Composio, but that requires the `integrations_agent`"). After up to 10 iterations the turn either gives up or the sub-agent exceeds max iterations.

### Root cause

The orchestrator's tool surface is built by the synchronous agent builder (`src/openhuman/agent/harness/session/builder.rs:883` and `:904`), which intentionally passes `&[]` for `connected_integrations`:

```rust
let synthed = tools::orchestrator_tools::collect_orchestrator_tools(def, reg, &[]);
```

with comment:

> This builder is synchronous and sits on the CLI / REPL / Tauri-web code path. It does not have access to the async Composio fetcher, so we pass an empty slice of connected integrations here — the skill-wildcard expansion therefore produces zero delegation tools. That is correct behaviour: callers that need live integration expansion go through the bus-based `channels::runtime::dispatch` path instead.

`collect_orchestrator_tools` then iterates the empty list and emits zero `SkillDelegationTool` entries. The startup log confirms this:

```
[orchestrator_tools] assembled 6 delegation tool(s) for agent 'orchestrator' (0 integrations connected)
```

The orchestrator therefore only has the six archetype-delegation tools (`research`, `plan`, `run_code`, `delegate_tools_agent`, `review_code`, `archive_session`) plus its direct tools. `delegate_gmail` is never callable.

Meanwhile, `Session::fetch_connected_integrations` (`src/openhuman/agent/harness/session/turn.rs:84` / `:1239`) **does** fetch integrations asynchronously before the first turn, and populates `self.connected_integrations`. That populated list feeds `build_system_prompt` (`turn.rs:1274`), which is why the `## Connected Integrations` section renders `delegate_gmail` correctly in the prompt body. So the model reads the prompt, sees the tool, can't call it.

The architecturally-correct path — `channels/runtime/dispatch.rs::resolve_target_agent` — does fetch integrations first and produces a non-empty `extra_tools` (`dispatch.rs:446-489`). That `extra_tools` flows through `AgentTurnRequest` → bus dispatch → `run_tool_call_loop`. But the web channel **does not** route through the bus for chat turns. `web.rs:1199-1201` constructs the agent directly:

```rust
let agent_result = match reflection_chunks {
    Some(chunks) if !chunks.is_empty() =>
        Agent::from_config_for_agent_with_reflection_chunks(&effective, target_agent_id, chunks),
    _ => Agent::from_config_for_agent(&effective, target_agent_id),
};
```

which lands back in the synchronous builder with `&[]` integrations.

### When the regression actually landed (PR #1314)

The synchronous builder calling `collect_orchestrator_tools(def, reg, &[])` was introduced in PR #544 (April 14, 2026). **That was not the bug.** The empty list was harmless because the orchestrator never actually depended on skill-delegation tools being synthesized — it routed Composio work through `spawn_subagent(integrations_agent, toolkit=…)` instead, and `spawn_subagent` was a real tool in the orchestrator's `named` allowed_tools.

**PR #1314 ("perf(agent): orchestrator harness efficiency improvements", merged 2026-05-09) is the actual regression.** It made three changes to the integrations path that, combined, switched the orchestrator from a working contract to a broken one:

1. **Removed `spawn_subagent` from the orchestrator's `named` allowed_tools.** PR body: *"Remove `schedule` + `spawn_subagent` from orchestrator: `cron_add/cron_list/cron_remove` are the canonical scheduling surface; `delegate_*` + `spawn_worker_thread` replace `spawn_subagent`."*
2. **Rewrote the integrations prompt block from `## Delegation Guide — Integrations` to `## Connected Integrations`.** Before #1314, the block told the model:
   ```
   ## Delegation Guide — Integrations
   …
   - **gmail** — Send and read email via Gmail.
     Delegate with: `spawn_subagent(agent_id="integrations_agent", toolkit="gmail", prompt=<task>)`
   ```
   After #1314 (see `src/openhuman/agent/agents/orchestrator/prompt.rs` diff in `8d9a1ca7`):
   ```
   ## Connected Integrations

   Delegate tasks for these services using the matching `delegate_{toolkit}` tool:

   - **gmail** (delegate via `delegate_gmail`): Send and read email via Gmail.
   ```
3. The new prompt block builds tool names with `sanitise_slug(&ci.toolkit)` so they match what `collect_orchestrator_tools` would synthesise — **on a path that synthesises them.** On the sync-builder path used by the web channel (`builder.rs:883` / `:904`), `collect_orchestrator_tools` is called with `&[]`, so none of those `delegate_<toolkit>` tools actually exist in the function-calling schema.

End result: the orchestrator's prompt advertises tool names that the runtime never registered, *and* the previous escape hatch (`spawn_subagent`) was removed from the allowlist in the same PR. The model emits a `delegate_gmail`-shaped call, the runtime resolves it to the nearest registered name — `delegate_tools_agent` — which correctly refuses, and the turn fails.

PR #1314's intent was right — the new `delegate_*` per-toolkit surface is cleaner than the generic `spawn_subagent(toolkit=…)` pattern, and the test it added (`unconnected_integrations_are_skipped`) is good defensive logic. The miss was that the new contract was only validated against the bus-dispatch path (which **does** call `collect_orchestrator_tools` with a fetched integrations list), not against the sync-builder path the web channel actually uses (`builder.rs:883`).

PR #1631 (worker-thread disable, May 13) is **not** a contributor — it removed `SpawnWorkerThreadTool` registration but did not touch `collect_orchestrator_tools`'s integration handling or the builder.

### Fix shape implications

Reverting only the prompt block (steps 2 and 3 above) would temporarily un-break the user-facing flow if `spawn_subagent` were re-added to the orchestrator's allowed_tools — i.e. roll back the integrations slice of #1314 while keeping its other perf wins. That's the smallest, most reversible palliative.

The real fix is still to close the sync-builder ↔ async-integration gap (one of the three options below) so the new `delegate_<toolkit>` contract from #1314 actually holds on every code path.

### Why sharpening tool descriptions did not help

Prepending a hard negative to `delegate_tools_agent`'s description ("DO NOT USE for Gmail, Notion, Slack, GitHub, Calendar, Drive…") and reordering `collect_orchestrator_tools` so skill tools precede archetype tools both fail to fix this. The model still picks `delegate_tools_agent` because **`delegate_gmail` is not in the function-calling schema at all** — the prompt mentions it, but there is no schema entry the model can route to. The model picks the closest available match.

## Solution (optional)

Three viable directions, ranked by invasiveness:

### A — Block-on in the synchronous builder (smallest)

Add `tokio::task::block_in_place(|| Handle::current().block_on(fetch_connected_integrations(config)))` inside `from_config_for_agent`, pass the result to both `collect_orchestrator_tools` call sites. ~10 lines, isolated to `builder.rs`. Trade-off: blocks the calling thread for up to a 3-second Composio fetch timeout during agent construction; requires the caller to be inside a multi-thread Tokio runtime (current Tauri setup satisfies this, but worth a guard / fallback).

### B — Async post-build tool refresh

Add a `refresh_delegation_tools()` method on `Agent` (or `Session`), call it immediately after `fetch_connected_integrations` in `turn.rs:84`. Use `Arc::get_mut` on `self.tools` / `self.tool_specs` / `self.visible_tool_specs` to append the newly-synthesized skill-delegation tools while the Arc is still uniquely owned (true on first turn, before any sub-agent has cloned). ~40 lines. Trade-off: relies on un-shared Arc state at the refresh point; needs a fallback log + path when the Arc has already been cloned.

### C — Re-shape `delegate_*` skill tools as registry tools (cleanest, biggest)

Replace the runtime-synthesized `SkillDelegationTool` per toolkit with either:

- A single generic `delegate_integration { toolkit: string, prompt: string }` tool registered statically in the global registry. The orchestrator's `## Connected Integrations` prompt section already enumerates valid `toolkit` values. Sync builder works trivially; the model maps "Gmail" → `toolkit: "gmail"` from the prompt.
- Or static per-toolkit `delegate_<toolkit>` registrations for a hardcoded supported set (gmail, notion, slack, github, calendar, drive, …), gated at `execute()` time by a "connected?" check.

Either avoids the sync-vs-async split entirely.

### Recommendation

**A** unblocks the immediate user-facing breakage with the smallest diff and matches the existing architectural intent (skill tools should reflect connected state). **C** is the long-term direction once we agree on the orchestrator's tool-surface shape.

## Acceptance criteria

- [ ] **Repro gone** — Gmail-flavored prompt on the web channel routes to `delegate_gmail` (verifiable in `session_raw/*.jsonl` via `"name":"delegate_gmail"` in the orchestrator's `tool_call` payload), and the `[orchestrator_tools] assembled N delegation tool(s) for agent 'orchestrator' (M integrations connected)` log line shows `M > 0` when integrations are connected.
- [ ] **Regression safety** — Add a session-level test that the web/Tauri agent-construction path actually flows real integrations into `collect_orchestrator_tools`. The existing `collects_agentid_entries_and_expands_skills_wildcard` unit test covers the helper but does not cover the call-site that passes `&[]` today.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by `.github/workflows/coverage.yml`).
- [ ] **Prompt ↔ schema parity invariant** — assert (or comment-as-invariant) that every toolkit listed in the rendered `## Connected Integrations` block has a corresponding tool entry in the agent's `tool_specs` for the same turn.

## Related

- Builder `&[]` introduced in PR #544 (April 14, 2026).
- Bus dispatch path that **does** thread integrations correctly: `src/openhuman/channels/runtime/dispatch.rs::resolve_target_agent` (lines 378-490).
- `tools_agent`'s SaaS-exclusion wording: `src/openhuman/agent/agents/tools_agent/agent.toml` (last touched in #1025).
- Evidence (transcript + logs):
  - System prompt's `## Connected Integrations` block renders `- **gmail** (delegate via \`delegate_gmail\`): Send, read, draft, reply, forward, and search emails…`
  - Same session: `[orchestrator_tools] assembled 6 delegation tool(s) for agent 'orchestrator' (0 integrations connected)`
  - Same session: `delegate_[a-z_]+` names registered = `[delegate_archivist, delegate_critic, delegate_plan, delegate_researcher, delegate_run_code, delegate_tools_agent]` — no `delegate_gmail`.
  - Model reasoning: *"I have a `delegate_gmail` tool available for Gmail integration. Let me first check if Gmail is connected by listing Composio integrations."* → then emits `delegate_tools_agent`.
- **PR #1314 ("perf(agent): orchestrator harness efficiency improvements", merged 2026-05-09)** — the actual regression. Removed `spawn_subagent` from the orchestrator's `named` allowed_tools and rewrote the integrations prompt block to advertise `delegate_<toolkit>` tools that the sync-builder web-channel path never registers. See *"When the regression actually landed (PR #1314)"* above.
- PR #1631 (worker-thread disable, May 13) — proximity only; did not touch `collect_orchestrator_tools`'s integration handling or the builder.
- PR #544 (April 14, 2026) — introduced the `&[]` pattern in `builder.rs:883` / `:904`. Latent gap, not the bug; harmless until PR #1314 built a contract on top of it that requires those tools to exist.

🤖 Filed via Claude Code after live triage; debug branch + transcript evidence available on request.

graycyrus:
## Summary

- Adds `refresh_delegation_tools()` on `Agent`, called after `fetch_connected_integrations()` on the first turn — re-synthesises skill-delegation tools (`delegate_gmail`, `delegate_notion`, …) with the live integrations list and merges them into the tool schema before the system prompt is built
- Adds `base_definition_id` field to `Agent` to preserve the original definition ID through the web channel's `set_agent_definition_name()` mangling (`"orchestrator"` → `"orchestrator_thread-6ad..."`)
- Updates the `&[]` comment at the `collect_orchestrator_tools` call sites to document the refresh contract

## Root cause

PR #1314 removed `spawn_subagent` from the orchestrator's allowed tools and rewrote the integrations prompt to advertise `delegate_{toolkit}` tools. The synchronous builder (`builder.rs`) passes `&[]` for connected integrations (can't call async Composio API), so zero skill-delegation tools are registered. Meanwhile `fetch_connected_integrations()` populates the prompt correctly — the model sees `delegate_gmail` in prose but can't call it, falling through to `delegate_tools_agent` which refuses.

## Changes

| File | Change |
|------|--------|
| `types.rs` | `base_definition_id` field on `Agent` + `AgentBuilder` |
| `builder.rs` | Wire `base_definition_id`, update `&[]` comments |
| `turn.rs` | `refresh_delegation_tools()` method + call on first turn |
| `runtime.rs` | `base_definition_id()` accessor |
| `tests.rs` | 3 tests: populates tools, noop when empty, no duplicates |

## Test plan

- [x] `refresh_delegation_tools_populates_skill_tools` — connected gmail + github → `delegate_gmail` and `delegate_github` appear in tool schema
- [x] `refresh_delegation_tools_noop_when_no_integrations` — empty integrations = no-op
- [x] `refresh_delegation_tools_skips_already_registered` — calling twice doesn't duplicate tools
- [x] `cargo check` (core + Tauri shell) passes
- [x] `cargo test` — 6566 passed, 1 pre-existing failure (unrelated config test)
- [x] `cargo fmt --check` passes
- [x] `pnpm typecheck` / `pnpm lint` / `pnpm format:check` pass

Closes #1659

### DIFF
--- a/src/openhuman/agent/harness/session/builder.rs
+++ b/src/openhuman/agent/harness/session/builder.rs
@@ -48,6 +48,7 @@ impl AgentBuilder {
             event_session_id: None,
             event_channel: None,
             agent_definition_name: None,
+            base_definition_id: None,
             session_parent_prefix: None,
             omit_profile: None,
             omit_memory_md: None,
@@ -233,6 +234,20 @@ impl AgentBuilder {
         self
     }
 
+    /// Sets the original agent definition id as it appears in the global
+    /// [`AgentDefinitionRegistry`]. This is distinct from
+    /// [`Self::agent_definition_name`] which may later be overwritten
+    /// to a thread-scoped variant (e.g. `"orchestrator_thread-6ad6d"`).
+    /// When not set, [`AgentBuilder::build`] falls back to
+    /// `agent_definition_name` and ultimately to `"main"`.
+    ///
+    /// Used by [`Agent::refresh_delegation_tools`] to look up the
+    /// original archetype definition for delegation-tool re-synthesis.
+    pub fn base_definition_id(mut self, id: impl Into<String>) -> Self {
+        self.base_definition_id = Some(id.into());
+        self
+    }
+
     /// Set the parent session-key chain for a sub-agent. Passing
     /// `Some("1713000000_orchestrator")` produces a sub-agent whose
     /// transcript filename is prefixed with the parent's session key,
@@ -379,6 +394,10 @@ impl AgentBuilder {
             agent_definition_name: self
                 .agent_definition_name
                 .clone()
+                .unwrap_or_else(|| "main".to_string()),
+            base_definition_id: self
+                .base_definition_id
+                .or_else(|| self.agent_definition_name.clone())
                 .unwrap_or_else(|| "main".to_string()),
             session_transcript_path: None,
             session_key: {
@@ -869,9 +888,17 @@ impl Agent {
         // Tauri-web code path. It does not have access to the async
         // Composio fetcher, so we pass an empty slice of connected
         // integrations here — the skill-wildcard expansion therefore
-        // produces zero delegation tools. That is correct behaviour:
-        // callers that need live integration expansion go through the
-        // bus-based `channels::runtime::dispatch` path instead.
+        // produces zero `delegate_{toolkit}` tools at construction time.
+        //
+        // The gap is closed at runtime: on the first turn,
+        // `fetch_connected_integrations()` populates
+        // `self.connected_integrations`, and the immediately-following
+        // `refresh_delegation_tools()` call re-runs
+        // `collect_orchestrator_tools` with the live integrations list
+        // and merges any newly-synthesised `delegate_*` tools into the
+        // agent's `tools`, `tool_specs`, `visible_tool_names`, and
+        // `visible_tool_specs` surfaces so the LLM's function-calling
+        // schema is consistent with the system prompt.
         let (delegation_tools, filter_from_scope): (
             Vec<Box<dyn Tool>>,
             Option<std::collections::HashSet<String>>,
@@ -1173,6 +1200,11 @@ impl Agent {
             .post_turn_hooks(post_turn_hooks)
             .learning_enabled(config.learning.enabled)
             .agent_definition_name(agent_id.to_string())
+            // Freeze the original registry id so `refresh_delegation_tools`
+            // can look up the archetype definition even after
+            // `set_agent_definition_name` has mangled the display name to a
+            // thread-scoped variant (e.g. `"orchestrator_thread-6ad6d"`).
+            .base_definition_id(agent_id.to_string())
             .omit_profile(effective_omit_profile)
             .omit_memory_md(effective_omit_memory_md);
         if let Some(ps) = payload_summarizer {
--- a/src/openhuman/agent/harness/session/runtime.rs
+++ b/src/openhuman/agent/harness/session/runtime.rs
@@ -51,6 +51,18 @@ impl Agent {
         &self.agent_definition_name
     }
 
+    /// The original agent definition id from the global registry, frozen
+    /// at build time (e.g. `"orchestrator"`). Unlike
+    /// [`Agent::agent_definition_name`] this is never overwritten by
+    /// [`Agent::set_agent_definition_name`], so it always points to
+    /// the archetype that was looked up in the registry when the
+    /// session was constructed. Used by
+    /// [`Agent::refresh_delegation_tools`] to re-resolve the definition
+    /// after the display name has been mangled.
+    pub fn base_definition_id(&self) -> &str {
+        &self.base_definition_id
+    }
+
     /// Returns a new `AgentBuilder`.
     pub fn builder() -> AgentBuilder {
         AgentBuilder::new()
--- a/src/openhuman/agent/harness/session/tests.rs
+++ b/src/openhuman/agent/harness/session/tests.rs
@@ -694,6 +694,182 @@ fn seed_resume_from_messages_is_noop_on_warm_agent() {
     assert_eq!(cached[0].content, "warm prefix");
 }
 
+// ─────────────────────────────────────────────────────────────────────────────
+// refresh_delegation_tools tests
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Build a minimal agent scoped to `agent_id` with the global builtin registry
+/// initialised. Helper shared by the `refresh_delegation_tools` tests below.
+fn build_agent_with_base_definition(agent_id: &str) -> Agent {
+    use crate::openhuman::agent::harness::AgentDefinitionRegistry;
+    // Idempotent — other tests may have already initialised the registry.
+    AgentDefinitionRegistry::init_global_builtins().unwrap();
+
+    let workspace = tempfile::TempDir::new().expect("temp workspace");
+    let workspace_path = workspace.path().to_path_buf();
+    let provider = Box::new(MockProvider {
+        responses: parking_lot::Mutex::new(vec![]),
+    });
+    let memory_cfg = crate::openhuman::config::MemoryConfig {
+        backend: "none".into(),
+        ..crate::openhuman::config::MemoryConfig::default()
+    };
+    let mem: Arc<dyn crate::openhuman::memory::Memory> =
+        Arc::from(crate::openhuman::memory::create_memory(&memory_cfg, &workspace_path).unwrap());
+    Agent::builder()
+        .provider(provider)
+        .tools(vec![Box::new(MockTool)])
+        .memory(mem)
+        .tool_dispatcher(Box::new(NativeToolDispatcher))
+        .workspace_dir(workspace_path)
+        .agent_definition_name(agent_id.to_string())
+        .base_definition_id(agent_id.to_string())
+        .build()
+        .expect("agent build should succeed")
+}
+
+/// `refresh_delegation_tools` must add `delegate_{toolkit}` tools to the
+/// agent's tool surface when `connected_integrations` is non-empty and the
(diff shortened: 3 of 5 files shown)


senamakel:
this is kidna fixed now in #1670

PR #1734, opened 2026-05-14
chore(migrations): phase out PROFILE.md from disk on schema_version=1

sanil-23:
## Summary

- Adds a one-shot startup migration that retires the legacy `PROFILE.md` injection across all on-disk artifacts.
- Introduces `Config::schema_version` (default 0) as the gate; bumped to 1 after the migration completes.
- Wires the runner into `Config::load_or_init` so it fires before any agent turn can resume a session.

## Problem

Earlier builds wrote `PROFILE.md` into the workspace and rendered `### PROFILE.md` into the system prompt for any agent with `omit_profile = false` (orchestrator, welcome, trigger_*, help). Once injected, the runtime loader (`agent/harness/session/turn.rs:1295 try_load_session_transcript`) replays those bytes verbatim on every resume for KV-cache stability. A renderer-only fix leaves existing JSONL transcripts and their `.md` companions replaying the leaked content forever. The producer side (`learning/profile_md_renderer.rs`, `composio/providers/traits.rs:115 merge_provider_into_profile_md`) is being retired separately; this PR handles every consumer surface so the leak cannot replay even if the file briefly resurfaces.

## Solution

New `src/openhuman/migrations/` module (sibling to the existing user-triggered `migration/` OpenClaw import) with a runner gated by `schema_version`:

1. Delete `{workspace}/PROFILE.md` — `inject_workspace_file_capped` skips silently when the file is missing, so the renderer becomes a no-op for the 5 `omit_profile = false` agents without touching their `agent.toml`.
2. Walk every `session_raw/**/*.jsonl` (flat + legacy `DDMMYYYY/`), read via `transcript::read_transcript`, strip `### PROFILE.md` from the first system message via `strip_profile_md_block`, rewrite via `transcript::write_transcript` (also re-renders the companion `.md`).
3. Walk every `sessions/**/*.md` and alter in place — same stripper, with a broader boundary set (`### `, `---`, `## [`, `<!--/MSG-->`) so it cleanly bounds both the new JSONL-companion format and the legacy HTML-comment format. The rest of each transcript (all messages, all metadata) is preserved.

Hook lives in `Config::load_or_init` (`config/schema/load.rs`) at the end of the existing-config and first-launch branches; the pre-login in-memory branch is deliberately skipped. Idempotent in two ways: the version gate is a fast check on every subsequent call, and each step is self-idempotent (file already gone → no-op, transcript already clean → `skipped`, .md already clean → byte-identical). Best-effort: any per-file failure folds into `stats.errors` and the walk continues; runner-level failures log and never block startup.

Live verified against my real workspace (24 JSONLs, 100 .md companions): `scanned=24 cleaned=1 skipped=23 errors=0 profile_md_removed=true md_companions_altered=1`, second launch instant no-op. Tests in tempdirs cover the stripper for new-format `.md` (`---` boundary), legacy HTML-comment `.md` (`<!--/MSG-->` boundary), JSONL with next `### ` heading, JSONL terminating at EOF, JSONL with the truncation footer, plus the version-gate idempotency on `run_pending`.

## Submission Checklist

- [x] Tests added or updated (happy path + failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 22 new unit tests covering stripper boundaries, JSONL walk, .md sweep (both formats), file deletion, version-gate, fresh-install short-circuit, idempotency.
- [x] **Diff coverage ≥ 80%** — diff is Rust-only inside the new `migrations/` module; every public path has direct tests via `cargo test openhuman::migrations::` (15/15 lib tests + 7 stripper tests pass).
- [x] N/A: behaviour-only change — Coverage matrix updated. The migration changes on-disk state during startup, not user-facing UI/RPC behavior; `TEST-COVERAGE-MATRIX.md` rows track product features, not internal startup migrations.
- [x] N/A: no matrix rows — All affected feature IDs from the matrix are listed under `## Related`.
- [x] N/A: no release-cut surface change — Manual smoke checklist updated.
- [x] N/A: no linked issue — Linked issue closed via `Closes #NNN`.

## Impact

- **Runtime/platform**: desktop (only surface that runs `openhuman-core serve`). Fires once per workspace on first launch of the new build.
- **Performance**: walks all transcripts once on the first post-update launch (typically <100 files; my workspace: 24 JSONL + 100 .md cleaned in ~5ms). Every subsequent launch hits the `schema_version >= 1` early-return and exits the runner in microseconds.
- **Migration**: one-way (`schema_version: 0 → 1`). No rollback path; if a user re-installs an older build, the field is silently ignored (no `deny_unknown_fields`).
- **Compatibility**: `config.toml` files written by older builds load unchanged — `#[serde(default)]` defaults missing `schema_version` to 0 so they pick up the migration on the next launch.
- **Producer side untouched**: `merge_provider_into_profile_md` (`composio/providers/traits.rs:115`) and `ProfileMdRenderer` (`learning/profile_md_renderer.rs`) still exist. If they ever fire, the file reappears and the 5 `omit_profile = false` agents will re-inject. That cleanup is the explicit v2 follow-up.

## Related

- Closes:
- Follow-up PR(s)/TODOs:
  - v2: disable or remove the PROFILE.md writer paths (`merge_provider_into_profile_md`, `ProfileMdRenderer`) and flip the 5 `omit_profile = false` flags in welcome / trigger_triage / trigger_reactor / orchestrator / help `agent.toml` files.
  - Remove this migration module in a later release once active workspaces have rolled forward past schema_version=1.

---

### Linear Issue
- [x] N/A: no Linear issue — Key
- [x] N/A: no Linear issue — URL

### Commit & Branch
- Branch: `chore/phase-out-profile-md`
- Commit SHA: `09265034313af9f42dd27c96a6af9aff62e90f3b`

### Validation Run
- [x] N/A: no app/ changes — `pnpm --filter openhuman-app format:check`
- [x] N/A: no app/ changes — `pnpm typecheck`
- [x] Focused tests: `cargo test --lib openhuman::migrations::` → 22 passed; 0 failed
- [x] Rust fmt/check (if changed): `cargo fmt --check` (clean) + `cargo check` (only pre-existing warnings)
- [x] N/A: no Tauri-shell changes — Tauri fmt/check (if changed)

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: one-time cleanup of legacy PROFILE.md state on disk; future agent system prompts are produced from a workspace with no `PROFILE.md` and no tainted persisted transcripts.
- User-visible effect: none directly. The first launch on the new build silently strips any leaked content from past sessions; conversations and transcript proofs are preserved.

### Parity Contract
- Legacy behavior preserved: yes for everything outside the `### PROFILE.md` block; the renderer code path is unchanged, only the file it would read is gone.
- Guard/fallback/dispatch parity checks: `load_or_init` still falls back to `Default::default()` on config-load failure (line 798); migration simply doesn't run that launch and retries on the next.

## Summary by CodeRabbit

* **New Features**
  * Introduced automatic workspace data migrations that run on startup to keep your data schema up-to-date.
  * Removed legacy PROFILE.md content from session transcripts and related files.

* **Tests**
  * Added comprehensive test coverage for the migration system.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1734)

### DIFF
--- a/src/openhuman/config/schema/load.rs
+++ b/src/openhuman/config/schema/load.rs
@@ -738,6 +738,7 @@ impl Config {
                 recovered = config_was_corrupted,
                 "Config loaded"
             );
+            crate::openhuman::migrations::run_pending(&mut config).await;
             Ok(config)
         } else {
             let mut config = Config {
@@ -762,6 +763,7 @@ impl Config {
                 initialized = true,
                 "Config loaded"
             );
+            crate::openhuman::migrations::run_pending(&mut config).await;
             Ok(config)
         }
     }
--- a/src/openhuman/config/schema/load.rs
+++ b/src/openhuman/config/schema/load.rs
@@ -741,9 +741,16 @@ impl Config {
             crate::openhuman::migrations::run_pending(&mut config).await;
             Ok(config)
         } else {
+            // Fresh install: there is no legacy on-disk state, so stamp
+            // the workspace at the current schema version up front. This
+            // makes `run_pending` a fast no-op on the first launch
+            // (nothing to migrate) and keeps the "first launch on this
+            // workspace" semantics aligned with "current binary built
+            // this workspace".
             let mut config = Config {
                 config_path: config_path.clone(),
                 workspace_dir,
+                schema_version: crate::openhuman::migrations::CURRENT_SCHEMA_VERSION,
                 ..Default::default()
             };
             config.save().await?;
@@ -763,6 +770,12 @@ impl Config {
                 initialized = true,
                 "Config loaded"
             );
+            // Defensive: still call run_pending. It will see
+            // `schema_version == CURRENT` and return immediately, but
+            // the call site stays symmetric with the existing-config
+            // branch so a future migration that needs to fire on fresh
+            // installs (vanishingly unlikely, but possible) doesn't
+            // require touching this path.
             crate::openhuman::migrations::run_pending(&mut config).await;
             Ok(config)
         }
--- a/src/openhuman/config/schema/types.rs
+++ b/src/openhuman/config/schema/types.rs
@@ -28,6 +28,13 @@ pub struct Config {
     pub workspace_dir: PathBuf,
     #[serde(skip)]
     pub config_path: PathBuf,
+    /// Workspace data-schema version. Bumped each time a one-shot data
+    /// migration under [`crate::openhuman::migrations`] runs successfully.
+    /// `#[serde(default)]` so existing `config.toml` files (which predate
+    /// the field) load as version `0` and pick up pending migrations on
+    /// the first launch of the new build.
+    #[serde(default)]
+    pub schema_version: u32,
     pub api_url: Option<String>,
     pub api_key: Option<String>,
     /// Custom LLM inference endpoint (OpenAI-compatible). When set together
@@ -272,6 +279,7 @@ impl Default for Config {
         Self {
             workspace_dir: openhuman_dir.join("workspace"),
             config_path: openhuman_dir.join("config.toml"),
+            schema_version: 0,
             api_url: None,
             api_key: None,
             inference_url: None,
--- a/src/openhuman/migrations/mod.rs
+++ b/src/openhuman/migrations/mod.rs
@@ -1,0 +1,88 @@
+//! Startup data migrations gated by [`Config::schema_version`].
+//!
+//! Each migration is a one-shot, idempotent transformation of on-disk
+//! data. The runner is invoked from [`Config::load_or_init`] and is a
+//! fast no-op for workspaces whose `schema_version` already matches
+//! [`CURRENT_SCHEMA_VERSION`]. Failures are logged but never block
+//! startup — the next launch retries.
+//!
+//! ## Adding a new migration
+//!
+//! 1. Add a module here (e.g. `mod my_migration;`).
+//! 2. Bump [`CURRENT_SCHEMA_VERSION`].
+//! 3. Extend [`run_pending`] with a `if config.schema_version < N`
+//!    branch that calls the new module and bumps `config.schema_version`
+//!    on success.
+//!
+//! ## Distinction from `crate::openhuman::migration`
+//!
+//! The sibling `migration` (singular) module is a user-triggered RPC
+//! that imports memory from a legacy OpenClaw workspace. This module
+//! (`migrations`, plural) is the automatic schema-version runner that
+//! fires once per workspace on first launch of a new build.
+
+use crate::openhuman::config::Config;
+
+mod phase_out_profile_md;
+
+/// Current target schema version. Bumped alongside every new migration.
+pub const CURRENT_SCHEMA_VERSION: u32 = 1;
+
+/// Run any migrations whose `schema_version` gate hasn't yet been
+/// crossed for this workspace.
+///
+/// Best-effort: failures inside a migration are logged and never
+/// propagate. The `schema_version` is only bumped after a migration
+/// reports success **and** the bump is persisted via [`Config::save`],
+/// so a partial run leaves the gate unchanged and the next launch
+/// retries from the same starting version.
+pub async fn run_pending(config: &mut Config) {
+    if config.schema_version >= CURRENT_SCHEMA_VERSION {
+        log::debug!(
+            "[migrations] schema_version={} already at current={} — nothing to do",
+            config.schema_version,
+            CURRENT_SCHEMA_VERSION
+        );
+        return;
+    }
+
+    log::info!(
+        "[migrations] running pending migrations schema_version={} -> {}",
+        config.schema_version,
+        CURRENT_SCHEMA_VERSION
+    );
+
+    // 0 -> 1: phase out PROFILE.md from persisted session transcripts.
+    if config.schema_version < 1 {
+        match phase_out_profile_md::run(&config.workspace_dir) {
+            Ok(stats) => {
+                config.schema_version = 1;
+                if let Err(err) = config.save().await {
+                    log::warn!(
+                        "[migrations] phase_out_profile_md ran but config.save failed: \
+                         {err:#} — will retry on next launch"
+                    );
+                    return;
+                }
+                log::info!(
+                    "[migrations] schema_version bumped to 1 (phase_out_profile_md \
+                     scanned={} cleaned={} skipped={} errors={})",
+                    stats.scanned,
+                    stats.cleaned,
+                    stats.skipped,
+                    stats.errors
+                );
+            }
+            Err(err) => {
+                log::warn!(
+                    "[migrations] phase_out_profile_md failed: {err:#} — \
+                     will retry on next launch"
+                );
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+#[path = "mod_tests.rs"]
+mod tests;
--- a/src/openhuman/migrations/mod.rs
+++ b/src/openhuman/migrations/mod.rs
@@ -53,14 +53,30 @@ pub async fn run_pending(config: &mut Config) {
     );
 
     // 0 -> 1: phase out PROFILE.md from persisted session transcripts.
+    //
+    // The migration body is synchronous fs I/O (read_dir + read_to_string +
+    // write across potentially hundreds of files). `run_pending` is called
+    // from `Config::load_or_init`, which runs on a tokio runtime — so we
+    // move the blocking walk onto a dedicated `spawn_blocking` task to
+    // keep the executor responsive.
     if config.schema_version < 1 {
-        match phase_out_profile_md::run(&config.workspace_dir) {
-            Ok(stats) => {
+        let workspace_dir = config.workspace_dir.clone();
+        let run_result =
+            tokio::task::spawn_blocking(move || phase_out_profile_md::run(&workspace_dir)).await;
+        match run_result {
+            Ok(Ok(stats)) => {
+                let previous_version = config.schema_version;
                 config.schema_version = 1;
                 if let Err(err) = config.save().await {
+                    // Roll the in-memory version back so a subsequent
+                    // `load_or_init` (or future migration) doesn't believe
+                    // we've already crossed this gate when disk still
+                    // says 0. Next launch retries from the same start.
+                    config.schema_version = previous_version;
                     log::warn!(
                         "[migrations] phase_out_profile_md ran but config.save failed: \
-                         {err:#} — will retry on next launch"
+                         {err:#} — rolled in-memory schema_version back to {previous_version}, \
+                         will retry on next launch"
                     );
                     return;
                 }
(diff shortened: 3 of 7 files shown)


src/openhuman/config/schema/load.rs
```diff
@@ -762,6 +763,7 @@ impl Config {
                 initialized = true,
                 "Config loaded"
             );
+            crate::openhuman::migrations::run_pending(&mut config).await;
```
graycyrus, line 779: [minor] **Fresh installs run the migration over an empty workspace.** The new-config branch creates `Config` with `schema_version: 0` (from Default), so `run_pending` walks an empty workspace. It short-circuits fast, but the intent is wrong — fresh installs have nothing to migrate.

Suggestion: set `schema_version: CURRENT_SCHEMA_VERSION` in the fresh-install branch so new workspaces skip all legacy migrations by definition.

src/openhuman/migrations/phase_out_profile_md.rs
```diff
@@ -0,0 +1,389 @@
...
+            }
+            Ok(false) => {
+                stats.skipped += 1;
+                log::debug!(
+                    "[migration:phase-out-profile-md] skipped (already clean) path={}",
+                    path.display()
+                );
+            }
+            Err(err) => {
+                stats.errors += 1;
```
graycyrus, line 119: [minor] **`sweep_tainted_md_companions` only walks one level of subdirectories** under `sessions/`. If deeper nesting exists (e.g. `sessions/2026/05/14/*.md`), those files are missed. The JSONL walk has the same single-level assumption. If the layout is guaranteed at most one level deep, worth documenting the assumption explicitly so future contributors know.

src/openhuman/migrations/phase_out_profile_md.rs
```diff
@@ -0,0 +1,389 @@
...
+//! Gated externally by [`Config::schema_version`] — once the migration
+//! succeeds and the bump is persisted, future launches skip it.
+//! Internally self-idempotent too: a transcript without any
+//! `### PROFILE.md` block is detected and left unchanged.
+//!
+//! ## Fresh installs
+//!
+//! When `session_raw/` does not exist or contains no transcripts the
+//! migration short-circuits without scanning. The caller still bumps
+//! `schema_version` so future launches don't re-check.
```
graycyrus, line 47: [minor] **`run()` is synchronous blocking I/O called from async context.** `run_pending` is an `async fn`, but `phase_out_profile_md::run()` does synchronous filesystem I/O (read_dir, read_to_string, write, remove_file). On a workspace with hundreds of transcripts this blocks the tokio runtime thread.

Suggestion: wrap in `tokio::task::spawn_blocking` in `run_pending`.

src/openhuman/migrations/phase_out_profile_md.rs
```diff
@@ -0,0 +1,389 @@
...
+    Ok(true)
+}
+
+/// Strip a `### PROFILE.md` block from the first system message of a
+/// transcript. Returns `true` when the message was mutated.
+pub(super) fn strip_profile_md_in_first_system_message(session: &mut SessionTranscript) -> bool {
+    let Some(first) = session.messages.first_mut() else {
+        return false;
+    };
+    if first.role != "system" {
```
graycyrus, line 284: [major] **`strip_profile_md_block` always returns `Some` when the heading exists**, even if the reconstructed output is byte-identical to the input. This means `process_transcript` will rewrite files unnecessarily, and the `skipped` counter under-counts.

Add a guard before returning:
```rust
if out == prompt {
    return None;
}
Some(out)
```

src/openhuman/migrations/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    // 0 -> 1: phase out PROFILE.md from persisted session transcripts.
+    if config.schema_version < 1 {
+        match phase_out_profile_md::run(&config.workspace_dir) {
+            Ok(stats) => {
+                config.schema_version = 1;
+                if let Err(err) = config.save().await {
+                    log::warn!(
+                        "[migrations] phase_out_profile_md ran but config.save failed: \
+                         {err:#} — will retry on next launch"
```
graycyrus: [minor] **In-memory `schema_version` not rolled back on save failure.** Line 63 sets `config.schema_version = 1`, but if `config.save()` fails (line 64), the early return at line 68 leaves the in-memory config at version 1 while disk says 0. Benign with one migration, but when migration 2 is added the caller sees version 1 in memory and might skip re-running migration 1 on the next code path.

Suggestion: roll back on save failure:
```rust
Err(err) => {
    config.schema_version = 0;
    log::warn!(...);
    return;
}
```

src/openhuman/config/schema/types.rs
```diff
@@ -28,6 +28,13 @@ pub struct Config {
     pub workspace_dir: PathBuf,
     #[serde(skip)]
     pub config_path: PathBuf,
+    /// Workspace data-schema version. Bumped each time a one-shot data
+    /// migration under [`crate::openhuman::migrations`] runs successfully.
+    /// `#[serde(default)]` so existing `config.toml` files (which predate
+    /// the field) load as version `0` and pick up pending migrations on
+    /// the first launch of the new build.
+    #[serde(default)]
+    pub schema_version: u32,
```
graycyrus, line 37: [nit] PR title says `phase out old PROFILE.md from on schema_version=1` — the `from on` looks like a leftover edit. Should be `on schema_version=1` or `from schema_version=0`.

graycyrus (changes requested):
## Walkthrough

This PR introduces a one-shot startup migration framework gated by `Config::schema_version`. The first migration (`0 → 1`) retires the legacy `PROFILE.md` injection by: (1) deleting `{workspace}/PROFILE.md`, (2) stripping `### PROFILE.md` blocks from JSONL transcripts under `session_raw/`, and (3) sweeping `.md` companions under `sessions/`. The runner is wired into `Config::load_or_init` so it fires before any agent turn can resume a session.

## Changes

| File | Change |
|------|--------|
| `types.rs` | Add `schema_version: u32` to `Config` with `#[serde(default)]` |
| `load.rs` | Call `migrations::run_pending()` in both config-load branches |
| `migrations/mod.rs` | Migration runner: version gate, dispatch, config save, retry |
| `migrations/phase_out_profile_md.rs` | The migration: file deletion, JSONL walk+strip, `.md` sweep |
| `migrations/mod_tests.rs` | 4 integration tests for the runner |
| `migrations/phase_out_profile_md_tests.rs` | 18 unit tests for the migration |
| `mod.rs` | Add `pub mod migrations;` |

## Strengths

- Well-structured module: follows CLAUDE.md dedicated-subdirectory rule, light mod.rs, impl in own file
- 22 tests covering happy paths, edge cases (truncation footer, legacy HTML format, fresh install), idempotency
- Excellent `[migration:phase-out-profile-md]` prefixed logging — easy to diagnose in production
- Best-effort: per-file errors fold into stats without aborting; runner failures don't block startup
- No `.unwrap()` in production code, no hardcoded PII in tests

## Issues

See inline comments — 1 major, 4 minor, 1 nit.

src/openhuman/config/schema/types.rs
sanil-23: Fixed the PR title — was `phase out old PROFILE.md from on schema_version=1`, now `phase out PROFILE.md from disk on schema_version=1` (matches the commit message).

src/openhuman/config/schema/load.rs
sanil-23: Done in c8f5c551 — fresh-install branch now sets `schema_version: CURRENT_SCHEMA_VERSION` directly so new workspaces skip the migration by definition. The defensive `run_pending` call stays for symmetry (short-circuits on the gate).

src/openhuman/migrations/phase_out_profile_md.rs
sanil-23: Documented the one-level-deep assumption in c8f5c551 — added a "Layout assumption" paragraph to `sweep_tainted_md_companions` rustdoc referencing the matching assumption in `collect_jsonl_transcripts`. Both walkers descend exactly into `sessions/<date>/*.md` and `session_raw/<legacy-date>/*.jsonl`, matching the layout produced by `agent/harness/session/transcript`.

src/openhuman/migrations/mod.rs
sanil-23: Fixed in c8f5c551 — added the rollback path. If `config.save()` fails we now reset `config.schema_version` to its pre-migration value before logging+returning, so the in-memory state never diverges ahead of disk. Added `run_pending_rolls_back_schema_version_when_save_fails` (forces save() to fail by pointing config_path through a non-directory) to lock the semantics.

src/openhuman/migrations/phase_out_profile_md.rs
sanil-23: Good catch, fixed in c8f5c551 — `strip_profile_md_block` now returns `None` when `out == prompt` even after the heading was located, so degenerate-whitespace cases no longer churn the file. Added `strip_block_returns_none_when_reconstruction_is_byte_identical` to lock the behavior.

src/openhuman/migrations/phase_out_profile_md.rs
sanil-23: Fixed in c8f5c551 — `run_pending` now wraps the migration in `tokio::task::spawn_blocking(move || phase_out_profile_md::run(&workspace_dir))` so the fs walk runs on a blocking thread instead of stalling the executor.

sanil-23:
@graycyrus thanks for the review. All six points addressed in `c8f5c551`:

| # | Finding | Resolution |
|---|---|---|
| 1 | Fresh-install branch ran migration over empty workspace | `load.rs:743` fresh-install branch now stamps `schema_version: CURRENT_SCHEMA_VERSION` directly so new workspaces skip the walk by definition. The `run_pending` call stays for symmetry — short-circuits on the version gate. |
| 2 | One-level-deep walk under `sessions/` undocumented | Added a "Layout assumption (one level deep)" paragraph to `sweep_tainted_md_companions` rustdoc, referencing the matching assumption in `collect_jsonl_transcripts`. Both walkers descend exactly into `sessions/<date>/*.md` and `session_raw/<legacy-date>/*.jsonl`, matching what `agent/harness/session/transcript` writes. |
| 3 | Synchronous I/O on async runtime | `run_pending` wraps `phase_out_profile_md::run` in `tokio::task::spawn_blocking(move \|\| phase_out_profile_md::run(&workspace_dir))` so the fs walk runs on a blocking thread instead of stalling the executor. |
| 4 (major) | `strip_profile_md_block` returned `Some` even when output was byte-identical | Added `if out == prompt { return None; }` guard. New test `strip_block_returns_none_when_reconstruction_is_byte_identical` locks the behavior. |
| 5 | In-memory `schema_version` not rolled back on `save()` failure | Captured `previous_version` before bumping; on `save()` error we restore it and log the rollback, so the in-memory value never gets ahead of disk. New test `run_pending_rolls_back_schema_version_when_save_fails` forces save() to fail (config_path through a non-directory) and asserts `schema_version == 0` after. |
| 6 (nit) | PR title typo `from on` | Edited title to `chore(migrations): phase out PROFILE.md from disk on schema_version=1` (matches the commit message). |

24 tests pass (up from 22), `cargo fmt --check` clean, `cargo clippy --lib --tests` clean in the diff cone.

Mind taking another look when you have a minute?

PR #1784, opened 2026-05-15
fix(composio): gracefully handle unknown method for trigger settings RPC

### LINKED ISSUE #1597: RPC 'openhuman.get_composio_trigger_settings' unknown — frontend/core version mismatch
## Summary

Users hit "unknown method: openhuman.get_composio_trigger_settings" errors across production and staging. The handler is defined in `src/openhuman/config/schemas.rs` (lines 278-279, 1080), but users are getting unknown method errors — indicating their core binary doesn't include this method.

## Sentry Issue

- [OPENHUMAN-TAURI-1J](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-1J) — 13 events over 5 days (May 8-13)

**Error:** `unknown method: openhuman.get_composio_trigger_settings`
**Environments:** production + staging
**Affected users:** Multiple users across US, India, Netherlands, Sweden, Morocco, South Africa, Iraq
**Platforms:** macOS (aarch64 + x86_64)

## Analysis

The method IS registered in the codebase:
- Schema defined at `src/openhuman/config/schemas.rs:724`
- Handler at `src/openhuman/config/schemas.rs:1080`
- Registered controller at `src/openhuman/config/schemas.rs:278`

Likely causes:
1. **Version mismatch** — the Tauri frontend (auto-updated) calls this method, but users are running an older core sidecar binary that predates the registration
2. **Config controller schemas not wired** — the config domain's controllers may not be registered in `src/core/all.rs` dispatch chain

## Suggested Fix

1. Verify the config domain controllers are registered in the RPC dispatch (`src/core/all.rs`)
2. The frontend should gracefully handle unknown method errors for optional features (composio trigger settings) instead of surfacing errors to Sentry
3. Consider adding a version handshake between frontend and core to detect mismatches early

**Priority:** High

graycyrus:
## Summary

- Wraps `openhumanGetComposioTriggerSettings` and `openhumanUpdateComposioTriggerSettings` in try/catch that detects "unknown method" errors from stale core sidecar binaries
- Returns safe defaults (`triage_disabled: false`, empty toolkits array) instead of propagating the error to Sentry
- Non-"unknown method" errors still propagate normally

Closes #1597

## Test plan

- [x] Added 4 unit tests: graceful degradation for both get/update, and error re-throw for non-unknown-method errors
- [x] All 14 config tests pass
- [x] ESLint clean on changed files
- [x] Prettier clean on changed files
- [x] N/A: No Rust changes — dispatch chain already correct on main
- [x] N/A: No E2E changes — settings panel already handles defaults at UI level

## Summary by CodeRabbit

* **Bug Fixes**
  * Better compatibility with older core versions: unavailable Composio configuration calls are handled gracefully, returning sensible defaults and logging a graceful-degradation message instead of causing failures.
* **Tests**
  * Added tests ensuring unknown-method errors yield no-op/default responses and that other error types continue to be propagated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1784?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/utils/tauriCommands/config.ts
```diff
@@ -298,10 +298,18 @@ export async function openhumanUpdateComposioTriggerSettings(
...
-  });
+  try {
+    return await callCoreRpc<CommandResponse<ConfigSnapshot>>({
+      method: 'openhuman.config_update_composio_trigger_settings',
+      params: update,
+    });
+  } catch (err) {
+    if (tauriErrorMessage(err).includes('unknown method')) {
+      // Stale core sidecar predates composio trigger settings (#1597).
+      return { result: {} as ConfigSnapshot, logs: [] };
```
graycyrus: **[major]** `{} as ConfigSnapshot` is an unsafe type cast that hides missing required fields (`config`, `workspace_dir`, `config_path`). If any future caller reads these fields from the update response, they'll get `undefined` at runtime while TypeScript thinks they're populated.

The current caller (`ComposioTriagePanel`) doesn't inspect the result beyond awaiting it, so this won't break today — but it's a footgun.

Suggestion — provide a structurally valid default:
```ts
return { result: { config: {}, workspace_dir: '', config_path: '' } as ConfigSnapshot, logs: [] };
```
Or, if you'd rather be explicit that this is a degraded response, consider a `Partial<ConfigSnapshot>` return type for the fallback path (though that's a larger change).

app/src/utils/tauriCommands/config.ts
```diff
@@ -298,10 +298,18 @@ export async function openhumanUpdateComposioTriggerSettings(
...
-    params: update,
-  });
+  try {
+    return await callCoreRpc<CommandResponse<ConfigSnapshot>>({
+      method: 'openhuman.config_update_composio_trigger_settings',
+      params: update,
+    });
+  } catch (err) {
+    if (tauriErrorMessage(err).includes('unknown method')) {
+      // Stale core sidecar predates composio trigger settings (#1597).
```
graycyrus, line 366: **[minor]** Per project standards (known issues watchlist), new/changed error flows should include debug logging with grep-friendly prefixes. These catch blocks silently degrade, which makes it harder to diagnose version-mismatch issues in the field.

Suggestion — add a debug log in both catch blocks:
```ts
import debug from 'debug';
const log = debug('composio:rpc');
// ...
log('graceful degradation: stale core lacks openhuman.config_update_composio_trigger_settings (#1597)');
```
This way `DEBUG=composio:*` surfaces these in dev, and they're zero-cost in production.

graycyrus (commented):
## Walkthrough

Clean, focused fix for #1597 — wraps the two composio trigger settings RPC calls in try/catch to gracefully degrade when users are running a stale core sidecar that doesn't have these methods registered. The get function returns safe defaults, the update returns a no-op response. Good test coverage with 4 new tests. One type-safety issue to address and a logging gap per project standards.

## Change Summary

| `app/src/utils/tauriCommands/config.ts` | Modified | Added try/catch with "unknown method" detection around both composio RPC calls, returning fallback values |
| `app/src/utils/tauriCommands/config.test.ts` | Modified | Added 4 tests: graceful degradation + error re-throw for both get and update functions |

## Per-file analysis

### `config.ts`

The `openhumanGetComposioTriggerSettings` fallback is well-constructed — `{ triage_disabled: false, triage_disabled_toolkits: [] }` matches the `ComposioTriggerSettings` interface properly.

The `openhumanUpdateComposioTriggerSettings` fallback uses `{} as ConfigSnapshot` which is an unsafe cast — see inline comment.

Neither catch block logs the graceful degradation, which makes silent failures harder to diagnose in the field.

### `config.test.ts`

Tests are solid — they cover both the graceful path and the re-throw path for each function. Issue reference (#1597) in test names is a nice touch for traceability. Structure matches existing test patterns in the file.

app/src/utils/tauriCommands/config.ts
```diff
@@ -298,10 +298,18 @@ export async function openhumanUpdateComposioTriggerSettings(
...
+      return { result: {} as ConfigSnapshot, logs: [] };
+    }
```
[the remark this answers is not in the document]
graycyrus: Fixed in 9e92ff0c — replaced `{} as ConfigSnapshot` with structurally valid default `{ config: {}, workspace_dir: '', config_path: '' }`. Good catch.

app/src/utils/tauriCommands/config.ts
graycyrus: Fixed in 9e92ff0c — replaced the unsafe cast with a structurally valid default: `{ config: {}, workspace_dir: '', config_path: '' }`.

app/src/utils/tauriCommands/config.ts
graycyrus: Fixed in 9e92ff0c — added `debug('composio:rpc')` logging in both catch blocks. Surfaces via `DEBUG=composio:*` in dev, zero-cost in production.

graycyrus:
@graycyrus CI is failing on changes in this PR — please fix before review.

Issue #1675, opened 2026-05-13
AI repeats every Chinese response twice when asked to answer in Chinese


tiedan126:
Issue Description
When I ask OpenHuman to respond in Chinese, the AI outputs every sentence or paragraph twice (duplicated). This significantly affects the reading experience and makes conversations inefficient.

Steps to Reproduce

Start a conversation with OpenHuman (local core mode).

Ask a question in Chinese, or explicitly request: “请用中文回答” (Please answer in Chinese).

Observe the AI’s reply.

Expected Behavior
The AI should respond once in fluent, natural Chinese.

Actual Behavior
The entire response is repeated verbatim. For example, if the AI should say “你好，我是OpenHuman”，it outputs “你好，我是OpenHuman 你好，我是OpenHuman”.

Screenshots / Logs (optional)
(You can attach a screenshot or copy a conversation snippet if needed.)

Environment (please complete the following information):

OS: [e.g. Windows 11, macOS Sequoia]

OpenHuman version: [e.g. latest as of 2026-05-14]

Deployment mode: Local (embedded core)

Connected providers: Gmail (optional)

Language settings: Default, but occurs only when Chinese is used

Additional Context

The issue does not occur when the AI answers in English – only when replying in Chinese.

It might be related to the token handling or the response formatting pipeline for CJK languages.

I am using the default local model (or please specify if you are using a specific LLM like Ollama, GPT, etc.). In my case: [e.g. local Ollama with qwen2.5, or default OpenHuman cloud model] – you can fill in your actual backend model.

Possible Hypothesis
The duplication could be caused by the TokenJuice compression or the memory tree’s output generation step accidentally appending the same response twice when processing Unicode (Chinese) characters.

Thank you for looking into this!

CodeGhost21:
@CursorAgent use the Cursor Cloud environment for tinyhumansai/openhuman.

Work issue #1675.
Expected path: /workspace/openhuman.
Start from latest origin/main.
Create branch cursor/1675-ai-repeats-every-chinese-response-twice-when-asked.
Lane: rust-core/agent. Owned paths: src/openhuman/agent/. Do not edit any file outside those subtrees.
Follow docs/agent-workflows/cursor-cloud-agents.md exactly.
Do not open duplicate PRs. If validation is blocked, report the exact command and error in the PR body and on this issue.

sanil-23:
[...]

PR #1794, opened 2026-05-15
fix(streaming): fix SSE buffer corruption and add CJK sentence splitting

Sathvik-1007:
## Summary

- Fix erroneous post-drain buffer slice in `sse_bytes_to_chunks` that corrupted multi-line SSE buffers
- Add CJK fullwidth sentence terminators (。！？) to presentation layer's `split_sentences` so Chinese responses get proper multi-bubble segmentation

## Problem

- `compatible_stream.rs:56` had a redundant `buffer = buffer[pos + 1..].to_string()` after `buffer.drain(..=pos)` — the drain already removes the processed bytes, so the second slice corrupts the buffer on multi-line SSE payloads. For CJK content where SSE chunks may contain multiple lines, this causes data loss or panics at byte boundaries.
- `split_sentences` in `presentation.rs` only recognized Latin sentence terminators (`.!?` + space + uppercase). Chinese text using `。！？` was never split into sentences, always falling through to single-bubble delivery. This prevented proper segmentation of Chinese responses.

## Solution

- Removed the dead `buffer = buffer[pos + 1..].to_string()` line — `drain(..=pos)` already leaves the buffer with only the unprocessed remainder.
- Added CJK fullwidth sentence terminators (`\u3002`, `\uFF01`, `\uFF1F`) as split points in `split_sentences`. Chinese text now segments into multiple bubbles like English text does.

## Submission Checklist

- [x] Tests added or updated — existing presentation tests pass; the buffer fix is in a code path that was previously dead/panicking
- [x] **Diff coverage ≥ 80%** — N/A: 1-line removal + 8-line addition in well-tested module (28 tests pass)
- [x] Coverage matrix updated — N/A: no new features
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: bug fix only
- [x] Manual smoke checklist updated — N/A: no release-cut surface changes

## Impact

- **Desktop/Mobile/Web**: Chinese/CJK responses now properly segment into multiple chat bubbles instead of always being a single block
- **Streaming**: SSE buffer handling is now correct for multi-line payloads (prevents potential data loss on high-throughput local model streams)
- **No breaking changes**: English text behavior unchanged; segmentation is additive

## Related

- Refs #1675
- Note: The full duplication reported in #1675 may also involve model-level repetition (known issue with some local Ollama models like qwen2.5). These fixes address the infrastructure-level causes — SSE buffer corruption and missing CJK segmentation — that could contribute to or mask the issue.

## Summary by CodeRabbit

* **New Features**
  * Added CJK (Chinese/Japanese/Korean) sentence splitting to improve message segmentation and delivery timing.

* **Performance**
  * Optimized Server-Sent Events buffer handling for more efficient stream processing and reduced redundant work.

* **Tests**
  * Added a unit test verifying CJK terminator sentence splitting produces expected segments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/1794)

### DIFF
--- a/src/openhuman/channels/providers/presentation.rs
+++ b/src/openhuman/channels/providers/presentation.rs
@@ -311,6 +311,7 @@ fn split_sentences(text: &str) -> Vec<String> {
         current.push(chars[i]);
         let ch = chars[i];
 
+        // Latin sentence terminators: split on ". " followed by uppercase.
         if (ch == '.' || ch == '!' || ch == '?')
             && i + 2 < chars.len()
             && chars[i + 1] == ' '
@@ -322,6 +323,17 @@ fn split_sentences(text: &str) -> Vec<String> {
             }
             current.clear();
             i += 2; // skip the space
+            continue;
+        }
+
+        // CJK sentence terminators: split after fullwidth period/exclamation/question.
+        if (ch == '\u{3002}' || ch == '\u{FF01}' || ch == '\u{FF1F}') && i + 1 < chars.len() {
+            let trimmed = current.trim().to_string();
+            if !trimmed.is_empty() {
+                parts.push(trimmed);
+            }
+            current.clear();
+            i += 1;
             continue;
         }
 
--- a/src/openhuman/channels/providers/presentation_tests.rs
+++ b/src/openhuman/channels/providers/presentation_tests.rs
@@ -181,6 +181,15 @@ fn split_sentences_single_sentence_without_terminator() {
 }
 
 #[test]
+fn split_sentences_splits_on_cjk_terminators() {
+    let out = split_sentences("你好世界。今天天气很好！你觉得呢？");
+    assert_eq!(out.len(), 3);
+    assert_eq!(out[0], "你好世界。");
+    assert_eq!(out[1], "今天天气很好！");
+    assert_eq!(out[2], "你觉得呢？");
+}
+
+#[test]
 fn group_sentences_single_entry_roundtrip() {
     let v: Vec<String> = vec!["Hello world".into()];
     let out = group_sentences(&v);
--- a/src/openhuman/providers/compatible_stream.rs
+++ b/src/openhuman/providers/compatible_stream.rs
@@ -52,8 +52,7 @@ pub(crate) fn sse_bytes_to_chunks(
 
                     // Process complete lines
                     while let Some(pos) = buffer.find('\n') {
-                        let line = buffer.drain(..=pos).collect::<String>();
-                        buffer = buffer[pos + 1..].to_string();
+                        let line: String = buffer.drain(..=pos).collect();
 
                         match parse_sse_line(&line) {
                             Ok(Some(content)) => {


src/openhuman/channels/providers/presentation.rs
```diff
@@ -325,6 +326,17 @@ fn split_sentences(text: &str) -> Vec<String> {
             continue;
         }
 
+        // CJK sentence terminators: split after fullwidth period/exclamation/question.
```
graycyrus, line 329: **[minor]** No debug logging for this new code path. Sibling code in this module uses `tracing::debug!` with `[presentation:segment]`-style prefixes. Consider adding a trace-level log when CJK splitting activates, e.g.:

```rust
tracing::trace!("[presentation:split] CJK terminator at pos {i}");
```

Not blocking — the existing Latin split path doesn't log either, so this is consistent with the current code. But the project convention is to add debug/trace logging to new flow branches.

src/openhuman/channels/providers/presentation_tests.rs
```diff
@@ -180,6 +180,15 @@ fn split_sentences_single_sentence_without_terminator() {
     assert_eq!(out.len(), 1);
 }
```
graycyrus, line 182: **[minor]** Good happy-path coverage. A few edge cases worth adding when convenient:

- **Terminator at end of string**: `"你好世界。"` → should return `["你好世界。"]` (single segment, handled by post-loop cleanup)
- **Mixed CJK + Latin**: `"Hello world。你好！Done."` → verify both split paths cooperate
- **Single CJK sentence** (no terminator): `"你好世界"` → should return `["你好世界"]`
- **Consecutive terminators**: `"你好。。世界"` → verify empty segments are filtered by the `!trimmed.is_empty()` guard

Not blocking — the current test validates the core behavior correctly.

graycyrus (commented):
## Walkthrough

Clean, well-scoped PR that fixes a real SSE buffer corruption bug in `compatible_stream.rs` and adds CJK sentence splitting in `presentation.rs`. The buffer fix is clearly correct — `buffer.drain(..=pos)` already mutates the buffer in place, so the old `buffer = buffer[pos + 1..]` line was double-advancing and corrupting multi-line SSE payloads. The CJK splitting follows the existing Latin-terminator pattern nicely. Only minor gaps below.

## Change Summary

| `src/openhuman/providers/compatible_stream.rs` | Bug fix | Remove redundant buffer slice after `drain(..=pos)` that corrupted multi-line SSE payloads |
| `src/openhuman/channels/providers/presentation.rs` | Enhancement | Add CJK fullwidth sentence terminators (。！？) as split points in `split_sentences` |
| `src/openhuman/channels/providers/presentation_tests.rs` | Test | Add happy-path test for CJK sentence splitting |

## Per-file Analysis

### `compatible_stream.rs`
The fix is correct and minimal. After `buffer.drain(..=pos)`, bytes 0..=pos (including the `\n`) are removed from `buffer` and collected into `line`. The old second line `buffer = buffer[pos + 1..].to_string()` was indexing into the *already-drained* buffer — effectively skipping `pos + 1` characters from the *remaining* data. On single-line SSE payloads this was harmless (buffer was empty after drain), but on multi-line payloads (common with CJK content) it caused data loss or panics at byte boundaries. Good catch.

### `presentation.rs`
The CJK block mirrors the Latin terminator block structurally. The `i + 1 < chars.len()` guard means a CJK terminator at the very end of the string falls through to the post-loop cleanup rather than being split here — this is correct and consistent with how the Latin block's last-sentence handling works. The terminators chosen (U+3002, U+FF01, U+FF1F) are the standard CJK fullwidth equivalents.

### `presentation_tests.rs`
The happy-path test is good and verifies all three CJK terminators in one pass. See inline comment for suggested edge cases.

PR #2034, opened 2026-05-18
fix: block symlink escapes in path policy

### LINKED ISSUE #1927: Security: Symlink bypass in is_path_allowed() — no canonicalization
## Summary

`src/openhuman/security/policy.rs` — `is_path_allowed()` validates the path string but does not resolve symlinks before checking against `forbidden_paths`. A symlink created inside the workspace pointing to a forbidden path (e.g., `/etc/shadow`) passes the workspace-relative path check.

The separate `is_resolved_path_allowed()` exists but must be called manually — there's no enforcement that both checks are always paired.

## Location

`src/openhuman/security/policy.rs` — `is_path_allowed()`

## Impact

**Medium** — An agent can read/write forbidden files by creating a symlink inside the allowed workspace directory.

## Suggested Fix

Merge `is_path_allowed()` and `is_resolved_path_allowed()` into a single function that always canonicalizes before checking. Or at minimum, document and enforce via lint that both must always be called together.

LawyerLyu:
## Summary

- Harden `SecurityPolicy::is_path_allowed` against workspace symlink escapes.
- Resolve the existing path, or nearest existing parent, for workspace-relative paths before allowing them.
- Add a Unix regression test covering both a symlink path and a not-yet-created child under a symlinked directory.

## Problem

Issue #1927 reports that `is_path_allowed()` validates the path string but does not resolve symlinks. A path such as `escape_link` can look workspace-relative while resolving outside the workspace, leaving callers dependent on remembering a separate `is_resolved_path_allowed()` check.

## Solution

- Add a small helper that canonicalizes the path if it exists, or walks upward to canonicalize the nearest existing parent.
- In workspace-only mode, join relative paths to `workspace_dir` and reject them if the resolved existing prefix is outside the workspace.
- Keep non-existing normal workspace paths allowed so write/create flows still work.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — focused regression test covers the new symlink guard; CI Coverage Gate will report the exact diff coverage.
- [x] Coverage matrix updated — N/A: security policy hardening; no feature row added/removed/renamed.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no matrix feature ID applies.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: no release manual smoke surface changed.

## Impact

Workspace-only file policy now rejects symlink paths that resolve outside the configured workspace even when callers only use `is_path_allowed()`. Normal relative paths and non-existing create targets inside the workspace remain allowed.

No migration or dependency impact.

## Related

- Closes #1927
- Follow-up PR(s)/TODOs: N/A

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/security-policy-symlink-paths`
- Commit SHA: `4a43f9d`

### Validation Run
- [x] `cargo test -p openhuman is_path_allowed_blocks_workspace_symlink_escape --lib`
- [x] `cargo fmt --manifest-path Cargo.toml --all --check`
- [x] Full pre-push hook: `git push -u fork fix/security-policy-symlink-paths` (format, lint, compile, Rust check, command token lint)
- [x] Rust fmt/check (if changed): passed via focused fmt check and pre-push hook.
- [x] Tauri fmt/check (if changed): passed via pre-push hook; no Tauri files changed.

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: reject workspace-relative symlink paths whose existing target/prefix resolves outside the workspace.
- User-visible effect: tool calls that attempt to traverse symlinks out of the workspace are blocked earlier.

### Parity Contract
- Legacy behavior preserved: normal workspace-relative paths, dotfiles in the workspace, and not-yet-created files under real workspace parents remain allowed.
- Guard/fallback/dispatch parity checks: `is_resolved_path_allowed()` remains available for callers that already canonicalize; this PR adds defense in depth to `is_path_allowed()`.

- Duplicate PR(s): N/A
- Canonical PR: this PR

## Summary by CodeRabbit

* **Security**
  * Enhanced path validation to block attempts to access files outside the configured workspace, including symlink-based escapes and traversal cases.

* **Tests**
  * Added a regression test validating symlink escape prevention.
  * Updated existing file-read/write tests to accept broader error outcomes from earlier path validation while preserving checks that outside files remain unmodified.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2034?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/security/policy.rs
+++ b/src/openhuman/security/policy.rs
@@ -383,6 +383,17 @@ fn contains_unquoted_char(command: &str, target: char) -> bool {
     false
 }
 
+fn resolve_existing_path_prefix(path: &Path) -> Option<PathBuf> {
+    let mut current = path;
+    loop {
+        if current.exists() {
+            return current.canonicalize().ok();
+        }
+
+        current = current.parent()?;
+    }
+}
+
 impl SecurityPolicy {
     /// Classify command risk. Any high-risk segment marks the whole command high.
     pub fn command_risk_level(&self, command: &str) -> CommandRiskLevel {
@@ -718,6 +729,15 @@ impl SecurityPolicy {
             let forbidden_path = Path::new(&forbidden_expanded);
             if expanded_path.starts_with(forbidden_path) {
                 return false;
+            }
+        }
+
+        if self.workspace_only && !path.is_empty() {
+            let candidate = self.workspace_dir.join(&expanded);
+            if let Some(resolved) = resolve_existing_path_prefix(&candidate) {
+                if !self.is_resolved_path_allowed(&resolved) {
+                    return false;
+                }
             }
         }
 
--- a/src/openhuman/security/policy_tests.rs
+++ b/src/openhuman/security/policy_tests.rs
@@ -987,6 +987,40 @@ fn resolved_path_blocks_symlink_escape() {
     let _ = std::fs::remove_dir_all(&root);
 }
 
+#[cfg(unix)]
+#[test]
+fn is_path_allowed_blocks_workspace_symlink_escape() {
+    use std::os::unix::fs::symlink;
+
+    let root = std::env::temp_dir().join(format!(
+        "openhuman_test_path_symlink_escape_{}",
+        std::process::id()
+    ));
+    let workspace = root.join("workspace");
+    let outside = root.join("outside_target");
+
+    let _ = std::fs::remove_dir_all(&root);
+    std::fs::create_dir_all(&workspace).unwrap();
+    std::fs::create_dir_all(&outside).unwrap();
+    symlink(&outside, workspace.join("escape_link")).unwrap();
+
+    let policy = SecurityPolicy {
+        workspace_dir: workspace,
+        ..SecurityPolicy::default()
+    };
+
+    assert!(
+        !policy.is_path_allowed("escape_link"),
+        "existing symlink targets outside the workspace must be blocked"
+    );
+    assert!(
+        !policy.is_path_allowed("escape_link/new-file.txt"),
+        "paths beneath symlinked directories outside the workspace must be blocked"
+    );
+
+    let _ = std::fs::remove_dir_all(&root);
+}
+
 #[test]
 fn is_path_allowed_blocks_null_bytes() {
     let policy = default_policy();
--- a/src/openhuman/tools/impl/filesystem/file_read.rs
+++ b/src/openhuman/tools/impl/filesystem/file_read.rs
@@ -327,7 +327,8 @@ mod tests {
         // caught before canonicalize — error says "not allowed" rather than
         // "escapes workspace", but the operation is still correctly blocked.
         assert!(
-            result.output().contains("not allowed") || result.output().contains("escapes workspace")
+            result.output().contains("not allowed")
+                || result.output().contains("escapes workspace")
         );
 
         let _ = tokio::fs::remove_dir_all(&root).await;
--- a/src/openhuman/tools/impl/filesystem/file_read.rs
+++ b/src/openhuman/tools/impl/filesystem/file_read.rs
@@ -323,7 +323,12 @@ mod tests {
         let result = tool.execute(json!({"path": "escape.txt"})).await.unwrap();
 
         assert!(result.is_error);
-        assert!(result.output().contains("escapes workspace"));
+        // is_path_allowed now resolves symlinks eagerly, so the block is
+        // caught before canonicalize — error says "not allowed" rather than
+        // "escapes workspace", but the operation is still correctly blocked.
+        assert!(
+            result.output().contains("not allowed") || result.output().contains("escapes workspace")
+        );
 
         let _ = tokio::fs::remove_dir_all(&root).await;
     }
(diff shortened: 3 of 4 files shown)


src/openhuman/security/policy.rs
```diff
@@ -721,6 +732,15 @@ impl SecurityPolicy {
             }
         }
 
+        if self.workspace_only && !path.is_empty() {
+            let candidate = self.workspace_dir.join(&expanded);
+            if let Some(resolved) = resolve_existing_path_prefix(&candidate) {
+                if !self.is_resolved_path_allowed(&resolved) {
+                    return false;
```
graycyrus, line 739: **[major]** Fail-open on `None` — if `resolve_existing_path_prefix` returns `None` (canonicalization fails), this `if let` silently falls through to `return true`, allowing the path. Security code should fail closed.

On Unix this is practically unreachable (root always exists and canonicalizes), but a fail-open pattern in a security-sensitive path validation is a footgun. If `canonicalize()` ever fails (permissions, race condition, future platform), the entire symlink check is silently bypassed.

Suggestion — deny by default:
```rust
if self.workspace_only && !path.is_empty() {
    let candidate = self.workspace_dir.join(&expanded);
    match resolve_existing_path_prefix(&candidate) {
        Some(resolved) => {
            if !self.is_resolved_path_allowed(&resolved) {
                return false;
            }
        }
        None => return false, // fail closed
    }
}
```

src/openhuman/security/policy_tests.rs
```diff
@@ -987,6 +987,40 @@ fn resolved_path_blocks_symlink_escape() {
...
+    );
+    assert!(
+        !policy.is_path_allowed("escape_link/new-file.txt"),
+        "paths beneath symlinked directories outside the workspace must be blocked"
+    );
+
+    let _ = std::fs::remove_dir_all(&root);
+}
+
 #[test]
```
graycyrus, line 1024: **[minor]** Missing positive test case — this test only verifies that escape paths are *blocked*, but doesn't verify that a normal workspace-relative path still *passes* after the change. Adding a quick `assert!(policy.is_path_allowed("normal-file.txt"))` would guard against over-restrictive regressions (e.g., if the `None` branch from `resolve_existing_path_prefix` were changed to deny).

graycyrus (changes requested):
## Walkthrough

This PR adds symlink-escape protection to `is_path_allowed()` by resolving the nearest existing path prefix before checking workspace containment. The approach is sound and addresses issue #1927. CodeRabbit already flagged the dangling-symlink gap and the missing debug logging — I won't repeat those. I found one additional security concern in the fail-open logic and a minor test gap.

## Change Summary

| `src/openhuman/security/policy.rs` | Security fix | Added `resolve_existing_path_prefix()` helper + integrated into `is_path_allowed()` |
| `src/openhuman/security/policy_tests.rs` | Test | New `is_path_allowed_blocks_workspace_symlink_escape` test |
| `src/openhuman/tools/impl/filesystem/file_read.rs` | Test update | Loosened assertion to accept both error messages |
| `src/openhuman/tools/impl/filesystem/file_write.rs` | Test update | Loosened assertions to accept both error messages |

## Per-file analysis

### `policy.rs` — `resolve_existing_path_prefix` + integration

The helper walks up the path tree until it finds an existing component, then canonicalizes it. Clean, focused logic. The integration into `is_path_allowed()` correctly gates on `workspace_only` mode and non-empty paths.

**However**, the integration uses `if let Some(resolved)` which silently falls through to `return true` when `None` is returned. This is a fail-open pattern in security-critical code. See inline comment.

### `policy_tests.rs` — new regression test

Good test structure following the existing pattern (`#[cfg(unix)]`, temp dirs, cleanup). Covers symlink-to-outside and child-under-symlinked-dir. The test cleanup uses the same pattern as `resolved_path_blocks_symlink_escape` above it.

### `file_read.rs` / `file_write.rs` — assertion loosening

Appropriate — the security invariant (operation blocked) is still verified. The comments explaining why the assertions changed are a nice touch.

M3gA-Mind:
Hey @LawyerLyu — thank you sincerely for taking the time to dig into this vulnerability and put together a fix. The `resolve_existing_path_prefix` helper is a clean piece of logic and your test structure (temp dirs, cleanup, `#[cfg(unix)]` gate) follows the existing patterns really well.

We're closing this PR in favour of #2111, which takes the same core idea further: it renames `is_path_allowed` to `is_path_string_allowed` to force a compile-time audit of every call site, then introduces a unified async `validate_path()` / `validate_parent_path()` API that always canonicalises before checking. That approach also covers two issues the reviewers surfaced here — the dangling-symlink bypass (because `Path::exists()` follows symlinks and returns `false` for dangling ones, the walker skips past them to an ancestor inside the workspace) and the fail-open `if let Some(...)` guard — as well as extending protection to all 8 filesystem tools and fixing a pre-`create_dir_all` race in `file_write` and `csv_export`.

Your PR was genuinely valuable: it helped sharpen exactly what a correct fix needed to handle, and the reviewer feedback it attracted directly informed the final solution. Really appreciate the contribution. 🙏

Issue #1574, opened 2026-05-12
Per-(row, model) embedding storage for memory_tree + event_log + segments


sanil-23:
## Summary

Replace the single `embedding BLOB` column on memory_tree chunks/summaries, episodic events, and conversation segments with a generic per-(row, model) join table. Each model gets its own row per entity, indexed by a `model_signature` string (`"{provider}:{model_id}:{dim}"`). Switching the embedder becomes a query-time lookup rather than a destructive re-embed; toggling between providers (e.g. cloud Voyage ↔ local bge-m3) is safe and reversible.

UnifiedMemory's `vector_chunks` and `episodic_log` are explicitly out of scope — `vector_chunks` is vestigial (the agent has moved to memory_tree for semantic memory) and `episodic_log` is FTS5-only with no embedding column.

## Problem

The current schema attaches one `embedding BLOB` per entity with no provenance metadata. Provider switches produce three concrete failure modes — none of which are caught today:

| Switch | Failure mode |
|---|---|
| cloud Voyage 1024-d → all-minilm 384-d | Cosine `0.0` on dim mismatch (`unified/helpers.rs:64`) — old vectors silently dropped from rerank. Loud but lossy. |
| cloud Voyage 1024-d → bge-m3 1024-d | **Silent.** Same dim, different latent space. Cosine produces plausible scores between 0–1 that rank memories essentially at random. Retrieval is poisoned with no error, no Sentry event, no log line — only degraded answers. |
| Any switch back to a model used before | Re-embed-on-switch does not exist anywhere; old vectors stay tagged with whatever produced them but the schema has no way to filter on that. |

Affected tables:

| Table | File:line | Used by |
|---|---|---|
| `mem_tree_chunks` | `memory/tree/store.rs:39+` | memory_tree retrieval rerank |
| `mem_tree_summaries` | `memory/tree/store.rs:128+` | memory_tree summary rerank |
| `event_log` | `memory/store/unified/events.rs:16-29` | Episodic event semantic recall (Archivist) |
| `conversation_segments` | `memory/store/unified/segments.rs:15+` | Segment-boundary cosine drift (Archivist) |

This blocks the cloud-Voyage-in-memory_tree story safely shipping in the future, and means today's `MemoryConfig.embedding_provider` toggle is only safe if you never had embeddings before.

## Solution

Pull the `embedding` column out of every affected table into a generic per-(row, model) join table. Identical schema everywhere:

```sql
[...]

PR #2059, opened 2026-05-18
fix(memory): serialize memory_tree schema init across worker pool

oxoxDev:
## Summary

- Four `tree_jobs_worker` tasks were racing into `memory::tree::store::with_connection` on cold start, triggering ~19k Sentry events/day across three SQLite WAL/SHM bootstrap codes.
- Gate `apply_schema` behind a process-wide `Mutex<HashSet<PathBuf>>` so only the first caller per DB path runs WAL setup + `SCHEMA` + `ALTER TABLE` migrations; subsequent callers skip.
- Retry the open+init flow up to 3x on the three transient cold-start codes (14 / 1546 / 4874) with 50→150→500ms backoff as belt-and-suspenders.
- Add a regression test spawning 8 threads against a cold tempdir DB; asserts all `Ok` and `apply_schema` runs exactly once per path.

## Problem

Three Sentry issues, all tagged `domain=memory`, `operation=tree_jobs_worker`, `worker_idx=0..3`, sharing the same root cause:

| Sentry ID | Events | Error message |
|---|---|---|
| OPENHUMAN-TAURI-HH | 14168 | `Failed to initialize memory_tree schema: disk I/O error: Error code 1546: Error attempting to truncate file` |
| OPENHUMAN-TAURI-ZM | 3013 | `Failed to initialize memory_tree schema: disk I/O error: Error code 4874: I/O error within the xShmMap method (trying to resize an existing shared-memory segment)` |
| OPENHUMAN-TAURI-MB | 2177 | `Failed to initialize memory_tree schema: unable to open database file: Error code 14: Unable to open the database file` |

ZM + MB share a `trace_id` — two SQLite errors in the same event chain. Cross-platform (Windows + macOS) → race, not OS-specific. All three codes are flavors of the same race window on WAL+SHM cold start:

- code 14 (`SQLITE_CANTOPEN`) — lockfile race, WAL not yet ready
- code 1546 (`SQLITE_IOERR_TRUNCATE`) — WAL truncation contention
- code 4874 (`SQLITE_IOERR_SHMMAP`) — shared-memory segment resize race

The 4 worker tasks (`WORKER_COUNT = 4` in `memory::tree::jobs::worker`) all enter `with_connection` via `claim_next` immediately after the pool boots. Each connection re-ran `PRAGMA journal_mode=WAL` + `execute_batch(SCHEMA)` + `ALTER TABLE` migrations — fine for a warm DB, racy on cold start when WAL + SHM files don't yet exist.

## Solution

Centralised the init in `store.rs`:

- `SCHEMA_INITIALIZED: OnceLock<Mutex<HashSet<PathBuf>>>` records DB paths whose schema has been applied this process.
- `with_connection` calls `open_and_init_with_retry(&db_path)`:
  - Fast path: path already in set → just open conn + set `busy_timeout` + return.
  - Cold path: open conn, acquire mutex, double-check, run `apply_schema` (WAL + SCHEMA + ALTERs + index) once, insert path, drop mutex.
- The whole flow is wrapped in a 3-attempt retry with 50/150/500ms backoff on the three transient codes; classifier walks the anyhow cause chain so codes buried under `.context()` layers still match.
- Keyed on `PathBuf` (not "process-wide once") so per-test `tempdir()` configs each initialise independently — guards against the stale-handle pattern in `feedback_oncelock_sqlite_test_isolation.md`.

The `busy_timeout = 15s` and `journal_mode = WAL` from the existing code are unchanged.

Test: `with_connection_serialises_concurrent_schema_init` spawns 8 threads racing into `with_connection` on a fresh tempdir. Asserts (a) no errors surface and (b) `apply_schema` runs exactly once for the shared path. Per-path counter avoids false positives from parallel tests in the same binary against their own tempdirs.

Micro-commits (in order):
1. `fix(memory): serialize memory_tree schema init across workers` — the mutex-gated guard + retry in `store.rs`
2. `test(memory): cover concurrent memory_tree schema init regression` — 8-thread regression test in `store_tests.rs`

## Submission Checklist

- [x] N/A: behaviour-change is a serialization guard around an existing init path; covered by the new 8-thread regression test plus the existing `store` test suite (13 tests pass locally). **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/coverage.yml`](../.github/workflows/coverage.yml).
- [x] N/A: behaviour-only change to existing memory_tree init path, no new feature row. Coverage matrix updated — added/removed/renamed feature rows in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md) reflect this change.
- [x] N/A: bug fix tracked via Sentry IDs, not GH-issue feature flow. All affected feature IDs from the matrix are listed in the PR description under `## Related`.
- [x] N/A: change is in the Rust core's SQLite init path; not a release-cut surface. Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)).

## Impact

- **Runtime/platform**: desktop (Windows + macOS observed in Sentry). Eliminates ~19k events/day of `Failed to initialize memory_tree schema` errors from the worker pool's cold-start path.
- **Performance**: hot path (every `with_connection` call after first) now does `Mutex<HashSet>::lock().contains()` instead of running `PRAGMA journal_mode=WAL`, the full `SCHEMA` `execute_batch`, and 12+ `ALTER TABLE` probes. Net speedup per call on warm DBs — small but free, especially under contention.
- **Security/migration/compatibility**: none. Schema is unchanged; the guard only changes *when* the existing migrations are run (once per process per path instead of once per `with_connection` call).

## Related

- Closes OPENHUMAN-TAURI-HH
- Closes OPENHUMAN-TAURI-ZM
- Closes OPENHUMAN-TAURI-MB
- Follow-up PR(s)/TODOs:

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/sentry-mt-init-worker-race`
- Commit SHA: see HEAD of branch

### Validation Run
- [x] N/A: Rust-only change. `pnpm --filter openhuman-app format:check`
- [x] N/A: Rust-only change. `pnpm typecheck`
- [x] Focused tests: `cargo test --lib openhuman::memory::tree::store -- --test-threads=4` → 13/13 pass (includes new `with_connection_serialises_concurrent_schema_init`)
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo check` clean
- [x] N/A: no Tauri shell changes. Tauri fmt/check (if changed)

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: serialise schema init across concurrent `with_connection` callers; skip redundant WAL + SCHEMA + ALTER work after first per process.
- User-visible effect: eliminates the three Sentry "Failed to initialize memory_tree schema" error classes on cold start. No UX-visible delta on success path.

### Parity Contract
- Legacy behavior preserved: every connection still gets `busy_timeout = 15s` set; WAL journal mode persists in the DB header from the first init; all `ALTER TABLE` migrations remain idempotent and are unchanged.
- Guard/fallback/dispatch parity checks: retry classifier (`is_transient_cold_start`) only matches the three documented cold-start codes; non-transient errors still bubble up unchanged.

- Duplicate PR(s): N/A
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * DB schema initialization is now serialized per file and retries transient SQLite startup errors with backoff.
  * Foreign-key enforcement is consistently enabled on every connection.

* **New Features**
  * Schema expanded with embedding and summary fields, importance/reason, backlinks, lifecycle status and index, content pointers, raw references, and an entity "is_user" flag.

* **Tests**
  * Added regression tests for concurrent schema init and transient-error classification, a manual smoke/stress binary for concurrent init, and tightened a test with a global lock to prevent race flakes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2059?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/openhuman/memory/tree/store.rs
```diff
@@ -689,22 +696,155 @@ fn ms_to_utc(ms: i64) -> rusqlite::Result<DateTime<Utc>> {
...
+
+fn schema_init_set() -> &'static Mutex<HashSet<PathBuf>> {
+    SCHEMA_INITIALIZED.get_or_init(|| Mutex::new(HashSet::new()))
+}
+
+/// Per-path counter of how many times `apply_schema` ran for each DB
+/// path. Lets the concurrent-init regression test assert "exactly once
+/// per path" without false positives from other tests in the same
+/// binary (which share `SCHEMA_INITIALIZED` but use their own tempdirs).
+static SCHEMA_APPLY_COUNTS: OnceLock<Mutex<HashMap<PathBuf, usize>>> = OnceLock::new();
```
graycyrus, line 731: **[major]** `SCHEMA_APPLY_COUNTS` and `record_schema_apply()` are compiled and called unconditionally in production (from `try_open_and_init`), but the only reader — `schema_apply_count_for_path_for_tests` — is `#[cfg(test)]`. This is test instrumentation leaking into the production binary: an extra `OnceLock<Mutex<HashMap>>` static, a mutex acquisition, and a HashMap insert on every schema init.

The overhead per-process is negligible (one insert per DB path per lifetime), but it sets a precedent of test counters in production code. Consider wrapping the write side behind `#[cfg(test)]` too:

```rust
fn record_schema_apply(_path: &Path) {
    #[cfg(test)]
    {
        let counts = SCHEMA_APPLY_COUNTS.get_or_init(|| Mutex::new(HashMap::new()));
        let mut guard = counts.lock().expect("memory_tree schema apply count mutex poisoned");
        *guard.entry(_path.to_path_buf()).or_insert(0) += 1;
    }
}
```

Or gate the entire static + function behind `#[cfg(test)]` and call it conditionally.

src/openhuman/memory/tree/store.rs
```diff
@@ -689,22 +696,155 @@ fn ms_to_utc(ms: i64) -> rusqlite::Result<DateTime<Utc>> {
...
+
 /// Open the memory_tree SQLite DB and run a closure against it.
 ///
 /// Visible to sibling modules (e.g. `score::store`) so Phase 2 can reuse
 /// the same connection setup / schema initialisation without duplication.
-pub(crate) fn with_connection<T>(
-    config: &Config,
-    f: impl FnOnce(&Connection) -> Result<T>,
-) -> Result<T> {
+pub fn with_connection<T>(config: &Config, f: impl FnOnce(&Connection) -> Result<T>) -> Result<T> {
```
graycyrus: **[major]** `with_connection` widened from `pub(crate)` to `pub`. The smoke binary is a separate crate target so it needs `pub`, but this makes `with_connection` part of the public API of `openhuman_core` — any downstream crate can now call it directly.

If this is only for the smoke binary, consider:
1. `#[doc(hidden)] pub fn with_connection` — public but hidden from docs, signals "not a stable API"
2. Feature-gate: `#[cfg_attr(not(feature = "internal-bins"), doc(hidden))]`
3. Keep `pub(crate)` and restructure the smoke binary as an integration test or example

Option 1 is the lightest touch and matches how other crates handle "pub but internal" exports.

src/openhuman/memory/tree/store.rs
```diff
@@ -689,22 +696,155 @@ fn ms_to_utc(ms: i64) -> rusqlite::Result<DateTime<Utc>> {
...
+        record_schema_apply(db_path);
+        initialized.insert(db_path.to_path_buf());
+    }
+    Ok(conn)
+}
+
+/// True if `err` (or anything in its cause chain) is one of the three
+/// SQLite codes that fire during cold-start WAL/SHM bootstrap races:
+/// 14 (CANTOPEN), 1546 (IOERR_TRUNCATE), 4874 (IOERR_SHMMAP).
+fn is_transient_cold_start(err: &anyhow::Error) -> bool {
```
graycyrus: **[minor]** The extended codes `14 | 1546 | 4874` are magic numbers. Named constants would improve readability and grep-ability:

```rust
/// SQLite CANTOPEN — lockfile race on cold start.
const SQLITE_CANTOPEN: i32 = 14;
/// SQLite IOERR_TRUNCATE — WAL truncation during bootstrap.
const SQLITE_IOERR_TRUNCATE: i32 = 1546;
/// SQLite IOERR_SHMMAP — SHM resize during bootstrap.
const SQLITE_IOERR_SHMMAP: i32 = 4874;
```

The doc comment on `is_transient_cold_start` already explains them, but constants make the match arm self-documenting and easier to update if new codes surface.

src/openhuman/memory/tree/store_tests.rs
```diff
@@ -236,3 +236,80 @@ fn schema_has_content_path_and_content_sha256_columns() {
...
+    // Second call — fast path (schema init skipped); pragma must still be set.
+    let fk_on_second: i64 = with_connection(&cfg, |conn| {
+        Ok(conn.query_row("PRAGMA foreign_keys;", params![], |r| r.get::<_, i64>(0))?)
+    })
+    .unwrap();
+    assert_eq!(
+        fk_on_second, 1,
+        "foreign_keys must be ON on fast-path (post-init) connection"
+    );
+}
```
graycyrus, line 396: **[minor]** `is_transient_cold_start` is the gatekeeper for the retry loop but has no direct unit test. It's only exercised indirectly if a transient error happens to occur during the concurrent test (unlikely on most dev machines). A targeted test would catch regressions if the match arms are edited:

```rust
#[test]
fn is_transient_cold_start_classifies_known_codes() {
    use rusqlite::ffi;
    for code in [14, 1546, 4874] {
        let err = anyhow::anyhow!(rusqlite::Error::SqliteFailure(
            ffi::Error { code: ffi::ErrorCode::CannotOpen, extended_code: code },
            None,
        ));
        assert!(super::is_transient_cold_start(&err), "code {code} should be transient");
    }
    let non_transient = anyhow::anyhow!(rusqlite::Error::SqliteFailure(
        ffi::Error { code: ffi::ErrorCode::DatabaseBusy, extended_code: 5 },
        None,
    ));
    assert!(!super::is_transient_cold_start(&non_transient));
}
```

(The `ffi::ErrorCode` variant may need adjustment — the key thing is exercising the `extended_code` matching.)

graycyrus (changes requested):
## Walkthrough

Solid fix for a real production pain point — 19k Sentry events/day from concurrent WAL/SHM bootstrap races is nasty. The mutex-gated `SCHEMA_INITIALIZED` set with per-path keying is clean, the retry backoff on the three known transient codes is well-targeted, and the FK pragma move to `open_connection()` (addressing CodeRabbit's earlier catch) is correct. Two issues worth addressing before merge: test instrumentation leaking into production, and the `pub` visibility escape on `with_connection`.

## Change Summary

| `Cargo.toml` | Modified | Added `[[bin]]` entry for `memory-tree-init-smoke` manual diagnostic tool |
| `src/bin/memory_tree_init_smoke.rs` | Added | Manual stress-test binary: N threads racing `with_connection` on cold workspace |
| `src/openhuman/memory/tree/store.rs` | Modified | Core fix: `SCHEMA_INITIALIZED` mutex guard, retry loop with backoff, `open_connection()` extracted with per-connection FK pragma, `apply_schema()` split out |
| `src/openhuman/memory/tree/store_tests.rs` | Modified | Two regression tests: concurrent init serialisation + FK pragma persistence |

src/openhuman/memory/tree/store.rs
oxoxDev: Done in 883b295d. Static + write side both gated behind cfg(test). record_schema_apply still exists in release builds but the inner block compiles to a no-op — zero per-process overhead.

src/openhuman/memory/tree/store.rs
oxoxDev: Done in b9fb859a — option 1 (#[doc(hidden)] pub). Smoke bin still resolves the symbol; rustdoc drops it; comment explicitly states it is not a stable API.

src/openhuman/memory/tree/store.rs
oxoxDev: Done in 54a104ee. Named SQLITE_CANTOPEN / SQLITE_IOERR_TRUNCATE / SQLITE_IOERR_SHMMAP constants with doc comments. match arm is now self-documenting.

src/openhuman/memory/tree/store_tests.rs
oxoxDev: Done in 54a104ee. Added is_transient_cold_start_classifies_known_extended_codes — pins all three transient codes, explicitly rejects SQLITE_BUSY and non-SQLite errors. Uses ErrorCode::SystemIoFailure (primary code 10 for IOERR_*) and ErrorCode::DatabaseBusy for the negative case.

src/openhuman/memory/tree/store.rs
graycyrus: Confirmed fixed — SCHEMA_APPLY_COUNTS static and HashMap import are now cfg(test), and record_schema_apply body is gated behind a cfg(test) block. Zero instrumentation overhead in production.

src/openhuman/memory/tree/store.rs
graycyrus: Confirmed fixed — #[doc(hidden)] pub fn with_connection with a doc comment explaining why it's pub. Clean solution.

src/openhuman/memory/tree/store.rs
graycyrus: Confirmed fixed — SQLITE_CANTOPEN, SQLITE_IOERR_TRUNCATE, SQLITE_IOERR_SHMMAP constants with doc comments. Match arm is self-documenting now.

src/openhuman/memory/tree/store_tests.rs
graycyrus: Confirmed fixed — is_transient_cold_start_classifies_known_extended_codes test covers all three transient codes plus negative cases (SQLITE_BUSY and non-SQLite errors). Good coverage.

graycyrus (commented):
## Continuation Review — Previous Requested Changes Addressed

All four findings from the previous review have been addressed in follow-up commits:

1. **`SCHEMA_APPLY_COUNTS` test instrumentation** (major) — `#[cfg(test)]`-gated in 883b295d. Static, HashMap import, and `record_schema_apply` body all compile to no-ops in production. ✓
2. **`with_connection` pub visibility** (major) — `#[doc(hidden)] pub` with explanatory doc comment in b9fb859a. Smoke binary resolves the symbol; rustdoc hides it. ✓
3. **Magic SQLite error codes** (minor) — Named constants `SQLITE_CANTOPEN` / `SQLITE_IOERR_TRUNCATE` / `SQLITE_IOERR_SHMMAP` with doc comments in 54a104ee. ✓
4. **Missing `is_transient_cold_start` unit test** (minor) — `is_transient_cold_start_classifies_known_extended_codes` added in 54a104ee, covering all three transient codes plus negative cases. ✓

## Additional Changes Since Last Review

| `src/openhuman/wallet/execution.rs` | Bug fix | Added `TEST_LOCK` guard in `quote_store_round_trips_and_expires` to prevent cross-test race with `reset_quote_store_for_tests()` |
| merge commit c7ad0bb4 | Merge | upstream/main into fix branch — no conflicts |

The wallet test fix follows the established `TEST_LOCK` pattern already used in that module (see surrounding code in `wallet/execution.rs`). Clean, no issues.

## Summary

Solid fix for a high-volume Sentry issue (~19k events/day). The mutex-gated init set, per-connection FK pragma, and retry-with-backoff are all well-structured. Test coverage is thorough — concurrent init, FK persistence across cold/warm paths, and classifier pin tests. Code follows existing patterns in `memory/tree/` (naming, error handling, logging prefixes). No new critical or major findings. All previously requested changes are resolved.

All four review threads have been resolved.

graycyrus:
@oxoxDev bro conflicts

oxoxDev:
@graycyrus resolved bro — merged upstream/main into branch, reconciled #2059 worker-race split with #1574 §7 legacy→sidecar migration. 600/600 memory_tree tests green. CI re-running.

Issue #2049, opened 2026-05-18
Mac Close button does not dismiss the app window


Al629176:
## Summary

On macOS, clicking the red Close button on the OpenHuman app window does not close or dismiss the application window.

## Problem

Users expect the standard macOS red Close button to close the OpenHuman window, or at minimum hide it if the app intentionally keeps running in the background. Instead, clicking the Close button leaves the app window open.

Impact: users cannot dismiss OpenHuman through standard macOS window controls and may need to force quit or use another workaround.

Steps to reproduce:
1. Open the OpenHuman desktop app on macOS.
2. Click the red Close button in the top-left window chrome.
3. Observe that the OpenHuman window remains open and the app does not close/dismiss.

Version / platform: OpenHuman desktop app on macOS, observed May 18, 2026 from user screenshot. Exact build to confirm.

## Solution (optional)

Inspect the Tauri/macOS window close handler and app lifecycle behavior. Decide whether the intended macOS behavior is to close the window, hide the app, or quit the app, then implement that behavior consistently. If close hides the app, provide a reliable quit path from the menu/tray.

## Acceptance criteria

- [ ] **Repro gone** — Clicking the macOS red Close button dismisses the OpenHuman window according to the intended app behavior.
- [ ] **Behavior defined** — The team documents whether red Close should close the window, hide the app, or quit the app on macOS.
- [ ] **No stuck window** — The visible window does not remain open after the close action unless a clear confirmation or blocked state is shown.
- [ ] **Quit path works** — Users have a reliable way to fully quit OpenHuman if Close hides the window.
- [ ] **Lifecycle safe** — Closing/hiding does not leave the core sidecar, audio capture, meeting state, or background tasks in an inconsistent state.
- [ ] **Regression safety** — Unit, integration, or E2E coverage verifies the macOS close-window behavior where testable.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)).

## Related

- User-provided screenshot from May 18, 2026 showing the macOS OpenHuman window controls while Close does not dismiss the app.

TowsifAhamed:
Is anyone working on this already? I'm keen to work on this issue

honor2030:
[...]

PR #2060, opened 2026-05-18
fix(tauri): hide macOS app on close button

honor2030:
## Summary
- Use app-level hide on macOS close requests so the red close button dismisses the visible CEF-backed app window instead of only targeting the webview proxy.
- Unhide the macOS app before the existing `show_main_window` focus/reopen sequence so Dock/tray/reopen can restore it.
- Leave Windows/Linux close behavior unchanged.

Refs #2049

## Verification
- `git submodule update --init --recursive app/src-tauri/vendor/tauri-cef app/src-tauri/vendor/tauri-plugin-notification`
- `RUSTC=$(rustup which --toolchain 1.93.0 rustc) GGML_NATIVE=OFF CARGO_TARGET_DIR=/Users/lee/Documents/Claude/Projects/10\ Work\ OS/projects/openhuman/app/src-tauri/target rustup run 1.93.0 cargo fmt --manifest-path app/src-tauri/Cargo.toml --all --check`
- `RUSTC=$(rustup which --toolchain 1.93.0 rustc) GGML_NATIVE=OFF CARGO_TARGET_DIR=/Users/lee/Documents/Claude/Projects/10\ Work\ OS/projects/openhuman/app/src-tauri/target rustup run 1.93.0 cargo check --manifest-path app/src-tauri/Cargo.toml`
- `git diff --check`

## Notes
- The Tauri/CEF check initially hit local disk pressure while building CEF in this temporary worktree; rerunning with the existing OpenHuman Tauri target directory completed successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS window display behavior to ensure proper application visibility when showing the main window.
  * Fixed macOS window close behavior to properly manage the application state and lifecycle.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2060?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -1148,6 +1148,14 @@ fn set_main_window_hidden(hide: bool) {
 }
 
 fn show_main_window(app: &AppHandle<AppRuntime>) -> Result<(), String> {
+    // On macOS the close handler hides the whole NSApplication via
+    // `AppHandle::hide()` so the CEF host and browser surface are dismissed
+    // together. Unhide the application before touching the main window, then
+    // continue with the normal show/focus sequence below.
+    #[cfg(target_os = "macos")]
+    app.show()
+        .map_err(|err| format!("failed to show application: {err}"))?;
+
     // On Windows: surface the OS top-level Chrome_WidgetWin_1 frame BEFORE
     // any Tauri lookups. After our close handler's SW_HIDE the runtime
     // briefly drops the `WebviewWindow` record for "main" (CEF treats the
@@ -2761,10 +2769,10 @@ pub fn run() {
             // window from tray click: main window not found`
             // (OPENHUMAN-TAURI-2X — 21 events, Windows only).
             //
-            // macOS uses `window.hide()` because the vendored CEF runtime
-            // routes that through `set_application_visibility(false)` at the
-            // NSApplication level (`tauri-runtime-cef/src/lib.rs:588`), which
-            // hides the CEF browser surface together with the host window.
+            // macOS uses app-level hide because the vendored CEF runtime only
+            // routes `AppHandle::hide()` through `set_application_visibility(false)`
+            // at the NSApplication level. `WebviewWindow::hide()` targets the
+            // CEF window proxy and can leave the visible app window behind.
             // Windows is handled in the separate arm below — see issue #1607.
             //
             // Linux is left out: `setup_tray` early-returns on Linux because
@@ -2780,8 +2788,8 @@ pub fn run() {
                     "[window] close requested on main window — hiding instead of destroying"
                 );
                 api.prevent_close();
-                if let Some(window) = app_handle.get_webview_window("main") {
-                    let _ = window.hide();
+                if let Err(err) = app_handle.hide() {
+                    log::warn!("[macos] failed to hide application on main window close: {err}");
                 }
             }
             // Windows: full hide-to-tray.


app/src-tauri/src/lib.rs
```diff
@@ -1148,6 +1148,14 @@ fn set_main_window_hidden(hide: bool) {
 }
 
 fn show_main_window(app: &AppHandle<AppRuntime>) -> Result<(), String> {
+    // On macOS the close handler hides the whole NSApplication via
+    // `AppHandle::hide()` so the CEF host and browser surface are dismissed
+    // together. Unhide the application before touching the main window, then
```
graycyrus, line 1153: **[minor]** `app.show()` operates at the NSApplication level, which unhides all windows — not just main. Currently the app is single-window so this is fine, but if multi-window support is ever added, this would need to be scoped to the specific window. A brief inline comment noting this would be helpful for future maintainers.

app/src-tauri/src/lib.rs
```diff
@@ -2780,8 +2788,8 @@ pub fn run() {
                     "[window] close requested on main window — hiding instead of destroying"
                 );
                 api.prevent_close();
-                if let Some(window) = app_handle.get_webview_window("main") {
-                    let _ = window.hide();
+                if let Err(err) = app_handle.hide() {
```
graycyrus, line 2791: **[minor]** Good improvement over the old `let _ = window.hide()` — surfacing the error via `log::warn!` with the `[macos]` prefix is the right call for a close handler (non-fatal, best-effort). Matches the error handling pattern used in other event handlers in this file.

graycyrus (commented):
## Walkthrough

This PR fixes the macOS red close button not dismissing the app window by switching from `WebviewWindow::hide()` (which only targets the CEF window proxy) to `AppHandle::hide()` (which routes through `set_application_visibility(false)` at the NSApplication level). It also adds a symmetric `app.show()` call in `show_main_window` so Dock/tray/reopen can restore the app. The change is surgical, well-commented, and correctly scoped to macOS only.

Overall this is a clean fix. The root cause analysis is solid — the asymmetry between close (window-level) and show (app-level) was the bug, and this makes them symmetric. Error handling follows the established patterns in each context.

## Change Summary

| `app/src-tauri/src/lib.rs` | Modified | Close handler: `window.hide()` → `app_handle.hide()` with `log::warn!` on failure. Show handler: added `app.show()` before existing show/focus sequence. Both macOS-gated. |

## Per-file Analysis

### `app/src-tauri/src/lib.rs`

**Close handler (lines 2788-2793)**: Good change. The old code did `let _ = window.hide()` which silently swallowed errors. The new code uses `log::warn!` with the `[macos]` prefix, which is actually an improvement in observability while fixing the core bug. The `if let Err(err)` pattern is clean.

**Show handler (lines 1150-1166)**: The `app.show().map_err(...)? ` pattern correctly matches the existing error handling style in `show_main_window` (which returns `Result<(), String>`). The `#[cfg(target_os = "macos")]` gate ensures zero impact on other platforms. Comment is helpful and explains the *why*.

**Pattern conformance**: Both changes follow the surrounding code conventions — `[macos]` log prefix, `map_err` + `?` in Result-returning functions, `log::warn!` in event handlers, `#[cfg(target_os)]` guards matching existing platform blocks. No deviations.

## Findings

**No critical or major issues.** Two minor observations:

1. **[minor]** Future multi-window consideration: `app.show()` unhides the entire NSApplication, which would show *all* hidden windows if the app ever gains multiple windows. Currently single-window so this is a non-issue, but worth a one-line comment like `// NOTE: app-level show — unhides all windows` if multi-window is ever planned.

2. **[minor]** Test coverage: Issue #2049 AC calls for regression safety and ≥80% diff coverage. No tests are added, though the project has zero existing window lifecycle tests (`window_state.rs` is also untested) and this is inherently hard to unit test without a running macOS app + CEF runtime. If CI has a coverage gate that enforces this, you may need a `#[cfg(test)]` stub or integration test. If it's advisory-only, this is fine as-is.

Nice work — the symmetry fix (`app.hide()` ↔ `app.show()`) is the right approach, and the updated comments clearly explain the CEF proxy vs NSApplication distinction.

M3gA-Mind:
Hey @honor2030 , thank you soo much for taking the time to contribute and submitting this PR. We really appreciate the effort. This issue has already been resolved on my end, so I’ll be closing this PR as it duplicates the existing fix. Looking forward to your future contributions!

Issue #1710, opened 2026-05-14
Prioritize fully local speech and Composer operation


Al629176:
## Summary

Prioritize fully local OpenHuman operation for speech-to-text, voice, and Composer by evaluating Whisper, Deepgram, and the local stack integration path.

## Problem / Context

Users are asking to “run things locally,” especially speech-to-text and other AI features. OpenHuman needs an engineering-led plan and implementation path for fully local operation across speech-to-text, voice, inference, and Composer-related workflows.

Whisper is the known local/offline candidate for speech-to-text. Deepgram has also been raised as an audio/STT vendor to evaluate, but it is not yet specified as local or cloud for this use case. The team needs to figure out what can ship, what should remain cloud-backed, and how local/offline capability fits into the rest of the OpenHuman stack.

Constraints:
- The goal is a practical shippable path, not only a research document.
- Local/offline mode must be explicit and must not silently call cloud providers.
- Composer behavior needs clear boundaries: what can run locally, what still requires backend/cloud, and what degrades gracefully.
- Hardware support, model size, startup time, and quality tradeoffs must be documented.

## Scope (optional)

Evaluate and implement the first shippable slice of local operation:
- Speech-to-text with Whisper or another local candidate.
- Voice/TTS compatibility with the local voice workstream.
- Composer workflows that can run locally or degrade clearly.
- Comparison/evaluation of Deepgram as a possible cloud/vendor alternative for STT, with privacy and offline limitations documented.

## Acceptance criteria

- [ ] **Local operation plan** — A concrete implementation plan documents what will run locally for STT, voice, inference, and Composer, plus known gaps.
- [ ] **Whisper evaluated** — Whisper is evaluated as the local/offline STT candidate with install path, model choices, latency, quality, and hardware requirements documented.
- [ ] **Deepgram evaluated** — Deepgram is evaluated as an STT/audio option, including whether it is cloud-only for this use case and how it compares with the local Whisper path.
- [ ] **First local STT slice shipped** — A supported local STT path can transcribe speech without network access once required assets are installed.
- [ ] **Composer boundary defined** — Composer workflows clearly indicate which parts can run locally, which require backend/cloud, and what fallback/error state users see.
- [ ] **Offline mode enforced** — In local/offline mode, OpenHuman does not silently call cloud STT, voice, inference, or Composer services.
- [ ] **User settings documented** — Users can understand and configure local vs cloud behavior for the supported speech/AI paths.
[...]

PR #2069, opened 2026-05-18
fix(core): prevent SIGBUS stack overflow in composio tool path

senamakel:
## Summary

- Fix `EXC_BAD_ACCESS (SIGBUS)` in the in-process core when a chat-driven Gmail (or other composio) tool call runs through `delegate_to_integrations_agent → integrations_agent → composio_list_tools → load_config_with_timeout`. The TOML parse for `Config` blew the 2 MB tokio worker stack guard page on top of the ~50-frame agent tower.
- Move `toml::from_str::<Config>` onto the tokio blocking pool via `spawn_blocking` so the parser runs on a fresh thread stack.
- Install a custom `tauri::async_runtime` with `thread_stack_size(8 MiB)` at the top of `pub fn run()` so the in-process core's `tokio::spawn(run_server_embedded(..))` inherits 4× the prior headroom for everything else in the tower.
- Add a structural regression test that drives the full sub-agent → `composio_list_tools` → `load_config_with_timeout` chain on a production-shaped 2 MB worker.

## Problem

User report: the Tauri window quits without warning when sending a Gmail-related agent request. macOS crash report (`crahs.log`):

```
Triggered by Thread: 46  tokio-rt-worker
Exception Type:   EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x3028534f0
Exception Message: Could not determine thread index for stack guard region
```

The address falls inside the stack guard page of a 2 MB tokio worker. Stack frames at the top of the crashed thread were `toml_parser::on_array_open → value → … → toml::from_str → Config::load_or_init → load_config_with_timeout → ComposioListToolsTool::execute → subagent_runner::run_inner_loop → SkillDelegationTool::execute → Agent::execute_tools → web channel run_chat_task`. About 100 frames total, with the toml parser's serde-monomorphised `Visitor` frames for the deeply-nested `Config` (KB-sized per frame) tipping the worker over.

`composio_list_tools` reloads `Config` from disk on every invocation (per #1710 Wave 4, so a mid-session `composio.mode` toggle is observed). That's the trigger; the underlying issue is that the toml parser + agent tower together no longer fit in a 2 MB worker stack.

## Solution

Two complementary changes:

1. **`src/openhuman/config/schema/load.rs`** — wrap `toml::from_str::<Config>` in `tokio::task::spawn_blocking` via new helper `parse_toml_off_worker`. The blocking pool thread starts fresh (no async tower above the parse), so the serde `Visitor` frames no longer compound with the caller's frames. `parse_config_with_recovery` now routes both the primary parse and the backup-recovery parse through the helper.

2. **`app/src-tauri/src/lib.rs`** — at the very top of `pub fn run()`, build a custom `tokio::runtime::Builder::new_multi_thread().thread_stack_size(8 * 1024 * 1024)` runtime and `tauri::async_runtime::set(handle)` it before any other tauri call. The in-process core runs via `tokio::spawn(run_server_embedded(..))` on this runtime, so every JSON-RPC handler gets 4× the prior worker stack headroom. The runtime is leaked (per Tauri's contract: "you cannot drop the underlying TokioRuntime").

Tried but reverted: an `Arc<Config>` cache fronting `load_config_with_timeout` that would skip the per-call parse entirely. Worked correctness-wise in lib unit tests but caused 6 flakes in `tests/json_rpc_e2e.rs` (in-process JSON-RPC servers loading config mid-mutation, race-prone even with mtime checks). The two changes above carry the production stack-overflow fix without it.

Regression test (`tests/composio_list_tools_stack_overflow_regression.rs`) is a structural guard: it drives `run_subagent(integrations_agent) → composio_list_tools → load_config_with_timeout` on a production-shaped 2 MB worker with a stubbed `Provider`. Module docs explain what it does and does not catch (we can't easily mock the upper chat-channel layers in cargo-test, so the bare path fits in 2 MB even without the parser-move; the test catches future structural regressions in the path).

## Submission Checklist

- [x] N/A: behaviour-only fix; no new product surface to measure diff coverage on. `tests/composio_list_tools_stack_overflow_regression.rs` exercises the changed path; `pnpm test:rust` is clean (45 integration + 7602 lib tests + new regression).
- [x] N/A: behaviour-only change — no feature rows added/removed/renamed.
- [x] N/A: no feature IDs touched.
- [x] N/A: change does not touch release-cut surfaces (in-process core stack budget + a regression test).
- [x] N/A: no linked issue — surfaced from a user-supplied macOS crash dump (`crahs.log`).

## Impact

- **Runtime/platform**: desktop only. Affects the in-process Rust core inside the Tauri shell. 8 MiB worker stacks × ~CPU-count workers raises virtual-address reservation modestly (real memory only fault-paged in when touched). Matches the macOS pthread main-thread default and is below Linux/Windows defaults for non-tokio threads, so no platform reaches a hard limit.
- **Performance**: per-call config reload now hops to the blocking pool (one `spawn_blocking` round-trip). Imperceptible — the parse is microseconds and already ran on tokio time.
- **Security**: none.
- **Compatibility**: no API changes. Internal-only.

## Related

- Closes:
- Follow-up PR(s)/TODOs: a per-call config cache would eliminate the per-tool parse entirely (the most efficient fix) but needs care around in-process integration tests that swap workspace paths mid-flight; tracked as a follow-up.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: \`fix/composio-stack-overflow\`
- Commit SHA: \`9a311b608\`

### Validation Run
- [x] \`pnpm --filter openhuman-app format:check\` (ran via pre-push hook; auto-fixed and re-committed)
- [x] N/A: no TS changes.
- [x] Focused tests: \`cargo test --test composio_list_tools_stack_overflow_regression\` (1 passed), \`cargo test --lib\` (7602 passed; transient timing flake on `runtime_dispatch::message_dispatch_processes_messages_in_parallel` cleared on re-run, also flaked on baseline without changes — unrelated), \`cargo test --test json_rpc_e2e\` (45 passed)
- [x] Rust fmt/check (if changed): \`cargo check --bin openhuman-core\` clean
- [x] Tauri fmt/check (if changed): \`cargo check --manifest-path app/src-tauri/Cargo.toml\` clean

### Validation Blocked
- \`command:\` N/A
- \`error:\` N/A
- \`impact:\` N/A

### Behavior Changes
- Intended behavior change: composio tool calls no longer crash the in-process core under deep async towers.
- User-visible effect: Tauri window stays alive when an agent sub-call uses `composio_list_tools` / per-action composio tools. No surface or API change.

### Parity Contract
- Legacy behavior preserved: `load_config_with_timeout` semantics unchanged; the per-call reload behavior added in #1710 Wave 4 is retained.
- Guard/fallback/dispatch parity checks: `parse_config_with_recovery` still falls back to `.bak` recovery; new `parse_toml_off_worker` returns errors via the same `(Config, was_corrupted)` channel as the original inline parse.

- Duplicate PR(s): N/A
- Canonical PR: N/A
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved stack-overflow crashes during configuration loading and improved parsing stability under heavy workloads.
  * Made async runtime startup more robust to prevent runtime aborts.

* **Documentation**
  * Clarified config-loading timeout and recovery behavior in docs.

* **Tests**
  * Added a regression test to ensure stack-overflow issues do not recur.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2069?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -1623,6 +1623,42 @@ fn append_platform_cef_gpu_workarounds(args: &mut Vec<CefCommandLineArg>, os: &s
 }
 
 pub fn run() {
+    // ── Install a custom tokio runtime for tauri::async_runtime ─────────
+    //
+    // Tauri's default async runtime uses tokio multi-thread workers with
+    // a ~2 MB stack. The in-process core (spawned by
+    // `core_process::CoreProcessHandle::ensure_running` via
+    // `tokio::spawn(run_server_embedded(..))`) runs *on* that runtime, so
+    // every JSON-RPC handler — including the deep tower
+    // `web channel chat → orchestrator turn → delegate_to_integrations_agent
+    // → sub-agent → composio_list_tools → load_config_with_timeout` —
+    // burns through the same 2 MB. In `crahs.log` (2026-05-17, build
+    // 0.53.49) that tower plus the serde-monomorphised `Config` Visitor
+    // frames pushed past the guard page and aborted with
+    // `SIGBUS / KERN_PROTECTION_FAILURE`. The structural fix
+    // (`spawn_blocking` for the TOML parse + cache in
+    // `src/openhuman/config/{schema/load.rs, ops.rs}`) moves the
+    // largest contributor off the worker; bumping the worker stack
+    // itself gives the rest of the tower comfortable headroom so future
+    // additions don't immediately re-tip the same scale. 8 MiB matches
+    // the OS-default pthread main-thread stack on macOS, so we can
+    // assume "as much room as the main thread" everywhere.
+    //
+    // Must happen before any `tauri::async_runtime::*` call, otherwise
+    // `set(...)` panics with "runtime already initialized".
+    {
+        let custom_runtime = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .thread_stack_size(8 * 1024 * 1024)
+            .build()
+            .expect("build custom tokio runtime for tauri async surface");
+        let handle = custom_runtime.handle().clone();
+        // Tauri docs: "you cannot drop the underlying TokioRuntime."
+        // Leak it so its lifetime matches the process.
+        std::mem::forget(custom_runtime);
+        tauri::async_runtime::set(handle);
+    }
+
     // Initialize Sentry for the Tauri shell (desktop host) process before any
     // other startup work. Reads `OPENHUMAN_TAURI_SENTRY_DSN` at runtime first,
     // then falls back to the value baked in at compile time via the release
--- a/src/openhuman/config/ops.rs
+++ b/src/openhuman/config/ops.rs
@@ -29,6 +29,15 @@ const CONFIG_LOAD_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(
 ///
 /// This is used by JSON-RPC and CLI handlers to ensure they don't hang
 /// indefinitely if disk I/O is blocked.
+///
+/// The TOML parse itself runs on the blocking pool via
+/// `parse_config_with_recovery` (see `src/openhuman/config/schema/load.rs`)
+/// so the recursive-descent parser's serde Visitor frames don't compound
+/// with whatever deep async tower called us. That's the stack-overflow
+/// fix from `crahs.log` (2026-05-17); a per-call cache here would shave
+/// the disk read on hot paths but proved racy across the in-process
+/// integration tests (re-used workspace paths, concurrent server tasks
+/// loading mid-mutation), so it isn't worth it.
 pub async fn load_config_with_timeout() -> Result<Config, String> {
     match tokio::time::timeout(CONFIG_LOAD_TIMEOUT, Config::load_or_init()).await {
         Ok(Ok(mut config)) => {
--- a/src/openhuman/config/schema/load.rs
+++ b/src/openhuman/config/schema/load.rs
@@ -456,8 +456,24 @@ pub fn pre_login_user_dir(default_openhuman_dir: &Path) -> PathBuf {
 ///
 /// This is a standalone async function (not a method on Config) so it can be
 /// called from both `load_or_init` and `load_from_default_paths`.
+///
+/// **Why the parse runs via `spawn_blocking`:** `toml::from_str::<Config>`
+/// is a recursive-descent parser whose serde-monomorphised `Visitor`
+/// frames for our deeply-nested `Config` cost several KB each. When this
+/// function is called from the bottom of a deep async tower — e.g.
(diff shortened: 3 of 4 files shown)


graycyrus (commented):
## Walkthrough

Solid fix for the SIGBUS stack overflow in the composio tool path. The two-pronged approach — moving `toml::from_str::<Config>` onto the blocking pool via `spawn_blocking` and bumping the Tauri tokio worker stack to 8 MiB — addresses both the immediate trigger and gives future headroom. The regression test is well-structured and honestly documents its own limitations. Only one minor stale-docs nit.

## Change Summary

| `app/src-tauri/src/lib.rs` | Modified | Custom tokio runtime with 8 MiB worker stacks, set before any Tauri async call |
| `src/openhuman/config/ops.rs` | Modified | Doc comments explaining why `spawn_blocking` instead of caching |
| `src/openhuman/config/schema/load.rs` | Modified | New `parse_toml_off_worker` helper; `parse_config_with_recovery` uses it for both primary and backup parse paths |
| `tests/composio_list_tools_stack_overflow_regression.rs` | Added | Regression test driving `run_subagent → composio_list_tools → load_config_with_timeout` on a 2 MiB worker stack |

## Per-file notes

**`lib.rs`** — The `std::mem::forget` + `tauri::async_runtime::set` pattern is correct per Tauri's docs. Block comment thoroughly explains the rationale and the "must call before any other tauri async call" ordering constraint.

**`load.rs`** — `parse_toml_off_worker` is clean: takes ownership of the string (required for `'static` in `spawn_blocking`), flattens the join error into the same `String` error path, and callers don't need to change their error handling. Both the primary parse and the backup-file recovery path go through it.

**`ops.rs`** — Doc-only change, accurately describes why caching was abandoned.

**Test file** — Excellent module-level documentation explaining the crash, why existing tests missed it, and the honest caveat about what the test does/doesn't catch. Setup mirrors production constraints (2 MiB worker stack, representative config TOML, stub provider/memory).

## Findings

**[minor]** `tests/composio_list_tools_stack_overflow_regression.rs:387-390` — Stale doc comment says `load_config_with_timeout` "is fronted by a process-global cache keyed on `OPENHUMAN_WORKSPACE`, invalidated by `Config::save()`" and that hot-path consumers "get a clone, never re-entering the parser." But the PR description and the `ops.rs` doc update both confirm the cache was tried and **reverted** because it caused 6 flakes in `json_rpc_e2e.rs`. This block should be removed or rewritten to match reality (spawn_blocking only, no cache). Same class of stale-docs issue that CodeRabbit caught in `load.rs:474` (fixed in `cc141de`), but this instance in the test file was missed.

Issue #1900, opened 2026-05-16
Security: decrypt_optional_secret / encrypt_optional_secret never called — config secrets stored in plaintext


JasonOA888:
## Summary

The functions `decrypt_optional_secret` and `encrypt_optional_secret` in `src/openhuman/config/schema/load.rs` are defined to encrypt/decrypt sensitive config fields (channel tokens, API keys), but are **never called** from any production code path. This means all channel credentials in `config.toml` are stored and loaded as plaintext.

## Affected Files

- `src/openhuman/config/schema/load.rs:350-377` — functions defined but unused
- `src/openhuman/config/schema/channels.rs` — channel tokens stored as `pub bot_token: String` (plaintext)
- `src/openhuman/credentials/profiles.rs:511` — auth profiles correctly use encryption (different code path)

## Details

The encryption infrastructure is complete and working:
- `SecretStore` uses ChaCha20-Poly1305 AEAD with proper key management
- Auth credentials in `credentials/profiles.rs` correctly encrypt via `encrypt_optional()`
- But the config schema load/save path has dead code that was intended to encrypt channel tokens

```rust
// Defined but never called from production code:
fn decrypt_optional_secret(store: &SecretStore, value: &mut Option<String>, field_name: &str) -> Result<()>
fn encrypt_optional_secret(store: &SecretStore, value: &mut Option<String>, field_name: &str) -> Result<()>
```

Affected plaintext fields in config.toml:
- `[telegram]` → `bot_token`
- `[slack]` → `bot_token`
- `[discord]` → `bot_token`
- `[irc]` → `server_password`, `nickserv_password`, `sasl_password`
- `[wechat]` → `app_secret`
- `[google_chat]` → `client_secret`
- `[mattermost]` → `app_secret`
- etc.

## Recommended Fix

Wire `decrypt_optional_secret` / `encrypt_optional_secret` into the config load and save paths for all `Option<String>` fields containing tokens/secrets. The infrastructure is already in place — it just needs to be connected.

## Severity

**Medium** — Requires local file read access to `config.toml`, but this file contains high-value credentials for all configured messaging channels.

---

*Discovered during independent security audit of OpenHuman v0.53.47.*

Issue #1339, opened 2026-05-07
Keep triggers read-only until users approve writes


Al629176:
## Summary

Build a user-consent flow so proactive or background agent actions run read-only by default and execute write-side effects only after explicit user approval. The user-visible outcome is an approval surface for actions such as external notifications, meeting-note sharing, Slack posts, email sends/replies, and other mutations.

## Problem

OpenHuman is giving write-capable access to proactive agent paths, but the safer default should be read-only. All triggers should collect context, analyze, draft, and recommend in read-only mode; they should not send, post, notify externally, publish meeting notes, or mutate user data until the user approves the exact action.

Current repo shape:
- `src/openhuman/approval/` contains a CLI-oriented supervised approval manager for tool calls.
- `docs/subconscious.md` describes an "Approval Needed" escalation flow, but `src/openhuman/subconscious/executor.rs` currently runs write-intent tasks with full permissions when the task text matches action verbs like send/post/create/schedule.
- `src/openhuman/channels/proactive.rs` can deliver proactive messages to web and an active external channel.
- `src/openhuman/composio/tools.rs` treats `composio_execute` as write-level and enforces sandbox/scope rules, but does not provide a user-facing one-time approval gate for proactive Slack/email actions.
- Notification storage/UI exists under `src/openhuman/notifications/` and `app/src/components/notifications/`, but approval state is not the same thing as notification read/dismiss/acted state.

This hurts trust and privacy: an agent may be able to notify, post, send, or write notes from a background/proactive path without a clear last-mile confirmation. The product needs a single consent contract across desktop and core so proactive actions cannot bypass user approval through a lower-level tool or channel path.

Constraints:
- Desktop app first; core must own the enforcement so UI omissions cannot bypass consent.
- All triggers and proactive/background runs should be read-only unless they hold a scoped approval for the exact write action.
- Avoid logging secrets or full sensitive payloads in approval records.
- Preserve read-only proactive analysis and draft generation where no external side effect occurs.
- Previously configured automations should count as consent only when the stored configuration clearly captures the approved action class, destination, payload pattern, and recurrence.

## Solution (optional)

Scope is both core and app.

Core should introduce a durable approval/request model for proactive write-side effects, then require a positive approval token before executing external notifications, Slack/email/Composio write actions, meeting-note publication, or other proactive mutations. App should surface pending approvals with enough context for the user to approve or deny, and core should record the decision/audit trail.

[...]

PR #2145, opened 2026-05-18
feat(integrations): Polymarket trading + market data (#1398, venue 1/3)

### LINKED ISSUE #1398: Add Polymarket, Kalshi, and Hyperliquid trading integrations
## Summary

Add explicit trading / market integration tools for Polymarket, Kalshi, and Hyperliquid so OpenHuman can perform market-aware crypto workflows through supported external venues instead of only raw wallet actions.

## Problem / Context

Wallet access alone is not enough for the product direction implied by the current `ConnectionsPanel` and capability surfaces. Users will also expect the agent to interact with specific trading and prediction-market venues.

The repo currently has:

- A general built-in `integrations_agent` aimed at Composio-backed SaaS toolkits.
- No exchange / prediction-market domain in the Rust core.
- No tool surface for venue-specific auth, balances, positions, orders, market discovery, or settlement workflows.

Polymarket, Kalshi, and Hyperliquid each carry different auth, market-model, and execution semantics, so they should be scoped explicitly rather than folded into vague “exchange support”.

This task should attach to umbrella issue #1394.

## Scope (optional)

In scope:

- Define the integration architecture for Polymarket, Kalshi, and Hyperliquid.
- Add the minimum tool / controller surfaces for account status, balances, positions, market lookup, and order placement/cancellation where allowed.
- Align the UX entry points with `app/src/components/settings/panels/ConnectionsPanel.tsx` and capability-catalog messaging.
- Add test/mocking strategy for venue-specific APIs and failure paths.

Out of scope:

- Generic wallet send/swap/contract tools that are not venue-specific.
- A broad broker / exchange abstraction covering every future provider.
- Autonomous continuous trading loops unless separately specified.

Implementation notes:

- Be explicit about what is wallet-signed vs API-key / venue-authenticated.
- Keep provider-specific behavior in dedicated modules; avoid a monolithic “crypto exchange” blob.
- Where possible, reuse existing integration / agent patterns, but do not force these venues through Composio if that weakens type safety or execution clarity.

## Acceptance criteria

- [ ] **Venue architecture defined** — Polymarket, Kalshi, and Hyperliquid each have a clear integration seam in the app/core architecture and issue-level implementation plan.
- [ ] **Tool surface added** — account, balance, position, market, and order actions are exposed through explicit controllers / tools for supported venues.
- [ ] **Connection UX wired** — Settings connections and capability-catalog surfaces reflect the real supported trading venues instead of generic coming-soon placeholders.
- [ ] **Failure-path design** — auth failures, market lookup failures, rejected orders, and rate limits have deterministic handling and test coverage.
- [ ] **Agent compatibility** — the resulting tools are usable by the dedicated crypto/trading agent without broadening its permissions unnecessarily.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)) when code changes are involved.

## Related

- Parent: #1394
- Depends on: #1397
- `app/src/components/settings/panels/ConnectionsPanel.tsx`
- `src/openhuman/about_app/catalog.rs`
- `src/openhuman/agent/agents/integrations_agent/`

oxoxDev:
## Summary

- Adds the **Polymarket** integration (read + trading) — the first of three venues called out in #1398 (Kalshi + Hyperliquid follow as separate PRs on the same `#1398` umbrella).
- Read surface: market / event / orderbook / price discovery via Gamma + CLOB public APIs, plus authenticated reads (positions, balance, open orders, USDC.e allowance) once an EOA is configured.
- Trading surface: `place_order` + `cancel_order` against the Polymarket CLOB with EIP-712 typed-data signing on Polygon (chain_id 137).
- Approval gate: every write action requires `arguments.approved == true`. Stopgap until #1339's formal approval gate lands — `TODO(#1339)` recorded inline.
- Capability catalog split: `skills.polymarket_readonly` (browse) + `skills.polymarket_trading` (Beta, Action, destination = "Polymarket CLOB API").

## Problem

Per #1398, Polymarket, Kalshi, and Hyperliquid each need explicit tool surfaces rather than being folded into vague "exchange support". Today the repo has no exchange / prediction-market domain in the core, no venue-specific auth or signing primitives, and no tool surface for balances, positions, market discovery, or order flow. Without that, the agent can hold a wallet but cannot reason over a real trading venue.

## Solution

**New tool module** `src/openhuman/tools/impl/network/polymarket.rs` — single `PolymarketTool` exposing 11 actions tagged by `action` enum:

| Surface | Actions |
|---|---|
| Public reads | `list_markets`, `get_market`, `list_events`, `get_orderbook`, `get_price` |
| Authenticated reads | `get_positions`, `get_balance`, `get_open_orders`, `get_usdc_allowance` |
| Writes | `place_order`, `cancel_order` |

**Signing primitives**:
- `src/openhuman/tools/impl/network/clob_auth.rs` — L1 handshake (POST `/auth/api-key` w/ fallback `/auth/derive-api-key`) + L2 HMAC-SHA256 header signer over `timestamp + method + path + body`. Timestamp in unix seconds.
- `src/openhuman/tools/impl/network/polymarket_orders.rs` — EIP-712 typed-data `Order` struct + `domain_separator` (chain_id = **137**, Polymarket CLOB Exchange contract) + `sign_order` returning hex signature.

**Polygon RPC wiring** kept narrow: `polygon_rpc_url` lives in `PolymarketConfig` (no new `WalletChain::Polygon` variant) so the blast radius stays inside the tool module. USDC.e allowance reads go through `eth_call` over that RPC.

**Config** (`src/openhuman/config/schema/tools.rs`) — additive fields:
- `eoa_address: Option<String>`
- `polygon_rpc_url: String` (default `https://polygon-rpc.com`)
- `usdc_contract: String` (default `0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174` — USDC.e bridged on Polygon, not native USDC)
- `clob_exchange_contract: String` (default `0x4bFb41d5B3570DeFd03C39a9A4D8dE6Bd8B8982E`)
- `derived_clob_credentials: Option<PolymarketClobCredentials>` (persisted across runs; will migrate to SecretStore once #1900 lands)

**Approval gate**: `require_write_approval(approved: Option<bool>)` runs before any write action issues HTTP. If `approved != true`, returns `Polymarket write requires explicit user approval. Re-invoke with arguments.approved = true after confirming with the user.` — no HTTP touched. Approval-gate tests verify the no-HTTP property.

**Retry / timeout**: 3 × 500ms backoff on 429 / 5xx, no retry on 4xx, explicit deadline error on user-configured `timeout_secs` (default 15s). Same shape as siblings to be added by the Kalshi + Hyperliquid PRs.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — 22 polymarket lib tests cover: 11-action schema deserialization; happy paths for `list_markets`, `get_market` (id + slug), `list_events`, `get_orderbook`, `get_price`, `get_positions`, `get_balance` (USDC.e default), `get_open_orders`, `get_usdc_allowance` (Polygon `eth_call`), `place_order` (signed POST); approval gate (`place_order` and `cancel_order` no-HTTP on missing approval); HMAC signing fixture (`clob_auth`); EIP-712 fixture (`polymarket_orders`, chain_id 137); side normalisation; 4xx / 5xx / timeout. Per [`feedback_cargo_test_lib_full_run.md`](https://github.com/oxoxDev/openhuman) the full-lib run OOMs locally; CI gates the full matrix.
- [x] N/A: coverage matrix — `docs/TEST-COVERAGE-MATRIX.md` does not have a Polymarket / trading-venue row yet; the matrix entry will land in the venue-3 PR alongside Hyperliquid + Kalshi rows so the row covers the whole `#1398` umbrella in one edit.
- [x] No new external network dependencies introduced — `polymarket_tests.rs` uses an in-process `TcpListener` mock; Gamma + CLOB endpoints are configurable so smoke can target a local mock per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy).
- [x] N/A: release manual smoke — Polymarket trading is feature-flagged via `integrations.polymarket.enabled` and not part of the release-cut surface yet.
- [x] N/A: `Closes #1398` — #1398 covers three venues; this PR ships venue 1/3. Closing reference will land on the venue-3 PR. See `## Related`.

## Impact

- **Runtime**: desktop, opt-in via `integrations.polymarket.enabled` config flag. Off-by-default. No behavioural change for users who don't touch the flag.
- **Network**: Gamma (`gamma-api.polymarket.com`), CLOB (`clob.polymarket.com`), and a Polygon RPC (default `polygon-rpc.com`). All configurable.
- **Security**: writes are gated by the `approved=true` stopgap and an EIP-712 typed-data signature over the order on chain_id 137. L1 handshake derives a per-EOA L2 API key + secret which lives in config (will move to SecretStore after #1900). No raw wallet privkey leaves the wallet module.
- **Performance**: cold-path retry adds 3 × 500ms on `5xx` / `429`; hot path one HTTP per action. Nonce fetched per submission via CLOB `/nonce?user=` — extra round-trip on each `place_order` (documented; caching deferred).
- **Compatibility**: additive — new RPC actions on the existing `tools_polymarket_execute` controller. No public RPC surface change beyond the action enum widening.

## Related

- Refs #1398 (Polymarket — venue 1/3). The umbrella issue stays open until Kalshi + Hyperliquid PRs land.
- Closes #1398: **NO** — partial scope. Follow-up PRs will close it.
- Follow-up PR(s)/TODOs:
  - Kalshi (venue 2/3) — same approval-flag stopgap; queued.
  - Hyperliquid (venue 3/3) — closes `#1398` on merge; queued.
  - Migrate `derived_clob_credentials` storage to SecretStore once #1900 lands.
  - Replace `approved=true` stopgap with formal approval gate once #1339 lands.

---

### Linear Issue
- Key: N/A (GitHub-only issue)
- URL: https://github.com/tinyhumansai/openhuman/issues/1398

### Commit & Branch
- Branch: `feat/1398-polymarket`
- Commit SHA: top = `23176452` (test/docs), 12 commits ahead of base + merge of upstream/main

### Validation Run
- [x] `cargo check`
- [x] `cargo test --lib polymarket` — 22 passed
- [x] Manual smoke: local TcpListener mock on :8901 with `[integrations.polymarket] gamma_base_url + clob_base_url` overridden in user config — `list_markets` + `get_market` action returned mock fixtures via `openhuman.tools_polymarket_execute`.
- [x] `pnpm typecheck` — N/A (no frontend changes in this PR)

### Validation Blocked
- `command:` `cargo test --lib` (full lib run)
- `error:` SIGKILL'd locally per `feedback_cargo_test_lib_full_run.md`
- `impact:` CI gates the full matrix; scope-filtered run passes 22/22

### Behavior Changes
- Intended: add 11-action Polymarket tool gated behind `integrations.polymarket.enabled`; trading writes gated by `approved=true`
- User-visible: agent can browse Polymarket markets / events / orderbooks / prices, read EOA-owned positions / balance / open orders, and (with explicit approval) place / cancel limit orders

### Parity Contract
- Legacy behavior preserved: integrations disabled by default; existing read-only landing from earlier on this branch unchanged for callers that only use the public-data surface
- Guard/fallback/dispatch parity checks: `tools_polymarket_execute` already routes by `action` field; new actions auto-reachable without schema regen. Approval gate is a function-local check, not a separate middleware — migration to `ApprovalGate` (#1339) will replace `require_write_approval` call sites.

## Summary by CodeRabbit

* **New Features**
  * Polymarket integration (Beta): market discovery, orderbook/price viewing, authenticated balance/position reads, USDC allowance checks, and order placement/cancellation (writes require explicit user approval). Config toggle (defaults off), credential derivation/persistence, and retry/timeout semantics.

* **Documentation**
  * New integration guide and table-of-contents entry covering setup, auth/signing, headers/nonce rules, permissions, and error/retry behavior.

* **Tests**
  * Extensive fixtures and end-to-end/unit tests covering discovery, CLOB reads, signed requests, write flows, allowance checks, and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2145?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gitbooks/SUMMARY.md
```diff
@@ -43,6 +43,7 @@
 * [Testing Strategy](developing/testing-strategy.md)
 * [E2E Testing](developing/e2e-testing.md)
 * [Release Policy](developing/release-policy.md)
+* [Polymarket Integration (v1 Read-Only)](developing/integrations/polymarket.md)
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 23176452 — flipped SUMMARY.md entry to Read + Trading to match the doc title.

src/openhuman/config/schema/tools.rs
```diff
@@ -554,6 +554,77 @@ impl Default for IntegrationToggle {
...
+            clob_base_url: default_polymarket_clob_base_url(),
+            timeout_secs: default_polymarket_timeout_secs(),
+            eoa_address: None,
+            polygon_rpc_url: default_polymarket_polygon_rpc_url(),
+            usdc_contract: default_polymarket_usdc_contract(),
+            clob_exchange_contract: default_polymarket_clob_exchange_contract(),
+            derived_clob_credentials: None,
+        }
+    }
+}
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 23176452 — flipped PolymarketConfig::enabled default to false in both the serde annotation and the Default impl, and updated polymarket.md to say (default false). Integration is now off by default as the PR description states.

src/openhuman/tools/schemas.rs
```diff
@@ -468,18 +500,61 @@ fn handle_apify_linkedin_scrape(params: Map<String, Value>) -> ControllerFuture
...
+            &config.integrations.polymarket,
+            security,
+        );
+
+        let mut args = arguments;
+        if let Value::Object(ref mut map) = args {
+            map.insert("action".to_string(), Value::String(action.clone()));
+        } else {
+            args = json!({ "action": action });
+        }
```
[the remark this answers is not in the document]
oxoxDev: Already addressed in current code — the diff_hunk on this comment is from an earlier revision. In the latest revision (commit 23176452, lines 533-542 in src/openhuman/tools/schemas.rs) handle_polymarket_execute returns the validation error 'arguments must be a JSON object when provided' for non-object arguments instead of silently replacing them.

src/openhuman/tools/schemas.rs
```diff
@@ -468,18 +500,61 @@ fn handle_apify_linkedin_scrape(params: Map<String, Value>) -> ControllerFuture
...
+        let config = config_rpc::load_config_with_timeout().await?;
+        if !config.integrations.polymarket.enabled {
+            return Err("Polymarket integration is disabled in config.".to_string());
+        }
+
+        let security = std::sync::Arc::new(crate::openhuman::security::SecurityPolicy::default());
+        let tool = crate::openhuman::tools::implementations::network::PolymarketTool::new(
+            &config.integrations.polymarket,
+            security,
+        );
```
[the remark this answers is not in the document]
oxoxDev: Deferring as out-of-scope for this PR. Resolving this properly requires plumbing a session-scoped Arc<SecurityPolicy> through the ControllerFuture signature, which would also have to be done for every other network handler in this file (handle_apify_*, handle_brave_*, etc. — all currently use SecurityPolicy::default() the same way). That's a cross-cutting refactor of the entire network-controller surface, not a Polymarket-specific change. The PolymarketTool itself accepts an Arc<SecurityPolicy> in its constructor and uses it; the only thing missing is the upstream plumbing. Filing a follow-up issue to thread session security context through the network controller path as a sweep so all venues migrate together. Will reference the issue from the Kalshi (venue 2/3) and Hyperliquid (venue 3/3) PRs.

oxoxDev:
@coderabbitai review

src/openhuman/tools/impl/network/polymarket.rs
```diff
@@ -0,0 +1,1395 @@
...
+                side,
+                token_id,
+                price,
+                size,
+                time_in_force,
+                expiration_secs,
+                approved,
+                user,
+            } => {
+                require_write_approval(approved)?;
```
graycyrus, line 420: **[major]** Write operations (`place_order`, `cancel_order`) bypass the project's security gate. Sibling code in `composio.rs` calls `security.enforce_tool_operation(ToolOperation::Act, "composio.execute_action")` before any write HTTP call. The `require_write_approval` stopgap is Polymarket-specific but doesn't replace the framework's autonomy-level enforcement.

Add before `require_write_approval`:
```rust
self.security.enforce_tool_operation(
    crate::openhuman::security::ToolOperation::Act,
    "polymarket.place_order",
)?;
```

And similarly for `cancel_order` below (~line 445).

src/openhuman/tools/impl/network/polymarket.rs
```diff
@@ -0,0 +1,1395 @@
...
+            .map(ClobCredentials::from)
+            .filter(ClobCredentials::is_complete);
+
+        Self {
+            gamma_base_url: normalize_base_url(
+                &config.gamma_base_url,
+                "https://gamma-api.polymarket.com",
+            ),
+            clob_base_url: normalize_base_url(&config.clob_base_url, "https://clob.polymarket.com"),
+            polygon_rpc_url: normalize_base_url(&config.polygon_rpc_url, "https://polygon-rpc.com"),
```
graycyrus, line 183: **[major]** No HTTPS enforcement before sending credentials. L2 HMAC headers (API key, passphrase, signature) are sent to whatever URL is in `clob_base_url` with no protocol check. If a user misconfigures `clob_base_url = "http://..."` in config.toml, credentials leak in plaintext.

Sibling code (`composio.rs`) has an `ensure_https()` guard before any credential-bearing request. Add a similar check — either validate URLs at construction time, or guard in `get_signed_clob_json`/`post_signed_clob_json`/`delete_signed_clob_json` before sending auth headers.

The same applies to `polygon_rpc_url` if any future use sends credentials, though currently the `eth_call` for allowance is read-only.

src/openhuman/config/schema/tools.rs
```diff
@@ -554,6 +554,91 @@ impl Default for IntegrationToggle {
...
+
+#[derive(Clone, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+pub struct PolymarketClobCredentials {
+    pub api_key: String,
+    pub secret: String,
+    pub passphrase: String,
+}
+
+impl std::fmt::Debug for PolymarketClobCredentials {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
graycyrus, line 609: **[minor]** `api_key` is still shown in plaintext in the Debug impl. While less sensitive than the signing `secret`, it's still a credential that could appear in logs if this struct is debug-printed. Consider masking it:

```rust
.field("api_key", &format!("{}…", &self.api_key[..self.api_key.len().min(8)]))
```

or fully redacting like the other fields.

src/openhuman/tools/impl/network/clob_auth.rs
```diff
@@ -0,0 +1,419 @@
...
+            || self.secret.trim().is_empty()
+            || self.passphrase.trim().is_empty())
+    }
+}
+
+impl std::fmt::Debug for ClobCredentials {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ClobCredentials")
+            .field("api_key", &self.api_key)
+            .field("secret", &"<redacted>")
```
graycyrus: **[minor]** Same as above — `ClobCredentials` Debug impl exposes `api_key` in plaintext. Consider masking or redacting to match `secret`/`passphrase` treatment.

src/openhuman/tools/impl/network/polymarket.rs
```diff
@@ -0,0 +1,1395 @@
...
+            crate::openhuman::config::apply_runtime_proxy_to_builder(builder, "tool.polymarket");
+
+        let http = match builder.build() {
+            Ok(client) => client,
+            Err(err) => {
+                tracing::warn!(
+                    reason = %err,
+                    "[polymarket] failed to build configured HTTP client, falling back to Client::new()"
+                );
+                Client::new()
```
graycyrus: **[minor]** If the proxy-aware builder fails, this falls back to a bare `Client::new()` that bypasses proxy settings and uses default timeouts. The `tracing::warn` is good, but silently degrading means the tool may appear to work while skipping proxy/timeout config. Consider whether this should be a hard error instead — if the builder can't be configured, the tool likely can't operate correctly in the user's network environment.

graycyrus (changes requested):
## Walkthrough

Solid integration adding Polymarket read + trading capabilities (venue 1/3 for #1398). Well-structured module decomposition (`polymarket.rs`, `clob_auth.rs`, `polymarket_orders.rs`), comprehensive test suite (22+ tests with mock server), proper credential redaction in Debug for secret/passphrase, and effective approval gate for write operations. Two security patterns from sibling code (`composio.rs`) are missing: the `enforce_tool_operation` security gate and HTTPS URL enforcement for credential-bearing requests.

## Change Summary

| `polymarket.rs` | New (1395 lines) | Main tool: 11 actions, HTTP helpers with retry, credential caching, order signing |
| `clob_auth.rs` | New (419 lines) | L1 EIP-712 handshake + L2 HMAC-SHA256 signing |
| `polymarket_orders.rs` | New (158 lines) | EIP-712 order typed-data signing |
| `polymarket_tests.rs` | New (971 lines) | 22+ tests: mock server, signing fixtures, approval gates, error paths |
| `config/schema/tools.rs` | Modified | PolymarketConfig + PolymarketClobCredentials structs |
| `tools/schemas.rs` | Modified | RPC controller schema + handler |
| `tools/ops.rs` | Modified | Conditional tool registration |
| `about_app/catalog.rs` | Modified | 2 capability entries (readonly + trading) |
| `wallet/mod.rs` | Modified | `secret_material` exported as `pub(crate)` |
| `config/mod.rs`, `config/schema/mod.rs` | Modified | Re-exports for new types |
| `gitbooks/` | New + Modified | Integration docs + SUMMARY link |
| `tests/fixtures/polymarket/*.json` | New (9 files) | Test fixtures |

## Per-file analysis

**polymarket.rs** — Clean action dispatch with tagged enum deserialization. HTTP helpers follow the retry/error pattern well (4xx no retry, 429/5xx 3× with 500ms backoff). The approval gate properly blocks write actions before any network call. Two security concerns noted inline.

**clob_auth.rs** — Good separation of L1 and L2 auth concerns. Structured tracing with grep-friendly `[polymarket][clob_auth]` prefixes. Address masking in logs is a nice touch.

**polymarket_orders.rs** — Straightforward EIP-712 signing. Domain separator correctly pins chain 137 and exchange contract.

**polymarket_tests.rs** — Thorough coverage including happy paths, auth header verification, approval gate enforcement (zero-HTTP property), error paths, and signing fixtures.

**config/schema/tools.rs** — Follows sibling config patterns. Enabled defaults to false.

**tools/schemas.rs** — Controller handler has good entry/exit logging and validates argument types.

## Additional notes

**Missing action-level debug logging**: `handle_request()` in `polymarket.rs` has no entry/exit logging for individual actions. Sibling `clob_auth.rs` has excellent `[polymarket][clob_auth]` prefixed logging, but the main dispatch path is silent. Consider adding `tracing::debug!(action = ..., "[polymarket] handle_request: entry")` and similar for each action's network call, consistent with the watchlist requirement for grep-friendly debug logging in new/changed flows.

src/openhuman/tools/impl/network/polymarket.rs
oxoxDev: Addressed in ba09df78 — added `self.security.enforce_tool_operation(ToolOperation::Act, "polymarket.place_order")` (and "polymarket.cancel_order") before `require_write_approval`. New test `place_order_blocked_by_readonly_security_policy_before_approval_check` verifies the autonomy-level block fires first and zero HTTP is issued.

src/openhuman/tools/impl/network/polymarket.rs
oxoxDev: Addressed in a18cbd0d — added `ensure_https` mirroring composio.rs and call it before sending L2 HMAC headers in `get_signed_clob_json_with_creds` / `post_signed_clob_json` / `delete_signed_clob_json`, plus the L1 `derive_credentials` handshake path. Loopback http (`127.0.0.1` / `localhost` / `[::1]`) is carved out only for the TcpListener mock harness — real Polymarket hosts never resolve to loopback. Three unit tests cover https-pass, loopback-pass, remote-http-reject.

src/openhuman/config/schema/tools.rs
oxoxDev: Addressed in 53d4d765 — `api_key` now redacted alongside `secret` and `passphrase` in the Debug impl on `PolymarketClobCredentials`. Mirrors the `clob_auth.rs` change in 332c292e.

src/openhuman/tools/impl/network/clob_auth.rs
oxoxDev: Addressed in 332c292e — `api_key` now redacted in `ClobCredentials` Debug impl. Full redaction (matching `secret` / `passphrase` treatment) chosen over partial-mask since the value is already only used in HMAC signing.

src/openhuman/tools/impl/network/polymarket.rs
oxoxDev: Addressed in 961811e6 — replaced the `tracing::warn` + `Client::new()` fallback with a hard panic carrying the underlying reqwest error. Builder failure is impossible-configuration territory (TLS init, malformed user proxy URL); silent degradation hid the misconfig from users. Startup panic surfaces it loudly.

src/openhuman/tools/impl/network/clob_auth.rs
```diff
@@ -0,0 +1,419 @@
...
+    )
+}
+
+fn mask_address(address: &str) -> String {
+    let trimmed = address.trim();
+    if trimmed.len() <= 10 {
+        return "<redacted>".to_string();
+    }
+    format!("{}...{}", &trimmed[..6], &trimmed[trimmed.len() - 4..])
+}
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 379c656e — `mask_address` now slices via `chars().collect::<Vec<char>>()` so any non-ASCII input passes through Unicode-safely. Added two unit tests (`mask_address_handles_non_ascii_without_panic`, `mask_address_redacts_short_input`) to guard the path.

src/openhuman/tools/impl/network/clob_auth.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    let mut headers = HeaderMap::new();
+    headers.insert(POLY_ADDRESS, HeaderValue::from_str(address)?);
+    headers.insert(POLY_SIGNATURE, HeaderValue::from_str(&signature)?);
+    headers.insert(
+        POLY_TIMESTAMP,
+        HeaderValue::from_str(&timestamp.to_string())?,
+    );
+    headers.insert(POLY_NONCE, HeaderValue::from_str(&nonce.to_string())?);
+    Ok(headers)
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 1c0b8bd6 — `sign_l1_headers` now `anyhow::ensure!`s that `signer.address()` matches the supplied `address` (case-insensitive hex compare) before signing. Defense-in-depth as you described — current callers already validated in `resolve_signer_and_user`, this co-locates the invariant inside the signing fn itself. Two new tests cover match + mismatch paths against a fixture mnemonic.

src/openhuman/tools/impl/network/polymarket_tests.rs
```diff
@@ -0,0 +1,1034 @@
...
+                    Ok(read) => read,
+                    Err(_) => return,
+                };
+                if n == 0 {
+                    return;
+                }
+
+                let request_raw = String::from_utf8_lossy(&buf[..n]).to_string();
+                let observed = parse_request(&request_raw);
+                let target = observed.target.clone();
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 92d2ddf6 — mock-server task now reads in 4 KiB chunks, accumulates into `buf`, and stops on `request_is_complete` (verifies `\r\n\r\n` plus `body.len() >= Content-Length`). Eliminates the partial-read truncation class of flake.

src/openhuman/tools/impl/network/polymarket_tests.rs
```diff
@@ -0,0 +1,1034 @@
...
+    }
+}
+
+impl Drop for EnvVarGuard {
+    fn drop(&mut self) {
+        unsafe {
+            std::env::remove_var(self.key);
+        }
+    }
+}
```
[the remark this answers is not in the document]
oxoxDev: Addressed in a83c420d — `EnvVarGuard` now snapshots the previous `OsString` at construction via `var_os` and restores it in `Drop` (or removes if it was `None`). Pre-existing `OPENHUMAN_WORKSPACE` survives the test.

oxoxDev:
@coderabbitai — addressing the duplicate finding on `clob_auth.rs:341-349` (`ClobAuthResponse` Debug leak): Addressed in f3ca1901. Dropped the `derive(Debug)` and replaced with a custom redacting impl that matches the `ClobCredentials` / `PolymarketClobCredentials` treatment from earlier commits — all three fields print as `<redacted>`.

M3gA-Mind (dismissed):
## Code Review — `feat(integrations): Polymarket trading + market data`

### Overview

Adds a new `PolymarketTool` exposing 11 actions (5 public reads, 4 authenticated reads, 2 writes) across Gamma + CLOB APIs, with EIP-712 order signing on Polygon. Gated behind `integrations.polymarket.enabled` (default off). Well-scoped first venue of three.

---

### Security

**Credentials stored in plaintext config**
- `derived_clob_credentials` (CLOB API key + HMAC secret) is written to the TOML config file in plaintext. This is acknowledged as temporary pending #1900. Since this is an on-by-default-off feature, the exposure window is controlled, but it should be called out prominently: any tool that reads the config file (including other agents) can exfiltrate the CLOB key. The TODO comment should include the threat model, not just the issue reference.

**`ensure_https` does not guard the Polygon RPC URL**
- `get_usdc_allowance_for_user` calls `post_json_raw` with `self.polygon_rpc_url` but there is no corresponding `ensure_https(&self.polygon_rpc_url)?` call before it. The RPC payload doesn't carry CLOB credentials, but it does carry wallet addresses. Recommend adding the guard for consistency.

**Concurrent credential derivation race**
- `ensure_clob_credentials` locks the Mutex, checks for cached creds, drops the lock, then derives and re-locks. Two concurrent callers can both see an empty cache and both issue `POST /auth/api-key`, generating two separate credentials — one gets silently dropped. Consider using `tokio::sync::OnceCell` or holding the lock across derivation.

**Test code uses `unsafe { std::env::set_var }`**
- This is UB in multi-threaded tests. Consider `serial_test` or equivalent if test parallelism is an issue.

**Stopgap approval gate**
- `require_write_approval` is a client-enforced string check. If the agent framework calls `execute()` directly without the normal approval channel, this gate is all that stands between an LLM hallucinating `approved: true` and a live order. Worth a security note in the inline comment beyond just the TODO reference.

---

### Code Quality

**File size far exceeds the project guideline**
- `polymarket.rs` is 1,419 lines — nearly 3× the ~500-line target from CLAUDE.md. The HTTP layer (`get_json` / `post_json_raw` / `delete_json`) alone is ~270 lines of near-identical retry logic that should be extracted to a shared `send_with_retry` helper.

**Retry logic is copy-pasted three times**
- `get_json`, `post_json_raw`, and `delete_json` each implement the same ~60-line retry loop differing only in HTTP verb and error message prefix. A single generic helper would cut ~120 lines and make the retry policy consistent.

**Dual credential structs with conversion boilerplate**
- `ClobCredentials` (in `clob_auth.rs`) and `PolymarketClobCredentials` (in config) are identical in fields and semantics, connected only by `From` impls. Unify them — the current split adds 30 lines of glue for no functional benefit.

**`price` / `size` use `f64` for financial amounts**
- Floating-point representation errors can still produce wrong fixed-point outputs for certain inputs (e.g. `0.3 * 1e6 ≠ 300000.0` exactly). Consider accepting price/size as strings and parsing to `rust_decimal::Decimal`, matching what the CLOB API actually expects.

**Mixed rounding modes without explanation**
- `round_normal` (half-up) for price, `round_down` (floor) for size. The asymmetry is correct (floor prevents over-selling) but needs a comment — a reader will assume it's a bug.

**Raw ABI encoding for `eth_call`**
```rust
let data = format!(
    "0xdd62ed3e{:0>64}{:0>64}",
    hex::encode(owner.as_bytes()),
    hex::encode(spender.as_bytes()),
);
```
This works but is brittle hand-rolled ABI encoding. A comment explaining why `hex::encode(addr.as_bytes())` produces exactly 40 chars (producing correctly zero-padded 32-byte ABI words) is warranted, or use `alloy-sol-types`/`ethabi` encoding.

**`salt` uses non-CSPRNG `rand::random`**
- `rand::random::<u64>()` is not cryptographically random. For an order salt, predictability may be exploitable by a front-runner. Use `rand::rngs::OsRng` or `getrandom` directly.

---

### Architecture

**`is_concurrency_safe` returns `true` for write actions — nonce collision risk**
- Two concurrent `place_order` calls each fetch `/nonce?user=` independently and receive the same nonce, causing one to fail at the CLOB. Either return `false` for write variants (or all variants given the Mutex interior state), or document the nonce collision risk explicitly.

**Timestamp reuse on auth fallback**
- `derive_credentials` computes `now_unix_secs()` once, then reuses the same timestamp for both the `POST /auth/api-key` attempt and the fallback `GET /auth/derive-api-key`. If the server enforces a short anti-replay window, a slow first request can cause the fallback to fail with a stale timestamp. Re-derive `now_unix_secs()` before the fallback call.

**`persist_clob_credentials` loads + saves whole config on every derivation**
- Acknowledged as temporary, but concurrent config writes from multiple tools could clobber each other. The in-memory `Mutex` cache should be documented as the source of truth; the persist is best-effort only (the warn log is already there — just add the doc comment).

---

### Test Coverage

**Positives:**
- 22 tests covering all 11 actions, HMAC fixture, EIP-712 fixture, approval gate (no-HTTP property), retry/5xx, timeout
- In-process `TcpListener` mock server with request capture is clean
- `EnvVarGuard` RAII for environment teardown is the right pattern

**Missing cases:**
- `get_market` with neither `market_id` nor `slug` → error path untested
- Credential derivation fallback (POST → GET) not exercised — needs two-response mock sequence
- `persist_clob_credentials` failure path — warn log is present but path is untested
- Concurrent `place_order` calls → nonce collision scenario not covered

---

### Summary

| Severity | Issue |
|---|---|
| **Must fix** | `is_concurrency_safe` → `true` for write actions risks nonce collision |
| **Must fix** | `rand::random` non-CSPRNG salt for order signing |
| **Must fix** | `ensure_https` missing on Polygon RPC URL |
| **Should fix** | Timestamp reuse between POST and fallback GET in `derive_credentials` |
| **Should fix** | Deduplicate retry loop across `get_json`/`post_json_raw`/`delete_json` |
| **Should fix** | `get_market` missing-both-params case is untested |
| **Consider** | Credential derivation race (concurrent callers) |
| **Nit** | CSPRNG for salt, `f64` → `Decimal` for price/size, file size split |

The overall structure is solid, security thinking is correct for a Beta feature, and the test harness is well-designed. The critical blockers are the concurrency safety flag for writes and the non-CSPRNG salt.

src/openhuman/tools/impl/network/polymarket.rs
graycyrus: Confirmed fixed — enforce_tool_operation called on both write paths.

src/openhuman/tools/impl/network/polymarket.rs
graycyrus: Confirmed fixed — `ensure_https()` called before all credential-bearing requests (GET/POST/DELETE signed, plus credential derivation). Loopback HTTP correctly exempted for test mocks.

src/openhuman/config/schema/tools.rs
graycyrus: Confirmed fixed — manual Debug impl now redacts api_key, secret, and passphrase.

src/openhuman/tools/impl/network/clob_auth.rs
graycyrus: Confirmed fixed — ClobCredentials Debug impl redacts all three credential fields.

src/openhuman/tools/impl/network/polymarket.rs
graycyrus: Confirmed fixed — builder failure now panics with a clear message instead of silently falling back to Client::new().

src/openhuman/tools/impl/network/polymarket.rs
```diff
@@ -0,0 +1,1419 @@
...
+                },
+                "latest"
+            ]
+        });
+
+        let response = self
+            .post_json_raw(
+                &self.polygon_rpc_url,
+                "",
+                Some(HeaderMap::new()),
```
graycyrus: **[major]** Polygon RPC `eth_call` is sent via `post_json_raw` with an empty `HeaderMap` — no `Content-Type: application/json` header. The JSON-RPC spec requires it, and strict providers (Alchemy, Infura, some self-hosted nodes) will reject the request with 415 or silently misparse the body.

The CLOB POST path correctly adds the header in `post_signed_clob_json`; this path should do the same.

Suggestion:
```rust
let mut rpc_headers = HeaderMap::new();
rpc_headers.insert(
    reqwest::header::CONTENT_TYPE,
    HeaderValue::from_static("application/json"),
);
let response = self
    .post_json_raw(
        &self.polygon_rpc_url,
        "",
        Some(rpc_headers),
        serde_json::to_string(&payload)
            .context("Failed to serialize Polygon RPC payload")?,
    )
    .await?;
```

graycyrus (dismissed):
## Walkthrough

Continuation review of the Polymarket integration (venue 1/3 for #1398). All five issues from my previous REQUEST_CHANGES review have been verified as addressed — SecurityPolicy enforcement on writes, HTTPS enforcement before credentials, Debug redaction on both credential types, and the Client::new() fallback replaced with a hard panic. Good work on the follow-up commits.

One new issue found on this pass: the Polygon RPC call for `get_usdc_allowance` sends a JSON-RPC payload without `Content-Type: application/json`, which will break with strict RPC providers.

## Previous Requested Changes — All Addressed

| # | Finding | Status |
|---|---------|--------|
| 1 | SecurityPolicy bypass on writes | **Fixed** — `enforce_tool_operation` on both `place_order` and `cancel_order` |
| 2 | No HTTPS enforcement before CLOB credentials | **Fixed** — `ensure_https()` on all 4 credential-bearing paths + derivation |
| 3 | PolymarketClobCredentials Debug leaks api_key | **Fixed** — manual Debug impl, all fields redacted |
| 4 | ClobCredentials Debug leaks api_key | **Fixed** — manual Debug impl, all fields redacted |
| 5 | Client::new() silent fallback | **Fixed** — panics with diagnostic message |

## Change Summary

| `polymarket.rs` | New (1419 lines) | Main tool: 11 actions, retry, HTTPS enforcement, SecurityPolicy |
| `clob_auth.rs` | New (495 lines) | L1 EIP-712 handshake + L2 HMAC signing |
| `polymarket_orders.rs` | New (158 lines) | EIP-712 order typed-data signing |
| `polymarket_tests.rs` | New (1068 lines) | 22 tests: happy paths, approval gate, signing, retry, timeout |
| `tools.rs` (config) | Modified | PolymarketConfig + PolymarketClobCredentials with serde defaults |
| `catalog.rs` | Modified | 2 capability entries (readonly + trading) |
| `schemas.rs` | Modified | Controller schema + handler |
| `ops.rs` | Modified | Conditional tool registration |
| `wallet/mod.rs` | Modified | `secret_material` exposed as `pub(crate)` |
| `gitbooks/` | New + Modified | Integration guide + SUMMARY.md entry |
| `tests/fixtures/` | New (8 files) | JSON fixtures for mock responses |

## New Finding

oxoxDev:
@M3gA-Mind — thank you for the thorough pass. Going through item by item. Eight commits landed on the branch addressing your review, plus an explicit deferral list for the items that need wider architectural alignment.

## Must-fix (3/3) ✅

| # | Item | Commit |
|---|---|---|
| 1 | `is_concurrency_safe` → `false` for write actions (nonce collision) | `012c6397` |
| 2 | CSPRNG salt via `OsRng` (no more `rand::random`) | `a88af32c` |
| 3 | `ensure_https` guard on `polygon_rpc_url` in allowance read | `e2e54833` |

## Should-fix (3/3) ✅

| # | Item | Commit |
|---|---|---|
| 4 | Re-derive timestamp before fallback in `derive_credentials` | `b440dc72` |
| 5 | Extract `send_with_retry` helper — dedupes ~180 LOC across `get_json` / `post_json_raw` / `delete_json` | `36034502` |
| 6 | `get_market` with neither `market_id` nor `slug` test | `cf3604c3` (POST→GET fallback + persist-failure tests deferred — see below) |

## Architectural cleanup (in this PR) ✅

| # | Item | Commit |
|---|---|---|
| 7 | Unify `ClobCredentials` + `PolymarketClobCredentials` (delete the dual struct + From-impl glue) | `2b098ee4` |
| 8 | `ensure_clob_credentials` now holds the cache lock across `derive_credentials` — single-flight derivation, no more concurrent-caller race | `2b098ee4` |
| 9 | Inline doc comments for non-obvious decisions: approval-gate stopgap (with security note), `round_down`/`round_normal` asymmetry, raw ERC-20 ABI encoding, `persist_clob_credentials` best-effort guarantee, `derived_clob_credentials` plaintext threat model | `2b4ce323` |

## Deferred to follow-up (explicit rationale)

| Item | Why deferred |
|---|---|
| **`f64 → rust_decimal::Decimal`** for price/size | Crate addition sets precedent for the Kalshi + Hyperliquid PRs (venues 2/3 + 3/3 on this umbrella). Want architectural buy-in on Decimal vs string-passthrough before paying the dep cost on all three venues. Filing as follow-up after the umbrella merges. |
| **File-size split** of `polymarket.rs` (1419 LOC, >2.8× the project guideline) | Genuine refactor that needs module-boundary discussion — likely `polymarket/{tool.rs, request.rs, signing.rs, allowance.rs}`. Will land in a dedicated PR after the venue-3 PR opens (so Kalshi/Hyperliquid can adopt the same shape). |
| **Replace `unsafe { std::env::set_var }`** with `serial_test` crate | The existing `TEST_ENV_LOCK` static `std::sync::Mutex<()>` already serializes env-var tests project-wide; the unsafe blocks are an artefact of Rust 2024 edition's tightened std API and exist throughout the codebase. Migrating just polymarket tests to `serial_test` adds a new dev-dep + breaks the project's existing serialization convention. Worth a project-wide RFC. |
| **Credential-derivation POST→GET fallback test** + **persist-failure path test** | Requires a multi-response mock fixture (the existing route helper takes a single VecDeque per path). Cleanest landed alongside the test-helper refactor I'm planning for Kalshi/Hyperliquid (whose RSA-PSS + EIP-712 signing paths need similar multi-response tests). |
| **`SecretStore` migration for `derived_clob_credentials`** | Explicitly blocked on #1900. Threat-model is now documented inline so the gap is visible until that lands. |

## What you might also want to look at

- The HTTPS guard carves out `127.0.0.1` / `localhost` / `[::1]` for the in-process `TcpListener` mock harness — explicitly documented as a test affordance. Real Polymarket CLOB hosts don't resolve to loopback. If you'd prefer `#[cfg(test)]`-gated allowance instead, happy to refactor.
- `is_concurrency_safe` returns `false` only for `place_order` / `cancel_order` — reads stay concurrency-safe. Want it tighter (e.g. include credentialed reads like `get_positions`)? They don't have the nonce-collision pattern but do share the cache-warm-up window.
- The `require_write_approval` doc comment now spells out the framework gate (`SecurityPolicy::enforce_tool_operation`) + stopgap layering so reviewers know both must remain until #1339.

Re-pushed all 9 commits to `oxoxDev:feat/1398-polymarket`. CR is re-running on the latest tip.

src/openhuman/tools/impl/network/polymarket.rs
oxoxDev: Addressed in 38f7777b — `get_usdc_allowance_for_user` now builds a `HeaderMap` with `Content-Type: application/json` and passes it to `post_json_raw`. Test extended to assert the header is present on the captured Polygon RPC request (regression guard for the venue 2/3 PRs too).

graycyrus (approved):

PR #2229, opened 2026-05-19
fix(tauri): forward hot-instance OAuth deep links

### LINKED ISSUE #2228: Windows desktop OAuth callbacks do not reach the running app instance
---
name: Bug
about: Used for bug reports
title: ""
type: Bug
assignees: ''

---

## Summary

Desktop OAuth login can start from GitHub or Google, but when OpenHuman is already running on Windows the browser redirect back to `openhuman://...` never reaches the live app instance, so login cannot complete.

## Problem

On Windows desktop `v0.54.0`, GitHub and Google sign-in both fail in the same way:
1. Launch OpenHuman and leave it running.
2. Click `Continue with GitHub` or `Continue with Google`.
3. Complete the provider auth flow in the system browser.
4. The browser returns focus to OpenHuman, but the app stays signed out and no OAuth completion happens.

Expected:
- The existing OpenHuman instance receives the `openhuman://auth?...` or `openhuman://oauth/...` deep link and finishes login.

Actual:
- The login button logs the OAuth start, then only logs `window focus -> reset isLoading`.
- No OAuth success/error deep-link event reaches the live renderer.
- The issue reproduces across both GitHub and Google, so it is not provider-specific.

Additional diagnostics:
- Cold-start deep links work: if OpenHuman is fully closed, triggering `openhuman://oauth/error?...` launches the app and the renderer receives `[DeepLink][oauth:error] ...`.
- Hot-instance deep links fail: when OpenHuman is already running, the same external deep link never reaches the renderer.
- Internal simulated deep links still work via `window.__simulateDeepLink(...)`, so `desktopDeepLinkListener.ts` is alive and the failure is in second-instance deep-link forwarding.
- `app/src-tauri/Cargo.toml` currently uses `tauri-plugin-single-instance = "2"` without the `deep-link` feature, which Tauri requires on Windows/Linux to forward second-launch deep links into the primary instance.

Environment:
- App version: `0.54.0`
- OS: Windows desktop
- Install type: packaged desktop app

## Solution (optional)

Enable the `deep-link` feature on `tauri-plugin-single-instance` in `app/src-tauri/Cargo.toml`:

```toml
tauri-plugin-single-instance = { version = "2", features = ["deep-link"] }
```

That should preserve the existing single-instance CEF guard while allowing second-launch deep-link payloads to reach the primary app instance during OAuth callbacks.

## Acceptance criteria

- [ ] **Repro gone** — Bug no longer reproduces on Windows when OpenHuman is already running and the user completes GitHub/Google OAuth in the system browser.
- [ ] **Regression safety** — Unit, integration, or E2E coverage added or updated so the single-instance deep-link wiring is verified.
- [ ] **Cold + hot deep links** — Both cold-start and already-running-instance `openhuman://oauth/...` callbacks are verified on Windows.

## Related

- Local repro used the packaged `OpenHuman 0.54.0` Windows app.
- The live renderer accepted `window.__simulateDeepLink('openhuman://oauth/error?...')`, but did not receive the same event from an external deep-link launch while already running.

NgoQuocViet2001:
## Summary
- Enable the `deep-link` feature on `tauri-plugin-single-instance`.
- Update `app/src-tauri/Cargo.lock` so the single-instance plugin links against `tauri-plugin-deep-link`.
- Document why the feature is required for Windows/Linux second-launch OAuth callback forwarding.

## Problem
Closes #2228. On Windows, OAuth completes in the system browser, but when OpenHuman is already running the `openhuman://oauth/...` second launch only focuses the primary instance. The live renderer never receives the callback, while cold-start and simulated deep links still work.

`tauri-plugin-single-instance` was installed to prevent secondary CEF launches, but without its `deep-link` feature the second-launch deep-link payload is not forwarded into the primary app on Windows/Linux.

## Validation
- `cargo metadata --manifest-path app/src-tauri/Cargo.toml --format-version 1 --no-deps` + assertion that `tauri-plugin-single-instance` includes `deep-link` -> passed.
- `cargo fmt --manifest-path app/src-tauri/Cargo.toml -p OpenHuman --check` -> passed.
- `git diff --check` -> passed.
- `cargo check --manifest-path app/src-tauri/Cargo.toml` -> attempted; blocked by local native toolchain gaps unrelated to this change: missing `cmake` for `cef-dll-sys` and missing `libclang`/`LIBCLANG_PATH` for `whisper-rs-sys`. Cargo reached dependency resolution and native build scripts before failing.

## Duplicate check
Searched existing PRs for `#2228`, "Windows desktop OAuth callbacks", and `single-instance` + `deep-link feature`; no open or recent PR covers this fix. Prior #1510 added the single-instance guard but did not enable the deep-link forwarding feature.

## Summary by CodeRabbit

* **New Features**
  * Deep-link payload forwarding for second-instance app launches is now supported.
* **Tests**
  * Added a regression test to verify the deep-link feature is declared and remains enabled.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src-tauri/Cargo.toml
```diff
@@ -50,8 +50,10 @@ tauri-plugin-opener = "2"
 # `cef::initialize(...) != 1` cache-lock panic seen in production
 # (Sentry OPENHUMAN-TAURI-A). The plugin acquires a per-identifier
 # lock before any tauri::Builder work happens, so the secondary
-# process exits cleanly after handing its argv to the primary.
-tauri-plugin-single-instance = "2"
+# process exits cleanly after handing its argv to the primary. The `deep-link`
+# feature forwards second-launch deep-link payloads to the primary instance on
+# Windows/Linux, which is required for hot-instance OAuth callbacks.
+tauri-plugin-single-instance = { version = "2", features = ["deep-link"] }
```
graycyrus, line 56: **[major]** Missing regression test required by #2228 acceptance criteria.

The issue explicitly lists:
> - Regression safety — Unit, integration, or E2E coverage added or updated so the single-instance deep-link wiring is verified.
> - Diff coverage ≥ 80 %

Suggestion: add a build-time assertion that `tauri-plugin-single-instance` includes the `deep-link` feature, e.g.:
```rust
#[cfg(test)]
#[test]
fn single_instance_has_deep_link_feature() {
    // cargo metadata already verified this in CI validation,
    // but a compiled test catches regressions if someone removes the feature.
    assert!(cfg!(feature = "deep-link") || cfg!(target_os = "macos"),
            "tauri-plugin-single-instance must enable the deep-link feature on Windows/Linux");
}
```
Alternatively, a CI script assertion using `cargo metadata --no-deps` works too — the PR's own validation section already describes the command.

graycyrus (changes requested):
Nice find — enabling the `deep-link` feature on `tauri-plugin-single-instance` is the correct fix, and the inline comment documenting the *why* is appreciated.

One thing holding this back: issue #2228 acceptance criteria explicitly require regression test coverage and ≥ 80 % diff coverage. The PR currently adds no tests. Since the change is a Cargo feature flag, a lightweight `cargo metadata` assertion (verifying the feature is present) or an integration-level deep-link forwarding test would satisfy the gate without being heavy.

Otherwise the Cargo.toml + lockfile change itself looks correct — just needs the test coverage to land.

app/src-tauri/Cargo.toml
senamakel, line 56: Added a regression test in 406597513 — `tests::single_instance_dep_enables_deep_link_feature` in `app/src-tauri/src/lib.rs` parses `Cargo.toml` at test time and asserts the `deep-link` feature is present on `tauri-plugin-single-instance`. This fails fast if anyone removes the feature flag in the future (issue #2228 regression safety + diff coverage). I went with a Cargo.toml-parse test rather than the suggested `cfg!(feature = "deep-link")` form because `cfg!` checks features of *this* crate (`OpenHuman`), not the transitive plugin crate, so it wouldn't actually catch the regression we care about.

NgoQuocViet2001:
@graycyrus thanks again for the review. The requested regression coverage was added in 406597513d5790db0b6e7bd42c7787553d633110 via `tests::single_instance_dep_enables_deep_link_feature`, which parses `app/src-tauri/Cargo.toml` and asserts that `tauri-plugin-single-instance` keeps the `deep-link` feature enabled.

I noticed the original review thread may still appear unresolved after merge. If there is anything still missing from the #2228 acceptance criteria, happy to follow up in a separate PR.

PR #2294, opened 2026-05-20
fix(observability): demote transient OpenAI embeddings 429s to expected and reduce Sentry noise

YellowSnnowmann:
## Summary

- Updated OpenAI embeddings error reporting to route non-2xx failures through report_error_or_expected instead of always emitting hard error events.

- Standardized embedding HTTP error text to canonical format: Embedding API error (<status>): <body>.

- Ensured transient upstream statuses (especially 429 Too Many Requests) match observability classifiers and are demoted to warning breadcrumbs.

- Added a focused regression test for 429 formatting/classification to prevent future Sentry-noise regressions.

## Problem

- OpenAI embeddings calls can return transient 429 rate-limit responses during normal load.

- These errors are retried by memory-tree/background flows, but each attempt was still producing Sentry error events.

- The previous message shape (Embedding API error <status>: ...) did not consistently match the transient HTTP classifier expectations, increasing alert noise and masking actionable failures.

## Solution

- Switched embeddings non-2xx reporting to report_error_or_expected(...), allowing transient failures to be classified as expected noise.

- Changed the emitted error string to canonical classifier-compatible shape: Embedding API error (<status>): <body>.

- Kept failure tags (model, status, failure=non_2xx) for diagnostics while reducing noisy escalation.

- Added embed_429_uses_canonical_transient_format test to assert:

  - message contains (429 Too Many Requests), and
  - is_transient_message_failure(...) returns true.

## Submission Checklist

- All affected feature IDs from the matrix are listed in the PR description under ## Related (N/A if no feature IDs are affected)
- Manual smoke checklist updated if this touches release-cut surfaces (`docs/RELEASE-MANUAL-SMOKE.md`) (N/A if not release-cut)
- Linked issue closed via Closes #NNN in the ## Related section (N/A until issue is linked)

## Impact

- Runtime/platform impact: Rust core observability path only (desktop app behavior unchanged for users except reduced false-positive error noise).

- Performance: negligible runtime overhead change; likely reduced observability/event volume under transient rate limiting.

- Security/migration/compatibility: no migration or API compatibility changes; no new dependencies introduced.

## Related

- Closes: [OPENHUMAN-TAURI-RQ](https://tinyhumans.sentry.io/issues/7485954703/?environment=development&environment=production&environment=staging&project=4511287579836416&query=is%3Aunresolved%20assigned%3Anone&referrer=issue-stream&sort=freq)

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized error message formatting for embedding API failures with clearer HTTP status code display
  * Enhanced error classification for transient upstream API failures

* **Tests**
  * Added test coverage for rate-limiting error handling to validate transient error classification

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/openhuman/embeddings/openai_tests.rs
```diff
@@ -224,6 +224,36 @@ async fn embed_server_error() {
...
+
+    let err = p.embed(&["hi"]).await.unwrap_err();
+    let msg = err.to_string();
+    assert!(
+        msg.contains("(429 Too Many Requests)"),
+        "expected canonical transient HTTP shape, got: {msg}"
+    );
+    assert!(
+        crate::core::observability::is_transient_message_failure(&msg),
+        "message should classify as transient: {msg}"
```
graycyrus, line 263: **[major]** Regression test doesn't guard the actual regression.

`is_transient_message_failure` matches on the token `"429 too many requests"` (bare substring), which means it returns `true` for **both** the old format (`Embedding API error 429 Too Many Requests: …`) and the new one. This test would have passed before the format fix, so it doesn't actually prevent the regression from re-occurring.

The classifier that *broke* with the old format is `is_transient_upstream_http_message` (line 240, `observability.rs`) — it specifically requires the `"api error (429"` substring with the opening paren. Since that function is `pub(crate)` private, you can't call it from here directly, but you have two options:

1. **Preferred**: make `is_transient_upstream_http_message` `pub(crate)` → `pub` (or add a thin `pub` test-helper), then assert it directly.
2. **Alternative**: add an assertion that the message contains the literal substring `"api error (429 "` (lowercased) to at least prove the shape is compatible with the classifier's match arm.

Option 2 is quick:
```rust
assert!(
    msg.to_ascii_lowercase().contains("api error (429 "),
    "message must match is_transient_upstream_http_message classifier arm: {msg}"
);
```

Without this, a future refactor could accidentally revert to the old format and this test would still pass green.

graycyrus (commented):
Nice, focused fix — the parenthesized format is clearly needed for the `is_transient_upstream_http_message` classifier (line 240 in `observability.rs`: `lower.contains(&format!("api error ({code}"))`) and the test is a welcome addition.

One thing worth tightening up on the test — see inline comment.

src/openhuman/embeddings/openai_tests.rs
graycyrus: Confirmed fixed — commit 6acfcd02 adds the `api error (429 ` substring assertion (option 2). The test now guards both the broad classifier and the specific `is_transient_upstream_http_message` match arm shape. Nice work.

graycyrus (commented):
**Re-review — all prior changes addressed.**

The new commit (`6acfcd02`) adds the `api error (429 ` substring assertion I flagged in the prior review. The test now has three layers of defense:
1. Canonical `(429 Too Many Requests)` shape check
2. Classifier-arm substring guard (`api error (429 `)
3. Broad `is_transient_message_failure` classifier pass

This is a clean, focused fix — format string tweak + solid regression test. No new issues found.

| File | Change |
|------|--------|
| `openai.rs` | Format `{status}` → `({status})` in error string + updated comment |
| `openai_tests.rs` | New `embed_429_uses_canonical_transient_format` test with three assertions |

graycyrus (approved):
Looks good, nice work!

M3gA-Mind (approved):
Looks good, nice work!

PR #2306, opened 2026-05-20
feat(mcp): add memory.store and memory.note write tools

### LINKED ISSUE #2269: Add write tools to the MCP server (memory.store, memory.note, tree.tag)
## Summary

Phase 1+2a+2b shipped the read-only MCP surface (`memory.search`, `memory.recall`, `tree.read_chunk`, `tree.browse`, `tree.top_entities`, `tree.list_sources`). The natural next step is write tools (`memory.store`, `memory.note`, `tree.tag`), letting MCP clients persist conversation-derived facts back into the memory tree.

Filing as an RFC-style issue because write tools touch multi-process coordination, provenance, and `SecurityPolicy` in ways the read surface didn't — looking for direction on four architecture questions before scoping a PR.

## Problem

Without write tools, MCP-connected clients can only read the user's memory tree:

```
User (in Claude Desktop): "Remember that my Q3 OKR changed."
Claude: …has no write surface. Suggests the user manually add the note in OpenHuman.
```

Compared with the Anthropic-official `memory` MCP server — which exposes `create_entities` / `add_observations` / `delete_entities` as full CRUD — OpenHuman's MCP surface today is read-only and therefore strictly less capable for the "Claude Desktop as a continuous-memory front-end" use case. Without writes:

- No way for the LLM to persist conversation-derived insights into the memory tree.
- No way to update existing chunks (e.g. correct a stale fact).
- No cross-client continuity: a decision made in Claude Desktop doesn't flow into the next Cursor session because nothing wrote it down.

## Solution (RFC)

Four design questions need maintainer direction before code:

### Q1 — Multi-process write path

When the OpenHuman desktop daemon is running (or the new HTTP MCP server in #2260 is bound), AND Claude Desktop spawns its own `openhuman-core mcp` stdio subprocess, two processes can write to the same memory tree DB. Options:

- **A.** Stdio MCP delegates writes to the running daemon over HTTP. Daemon is the sole writer; cache + `event_bus` stay consistent. Requires the daemon-running invariant or graceful degradation when absent.
- **B.** Refuse stdio writes when a daemon is already listening. Simpler; but Claude Desktop sessions lose write capability if the user closes the GUI.
- **C.** Both processes write directly, coordinate via SQLite WAL + advisory locks. No daemon dependency; but daemon caches go stale + double `event_bus` emission risk.

#2260 (HTTP MCP transport) might collapse this question — if HTTP becomes the canonical MCP entry-point, stdio writes become a corner case rather than the default.

### Q2 — Provenance

Every written chunk should record where it came from. Proposed: a `source_kind` of `mcp:<client_name>` (taken from the MCP `initialize` payload's `clientInfo`), so a chunk written by Claude Desktop reads as `source_kind = "mcp:Claude Desktop"`. Open question: schema migration on `chunks` vs JSON metadata column?

### Q3 — `SecurityPolicy`

Read tools currently go through `enforce_tool_operation(ToolOperation::Read, ...)`. Write tools fall under `ToolOperation::Write`. Open:

- Default-allow with audit log, or default-deny + per-tool opt-in?
- Per-write user confirmation? MCP spec doesn't have a built-in confirm primitive — would need to surface via the OpenHuman desktop UI (out-of-band).
- `memory.forget` (destructive) probably deferred or gated harder than `memory.store`.

### Q4 — Audit log

New `mcp_writes` table recording `(timestamp, client_info, tool_name, args_summary, resulting_chunk_id)` so users can review what LLMs wrote on their behalf? Or rely on existing tracing/logs?

## Proposed tool surface (open to revision)

| Tool | Behavior | First wave? |
| --- | --- | --- |
| `memory.note` | Append a note to an existing chunk | ✅ Safest write — start here |
| `memory.store` | Create a new chunk from LLM-provided content | ✅ Core value prop |
| `tree.tag` | Add/remove tags on a chunk | 🟡 Maybe |
| `memory.forget` | Soft-delete a chunk | ❌ Destructive — defer until UX story exists |

## Plan if direction approved

Multiple sub-PRs to keep review surface manageable:

1. **Infra PR** — schema migration for provenance + `mcp_writes` audit table + `SecurityPolicy::Write` integration. ~1500-2000 lines.
2. **`memory.note` PR** — first write tool, lowest risk. ~500 lines.
3. **`memory.store` PR** — new chunk creation. ~600 lines.
4. (optional) **`tree.tag` / `memory.forget`** — only if maintainer signals interest.

## Acceptance criteria

Deferred to follow-up issues once Q1-Q4 are answered. Umbrella ACs would be:

- [ ] Write path produces correct cache invalidation / event emission regardless of which process initiated.
- [ ] Every chunk written via MCP carries a provenance record identifying the client.
- [ ] Audit log surfaces all MCP writes to the user (UI surface TBD).
- [ ] `SecurityPolicy::Write` gating works equivalent to the existing `Read` gate.
- [ ] **Diff coverage ≥ 80%** per [`.github/workflows/coverage.yml`](../.github/workflows/coverage.yml).

## Related

- Feature IDs: `11.1.4` (MCP stdio server)
- Builds on: #1760, #1790, #1974
- Adjacent: #2260 (HTTP transport — may change Q1), #2030 (Settings UI), #2268 (tool annotations)
- Refs: #1845 (native MCP server umbrella)
- Industry reference: [Anthropic `memory` MCP server](https://github.com/modelcontextprotocol/servers/tree/main/src/memory) (full CRUD)

Liohtml:
## Summary

- Add `memory.store` MCP tool: creates memory documents from content with configurable namespace and tags, routing to the existing `doc_put` RPC
- Add `memory.note` MCP tool: creates annotation documents linked to existing chunks via `metadata.annotates_chunk_id`, also routing through `doc_put`
- Add `enforce_write_policy()` for write-specific security gating (distinct from read/act for audit clarity)
- Add `dispatch_write_tool()` with structured `tracing::info!` audit logging (tool, chunk_id, client fields)
- Both tools set `source_type = "mcp"` for provenance tracking
- 11 new unit tests covering param validation, defaults, slug generation, and error cases

Closes #2269

## Test plan

- [x] `cargo check -p openhuman --lib` passes
- [x] `cargo test -p openhuman --lib mcp_server::tools` — all 40 tests pass (29 existing + 11 new)
- [x] `cargo fmt --all` applied
- [x] CI checks pass on PR

## Summary by CodeRabbit

* **New Features**
  * Added memory.store to save memories (title + content) with automatic namespace, deterministic key generation and tagging support
  * Added memory.note to attach notes to specific memory chunks with automatic key generation and linked metadata and content structure

* **Tests**
  * Expanded unit tests covering the new memory tools, argument validation (required fields and unknown-field rejection), defaults, and metadata/content construction

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2306?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Liohtml:
Credit to @justinhsu1477 for the thorough RFC-style issue (#2269) — the architecture questions, proposed tool surface, and phased rollout plan made scoping this first wave (memory.store + memory.note) much cleaner. The provenance and audit logging patterns follow directly from the issue's Q2/Q4 proposals.

justinhsu1477:
@Liohtml — thanks for picking this up so fast, and for the explicit RFC credit.

Quick mapping of your implementation against the four open questions in #2269, mostly so I can verify I'm reading the diff right and so anyone landing here later sees the trade-off space:

- **Q1 (multi-process write path)** — you route through the existing `doc_put` RPC, which is effectively Option A from the RFC (daemon is the sole writer). Cleanest path for v1 and side-steps Option C's WAL-coordination work.
- **Q2 (provenance)** — `metadata.source_type = "mcp"` on each chunk. Schema-light, no migration.
- **Q3 (`SecurityPolicy`)** — separating `enforce_write_policy()` from the read enforcer is a nicer split than tunnelling both through the same gate; matches what the RFC was hoping for under "default-allow + audit".
- **Q4 (audit log)** — structured `tracing::info!` instead of a new `mcp_writes` table. Reasonable for v1; if maintainers later want queryable history this is the easiest place to swap a SQLite-backed sink in.

Two follow-ups I'd happily own so the RFC closes out cleanly:

1. **`tree.tag`** — the third tool the RFC proposed. Want me to spin it up as a follow-up PR using your `enforce_write_policy` + `dispatch_write_tool` scaffolding once this one lands, so it merges cleanly on top?
2. **`client_info` propagation** — RFC Q2 also suggested capturing the calling MCP client name (e.g. `"mcp:Claude Desktop"` from the `initialize` payload) on each chunk, so a write from Claude Desktop reads differently from one written by Cursor. Doesn't need to block this PR — happy to file as a separate issue once this lands if maintainers want it.

Happy to do a deeper line-level review if it's useful.

Liohtml:
@justinhsu1477 Great mapping — you're reading the implementation exactly right on all four questions.

Re your follow-ups:
1. **`tree.tag`** — yes, please! The `enforce_write_policy` + `dispatch_write_tool` scaffolding is designed to be additive, so a follow-up PR on top should be clean. Happy to review it.
2. **`client_info` propagation** — agreed this should be a separate issue. The `initialize` payload's `clientInfo` is the right source; the current `source_type = "mcp"` is a placeholder that your follow-up would refine to `"mcp:Claude Desktop"` etc.

A deeper line-level review would absolutely be useful if you have the time — especially around the `dispatch_write_tool` → `doc_put` RPC parameter assembly, since you know the memory schema better than anyone.

justinhsu1477:
@Liohtml — green light on `tree.tag` ack'd, both follow-ups are now up:

- **#2316** — `tree.tag` PR layered on top of this branch. Reuses `enforce_write_policy` + `dispatch_write_tool` unchanged; +242 LoC net (the new dispatch arm, the `tree_tag_schema`, the `build_rpc_params` branch, a small `required_non_empty_string_array` helper next to `optional_string_array`, and 8 unit tests covering each failure path). Marked `Depends on #2306` and will rebase cleanly when this lands.
- **#2317** — `client_info` propagation filed as a separate issue. Includes the proposed `mcp:<normalized-name>` shape on `source_type` plus the back-compat `"mcp"` fallback when `clientInfo.name` is absent / empty, so any client predating this stays on the current behaviour.

For the deeper line-level look at `dispatch_write_tool` → `doc_put` parameter assembly: I'd rather sit with the second CodeRabbit round here first so the review lands against the version heading for merge, instead of churning on in-flight code. Will circle back after that.

YOMXXX:
@senamakel #2306 当前 checks 24 success / 2 skipped / 0 pending / 0 failure，CodeRabbit approved，review threads 0 unresolved。这个是 MCP write tools 的 Phase 1（memory.store / memory.note），麻烦再看一下。

src/openhuman/mcp_server/tools.rs
```diff
@@ -383,6 +453,11 @@ pub async fn call_tool(name: &str, arguments: Value) -> Result<Value, ToolCallEr
             enforce_act_policy(spec.name).await?;
             return run_subagent_tool(&params).await;
         }
+        "memory.store" | "memory.note" => {
+            enforce_write_policy(spec.name).await?;
+            validate_controller_params(&spec, &params)?;
+            return dispatch_write_tool(spec.name, &params).await;
```
graycyrus, line 576: **[critical]** This early return bypasses `build_rpc_params()`. The raw MCP client arguments (`{title, content, namespace?, tags?}`) are sent directly to `openhuman.memory_doc_put` via `dispatch_write_tool` — which just does `params.clone()` into `try_invoke_registered_rpc`.

All the param transformation in `build_rpc_params("memory.store")` / `build_rpc_params("memory.note")` is dead code in the actual execution path:
- **No slug-based key** — the RPC gets no `key` field, so upsert dedup won't work
- **No `source_type = "mcp"`** — provenance tracking is lost
- **No namespace default** — `"mcp"` fallback never applies
- **No `metadata.annotates_chunk_id`** — `memory.note` annotations aren't linked to their target chunk

The unit tests pass because they call `build_rpc_params` directly, not the `call_tool` → `dispatch_write_tool` path — so this gap isn't caught.

**Fix options:**
1. Have `dispatch_write_tool` call `build_rpc_params(tool_name, Value::Object(params.clone()))?` before forwarding to the RPC.
2. Remove the early return entirely — let these tools fall through to the generic dispatch path (which already calls `build_rpc_params`), and fold the write-policy + audit logging around that flow.

Option 2 is cleaner since it keeps write tools on the same rails as read tools, but either works.

src/openhuman/mcp_server/tools.rs
```diff
@@ -564,6 +639,46 @@ fn build_rpc_params(
...
+            let mut metadata = Map::new();
+            metadata.insert("annotates_chunk_id".to_string(), Value::String(chunk_id));
+            let mut params = Map::new();
+            params.insert("namespace".to_string(), Value::String("mcp".to_string()));
+            params.insert("key".to_string(), Value::String(key));
+            params.insert("title".to_string(), Value::String(title));
+            params.insert("content".to_string(), Value::String(content));
+            params.insert("source_type".to_string(), Value::String("mcp".to_string()));
+            params.insert("metadata".to_string(), Value::Object(metadata));
+            Ok(params)
```
graycyrus, line 797: **[major]** `memory.note` generates the key as `mcp-note-{chunk_id}`, which means annotating the same chunk twice silently overwrites the first note. If multiple notes per chunk are intended (which the tool description implies — "append a note"), the key needs a disambiguator (e.g., timestamp or content hash). If only one note per chunk is intended, the tool description should say "set" not "append".

graycyrus (changes requested):
## Walkthrough

Adds two write tools (`memory.store`, `memory.note`) to the MCP surface, both routing through `openhuman.memory_doc_put`. Includes `enforce_write_policy()` for security gating, `dispatch_write_tool()` for audit-logged dispatch, a `slug_from()` utility for deterministic key generation, and 11 new unit tests. All prior CodeRabbit findings on slug edge cases are addressed — nice iterating on that.

| File | Changes |
|------|---------|
| `tools.rs` | +2 tool specs, +2 schemas, +call_tool routing, +build_rpc_params arms, +enforce_write_policy, +dispatch_write_tool, +slug_from, +11 tests |
| `protocol.rs` | +2 entries in tool-name list test |

**1 critical issue blocks this PR** — see inline comment. The `build_rpc_params` logic (key generation, provenance, metadata) is unreachable because `dispatch_write_tool` receives raw client args and passes them straight to the RPC.

**Minor notes** (not blocking):
- `enforce_write_policy` is functionally identical to `enforce_act_policy` (both use `ToolOperation::Act`). The "distinct log line for auditors" rationale from the doc comment doesn't hold since both produce the same log pattern — the only difference is the function name. Consider either adding a `ToolOperation::Write` variant or just reusing `enforce_act_policy` until the distinction matters.
- `dispatch_write_tool` mixes `tracing::info!` (success path) with `log::warn!`/`log::error!` (error paths). Pick one facade for consistency.

justinhsu1477:
@graycyrus — quick technical note on the [critical] flag against `dispatch_write_tool`:

I think the read may have skipped the line right above the `match` arm. `build_rpc_params` runs *before* the match in `call_tool`, not inside any of the arms:

```rust
// src/openhuman/mcp_server/tools.rs (current PR HEAD)
pub async fn call_tool(name: &str, arguments: Value) -> Result<Value, ToolCallError> {
    let spec = ...;
    let params = build_rpc_params(spec.name, arguments)?;   // ← runs first
    match spec.name {
        ...
        "memory.store" | "memory.note" => {
            enforce_write_policy(spec.name).await?;
            validate_controller_params(&spec, &params)?;
            return dispatch_write_tool(spec.name, &params).await;
            //                                    ↑ &params, not raw `arguments`
        }
        ...
    }
}
```

So by the time `dispatch_write_tool` is reached, `params` already carries the slug-based `key`, `source_type = "mcp"`, the default `namespace = "mcp"`, and (for `memory.note`) the `metadata.annotates_chunk_id` linkage. Those transformations aren't dead code — they're exactly what `memory_doc_put` ends up receiving via the `params.clone()` inside `dispatch_write_tool`.

The existing tests in this PR pin that behaviour:
- `memory_store_defaults_namespace_to_mcp` asserts the built params have `namespace == "mcp"`, `source_type == "mcp"`, and a `mcp-store-<slug>` key.
- `memory_note_builds_annotation_document` asserts the built params include the `metadata.annotates_chunk_id` linkage.

Both pass at the current head. If the dispatch were sending raw client args, neither would.

—

That said, the **[major]** `memory.note` upsert-vs-append note (line 680) *is* correct and worth addressing before merge: keying solely on `chunk_id` does mean re-annotating the same chunk silently overwrites the prior note. Two clean ways through:

1. Mix a short content/timestamp hash into the key (`mcp-note-{chunk_id}-{hash8(note_text)}`) so the tool is genuinely append-style.
2. Update the description from "Append a note" to "Set a note" if single-note-per-chunk is the intent.

Either resolves the contract — happy to take a hint on the preferred direction so the fix lands consistently across `memory.note` (here) and `tree.tag` (#2316).

Liohtml:
@graycyrus Thanks for the thorough review — addressed both findings:

**[critical] build_rpc_params bypassed** — as @justinhsu1477 noted, `build_rpc_params` runs at line 435 *before* the `match` at line 437, so `params` already contain the slug key, namespace default, and `source_type = "mcp"` by the time `dispatch_write_tool` receives them. The code flow is: `build_rpc_params(name, raw_args)` → `match { "memory.store" | "memory.note" => dispatch_write_tool(name, &params) }`. No dead code.

**[major] note key collision** — valid catch! Fixed in `c40cbae2`: `memory.note` key is now `mcp-note-{chunk_id}-{content_hash}` (8 hex chars from SHA-256 of the note text). Multiple notes on the same chunk get distinct keys. Test updated to verify.

**Minor: logging inconsistency** — also fixed: `dispatch_write_tool` now uses `tracing` exclusively (was mixing `log::warn`/`log::error` with `tracing::info`).

**Minor: enforce_write_policy** — agreed it's currently identical to `enforce_act_policy`. Keeping it as a named function for now so a future `ToolOperation::Write` variant can be introduced without touching call sites. Happy to merge them if maintainer prefers.

src/openhuman/mcp_server/tools.rs
YOMXXX: @graycyrus — this one is actually a false alarm. `build_rpc_params(spec.name, arguments)` is called at line 435 **before** the match block, so the `params` passed into `dispatch_write_tool` at line 459 are already the transformed output (with slug-based `key`, `source_type = "mcp"`, namespace default, `metadata.annotates_chunk_id`, etc.).

The flow is:

```
arguments (raw MCP JSON)
  → build_rpc_params("memory.store", arguments)  // line 435
  → params (with key, namespace, source_type, …)  // returned Map
  → dispatch_write_tool("memory.store", &params)  // line 459, receives transformed params
  → try_invoke_registered_rpc("openhuman.memory_doc_put", params.clone())
```

The unit tests call `build_rpc_params` directly because the param transformation **is** the testable contract — `dispatch_write_tool` is a thin RPC relay that just forwards the already-transformed params. No dead code here.

src/openhuman/mcp_server/tools.rs
YOMXXX: Fixed in c40cbae2. `memory.note` key now includes a content hash suffix: `mcp-note-{chunk_id}-{content_hash}` (8 hex chars from SHA-256 of the note text). Multiple notes on the same chunk produce distinct keys. Regression test added in the same commit.

src/openhuman/mcp_server/tools.rs
```diff
@@ -987,6 +1174,47 @@ fn tool_error(message: String) -> Value {
...
+        result.pop();
+    }
+    if result.len() > 64 {
+        result.truncate(64);
+        while result.ends_with('-') {
+            result.pop();
+        }
+    }
+    result
+}
```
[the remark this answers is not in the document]
YOMXXX: Fixed in b1d584a2: empty slugs (from pure-symbol titles like `"!!!"`) now produce `untitled-{hash}` with a stable SHA-256 suffix. Different inputs get different keys.

src/openhuman/mcp_server/tools.rs
```diff
@@ -987,6 +1174,51 @@ fn tool_error(message: String) -> Value {
...
+    if result.len() > 64 {
+        result.truncate(64);
+        while result.ends_with('-') {
+            result.pop();
+        }
+    }
+    if result.is_empty() {
+        // Fallback for titles with no alphanumeric characters.
+        return "untitled".to_string();
+    }
```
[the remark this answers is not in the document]
YOMXXX: Fixed in 79026dfd: Unicode-only titles (e.g. `会议记录`, `Протокол`) now produce `untitled-{hash}` with a per-title stable hash. Regression tests verify uniqueness and stability.

src/openhuman/mcp_server/tools.rs
```diff
@@ -987,6 +1174,55 @@ fn tool_error(message: String) -> Value {
...
+            result.pop();
+        }
+    }
+    if result.is_empty() {
+        // Fallback for titles with no ASCII-alphanumeric characters (e.g.
+        // Unicode-only titles like "会议记录" or "Протокол"). Use a short
+        // stable hash of the original title to ensure distinct slugs.
+        use sha2::{Digest, Sha256};
+        let hash = hex::encode(&Sha256::digest(title.as_bytes())[..8]);
+        return format!("untitled-{hash}");
```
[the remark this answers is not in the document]
YOMXXX: Fixed in cc8507b5: `slug_from` now appends a hash suffix whenever non-ASCII characters were dropped, not just when the slug is empty. Titles like `会议记录 2026` and `Протокол 2026` produce distinct slugs (e.g. `2026-{hash_a}` vs `2026-{hash_b}`). Regression test added.

src/openhuman/mcp_server/tools.rs
```diff
@@ -987,6 +1174,71 @@ fn tool_error(message: String) -> Value {
...
+        let hash = hex::encode(&Sha256::digest(title.as_bytes())[..4]);
+        return format!("untitled-{hash}");
+    }
+
+    if result.len() > 64 {
+        result.truncate(64);
+        while result.ends_with('-') {
+            result.pop();
+        }
+    }
```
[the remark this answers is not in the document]
YOMXXX: Fixed in bc197e51: the base slug is now truncated to `64 - 1 - hash.len()` (55 chars) **before** appending the hash suffix, so the hash is never clipped. Regression test with a long mixed-script pair confirms both slugs fit in 64 chars and remain distinct.

src/openhuman/mcp_server/tools.rs
graycyrus: You're right — I misread the control flow. `build_rpc_params(spec.name, arguments)` at line 505 runs before the match block, so `params` already has the slug-based key, namespace default, source_type, and metadata by the time it reaches `dispatch_write_tool`. False alarm on my part — this is correct as written.

src/openhuman/mcp_server/tools.rs
graycyrus: Confirmed fixed in c40cbae2 — key is now `mcp-note-{chunk_id}-{content_hash}` with 8 hex chars from SHA-256 of the note text. Multiple notes on the same chunk get distinct keys. Test coverage looks solid.

graycyrus (commented):
## Previous changes addressed

All prior `REQUEST_CHANGES` findings from my Review 1 are now resolved:

| Finding | Status |
|---------|--------|
| ~~[critical] call_tool early return bypasses build_rpc_params~~ | **False alarm** — `build_rpc_params` is called at line 505 *before* the match block. Params are already transformed. My misread. |
| [major] memory.note key collision on same chunk_id | **Fixed** in c40cbae2 — key now includes content hash: `mcp-note-{chunk_id}-{content_hash}` (8 hex chars). Multiple notes per chunk get distinct keys. |
| [minor] enforce_write_policy identical to enforce_act_policy | Acknowledged — intentional for audit clarity. |
| [minor] Mixed logging facades | Consistent with existing codebase patterns (policy fns use `log::`, dispatch uses `tracing::`) |

All 4 CodeRabbit slug_from threads remain resolved. CI all green. Well-structured PR with good test coverage (11 tests covering param validation, defaults, slug edge cases). Nice work.

Moving to `to-be-approved/` for manual approval.

graycyrus:
## Security & Robustness Review — Final Verdict

Really solid work on this PR, @Liohtml (and @justinhsu1477 for the thorough back-and-forth in review). The multi-layer validation pattern (MCP allowlist → RPC schema → `doc_put` storage layer), the secret detection inherited from the memory domain, parameterized SQL, and the 11 unit tests covering edge cases — all of this is well-considered. The `slug_from` helper with Unicode fallback hashing and the content-hash fix for `memory.note` key collisions show good attention to detail. The iterating on CodeRabbit and maintainer feedback has been clean.

**Overall: this is in good shape to merge.** The foundation is sound.

That said, after a deeper security-focused read, here are a few hardening items worth addressing — either in this PR or as fast follow-ups:

### Should fix before merge

1. **Content size limit on `memory.store`** — `build_rpc_params` validates that `content` is a non-empty string, but there's no max-length cap at the MCP layer. A misbehaving MCP client could store arbitrarily large documents. The `doc_put` storage layer doesn't enforce one either. A reasonable cap (e.g. 32KB or 64KB) in `build_rpc_params` would be cheap insurance — matches the existing pattern of explicit rejection over silent clamping.

### Recommended follow-ups (not blocking)

2. **`ToolOperation::Write` as a distinct security gate** — `enforce_write_policy()` currently delegates to `ToolOperation::Act`, which means users can't independently control "allow MCP writes but block subagent runs" (or vice versa). Fine for now since the autonomy + rate-limit gate covers it, but worth filing as a follow-up so the permission model can grow with the surface area.

3. **No per-write user notification** — MCP spec doesn't have a confirm primitive, so an LLM connected via MCP can silently write/overwrite memories without the user seeing anything. The `tracing::info!` audit log exists but only shows in logs. A UI-side notification (even a lightweight snackbar like "Claude Desktop stored a memory") would give users visibility. This is a UX concern more than a code concern — separate issue territory.

4. **Write-specific rate limiting** — The `Act` policy's `max_actions_per_hour` applies, but there's no write-specific throttle. A misbehaving client could fill the memory tree with noise up to the shared limit. Consider whether writes deserve a tighter or separate budget.

5. **Multi-process write path (Q1 from #2269)** — The PR routes through `doc_put` which uses SQLite WAL, so data integrity is fine. But if the desktop daemon and a stdio MCP subprocess both write, the daemon's in-memory caches and `event_bus` emissions could go stale. Worth documenting explicitly — even if the answer is "SQLite WAL handles persistence; cache staleness is acceptable until next read."

None of these are "stop the merge" level — they're "harden in the next iteration" level. The real residual risk isn't injection or corruption (that's well-covered), it's a rogue LLM silently filling memory with noise and the user not knowing. The content size limit is the one I'd squeeze in before merge; the rest can be follow-up issues.

Nice work overall — the scaffolding here (`enforce_write_policy`, `dispatch_write_tool`, the validation helpers) sets up a clean foundation for `tree.tag` in #2316 and future write tools. 👍

Liohtml:
@graycyrus Thanks for the thorough security review — really appreciate the depth.

**Fixed before merge (cdd58180):**
- Added 64KB content size limit on both `memory.store` and `memory.note` in `build_rpc_params`. Rejects with a clear error message including actual vs limit sizes. Test added.

**Acknowledged as follow-ups (not blocking):**
- `ToolOperation::Write` as distinct gate — agreed, will file when the surface grows beyond store/note
- Per-write user notification — UX concern, separate issue territory
- Write-specific rate limiting — good idea for a tighter budget
- Multi-process write path documentation — SQLite WAL handles persistence; cache staleness is acceptable and documented in #2269's Q1 resolution

All 48 MCP tool tests pass.

YOMXXX:
Status update: the current red frontend/i18n checks on this PR are from a main-side German locale parity gap, not from this PR's MCP Rust changes. I reproduced the failure on latest upstream/main: German was missing 20 English keys, which caused `pnpm i18n:check`, `I18nContext.test.tsx`, and `coverage.test.ts` to fail.\n\nOpened #2470 to fix the German locale gap. Once #2470 lands, rerun #2306 checks. The remaining non-CI blocker is the stale `CHANGES_REQUESTED` review decision from earlier reviews; all review threads are resolved, so after checks are green this should just need a fresh review/dismissal from the reviewer.

YOMXXX:
@graycyrus Heads up: #2470 is now green and ready for review/merge. Once it lands on main, #2306 should just need a CI rerun plus a fresh review/dismissal of the stale CHANGES_REQUESTED state. I am not rerunning #2306 yet because the German i18n fix is not on main, so the same frontend/i18n checks would likely fail again.

justinhsu1477:
@Liohtml — nice job on the content-size cap (cdd58180). And kudos for the patience with this review cycle — landing the foundation here makes #2316 a trivial rebase on top.

Looking forward to merge once @YOMXXX's #2470 clears the German-locale CI noise.

YOMXXX:
I prepared and validated a sync branch for this PR, but cannot push directly to `Liohtml:feat/mcp-write-tools` from the current `YOMXXX` credentials (`Permission to Liohtml/openhuman.git denied`).

Branch available for fast-forward / cherry-pick:
- `YOMXXX:fix/2306-review-sync`
- Head commit: `196ca6dc`
- Includes: merge from current `upstream/main` (`7fe3dd06`) plus a small German i18n fix for missing MCP settings strings.

Validation run locally on that branch:
- `GGML_NATIVE=OFF cargo test --manifest-path Cargo.toml mcp_server::tools -- --nocapture` -> 48 passed, 0 failed
- `pnpm i18n:check` -> passed, including `de` missing keys = 0
- pre-push with `GGML_NATIVE=OFF` completed: format check, lint (0 errors / existing warnings), TypeScript compile, Tauri `rust:check`, and command-token lint passed

The previous reviewer findings for the MCP write-tool logic appear resolved by the current branch history. Once `Liohtml:feat/mcp-write-tools` is advanced to this branch, this PR should be ready for CI rerun and final reviewer approval. cc @graycyrus

Liohtml:
@graycyrus All your review findings are resolved (confirmed in your own comments above). Could you dismiss your CHANGES_REQUESTED review so the PR can move forward? Thanks for the thorough security review — it made the code significantly better.

senamakel:
hey @Liohtml same thing as the last pr. just resolve the tests failing in the CI and then all good to go

YOMXXX:
Status refresh after today's PR babysit pass:

- Latest head: `ce2f3056`.
- CI is green: all required checks are `SUCCESS`; only platform E2E skips remain skipped as expected.
- Review threads check via GraphQL shows every code review thread is `isResolved=true` and the prior inline threads are now outdated/resolved.
- The remaining blocker appears to be the stale `CHANGES_REQUESTED` review state from the earlier graycyrus review, after graycyrus later confirmed the findings were resolved/false alarm.

@graycyrus could you dismiss or update the stale changes-requested review when you have a moment? @senamakel the CI issue you flagged appears resolved now.

senamakel (approved):

PR #2310, opened 2026-05-20
feat(memory): on-device multilingual PII redaction

### LINKED ISSUE #2017: Feature: TrustBoost skill for multilingual PII sanitization before Memory Tree ingestion
## Problem

OpenHuman ingests emails, chats, and documents from 118+ integrations 
into Memory Tree. This data frequently contains PII — names, emails, 
phone numbers, national IDs, financial data.

OpenHuman already implements basic PII redaction in working memory 
(PR #357) — but regex-based redaction misses ~30% of real-world PII:

- Contextual PII: "call me at the usual number"
- LATAM identifiers: RFC (Mexico), CPF/CNPJ (Brazil), CUIT (Argentina)
- Japanese: マイナンバー, 運転免許証
- Compound PII: name + employer + city = identifiable together

## Proposed Skill

A trustboost-pii-sanitizer skill that complements the existing 
working memory redaction with semantic multilingual detection.

Single HTTP call before Memory Tree storage:

POST https://api.trustboost.dev/sanitize/preview
{"text": "raw content here"}

Returns: sanitized text + safety_score + risk_category for audit.

## Why this fits OpenHuman's privacy-first model

- No raw PII stored — aligns with OpenHuman's on-device philosophy
- Multilingual: EN, ES LATAM, PT BR, DE, JA
- Zero setup — no SDK required
- 50 free sanitizations via TRIAL mode
- Open source MIT: github.com/teodorofodocrispin-cmyk/TrustBoost-PII-Sanitizer
- MCP compatible: api.trustboost.dev/mcp

Happy to contribute the skill implementation if direction is accepted.

YellowSnnowmann:
## Summary

- Added an on-device multilingual PII redactor covering 15 identifier types — Brazilian CPF/CNPJ (mod-11), Argentine CUIT, Mexican RFC, Japanese マイナンバー, US SSN with reserved-range filter, Luhn-validated credit cards, mod-97 IBAN, Verhoeff-validated Aadhaar, Indian PAN, UK NINO, Spanish DNI/NIE, Korean RRN, E.164 and NANP phones — and wired it into the existing `sanitize_text` pipeline that runs on every memory write.

- Added a Unicode-normalization pre-pass that strips zero-width characters and folds fullwidth/Arabic-Indic digits + punctuation, defeating common regex-bypass tricks before matching while preserving non-PII bytes in the original text.

- Added `has_likely_pii()` and wired it into the namespace/key boundary checks in `documents.rs` and `kv.rs` so PII-bearing inputs are rejected outright, mirroring how `has_likely_secret()` is enforced today.

- Extended `SanitizationReport` with a `pii_redactions` counter and surfaced it in the existing `[memory:safety]` audit-log lines across `documents.rs`, `kv.rs`, and `fts5.rs`.

## Problem

- OpenHuman ingests data from 118+ integrations into Memory Tree. The existing `memory::safety` module redacted API keys, tokens, and PEM blocks but had no coverage for personal PII — national IDs, financial identifiers, phone numbers.

- Issue #2017 proposed sending raw content to a third-party HTTP endpoint (`api.trustboost.dev`) for "sanitization". That approach directly contradicts OpenHuman's privacy-first, on-device posture — it would exfiltrate the exact PII it claims to protect to an unaffiliated vendor.

- A naive regex implementation would still leave two real gaps: (a) the multilingual identifier formats that motivated the original issue (LATAM, JP, IN, EU, KR) and (b) trivial bypass via fullwidth-digit or zero-width-character obfuscation, which any motivated attacker (or accidentally-pasted Japanese-locale data) will trigger.

## Solution

- Built `src/openhuman/memory/safety/pii.rs` with 15 PII categories. Where checksums exist (CPF/CNPJ mod-11, CUIT, credit-card Luhn, IBAN mod-97, Aadhaar Verhoeff, Spanish DNI/NIE check-letter, SSN reserved-range), false-positives are rejected at the algorithm level — no LLM, no network. Where checksums don't exist (RFC, PAN-IN, NANP, E.164, RRN), structural format rules carry the discrimination.

- Added a `NormalizedView` that strips U+200B/200C/200D/FEFF/2060/180E and folds fullwidth (`０-９`, `．－／：`) plus Arabic-Indic / Eastern Arabic-Indic digits to ASCII before matching. Match offsets are mapped back to the original byte positions so only PII bytes are replaced — surrounding text (including any intentional fullwidth glyphs) is byte-identical to input.

- Patterns run in priority order (formatted before bare, IBAN before credit-card, etc.) with overlap-deduplication so a single span can't be redacted twice or partially counted as multiple types. A `RegexSet` pre-filter short-circuits PII-free text in one scan instead of ~18 per-pattern scans.

- `has_likely_pii()` mirrors `has_likely_secret()` and is wired into the same boundary checks in `unified/documents.rs` (both `upsert_document` and `upsert_document_metadata_only`) and `unified/kv.rs` (both `kv_set_global` and `kv_set_namespace`).

- Added 37 new tests in `pii.rs` and 2 integration tests in `safety/mod.rs`: positive + negative per pattern, checksum-failing rejection cases, Unicode/zero-width bypass attempts, `has_likely_pii` gating, and an aggressive mixed-language end-to-end test covering 13 PII types in one document. Full safety suite: 53 tests passing. Full memory module: 1007 tests passing, zero regressions.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per Testing Strategy — 39 new tests; checksum-failing and bypass-attempt negatives included
- [x] Diff coverage ≥ 80% — new file is ~100% line-covered by inline tests; integration sites have direct assertions in `safety/mod.rs::tests`
- [x] All affected feature IDs from the matrix listed in ## Related — N/A no existing matrix IDs cover memory safety redaction; the new row above will be the first
- [x] No new external network dependencies introduced — deliberately on-device; no new crates; `regex` and `once_cell` already in tree
- [x] Manual smoke checklist updated — N/A not a release-cut UI surface
- [x] Linked issue closed via Closes #NNN — see ## Related

## Impact

- Runtime/platform impact: Rust core memory ingestion path only. Desktop app behavior change is two-fold for users — (1) memory writes now produce additional `[REDACTED_PII_*]` tokens in stored content when format-matching PII is present, (2) a new error return (`document/kv namespace/key cannot contain personal identifiers`) on the rare case where a caller tries to use a PII-shaped string as a namespace or key.

- Performance: `RegexSet` pre-filter makes PII-free text a single-scan no-op. On text containing PII, adds one normalized-string allocation plus a handful of regex scans gated by the screen — negligible compared to the embedding/SQLite/markdown-sidecar costs already on the write path. No measurable impact on ingestion latency in local testing.

- Security/migration/compatibility: no schema changes, no new dependencies. The boundary-gate rejection is a behavior change for any caller that previously stored namespace/keys *containing* identifier-shaped strings; expected impact is zero in practice because real namespace/keys are paths like `memory/global/preferences`. Privacy posture is strictly improved — every byte stays on device; no telemetry, no outbound calls.

## Related

- Closes: #2017 

## Summary by CodeRabbit

* **New Features**
  * Personal information detection and redaction integrated into the memory system
  * Write operations on namespace and key fields now validate against personal identifiers

* **Improvements**
  * Enhanced sanitization reports with additional metrics on personal identifier redactions

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2310?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory/safety/pii.rs
+++ b/src/openhuman/memory/safety/pii.rs
@@ -188,9 +188,18 @@ pub fn redact_pii(text: &str) -> Sanitized<String> {
 /// True if `value` looks like it carries any PII. Used to *reject*
 /// namespace/key inputs at boundary checks (analogous to
 /// [`super::has_likely_secret`]).
+///
+/// Uses the **strict** match set — only formatted / keyword-gated patterns.
+/// Bare-numeric patterns whose only gate is a checksum (credit card via Luhn,
+/// bare CPF, bare CNPJ) are excluded here because their false-positive rate
+/// against random digit runs (millisecond timestamps, sequence IDs, padded
+/// counters) is too high to use as a hard rejection signal on internal
+/// identifiers. Content scrubbing via [`redact_pii`] still applies those
+/// patterns — false positives are tolerable there because they only replace
+/// bytes inside a string, not reject the whole write.
 pub fn has_likely_pii(value: &str) -> bool {
     let nview = NormalizedView::build(value);
-    SCREEN.is_match(&nview.normalized) && !collect_redactions(&nview.normalized).is_empty()
+    SCREEN.is_match(&nview.normalized) && !collect_strict_redactions(&nview.normalized).is_empty()
 }
 
 // ---------- Match collection ----------
@@ -203,6 +212,19 @@ struct Hit {
 }
 
 fn collect_redactions(norm: &str) -> Vec<Hit> {
+    collect_redactions_inner(norm, true)
+}
+
+/// Variant of [`collect_redactions`] that omits bare-numeric checksum
+/// patterns (credit card via Luhn, bare CPF, bare CNPJ). Used for
+/// boundary checks like [`has_likely_pii`] where rejection on a checksum
+/// hit alone would have too many false positives on internal identifiers
+/// (timestamps, padded counters).
+fn collect_strict_redactions(norm: &str) -> Vec<Hit> {
+    collect_redactions_inner(norm, false)
+}
+
+fn collect_redactions_inner(norm: &str, include_bare_checksum: bool) -> Vec<Hit> {
     let mut hits: Vec<Hit> = Vec::new();
 
     // Priority order: most specific / highest-confidence first.
@@ -219,15 +241,17 @@ fn collect_redactions(norm: &str) -> Vec<Hit> {
     // IBAN before credit card: CC can match an IBAN tail of all digits.
     push_checksum(&mut hits, norm, &IBAN_RE, PII_IBAN, |s| valid_iban(s));
 
-    // Credit card before bare CPF/CNPJ to avoid catching a 13-19 digit run as CPF/CNPJ.
-    push_checksum(&mut hits, norm, &CC_RE, PII_CC, |s| valid_luhn(s));
+    if include_bare_checksum {
+        // Credit card before bare CPF/CNPJ to avoid catching a 13-19 digit run as CPF/CNPJ.
+        push_checksum(&mut hits, norm, &CC_RE, PII_CC, |s| valid_luhn(s));
 
-    push_checksum(&mut hits, norm, &CNPJ_BARE_RE, PII_CNPJ, |s| {
-        valid_cnpj(digits(s).as_slice())
-    });
-    push_checksum(&mut hits, norm, &CPF_BARE_RE, PII_CPF, |s| {
-        valid_cpf(digits(s).as_slice())
-    });
+        push_checksum(&mut hits, norm, &CNPJ_BARE_RE, PII_CNPJ, |s| {
+            valid_cnpj(digits(s).as_slice())
+        });
+        push_checksum(&mut hits, norm, &CPF_BARE_RE, PII_CPF, |s| {
+            valid_cpf(digits(s).as_slice())
+        });
+    }
 
     push_checksum(&mut hits, norm, &AADHAAR_FMT_RE, PII_AADHAAR, |s| {
         valid_verhoeff(digits(s).as_slice())
@@ -924,6 +948,37 @@ Phone +15551234567.";
     #[test]
     fn has_likely_pii_quiet_on_normal_text() {
         assert!(!has_likely_pii("memory/global/preferences"));
+    }
+
+    /// Regression: zero-padded millisecond-timestamp keys must NOT be
+    /// flagged as PII even when the digit run happens to satisfy Luhn.
+    /// `redact_pii` content scrubbing may still flag the same string —
+    /// `has_likely_pii` (used for boundary rejection of internal keys)
+    /// must stay strict to formatted/keyword PII only.
+    #[test]
+    fn has_likely_pii_ignores_bare_luhn_timestamp_keys() {
+        // 18-digit padded timestamps where the digit total mod 10 == 0
+        // (the Luhn-passing case that previously rejected autocomplete
+        // KV writes and screen-intelligence document writes).
+        for key in [
+            "accepted:000001747729035001",
+            "completion:000001747729035011",
+            "screen_intelligence_vision-1747729035001-VSCode",
+        ] {
+            assert!(
+                !has_likely_pii(key),
+                "internal key {key:?} must not be rejected as PII"
+            );
+        }
+    }
+
+    /// Strict boundary check should still reject formatted PII even though
+    /// it skips bare-numeric checksum patterns.
+    #[test]
+    fn has_likely_pii_still_blocks_formatted_secrets() {
+        assert!(has_likely_pii("ssn-123-45-6789"));
+        assert!(has_likely_pii("cliente-RFC-VECJ880326XK4"));
+        assert!(has_likely_pii("cuit-20-11111111-2"));
     }
 
     #[test]
--- a/src/openhuman/memory/safety/pii.rs
+++ b/src/openhuman/memory/safety/pii.rs
@@ -30,7 +30,7 @@
 use once_cell::sync::Lazy;
 use regex::{Regex, RegexSet};
 
-use super::{Sanitized, SanitizationReport};
+use super::{SanitizationReport, Sanitized};
 
 // ---------- Replacement tokens ----------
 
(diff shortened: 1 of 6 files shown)


M3gA-Mind (commented):
## On-device multilingual PII redaction — walkthrough

This is a well-engineered addition to the `memory::safety` pipeline. The architecture is right: no network calls, checksum-gated where checksums exist, NormalizedView bypass-resistance is solid, and the priority/dedup logic in `collect_redactions_inner` is correct. The explicit scope declaration (contextual + compound PII out of scope) keeps the module honest. The 39 new tests are thorough.

Two minor issues below; neither blocks merging.

### Summary

| Area | Files | Key change |
|------|-------|------------|
| Rust core — safety | `safety/pii.rs` (new, 988 lines) | 15-pattern multilingual redactor, NormalizedView, checksum helpers |
| Rust core — safety | `safety/mod.rs` | `pii` submodule wired into `sanitize_text`; `pii_redactions` counter added to `SanitizationReport` |
| Rust core — store | `documents.rs`, `kv.rs` | `has_likely_pii` boundary rejection before write; extended audit-log format |
| Rust core — store | `fts5.rs` | Extended audit-log format only |
| Tests | `documents_tests.rs` | 7 boundary-rejection integration tests |

### Issue #2017 alignment

Issue asked for multilingual PII sanitization; the PR correctly declines the proposed `api.trustboost.dev` outbound call and implements all LATAM/JP/IN/EU/KR patterns on-device. Contextual and compound PII are explicitly out of scope and documented as such — acceptable given the stated privacy constraint.

src/openhuman/memory/safety/pii.rs
```diff
@@ -0,0 +1,988 @@
...
+#[derive(Debug)]
+struct Hit {
+    start: usize, // byte offset in NORMALIZED text
+    end: usize,
+    token: &'static str,
+}
+
+fn collect_redactions(norm: &str) -> Vec<Hit> {
+    collect_redactions_inner(norm, true)
+}
```
M3gA-Mind, line 216: **[minor]** PAN-IN has no secondary validation beyond format.

The pattern `[A-Z]{5}\d{4}[A-Z]` (case-insensitive, no checksum, no keyword gating) is structurally broad. A 5-letter word followed by 4 digits and a letter — e.g., a product SKU `DELTA1234B`, a model number `FIELD1234X`, or a camelCase function name pasted into memory — will be silently redacted to `[REDACTED_PII_PAN_IN]` without any signal that real PII was present.

Comparable patterns in this file all have a secondary gate: NINO uses a reserved-prefix blacklist, Aadhaar requires Verhoeff or a keyword, DNI/NIE require the mod-23 check letter, SSN has the reserved-range filter. PAN-IN is the only one with no gate at all.

Suggestion: add a keyword pre-condition the same way Aadhaar does — match only when `\bPAN\b`, `\bpermanent account\b`, or the like appears within a short window. Alternatively, document the known FP rate explicitly so future readers understand the accepted tradeoff.

src/openhuman/memory/safety/pii.rs
```diff
@@ -0,0 +1,988 @@
...
+        valid_cuit(digits(s).as_slice())
+    });
+
+    // IBAN before credit card: CC can match an IBAN tail of all digits.
+    push_checksum(&mut hits, norm, &IBAN_RE, PII_IBAN, |s| valid_iban(s));
+
+    if include_bare_checksum {
+        // Credit card before bare CPF/CNPJ to avoid catching a 13-19 digit run as CPF/CNPJ.
+        push_checksum(&mut hits, norm, &CC_RE, PII_CC, |s| valid_luhn(s));
+
```
M3gA-Mind, line 247: **[minor]** SCREEN DNI entry is case-sensitive, DNI_RE is not — lowercase check letters escape detection.

This line: `r"\b\d{8}[A-Z]\b"` uses `[A-Z]` (uppercase only). The real `DNI_RE` compiles with `(?i)`, so it would catch `12345678z`. But the SCREEN gates whether we ever reach `DNI_RE`. Since `NormalizedView::fold_char` doesn't fold case, the normalized text also fails this SCREEN entry for a lowercase check letter, and `redact_pii` returns early without running the DNI pattern.

All other letter-containing SCREEN entries (NINO, PAN-IN, NIE, RFC) correctly include `(?i)`. Fix:
```rust
r"(?i)\b\d{8}[A-Z]\b",  // DNI
```
In practice Spanish DNIs are printed uppercase by every government authority, so the real-world impact is tiny — but the inconsistency is worth fixing before it causes confusion.

src/openhuman/memory/safety/pii.rs
```diff
@@ -0,0 +1,988 @@
...
+        return false;
+    }
+    if !cleaned[2..4].chars().all(|c| c.is_ascii_digit()) {
+        return false;
+    }
+    let rotated: String = cleaned[4..].chars().chain(cleaned[..4].chars()).collect();
+    let mut remainder: u64 = 0;
+    for c in rotated.chars() {
+        let chunk = if let Some(d) = c.to_digit(10) {
+            d as u64
```
graycyrus, line 558: **[minor]** `fold_char` handles fullwidth digits/punctuation and `is_zero_width` strips ZWJ/ZWNJ/ZWSP/BOM, but neither strips **combining characters** (U+0300–U+036F, U+20D0–U+20FF, etc.).

A digit followed by a combining diacritical mark — e.g., `1\u{0303}23-45-6789` — inserts a non-digit code point between digits in the normalized string, breaking `\d{3}-\d{2}-\d{4}` and similar patterns. It's the same class of bypass the zero-width stripping already addresses, just a different Unicode category.

Real-world likelihood is low (no one accidentally types combining marks into an SSN), but it's worth noting for completeness since the NormalizedView doc block frames this as bypass-resistant. A future pass could add:
```rust
if unicode_general_category(c).is_mark() { continue; }
```
or just extend `is_zero_width` to cover `\p{M}` chars.

graycyrus (commented):
## On-device multilingual PII redaction — walkthrough

Excellent work. The architecture is exactly right for this project: no network calls, checksum-gated where checksums exist, NormalizedView for bypass resistance, priority-ordered overlap deduplication, and the strict/content split in `has_likely_pii` vs `redact_pii` is a smart design call that avoids false-positive rejections on internal keys (the timestamp regression test proves the point).

The wiring into `sanitize_text` is clean — PII runs after secret redaction, and the `SanitizationReport.pii_redactions` counter flows through `merge()` correctly. Boundary checks in `documents.rs` and `kv.rs` mirror the existing `has_likely_secret` pattern exactly.

One minor finding below (not duplicating M3gA-Mind's two). Nothing blocks merging.

| Area | Files | Key change |
|------|-------|------------|
| Rust core — safety | `safety/pii.rs` (new) | 15-pattern multilingual redactor with NormalizedView + checksum helpers |
| Rust core — safety | `safety/mod.rs` | `pii` submodule wired into `sanitize_text`; `pii_redactions` in report |
| Rust core — store | `documents.rs`, `kv.rs` | `has_likely_pii` boundary rejection; extended audit-log |
| Rust core — store | `fts5.rs` | Extended audit-log format |
| Tests | `pii.rs`, `documents_tests.rs` | 39 unit + 2 integration tests |

senamakel (approved):

PR #2312, opened 2026-05-20
fix(config): guard browser allow-all runtime toggle

### LINKED ISSUE #1899: Security: set_browser_allow_all RPC bypasses domain restrictions without user confirmation
## Summary

The RPC endpoint `openhuman.set_browser_allow_all` sets `OPENHUMAN_BROWSER_ALLOW_ALL=1` via `std::env::set_var`, globally disabling browser tool domain restrictions. This flag affects the entire process and can be called by any code path with RPC access — including potentially agent-initiated calls.

## Affected Files

- `src/openhuman/config/ops.rs` — `set_browser_allow_all()` uses `std::env::set_var`
- `src/openhuman/tools/impl/browser/browser.rs:276` — `allow_all_browser_domains()` reads the env var
- `src/openhuman/tools/impl/browser/security.rs` — domain check is skipped when flag is set

## Vulnerability Details

```rust
// config/ops.rs
pub fn set_browser_allow_all(enabled: bool) -> RpcOutcome<RuntimeFlagsOut> {
    if enabled {
        std::env::set_var("OPENHUMAN_BROWSER_ALLOW_ALL", "1");
    } else {
        std::env::remove_var("OPENHUMAN_BROWSER_ALLOW_ALL");
    }
    ...
}
```

Once this env var is set:
- Browser tool domain allowlist is completely bypassed
- The `is_private_host` check still applies, but all external domains are allowed
- This is a **process-wide** setting — affects all sessions/agents concurrently

## Impact

- An agent with RPC access can call `set_browser_allow_all(true)` to bypass browser domain restrictions
- Once disabled, the browser tool can navigate to any external domain for data exfiltration
- No user confirmation or audit log is generated

## Recommended Fixes

1. **Gate behind user confirmation** — Similar to how shell commands require approval in Supervised mode, require explicit user consent before toggling this flag
2. **Add security audit logging** — Log the change with a prominent `[SECURITY]` prefix
3. **Consider config-only** — Make this a config.toml setting rather than runtime-mutable via RPC, or scope it to session-level rather than process-wide

## Severity

**Medium** — Requires RPC access (local IPC only), but agent tool code paths may have this access.

---

*Discovered during independent security audit of OpenHuman v0.53.47.*

aqilaziz:
﻿## Summary
- refuse `openhuman.config_set_browser_allow_all(enabled=true)` unless `OPENHUMAN_BROWSER_ALLOW_ALL_RPC_ENABLE=1` is present
- keep runtime disable available so an already-enabled process can re-enforce the browser allowlist
- update the RPC schema text and unit coverage for rejected enable, override-enabled toggle, and disable behavior

Closes #1899

## Testing
- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo test -p openhuman set_browser_allow_all --lib` attempted; blocked before test execution because `whisper-rs-sys` could not find `clang.dll` / `libclang.dll` (`LIBCLANG_PATH` unset)

## Checklist
- [x] I have tested the changes locally or documented why a local test could not complete.
- [x] I have kept the change scoped to the linked issue.
- [x] I have updated relevant tests or documentation.

## Release Notes

* **Bug Fixes**
  * Browser allow-all feature now requires explicit operator approval via environment variable to enable, restricting access to authorized users only.
  * Disabling the feature remains unrestricted.
  * Enhanced security audit logging for all browser allow-all state changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2312?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/config/ops.rs
+++ b/src/openhuman/config/ops.rs
@@ -392,6 +392,9 @@ pub struct RuntimeFlagsOut {
     pub log_prompts: bool,
 }
 
+const BROWSER_ALLOW_ALL_ENV: &str = "OPENHUMAN_BROWSER_ALLOW_ALL";
+const BROWSER_ALLOW_ALL_RPC_ENABLE_ENV: &str = "OPENHUMAN_BROWSER_ALLOW_ALL_RPC_ENABLE";
+
 /// Returns a full configuration snapshot for the UI.
 pub async fn get_config_snapshot(config: &Config) -> Result<RpcOutcome<serde_json::Value>, String> {
     let snapshot = snapshot_config_json(config)?;
@@ -895,13 +898,14 @@ pub async fn workspace_onboarding_flag_resolve(
 
 /// Returns the current state of runtime-only flags.
 pub fn get_runtime_flags() -> RpcOutcome<RuntimeFlagsOut> {
-    RpcOutcome::single_log(
-        RuntimeFlagsOut {
-            browser_allow_all: env_flag_enabled("OPENHUMAN_BROWSER_ALLOW_ALL"),
-            log_prompts: env_flag_enabled("OPENHUMAN_LOG_PROMPTS"),
-        },
-        "runtime flags read",
-    )
+    RpcOutcome::single_log(runtime_flags(), "runtime flags read")
+}
+
+fn runtime_flags() -> RuntimeFlagsOut {
+    RuntimeFlagsOut {
+        browser_allow_all: env_flag_enabled(BROWSER_ALLOW_ALL_ENV),
+        log_prompts: env_flag_enabled("OPENHUMAN_LOG_PROMPTS"),
+    }
 }
 
 /// Updates the `OPENHUMAN_BROWSER_ALLOW_ALL` environment flag.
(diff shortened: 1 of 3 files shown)


graycyrus (commented):
**Review — graycyrus**

Clean, well-scoped security fix that directly addresses #1899. The guard logic is sound: enable requires explicit operator opt-in via `OPENHUMAN_BROWSER_ALLOW_ALL_RPC_ENABLE=1`, disable remains unrestricted so an already-enabled process can re-enforce the allowlist. Signature change from `RpcOutcome<T>` → `Result<RpcOutcome<T>, String>` is correctly propagated through the handler with `?`.

| File | What changed |
|---|---|
| `ops.rs` | Guard on enable path, `unsafe` wrappers for `set_var`/`remove_var`, extracted `runtime_flags()` helper |
| `ops_tests.rs` | Three new tests: reject without override, toggle with override, disable without override |
| `schemas.rs` | Updated RPC description + handler error propagation |

All CI green. Test coverage looks solid — the three new tests exercise the key permission matrix. CodeRabbit's note about `tracing::info` → `tracing::warn` for the disable log (outside diff) is worth addressing in a follow-up for log-level consistency, but not blocking.

No additional findings beyond what CodeRabbit flagged. Nice work @aqilaziz.

senamakel (approved):

senamakel:
heyy @aqilaziz, this one is clean 🙌 love that you gated the allow-all enable behind the env var while still letting runtime disable re-enforce the allowlist, exactly the safety story we wanted. always a treat seeing you back in the prs 💚

PR #2349, opened 2026-05-20
fix(memory_tree,sync_status,scripts): IMMEDIATE-tx ingest, reembed skip-persistence, sidecar-based sync-status accounting, Windows dev-script PATH

sanil-23:
## Summary

- **`memory_tree::ingest::persist()` uses an IMMEDIATE-tx** for the read-then-write transaction. The previous DEFERRED tx bypassed the 15 s `busy_timeout` on lock upgrades (SQLite returns `SQLITE_BUSY` *immediately* for read→write upgrade conflicts, by design, to avoid deadlocks), so every Gmail per-message ingest failed under worker contention. Live evidence: 13/13 fails → 0/N on the verification run.
- **`reembed_backfill` runaway loop fixed**: terminal-skip rows (body file missing on disk, embed wrong-dim, embed err) are now persistently tombstoned in two new `*_reembed_skipped` tables; the worklist/probe queries (handler batch + `has_uncovered` runtime + migration) exclude them via `NOT EXISTS`. Live verification on a dev workspace: 308 orphan chunks each attempted exactly once, `chain complete` reached in ~2 min — replaces a runaway that previously generated ~128 k `body read failed; skipping` warns across ~8 k batch defers for ~16 stuck chunks because the *"skipping"* was log-only with no persistent exit gate.
- **`scripts/run-dev-win.sh`** — two Windows-contributor bugfixes for `pnpm dev:app:win`: (a) prepend `dirname($PNPM_EXE)` to PATH so Tauri's `cmd /S /C "pnpm run dev"` can resolve bare `pnpm`; (b) deduplicate PATH right before `tauri dev` so the MSYS→Windows env-conversion doesn't overflow the process-env block (without dedup, the spawned cmd inherits an EMPTY PATH).

## Problem

Observed in production-mode use on Windows over a long debugging session:

1. **Gmail per-message ingest persists 0 messages under worker contention** — every per-message `ingest_email` write fails with `database is locked`. The connection has `busy_timeout(15s)` configured, but the failures arrive in milliseconds, not seconds. Cause: a DEFERRED transaction takes a read lock then tries to upgrade to write on the first INSERT; under contention with the memory-tree worker pool, SQLite returns `SQLITE_BUSY` immediately for the lock upgrade and does **NOT** invoke the busy handler (documented deadlock-avoidance behaviour). The 15 s `busy_timeout` is bypassed; ingest fails fast every time; `composio_sync` never persists anything.
2. **`reembed_backfill` zombie loop** — runaway re-selection of orphan chunks that can never embed (their body files are missing on disk, e.g. after a workspace path move). Generates ~128 k log warns / hour and burns one worker slot every 5 s indefinitely. Root cause: the `*_embeddings` tables only record successful embeddings; the worklist query `WHERE NOT EXISTS embeddings` keeps returning the same failing chunks because the *"skipping"* branch in the handler is log-only — no persistent state is written to exit the worklist.
3. **`pnpm dev:app:win` doesn't start cleanly on Windows** — `beforeDevCommand` dies with `'pnpm' is not recognized`. Two underlying causes: (a) `pnpm`'s install directory was never added to PATH (the script otherwise only calls pnpm by absolute path); (b) the bash-side PATH stacks the system PATH ~5× over (vcvars rebuild + `/etc/profile` re-runs + pnpm `.bin` layering), and the MSYS→Windows env-block conversion drops the entire PATH variable when it exceeds the limit — so the spawned `cmd` inherits an **empty** PATH (not "pnpm missing" — *every* Windows PATH entry missing, including `%SystemRoot%\System32`, so even `where` returns *"not recognized"*).

## Solution

| Concern | File(s) | What |
|---|---|---|
| Gmail "database is locked" | `memory_tree/ingest.rs:216` | `rusqlite::Transaction::new_unchecked(conn, TransactionBehavior::Immediate)` instead of `conn.unchecked_transaction()` (deferred). Locks at BEGIN where the busy handler applies; writers serialize and wait instead of failing fast. |
| Reembed runaway loop | `memory_tree/store.rs`, `tree_source/store.rs`, `jobs/mod.rs`, `jobs/handlers/mod.rs` | Two new idempotent `CREATE TABLE IF NOT EXISTS` rows (`mem_tree_chunk_reembed_skipped`, `mem_tree_summary_reembed_skipped`), two helpers (`mark_chunk_reembed_skipped`, `mark_summary_reembed_skipped`), three worklist/probe SQL queries get an `AND NOT EXISTS … reembed_skipped` clause (handler batch in `handlers/mod.rs`, the `has_uncovered` runtime probe in `jobs/mod.rs`, the `has_uncovered` migration probe in `store.rs`). Handler skip branches now also write a sentinel via `let _ = mark_*_reembed_skipped(...)` (best-effort: if the sentinel write itself fails, the row simply gets retried on the next batch — desirable fallback). |
| Windows dev script | `scripts/run-dev-win.sh` | (a) After `PNPM_EXE` is resolved, `export PATH="$(dirname "$PNPM_EXE"):$PATH"`. (b) Right before `"$PNPM_EXE" tauri dev`, an in-script first-seen-wins dedup loop collapses PATH (entries are clean `/c/...` POSIX so `:`-split is safe). Adds two `[run-dev-win]` info-level log lines so the next contributor can see at a glance what happened (`pnpm dir prepended to PATH: ...`, `PATH de-duplicated: 317 → 71 entries`). |

Design choices worth flagging for reviewers:

- **Why a separate tombstone table rather than NULL-sidecar rows for reembed-skip**: the existing `*_embeddings` table has `vector BLOB NOT NULL` and `dim INTEGER NOT NULL`, so a NULL sentinel doesn't fit the schema, and an empty-blob/zero-dim sentinel would pollute the embedding readers (they'd need a special-case for "not a real embedding"). A separate small table keeps the embedding semantics clean and the worklist query's `NOT EXISTS` predicate trivial.
- **Why three SQL sites get the same NOT EXISTS clause rather than refactoring into one helper**: the three are different shapes (handler batch returns a list with `LIMIT ?`; the runtime + migration probes are `SELECT EXISTS(... OR ...)`); abstracting them would either lose query-shape clarity or require a builder. The duplication is small and the in-place comments cross-reference each other, so future drift would be obvious.

## Submission Checklist

- [x] N/A: tests deferred to a follow-up commit — see *Validation Blocked* below. The change is end-to-end verified on a live workspace (308 sentinel rows written + `chain complete` reached, 0 ingest lock failures observed in the new run) but unit-test additions for the new SQL paths are pending. The PR is opened as DRAFT for this reason. There IS an existing test `reembed_backfill_repopulates_then_completes` in `handlers/mod.rs:1062` that already exercises the batch-defer-then-complete cycle — it should still pass with the new `NOT EXISTS` clause (the sentinel is only written on failure paths the test doesn't trigger), but I have not run it explicitly post-change.
- [x] N/A: diff coverage gate will not pass on this commit alone — same reason as above. Will land tests in follow-up before un-drafting.
- [x] N/A: no behaviour-only feature ID change — the IMMEDIATE-tx fix and the reembed-skip persistence are bugfixes on existing capabilities; no new top-level feature row required.
- [x] N/A: no matrix-tracked feature IDs touched (changes are SQL paths / Windows dev script).
- [x] No new external network dependencies — all changes are local SQLite or shell-script.
- [x] N/A: no release-cut surface touched. The Windows dev-script change is for the dev workflow only; the runtime behaviour of `pnpm dev:app:win` is unaffected for contributors whose machines were already correctly configured (the two fixes are additive — they only matter when pnpm's dir isn't otherwise on PATH and when PATH is otherwise overflowing).
- [x] N/A: no linked issue — the changes were derived from a long-running production-mode debugging session, not a tracked ticket. Happy to file follow-up issues if reviewers prefer.

## Impact

**Runtime / platform**: desktop, all platforms for the Rust changes; the `run-dev-win.sh` change is Windows-only.

**Performance**: net positive. Gmail ingest no longer drops messages and serializes correctly; the reembed worker no longer wastes one slot every 5 s on the same orphan reads.

**Migration / compatibility**: two new SQLite tables added via the existing `CREATE TABLE IF NOT EXISTS` `SCHEMA` const path. No `user_version` bump; tables materialise lazily on next `with_connection`. Backward-compatible: older binaries opening the DB will simply ignore the new tables.

**Security**: no privilege changes.

## Related

- Closes:
- Follow-up PR(s)/TODOs:
  - Unit tests for the IMMEDIATE-tx ingest path under simulated lock contention and for the reembed sentinel write/read cycle.
  - One-shot data cleanup or admin RPC for the ~308 sentinel rows now on the dev workspace (cosmetic — the rows are correct, just noisy in diagnostics).

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/memory-tree-and-win-dev`
- Commit SHA: `9b9f54ea`

### Validation Run
- [x] N/A: `pnpm --filter openhuman-app format:check` not run — Windows working tree has the documented ~600-file CRLF/LF drift; `format:check` would rewrite unrelated files. Pre-push hook bypassed via `--no-verify` per the same documented pattern.
- [x] N/A: `pnpm typecheck` not run — no TypeScript surface changed in this PR.
- [x] Focused tests: the existing `reembed_backfill_repopulates_then_completes` integration-style test in `jobs/handlers/mod.rs:1062` should still pass with the new `NOT EXISTS` clause (it triggers the success path the sentinel is not written on); I have not run it explicitly post-change.
- [x] Rust fmt/check: `cargo check` ran implicitly via `pnpm dev:app:win` rebuild — 0 errors. `cargo fmt` not explicitly run; will run in the follow-up that adds tests.
- [x] N/A: Tauri fmt/check — no `app/src-tauri/` Rust changes in this PR.

### Validation Blocked
- `command:` `cargo fmt` and unit-test additions for the IMMEDIATE-tx and reembed-skip paths
- `error:` n/a — not blocked technically; bundled with the test-coverage follow-up before un-drafting.
- `impact:` PR is opened as DRAFT; reviewer feedback welcome on the schema + worklist-SQL approach before I invest in the test suite.

### Behavior Changes
- Intended behavior change:
  1. Gmail per-message ingest no longer drops messages under worker contention — the IMMEDIATE-tx makes the 15 s busy_timeout actually apply on the read-then-write block in `persist()`.
  2. `reembed_backfill` chain terminates instead of looping forever on orphan chunks — terminal failures are persistently tombstoned and excluded from the worklist.
  3. `pnpm dev:app:win` actually starts on a fresh Windows checkout without manual workarounds.
- User-visible effect:
  - Gmail sync writes actually persist; *"Memory Sources"* counts grow instead of staying at zero.
  - Logs are no longer flooded with `[memory_tree::jobs] reembed_backfill: chunk … body read failed; skipping` (was ~128 k warns / run before, now once per orphan per signature).
  - New contributors on Windows can run `pnpm dev:app:win` and have a Tauri shell open.

### Parity Contract
- Legacy behavior preserved:
  - `mem_tree_chunk_reembed_skipped` / `mem_tree_summary_reembed_skipped` are created idempotently — older code opening the DB pre-this-change works unchanged; rolling back this commit removes only the additive tables.
  - The `NOT EXISTS … reembed_skipped` clauses are pure subtraction from the worklist — every chunk that previously appeared still appears unless it has a terminal-failure sentinel, which only the new handler code can write.
  - `mark_*_reembed_skipped` is best-effort (`let _ =`): a transient write failure does not crash the batch; the row simply gets retried on the next batch (desired fallback).
  - IMMEDIATE-tx is strictly safer than DEFERRED: it never deadlocks (acquires write lock at BEGIN), and `busy_timeout` already governs the wait. Existing successful writers under no contention behave identically.
  - `run-dev-win.sh` changes are additive and only fire conditions that already produced errors (pnpm-not-on-PATH; PATH overflow). Contributors whose machines didn't hit those conditions see no behaviour change.
- Guard/fallback/dispatch parity checks: the worklist SQL was updated in all 3 sites that reference the same predicate (handler batch + `has_uncovered` runtime probe + `has_uncovered` migration probe) to avoid drift between sites; each unconditionally adds `AND NOT EXISTS … reembed_skipped` so the three remain semantically aligned.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: n/a

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows dev startup: improved PATH handling and de-duplication so the package manager is reliably found and startup logs PATH state.
  * Embedding backfill: unembeddable chunks/summaries are permanently marked per model signature and excluded from reprocessing, preventing infinite retry loops.
  * Sync status reporting: pending/processed counts now reflect migrated sidecars, fixing incorrect status totals.
  * Ingest stability: transaction behavior adjusted to reduce concurrency/busy-timeout issues during persistence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2349?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/run-dev-win.sh
+++ b/scripts/run-dev-win.sh
@@ -494,8 +494,13 @@ echo "[run-dev-win] pnpm resolved to: $PNPM_EXE"
 # ever calls pnpm by absolute path, so its dir was never on PATH and Tauri
 # dies with "'pnpm' is not recognized". Prepend the resolved pnpm's dir — it
 # ships pnpm.CMD alongside the bash shim, which cmd.exe uses.
-export PATH="$(dirname "$PNPM_EXE"):$PATH"
-echo "[run-dev-win] pnpm dir prepended to PATH: $(dirname "$PNPM_EXE")"
+# Split the dirname computation out of the export so a `dirname` failure
+# surfaces with a non-zero exit (SC2155) instead of being swallowed by the
+# enclosing `export`. `dirname` on a validated absolute path is reliable
+# in practice, but the strict-mode posture is worth the extra line.
+PNPM_DIR="$(dirname "$PNPM_EXE")"
+export PATH="$PNPM_DIR:$PATH"
+echo "[run-dev-win] pnpm dir prepended to PATH: $PNPM_DIR"
 echo "[run-dev-win] node on bash PATH:    $(command -v node 2>/dev/null || echo '<not found>')"
 echo "[run-dev-win] node.exe on bash PATH: $(command -v node.exe 2>/dev/null || echo '<not found>')"
 
--- a/scripts/run-dev-win.sh
+++ b/scripts/run-dev-win.sh
@@ -499,8 +499,18 @@ echo "[run-dev-win] pnpm resolved to: $PNPM_EXE"
 # enclosing `export`. `dirname` on a validated absolute path is reliable
 # in practice, but the strict-mode posture is worth the extra line.
 PNPM_DIR="$(dirname "$PNPM_EXE")"
-export PATH="$PNPM_DIR:$PATH"
-echo "[run-dev-win] pnpm dir prepended to PATH: $PNPM_DIR"
+# `dirname` returns `.` for a bare filename (e.g. if PNPM_EXE somehow
+# resolved to just "pnpm" without a path component). Prepending `.` would
+# inject the current working directory into PATH on a Windows dev machine
+# — a privilege-escalation-flavoured surprise. Skip the prepend in that
+# case (and on the also-degenerate empty result); the absolute-path call
+# sites elsewhere in this script still work.
+if [[ -n "$PNPM_DIR" && "$PNPM_DIR" != "." ]]; then
+  export PATH="$PNPM_DIR:$PATH"
+  echo "[run-dev-win] pnpm dir prepended to PATH: $PNPM_DIR"
+else
+  echo "[run-dev-win] pnpm dir not prepended to PATH (PNPM_EXE has no path component: $PNPM_EXE)"
+fi
 echo "[run-dev-win] node on bash PATH:    $(command -v node 2>/dev/null || echo '<not found>')"
 echo "[run-dev-win] node.exe on bash PATH: $(command -v node.exe 2>/dev/null || echo '<not found>')"
 
--- a/scripts/run-dev-win.sh
+++ b/scripts/run-dev-win.sh
@@ -488,6 +488,14 @@ if [[ -z "$PNPM_EXE" ]]; then
   exit 1
 fi
 echo "[run-dev-win] pnpm resolved to: $PNPM_EXE"
+
+# `cargo tauri dev` runs its beforeDevCommand (`pnpm run dev`) via a native
+# `cmd /S /C` that resolves bare `pnpm` off PATH. This script otherwise only
+# ever calls pnpm by absolute path, so its dir was never on PATH and Tauri
+# dies with "'pnpm' is not recognized". Prepend the resolved pnpm's dir — it
+# ships pnpm.CMD alongside the bash shim, which cmd.exe uses.
+export PATH="$(dirname "$PNPM_EXE"):$PATH"
+echo "[run-dev-win] pnpm dir prepended to PATH: $(dirname "$PNPM_EXE")"
 echo "[run-dev-win] node on bash PATH:    $(command -v node 2>/dev/null || echo '<not found>')"
 echo "[run-dev-win] node.exe on bash PATH: $(command -v node.exe 2>/dev/null || echo '<not found>')"
 
@@ -575,6 +583,25 @@ else
   echo "[run-dev-win] WARNING: invalid OPENHUMAN_DEV_PORT='$raw_dev_port'; falling back to 1420" >&2
   DEV_PORT=1420
 fi
+
+# Tauri spawns beforeDevCommand (`pnpm run dev`) via a native `cmd /S /C`
+# inheriting THIS process's env. By here PATH has the full system PATH stacked
+# several times over (vcvars rebuild + Git-Bash /etc/profile re-runs + pnpm
+# .bin layering); the MSYS→Windows conversion overflows the process
+# environment-block limit, so the child inherits an EMPTY PATH and Tauri dies
+# with "'pnpm' is not recognized" (even `where` is gone). Collapse PATH to
+# first-seen entries (clean POSIX `/c/...` entries, so ':' split is safe).
+_dedup_seen=":"
+_dedup_new=""
+IFS=':' read -ra _dedup_parts <<< "$PATH"
+for _dp in "${_dedup_parts[@]}"; do
+  [[ -z "$_dp" ]] && continue
+  case "$_dedup_seen" in *":$_dp:"*) continue ;; esac
+  _dedup_seen="${_dedup_seen}${_dp}:"
+  _dedup_new="${_dedup_new:+$_dedup_new:}$_dp"
+done
+export PATH="$_dedup_new"
+echo "[run-dev-win] PATH de-duplicated: ${#_dedup_parts[@]} → $(awk -v RS=: 'END{print NR}' <<< "$_dedup_new") entries"
 
 if (( DEV_PORT != 1420 )); then
   echo "[run-dev-win] OPENHUMAN_DEV_PORT=$DEV_PORT — overriding tauri devUrl"
--- a/src/openhuman/memory/tree/jobs/handlers/mod.rs
+++ b/src/openhuman/memory/tree/jobs/handlers/mod.rs
@@ -27,6 +27,7 @@ use crate::openhuman::memory::tree::score::extract::build_summary_extractor;
 use crate::openhuman::memory::tree::score::store as score_store;
 use crate::openhuman::memory::tree::store as chunk_store;
 use crate::openhuman::memory::tree::tree_global::digest::{self, DigestOutcome};
+use crate::openhuman::memory::tree::tree_source::store as summary_store;
 use crate::openhuman::memory::tree::tree_source::{
     build_summariser, get_or_create_source_tree, LabelStrategy, LeafRef,
 };
@@ -705,15 +706,18 @@ async fn handle_reembed_backfill(config: &Config, job: &Job) -> Result<JobOutcom
                     log::warn!(
                         "[memory_tree::jobs] reembed_backfill: summary {id} embed wrong dim, skipping (sig={active_sig})"
                     );
-                    let _ = crate::openhuman::memory::tree::tree_source::store::mark_summary_reembed_skipped(
-                        config, id, &active_sig, "embed wrong dim",
+                    let _ = summary_store::mark_summary_reembed_skipped(
+                        config,
+                        id,
+                        &active_sig,
+                        "embed wrong dim",
                     );
                 }
                 Err(e) => {
                     log::warn!(
                         "[memory_tree::jobs] reembed_backfill: summary {id} embed failed: {e}; skipping (sig={active_sig})"
                     );
-                    let _ = crate::openhuman::memory::tree::tree_source::store::mark_summary_reembed_skipped(
+                    let _ = summary_store::mark_summary_reembed_skipped(
                         config,
                         id,
                         &active_sig,
@@ -725,7 +729,7 @@ async fn handle_reembed_backfill(config: &Config, job: &Job) -> Result<JobOutcom
                 log::warn!(
                     "[memory_tree::jobs] reembed_backfill: summary {id} body read failed: {e}; skipping (sig={active_sig})"
                 );
-                let _ = crate::openhuman::memory::tree::tree_source::store::mark_summary_reembed_skipped(
+                let _ = summary_store::mark_summary_reembed_skipped(
                     config,
                     id,
                     &active_sig,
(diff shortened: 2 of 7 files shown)


src/openhuman/memory/tree/jobs/handlers/mod.rs
```diff
@@ -625,40 +643,95 @@ async fn handle_reembed_backfill(config: &Config, job: &Job) -> Result<JobOutcom
...
-                ),
-                Err(e) => log::warn!(
-                    "[memory_tree::jobs] reembed_backfill: summary {id} embed failed: {e}; skipping"
-                ),
+                Ok(_) => {
+                    log::warn!(
+                        "[memory_tree::jobs] reembed_backfill: summary {id} embed wrong dim, skipping (sig={active_sig})"
+                    );
+                    let _ = crate::openhuman::memory::tree::tree_source::store::mark_summary_reembed_skipped(
+                        config, id, &active_sig, "embed wrong dim",
```
M3gA-Mind: **[minor]** The chunk tombstone calls above use `chunk_store::mark_chunk_reembed_skipped(...)` (module alias), but the summary side spells out the full `crate::openhuman::memory::tree::tree_source::store::mark_summary_reembed_skipped(...)` path. This repeats three times in the summary loop and is noisy.

Add a `use` at the top of the file:
```rust
use crate::openhuman::memory::tree::tree_source::store as summary_store;
```
Then call `summary_store::mark_summary_reembed_skipped(...)` to match the `chunk_store::` pattern.

M3gA-Mind (changes requested):
## Review — PR #2349: IMMEDIATE-tx ingest + reembed skip-persistence + Windows PATH fixes

Three real, production-observed bugs fixed here, and the root-cause analysis in the PR description is thorough and accurate. The IMMEDIATE-tx change is the right fix for the SQLite busy-handler bypass, the tombstone table design is clean (FK + ON DELETE CASCADE, idempotent upsert, index on signature), and the Windows PATH dedup logic is correct. Approach approved — one blocker before un-drafting.

_CodeRabbit flagged SC2155 on the `export PATH=$(dirname ...)` line in `run-dev-win.sh` — skipped here._

---

### **[major]** Tombstone exclusion path has no test coverage

The tombstone mechanism is the heart of the reembed fix, but no test exercises it. The existing `reembed_backfill_repopulates_then_completes` (line 1062) covers only the success path — it never triggers the failure branches where `mark_chunk_reembed_skipped` / `mark_summary_reembed_skipped` are called, so the critical new invariants are unverified:

1. A terminal failure (body file missing, wrong embed dim) actually writes a tombstone row.
2. The next batch's worklist `NOT EXISTS … reembed_skipped` clause excludes it.
3. `has_uncovered` (in both `jobs/mod.rs` and the migration probe in `store.rs`) returns `false` when every remaining unembedded chunk/summary has a tombstone — i.e., the chain terminates instead of re-arming forever.

Without (3) in particular, a test regression could silently reintroduce the infinite-rearm loop. A dedicated test that seeds an orphan chunk, runs one backfill batch, asserts the tombstone row was written, runs the `has_uncovered` probe, and asserts it returns `false` would close this gap cleanly. The test harness already has `reembed_backfill_repopulates_then_completes` as a scaffold to follow.

---

### **[minor]** Fully-qualified `crate::...` path for summary tombstone calls is asymmetric

See inline comment — chunk side uses the `chunk_store::` alias, summary side spells out the full `crate::openhuman::memory::tree::tree_source::store::` path. Just add a `use` import at the top of the handler.

---

### **[minor]** No built-in path to un-tombstone a row

Once a `(chunk_id, model_signature)` is tombstoned, the only way to clear it is direct `DELETE FROM mem_tree_chunk_reembed_skipped` SQL — there's no admin RPC or helper. The PR mentions a follow-up cleanup script for the 308 dev-workspace rows. Worth filing the issue now so it doesn't get lost; if a contributor moves their workspace back the orphan chunks will silently stay skipped until tombstones are cleared manually.

---

**Summary table**

| File | Finding | Severity |
|------|---------|----------|
| `jobs/handlers/mod.rs` | No test for tombstone write / worklist exclusion / `has_uncovered` termination | major |
| `jobs/handlers/mod.rs` | Summary tombstone calls use fully-qualified `crate::...` path instead of a `use` import (chunk side uses `chunk_store::` alias) | minor |
| `store.rs` / `tree_source/store.rs` | No un-tombstone helper; manually clear tombstones if workspace paths are repaired | minor |

The code logic itself is solid — no semantic bugs found. This is a test-coverage gate only.

src/openhuman/memory/tree/jobs/handlers/mod.rs
sanil-23: Done in 09331d87 — added `use crate::openhuman::memory::tree::tree_source::store as summary_store;` at the top of `jobs/handlers/mod.rs` alongside the existing `chunk_store` alias, and converted the three `mark_summary_reembed_skipped` calls in the reembed loop. Kept the change scoped to the calls this PR introduced — the older `tree_source::store::` fully-qualified calls at lines 465 / 486 / 531 / 745 are pre-existing (from PR #2153) and out of scope here.

sanil-23:
Thanks for the thorough review — all three points addressed in **09331d87**.

### `[major]` Tombstone exclusion path test coverage

Added `reembed_backfill_tombstones_orphan_and_terminates` in `src/openhuman/memory/tree/jobs/handlers/mod.rs` (right after the existing `reembed_backfill_repopulates_then_completes` scaffold you pointed at). It exercises the three invariants you called out, end-to-end:

1. Seeds a chunk row + staged path, then `fs::remove_file` on the body to reproduce the in-wild orphan failure mode.
2. Runs `handle_reembed_backfill` once → asserts `JobOutcome::Defer` and that `mem_tree_chunk_reembed_skipped` has exactly one row keyed by `(chunk_id, sig)`.
3. Runs the handler again → asserts `JobOutcome::Done` (the `NOT EXISTS … reembed_skipped` clause in the worklist excludes the tombstone).
4. Runs the `ensure_reembed_backfill` migration probe inline → asserts `has_uncovered = false`, so the chain can't re-arm on the next config save.

That third assertion is the critical one — without it (3) the original 16-orphan → ~128 k warn → ~8 k defer runaway could silently regress and still pass the other two assertions.

### `[minor]` `summary_store` alias

Added the `use crate::openhuman::memory::tree::tree_source::store as summary_store;` next to the existing `chunk_store` alias, and converted the three `mark_summary_reembed_skipped` calls in the reembed loop to use it. Kept the scope tight — the older fully-qualified `tree_source::store::` calls at lines 465 / 486 / 531 / 745 in the same file are pre-existing (PR #2153) and out of this PR's diff, so I didn't sweep them up to keep the review-fix diff focused.

### `[minor]` No path to un-tombstone

Filed **#2358** to track. Scoped to a backend helper + optional admin RPC; explicitly noted the "sticky tombstone is intentional, manual operator action is the right interface" framing so we don't accidentally autocorrect them on the next PR.

### CodeRabbit nit (SC2155)

While here, split `export PATH="$(dirname "$PNPM_EXE"):$PATH"` in `scripts/run-dev-win.sh` into a separate `PNPM_DIR=$(…)` + `export PATH=…` so a `dirname` failure surfaces with a real exit code instead of being eaten by the enclosing `export`. Single-line nit, fits the same review-fix commit.

### Validation

- `cargo test --lib openhuman::memory::tree` → 598 passed (including the new test)
- `cargo fmt --check` → clean
- `cargo clippy -p openhuman --no-deps` (matches the `Rust Quality` CI step) → no new warnings on touched files

scripts/run-dev-win.sh
```diff
@@ -488,6 +488,19 @@ if [[ -z "$PNPM_EXE" ]]; then
...
+# ever calls pnpm by absolute path, so its dir was never on PATH and Tauri
+# dies with "'pnpm' is not recognized". Prepend the resolved pnpm's dir — it
+# ships pnpm.CMD alongside the bash shim, which cmd.exe uses.
+# Split the dirname computation out of the export so a `dirname` failure
+# surfaces with a non-zero exit (SC2155) instead of being swallowed by the
+# enclosing `export`. `dirname` on a validated absolute path is reliable
+# in practice, but the strict-mode posture is worth the extra line.
+PNPM_DIR="$(dirname "$PNPM_EXE")"
+export PATH="$PNPM_DIR:$PATH"
+echo "[run-dev-win] pnpm dir prepended to PATH: $PNPM_DIR"
```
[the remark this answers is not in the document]
sanil-23: Addressed in e48ee70a — wrapped the prepend in `if [[ -n "$PNPM_DIR" && "$PNPM_DIR" != "." ]]` and added a fall-through `echo` so the bare-name case is debuggable from the run log instead of silently leaving PATH unmodified. The absolute-path call sites elsewhere in the script still work unchanged.

graycyrus (commented):
**Review by @graycyrus**

Solid PR — all three fixes are well-motivated by production evidence and the implementation is clean.

### Walkthrough

| Area | Files | What |
|------|-------|------|
| Rust core | `ingest.rs` | DEFERRED → IMMEDIATE tx — correct fix for SQLite busy-handler bypass on read→write lock upgrades |
| Rust core | `store.rs`, `tree_source/store.rs`, `handlers/mod.rs`, `jobs/mod.rs` | Tombstone tables + `NOT EXISTS` clauses across all 3 worklist/probe sites — properly synchronized |
| Scripts | `run-dev-win.sh` | PATH prepend guard + dedup loop — defensive and well-commented |

### Notes

- **IMMEDIATE tx** is strictly safer than DEFERRED here — acquires write lock at `BEGIN` where `busy_timeout` applies. No deadlock risk since the tx is short-lived.
- **Tombstone design** (separate tables vs NULL sidecars) is the right call — keeps `*_embeddings` schema clean and the `NOT EXISTS` predicates trivial.
- **All 3 SQL sites** (handler batch worklist, `has_uncovered` runtime probe, migration probe) are properly updated with matching `NOT EXISTS … reembed_skipped` clauses. The in-code cross-references make future drift obvious.
- **Test** `reembed_backfill_tombstones_orphan_and_terminates` pins the critical regression path: tombstone write → worklist exclusion → chain termination → migration probe agreement.
- **CodeRabbit's `.` guard concern** was addressed in e48ee70 with a soft-warning approach — appropriate since absolute-path call sites still work in the degenerate case.

No critical or major findings — nice work. Pending: unit tests for the IMMEDIATE-tx path under simulated lock contention (acknowledged in PR description as follow-up before un-drafting).

graycyrus (dismissed):
Looks good, nice work!

sanil-23:
@coderabbitai addressed in a45c5db6.

The "gate the wave anchor on unresolved chunks inside the window" finding is fixed: `provider_pending.pending` is now scoped to chunks within `WAVE_WINDOW_MS` of the provider's most recent chunk (joins `provider_max`, reuses the `?1` window param), so a stale out-of-window straggler can't resurrect an "active" wave. Lifetime `chunks_pending` is unchanged (still counts old stragglers).

Added the regression test you suggested — `stale_out_of_window_pending_does_not_open_a_wave`: old unresolved (30 min ago) + recent resolved ⇒ `chunks_pending=1`, `batch_total=0`. Module suite: 7 passed.

M3gA-Mind (approved):

PR #2394, opened 2026-05-21
fix(onboarding): surface continue-to-chat failures

aqilaziz:
## Summary
- return the ContextPage completeAndExit promise to the context step instead of swallowing it at the page boundary
- put the manual Continue to chat button into loading state while completion RPCs are in flight
- show the existing friendly error card and capture Sentry when manual completion rejects

Refs #2081

## Checks
- [x] prettier --check focused onboarding files
- [x] pnpm test:unit src/pages/onboarding/pages/__tests__/ContextPage.test.tsx src/pages/onboarding/steps/__tests__/ContextGatheringStep.test.tsx
- [x] git diff --check

## Summary by CodeRabbit

* **New Features**
  * Loading state now displays during context step completion in onboarding, indicating active processing.

* **Bug Fixes**
  * Enhanced error handling when context step completion encounters issues, with improved error state display to users.

* **Tests**
  * Updated onboarding tests to validate completion error scenarios and loading state behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/pages/onboarding/pages/ContextPage.tsx
+++ b/app/src/pages/onboarding/pages/ContextPage.tsx
@@ -1,4 +1,3 @@
-import * as Sentry from '@sentry/react';
 import { useNavigate } from 'react-router-dom';
 
 import { trackEvent } from '../../../services/analytics';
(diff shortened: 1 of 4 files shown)


graycyrus (commented):
Nice fix. Moving the error handling from `ContextPage` down into `ContextGatheringStep` is the right call — that's where the UI state lives, so that's where the loading spinner and error card belong.

The `Promise.resolve(onNext())` wrapper handles the `void | Promise<void>` return type cleanly, the double-click guard via `isCompleting` is solid, and the Sentry capture + tags are preserved from the old page-level catch.

Tests cover the failure path well. No issues found — this is ready for approval.

graycyrus:
Hey @aqilaziz — thanks for tackling #2081, the loading state on the button is a nice touch and the Sentry capture cleanup is solid work.

After reviewing this more closely though, we're going to pass on showing the error card for this flow. Here's the thinking:

The context-gathering pipeline is an **optional background process** — the user's chat is fully ready regardless of whether it succeeds or fails. Showing an error card ("Your chat is ready...") for something the user didn't explicitly ask for and doesn't need to care about adds unnecessary friction right at the moment we want onboarding to feel smooth and frictionless.

The previous behavior (Sentry capture + silent continuation) was actually closer to the right UX here — if the completion RPCs fail, the best thing to do is **silently move on to chat** and let the team track it via Sentry. The user shouldn't be interrupted with a "something went wrong" card for a process they don't even know is running.

If you'd like to iterate on this, the part worth keeping would be:
- The **loading spinner** on the "Continue to chat" button (good click feedback)
- The **double-click guard** (`if (isCompleting) return`)
- If the RPC fails, just **navigate to chat anyway** instead of showing the error card

But as-is, the error card display is a UX regression for onboarding flow, so closing this one out. Appreciate the effort and the clean test coverage — hope this context helps for future contributions!

Issue #2117, opened 2026-05-18
Feature request: Enable Chinese (zh-CN) locale in release builds


anthem216:
Summary
The `zh-CN` (Simplified Chinese) translation files are present in the source code under `app/src/lib/i18n/zh-CN/` (split into 5 chunks, ~100KB total), but the `v0.53.43` release build does **not** include i18n data — the locale switcher is not accessible in the UI, and `localeSlice` is not registered in the Redux store.

 Background
- `I18nProvider` exists in source (`app/src/lib/i18n/I18nContext.tsx`) and reads from `state.locale.current`
- `LanguageSelect` component (`app/src/components/LanguageSelect.tsx`) supports 11 languages including `zh-CN`
- The built JS bundle (`index-DpwwJ3HS.js`) does **not** contain any `locale`/`zh-CN`/`detectLocale` strings
- `document.documentElement.lang` is hardcoded to `"en"` in the shipped build

Ask
Please enable i18n in the next release build so that:
1. `localeSlice` is registered in the Redux store
2. The Language picker is exposed in the Settings / Account UI
3. `zh-CN` translations are bundled in the production build

Impact
There is significant interest in a Chinese UI from users in China. Having this feature properly bundled would make OpenHuman accessible to a much wider audience.

Thank you for this amazing project!

M3gA-Mind:
This feature is fully implemented in the current codebase (v0.54.x):

1. **`localeSlice` is registered** in `app/src/store/index.ts` with `persistedLocaleReducer` and auto-detects the user's browser locale via `detectLocale()`
2. **`I18nProvider`** is active in `App.tsx`, wrapping the full component tree
3. **`LanguageSelect`** is exposed in `SettingsHome.tsx` (the Settings → Language row) and in `BootCheckGate.tsx`
4. **zh-CN translations** are 100% complete — all 1,947 keys from `en.ts` are present in the 5 zh-CN chunk files
5. The built bundle includes `detectLocale`, `zh-CN`, and all locale strings

No code changes needed. Closing as already resolved.

PR #2403, opened 2026-05-21
ci(i18n): add zh-CN desktop bundle guard

aqilaziz:
## Summary
- add a bundle scanner that fails when production JS assets omit the zh-CN locale key or Simplified Chinese picker label
- expose the check as pnpm i18n:bundle:check for release validation after app/dist is built
- keep the guard independent from source-only i18n coverage so release bundles can be checked directly

Refs #2117

## Checks
- [x] node --check scripts/verify-i18n-bundle.mjs
- [x] prettier --check package.json scripts/verify-i18n-bundle.mjs
- [x] pnpm --filter openhuman-app build
- [x] pnpm i18n:bundle:check
- [x] git diff --check

## Summary by CodeRabbit

* **Chores**
  * Added a new maintenance script entry for build checks.
* **Style**
  * Updated Chinese locale label in the language selector from "中文" to "简体中文".

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2403?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/components/LanguageSelect.tsx
+++ b/app/src/components/LanguageSelect.tsx
@@ -8,7 +8,7 @@ import { setLocale } from '../store/localeSlice';
 const LOCALE_OPTIONS: Array<{ value: Locale; flag: string; label: string }> = [
   { value: 'en', flag: '🇬🇧', label: 'English' },
   { value: 'ko', flag: '🇰🇷', label: '한국어' },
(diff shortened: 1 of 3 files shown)


graycyrus (approved):
Looks good, nice work!

Issue #2377, opened 2026-05-20
Remote core mode: OAuth callback deep link never delivers JWT to core, causing infinite onboarding loop


chacal1231:
Remote core OAuth callback never delivers JWT, causing infinite onboarding loop
Summary
When using a remote Docker-hosted core, the OAuth deep link openhuman://auth?token=... never delivers the JWT to the core after Google/GitHub login, leaving the user stuck in an infinite onboarding loop.
Problem
What happened: After completing Google/GitHub OAuth, the browser displays openhuman://auth?token=JWT&key=auth but the deep link is never handled — the JWT is never sent to the remote core via openhuman.auth_store_session. The core logs no backend session token; run auth_store_session first and fires auth_clear_session, resetting the onboarding flow.
What was expected: The OAuth callback should deliver the JWT to the configured remote core automatically, the same way it works with a local core.
Steps to reproduce:

Deploy openhuman-core via Docker behind an nginx SSL reverse proxy
Open the Windows desktop app, select "Run on the Cloud (Complex)"
Enter https://your-domain:port/rpc + core token → Test Connection returns OK → click Continue
Click Google or GitHub login → OAuth completes in external browser
Browser shows openhuman://auth?token=JWT&key=auth but app returns to Welcome screen
Repeat indefinitely — loop never breaks

Version / platform: ghcr.io/tinyhumansai/openhuman-core:latest (v0.54.0), Windows desktop client, Ubuntu server with nginx SSL proxy.
Solution (optional)
The openhuman:// deep link handler does not forward the JWT to the remote core when runtime mode is cloud. The app should detect the remote runtime mode and call openhuman.auth_store_session with the token from the deep link callback instead of assuming a local session.
Workaround: Manually extract the JWT from the URL shown in the browser after OAuth and inject it:

bashcurl -s -X POST https://your-domain/rpc \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer CORE_TOKEN" \
  -d '{"jsonrpc":"2.0","method":"openhuman.auth_store_session","params":{"token":"JWT_HERE"},"id":1}'
Acceptance criteria

 Repro gone — OAuth JWT is delivered to remote core after login without manual intervention.
 Regression safety — E2E test added covering remote runtime OAuth callback flow.
 Diff coverage ≥ 80% — Fix PR meets changed-lines coverage gate.
 Deep link handler — openhuman://auth callback correctly calls auth_store_session on the configured remote core regardless of runtime mode.

[...]

Issue #2374, opened 2026-05-20
Cannot signin with google or github


nishantmshah:
Use a concise sentence-case title that describes the broken behavior. Do not add `Bug` or bracket prefixes to the title.

## Summary

What failed, in one or two sentences (user-visible symptom or test failure).

## Problem

What happened vs what you expected, impact, and **steps to reproduce** (ordered, minimal). Include **version / platform** (app version, OS, desktop vs dev) if known.

## Solution (optional)

Suspected cause, workaround, or proposed fix. Skip if unknown.

## Acceptance criteria

- [ ] **Repro gone** — Bug no longer reproduces on the stated environment (or root cause documented if intentional).
- [ ] **Regression safety** — Unit, integration, or E2E coverage added or updated if this should not come back.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)).
- [ ] **…** — Other verify-before-close items.

## Related

Links to issues, PRs, logs, or prior discussion.

vRITHNER:
Can sign-in with Google or GitHub too

YellowSnnowmann:
This appears to be the same OAuth redirect issue reported in #2590 and #2521. The root cause was that the v0.54.0 installer was built before the OAuth fixes landed on `main`:

- #2469 — `fix(tauri): forward Windows local-runtime OAuth callbacks`
- #2511 — `feat(auth): loopback OAuth redirect with deep-link fallback`
- #2550 — `fix(oauth): make loopback redirect actually work, plus settings cleanup`

All three are merged. The fix will be available in the next release build. Closing as duplicate of #2590.

YellowSnnowmann:
> Can sign-in with Google or GitHub too

Issue #2298, opened 2026-05-20
Unable to log in


GuanXcode:
15:23:36:WRN:log [auth] unauthorized request to /auth — missing or wrong bearer token log.target="openhuman_core::core::auth" log.module_path="openhuman_core::core::auth" log.file="D:\\a\\openhuman\\openhuman\\src\\core\\auth.rs" log.line=164

udnn1:
same

Rewindbb33:
(workaround to get it working) run as admin (windows x64) in cmd :

reg add "HKEY_CLASSES_ROOT\WOW6432Node\openhuman" /ve /d "URL:OpenHuman Protocol" /f

reg add "HKEY_CLASSES_ROOT\WOW6432Node\openhuman" /v "URL Protocol" /d "" /f

reg add "HKEY_CLASSES_ROOT\WOW6432Node\openhuman\shell\open\command" /ve /d ""C:\Program Files\OpenHuman\OpenHuman.exe" "%1"" /f

Flare26:
Workaround didn't work for me :(

yashwanthkotla:
Run these commands in an Administrator PowerShell window

**reg add "HKEY_CLASSES_ROOT\WOW6432Node\openhuman" /ve /d "URL:OpenHuman Protocol" /f
reg add "HKEY_CLASSES_ROOT\WOW6432Node\openhuman" /v "URL Protocol" /d "" /f
reg add --% "HKEY_CLASSES_ROOT\WOW6432Node\openhuman\shell\open\command" /ve /d "\"C:\Program Files\OpenHuman\OpenHuman.exe\" \"%1\"" /f**

Kill background tasks: Run this to clear any frozen processes:

**Stop-Process -Name "OpenHuman" -Force -ErrorAction SilentlyContinue**

Copy the auth link: On the browser login page, click the fallback button to copy the link (starts with openhuman://auth...).

Manually run it: Press **Win + R**, paste the link into the Run dialog box, and press Enter.

wahaha1213:
一个本地化的应用为什么要登录，登了半天登录不上。

dino-zhao:
log file shows that login success,but the app return  initial waitting page

weichaowill-maker:
> 一个本地化的应用为什么要登录，登了半天登录不上。

大哥，你解决了吗，求帮忙

DesignStack:
[...]

PR #2449, opened 2026-05-21
fix(windows): make pnpm dev:app:win work behind TLS-inspecting proxies

M3gA-Mind:
## Summary

- Windows dev (`pnpm dev:app:win`) is broken end-to-end on machines behind a corporate TLS-inspection proxy and on workspaces whose path contains spaces (e.g. `E:\Office Files\...`).
- Two separate failure classes are addressed:
  - **TLS chain:** rustls + bundled webpki-roots reject the OS-trusted (but Mozilla-unknown) CA presented by the proxy, so every HTTPS/WSS call to `api.tinyhumans.ai` fails with `UnknownIssuer` — auth, sockets, composio, integrations.
  - **Dev script chain:** `bash -> pnpm -> cmd.exe -> cargo-tauri -> cmd.exe -> tauri.CMD -> vite.CMD` accumulates a small bug at almost every layer (broken pnpm self-update placeholder, quoted-path mangling, `node_modules/.bin` resolved against the wrong cwd, ASCII space causing `E:\Office is not recognized`, missing `node` on cargo-tauri's child PATH, etc.).
- All TLS swaps are gated `#[cfg(target_os = "windows")]` so macOS / Linux keep their original `rustls` path and avoid a new OpenSSL runtime dep.

## Problem

When running the desktop dev shell on a Windows host whose network re-signs TLS with a private CA (corporate AV / DLP, captive portal, ZScaler-style proxy), the bundled `OpenHuman.exe` cannot speak to any backend service:

```
[socket] Connection failed (attempt 1/5):
  WebSocket connect: IO error: invalid peer certificate: UnknownIssuer

[composio] list_connections failed:
  error sending request for url (https://api.tinyhumans.ai/agent-integrations/...)
  -> client error (Connect) -> invalid peer certificate: UnknownIssuer
```

Concurrently, `pnpm dev:app:win` itself fails to launch. The chain produces (in order, fixed by separate commits):

1. `cef-dll-sys` build script cannot download CEF (same rustls TLS issue, from `ureq`).
2. `aws-lc-sys` C feature probe fails on MSVC 14.44 (stdalign / C11 dialect).
3. `'C:/Program' is not recognized` — pnpm script-shell mangles the quoted bash path in `dev:app:win`.
4. `'This: command not found'` — broken `pnpm` placeholder in `%LOCALAPPDATA%\pnpm\.tools\...\node_modules\@pnpm\exe\pnpm`.
5. `'vite' is not recognized` — pnpm only prepends `./node_modules/.bin` (relative) for script children; it resolves against the wrong cwd.
6. `'tauri' is not recognized` — which pnpm shim variant wins determines whether `node_modules/.bin` lands on PATH.
7. `'pnpm' is not recognized` — cargo-tauri's `cmd.exe /S /C` child for `beforeDevCommand` strips PATH entries.
8. `'\"...\\pnpm\"' is not recognized` — Rust `Command` argv escaping rewrites embedded `"` to `\"` for cmd.
9. `'E:\Office' is not recognized` — bare absolute path breaks on workspaces with spaces.
10. `'node' is not recognized` — vite shim falls back to `node` (PATH lookup) when no sibling `node.exe`.

### TLS backend selection (gated to Windows)

Centralise platform-conditional TLS in every reqwest builder and the `tokio-tungstenite` Cargo feature:

```rust
let builder = Client::builder();
#[cfg(target_os = "windows")]
let builder = builder.use_native_tls();
#[cfg(not(target_os = "windows"))]
let builder = builder.use_rustls_tls();
```

```toml
[target.'cfg(windows)'.dependencies]
tokio-tungstenite = { version = "0.24", features = ["connect", "handshake", "native-tls"] }

[target.'cfg(not(windows))'.dependencies]
tokio-tungstenite = { version = "0.24", features = ["connect", "handshake", "rustls-tls-webpki-roots"] }
```

Touched HTTP-client construction sites:

- `src/api/rest.rs`
- `src/openhuman/app_state/ops.rs`
- `src/openhuman/config/schema/proxy.rs` — central factory used by composio, channels, agent paths
- `src/openhuman/integrations/{client, searxng, seltz}.rs`
- `src/openhuman/inference/provider/compatible.rs` (two sites)

After the swap: schannel on Windows reads the OS trust store, so corporate-CA-resigned certs validate cleanly. macOS / Linux stay on rustls, no behaviour change.

### `scripts/run-dev-win.sh` hardening

Each commit peels off one layer of the chain. End state: `bash -> cargo-tauri.exe -> cmd /S /C -> run-vite.bat -> node.exe vite.js`, hermetic absolute paths throughout, no dependency on whichever pnpm self-managed shim happens to win `command -v`.

Key changes:

- Prepend `$APP_DIR/node_modules/.bin` and the resolved `$PNPM_EXE` dir to `PATH_PREFIX`.
- Invoke `.cache/cargo-install/bin/cargo-tauri.exe` directly instead of `pnpm tauri dev`.
- Override `beforeDevCommand` via a one-shot `run-vite.bat` written to a space-free `%TEMP%` directory; the wrapper itself calls `node.exe vite.js` so neither `node` nor `vite.CMD` need to be on cargo-tauri's child PATH.

## Submission Checklist

- [x] N/A: Tests added or updated — TLS gate is `cfg`-only with no runtime-testable code path; dev-script changes are bash-only.
- [x] N/A: Diff coverage >= 80% — changes are `cfg`-only TLS toggles and bash script; no new executable TS/Rust lines added.
- [x] N/A: Coverage matrix updated — behaviour-only change on Windows; macOS/Linux observable behaviour unchanged.
- [x] N/A: All affected feature IDs from the matrix listed — no feature row maps to dev launcher or per-target TLS backend selection.
- [x] No new external network dependencies introduced — native-tls binds to system schannel/SecureTransport/libssl, all present on every supported platform.
- [x] N/A: Manual smoke checklist updated — release-cut surface unchanged.
- [x] N/A: Linked issue closed via `Closes #NNN` — linked tickets have broader scope than this PR fully addresses; using `Refs` instead of `Closes`.

## Impact

- **Runtime/platform:** Windows binaries now build with `native-tls` (schannel) for both reqwest and tokio-tungstenite. macOS / Linux unchanged — still `rustls` + webpki-roots, no OpenSSL runtime dep added.
- **Performance:** negligible. schannel and rustls have comparable handshake costs; this PR does not move call counts.
- **Security:** strictly improves Windows behaviour — the trust store now matches the OS one a user already audits via Windows Certificate Manager. No new attack surface.
- **Migration:** none. Existing config + state directories work as-is.

## Related

- Refs #2374 — "Cannot signin with google or github" (Google / GitHub OAuth login failures)
- Refs #2298 — "Unable to log in"
- Refs #2377 — "Remote core mode: OAuth callback deep link never delivers JWT to core, causing infinite onboarding loop"
- Follow-up PR(s)/TODOs: `aws-lc-sys`'s stdalign C-feature probe is worked around in this branch via `CFLAGS_x86_64_pc_windows_msvc="/std:c11"` set externally — a small upstream PR to aws-lc-sys to make the probe tolerant of pre-C11 dialects would let us drop that.

---

### Linear Issue

- Key: N/A (ad-hoc Windows dev unblocking session)
- URL: N/A

### Commit & Branch

- Branch: `sync/window-upstream`
- Commit SHA: `8b179199`

### Validation Run

- [x] `pnpm format:check` — passes (verified this session; Prettier reports all files use correct code style).
- [x] `pnpm typecheck` — passes (verified this session; tsc --noEmit exits cleanly).
- [x] N/A: Focused Rust tests (gmail/delete domains) — these domains are not touched by this PR; CI Rust Core Tests covers the integration paths.
- [x] Rust fmt/check — `cargo fmt --check` (exit 0) and `cargo check --manifest-path Cargo.toml` (clean finish) both pass (verified this session).
- [x] Tauri fmt/check — `cargo fmt --check` and `cargo check --manifest-path app/src-tauri/Cargo.toml` both pass (verified this session).

### Validation Blocked

- `command:` `pnpm install` (initial) and `cargo install --locked --path app/src-tauri/vendor/tauri-cef/crates/tauri-cli` (CEF download).
- `error:` `UNABLE_TO_VERIFY_LEAF_SIGNATURE` (pnpm), `UnknownIssuer` (cargo) — same root cause as the runtime TLS bug this PR fixes.
- `impact:` Local-only on this contributor's network. The committed code does not depend on the local mirror or `NODE_OPTIONS=--use-system-ca` workarounds used to bootstrap the build.

### Behavior Changes

- Intended behavior change: on Windows, every reqwest client and the WebSocket connector now consult the Windows certificate store instead of Mozilla webpki-roots. No change elsewhere.
- User-visible effect: Windows users in TLS-inspected environments can sign in, sync over WebSocket, and reach composio / channel APIs. Users on clean networks see no difference; macOS / Linux see no difference.

### Parity Contract

- Legacy behavior preserved: macOS / Linux still build and link rustls only; tungstenite still rejects `wss` without an explicit TLS feature; `integrations::client` still pins `http1_only` per the original comment.
- Guard/fallback/dispatch parity checks: all four TLS call sites pair `#[cfg(windows)]` with an explicit `#[cfg(not(windows))]` arm that selects the historical rustls path — no platform falls through with neither branch.

## Summary by CodeRabbit

* **Chores**
  * HTTP clients now consistently choose a platform-appropriate TLS backend: Windows uses the system native certificate store, other platforms use rustls. This improves compatibility with platform TLS stacks and yields more predictable, secure network behavior across features that make HTTP requests.
  * Windows development startup script updated to reliably locate local tool shims and run the vendored dev runner, making local dev startup more predictable and robust.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2449?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src-tauri/Cargo.lock
```diff
@@ -4309,6 +4320,22 @@ dependencies = [
  "windows-sys 0.61.2",
 ]
 
+[[package]]
```
M3gA-Mind, line 4323: **[major]** `motosan-ai-oauth` appears here as a direct dependency of the `openhuman` Tauri package — but `app/src-tauri/Cargo.toml` isn't in the diff, so there's no visible `Cargo.toml` line explaining why it was added.

This PR is a TLS-and-dev-script fix; an OAuth library has no obvious connection to either. A few possibilities:
- It was already in `Cargo.toml` but the lockfile was stale and got regenerated as a side-effect of the `cargo check` run mentioned in the submission checklist — in which case the `Cargo.toml` line predates this PR and is fine, but please confirm.
- It was added as part of an unrelated in-progress feature on this branch and accidentally bundled in.

Either way, please explain the provenance. An unreviewed OAuth library (PKCE flows, base64, rand, sha2) landing in the Tauri binary without explanation is a scope and security concern.

src/openhuman/config/schema/proxy.rs
```diff
@@ -441,7 +441,14 @@ pub fn build_runtime_proxy_client(service_key: &str) -> reqwest::Client {
...
+    // is honored — see [`crate::api::rest`]. macOS / Linux keep rustls.
+    let raw = reqwest::Client::builder();
+    #[cfg(target_os = "windows")]
+    let raw = raw.use_native_tls();
+    #[cfg(not(target_os = "windows"))]
+    let raw = raw.use_rustls_tls();
+    let builder = apply_runtime_proxy_to_builder(raw, service_key);
     let client = builder.build().unwrap_or_else(|error| {
         tracing::warn!(service_key, "Failed to build proxied client: {error}");
         reqwest::Client::new()
```
M3gA-Mind: **[minor]** The `.unwrap_or_else` fallback calls `reqwest::Client::new()`, which selects TLS from reqwest's feature-flag defaults — not the explicit platform choice made 10 lines up. When both `native-tls` and `rustls-tls` are compiled in, `Client::new()` picks rustls, so on Windows the error path silently reverts to the corporate-CA-breaking behavior this PR is fixing.

Suggestion — apply the same selection to the fallback:
```rust
.unwrap_or_else(|error| {
    tracing::warn!(service_key, "Failed to build proxied client: {error}");
    let fb = reqwest::Client::builder();
    #[cfg(target_os = "windows")]
    let fb = fb.use_native_tls();
    #[cfg(not(target_os = "windows"))]
    let fb = fb.use_rustls_tls();
    fb.build().unwrap_or_default()
})
```
(Same issue at the second fallback in `build_runtime_proxy_client_with_timeouts`.)

src/api/rest.rs
```diff
@@ -78,10 +78,21 @@ fn build_backend_reqwest_client() -> Result<Client> {
...
-        .default_headers(default_headers)
-        .use_rustls_tls()
+    // TLS backend selection:
+    //   - Windows: schannel (native-tls) so the Windows cert store is
+    //     honored, including any corporate CA installed by AV / TLS-
+    //     inspecting proxies that re-sign certs with a private root.
+    //     rustls + webpki-roots only knows Mozilla CAs and fails such
+    //     environments with `UnknownIssuer`.
+    //   - macOS / Linux: keep rustls + webpki-roots — these platforms
+    //     don't hit the corporate-MITM scenario in normal distribution,
```
M3gA-Mind: **[major]** The 4-line `#[cfg(windows)]` / `#[cfg(not(windows))]` TLS selection block is copy-pasted across 6 files and 8 call sites (`rest.rs`, `ops.rs`, `proxy.rs` ×2, `compatible.rs` ×2, `client.rs`, `searxng.rs`, `seltz.rs`). If the policy ever needs to change — say, Android support lands or a SecureTransport regression forces native-tls on macOS — all 8 sites need independent updates.

A one-line extraction would prevent future drift:
```rust
// e.g. in src/openhuman/tls.rs
pub fn tls_client_builder() -> reqwest::ClientBuilder {
    let b = reqwest::Client::builder();
    #[cfg(target_os = "windows")]
    let b = b.use_native_tls();
    #[cfg(not(target_os = "windows"))]
    let b = b.use_rustls_tls();
    b
}
```
Then each call site becomes `tls_client_builder().http1_only().timeout(...).build()` — readable and single-source-of-truth.

scripts/run-dev-win.sh
```diff
@@ -599,28 +617,89 @@ else
...
+  echo "[run-dev-win] tauri:ensure should have installed it. Aborting." >&2
+  exit 1
+fi
 
+# Build a tauri.conf.json `-c` JSON merge that:
+#  - pins `beforeDevCommand` to the absolute pnpm path so cargo-tauri's
+#    cmd.exe child can find pnpm regardless of any PATH stripping
+#    between bash → cargo-tauri → cmd. The default in tauri.conf.json
+#    is `"pnpm run dev"` (bare name) which depends on PATH.
+#  - overrides `devUrl` when OPENHUMAN_DEV_PORT is non-default.
```
M3gA-Mind, line 654: **[minor]** The old PATH deduplication block explicitly documented *why* it existed — MSYS→Windows conversion was overflowing the environment-block limit and leaving child processes with an empty PATH. That block is gone now.

The new approach (direct `cargo-tauri.exe` invocation, absolute paths in the `.bat` wrapper) is better, but a brief note would preserve the institutional knowledge:
```bash
# Direct invocation avoids the pnpm→cmd→bash PATH-accumulation that
# previously overflowed the env block. The wrapper uses absolute paths
# so beforeDevCommand does not need PATH at all.
```
Otherwise the next person who hits a related PATH issue won't know the history.

M3gA-Mind (commented):
## Review Summary

This is a well-motivated PR tackling two genuinely painful Windows problems. The TLS gating is logically correct — `#[cfg(target_os = "windows")]` / `#[cfg(not(target_os = "windows"))]` pairs are clean, the platform story (schannel for Windows, rustls+webpki everywhere else) is sound, and the bash script refactoring eliminates a legitimately fragile call chain. CodeRabbit's %TEMP% space-guard concern is already addressed in the new code.

Two things need resolution before merge:

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **[major]** | `app/src-tauri/Cargo.lock` | `motosan-ai-oauth` appears as a direct dep of `openhuman` (Tauri app), not explained in the PR |
| 2 | **[major]** | All 6 Rust files | TLS selector copy-pasted 8× — extract to a shared helper |
| 3 | **[minor]** | `proxy.rs:454,480` | Error-path `Client::new()` fallback bypasses platform TLS selection |
| 4 | **[minor]** | `run-dev-win.sh` | PATH dedup block removed without noting the assumption change |

The Windows dev experience improvement is real and the approach is correct — just needs the items above tidied up.

src/openhuman/integrations/client.rs
```diff
@@ -100,8 +100,16 @@ impl IntegrationClient {
...
         // rustls succeeds — so the integrations client was the odd one
         // out with raw "error sending request" failures.
-        let http_client = reqwest::Client::builder()
-            .use_rustls_tls()
+        // Windows: native TLS (schannel) so the OS cert store is honored
+        // — needed when corporate TLS interception re-signs certs with a
+        // private CA. macOS / Linux: keep rustls (preserves the staging
+        // handshake compatibility documented above).
+        let http_builder = reqwest::Client::builder();
+        #[cfg(target_os = "windows")]
```
graycyrus: **[major]** `composio/client.rs:462` has the same `use_rustls_tls()` pattern but wasn't updated.

The PR description explicitly names composio as a failing path:
> `[composio] list_connections failed: ... UnknownIssuer`

...and the comment at that site says *"Mirror the TLS settings of the shared client"* — those settings just changed, but this site didn't follow.

`src/openhuman/composio/client.rs` around line 461:
```rust
let http_client = reqwest::Client::builder()
    .use_rustls_tls()  // needs the same #[cfg] gate
```

Apply the same pattern here:
```rust
let http_builder = reqwest::Client::builder();
#[cfg(target_os = "windows")]
let http_builder = http_builder.use_native_tls();
#[cfg(not(target_os = "windows"))]
let http_builder = http_builder.use_rustls_tls();
let http_client = http_builder
    .http1_only()
    // ...rest unchanged
```

(Noting this on `integrations/client.rs` since it's the closest touched file — the actual fix is in `composio/client.rs`.)

graycyrus (changes requested):
Thorough, well-documented PR — the platform-conditional TLS gating and the `run-dev-win.sh` overhaul are exactly right. CI is fully green. One call site was missed.

### Change summary

| Area | Files | What changed |
|------|-------|--------------|
| Cargo deps | `Cargo.toml`, lockfiles | Platform-split `tokio-tungstenite`: `native-tls` on Windows, `rustls` elsewhere |
| Rust HTTP clients (x7) | `rest.rs`, `ops.rs`, `proxy.rs`, `compatible.rs`, `client.rs`, `searxng.rs`, `seltz.rs` | `#[cfg]`-gated TLS backend: `use_native_tls()` on Windows, `use_rustls_tls()` elsewhere |
| Dev script | `run-dev-win.sh` | Invoke `cargo-tauri.exe` directly, hermetic vite wrapper via temp `.bat`, PATH hardening |

All `#[cfg]` pairs are symmetric and the Cargo feature split is correct. The script space-guard for `VITE_WRAPPER_WIN` is solid.

One site was missed — see inline comment.

app/src-tauri/Cargo.lock
M3gA-Mind, line 4323: **Question (answered):** @M3gA-Mind asked about `motosan-ai-oauth` appearing as a direct dep of the Tauri app without explanation in the diff.

Root cause: the dependency is declared in the root `Cargo.toml` at line 98 (`motosan-ai-oauth = { version = "0.2", features = ["codex"] }`), which is a pre-existing addition unrelated to this PR. It surfaced in `app/src-tauri/Cargo.lock` because the lock file was regenerated as a side-effect of the `cargo check` run mentioned in the submission checklist — exactly the first scenario you anticipated. No new OAuth library was introduced by this PR; the root `Cargo.toml` line predates this branch.

M3gA-Mind (commented):
pr-manager: All actionable reviewer comments addressed in commit be9b0d77. One informational item below for the reviewer's awareness.

M3gA-Mind:
@M3gA-Mind unresolved review feedback from graycyrus — please address before we review.

M3gA-Mind (commented):
pr-manager: All remaining actionable comments applied in commit 8bf383f8. One item deferred for human decision — see inline thread below.

src/openhuman/tls.rs
```diff
@@ -0,0 +1,34 @@
+//! Platform-conditional TLS backend selection for reqwest clients.
```
M3gA-Mind, line 1: **Deferred:** CodeRabbit flagged that `src/openhuman/tls.rs` should live in `src/openhuman/tls/mod.rs` per the CLAUDE.md module-layout rule ("new standalone `*.rs` files at `src/openhuman/` root are not a template"). The previous agent and the task instructions explicitly acknowledge `tls.rs` as grandfathered alongside `dev_paths.rs` and `util.rs` for this PR, since it is a small single-function helper extracted to eliminate duplication. If the team prefers the subdirectory form before merge, the change is mechanical: `mkdir src/openhuman/tls && mv src/openhuman/tls.rs src/openhuman/tls/mod.rs` — no callers change because the import path `crate::openhuman::tls::tls_client_builder` stays identical. Please confirm whether to apply that rename before merge.

PR #2457, opened 2026-05-21
fix(config): preserve built-in reserved-slug cloud_providers across settings saves

YellowSnnowmann:
## Summary

- `handle_inference_update_model_settings` and `handle_update_model_settings` now silently drop reserved-slug `cloud_providers` entries (`openhuman`, `cloud`, `pid`) from incoming patches instead of failing the entire request with `Err("slug 'X' is reserved...")`.
- `apply_model_settings` now preserves reserved-slug entries from the stored config when replacing `cloud_providers`, so the migration-seeded built-in OpenHuman entry survives every settings save.
- Empty-slug validation behavior is preserved — empty slugs still pass the filter and hit the explicit "slug must not be empty" error.
- Added two regression tests pinning the preservation contract and the no-double-add defense.
- Eliminates Sentry **TAURI-RUST-5** (~7,183 events / 14 days, ~512/day) at the source.
- Fixes user-facing bug: AI Settings saves were silently failing for every user post-migration, dropping their actual edits.

## Problem

- The migration `unify_ai_provider_settings` seeds a `cloud_providers` entry with `slug: "openhuman"` into every user's config — this is the built-in OpenHuman provider and is referenced by `primary_cloud` and the per-workload routing fields.

- The frontend's AI Settings screen uses a read-modify-write pattern: it reads the full `cloud_providers` list (which includes the seeded built-in), the user edits one of their custom providers, and the frontend echoes the whole list back on save.

- Both schema handlers then ran every incoming entry through `is_slug_reserved()`:

```rust
if is_slug_reserved(&slug) {
    return Err(format!("slug '{}' is reserved and cannot be used for a custom provider", slug));
}
```

- Because the entries were processed via `.collect::<Result<Vec<_>, String>>().transpose()?`, a single rejected entry failed the **entire** RPC call. The user's actual edit (e.g. a new API key) was discarded, and `jsonrpc.rs` reported the error to Sentry via `report_error_or_expected` since the message didn't match any expected-error classifier.

- Result: every save-settings click → one Sentry event + a silently failed user edit. **7,183 events in 14 days**, on track to be one of the noisiest issues on the dashboard. Sentry link: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/45

- The system was rejecting data its own migration had seeded.

## Solution

Two coordinated changes that together fix the user experience and the Sentry noise without altering the reserved-slug contract:

1. **Schema handlers** (`src/openhuman/inference/schemas.rs`, `src/openhuman/config/schemas.rs`) — insert a `.filter()` step before `.map()` that drops non-empty reserved slugs silently:

   ```rust
   .filter(|entry| {
       let trimmed = entry.slug.trim();
       trimmed.is_empty() || !is_slug_reserved(trimmed)
   })
   ```

   The filter is intentionally constructed so empty slugs **pass through** and still hit the explicit "slug must not be empty" `Err` inside `.map()` — preserving the validation surface for genuinely malformed frontend payloads.

2. **`apply_model_settings`** (`src/openhuman/config/ops.rs`) — before replacing `config.cloud_providers` with the patch, snapshot any existing reserved-slug entries from the stored config and re-inject them afterwards. A `iter().any(|e| e.slug == entry.slug)` guard prevents duplicates if a CLI/test path bypasses the schema filter.

### Design decisions / tradeoffs

- **Silent filter vs. explicit error**: The frontend has no good way to distinguish "I authored this `openhuman` entry on purpose" from "this came from the read-modify-write cycle." Returning a friendly error per-entry is more disruptive than silently filtering, and the user can't create a meaningful `openhuman` provider anyway — the slug is reserved. Silent filter wins.
- **Preserve-on-apply rather than merge-in-handler**: The schema handler is shape-focused (param → patch); ownership of "what lives in stored config" belongs to `apply_model_settings`. Putting the preservation logic there means CLI / tests / future RPC paths that build a `ModelSettingsPatch` directly all benefit from the same invariant.
- **Why not classify the message as expected in `expected_error_kind`?** That would silence Sentry but leave the user with a broken save. The real bug is the failing update, not the alert.

## Submission Checklist

  - `apply_model_settings_preserves_existing_reserved_slug_cloud_providers` — happy path: built-in survives a normal save
  - `apply_model_settings_does_not_double_add_reserved_entries` — edge case: defensive de-dup
- **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/coverage.yml`](../.github/workflows/coverage.yml). Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge.
- Coverage matrix updated — added/removed/renamed feature rows in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md) reflect this change (or `N/A: behaviour-only change`)
  - `N/A: behaviour-only fix to an existing feature row (AI Settings / cloud_providers)`
- No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy))
- Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md))
  - `N/A: bug fix on an existing flow, smoke step "Save AI settings without error" already covers the regression`
- Linked issue closed via `Closes #NNN` in the `## Related` section

## Impact

- **Runtime/platform**: Desktop (Windows / macOS / Linux). No mobile / web / CLI behavior change beyond the shared core path.
- **Performance**: Negligible. One additional `iter().filter().cloned().collect()` over `cloud_providers` per settings-save call. The list is tiny (typically < 10 entries).
- **Security**: None. The reserved-slug contract (`is_slug_reserved`) is unchanged; users still cannot persist a user-controlled provider with a reserved slug. The filter drops such entries rather than allowing them through.
- **Migration / compatibility**: Backward-compatible. No schema change, no on-disk format change. Configs already on disk continue to work — in fact, this fix is what makes them work correctly going forward.
- **Observability**: Drops Sentry TAURI-RUST-5 to zero (~512 events/day eliminated). No new error-reporting paths added.
- **User-visible**: AI Settings → Save now actually persists user edits for every user who has run the `unify_ai_provider_settings` migration (i.e. everyone on a current build).

## Related

- Sentry: [TAURI-RUST-5](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/45)

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud provider configurations are preserved during settings updates, preventing loss of reserved built-in entries and avoiding duplicate entries.
  * Requests that include reserved provider identifiers are now filtered rather than causing the entire update to be rejected; providers with empty names are still rejected.

* **Tests**
  * Added tests covering cloud provider preservation and defensive duplicate handling; updated an end-to-end test to assert provider survival after updates.

* **Localization**
  * Added new German translations for provider and developer-menu (MCP server) UI text.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2457?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


M3gA-Mind:
@YellowSnnowmann this PR has merge conflicts with main — please rebase/resolve before review.

PR #2489, opened 2026-05-22
fix(inference): fail closed when BYOK intent cannot resolve a provider

### LINKED ISSUE #2483: Fail closed when custom inference routing cannot leave the hosted backend
## Summary

Custom/BYOK inference setup should not silently fall back to the hosted OpenHuman backend when the user intended direct provider routing but the provider entry or workload route is incomplete.

## Problem

Recent feedback on #2422 calls out the highest-risk trust-boundary failure: a user can believe prompts are using their own provider or local/direct route while unresolved workload routing still falls back to the managed OpenHuman backend.

Current source context that needs a focused fix:

- `src/openhuman/config/schema/types.rs` documents `inference_url` as the direct inference endpoint, separate from the product backend.
- `src/openhuman/inference/provider/factory.rs` resolves empty or `cloud` workload provider strings through `resolve_primary_cloud_provider_string`.
- `resolve_primary_cloud_provider_string` ultimately falls back to `openhuman` when there is no primary/custom provider match.
- `src/openhuman/inference/provider/factory_test.rs` currently preserves this behavior in `legacy_inference_url_without_matching_provider_stays_on_openhuman_primary`.

Minimal source-level reproduction:

1. Configure a non-OpenHuman `inference_url` intent, but leave no matching custom `cloud_providers` entry and no explicit workload route.
2. Resolve a workload such as `provider_for_role("reasoning", &config)`.
3. The resolved provider remains `openhuman` instead of failing closed or surfacing a configuration error.

Impact: this can route prompts through the hosted backend after the user intended BYOK/custom/direct inference, which contradicts the trust model clarified in #2426.

## Solution (optional)

Fail closed when a user has expressed custom/direct inference intent but the custom provider cannot be resolved. The core should return an actionable configuration error or degraded-state diagnostic instead of silently selecting the managed OpenHuman backend. The app settings/onboarding surface should also show that custom inference is incomplete until a provider entry, credential, model, and workload route are valid.

Keep explicit managed-backend use working when the user chooses OpenHuman as the provider.

## Acceptance criteria

- [ ] **Repro gone** — A custom/BYOK inference intent with missing or invalid provider mapping does not resolve to `openhuman` silently.
- [ ] **Explicit backend still works** — Users who intentionally choose the OpenHuman managed backend still route through it.
- [ ] **Actionable diagnostics** — Core and Settings → AI expose a clear incomplete-custom-provider error state.
- [ ] **Regression safety** — Provider factory and model-settings tests cover incomplete custom provider routing, explicit OpenHuman routing, and successful custom provider routing.
- [ ] **Diff coverage ≥ 80%** — The implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)).

## Related

- Follow-up from #2422, especially the 2026-05-21 comment about BYOK silently degrading to the hosted path.
- #2426 documents the current local-vs-managed boundary but intentionally does not change routing behavior.
- #1819 tracks the broader local-first connector / OAuth broker direction, which is related but separate from this inference-routing fail-closed behavior.

M3gA-Mind:
## Summary

- When `inference_url` pointed at a non-openhuman endpoint but no matching `cloud_providers` entry existed, `resolve_primary_cloud_provider_string` silently fell back to `PROVIDER_OPENHUMAN` — routing prompts through the managed backend even when the user intended BYOK/custom routing.
- This PR adds fail-closed behavior: a `BYOK_INCOMPLETE_SENTINEL` string is returned instead, which `create_chat_provider_from_string` catches early and turns into an actionable `BYOK_INCOMPLETE` error naming the configured `inference_url` and telling the user exactly how to fix it.
- Explicit workload routes (`reasoning_provider = "openhuman"`) and explicitly managed-backend configs are unaffected.

## Changes

| File | Change |
|------|--------|
| `factory.rs` | `BYOK_INCOMPLETE_SENTINEL` constant; `has_custom_inference_intent()`; sentinel path in `resolve_primary_cloud_provider_string`; early sentinel detection + error in `create_chat_provider_from_string` |
| `factory_test.rs` | Rename `legacy_…_stays_on_openhuman_primary` → expects sentinel; 6 new BYOK regression tests |
| `provider/mod.rs` | Re-export `BYOK_INCOMPLETE_SENTINEL` |

## Test plan

- [x] `cargo check -p openhuman` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test -p openhuman -- factory_test` — 54/54 pass (all pre-existing + 6 new BYOK tests)

## Related

Closes #2483

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail-closed behavior added when a custom/incoming inference URL lacks a matching cloud-provider route, preventing silent fallback to the managed backend.
  * Error messages improved to clearly identify the problematic inference URL and direct remediation toward the cloud provider route and inference_url configuration.
  * Routing now respects explicit per-workload provider settings and surfaces configuration issues earlier.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2489?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/inference/provider/factory.rs
+++ b/src/openhuman/inference/provider/factory.rs
@@ -35,6 +35,12 @@ use crate::openhuman::inference::provider::ProviderRuntimeOptions;
 pub const PROVIDER_OPENHUMAN: &str = "openhuman";
 /// Prefix for Ollama-local providers: `"ollama:<model>"`.
 pub const OLLAMA_PROVIDER_PREFIX: &str = "ollama:";
+/// Sentinel returned when a user has expressed custom/BYOK inference intent
+/// (via a non-openhuman `inference_url`) but no matching `cloud_providers`
+/// entry was found. Passed through `provider_for_role` and caught early in
+/// `create_chat_provider_from_string` to produce a clear configuration error
+/// instead of silently routing through the managed OpenHuman backend.
+pub const BYOK_INCOMPLETE_SENTINEL: &str = "__byok_incomplete__";
 
 fn is_abstract_tier_model(model: &str) -> bool {
     use crate::openhuman::config::{
@@ -148,6 +154,24 @@ pub fn create_chat_provider_from_string(
         role,
         p
     );
+
+    // Fail-closed: BYOK intent was detected upstream but no matching provider
+    // entry was found. Surface a clear configuration error instead of silently
+    // routing through the managed OpenHuman backend.
+    if p == BYOK_INCOMPLETE_SENTINEL {
+        let inference_url = config
+            .inference_url
+            .as_deref()
+            .filter(|s| !s.trim().is_empty())
+            .unwrap_or("<unset>");
+        anyhow::bail!(
+            "[chat-factory] BYOK_INCOMPLETE: inference_url is set to a custom/direct endpoint \
+             ({inference_url}) but no matching cloud_providers entry was found for role '{role}'. \
+             To complete BYOK setup add a cloud_providers entry whose endpoint matches \
+             {inference_url} (or use a workload-specific route). \
+             To use the OpenHuman managed backend instead, clear inference_url from config."
+        );
+    }
 
     // Empty / legacy "cloud" sentinel → primary cloud target.
     if p.is_empty() || p == "cloud" {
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
Addressed the two CodeRabbit review items in commit 08473ee3:

1. **Log verbosity** — downgraded both BYOK sentinel log calls from `warn` to `debug`. These fire during a configuration-in-progress state (user has set `inference_url` but hasn't added the matching `cloud_providers` entry yet), which is expected during BYOK setup and not an operational error.

2. **URL redaction** — replaced `inference_url={:?}` (full URL including path/query) with `host={}` via a new `redact_inference_url()` helper that extracts only the host component. This prevents any credentials embedded in query strings from appearing in logs.

3. **Test precision** — changed `assert_ne!(provider_for_role("reasoning", &config), BYOK_INCOMPLETE_SENTINEL)` to `assert_eq!(provider_for_role("reasoning", &config), "custom:gpt-4o")` so the test locks the exact routing outcome rather than just ruling out the sentinel.

laith-max (approved):

laith-max (approved):

PR #2514, opened 2026-05-23
feat(analytics): forward x-tauri-version and x-core-version on every backend request

senamakel:
## Summary

- Pairs the new backend middleware in [tinyhumansai/backend#738](https://github.com/tinyhumansai/backend/pull/738), which captures `x-tauri-version` / `x-core-version` / `x-web-version` on every request and threads them through inference logs + analytics.
- Today desktop-originated requests only carry **one** version header, so backend telemetry can't tell which shell build a given core-version request came from (or vice versa). This change makes every desktop request carry both.

## Changes

- **Frontend (`app/src/services/clientVersionHeaders.ts`)** — in Tauri mode `getClientVersionHeaders()` now resolves the core version once via `openhuman.update_version` RPC (cached, retries on transient failure) and returns both `x-tauri-version` and `x-core-version`. Web mode is unchanged (`x-web-version` only).
- **Rust core (`src/api/rest.rs`)** — `build_backend_reqwest_client()` additionally reads `OPENHUMAN_TAURI_VERSION` from the environment and installs `x-tauri-version` as a default reqwest header next to the existing `x-core-version`.
- **Tauri shell (`app/src-tauri/src/core_process.rs`)** — exports `OPENHUMAN_TAURI_VERSION=env!("CARGO_PKG_VERSION")` before spawning the embedded core task, so the in-process core sees the shell build that hosts it.

Result: every backend request originating from the desktop app now carries both `x-tauri-version` and `x-core-version`, whether it came from the React app (`apiClient` / `authApi`) or from the Rust core's `BackendOAuthClient` (auth, inference, analytics, channels, billing, …).

## Test plan

- [x] `pnpm test src/services/__tests__/apiClient.test.ts src/services/api/__tests__/authApi.test.ts` — 7 passing (new assertion: Tauri requests emit `x-tauri-version` and `x-core-version` together).
- [x] `cargo test --lib -- rest` — 64 passing, including the new `backend_client_sends_x_tauri_version_when_env_set` regression.
- [x] `pnpm typecheck`
- [x] `cargo check` (core + Tauri shell)
- [ ] Manual: open the desktop app against staging, hit any RPC that calls the backend, confirm both `x-tauri-version` and `x-core-version` land on the resulting `InferenceLog.clientVersions` / perf analytics event.

## Release Notes

* **Chores**
  * Enhanced API request version tracking by automatically including both Tauri desktop version and core service version in request headers, enabling better system identification and diagnostics across client and server components.

* **Tests**
  * Expanded test coverage for client version headers with comprehensive validation of concurrent version retrieval from multiple sources, error handling, and retry behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2514?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src-tauri/src/core_process.rs
+++ b/app/src-tauri/src/core_process.rs
@@ -217,6 +217,10 @@ impl CoreProcessHandle {
                     // the same env, matching what a child sidecar would have
                     // received via Command::env.
                     std::env::set_var("OPENHUMAN_CORE_TOKEN", self.rpc_token.as_str());
+                    // Surface the Tauri shell version to the in-process core so
+                    // backend-bound HTTP requests can attach `x-tauri-version`
+                    // analytics headers alongside `x-core-version`.
+                    std::env::set_var("OPENHUMAN_TAURI_VERSION", env!("CARGO_PKG_VERSION"));
                     *self.active_port.write() = port;
                     *self.last_port_fallback.write() = None;
 
--- a/app/src/services/__tests__/apiClient.test.ts
+++ b/app/src/services/__tests__/apiClient.test.ts
@@ -2,7 +2,10 @@ import { getVersion } from '@tauri-apps/api/app';
 import { isTauri } from '@tauri-apps/api/core';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { callCoreRpc } from '../coreRpcClient';
+
 vi.mock('@tauri-apps/api/app', () => ({ getVersion: vi.fn() }));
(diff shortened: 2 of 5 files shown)


sanil-23:
Good catch — wrapping in an RAII guard prevents env var leakage on test panic. Applied the fix.

sanil-23:
Good catch from CodeRabbit — wrapping in an RAII guard prevents env var leakage on test panic. Applied the fix in a follow-up PR: #2515 (or whatever number it gets).

PR #2546, opened 2026-05-23
feat(core): opt-in memory deletion on channel / composio disconnect

### LINKED ISSUE #2515: Offer memory deletion when disconnecting a channel or Composio integration
---
## Summary

When users disconnect a channel (Discord, Telegram, Slack, iMessage, WhatsApp) or a
Composio integration (Gmail, Notion, Google Drive), give them an opt-in checkbox to
also delete the ingested memory chunks from that source. On confirm, every
`mem_tree_chunks` row scoped to the source — along with its embeddings, score row, and
entity-index entries — is removed.

## Problem

Today `disconnect_channel` (`src/openhuman/channels/controllers/ops.rs`) and
`composio_delete_connection` (`src/openhuman/composio/ops.rs`) only revoke credentials
and clear `config.toml`. The memory tree is never touched.

This hurts users who disconnect expecting a clean break:

- Privacy expectation gap — ingested chats, emails, and documents continue to live
in `chunks.db` indefinitely, indexed for retrieval and feeding entity hotness scoring.
- No source-scoped escape hatch — there is no user-facing way to scrub memory tied
to a single source. The only existing tool is `memory.clear_namespace`, which is
all-or-nothing.
- Compounds over time — the longer a source is connected before disconnect, the more
stale content lingers and shapes future agent responses.

> Constraint: deletion must be opt-in (default off). Some users want to revoke a
> credential while keeping the historical knowledge the agent has already learned.

## Solution (optional)

Scope: core + app.

### 1. Core store helper — `src/openhuman/memory/tree/store.rs`

Add a new function:

```rust
pub fn delete_chunks_by_source(
    config: &Config,
    source_kind: SourceKind,
    source_id: &str,
) -> Result<usize>
```

- The existing index `idx_mem_tree_chunks_source` ON `mem_tree_chunks(source_kind, source_id)` makes the lookup efficient.
- `ON DELETE CASCADE` foreign keys on `mem_tree_chunk_embeddings`, `mem_tree_score`, and
`mem_tree_entity_index` clean derived rows automatically.
- Wrap in a single transaction; return the number of chunks deleted.

#### RPC: `disconnect_channel`

`src/openhuman/channels/controllers/ops.rs`

Change: Accept `clear_memory: bool` (default false). When true, after credentials are
revoked, enumerate every `source_id` belonging to the channel (e.g. every `slack:#…`
id) and call `delete_chunks_by_source` for each.

---

#### RPC: `composio_delete_connection`

`src/openhuman/composio/ops.rs`

Change: Same `clear_memory: bool` parameter; same guarantee for Gmail / Notion / Drive
sources.

### 3. App — disconnect confirmation dialog

- Add an unchecked-by-default checkbox: "Also delete all memory ingested from this
source".
- Copy makes the irreversibility explicit.
- Wire the flag through `coreRpcClient` to the new RPC parameter.

Backwards compatible: `clear_memory` defaults to false, so every existing flow behaves
exactly as today.

## Acceptance criteria

- `delete_chunks_by_source` lands in `memory/tree/store.rs` — deletes from
`mem_tree_chunks` filtered by `(source_kind, source_id)` inside a transaction, returns
rows deleted; cascade FKs verified to also remove the matching embeddings, score
rows, and entity-index rows.
- `disconnect_channel` accepts `clear_memory: bool` — true path deletes every chunk for
every `source_id` owned by that channel; false (default) path leaves memory untouched.
- `composio_delete_connection` accepts `clear_memory: bool` — same guarantee for Gmail /
Notion / Drive sources.
- Disconnect confirmation dialog exposes the option — unchecked by default; copy
makes irreversibility explicit; flag is forwarded through `coreRpcClient`.
- Failures surface clearly — if `clear_memory=true` and the deletion step fails after
credentials are already revoked, the user sees an actionable error instead of a
silent half-clean state.
- Diff coverage ≥ 80% — the implementing PR meets the changed-lines coverage gate
(Vitest + cargo-llvm-cov, enforced by `.github/workflows/coverage.yml`
(`../../.github/workflows/coverage.yml`)).

## Related

None found — no prior issue or PR addresses memory cleanup on disconnect.

---

I'd like to take this on. If a maintainer can confirm the direction (specifically
the opt-in default and the cascade-based deletion path), I'll fork, branch off main,
and open a PR.

mengqiuzhen:
## Summary

Add opt-in `clear_memory` parameter to `channels.disconnect` and `composio.delete_connection` RPCs. When set, all memory tree chunks indexed under the source are deleted in a single SQLite transaction. A confirmation checkbox in the UI (unchecked by default) makes the irreversibility explicit. The RPC response includes a `memory_clear` field so callers can surface partial failures.

Closes #2515.

## Changes

- **Rust core**: `delete_chunks_by_source` helper in `memory/tree/store.rs`, transaction-safe with explicit cascade cleanup
- **RPC**: `disconnect_channel` + `composio_delete_connection` accept `clear_memory: bool` (default false)
- **Frontend**: confirmation dialog with unchecked checkbox in DiscordConfig, TelegramConfig, ComposioConnectModal
- **i18n**: disconnect confirmation strings for en + zh-CN + 11 other locales
- **Tests**: 9 new frontend tests + Rust unit tests for delete/store/clear_memory flows

## Test plan

- [x] `pnpm test --run` — 342 files / 3332 tests passing (all new tests pass)
- [x] `cargo fmt --all -- --check` — clean
- [x] Rust unit tests for `delete_chunks_by_source`, `composio_toolkit_source_kind`, `disconnect_channel` with `clear_memory`
- [x] Frontend unit tests for `deleteConnection`, `disconnectChannel`, disconnect confirmation UI (3 components)

## Summary by CodeRabbit

* **New Features**
  * Two-step disconnect confirmation across integrations (Discord, Telegram, Composio) with an optional "clear memory" checkbox; disconnect now respects the chosen clear-memory option.
  * Backend now supports optional purging of ingested memory when clearing a connection.

* **Localization**
  * Added disconnect-related translations for 13+ languages.

* **Tests**
  * New and expanded tests covering confirmation UI, clear-memory flows, and backend delete behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2546?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


senamakel:
great pr! merging shortly 🙏

src/openhuman/memory_store/chunks/store.rs
```diff
@@ -1727,6 +1727,48 @@ pub fn clear_reembed_skipped_for_signature(
...
+                 (SELECT id FROM mem_tree_chunks \
+                  WHERE source_kind = ?1 AND source_id LIKE ?2)"
+            );
+            tx.execute(&sql, params![source_kind.as_str(), source_id_pattern])?;
+        }
+        let deleted = tx.execute(
+            "DELETE FROM mem_tree_chunks WHERE source_kind = ?1 AND source_id LIKE ?2",
+            params![source_kind.as_str(), source_id_pattern],
+        )?;
+        tx.commit()?;
```
senamakel: **Deferred from PR review:** delete_chunks_by_source only removes SQLite rows but not chunk body files at content_path. The single-chunk delete path in read_rpc.rs handles file cleanup but this bulk delete does not. Recommend a follow-up to add file cleanup here.

senamakel:
**Deferred from PR review (architectural concern):** @coderabbitai flagged that `composio_toolkit_source_kind` in `composio/ops.rs` uses toolkit-wide LIKE prefixes (`gmail:%`, `notion:%`) which could delete memory chunks belonging to other active connections of the same toolkit. This is a valid concern for multi-connection scenarios but requires understanding how source_ids are structured per-connection vs per-toolkit — recommend a follow-up issue.

PR #2553, opened 2026-05-24
feat(mcp): add mcp_writes audit log table (draft, RFC #2536)

justinhsu1477:
Draft scaffold for the MCP write audit log proposed in #2536 (Phase 3 RFC #2269 Q4).

Opening as **draft** now to claim territory and give @senamakel a concrete artifact to react to alongside the four design questions on #2536. The implementation here follows the lazy-consensus defaults I posted in that thread — happy to revise if the consensus lands elsewhere.

## What this PR does

Replaces the `tracing::info!`-only audit trail for the three MCP write tools (`memory.store`, `memory.note`, `tree.tag`) with a persistent SQLite-backed `mcp_writes` table colocated in the existing memory-tree DB.

### Schema (in `tree::store::SCHEMA`)

```sql
CREATE TABLE IF NOT EXISTS mcp_writes (
    id                  INTEGER PRIMARY KEY AUTOINCREMENT,
    timestamp_ms        INTEGER NOT NULL,
    client_source_type  TEXT NOT NULL,   -- e.g. "mcp:claude-desktop"
    tool_name           TEXT NOT NULL,   -- "memory.store" | "memory.note" | "tree.tag"
    args_summary        TEXT,            -- slim per-tool identifier JSON
    resulting_chunk_id  TEXT,            -- document_id on success, NULL on failure
    success             INTEGER NOT NULL,
    error_message       TEXT             -- ≤ 1 KiB, NULL on success
);
CREATE INDEX IF NOT EXISTS idx_mcp_writes_timestamp ON mcp_writes (timestamp_ms DESC);
CREATE INDEX IF NOT EXISTS idx_mcp_writes_client    ON mcp_writes (client_source_type);
CREATE INDEX IF NOT EXISTS idx_mcp_writes_tool      ON mcp_writes (tool_name);
```

Additive `CREATE TABLE IF NOT EXISTS` — applied on first connection through the existing `with_connection` cache. No separate migration step.

### Wiring

- `dispatch_write_tool` now records both successful and failed dispatches. Failed audit inserts are logged at `warn!` and **never** propagate — write availability is not degraded by the audit subsystem.
- New `tools::call_tool_with_session(name, args, client_source_type)` threads `McpSession::source_type()` (from @YOMXXX-G's #2332) through so per-client provenance (`mcp:claude-desktop` / `mcp:cursor` / fallback `mcp`) lands in the row.
- `tools::call_tool` kept as a back-compat wrapper for any callers we haven't threaded yet.

[...]

PR #2566, opened 2026-05-24
[codex] Add MCP write audit log

### LINKED ISSUE #2536: Add `mcp_writes` audit log table for queryable MCP write history (Phase 3 #2269 Q4)
## Summary

Add a persistent `mcp_writes` audit log table that records every successful MCP write tool invocation (`memory.store`, `memory.note`, `tree.tag`) so users can review what LLMs connected via MCP have written to their Memory Tree. Replaces the current `tracing::info!`-only audit trail with a queryable surface.

This closes out **Q4 from the Phase 3 RFC (#2269)** — the only one of the four RFC questions that doesn't have an implementation answer yet:

- ✅ **Q1** multi-process write path → Liohtml's #2306 picked routing through `doc_put` (daemon-sole-writer)
- ✅ **Q2** provenance → `source_type = "mcp:<client>"` (#2306 placeholder + YOMXXX's #2332 `clientInfo.name` capture)
- ✅ **Q3** `SecurityPolicy` → distinct `enforce_write_policy()` gating on `ToolOperation::Act` (#2306)
- ⏳ **Q4** audit log → 🟡 **this issue**

## Problem

After #2306 / #2316 / #2332 landed, the MCP write surface is functionally complete but only emits ephemeral `tracing::info!` lines for write events. Quoting @graycyrus's #2306 review verbatim:

> "The `tracing::info!` audit log exists but only shows in logs. A UI-side notification ... would give users visibility."
>
> "The real residual risk isn't injection or corruption (that's well-covered), it's a rogue LLM silently filling memory with noise and the user not knowing."

Concrete user-impact gaps without a queryable audit log:

1. **No accountability** — a user can't answer "what did Claude Desktop write into my memory this week?" without `grep`-ing log files.
2. **No compliance story** — enterprise users with audit-trail requirements can't satisfy them with a tracing-only sink.
3. **No abuse detection** — a misbehaving client filling memory at the per-hour rate limit shows up only as log spam, not a queryable signal.
4. **No undo path** — even if a user spots a problem write in the log, there's no structured handle to address it.

## Proposed solution

Add a new SQLite table next to the existing memory tree DB (same SQLite handle to avoid a second connection):

```sql
CREATE TABLE mcp_writes (
    id                  INTEGER PRIMARY KEY AUTOINCREMENT,
    timestamp_ms        INTEGER NOT NULL,
    client_info         TEXT    NOT NULL,   -- "mcp:claude-desktop" / "mcp:cursor" / fallback "mcp"
    tool_name           TEXT    NOT NULL,   -- "memory.store" / "memory.note" / "tree.tag"
    args_summary        TEXT,               -- JSON object with non-PII identifying args (see below)
    resulting_chunk_id  TEXT,               -- document_id returned by memory_doc_put
    success             INTEGER NOT NULL,   -- 1 success, 0 failure
    error_message       TEXT                -- populated only when success == 0
);

CREATE INDEX idx_mcp_writes_timestamp ON mcp_writes (timestamp_ms DESC);
CREATE INDEX idx_mcp_writes_client    ON mcp_writes (client_info);
CREATE INDEX idx_mcp_writes_tool      ON mcp_writes (tool_name);
```

### What goes into `args_summary`

Per-tool slim JSON that captures the **identifying** args without duplicating the document content (which is already persisted via `memory_doc_put`):

| Tool | `args_summary` shape |
| --- | --- |
| `memory.store` | `{ "title": "<first 128 chars>", "namespace": "...", "tag_count": N }` |
| `memory.note` | `{ "chunk_id": "...", "note_text_length": N }` |
| `tree.tag` | `{ "chunk_id": "...", "tags": [...] }` |

`args_summary` deliberately avoids storing raw note/store content twice — the actual content lives in the memory tree itself; the audit table just records the metadata enough to identify and join.

### Write flow

Inside `dispatch_write_tool` in `src/openhuman/mcp_server/tools.rs`:

```rust
// Existing: dispatch to doc_put
match all::try_invoke_registered_rpc(rpc_method, params.clone()).await {
    Some(Ok(value)) => {
        let document_id = value.get("document_id").and_then(Value::as_str);
        // NEW: audit record after successful write
        let _ = mcp_audit::record_write(McpWriteRecord {
            timestamp_ms: now_ms(),
            client_info: session.source_type(),    // from McpSession
            tool_name: tool_name.to_string(),
            args_summary: summarize_args(tool_name, params),
            resulting_chunk_id: document_id.map(str::to_string),
            success: true,
            error_message: None,
        }).await;
        tracing::info!(tool = tool_name, chunk_id = document_id, "write success");
        Ok(tool_success(value))
    }
    Some(Err(message)) => {
        // NEW: audit failed writes too — abuse detection signal
        let _ = mcp_audit::record_write(McpWriteRecord {
            success: false,
            error_message: Some(message.clone()),
            ...
        }).await;
        ...
    }
    ...
}
```

The audit insert uses `let _ = ... .await` (best-effort) — see Q2 below.

### Query surface

New RPC: `openhuman.mcp_audit_list` (read-only):

```rust
PutMcpAuditListParams {
    limit: u32,                        // default 50, max 500
    offset: u32,                       // default 0
    since_ms: Option<u64>,             // optional time-window filter
    client_filter: Option<String>,     // optional "mcp:claude-desktop" filter
    tool_filter: Option<String>,       // optional "memory.store" filter
    success_only: Option<bool>,        // default None (both); useful for UI's "show failures"
}
```

Returns `Vec<McpWriteRecord>` ordered by `timestamp_ms DESC`.

### Q1 — Storage backend

**A.** Add the table to the **existing memory tree SQLite** (single handle, single migration, transactional with the write itself). Easiest deployment.

**B.** Separate `mcp_audit.sqlite` file (isolated; survives memory tree corruption / reset; can be wiped independently for privacy). Bit more plumbing.

**C.** Reuse one of the existing KV stores or event_bus persistence layers if there's a natural home I haven't found.

The implementation sketch assumes **A**, but `mcp_writes` is a write-heavy append-only log with very different access patterns from the chunk tree — splitting may be cleaner.

### Q2 — Audit-write coupling on failure

**A. Best-effort audit (write succeeds even if audit fails)** — the sketch above. Audit failure is logged but not propagated. Pro: write availability never degrades; con: theoretical race where the chunk lands but the audit row doesn't, breaking the "every write is auditable" guarantee.

**B. Transactional (audit-then-write inside one SQLite tx)** — strict guarantee, but couples write latency to the audit table and complicates the `memory_doc_put` RPC boundary.

**C. Write-then-audit, abort-on-audit-failure** — between A and B; would require rolling back the just-completed write, which `doc_put` doesn't currently support cleanly.

Preference: **A** unless maintainers want the stronger guarantee for compliance reasons.

### Q3 — Retention policy

**A.** No automatic prune (table grows forever). Simplest; trusts user disk space.

**B.** Rolling window — e.g. delete rows older than 90 days during a daily job (could piggyback on the existing memory tree scheduler).

**C.** Size-bounded — drop oldest rows when the table exceeds N MB.

**D.** Per-user opt-in retention setting (under `config.mcp.audit_retention_days`).

For v1, leaning **A** (no prune) with an explicit "retention is a follow-up" note — simpler scope, doesn't lock us into a policy before we see real usage volume.

### Q4 — Query surface

**A. Internal RPC only** (`openhuman.mcp_audit_list`) — OpenHuman's own UI or CLI can read; MCP clients themselves cannot see their audit history. Tightest blast radius.

**B. Also expose as an MCP tool** (e.g. `audit.list_writes`) — lets a client like Claude Desktop reflect on its own writes, useful for "show me what you stored last session" UX. But also means an MCP client can see what other clients wrote to the same user's memory tree.

**C. Hybrid** — MCP tool exposes only the current client's own writes (filtered by `client_info == session.source_type()`), while the internal RPC sees all.

Strong preference for **A** in v1 (smallest surface, fewest privacy decisions). **C** is the right long-term shape; **B** is too permissive.

## Acceptance criteria

- [ ] `mcp_writes` table created via SQLite migration; existing users migrate cleanly on first launch after upgrade.
- [ ] Every successful `dispatch_write_tool` call inserts a row with the correct `client_info` (from `McpSession`), `tool_name`, `args_summary`, and `resulting_chunk_id`.
- [ ] Failed writes also recorded (`success: 0`, `error_message` populated) — abuse-detection signal.
- [ ] `args_summary` does not duplicate the chunk content (only identifying metadata).
- [ ] `openhuman.mcp_audit_list` RPC registered and returns records ordered by `timestamp_ms DESC`, with `limit` / `offset` / filter support.
- [ ] Unit tests cover: insert success, insert failure, query with each filter, ordering, limit + offset.
- [ ] **Diff coverage ≥ 80%**.
- [ ] No new UI surface — that's the follow-up.

## Out of scope (follow-ups)

- **UI surface** for browsing the audit log — Channels tab? Settings panel? Dedicated "Memory Activity" view? Belongs to a separate issue once the data layer is in place. @graycyrus already flagged the UI notification angle on #2306; an audit list view is the natural companion.
- **Per-write user confirmation** — the MCP spec doesn't have a confirm primitive (also discussed on #2269 Q3); confirmation is an OS-level out-of-band concern.
- **Audit log for read tools** (`memory.search`, `tree.read_chunk`, etc.) — debatable whether reads need audit; defer until we see a request.
- **Retention policy** — see Q3; punted to "follow-up after we have usage data".

## Related

- #2269 (Phase 3 RFC) — this issue is Q4 of that RFC.
- #2306 (Liohtml's `memory.store` / `memory.note`) — answered Q1, Q2 placeholder, Q3.
- #2316 (`tree.tag`) — third write tool; same audit hook applies uniformly.
- #2332 (YOMXXX's `clientInfo.name` capture) — provides the `client_info` field this audit table records.
- #2317 (closed) — companion follow-up that #2332 implemented.
- @graycyrus #2306 review: "if maintainers later want queryable history this is the easiest place to swap in".

Felyx-Fu:
## Summary

- add a persistent `mcp_writes` table to the memory-tree SQLite schema
- record successful and failed MCP write tool calls from `memory.store`, `memory.note`, and `tree.tag`
- add internal-only `openhuman.mcp_audit_list` RPC with limit, offset, since, client, tool, and success filters
- thread `McpSession::source_type()` into MCP write dispatch so audit rows and write provenance use the connected client label

Closes #2536.

## Validation

- `cargo test mcp_audit --lib`
- `cargo test summarize_write_args --lib`
- `cargo test extract_document_id --lib`
- `cargo test schema_for_rpc_method_finds_internal_mcp_audit_list --lib`
- `cargo test rpc_method_from_parts_does_not_expose_internal_mcp_audit_list --lib`
- `pnpm format:check`
- `pnpm rust:check`
- `pnpm typecheck`
- `pnpm lint` (passes with existing warnings)
- `pnpm --dir app run lint:commands-tokens`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Comprehensive MCP audit logging for write operations (timestamps, caller identity, redacted args summary, results/errors)
  * Queryable, filterable, paginated audit retrieval with newest-first ordering and safety limits
  * Audit inspection exposed as an internal-only RPC endpoint (registered on dispatch path but not publicly discoverable)
  * Tool calls now propagate session/client origin so audits capture caller provenance

* **Tests**
  * Added coverage for storage, querying, redaction, ordering, pagination, and internal RPC wiring

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2566?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


YOMXXX:
Review/CI triage note: this red run is a real compile failure, not the earlier checkout/GHCR infra class. The current first blocker is in `src/openhuman/mcp_server/tools.rs`: `load_write_config(...) -> Result<Config, ToolCallError>` matches `Ok(config) => config`, but that arm needs to return `Ok(config)`. That single E0308 failure is enough to fan out into Rust Quality, core tests, Tauri build, and E2E failures.\n\nAlso flagging reviewer context: this overlaps with draft #2553 for #2536. If this PR stays canonical, I would first fix the compile error and then ask the #2553 author/reviewer to either close the draft or explicitly keep it as an alternative design thread so reviewers do not review two divergent audit-log implementations in parallel.

justinhsu1477:
Author of #2553 (the parallel draft for #2536) here — flagging coordination context, not pushing for territory.

**RFC design context** (from the lazy-consensus comment on #2536):
- **Q1 = A** — colocated in the existing memory-tree SQLite, additive `CREATE TABLE IF NOT EXISTS` (no separate migration step). ✅ both PRs picked this.
- **Q2 = A** — best-effort audit; insert failures are logged but never propagate to the underlying write. ✅ both PRs picked this.
- **Q3 = A** — no auto-prune in v1; revisit once we see usage volume.
- **Q4 = A (internal RPC only, deferred in #2553)** — the `openhuman.mcp_audit_list` surface you've shipped in this PR.

So we're aligned on Q1–Q3. The split is Q4: my draft deliberately deferred the query RPC pending ratification (kept a `recent()` helper as `#[cfg(test)]` only); yours commits to A (internal-only RPC, no MCP-client-side exposure) in v1. That's a reasonable read of the lazy consensus.

**One safety detail worth checking** that bit me in #2553: `error_message` truncation needs to land on a UTF-8 char boundary, not a raw byte offset. A 4-byte char like `🦀` sliced mid-codepoint produces invalid UTF-8 and SQLite happily stores garbage. The pattern I used:

```rust
let mut end = ERROR_MESSAGE_MAX_BYTES;
while end > 0 && !msg.is_char_boundary(end) {
    end -= 1;
}
record.error_message = Some(msg[..end].to_string());
```

Plus a regression test (`record_handles_multibyte_truncation_safely`) that repeats `🦀` past the cap and asserts `is_char_boundary(stored.len())`. Worth checking your store doesn't have the same naive-slice risk.

**On the "writes vs write attempts" question @YOMXXX raised**: I'd argue both should be recorded. Failure rows are the abuse-detection signal — a misbehaving MCP client repeatedly bouncing off the policy gate or `doc_put` validation will show up as a burst of `success = false` rows that pure "successful write history" framing would miss. Both PRs already do this; just worth ratifying explicitly so the RPC docs match.

Happy to close #2553 once @senamakel picks the canonical path. Either outcome works — flagging this so the design context from the RFC thread doesn't get lost in the handoff.

Felyx-Fu:
@justinhsu1477 Agreed on both points, thanks for carrying the #2553/RFC context over here.

I tightened this up in the latest commits:

- `error_message` is now capped at 1 KiB in `mcp_audit::store::record_write`, and truncation walks back to a UTF-8 char boundary before slicing. I added `record_handles_multibyte_error_truncation_safely` with a repeated `\u{1F980}` payload to lock that down.
- The `mcp_audit.list` controller schema now explicitly describes the rows as write attempts, including successful writes plus rejected/failed attempts, and clarifies that `success_only` filters for succeeded attempts.

I also fixed the `load_write_config` `Ok(config)` compile regression and updated the audit store to use the current `memory_store::chunks::store` connection path after the main-branch memory refactor.

Validation on this branch:

- `cargo fmt --manifest-path Cargo.toml --check`
- `git diff --check`
- `cargo check --manifest-path Cargo.toml -p openhuman --lib`
- `cargo test --manifest-path Cargo.toml -p openhuman --lib record_handles_multibyte_error_truncation_safely -- --nocapture`

src/openhuman/mcp_server/tools.rs
```diff
@@ -1202,19 +1287,216 @@ async fn dispatch_write_tool(
...
-            Ok(tool_error(format!(
-                "{} is unavailable: mapped RPC method `{}` is not registered",
-                tool_name, rpc_method
-            )))
+            Ok(tool_error(format!("{tool_name} is unavailable: {message}")))
+        }
+    }
+}
+
+fn audit_write(config: &Config, record: NewMcpWriteRecord) {
```
senamakel, line 1313: Acknowledged nitpick: `tools.rs` is ~2750 lines — well beyond the ~500 line guideline. Extracting the write-dispatch and audit helpers (`dispatch_write_tool`, `audit_write`, `audit_write_rejection`, `summarize_write_args`, etc.) into a sibling module like `mcp_server/write_dispatch.rs` would be a solid follow-up. Deferring for a separate PR since it's a heavy refactor orthogonal to the audit feature.

YOMXXX: Tracked as #2604. Keeping this out of #2566 because the extraction is a mechanical-but-large module split; the audit behavior can land independently, then the refactor can preserve the write/audit contract with focused regression coverage.

PR #2639, opened 2026-05-25
feat(mcp): search + keyboard navigation for installed MCP servers

aashir-athar:
## Summary

- New **search/filter feature** for the installed MCP servers list in `Channels → ChannelConfigPanel → McpServersTab`. As users install more MCP servers, scanning the left-pane list becomes a real chore — this PR adds a controlled search input, live filtering, an accessible result counter, and arrow-key navigation across the list.
- New component `McpServerSearch.tsx`: controlled `<input type="search">` wrapped in a `role="search"` landmark, with a clear button that appears when the value is non-empty. Intentionally **no global keyboard shortcut** to avoid colliding with the app-wide `CommandProvider` in `App.tsx`.
- Extended `InstalledServerList.tsx`: optional `filter` prop case-insensitively matches against `display_name`, `qualified_name`, and `description`. New "X of Y servers" indicator announced via `role="status" aria-live="polite"`. New "No servers match \"{query}\"" empty state. ArrowUp/ArrowDown move focus across the visible server buttons (clamped at the edges; only the filtered set is traversed).
- 6 new i18n keys under the `mcp.installed.search.*` namespace, mirrored across all 13 locale chunks (English values used as the untranslated placeholder per the repo's standard pattern; counted in each locale's "untranslated" total exactly like the existing keys).
- 18 new Vitest tests (7 for `McpServerSearch`, 11 for the new behaviour in `InstalledServerList`).

## Problem

The MCP feature surface ships today via `pages/Channels.tsx` → `ChannelConfigPanel.tsx` → `McpServersTab.tsx`. The left pane is a flat scrolling list of installed servers (`InstalledServerList`). With three or four installs it's fine. The moment a user installs the eighth or fifteenth — which is the actual usage shape with the Smithery catalog — finding a specific server requires either scrolling or page-search via the browser/CEF shortcut.

There's also no keyboard-only path for moving through the list. Each server is a `<button>` so Tab moves between them, but power users with even a small inventory want ArrowUp/ArrowDown.

These are the two paper cuts this PR addresses. They're contained, complementary, and they scale with the platform's growth — every Smithery server added makes them more valuable.

### New: `McpServerSearch.tsx`

Controlled component with two props (`value`, `onChange`). Renders:

- A `<div role="search" aria-label={t('mcp.installed.search.landmarkAria')}>` landmark so assistive tech can jump straight to it.
- `<input type="search">` with `aria-label` and placeholder, both translated.
- A clear `<button>` that only appears when `value.length > 0`, with `aria-label` and a decorative `<svg aria-hidden="true">` X icon.

**Intentional non-features:**

- **No global Cmd/Ctrl+K binding.** The app already has a `CommandProvider` in `App.tsx`; binding another global shortcut here would silently override or fight with it. Users click or Tab into the field.

### Extended: `InstalledServerList.tsx`

New optional `filter?: string` prop. When set, the list:

1. Trims and lowercases the filter.
2. Builds a haystack per server from `display_name + qualified_name + description` and matches case-insensitively (substring).
3. Renders, in order:
   - If `filter` is non-empty: a `<p role="status" aria-live="polite">` showing `"{shown} of {total} servers"` so screen readers announce the new total as the user types.
   - If the filtered set is empty: a centered "No servers match \"{query}\"" message. Distinct from the existing zero-installed empty state, which still wins when `servers.length === 0` regardless of filter.
   - Otherwise: the existing `<ul>` of server buttons.
4. The original (zero-installed) empty state is preserved untouched and still wins over the no-match state — verified by an explicit test.

Each server button gets a stable `data-server-id={server.server_id}` attribute and an `onKeyDown` handler:

- **ArrowDown / ArrowUp** find the next/previous sibling button via `listRef.current.querySelectorAll('button[data-server-id]')`, call `focus()`, and `preventDefault`. Clamps at the edges (no wrap, simpler mental model; matches WAI-ARIA listbox guidance for explicit single-selection lists).
- **Other keys are not intercepted** — Enter/Space still activate the button, Tab still escapes the list. Verified by an explicit test that asserts `fireEvent.keyDown` returns `true` for an unrelated key.

### Wiring in `McpServersTab.tsx`

A new `useState<string>('')` in the tab owns the filter. The `<McpServerSearch>` renders directly above `<InstalledServerList>` in the left pane, but **only when `servers.length > 0`** — there's nothing to search when nothing is installed, and the search box would visually compete with the existing "Browse catalog" empty-state CTA.

The filter is intentionally **not persisted** — it's a transient scan helper, not a saved view. Reloading the tab or revisiting the page returns to an unfiltered list.

### i18n

6 new keys under `mcp.installed.search.*` added to `app/src/lib/i18n/en.ts` AND to `app/src/lib/i18n/chunks/en-1.ts` (the runtime chunk that holds the `mcp.installed.*` namespace). Per the project's parity rule enforced by `scripts/i18n-coverage.ts`, the same six keys are also added to all 12 non-English locale chunks (`ar-1.ts` … `zh-CN-1.ts`) with English values as untranslated placeholders. They show up in each locale's "untranslated" count exactly like every other not-yet-translated key.

Verified locally:

```
$ pnpm i18n:check
…
## zh-CN  (2988 keys)
  missing:        0
  extra:          0
  …
  per-chunk:      1:1197/1197  2:387/387  3:389/389  4:391/391  5:629/629
EXIT: 0
```

## Submission Checklist

- [x] **Tests added** — 18 new tests in two files. 7 cover the new `McpServerSearch` (landmark, input attributes, clear button visibility / firing, controlled value, decorative-icon `aria-hidden`). 11 cover the new behaviour on `InstalledServerList` (filter by name / qualified_name / description / undefined-description; whitespace trimming; no-match state; "X of Y" with correct `role="status"` + `aria-live="polite"` attributes; count hidden when filter empty or whitespace-only; original empty state precedence; ArrowDown/Up focus moves; clamping at edges; unrelated keys not intercepted; arrow nav restricted to visible items).
- [x] **Diff coverage ≥ 80%** — every new branch in `InstalledServerList.tsx` and every new line in `McpServerSearch.tsx` is exercised by the new tests. Local Vitest: **99/99 passing**, including all pre-existing MCP component tests (no regression).
- [x] Coverage matrix updated — **N/A: enhancement to an existing feature row, no new feature ID added or removed.**
- [x] All affected feature IDs listed under `## Related` — **N/A: enhancement to existing MCP feature surface.**
- [x] No new external network dependencies introduced — purely client-side filter and keyboard nav; no new API calls, no new packages.
- [x] Manual smoke checklist updated if release-cut surfaces touched — **N/A: in-app component only, no release-cut surface.**
- [x] Linked issue closed via `Closes #NNN` in `## Related` — no specific issue; this is an organic UX improvement on the MCP surface.

## Impact

- **Runtime/platform**: Desktop only — `McpServersTab` is only rendered in `ChannelConfigPanel` which only shows on desktop. No iOS / web impact.
- **Performance**: Filter is O(n × m) where n = installed servers and m = filter length. With realistic n (~10–100 installs) and m (~tens of chars), this is microseconds per keystroke. `useMemo` guards the filtered list so the filter only re-runs when `servers` or `trimmedFilter` change.
- **Security / migration / compatibility**: None.
- **Backward compatibility**: `filter` is an **optional** prop with default `''`. Any existing caller of `InstalledServerList` that doesn't pass it gets exactly the previous behaviour byte-for-byte (verified via the existing test suite passing unchanged).
- **i18n**: English text everywhere by default; the same 6 keys exist in every locale's chunks as untranslated placeholders ready for native-speaker translation in follow-up PRs.

## Related

- Closes:
- Follow-up PR(s)/TODOs:
  - Native-speaker translation of the 6 new `mcp.installed.search.*` keys across the 12 non-English locales (they currently fall back to the English values; flagged in each locale's `untranslated` count).
  - Optional future: also surface tool names in the filter haystack (search by "git" → matches any server that ships a tool named `git_*`). Needs the tool inventory pre-loaded across all connected servers; out of scope for this PR (would change the data model).

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `feat/mcp-servers-search`
- Commit SHA: `13d70bda` (will be referenced from this PR)

### Validation Run

- [x] `pnpm --filter openhuman-app lint` — **clean** (exit 0).
- [x] `pnpm vitest run src/components/channels/mcp/` — **99/99 passing** including the 18 new tests and the pre-existing `Skills.mcp-coming-soon.test.tsx` (the latter confirms no regression to the intentionally-pinned Skills-page placeholder).
- [x] `pnpm i18n:check` — **exit 0**, every locale at `1:1197/1197` parity, 0 missing keys, 0 extra keys, 0 drift.

### Validation Blocked

- `command:` `pnpm --filter openhuman-app format:check` (which chains `cargo fmt --check` via the app's `format:check` script) and the husky pre-push hook (which runs `pnpm format` → `cargo fmt`).
- `impact:` Used `git push --no-verify` per CLAUDE.md's allowance for unrelated pre-existing breakage. **This PR touches zero Rust files**, so `cargo fmt --check` is not actually applicable to this change set — CI's Linux runner will run it as a no-op for the changed files. Prettier on the changed TS files runs in CI on a clean Linux checkout where line endings are LF; verified locally that the new files were written with LF.

### Behavior Changes

- Intended behavior change: Search/filter and keyboard navigation become available in the installed MCP server list when at least one server is installed.
- User-visible effect: Faster scanning at any install count >1; keyboard-only users can move through the list with Arrow keys; screen-reader users hear the live-region announce the filtered count as they type.

### Parity Contract

- Legacy behavior preserved: The `filter` prop defaults to `''`. When omitted (every existing call site), `InstalledServerList` renders identically to the previous version — verified by all pre-existing tests still passing unchanged.
- Guard/fallback/dispatch parity checks: The original zero-installed empty state still wins over the new no-match state (explicit test). The fallback to `'disconnected'` for missing status entries is unchanged.

- Canonical PR: This one.
- Resolution: N/A.

## Summary by CodeRabbit

* **New Features**
  * Local search UI to filter installed servers (matches name, identifier, description); clear button resets filter. Live "X of Y" announcements while filtering; preserves original empty-state when no servers exist.
* **Bug Fixes**
  * Status indicator tooltips now use localized status text casing.
* **Tests**
  * Expanded tests for filtering behavior, keyboard (ArrowUp/Down) navigation, status tooltip titles, and accessible messaging.
* **Localization**
  * Added translations for the search UI in 16 languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2639?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oxoxDev (changes requested):
## Walkthrough

Adds `McpServerSearch` component + filter + ArrowUp/ArrowDown nav + result counter to the installed MCP servers list. New `role="search"` landmark on the input, `role="status" aria-live="polite"` on the count. CodeRabbit raised `title={status}` (raw enum leak) — author addressed by routing through `STATUS_I18N_KEYS` + `channels.status.*` i18n keys, and tests updated to assert i18n'd labels. 18 new Vitest tests. 19 files +614/-43. CodeRabbit re-APPROVED post-fix. CI: Rust Core failures pre-existing on jobs unrelated to a frontend PR.

## Verified

- Status tooltip fix: `title={t(STATUS_I18N_KEYS[status])}` at L477 + matching test updates ✓
- All 6 new `mcp.installed.search.*` keys present in `en.ts` + 13 locale `-1.ts` chunks (ar/bn/de/en/es/fr/hi/id/it/ko/pt/ru/zh-CN)
- Intentional non-feature noted: no global Cmd/K binding (would collide with `CommandProvider`)
- Trim + lowercase + 3-field haystack (display_name + qualified_name + description) match logic is clean

## Major findings

- **i18n parity gap** — `pl-1.ts` (Polish) has 0 new keys vs 6 in every other locale + `en.ts`. Polish locale was added to `upstream/main` AFTER this PR branched (`4043d03e feat(i18n): add Polish locale #2731` on May 27, branch-base is May 26). Will fail `pnpm i18n:check` once rebased. Author needs to rebase + add the 6 `mcp.installed.search.*` keys to `pl-1.ts` using English as placeholder per CLAUDE.md pattern.

## Nits

- `InstalledServerList.tsx` keyboard handler — tests cover ArrowDown / ArrowUp + edge clamping, but no test for what happens when focus is on the empty-state's "Browse catalog" button and user hits ArrowUp. Likely a no-op (no buttons above), but worth pinning.

## Questions

- ArrowUp / ArrowDown only traverse the filtered set per PR body. If user filters to 1 result then clears the filter, does focus stay on the same server or reset? Worth a one-line test if you have a preference.

graycyrus:

aashir-athar:
Thanks @oxoxDev! Addressed in the latest push:

**Major — i18n parity (pl):** rebased onto current main. `pl-1.ts` uses a `...en1` spread, so it inherits every English key with fallback automatically — `i18n:check` passes EXIT 0 with 0 missing across all locales (incl. pl) without per-key edits. The 12 explicit locales carry the `mcp.installed.search.*` keys. (My #2655 inventory feature landed on main in the meantime; the merge reconciles search + inventory wiring/keys in McpServersTab as a union.)

On the ArrowUp-from-empty-state nit: it's a no-op by construction — keyboard nav only iterates the rendered result rows, and the empty state renders no rows to move focus between. Happy to pin it with a test in a follow-up.

Validation: 37 search/list tests pass, tsc clean, eslint clean, prettier clean.

PR #2641, opened 2026-05-25
feat(mcp): connection health toolbar with bulk Retry All / Disconnect All actions

aashir-athar:
## Summary

- New **MCP Connection Health Toolbar** in the left pane of `Channels → ChannelConfigPanel → McpServersTab`. Surfaces aggregate per-state counts (connected / connecting / error / idle) and exposes two bulk actions that previously required clicking into each server one at a time.
- **Retry All `(N)`** — visible only when there are errored servers. Iterates through them via `mcpClientsApi.connect(id)` wrapped in `Promise.allSettled` so one bad apple doesn't abort the batch, then refreshes statuses once at the end.
- **Disconnect All `(N)`** — visible only when there are connected servers. Gated behind a `role="dialog" aria-modal` confirmation dialog before the bulk RPC fires.
- The status summary is announced through a `role="status" aria-live="polite"` region so screen-reader users hear updates as the 5-second polling loop refreshes. Status dots are `aria-hidden`. Buttons have descriptive `aria-label`s with interpolated counts.
- 15 new i18n keys under `mcp.health.*` mirrored across all 13 locale chunks (English values used as untranslated placeholders per the repo's standard pattern).
- 17 new Vitest tests covering count aggregation, conditional button rendering, dialog open/cancel/confirm flow, partial-failure orchestration via `Promise.allSettled`, error surfacing through `role="alert"`, and the a11y attribute contract on every interactive element.

## Problem

MCP servers in this app can drift into `error` state for legitimate reasons — a key expired, a managed runtime restart cycle, a flaky upstream HTTP-remote server. Today, recovering means clicking each errored server individually in the list, navigating to its detail view, and hitting connect. With three or four servers it's tedious; with the ten-plus shape that real Smithery users hit, it's a chore.

Disconnect-all is symmetric — useful during debugging ("what changes if I cut all MCP connections?"), useful during shutdown, useful for the user who just installed a server with bad env vars and wants a clean reset across the board.

There's also no aggregate at-a-glance status. The list shows per-server dots, but counting "how many am I currently connected to?" requires scanning. For users with a dozen servers, that scan happens visually every time they open the tab.

### New component `McpConnectionHealthToolbar.tsx`

Renders **only** when `statuses.length > 0` — preserves the existing "empty install + Browse catalog CTA" empty state by not competing for vertical space.

**Layout** (compact, fits the 224px left pane):

```
┌────────────────────────────────────────┐
│ HEALTH       Retry all (1)  Disconnect │
│                                all (3) │
│ ● 3 connected  ● 1 error  ● 2 idle    │
└────────────────────────────────────────┘
```

Status pills use the existing color vocabulary from `McpStatusBadge` (sage/amber/coral/stone). Dots flow on a `flex flex-wrap` row so the layout doesn't break at narrower widths.

**Aggregation** (in `computeHealthCounts`, exported as a pure helper):

- Single pass over `statuses` producing `connectedIds`, `errorIds`, and counts for `connecting` / `disconnected`. `useMemo` keyed on `statuses` so the bulk action targets don't recompute on unrelated re-renders.
- `connectedIds` and `errorIds` are arrays (not just counts) — the same memoization payload powers both the visual badges and the bulk-action callbacks, so there's no risk of the displayed count diverging from the IDs that would be acted on.

**Bulk action flow** (Retry All):

1. Click `Retry all (N)` — calls `onReconnect(errorIds)`; sets `isOperating=true` to disable both buttons.
2. Parent (`McpServersTab.handleBulkReconnect`) does `Promise.allSettled(errorIds.map(id => mcpClientsApi.connect(id)))` followed by `fetchStatuses()`.
3. On resolve, `isOperating=false`; the next render reflects the new statuses; the buttons re-enable.
4. On parent throw: `setOpError(err.message)` displays the failure in a `role="alert"` paragraph below the summary. Falls back to `t('mcp.health.opErrorGeneric')` for non-Error throws.

**Bulk action flow** (Disconnect All):

1. Click `Disconnect all (N)` — opens confirmation dialog with `role="dialog" aria-modal="true"` and explicit `aria-labelledby` / `aria-describedby` referencing the title and body IDs.
2. Cancel → close dialog, `onDisconnect` never fires.
3. Confirm → close dialog, then run the same `Promise.allSettled` orchestration via `onDisconnect(connectedIds)`.

### Wiring in `McpServersTab.tsx`

- Imports the new component.
- Adds `handleBulkReconnect` and `handleBulkDisconnect` as `useCallback`s alongside the existing `handleInstallSuccess` / `handleUninstalled` handlers. Both call `Promise.allSettled` on per-server RPC then `fetchStatuses()` for one round-trip refresh.
- Renders `<McpConnectionHealthToolbar>` above `<InstalledServerList>` in the left pane, after the existing `loadError` block.

### i18n

15 new keys under the `mcp.health.*` namespace added to `app/src/lib/i18n/en.ts` AND to `app/src/lib/i18n/chunks/en-1.ts` (the runtime chunk that already holds the existing `mcp.*` keys). All 12 non-English locale chunks (`ar-1.ts` … `zh-CN-1.ts`) get the same six keys with the English value as the untranslated placeholder, per the project pattern that `scripts/i18n-coverage.ts` enforces.

Verified locally:

```
$ pnpm i18n:check
…
## zh-CN  (2997 keys)
  missing: 0   extra: 0   drifted chunks: 0
  per-chunk: 1:1206/1206  2:387/387  3:389/389  4:391/391  5:629/629
EXIT: 0
```

## Submission Checklist

- [x] **Tests added** — 17 new tests in `McpConnectionHealthToolbar.test.tsx`. Cover: empty-state return-null path; correct count aggregation across mixed status sets; conditional rendering of connecting / error pills only when non-zero; conditional rendering of Retry All only when errors exist; conditional rendering of Disconnect All only when connected exist; "Retry all" calls `onReconnect` with exactly the errored IDs; "Disconnect all" opens a confirmation dialog rather than firing directly; cancel closes the dialog without calling `onDisconnect`; confirm fires `onDisconnect` with exactly the connected IDs and closes the dialog; both action buttons are disabled during a pending bulk operation; errors thrown from `onReconnect` surface via `role="alert"`; non-Error throws fall back to a generic message; the summary region has `role="status"` + `aria-live="polite"` + a descriptive `aria-label`; the confirmation dialog body interpolates the correct connected count.
- [x] **Diff coverage ≥ 80%** — every branch in `McpConnectionHealthToolbar.tsx` and the new `handleBulkReconnect` / `handleBulkDisconnect` callbacks in `McpServersTab.tsx` are exercised by the test suite. Local Vitest: **91/91 passing** across the full MCP suite (8 test files); no regression in any pre-existing test.
- [x] Coverage matrix updated — **N/A: enhancement to existing MCP feature row, no new feature ID added or removed.**
- [x] All affected feature IDs listed under `## Related` — **N/A.**
- [x] No new external network dependencies introduced — uses the existing `mcpClientsApi.connect` / `mcpClientsApi.disconnect` round-tripped through the existing JSON-RPC.
- [x] Manual smoke checklist updated if release-cut surfaces touched — **N/A.**
- [x] Linked issue closed via `Closes #NNN` in `## Related` — no specific issue; organic UX improvement on the MCP surface.

## Impact

- **Runtime/platform**: Desktop only — `McpServersTab` is desktop-only via `ChannelConfigPanel`. No iOS / web impact.
- **Performance**: Status aggregation is O(n) per render, memoized. Bulk actions are N parallel RPCs via `Promise.allSettled` — strictly equivalent to a user manually triggering N connects/disconnects, just dispatched together. No new polling loop.
- **Security**: Disconnect is gated behind a confirmation dialog (`role="dialog" aria-modal`). The dialog body explicitly states secrets and configurations are preserved, addressing the "did I just lose my install state?" anxiety.
- **Backward compatibility**: The toolbar returns `null` when `statuses.length === 0`, preserving the existing zero-installed flow byte-for-byte. All pre-existing MCP tests still pass unchanged.
- **i18n**: English values everywhere by default; the same 15 keys exist in every locale's chunks as untranslated placeholders, ready for native-speaker translation in follow-up PRs (they count toward each locale's `untranslated` total).
- **A11y**: Live region for status changes; explicit `aria-label`s with interpolated counts on action buttons; `role="dialog" aria-modal aria-labelledby aria-describedby` on the confirmation modal; decorative `aria-hidden` status dots; `role="alert"` for operation errors.

## Related

- Closes:
- Follow-up PR(s)/TODOs:
  - Native-speaker translation of the 15 new `mcp.health.*` keys across the 12 non-English locales (currently fall back to English placeholders; flagged in each locale's `untranslated` count).
  - Optional future: a "Refresh now" button next to the toolbar for users who want an immediate poll instead of waiting up to 5 seconds for the next tick. Skipped here to keep scope tight.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `feat/mcp-connection-health-toolbar`
- Commit SHA: (filled by GitHub after push)

### Validation Run

- [x] `pnpm --filter openhuman-app lint` — **clean for new files**. The full repo currently produces 62 warnings (0 errors), of which zero are in the files this PR adds or modifies; the single warning at `McpServersTab.tsx:70:18` is the pre-existing `setLoading(false)` inside the initial-load `useEffect`, untouched by this PR.
- [x] `pnpm vitest run src/components/channels/mcp/` — **91/91 passing** including the 17 new tests and all pre-existing MCP component tests (no regression).
- [x] `pnpm i18n:check` — **exit 0**, every locale at `1:1206/1206` parity, 0 missing keys, 0 extra keys, 0 drift.

### Validation Blocked

- `command:` `pnpm --filter openhuman-app format:check` (chains `cargo fmt --check` via the workspace script) and the husky pre-push hook (which runs `pnpm format` → `cargo fmt`).
- `impact:` Used `git push --no-verify` per CLAUDE.md's allowance for unrelated pre-existing breakage. **This PR touches zero Rust files**, so `cargo fmt --check` would have been a no-op for the changed files. Prettier on the changed TS files runs in CI on a clean Linux checkout with LF line endings; verified locally that the new files were written with LF.

### Behavior Changes

- Intended behavior change: Surfaces aggregate MCP connection health in the tab pane; introduces bulk Retry All / Disconnect All as new user actions gated behind appropriate state (errors exist / connections exist) and confirmation (for disconnect).
- User-visible effect: At-a-glance health for power users with many MCP servers; one-click recovery of bulk error states; one-confirmation cut-all-connections for debugging or reset.

### Parity Contract

- Legacy behavior preserved: When `statuses.length === 0`, the toolbar returns `null` — the existing layout renders exactly as before. All pre-existing MCP tests pass unchanged. Per-server connect/disconnect via the existing detail view is untouched.
- Guard/fallback/dispatch parity checks: `Promise.allSettled` ensures partial failures within a bulk action don't abort the rest; the next poll tick reconciles any drift; thrown errors at the orchestration boundary surface through `role="alert"` rather than being swallowed.

- Canonical PR: This one.
- Resolution: N/A.

## Summary by CodeRabbit

* **New Features**
  * Added MCP connection health toolbar showing real-time server status counts (connected, connecting, error, disconnected).
  * Introduced bulk "Retry all" and "Disconnect all" actions with confirmation dialog for safer operations.
  * Error alerts display partial failure details when bulk operations encounter issues.

* **Internationalization**
  * Added health UI translation support across 12+ languages including Arabic, Bengali, German, Spanish, French, Hindi, Indonesian, Italian, Korean, Portuguese, Russian, and Chinese.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2641?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oxoxDev (changes requested):
## Walkthrough

New `McpConnectionHealthToolbar` component above the installed servers list — aggregate per-state counts (connected/connecting/error/idle), Retry All button (visible when errors exist), Disconnect All (gated behind `role="dialog" aria-modal` confirmation). Bulk RPCs use `Promise.allSettled` so one bad apple doesn't abort the batch. Status summary in `role="status" aria-live="polite"` region. 15 new `mcp.health.*` i18n keys + 17 Vitest tests. CodeRabbit dismissed initial review then APPROVED. CI: only Rust Core Windows fails (pre-existing main flake; unrelated to FE PR).

## Verified

- `computeHealthCounts` is a pure exported helper, `useMemo` keyed on `statuses` — counts can't diverge from the IDs the bulk actions target.
- Both bulk handlers wired in `McpServersTab.tsx:140-156` via `Promise.allSettled` over `serverIds.map(mcpClientsApi.connect/disconnect)` — orchestration matches PR claim.
- Toolbar renders only when `statuses.length > 0` → preserves the "Browse catalog" empty-state CTA.
- 13 locale `-1.ts` chunks + `en.ts` all carry 15 keys ✓

## Major findings

- **i18n parity gap** — `pl-1.ts` (Polish) has 0 of the 15 new `mcp.health.*` keys vs 15 in every other locale. Same root cause as #2639: Polish locale was added to `upstream/main` (commit `4043d03e`, PR #2731) AFTER this branch was cut on May 25. Will fail `pnpm i18n:check` once rebased. Same fix: rebase + add the 15 `mcp.health.*` keys to `pl-1.ts` using English values as placeholders per the standard pattern.

## Nits

- `McpConnectionHealthToolbar:80` — `confirmDisconnect` state has no escape-key handler test. Esc should close the dialog without firing the bulk RPC. Worth a one-line test using `fireEvent.keyDown(dialog, { key: 'Escape' })`.
- The Retry All flow calls `Promise.allSettled` but doesn't surface per-server failure counts. After "Retry All (5)" runs, user only knows 5 attempts happened, not how many succeeded. Consider exposing `{ succeeded, failed }` in `opError` so the toast/alert region tells the user "3 succeeded, 2 failed" — actionable, otherwise they re-scan dots.

## Questions

- The 5-second polling loop continues running while the confirmation dialog is open. If a server transitions to disconnected mid-dialog, the count behind the modal silently changes. Acceptable, or should the dialog snapshot the count at open time?

graycyrus:

aashir-athar:
Thanks @oxoxDev! Addressed in the latest push:

**Nit — Esc on the Disconnect-all dialog:** added an Escape handler that closes the confirmation without firing the bulk RPC, matching the modal-dismiss affordance of the other MCP dialogs. The listener attaches only while the dialog is open and is declared before the component's early return (rules of hooks). New test pins Esc → dialog closed + `onDisconnect` not called.

**Major — i18n parity (pl):** rebased onto current main. `pl-1.ts` uses a `...en1` spread so it inherits every key with English fallback automatically — `i18n:check` passes EXIT 0 with 0 missing (incl. pl), no per-key edits needed. (My #2655 inventory feature landed on main in the meantime, so the merge also reconciles the toolbar + inventory wiring/keys as a union.)

The Retry-All success/failure-count nit and the dialog-snapshot question are good follow-ups; happy to take them in a focused PR so this one stays scoped to the toolbar.

Validation: i18n:check EXIT 0, 18 toolbar tests pass, tsc clean, eslint clean, prettier clean.

app/src/components/channels/mcp/McpConnectionHealthToolbar.tsx
```diff
@@ -0,0 +1,250 @@
...
+          <div className="bg-white dark:bg-neutral-900 rounded-xl shadow-xl max-w-sm w-full p-4">
+            <h2
+              id="mcp-disconnect-all-title"
+              className="text-sm font-semibold text-stone-900 dark:text-neutral-100 mb-2">
+              {t('mcp.health.disconnectConfirm.title')}
+            </h2>
+            <p
+              id="mcp-disconnect-all-body"
+              className="text-xs text-stone-600 dark:text-neutral-300 mb-4">
+              {t('mcp.health.disconnectConfirm.body').replace(
```
senamakel, line 224: Re: @oxoxDev's open question about the 5s poll continuing while the Disconnect-all confirm dialog is open —

Current behavior is intentional and I'd argue it's the safer of the two options: `runDisconnectAll` reads the **live** memoized `counts.connectedIds` at confirm time (not a value captured at dialog-open), and the dialog body interpolates the same live `counts.connectedCount`. So if a server drops to `disconnected` mid-dialog, both the displayed count and the actual disconnect target stay in sync — we never disconnect a server the user can no longer see as connected, and we never act on a stale snapshot.

Snapshotting at open time would *freeze* the body text but then act on a list that may have changed, which is the failure mode worth avoiding. The only cosmetic wrinkle is the body number can tick while the modal is open; if that's distracting we could freeze the *displayed* number while still acting on the live set, but I'd lean toward leaving it as-is for correctness. Deferring the final call to you — happy to add the display-freeze if you prefer.

graycyrus:

PR #2644, opened 2026-05-25
feat(mcp): interactive Tool Execution Playground (schema viewer + JSON args editor + run + history)

aashir-athar:
## Summary

- New **Tool Execution Playground** — an interactive modal that opens from any tool in the connected-server detail view, lets the user compose JSON arguments against the tool's input schema, executes the call live via `mcpClientsApi.toolCall`, and renders the structured result. Closes the long-standing gap between "I can see this tool's schema" and "I can actually invoke it without writing a chat prompt or curl call".
- New component `McpToolPlayground.tsx` (432 lines) — modal dialog with: collapsible JSON-schema viewer; auto-focused JSON args textarea with live-clearing parse errors; **Format** action (pretty-prints valid JSON, leaves invalid untouched); **Run** button + Cmd/Ctrl+Enter shortcut from the textarea; result panel with role/aria-live varying by success vs error; copy-to-clipboard with transient "Copied" state; collapsible in-session **invocation history** capped at 10 entries with one-click "Load" to restore prior args.
- `McpToolList.tsx` extended with an **optional** `onTryTool(tool)` callback — when omitted the list renders as before (preserved for any other call site); when provided each tool gets an accessibly-labelled "Try" button next to its name.
- `InstalledServerDetail.tsx` wires the playground in: hosts the `playgroundTool` state, passes `onTryTool` to the list **only when the server is connected**, renders `<McpToolPlayground>` conditionally with proper lifecycle binding.
- **20 new i18n keys** under `mcp.toolList.*` (2) and `mcp.playground.*` (18), mirrored across all 13 locale chunks per the repo's `i18n:check` parity rule.
- **32 new Vitest tests** (28 in `McpToolPlayground.test.tsx`, 4 added to `McpToolList.test.tsx`) plus an existing `McpToolList` test rewritten to match the new row structure.

## Problem

Today, the only way for a user to test whether an installed MCP server's tool actually does what its description promises is to:

1. Open the agent chat
2. Type a natural-language prompt that the agent might or might not interpret as a request to call that exact tool
3. Hope the agent passes the right arguments
4. Read the agent's summary of the result

This is an *atrocious* feedback loop for someone debugging a freshly-installed server, validating env-var changes, or just trying to understand a tool's behaviour. The existing `mcpClientsApi.toolCall` RPC has been wired since #2559 — there was simply no UI surface to invoke it directly.

The MCP detail view also hits a wall once the server is connected and its tool list is rendered: the list shows a name + description and that's it. No way to peek at the input schema, no way to invoke. Every tool is read-only documentation.

### `McpToolPlayground.tsx` — the modal

A self-contained modal opened by the parent setting `playgroundTool: McpTool | null`. Lifecycle bound to the modal's lifetime: focus management, Esc binding, click-outside binding, history state all live inside the component and die with it.

**Args editor**:

- `<textarea>` controlled via `argsJson` state, initialised to `'{}'` so the empty state is also valid JSON.
- `onChange` live-clears any stale parse error from the previous Run.
- `aria-label`, `aria-describedby` pointing to the help text, `spellCheck={false}` (it's code, not prose).
- **Format** button: `JSON.stringify(JSON.parse(raw), null, 2)` — pretty-prints valid JSON, returns the original string untouched if invalid (never throws on Format).

**Run flow**:

1. Parse the args; whitespace-only input is treated as `{}`. Bad JSON → `setParseError`, render `role="alert"`, abort.
2. Set `isRunning`, clear prior result.
3. `await mcpClientsApi.toolCall({ server_id, tool_name, arguments: parsed })`.
4. Success → `setResultText(JSON.stringify(callResult.result, null, 2))`, `setResultIsError(callResult.is_error)`, push to history (cap 10, most-recent-first).
5. Thrown exception → render as error result via `role="alert"`, also pushed to history (so a failed try is recoverable).
6. Always `setIsRunning(false)` in `finally`.

**Keyboard**: Cmd/Ctrl+Enter inside the textarea triggers Run (`preventDefault` so the modifier+Enter doesn't insert a newline). Esc on the document closes (gated through React's effect cleanup so it doesn't survive unmount).

**Result panel**:

- Success: `role="status" aria-live="polite"` — screen readers announce updates as new invocations land.
- Error (tool returned `is_error: true` OR an exception was thrown): `role="alert" aria-live="assertive"`.
- Both render the result as formatted JSON inside a `<pre>`; the panel ships with a **copy-to-clipboard** button that flips to "Copied" for ~1.5s on success and silently no-ops on browsers/contexts without clipboard access.

**History**:

- Collapsible (default hidden). Heading shows `History (N)` so the count is visible without expanding.
- Each entry: status dot (sage / coral) for success / error, local-timezone HH:MM:SS, args preview, "Load" button that restores that entry's args into the textarea and re-focuses it.
- Cap at 10; oldest fall off; first run after the cap is silently dropped to keep the in-session memory bounded. Errors are recorded too — a failed try is still a try.

### `McpToolList.tsx` — the entry point

Optional `onTryTool?: (tool: McpTool) => void` prop:

- When undefined, the list renders exactly as before — every other call site keeps working unchanged (verified by the existing tests passing).
- When defined, each tool row gets a "Try" button (right-aligned, primary-coloured, full `aria-label="Open execution playground for {name}"`).

One pre-existing test (`does not render description paragraph when description is undefined`) used a `p + p` selector against the previous flat row structure; updated to match the new row structure (`<div>` wrapper for name + Try, then optional `<p>` for description) without weakening the assertion.

### `InstalledServerDetail.tsx` — the orchestrator

Three small additions:

- Import `McpToolPlayground` + a `playgroundTool` state slot.
- Pass `onTryTool={status === 'connected' ? setPlaygroundTool : undefined}` so the Try button is gated on a live connection (a disconnected server has no tools, but this is belt-and-suspenders against future state drift).
- Conditionally render `<McpToolPlayground serverId={server.server_id} tool={playgroundTool} onClose={() => setPlaygroundTool(null)} />` so the modal's lifecycle is bound to a specific tool invocation session.

### i18n

20 new keys: 2 in `mcp.toolList.*` (Try button text + per-tool aria-label with `{name}`), 18 in `mcp.playground.*` (title with `{name}`, close, schema viewer, args label/help, run shortcut hint, Format, run/running/result/result-error/copy/copied/history/history-empty/history-load/invalid-json/unexpected-error). Added to `en.ts` and to all 13 locale chunks (`en-1.ts` … `zh-CN-1.ts`). English values used as untranslated placeholders for non-English locales per the existing pattern.

Verified locally:

```
$ pnpm i18n:check
…
## zh-CN  (2997 keys)
  missing: 0   extra: 0   drifted chunks: 0
  per-chunk: 1:1211/1211  2:387/387  3:389/389  4:391/391  5:629/629
EXIT: 0
```

## Submission Checklist

- [x] **Tests added** — 32 new tests (28 in `McpToolPlayground.test.tsx`, 4 in `McpToolList.test.tsx`). Covers: modal a11y attributes (`role="dialog"`, `aria-modal`, `aria-labelledby`); description present / absent; close via Esc / button / backdrop click (and **not** via clicking the dialog card itself); auto-focus on mount; collapsible schema viewer; default `{}` args; invalid-JSON refusal with `role="alert"`; whitespace-only args treated as `{}`; Format handling for valid and invalid input; success result rendering with `role="status"`; `is_error: true` rendering with `role="alert"`; thrown RPC exception rendering; Error vs non-Error fallback messages; Run button disabled state during in-flight calls; Cmd+Enter / Ctrl+Enter shortcuts; bare Enter doesn't run; history recording; Load restores args; history cap at 10 with oldest falling off; error invocations recorded; empty-history placeholder; per-tool Try button rendering when `onTryTool` is provided and absent when omitted; Try invokes the callback with the right tool object; Try renders even for description-less tools.
- [x] **Diff coverage ≥ 80%** — every branch in `McpToolPlayground.tsx` and every new line in `McpToolList.tsx` is exercised by the new tests. Local Vitest: **106/106 passing** across the full MCP suite (8 test files); no regression in any pre-existing test.
- [x] Coverage matrix updated — **N/A: enhancement to existing MCP feature row, no new feature ID.**
- [x] All affected feature IDs listed under `## Related` — **N/A.**
- [x] No new external network dependencies introduced — uses the existing `mcpClientsApi.toolCall` over the existing JSON-RPC; the optional clipboard write uses `navigator.clipboard.writeText` which is best-effort and silently degrades.
- [x] Manual smoke checklist updated if release-cut surfaces touched — **N/A.**
- [x] Linked issue closed via `Closes #NNN` in `## Related` — no specific issue; organic UX improvement.

## Impact

- **Runtime/platform**: Desktop only — the playground lives inside `InstalledServerDetail` which only renders in `Channels → ChannelConfigPanel`. No iOS / web impact.
- **Performance**: Modal lifecycle is bound to `playgroundTool`. History is a bounded in-memory array (cap 10). JSON parse / stringify happens on Run / Format only — never per keystroke. Schema viewer is collapsible (closed by default).
- **Security**: The playground only invokes tools on the user's **already-installed, already-connected** MCP servers; the agent's existing autonomy gates and per-tool policies still apply on the core side via the same RPC the agent uses. The clipboard write is best-effort and only fires when the user clicks Copy.
- **Backward compatibility**: `McpToolList`'s `onTryTool` is an **optional** prop. Every existing call site (just `InstalledServerDetail` today) keeps working unchanged when the prop is omitted — verified by the existing test suite passing without modification.
- **A11y**: Modal dialog with proper attributes; auto-focus on the input on open; Esc / backdrop close; result region uses `role="status"` / `role="alert"` with matching `aria-live` polite/assertive; all interactive buttons have `aria-label`s; status dots and icons are `aria-hidden="true"`; Try button aria-label includes the tool name for context.
- **i18n**: English text by default; 20 new keys exist in every locale's chunks as untranslated placeholders ready for native-speaker translation in follow-up PRs.

## Related

- Closes:
- Follow-up PR(s)/TODOs:
  - Native-speaker translation of the 20 new `mcp.toolList.tryTool*` / `mcp.playground.*` keys across the 12 non-English locales (they currently fall back to English placeholders; flagged in each locale's `untranslated` count).
  - Optional future: a JSON-schema-driven form renderer so simple tools (string / number / boolean / enum / required-only objects) can be invoked without typing raw JSON. Skipped here to keep scope tight and avoid re-implementing JSON-schema coercion semantics.
  - Optional future: persist a small per-tool "last-good args" snapshot across modal closes so users don't lose their work when they close + reopen the modal.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `feat/mcp-tool-playground`
- Commit SHA: (filled by GitHub after push)

### Validation Run

- [x] `pnpm --filter openhuman-app lint` — **clean for new files**. Full repo currently shows 62 warnings + 0 errors (one error I introduced — duplicate React imports — was fixed before commit by merging into a single import). Zero errors and zero warnings now attributable to the files this PR adds or modifies.
- [x] `pnpm vitest run src/components/channels/mcp/` — **106/106 passing** including 32 new tests; pre-existing `Skills.mcp-coming-soon.test.tsx` still passes (confirms no regression to the intentionally-pinned Skills-page placeholder).
- [x] `pnpm i18n:check` — **exit 0**, every locale at `1:1211/1211` parity, 0 missing keys, 0 extra keys, 0 drift.

### Validation Blocked

- `command:` `pnpm --filter openhuman-app format:check` (chains `cargo fmt --check`) and the husky pre-push hook (which runs `pnpm format` → `cargo fmt`).
- `impact:` Used `git push --no-verify` per CLAUDE.md's allowance for unrelated pre-existing breakage. **This PR touches zero Rust files**, so `cargo fmt --check` would have been a no-op for the changed files. Prettier on the changed TS files runs in CI on a clean Linux checkout with LF line endings; the two new files were authored with LF.

### Behavior Changes

- Intended behavior change: Users on the MCP Servers tab who select a connected server now see a "Try" button next to each tool in the inventory. Clicking it opens an interactive playground modal where they can compose JSON arguments, run the tool live, inspect the structured result, and iterate. The in-session history lets them retry previous args without retyping.
- User-visible effect: A genuine debugging / exploration surface for MCP that previously didn't exist. The agent chat is no longer the only path to invoking a tool.

### Parity Contract

- Legacy behavior preserved: `McpToolList`'s API is backward-compatible (the new `onTryTool` prop is optional). When omitted the list renders byte-for-byte as before — verified by every pre-existing `McpToolList` test still passing without modification. `InstalledServerDetail`'s existing flows (connect / disconnect / uninstall / config assistant) are untouched.
- Guard/fallback/dispatch parity checks: Tool invocations go through the same `mcpClientsApi.toolCall` RPC the agent uses; the core's autonomy gate and per-tool security policy still apply. The clipboard write degrades silently on platforms without `navigator.clipboard`. Esc handler is bound on document and torn down in the effect cleanup so it doesn't survive unmount.

- Canonical PR: This one.
- Resolution: N/A.

## Summary by CodeRabbit

* **New Features**
  * Tool Execution Playground: run MCP tools via a modal with JSON args, format/validate input, view/copy results, and per-session invocation history (up to 10).
  * "Try" buttons on tool lists open the playground only when the server is connected.
  * UI translations added for the playground in many languages.

* **Bug Fixes**
  * Playground now closes and clears state on disconnect, uninstall, or status changes.

* **Tests**
  * Comprehensive tests for playground UI, parsing, run flow, history, accessibility, and integration with server status.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2644?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oxoxDev (changes requested):
## Walkthrough

Adds `McpToolPlayground` modal — schema viewer + JSON args textarea + Run + result + copy-to-clipboard + in-session invocation history (cap 10). New `onTryTool` opt-in prop on `McpToolList` (preserves callers that don't pass it). Wired into `InstalledServerDetail` only when `status === 'connected'`. Cmd/Ctrl+Enter shortcut, Format pretty-print, parse-error live-clear. 23 new i18n keys (`mcp.toolList.*` + `mcp.playground.*`), 32 new Vitest tests. CodeRabbit re-APPROVED post initial CHANGES_REQUESTED. CI: only Rust Core jobs fail (pre-existing main flake; unrelated to FE PR).

## Verified

- Copy-status reset on Run: `setCopyStatus('idle')` at `McpToolPlayground.tsx:151` inside `handleRun` ✓ (CodeRabbit minor closed)
- Tool list gated at render: `tools={status === 'connected' ? tools : []}` + `onTryTool={status === 'connected' ? setPlaygroundTool : undefined}` at `InstalledServerDetail.tsx:236-237` ✓ (prevents opening new playground when disconnected)

## Major findings

1. **i18n parity gap** — `pl-1.ts` has 0 of the 23 new `mcp.{toolList,playground}.*` keys vs 23 in every other locale + `en.ts`. Same stale-branch cause as #2639/#2641 (Polish landed via #2731 on May 27 after this branched May 25). Will fail `pnpm i18n:check` once rebased. Same fix: rebase + add 23 keys to `pl-1.ts` with English placeholders.
2. **Disconnect-mid-playground hole** — CodeRabbit flagged that the modal can stay open across a disconnect. Author addressed the *opening* path (gated at `onTryTool`) but did NOT close an *already-open* modal when status flips to non-connected. There's no `useEffect` watching `status` to clear `playgroundTool`. If the user opens the playground, then the server disconnects (network blip, auth expiry), the modal stays mounted and a "Run" press sends a toolCall RPC against a disconnected server. Add:
   ```ts
   useEffect(() => {
     if (status !== 'connected' && playgroundTool) {
       setPlaygroundTool(null);
     }
   }, [status, playgroundTool]);
   ```
   Plus a test for it. Half-closes CodeRabbit's original concern as written.

## Nits

- `handleRun:135` — `let parsed: unknown` declared early but assigned inside a try. Tighten with `const parsed = JSON.parse(argsJson)` inside the try and return on catch.
- 432-line `McpToolPlayground.tsx` — over CLAUDE.md's preferred ≤500 line ceiling but close to it. Worth splitting `InvocationHistory` into its own component when the next feature lands (don't block this PR on it).

## Questions

- The in-session history is capped at 10 entries — is the cap exposed as a prop, or hard-coded? If hard-coded, worth a `const HISTORY_CAP = 10;` at module scope so a future tweak doesn't need to grep the magic number.
- Cmd/Ctrl+Enter handler — does it fire when focus is inside the schema viewer's collapsed `<details>`? If yes, that's expected; just confirming the keyboard shortcut scope.

graycyrus (commented):

aashir-athar:
Thanks for the thorough review @oxoxDev! Addressed in the latest push:

**Major #2 — disconnect-mid-playground:** a staged playground tool is now dropped whenever `status` leaves `connected` by *any* path (poll, prop change, auth expiry), not just the explicit handlers. Implemented via React's "adjust state while rendering" pattern (track previous status, reset on change) rather than a status-watching `useEffect` — same result without an extra render pass or a `react-hooks/set-state-in-effect` warning. New test asserts the modal does **not** reappear on reconnect (proving state cleared, not just hidden by the render gate).

**Nit — `let parsed`:** extracted a pure, exported `parseToolArgs()` returning a discriminated result, so JSON-parse failures (user input) stay cleanly separate from RPC failures (the tool call). Unit-tested directly. (Kept them separate deliberately — folding the parse into the RPC `try` would conflate the two error surfaces.)

**Nit — history cap:** already a module const (`HISTORY_LIMIT = 10`).

**Major #1 — i18n parity (pl):** rebased onto current main. Worth noting: `pl-1.ts` uses a `...en1` spread, so it inherits every English key with fallback automatically — `i18n:check` passes EXIT 0 with 0 missing (incl. pl) without per-key edits. The 12 explicit locales already carry the feature keys.

Validation: 47 MCP tests pass, tsc clean, eslint clean (0 warnings), prettier clean.

PR #2658, opened 2026-05-26
fix(settings): blur recovery phrase by default

### LINKED ISSUE #2657: Recovery phrase should require manual reveal
## Problem

when you open Settings --> Account -->Recovery Phrase, the 12-word seed phrase is shown in plain text right away. what makes it easy for someone nearby to glance at it before the user even has a chance to hide it. It is a small privacy leak but not great for a recovery phrase screen.

## Solution

updated RecoveryPhrasePanel.tsx so the phrase is blurred by default and pointer events are disabled, preventing the words from being highlighted or copied before the user explicitly reveals them. i also added a centered eye-slash SVG overlay. it is a one-way reveal: click the icon, the blur goes away, the icon disappears for good and the user can read/copy the words normally. i also disabled the Copy to Clipboard button while the words are hidden so it fades to 50% opacity and shows the not-allowed cursor until the phrase is explicitly revealed.

iamrhn:
Fixes #2657 

### Added a default blur/reveal flow for the recovery phrase screen in RecoveryPhrasePanel.tsx.

- recovery phrase is now blurred by default
- pointer events are disabled while hidden
- added centered eye-slash reveal overlay
- reveal is one-way and requires explicit user action
- disabled “Copy to Clipboard” until the phrase is revealed
- copy button now fades and shows the not-allowed cursor while locked

This prevents the recovery phrase from being visible or copied immediately when opening the panel.

### Screenshots

image
image

## Summary by CodeRabbit

* **New Features**
  * Recovery phrase is blurred and non-interactive by default in generate mode; users can reveal it with a new "Reveal recovery phrase" control.
  * Copy-to-clipboard is disabled until the phrase is revealed; disabled styling updated for improved clarity.

* **Chores**
  * Added "Reveal recovery phrase" translation entries across multiple language files to enable localization.

* **Tests**
  * UI test added to verify reveal behavior and that the copy button becomes enabled after reveal.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2658?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/components/settings/panels/RecoveryPhrasePanel.test.tsx
+++ b/app/src/components/settings/panels/RecoveryPhrasePanel.test.tsx
@@ -1,0 +1,57 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import RecoveryPhrasePanel from './RecoveryPhrasePanel';
+
+const navigateBackMock = vi.fn();
+
+vi.mock('../hooks/useSettingsNavigation', () => ({
+  useSettingsNavigation: () => ({
+    navigateBack: navigateBackMock,
+    breadcrumbs: [],
+  }),
+}));
+
+vi.mock('../../../lib/i18n/I18nContext', () => ({
+  useT: () => ({ t: (key: string) => key }),
+}));
+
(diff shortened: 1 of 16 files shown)


senamakel:
so fly!!

PR #2683, opened 2026-05-26
fix(agent): resumable checkpoint at tool-call cap + tolerant tool parsing/resolution

sanil-23:
## Summary

- **A1 (HIGH):** at the tool-call iteration cap, the live chat turn now emits a *resumable checkpoint* (model-written "done so far / next steps", native tools disabled, streamed) as the final assistant message instead of aborting — the transcript stays well-formed and the user's next message resumes the task.
- **A1 guard:** a `0-text / 0-tool-call` completion is surfaced as an error instead of being accepted as a blank reply.
- **A2 (MED):** `LazyToolkitResolver` tolerates near-miss tool slugs (case / separator drift) so `integrations_agent` stops dropping valid Gmail/Slides calls.
- **A3 (MED):** `parse_tool_call_value` accepts argument-key aliases (`args`/`parameters`/`input`); tool name kept strict.

## Problem

From tester logs (`bug-report-2026-05-26`, M-chip tester):

- **A1** — hitting `max_tool_iterations` (10) returned `Err(MaxIterationsExceeded)` *after* the transcript had been persisted mid-tool-cycle. The thread was left ending on an unterminated assistant tool-call; the **next** message replayed that history, the gateway returned an empty completion, the loop accepted it as the final answer (`final_chars=0`), and the thread silently wedged — every subsequent message returned blank. (In the log, `thread-d6f29915`: 4 healthy turns, then permanent silence.)
- **A2** — `integrations_agent` repeatedly logged `tool not in allowlist` for valid actions (7 occurrences) because the lazy resolver matched slugs exact-only; each drop burned an iteration, accelerating A1.
- **A3** — a malformed `<tool_call>` dropped the call and wasted an iteration.

## Solution

- **A1** — on reaching the cap, `turn.rs` calls a new `summarize_iteration_checkpoint` (provider call with `tools: None`, streamed) to produce the checkpoint, persists it as the final assistant message, and returns `Ok(checkpoint)`. Full tool-call history is retained, so resumption needs no heuristic — the model continues from the checkpoint on the next user message. `build_deterministic_checkpoint` (from the turn's `ToolCallRecord`s) is the fallback if the summary call returns empty, guaranteeing the thread can never be left unterminated. Separately, an empty completion in the normal loop now returns an error rather than `Ok("")`.
- **A2** — `LazyToolkitResolver::find_action` resolves exact → case-insensitive → unique normalized-slug (alphanumeric, lowercased); fabricated slugs stay unresolved and route to the existing "available tools" error so the model self-corrects.
- **A3** — argument-key aliases via `first_args_by_keys`; `name` resolution unchanged to avoid misreading a plain JSON answer as a tool call in the whole-response parse path.

Scope note: the legacy `run_tool_call_loop` (`tool_loop.rs`, superseded for chat dispatch and now test-only) still returns `MaxIterationsExceeded`; only the live session-turn path changed.

## Submission Checklist

> If a section does not apply, marked `N/A` with a one-line reason.

- [x] Tests added or updated (happy path + failure/edge) — checkpoint-at-cap + well-formed transcript, empty-response error, deterministic fallback (`turn_tests.rs`); arg-alias recovery (`parse_tests.rs`); near-miss resolve + `normalize_slug` (`ops_tests.rs`). The prior `turn_errors_when_max_tool_iterations_are_exceeded` test was updated to assert the new checkpoint behaviour.
- [x] **Diff coverage ≥ 80%** — added unit tests target the changed lines (checkpoint path, fallback, empty guard, resolver, arg-alias); CI `diff-cover` enforces the exact gate.
- [x] Coverage matrix updated — `N/A: behaviour-only bug fix; no feature row added/removed/renamed.`
- [x] All affected feature IDs listed under Related — `N/A: no matrix feature touched.`
- [x] No new external network dependencies — none added.
- [x] Manual smoke checklist updated — `N/A: no release-cut surface changed.`
- [x] Linked issue closed via `Closes #NNN` — `N/A: no tracking issue filed; sourced from bug-report-2026-05-26 (private tester logs).`

## Impact

- Desktop chat: hitting the tool-call cap now yields a usable checkpoint and a resumable thread instead of a dead conversation. No API/schema changes, no migration. Rust-core only (no frontend changes).

## Related

- Closes: N/A (no linked issue; bug-report-2026-05-26 A1/A2/A3)
- Follow-up PR(s)/TODOs: A3 — capture real malformed `<tool_call>` bodies (now logged via `body_chars`) before any further parser widening.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/agent-tool-loop-resilience`
- Commit SHA: `5f458c73`

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool-call argument keys now accept common aliases while generic aliases are rejected for bare top-level responses; malformed tool-call warnings include tag body length.
  * Agent turns that hit iteration caps now produce resumable assistant checkpoints (with deterministic fallback) and persist a final assistant entry instead of failing.
  * Empty provider responses with no tool calls now surface an error rather than returning empty text.
  * Tool name resolution tolerates common slug variations when unambiguous.

* **Tests**
  * Added tests for argument-key aliasing, bare-response restrictions, slug normalization, checkpoint fallback, and empty-response errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2683?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/agent/harness/parse.rs
+++ b/src/openhuman/agent/harness/parse.rs
@@ -456,11 +456,13 @@ pub(crate) fn parse_tool_calls(response: &str) -> (String, Vec<ParsedToolCall>) 
 
             if !parsed_any {
                 // body_chars only (never the body itself — it may carry tool
-                // arguments with user data). Surfaces how often the model
-                // emits an unparseable tool-call tag (bug-report-2026-05-26 A3).
+                // arguments with user data). Stable `[agent_parse]` prefix so
+                // it aggregates with the other harness log families. Surfaces
+                // how often the model emits an unparseable tool-call tag
+                // (bug-report-2026-05-26 A3).
                 tracing::warn!(
-                    "Malformed <tool_call> JSON: expected tool-call object in tag body (body_chars={})",
-                    inner.chars().count()
+                    body_chars = inner.chars().count(),
+                    "[agent_parse] malformed <tool_call> JSON: expected tool-call object in tag body"
                 );
             }
 
--- a/src/openhuman/agent/harness/parse.rs
+++ b/src/openhuman/agent/harness/parse.rs
@@ -26,6 +26,27 @@ pub(crate) fn parse_arguments_value(raw: Option<&serde_json::Value>) -> serde_js
     }
 }
 
+/// Object keys that may carry the tool **arguments**, in priority order.
+/// Models drift from the canonical `arguments` to `args`/`parameters`/etc.;
+/// accepting these recovers an otherwise well-formed call (with a correct
+/// `name`) instead of dropping it and burning an agent iteration
+/// (bug-report-2026-05-26 A3). The tool **name** is deliberately left
+/// strict — widening it would risk misreading a plain JSON answer as a
+/// tool call in the whole-response parse path.
+const TOOL_ARG_KEYS: &[&str] = &["arguments", "args", "parameters", "params", "input"];
+
+/// Normalized arguments for the first present key among [`TOOL_ARG_KEYS`]
+/// (via [`parse_arguments_value`], which tolerates both stringified and
+/// object JSON). Empty-object default when none are present.
+fn first_args_by_keys(obj: &serde_json::Value) -> serde_json::Value {
+    for key in TOOL_ARG_KEYS {
+        if let Some(v) = obj.get(*key) {
+            return parse_arguments_value(Some(v));
+        }
+    }
+    parse_arguments_value(None)
+}
+
 pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedToolCall> {
     if let Some(function) = value.get("function") {
         let name = function
@@ -35,7 +56,7 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
             .trim()
             .to_string();
         if !name.is_empty() {
-            let arguments = parse_arguments_value(function.get("arguments"));
+            let arguments = first_args_by_keys(function);
             return Some(ParsedToolCall {
                 name,
                 arguments,
@@ -55,7 +76,7 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
         return None;
     }
 
-    let arguments = parse_arguments_value(value.get("arguments"));
+    let arguments = first_args_by_keys(value);
     Some(ParsedToolCall {
         name,
         arguments,
@@ -388,7 +409,13 @@ pub(crate) fn parse_tool_calls(response: &str) -> (String, Vec<ParsedToolCall>) 
             }
 
             if !parsed_any {
-                tracing::warn!("Malformed <tool_call> JSON: expected tool-call object in tag body");
+                // body_chars only (never the body itself — it may carry tool
+                // arguments with user data). Surfaces how often the model
+                // emits an unparseable tool-call tag (bug-report-2026-05-26 A3).
+                tracing::warn!(
+                    "Malformed <tool_call> JSON: expected tool-call object in tag body (body_chars={})",
+                    inner.chars().count()
+                );
             }
 
             remaining = &after_open[close_idx + close_tag.len()..];
--- a/src/openhuman/agent/harness/parse.rs
+++ b/src/openhuman/agent/harness/parse.rs
@@ -48,6 +48,27 @@ fn first_args_by_keys(obj: &serde_json::Value) -> serde_json::Value {
 }
 
 pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedToolCall> {
+    // Default to the permissive (tagged) behaviour: callers that reach a
+    // value through an explicit tool-call marker (`tool_calls` array,
+    // `<tool_call>` tags, ```tool_call blocks) accept the arg-key aliases.
+    parse_tool_call_value_aliased(value, true)
+}
+
+/// Parse a single JSON value as a tool call.
+///
+/// `allow_arg_aliases` controls whether the generic argument-key aliases in
+/// [`TOOL_ARG_KEYS`] (notably the very generic `input`) are honoured for a
+/// **bare** `{ "name": .., .. }` object. The whole-response fallback path
+/// (`parse_tool_calls` on a top-level JSON object) passes `false`: there, a
+/// normal model reply such as `{"name":"Alice","input":"hi"}` must not have
+/// its `input` slurped into tool arguments and routed to execution
+/// (bug-report-2026-05-26 A3 follow-up). The `function`-wrapped shape stays
+/// permissive regardless — the `function` key is an unambiguous tool-call
+/// marker.
+fn parse_tool_call_value_aliased(
+    value: &serde_json::Value,
+    allow_arg_aliases: bool,
+) -> Option<ParsedToolCall> {
     if let Some(function) = value.get("function") {
         let name = function
             .get("name")
@@ -76,7 +97,13 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
         return None;
     }
 
-    let arguments = first_args_by_keys(value);
+    let arguments = if allow_arg_aliases {
+        first_args_by_keys(value)
+    } else {
+        // Singleton whole-response fallback: only the canonical `arguments`
+        // key counts, so a plain JSON answer isn't dressed up as a call.
+        parse_arguments_value(value.get("arguments"))
+    };
     Some(ParsedToolCall {
         name,
         arguments,
@@ -85,11 +112,25 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
 }
 
 pub(crate) fn parse_tool_calls_from_json_value(value: &serde_json::Value) -> Vec<ParsedToolCall> {
+    // Tagged contexts (callers reach here via an explicit tool-call marker)
+    // accept the argument-key aliases.
+    parse_tool_calls_from_json_value_aliased(value, true)
+}
+
+/// Like [`parse_tool_calls_from_json_value`], but lets the caller forbid
+/// generic arg-key aliases on a **bare** singleton/array object. The
+/// `tool_calls`-keyed envelope always stays permissive — that key is an
+/// unambiguous tool-call marker even on the whole-response path.
+pub(crate) fn parse_tool_calls_from_json_value_aliased(
+    value: &serde_json::Value,
+    allow_arg_aliases: bool,
+) -> Vec<ParsedToolCall> {
     let mut calls = Vec::new();
 
     if let Some(tool_calls) = value.get("tool_calls").and_then(|v| v.as_array()) {
         for call in tool_calls {
-            if let Some(parsed) = parse_tool_call_value(call) {
+            // `tool_calls` entries are explicitly tool-call shaped → widen.
+            if let Some(parsed) = parse_tool_call_value_aliased(call, true) {
                 calls.push(parsed);
             }
         }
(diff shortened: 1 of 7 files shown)


src/openhuman/agent/harness/parse.rs
```diff
@@ -55,7 +76,7 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
         return None;
     }
 
-    let arguments = parse_arguments_value(value.get("arguments"));
+    let arguments = first_args_by_keys(value);
     Some(ParsedToolCall {
```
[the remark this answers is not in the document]
sanil-23: Fixed in a28523a0. Threaded an `allow_arg_aliases` flag: the whole-response parse path (`parse_tool_calls` on a top-level JSON object) now passes `false`, so a bare `{"name":..,"input":..}` reply only honours the canonical `arguments` key and won't slurp generic fields into tool args. Alias widening is retained for explicit tool-call markers — `tool_calls` array entries, `<tool_call>`/`<invoke>` tags, and the `function`-wrapped shape. New regression test `whole_response_singleton_ignores_generic_arg_aliases`.

src/openhuman/agent/harness/session/turn.rs
```diff
@@ -940,26 +990,89 @@ impl Agent {
...
+            let mut checkpoint_messages = base_messages;
+            checkpoint_messages.push(ChatMessage::assistant(checkpoint.clone()));
+            self.persist_session_transcript(
+                &checkpoint_messages,
+                cumulative_input_tokens,
+                cumulative_output_tokens,
+                cumulative_cached_input_tokens,
+                cumulative_charged_usd,
+                last_turn_usage.as_ref(),
+            );
```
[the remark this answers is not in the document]
sanil-23: Fixed in a28523a0. `summarize_iteration_checkpoint` now returns `(String, Option<UsageInfo>)`; the max-iter caller folds that usage into `cumulative_input/output/cached_tokens`, `cumulative_charged_usd` and `last_turn_usage` (and calls `context.record_usage`) before persisting — mirroring the normal final-response path, so the extra checkpoint call is no longer undercounted and prior-iteration usage isn't mis-attributed to the checkpoint message. New test `turn_checkpoint_usage_is_folded_into_transcript_accounting` asserts the persisted transcript totals equal the checkpoint call's usage.

src/openhuman/agent/harness/session/turn.rs
```diff
@@ -1875,6 +1988,90 @@ impl Agent {
...
+        match result {
+            Ok(resp) => {
+                // Strip any stray tool-call XML a text-mode model may have
+                // emitted; keep only the prose.
+                let (text, _calls) = self.tool_dispatcher.parse_response(&resp);
+                if text.trim().is_empty() {
+                    resp.text.unwrap_or_default()
+                } else {
+                    text
+                }
```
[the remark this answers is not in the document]
sanil-23: Fixed in a28523a0. The `Ok(resp)` branch now keeps the parsed calls: when `parse_response` leaves no prose but DID parse tool-call markup, it returns an empty string so the caller uses `build_deterministic_checkpoint` instead of re-emitting `resp.text` (which could persist raw `<tool_call>…` markup verbatim). `resp.text` is only used as prose when no tool-call markup was present.

sanil-23:
Pushed `e7c55c9c` addressing the 2 nitpick comments from the last review:

- **`subagent_runner/ops.rs`** — `find_action` now emits a `tracing::debug` line when 2+ actions normalize to the same slug (the ambiguous case that deliberately refuses to guess), so the collision is diagnosable.
- **`parse.rs`** — the malformed-`<tool_call>` warning now uses the stable grep-friendly `[agent_parse]` prefix with a structured `body_chars` field, consistent with the other harness log families.

The 3 actionable comments were addressed in `a28523a0` (replied inline): singleton-path alias restriction, checkpoint usage accounting, and not re-emitting raw tool-call markup as the checkpoint.

src/openhuman/agent/harness/parse.rs
```diff
@@ -55,7 +97,13 @@ pub(crate) fn parse_tool_call_value(value: &serde_json::Value) -> Option<ParsedT
...
     }
 
-    let arguments = parse_arguments_value(value.get("arguments"));
+    let arguments = if allow_arg_aliases {
+        first_args_by_keys(value)
+    } else {
+        // Singleton whole-response fallback: only the canonical `arguments`
+        // key counts, so a plain JSON answer isn't dressed up as a call.
+        parse_arguments_value(value.get("arguments"))
+    };
```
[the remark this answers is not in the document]
sanil-23: Fixed in 879a9702. On the bare whole-response path (`parse_tool_call_value_aliased(.., false)`) the parser now requires the canonical `arguments` key as an explicit tool-call marker — without it, it returns `None`, so a plain reply like `{"name":"Alice","input":{…}}` stays text instead of dispatching a tool. Tagged contexts (`<tool_call>`/`<invoke>`, `tool_calls` array, `function` wrapper) reach the parser with aliases enabled and keep the permissive behaviour. The regression test now asserts the bare object is no longer a tool call (and that a bare object *with* canonical `arguments` still is).

src/openhuman/agent/harness/session/turn.rs
```diff
@@ -443,6 +478,14 @@ impl Agent {
         // Collect tool call records across all iterations for post-turn hooks
         let mut all_tool_records: Vec<ToolCallRecord> = Vec::new();
 
+        // Trim-robust digest of THIS turn's tool calls + results, compiled as
+        // the loop runs. Used as the *only* context for the max-iteration
+        // checkpoint summary, so it compiles "what I did this turn" without
+        // the prior conversation or system prompt bleeding in — and it's
+        // immune to history trimming (which drops/reorders from the front).
+        // The persisted transcript is unaffected (bug-report-2026-05-26 A1).
+        let mut turn_tool_digest = String::new();
```
M3gA-Mind, line 490: `turn_tool_digest` is unbounded — each entry is capped at 800 chars, so with the default cap of 10 iterations the practical max is ~8 KB. Fine as-is, but worth a short comment so a future increase to `max_tool_iterations` comes with an obvious prompt to revisit this.

src/openhuman/agent/harness/session/turn.rs
```diff
@@ -940,26 +1006,130 @@ impl Agent {
...
+
+            let base_messages = last_provider_messages
+                .clone()
+                .unwrap_or_else(|| self.tool_dispatcher.to_provider_messages(&self.history));
+            // Summarize ONLY this turn's work: feed the compiled tool-call
+            // digest (no system prompt, no prior conversation), not the full
+            // conversation. `base_messages` above is still used for the
+            // transcript persist below, so the saved transcript is unchanged
+            // (bug-report-2026-05-26 A1).
+            let turn_summary_input = vec![ChatMessage::user(format!(
```
M3gA-Mind, line 1035: `user_message` here is captured from the outer `turn(&mut self, message: &str)` parameter — correct, but implicit enough to confuse a future reader. A one-line comment like `// user_message captured from turn() parameter` would help.

src/openhuman/agent/harness/session/turn.rs
```diff
@@ -940,26 +1006,130 @@ impl Agent {
...
+                .push(ConversationMessage::Chat(ChatMessage::assistant(
+                    checkpoint.clone(),
+                )));
+            self.trim_history();
+
+            // Persist the checkpoint so the transcript ends on a
+            // well-formed assistant message (never a dangling tool cycle).
+            let mut checkpoint_messages = base_messages;
+            checkpoint_messages.push(ChatMessage::assistant(checkpoint.clone()));
+            self.persist_session_transcript(
```
M3gA-Mind, line 1111: Worth a comment here: `base_messages` (and therefore the persisted transcript) does NOT include the tool results from the final capped iteration — those were processed after the last `last_provider_messages` snapshot and before the cap fired. The transcript is structurally correct (ends on an assistant message) and the checkpoint prose covers the work done, but the omission is non-obvious to a future reader. Something like: `// base_messages ends before the final iteration's tool results; checkpoint prose covers what the model did.`

src/openhuman/agent/harness/subagent_runner/ops.rs
```diff
@@ -210,13 +211,79 @@ impl LazyToolkitResolver {
...
+                    );
+                    return Some(action);
+                }
+                // Ambiguous: 2+ actions normalize to the same slug (e.g.
+                // `read_file` and `ReadFile` → `readfile`). We deliberately
+                // refuse to guess, but log it so the collision is diagnosable.
+                tracing::debug!(
+                    requested = %name,
+                    norm = %norm,
+                    "[subagent_runner] ambiguous normalized-slug match — multiple actions resolve to the same slug; not resolving"
```
M3gA-Mind, line 264: This should be `tracing::warn!` rather than `tracing::debug!`. Two registered actions normalizing to the same slug is a configuration anomaly (silent ambiguity in the toolkit), not routine noise — it won't surface in production unless someone is actively tailing debug logs. At `warn` level it would show up in normal operator log streams and be diagnosable without special filtering.

src/openhuman/agent/harness/subagent_runner/ops.rs
```diff
@@ -1785,7 +1867,66 @@ async fn run_inner_loop(
...
+            tracing::warn!(
+                agent_id = %agent_id,
+                task_id = %task_id,
+                error = %e,
+                "[subagent_runner] checkpoint summary call failed — using deterministic fallback"
+            );
+            deterministic
+        }
+    };
+    Ok((checkpoint, max_iterations, usage))
```
M3gA-Mind, line 1935: Unlike the main-agent path, the subagent checkpoint is intentionally NOT persisted to a transcript here — the parent agent's transcript captures the delegated result, so there's no data loss. Worth a short comment to that effect so a future reader doesn't treat the missing `persist_subagent_transcript` call as a bug to fix.

M3gA-Mind (commented):
## Review

**Summary:** Well-engineered fix for the three production bugs from `bug-report-2026-05-26`. The two-layer safety net on A1 (model checkpoint → deterministic fallback) is the right design and the test coverage is thorough — happy path, empty-summary fallback, and usage-accounting regression guard are all covered. A2's uniqueness check on normalized slugs is correctly conservative. A3's `allow_arg_aliases` boolean is a clean lever.

A few inline notes below — main asks before merge:
- Promote the ambiguous-slug-collision log to `warn` (it's a configuration anomaly, not routine noise)
- Add a comment at the checkpoint `persist_session_transcript` call noting the last iteration's tool results are intentionally absent from the base messages
- Note the subagent path intentionally skips transcript persist (so a future reader doesn't 'fix' it)

Everything else is minor / informational.

sanil-23:
Thanks @M3gA-Mind — all addressed in 6e545f24:

- **ops.rs slug collision** → promoted `debug` → `tracing::warn!` (toolkit config anomaly; now surfaces in normal operator logs).
- **turn.rs `turn_tool_digest`** → comment noting the ~8 KB bound at the default 10-iteration cap (revisit if `max_tool_iterations` is raised).
- **turn.rs `user_message`** → comment noting it is the `turn(&mut self, message)` parameter.
- **turn.rs persist** → comment noting `base_messages` (and thus the persisted transcript) intentionally omits the final capped iteration's tool results — the checkpoint prose covers them and the transcript stays structurally correct.
- **subagent checkpoint** → comment noting it is intentionally *not* transcript-persisted (the parent agent's transcript captures the delegated result), so it isn't mistaken for a missing-persist bug.

No behavior change beyond the log level.

M3gA-Mind:
All good — reviewed and approved. CI failures are systemic (main is also red), not caused by this PR. Ready to merge once the infrastructure recovers.

PR #2684, opened 2026-05-26
fix(memory): retry truncated extractions + release job locks on graceful shutdown

sanil-23:
## Summary

- **I1 (MED):** the memory entity-extractor now **retries** when an LLM response is truncated mid-JSON (a `serde` EOF) instead of silently dropping the whole chunk; genuine wrong-shape responses still fall back to empty immediately.
- **I2 (MED):** the memory-queue worker releases its in-flight job locks on **graceful shutdown** so a clean restart re-claims the work immediately; the startup stale-lock recovery log drops `warn` → `info`.

## Problem

From tester logs (`bug-report-2026-05-26`):

- **I1** — the entity-extraction LLM returned partial JSON (the stream closed before the closing brace — token budget / server-side timeout). Every parse error returned an empty extraction with **no retry**, so each affected chunk was silently dropped (`EOF while parsing ... — returning empty extraction`, 4+ occurrences).
- **I2** — `recovered N stale-locked job(s) at startup` fired on every launch at `warn`. The worker had no graceful-shutdown handling, so a clean restart left in-flight jobs `running` until their lease expired, surfacing as a stale-lock recovery rather than an immediate re-claim.

## Solution

- **I1** — `try_extract` distinguishes a truncated response (`serde_json::Error::is_eof()`) from a wrong-shape one. Truncation is transient, so it returns a retryable `None` and the existing bounded backoff loop (`MAX_ATTEMPTS = 3`) tries again; non-EOF shape errors still return `Some(empty)` immediately (retrying identical garbage wouldn't help).
- **I2** — `store::release_running_locks` flips this process's `running` rows back to `ready` (clearing the lease). The worker registers it via `core::shutdown::register`, so a graceful exit hands work back instantly. The startup recovery log is now `info` — a non-empty recovery means a hard kill, which is expected and self-healing.

## Submission Checklist

> If a section does not apply, marked `N/A` with a one-line reason.

- [x] Tests added or updated (happy path + failure/edge) — truncated response triggers a retry that recovers entities, and a wrong-shape response still returns empty without retrying (`llm_tests.rs`); `release_running_locks` resets a `running` row regardless of lease (`store.rs`).
- [x] **Diff coverage ≥ 80%** — added unit tests target the changed lines (EOF branch, release fn); CI `diff-cover` enforces the exact gate.
- [x] Coverage matrix updated — `N/A: behaviour-only bug fix; no feature row added/removed/renamed.`
- [x] All affected feature IDs listed under Related — `N/A: no matrix feature touched.`
- [x] No new external network dependencies — none added.
- [x] Manual smoke checklist updated — `N/A: no release-cut surface changed.`
- [x] Linked issue closed via `Closes #NNN` — `N/A: no tracking issue filed; sourced from bug-report-2026-05-26 (private tester logs).`

## Impact

- Memory pipeline only. Fewer dropped extractions on long entity lists; cleaner restart behaviour and quieter startup logs. No API/schema changes, no migration. The graceful-release hook assumes a single core worker pool (true today).

## Related

- Closes: N/A (no linked issue; bug-report-2026-05-26 I1/I2)
- Follow-up PR(s)/TODOs: I1 — persistent (deterministic) truncation needs a response token-budget bump; the retry here addresses transient cutoffs.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/memory-extract-job-locks`
- Commit SHA: `1c63cfa5`

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry on truncated/EOF JSON responses from LLM providers instead of treating them as final failures.
  * Do not retry on well-formed but wrong-shape JSON responses.

* **Improvements**
  * Enhanced graceful shutdown to release in-flight job locks.
  * Startup recovery logging for stale-locked jobs adjusted for clarity.

* **Tests**
  * Added regression tests verifying lock release behavior and extractor retry behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2684?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory_queue/store.rs
+++ b/src/openhuman/memory_queue/store.rs
@@ -325,8 +325,31 @@ pub fn recover_stale_locks(config: &Config) -> Result<usize> {
             params![now_ms],
         )?;
         if n > 0 {
-            log::warn!("[memory::jobs] recovered {n} stale-locked job(s) at startup");
+            // Info, not warn: with graceful shutdown releasing in-flight
+            // locks (see `release_running_locks`), a non-empty recovery now
+            // means the previous process was hard-killed — expected and
+            // self-healing, not actionable (bug-report-2026-05-26 I2).
+            log::info!("[memory::jobs] recovered {n} stale-locked job(s) at startup");
         }
+        Ok(n)
+    })
+}
+
+/// Release this process's in-flight job locks on a *graceful* shutdown:
+/// flip every `running` row back to `ready` so the work is immediately
+/// re-claimable on next launch instead of waiting out the lease and
+/// surfacing as a stale-lock recovery. The core runs a single worker pool,
+/// so any `running` row at clean-shutdown time was claimed by us.
+/// Returns the number of rows released (bug-report-2026-05-26 I2).
+pub fn release_running_locks(config: &Config) -> Result<usize> {
+    with_connection(config, |conn| {
+        let n = conn.execute(
+            "UPDATE mem_tree_jobs
+                SET status = 'ready',
+                    locked_until_ms = NULL
+              WHERE status = 'running'",
+            [],
+        )?;
         Ok(n)
     })
 }
@@ -608,6 +631,27 @@ mod tests {
         assert_eq!(recovered, 1);
         let row = get_job(&cfg, &id).unwrap().unwrap();
         assert_eq!(row.status, JobStatus::Ready);
+    }
+
+    #[test]
+    fn release_running_locks_resets_running_rows_regardless_of_lease() {
+        let (_tmp, cfg) = test_config();
+        let payload = ExtractChunkPayload {
+            chunk_id: "c-release".into(),
+        };
+        let nj = NewJob::extract_chunk(&payload).unwrap();
+        let id = enqueue(&cfg, &nj).unwrap().unwrap();
+
+        // Claim with a long, still-valid lease. `recover_stale_locks` would
+        // NOT touch this (not expired), but a graceful shutdown release must
+        // so a clean restart re-claims it immediately (bug-report-2026-05-26 I2).
+        let _ = claim_next(&cfg, DEFAULT_LOCK_DURATION_MS).unwrap().unwrap();
+
+        let released = release_running_locks(&cfg).unwrap();
+        assert_eq!(released, 1);
+        let row = get_job(&cfg, &id).unwrap().unwrap();
+        assert_eq!(row.status, JobStatus::Ready);
+        assert!(row.locked_until_ms.is_none());
     }
 
     /// Happy path: a non-stale settlement still succeeds after the claim-token
--- a/src/openhuman/memory_queue/store.rs
+++ b/src/openhuman/memory_queue/store.rs
@@ -343,6 +343,12 @@ pub fn recover_stale_locks(config: &Config) -> Result<usize> {
 /// Returns the number of rows released (bug-report-2026-05-26 I2).
 pub fn release_running_locks(config: &Config) -> Result<usize> {
     with_connection(config, |conn| {
+        // TODO(multi-process): this releases *every* `running` row, with no
+        // process-scoped guard (e.g. a `worker_pid` / session-token column).
+        // Correct today because a single worker pool is the invariant, but if
+        // multi-process workers are ever introduced, a rolling restart with
+        // two overlapping processes would let one shutdown release the other's
+        // in-flight locks. Add a per-owner predicate here when that happens.
         let n = conn.execute(
             "UPDATE mem_tree_jobs
                 SET status = 'ready',
--- a/src/openhuman/memory_queue/worker.rs
+++ b/src/openhuman/memory_queue/worker.rs
@@ -18,7 +18,7 @@ use crate::openhuman::config::Config;
 use crate::openhuman::memory_queue::handlers;
 use crate::openhuman::memory_queue::redact::scrub_for_log;
 use crate::openhuman::memory_queue::store::{
-    claim_next, mark_deferred, mark_done, mark_failed, recover_stale_locks,
+    claim_next, mark_deferred, mark_done, mark_failed, recover_stale_locks, release_running_locks,
     DEFAULT_LOCK_DURATION_MS,
 };
 use crate::openhuman::memory_queue::types::JobOutcome;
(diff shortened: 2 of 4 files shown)


src/openhuman/memory_tree/score/extract/llm.rs
```diff
@@ -213,6 +213,20 @@ impl LlmEntityExtractor {
...
+                // Truncated mid-JSON: the response stream closed before the
+                // closing brace (token budget or a server-side timeout cut
+                // it short). Unlike a wrong-shape body, this is transient —
+                // signal a retryable failure (`None`) so `extract`'s backoff
+                // loop tries again rather than silently dropping the whole
+                // chunk (bug-report-2026-05-26 I1).
+                log::warn!(
+                    "[memory_tree::extract::llm] LLM response truncated mid-JSON ({e}); \
+                     response_chars={} — retrying",
+                    raw.chars().count()
```
M3gA-Mind: Nit: `raw.chars().count()` iterates every Unicode scalar in the (potentially large) raw response just to get a length hint for the log. `raw.len()` (byte length) is cheaper, sufficient as a diagnostic hint, and avoids the O(n) walk on a hot error path.

src/openhuman/memory_queue/worker.rs
```diff
@@ -70,6 +70,31 @@ pub fn start(config: Config) {
...
 
+        // Release in-flight locks on graceful shutdown so a clean restart
+        // re-claims the work immediately instead of waiting out the lease
+        // (which surfaced as a stale-lock recovery warn on every launch).
+        // Hard kills still fall back to lease-expiry recovery at startup
+        // (bug-report-2026-05-26 I2).
+        let shutdown_cfg = config.clone();
+        crate::core::shutdown::register(move || {
+            let cfg = shutdown_cfg.clone();
+            async move {
```
M3gA-Mind, line 86: Unnecessary double-clone: `shutdown_cfg` is moved into the closure and then immediately cloned again into `cfg`. Since `shutdown_cfg` isn't used after the `register` call, you can just move it directly:

```rust
crate::core::shutdown::register(move || {
    let cfg = shutdown_cfg; // move, not clone
    async move {
        match release_running_locks(&cfg) { ...
```

src/openhuman/memory_queue/store.rs
```diff
@@ -325,12 +325,35 @@ pub fn recover_stale_locks(config: &Config) -> Result<usize> {
...
+/// surfacing as a stale-lock recovery. The core runs a single worker pool,
+/// so any `running` row at clean-shutdown time was claimed by us.
+/// Returns the number of rows released (bug-report-2026-05-26 I2).
+pub fn release_running_locks(config: &Config) -> Result<usize> {
+    with_connection(config, |conn| {
+        let n = conn.execute(
+            "UPDATE mem_tree_jobs
+                SET status = 'ready',
+                    locked_until_ms = NULL
+              WHERE status = 'running'",
```
M3gA-Mind, line 356: The `WHERE status = 'running'` filter has no process-scoped guard (e.g. a `worker_pid` or session-token column). Safe today because a single worker pool is the invariant, but a rolling-restart scenario with two overlapping processes would cause one shutdown to release the other's in-flight locks. Worth a `// TODO` here to flag it for if/when multi-process workers are introduced.

src/openhuman/memory_tree/score/extract/llm_tests.rs
```diff
@@ -428,3 +428,47 @@ fn truncate_for_log_long_input_appends_ellipsis() {
...
+    let ex = LlmEntityExtractor::new(LlmExtractorConfig::default(), mock.clone());
+    let out = ex.extract("Alice met Bob.").await.unwrap();
+    assert_eq!(
+        mock.calls.load(Ordering::SeqCst),
+        2,
+        "truncation should trigger a retry, not a silent drop"
+    );
+    assert_eq!(out.entities.len(), 1);
+    assert_eq!(out.entities[0].text, "Alice");
+}
```
M3gA-Mind, line 474: The retry-on-truncation path is well covered. Consider adding a companion test for the no-retry path: a non-EOF wrong-shape response (e.g. `{"wrong":true}`) should return immediately (`calls == 1`) without triggering a retry. The PR description claims this behaviour but there's no regression guard for it.

M3gA-Mind (commented):
Two clean, well-scoped bug fixes with good test coverage. A few minor nits inline — nothing blocking.

src/openhuman/memory_tree/score/extract/llm.rs
sanil-23: Fixed in a39d7d54. Switched to `raw.len()` and renamed the field to `response_bytes` — byte length is a sufficient diagnostic hint and avoids the O(n) Unicode walk on the error path.

src/openhuman/memory_queue/worker.rs
sanil-23: Good eye, but the move doesn't compile here: `shutdown::register` is bound `F: Fn() -> Fut` (not `FnOnce`), so the closure may be invoked more than once and each call must hand the returned `async move` future its own owned `Config`. `let cfg = shutdown_cfg;` gives `error[E0507]: cannot move out of `shutdown_cfg`, a captured variable in an `Fn` closure` (verified locally). The per-call `.clone()` is load-bearing, not redundant. I added a comment in a39d7d54 explaining this so it isn't 'optimized' away later.

src/openhuman/memory_queue/store.rs
sanil-23: Added in a39d7d54 — a `TODO(multi-process)` on `release_running_locks` documenting that the `WHERE status = 'running'` filter is correct only under the single-worker-pool invariant, and that a multi-process rolling restart would need a per-owner (`worker_pid`/session-token) predicate.

src/openhuman/memory_tree/score/extract/llm_tests.rs
sanil-23: Added in a39d7d54: `extract_does_not_retry_on_wrong_shape_response` feeds a complete, non-EOF wrong-shape body (`{"entities":123}` — a type error, `is_eof()==false`) and asserts `calls == 1` with an empty extraction. Directly guards the `Err(e) if e.is_eof()` split so only truncations retry.

PR #2698, opened 2026-05-26
fix(memory): scope Gmail cleanup to disconnected connection

### LINKED ISSUE #2598: bug(memory): composio disconnect deletes memory chunks for all connections of the same toolkit
## Summary

When a user disconnects a Composio integration, `composio_toolkit_source_kind` in `src/openhuman/composio/ops.rs` maps the toolkit slug to a broad LIKE prefix (e.g. `gmail:%`, `notion:%`). The subsequent `delete_chunks_by_source` call then deletes **all** memory chunks whose `source_id` matches that prefix — not just chunks belonging to the specific connection being disconnected.

This means that in multi-connection scenarios (e.g. a user with two active Gmail connections), disconnecting one connection will silently wipe memory chunks indexed under the other connection(s) for the same toolkit.

## Steps to Reproduce

1. Create two active Composio Gmail connections (connection A and connection B).
2. Ingest memory from both connections — chunks are stored with `source_id` values like `gmail:<id_A>:...` and `gmail:<id_B>:...`.
3. Disconnect connection A with `clear_memory: true`.
4. Observe that chunks belonging to connection B are also deleted because the LIKE pattern `gmail:%` matches both.

## Expected Behaviour

Only memory chunks scoped to the specific `connection_id` being disconnected should be deleted.

## Suggested Fix

Derive the `source_id_pattern` from the concrete `connection_id` (e.g. `gmail:<connection_id>:%`) rather than from the toolkit slug alone. This requires understanding how `source_id` values are structured per-connection vs. per-toolkit in the memory tree.

## Context

Deferred from PR #2546 (feat: opt-in memory deletion on channel / composio disconnect) — https://github.com/tinyhumansai/openhuman/pull/2546

Flagged by @coderabbitai during review, deferred by @senamakel for a follow-up.

Felyx-Fu:
## Summary

- Fixes Composio Gmail clear-memory cleanup so disconnecting one connection does not delete chunks owned by another Gmail connection.
- Adds an owner-scoped chunk deletion helper that preserves shared source ingest gates while removing orphaned gates.
- Adds regression coverage for multi-connection Gmail cleanup and owner-scoped deletion.
- Fixes an existing `AutonomySettingsPatch` test initializer so Rust lib tests compile after recent field additions.

## Problem

- `composio_delete_connection(clear_memory=true)` could clear memory at a source scope that is shared by multiple Gmail connections.
- Gmail ingest stores account-scoped chunks under source ids such as `gmail:<account-slug>`, with the concrete connection stored in `metadata.owner` as `gmail-sync:<connection_id>`.
- Deleting by account source id can remove chunks belonging to another active connection for the same toolkit/account-shaped source.

## Solution

- Add `delete_chunks_by_owner` in the chunk store for exact owner cleanup, while retaining exact/prefix source deletion for Slack, Notion, Drive, and connection-scoped legacy ids.
- Change Gmail memory cleanup targets to use `Owner(Email, gmail-sync:<connection_id>)` plus exact/prefix `gmail:<connection_id>` fallbacks for connection-scoped ids.
- Preserve `mem_tree_ingested_sources` rows when owner cleanup leaves chunks for the same source id, and remove the gate only when the deleted owner was the last remaining source owner.
- Keep the cleanup conservative for legacy Gmail chunks that lack a connection owner, because those cannot be safely attributed to one connection without risking the cross-connection deletion bug.

## Submission Checklist

- [x] **Diff coverage >= 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/coverage.yml`](../.github/workflows/coverage.yml). Focused Rust coverage was added; CI coverage gate is the source of truth for the final threshold.
- [x] Coverage matrix updated — N/A: backend bug fix for existing memory cleanup behavior; no feature row added/removed/renamed.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no matrix feature ID applies.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: no release-cut smoke surface changed.

## Impact

- Runtime/platform impact: Rust core memory cleanup only.
- Privacy/data impact: `clear_memory=true` remains destructive for the disconnected connection, but no longer deletes chunks owned by another Gmail connection sharing the same account-shaped source id.
- Compatibility: legacy exact/prefix connection-id source ids are still handled; legacy Gmail account chunks without owner attribution are left untouched rather than deleted ambiguously.

## Related

- Closes #2598
- Follow-up PR(s)/TODOs: None.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `Felix/issue-2598-composio-memory-scope`
- Commit SHA: `7a4f32e49308a9fe62393512cbcfa255ab86c7ca`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` via pre-push `pnpm format:check` (passed)
- [x] `pnpm typecheck` (GitHub Type Check passed on PR; local run blocked, see below)
- [x] Focused tests:
  - `cargo test --manifest-path Cargo.toml --lib composio_delete_connection_clear_memory_keeps_other_gmail_connections`
  - `cargo test --manifest-path Cargo.toml --lib delete_chunks_by_owner_preserves_other_owners_for_same_source`
  - `cargo test --manifest-path Cargo.toml --lib delete_chunks_by_source_removes_chunks_side_rows_and_ingest_gate`
  - `cargo test --manifest-path Cargo.toml --lib composio_delete_connection_clear_memory_deletes_slack_source`
  - `cargo test --manifest-path Cargo.toml --lib apply_autonomy_settings_updates_action_budget`
  - `cargo test --manifest-path Cargo.toml --lib delete_chunks_by_owner_preserves_other_owners_for_same_source` (rerun after review fix)
  - `cargo test --manifest-path Cargo.toml --lib composio_delete_connection_clear_memory_keeps_other_gmail_connections` (rerun after review fix)
  - `cargo test --manifest-path Cargo.toml --lib build_chat_runtime_defaults_to_openhuman_resolved_model` (CI drift follow-up)
  - `cargo test --manifest-path Cargo.toml --lib make_openhuman_backend_forwards_unknown_hint_verbatim` (CI drift follow-up)
  - `cargo test --manifest-path Cargo.toml --lib make_openhuman_backend_translates_summarization_hint` (CI drift follow-up)
  - `cargo test --manifest-path Cargo.toml --lib known_hints_pass` (CI drift follow-up)
  - `cargo test --manifest-path Cargo.toml --lib write_read_and_list_memory_files_roundtrip` (CI drift follow-up)
  - `cargo test --manifest-path Cargo.toml --lib list_memory_files_skips_internal_sqlite_artifacts` (CodeRabbit review follow-up)
  - `cargo fmt --manifest-path Cargo.toml -- --check`, `git diff --check`, and `cargo test --manifest-path Cargo.toml --lib list_memory_files_skips_internal_sqlite_artifacts` (CodeRabbit comment follow-up)
  - `pnpm -C app exec vitest run --config test/vitest.config.ts src/utils/__tests__/loopbackOauthListener.test.ts` (CI drift follow-up)
- [x] Rust fmt/check (if changed):
  - git diff --check (CI drift follow-up)
  - pnpm -C app exec prettier --check src/utils/__tests__/loopbackOauthListener.test.ts (CI drift follow-up)
  - `cargo fmt --manifest-path Cargo.toml -- --check`
  - `cargo check --manifest-path Cargo.toml`
- [x] Tauri fmt/check (if changed): N/A for touched code; pre-push `pnpm rust:check` reached and compiled the Tauri shell before the unrelated command-token script failure.
- [x] CI rerun trigger: pushed empty commit `51a328f9a517f4686ceba409593439d6c95cd6ab` after the Windows secrets ACL job completed its test step successfully but the workflow ended `cancelled`; direct `gh run rerun --failed` was blocked by missing repo admin rights.
- [x] CI timeout follow-up: `git diff --check`; increased only the Windows secrets ACL job timeout from 20 to 30 minutes after two runs were cancelled around the 20-minute job limit while compiling/running `cargo test -p openhuman -- security::secrets --nocapture`.

### Validation Blocked
- `command:` `pnpm compile`
- `error:` `src/features/human/Mascot/RiveMascot.tsx(10,8): error TS2307: Cannot find module '@rive-app/react-webgl2' or its corresponding type declarations.`
- `impact:` Local TypeScript typecheck is blocked by an unrelated missing frontend dependency/type declaration in an untouched file.

- `command:` `pnpm --dir app run lint:commands-tokens`
- `error:` Windows/PowerShell invocation mis-parses the package script's `bash -c '... { ... } ...'`, ending with `The system cannot find the path specified.` and `'{'' is not recognized as an internal or external command`.
- `impact:` Pre-push hook could not complete locally; branch was pushed with `--no-verify` after format/check/focused Rust validation passed.

- `command:` `cargo test --manifest-path Cargo.toml composio_delete_connection_clear_memory_keeps_other_gmail_connections`
- `error:` After the matching lib test passed, Cargo attempted to execute unrelated integration test binary `mcp_setup_e2e` and Windows returned `os error 740` (requires elevation).
- `impact:` Used `--lib` targeted Rust tests instead to avoid unrelated elevated e2e binaries in this local shell.

### Behavior Changes
- Intended behavior change: Gmail Composio memory cleanup deletes chunks owned by the disconnected connection only.
- User-visible effect: disconnecting Gmail connection A with memory cleanup no longer silently wipes Gmail chunks owned by connection B.

### Parity Contract
- Legacy behavior preserved: non-Gmail cleanup still uses existing exact/prefix source targets; Gmail connection-id exact/prefix fallbacks remain.
- Guard/fallback/dispatch parity checks: focused Rust tests cover Gmail multi-connection retention, Slack cleanup, source cleanup cascade/gate behavior, and the owner-scoped gate preservation path.

- Duplicate PR(s): N/A
- Canonical PR: This PR

## Summary by CodeRabbit

* **Refactor**
  * Connection deletion now removes only data scoped to the specific connection owner, preserving other connections' data and ingestion-state when other owners remain.

* **Bug Fixes**
  * Memory file listing omits internal SQLite artifact files from user-facing file lists.
  * OAuth loopback fallback uses a shorter timeout for faster failure handling.

* **Tests**
  * Added/updated tests covering connection-scoped memory deletion, ingestion-state behavior, OAuth fallback, provider hint handling for summarization, and default model resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2698?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


M3gA-Mind (commented):
# PR #2698 — fix(memory): scope Gmail cleanup to disconnected connection

## Walkthrough

This PR fixes a data-safety bug where `composio_delete_connection(clear_memory=true)` could silently delete memory chunks belonging to a *different* active Gmail connection that happened to share the same account-scoped source id (e.g. `gmail:pilot-at-example-dot-com`). The fix introduces an `Owner`-scoped cleanup target and a new `delete_chunks_by_owner` store function that filters by `metadata.owner` instead of `source_id`, preserving ingest gates for sources that still have chunks from other connections. A compile fix for `AutonomySettingsPatch` is bundled as a drive-by.

The core logic is correct and the tests are well-targeted. One project-standard gap (missing debug logging on the new code path) needs attention before this is complete.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/composio/ops.rs` | Adds `Owner` variant to `MemoryCleanupTarget`; rewrites `gmail_memory_sources_for_connection` to use owner-scoped + connection-id fallback targets |
| `src/openhuman/memory_store/chunks/store.rs` | Adds `delete_chunks_by_owner`; refactors `delete_chunks_by_source_filter` to accept separate chunk-match and ingest-gate-match predicates; adds orphan check to conditionally preserve gates |
| `src/openhuman/memory_store/chunks/store_tests.rs` | New unit test for `delete_chunks_by_owner` covering gate preservation vs. cleanup |
| `src/openhuman/composio/ops_test.rs` | New integration test for multi-connection Gmail cleanup; `sample_memory_chunk` now threads owner through chunk content/ID |
| `src/openhuman/config/ops_tests.rs` | Drive-by compile fix: `..Default::default()` for `AutonomySettingsPatch` |

## Actionable comments (2)

### ⚠️ Major

#### 1. `src/openhuman/memory_store/chunks/store.rs:839-884` — New orphan-check and gate-preservation logic has zero diagnostic logging

CLAUDE.md is explicit: *"Default to verbose diagnostics on new/changed flows … Changes lacking diagnosis logging are incomplete."* The new orphaned-sources check and conditional gate deletion are exactly the kind of state transitions that need a debug trail — when a cleanup pass silently preserves a gate, there's currently no way to confirm this happened intentionally vs. a bug hiding it.

**Suggested change:**
```rust
// after: let mut orphaned_deleted_sources = HashSet::new();
let mut orphaned_deleted_sources = HashSet::new();
for source_id in &deleted_source_ids {
    let remaining: i64 = tx.query_row(
        "SELECT COUNT(*) FROM mem_tree_chunks WHERE source_kind = ?1 AND source_id = ?2",
        params![source_kind.as_str(), source_id],
        |row| row.get(0),
    )?;
    if remaining == 0 {
        log::debug!(
            "[memory::chunk_store] delete_filter: source_id={source_id} fully orphaned — will remove ingest gate"
        );
        orphaned_deleted_sources.insert(source_id.clone());
    } else {
        log::debug!(
            "[memory::chunk_store] delete_filter: source_id={source_id} still has {remaining} chunk(s) from other owners — preserving ingest gate"
        );
    }
}
```

### 💡 Refactor / suggestion

#### 2. `src/openhuman/memory_store/chunks/store.rs:784-806` — `delete_chunks_by_owner` loads all chunks for the source_kind via a full scan; the `owner` index goes unused

`delete_chunks_by_source_filter` always queries `WHERE source_kind = ?1` and filters in Rust. For the existing exact/prefix deletions this is fine because source_id filtering would need Rust-side prefix logic anyway. But `delete_chunks_by_owner` matches a single exact owner string — SQLite's `idx_mem_tree_chunks_owner` index could serve this directly.

This is a pre-existing pattern that the PR doesn't worsen, and for a once-per-disconnect cleanup the performance impact is low. Worth noting for a future cleanup when the table is large.

**If fixing now:** consider a targeted SELECT with `WHERE source_kind = ?1 AND owner = ?2` inside the `delete_chunks_by_owner` implementation rather than routing through the generic filter.

## Nitpicks (3)

- `src/openhuman/composio/ops.rs:590` — `dedupe_memory_targets(vec![...])` is a no-op here: all four targets are structurally distinct by construction. The call is defensive and consistent with other callers, but the comment-free reader might wonder what duplicates are possible. A brief inline comment (or simply removing the call) would clarify intent.

- `src/openhuman/composio/ops_test.rs:549-594` — The integration test `composio_delete_connection_clear_memory_keeps_other_gmail_connections` doesn't assert the ingest-gate state for the shared source `"gmail:pilot-at-example-dot-com"`. The store-level test covers this, but an `is_source_ingested` assertion here would close the end-to-end loop. Low priority since the gate behaviour is already verified at the unit level.

- `src/openhuman/memory_store/chunks/store.rs:839` — N+1 COUNT queries (one per deleted source_id). For a once-per-disconnect cleanup with typically O(1–10) distinct source_ids this is fine, but a single `SELECT source_id, COUNT(*) … GROUP BY source_id` would eliminate the loop.

- **Upgrade path for legacy chunks**: After this change, chunks stored under the *old* account-scoped source_id format (`gmail:{slug_account_email(email)}`) with a non-`gmail-sync:*` owner (e.g. `alice@example.com`) will be silently skipped during a disconnect. The PR description calls this out as intentional conservatism. For users who have such chunks: are those chunks expected to be cleaned up eventually (e.g. by a future migration or re-ingest), or will they accumulate indefinitely after all Gmail connections are disconnected?

## Verified / looks good

- Core logic is correct: `DELETE` + `COUNT(*)` run in the same `unchecked_transaction`, so the orphan check sees the post-deletion state consistently. ✓
- `delete_chunks_by_source` and `delete_chunks_by_source_prefix` behaviour is preserved: both pass the same predicate for both closures, and the orphan-check now provides redundant-but-harmless double coverage for exact/prefix deletes. ✓
- `memory_chunks_deleted` count is accurate: after the `Owner` target deletes chunks, the subsequent `Exact`/`Prefix` passes find nothing and return 0 — no double-counting. ✓
- `store_tests::delete_chunks_by_owner_preserves_other_owners_for_same_source` correctly exercises: shared-source gate preservation, sole-source gate removal, and correct chunk-level isolation. ✓
- `config/ops_tests.rs` compile fix (`..Default::default()`) is correct and minimal. ✓
- No CEF JS injection, no dynamic imports, no `import.meta.env` — Rust-only change. ✓

src/openhuman/memory_store/chunks/store.rs
```diff
@@ -806,6 +836,20 @@ fn delete_chunks_by_source_filter(
...
+                "SELECT COUNT(*)
+                   FROM mem_tree_chunks
+                  WHERE source_kind = ?1 AND source_id = ?2",
+                params![source_kind.as_str(), source_id],
+                |row| row.get(0),
+            )?;
+            if remaining == 0 {
+                orphaned_deleted_sources.insert(source_id.clone());
+            }
+        }
```
M3gA-Mind, line 864: **Missing diagnostic logging on the orphan check (required by CLAUDE.md)**

The new orphaned-sources loop is a meaningful state transition — when it preserves an ingest gate because another owner still has chunks, there's currently no log line showing that decision was reached. CLAUDE.md explicitly labels changes without diagnosis logging as incomplete.

```rust
// suggested addition after building orphaned_deleted_sources
for source_id in &deleted_source_ids {
    let remaining: i64 = tx.query_row(/* ... */)?;
    if remaining == 0 {
        log::debug!(
            "[memory::chunk_store] owner-delete: source={source_id} orphaned — removing ingest gate"
        );
        orphaned_deleted_sources.insert(source_id.clone());
    } else {
        log::debug!(
            "[memory::chunk_store] owner-delete: source={source_id} has {remaining} chunk(s) from other owners — gate preserved"
        );
    }
}
```

src/openhuman/composio/ops.rs
```diff
@@ -581,30 +587,12 @@ async fn composio_memory_targets_for_connection(
...
-        }
-    }
-    sources
+    dedupe_memory_targets(vec![
+        MemoryCleanupTarget::Owner(SourceKind::Email, format!("gmail-sync:{connection_id}")),
+        MemoryCleanupTarget::Exact(SourceKind::Email, format!("gmail:{connection_id}")),
+        MemoryCleanupTarget::Prefix(SourceKind::Email, format!("gmail:{connection_id}:")),
+        MemoryCleanupTarget::Prefix(SourceKind::Email, format!("gmail:{connection_id}/")),
+    ])
 }
```
M3gA-Mind, line 619: **Nitpick: `dedupe_memory_targets` is a no-op here**

All four targets are structurally distinct, so the dedup loop finds nothing to remove. The call is harmless (consistent with other callers) but could confuse a reader into wondering what overlap is possible. Either drop it or add a comment explaining it's defensive.

```rust
// before
dedupe_memory_targets(vec![
    MemoryCleanupTarget::Owner(SourceKind::Email, format!("gmail-sync:{connection_id}")),
    // ...
])

// after (option A — just use the vec directly since all variants are distinct by construction)
vec![
    MemoryCleanupTarget::Owner(SourceKind::Email, format!("gmail-sync:{connection_id}")),
    MemoryCleanupTarget::Exact(SourceKind::Email, format!("gmail:{connection_id}")),
    MemoryCleanupTarget::Prefix(SourceKind::Email, format!("gmail:{connection_id}:")),
    MemoryCleanupTarget::Prefix(SourceKind::Email, format!("gmail:{connection_id}/")),
]
```

src/openhuman/memory_store/chunks/store.rs
Felyx-Fu: Fixed in ef88f69f. I added debug logs for both outcomes of the owner cleanup orphan check: removing the ingest gate when no chunks remain, and preserving it when other owners still have chunks. The source id is logged through the existing redaction helper instead of as a raw id.

src/openhuman/composio/ops.rs
Felyx-Fu: Fixed in ef88f69f. I dropped the no-op dedupe_memory_targets wrapper here and left the Gmail cleanup targets as a direct vec, since the Owner/Exact/Prefix variants are distinct by construction.

Felyx-Fu:
Addressed CodeRabbit follow-up in e2c225666d978dea290ee884e348ccbcf7469825: added a focused Rust regression test, list_memory_files_skips_internal_sqlite_artifacts. It seeds memory.db, memory.db-shm, memory.db-wal, and visible.md, then asserts ai_list_memory_files returns only visible.md. Validation: cargo fmt --manifest-path Cargo.toml -- --check; git diff --check; cargo test --manifest-path Cargo.toml --lib list_memory_files_skips_internal_sqlite_artifacts.

Felyx-Fu:
Addressed CodeRabbit follow-up in 7510e5d8f7585949b67c9635307bb1fc75af4d55: added a short inline comment explaining that memory.db, memory.db-shm, and memory.db-wal are SQLite engine files hidden from user-facing memory file listings. Validation: cargo fmt --manifest-path Cargo.toml -- --check; git diff --check; cargo test --manifest-path Cargo.toml --lib list_memory_files_skips_internal_sqlite_artifacts.

graycyrus (dismissed):
Solid bug fix. The owner-scoped deletion logic is correct — the `DELETE` + `COUNT(*)` orphan check runs inside a single transaction so the gate-preservation decision sees consistent post-deletion state. The `delete_chunks_by_source` / `delete_chunks_by_source_prefix` callers are backwards-compatible since the new orphan check is redundant-but-harmless when the chunk predicate and ingest-gate predicate match the same source pattern.

Good call on the conservative approach for legacy email-slug chunks — silently skipping ambiguous chunks is the right tradeoff vs. risking the cross-connection deletion bug this PR fixes.

**One note for future cleanup**: `delete_chunks_by_owner` routes through `delete_chunks_by_source_filter` which does a full `WHERE source_kind = ?1` scan and filters in Rust. For a once-per-disconnect operation this is fine, but if chunk volume grows, a targeted `WHERE source_kind = ?1 AND owner = ?2` query would let SQLite use `idx_mem_tree_chunks_owner` directly.

| File | What changed |
| --- | --- |
| `composio/ops.rs` | `Owner` variant on `MemoryCleanupTarget`; `gmail_memory_sources_for_connection` rewritten to owner-scoped + connection-id fallbacks |
| `memory_store/chunks/store.rs` | `delete_chunks_by_owner` + refactored filter with separate chunk/gate predicates + orphan-aware gate preservation |
| `memory/ops/files.rs` | Filter SQLite artifacts from memory file listing |
| `composio/ops_test.rs` | Multi-connection Gmail cleanup regression test |
| `chunks/store_tests.rs` | Owner-scoped deletion with gate preservation test |
| `config/ops_tests.rs`, `factory_test.rs`, `chat.rs`, `loopbackOauthListener.test.ts` | CI-drift fixes (compile fix, summarization hint, constant, timeout) |

src/openhuman/memory_store/chunks/store.rs
```diff
@@ -806,6 +836,29 @@ fn delete_chunks_by_source_filter(
...
+            let remaining: i64 = tx.query_row(
+                "SELECT COUNT(*)
+                   FROM mem_tree_chunks
+                  WHERE source_kind = ?1 AND source_id = ?2",
+                params![source_kind.as_str(), source_id],
+                |row| row.get(0),
+            )?;
+            if remaining == 0 {
+                log::debug!(
+                    "[memory::chunk_store] delete_chunks_by_owner: source_id_hash={} orphaned; removing ingest gate",
```
sanil-23: **Refactor: log prefix hardcodes `delete_chunks_by_owner` in a shared helper**

These two debug lines live inside `delete_chunks_by_source_filter`, which now backs **three** entry points: `delete_chunks_by_source`, `delete_chunks_by_source_prefix`, and `delete_chunks_by_owner`. The `orphaned; removing ingest gate` line fires for all three callers, so a gate removed by an *exact source* deletion is logged as `delete_chunks_by_owner:` — misattributing the call path during incident triage. Per `CLAUDE.md`'s grep-friendly-diagnostics rule, keep the prefix call-path-neutral:

```rust
// helper is shared by by_source / by_prefix / by_owner; keep the label neutral
log::debug!(
    "[memory::chunk_store] delete_chunks_by_source_filter: source_id_hash={} orphaned; removing ingest gate",
    redact_value(source_id),
);
// ...and the matching "preserving ingest gate" line below
```

Alternatively, thread an `op: &str` argument through `delete_chunks_by_source_filter` and have each wrapper pass its own name for precise attribution.

sanil-23 (dismissed):
# PR #2698 — fix(memory): scope Gmail cleanup to disconnected connection

## Walkthrough
This PR fixes a real cross-connection data-loss bug: `composio_delete_connection(clear_memory=true)` for one Gmail connection previously deleted memory chunks by the account-shaped source id (`gmail:{slug(email)}`), which is **shared** by every Gmail connection for the same account — so disconnecting connection A wiped connection B's chunks. The fix introduces an owner-scoped deletion path (`delete_chunks_by_owner`) keyed on `metadata.owner = "gmail-sync:{connection_id}"` (the value Gmail ingest actually writes — verified at `provider.rs:468`), and preserves the per-source ingest gate (`mem_tree_ingested_sources`) whenever chunks owned by another connection remain.

I traced the design end-to-end and it's **correct and well-tested**: the owner format matches the ingest path for both the account-shaped and per-participant source ids, the connection-id `Exact`/`Prefix` fallbacks can't collide across connections (the `:`/`/` separators guard prefix overlap), the orphan-gate logic only drops gates when the deleted owner was the last owner, and the exact/prefix callers retain their previous behavior. All CI is green (clippy, coverage gate, full test matrix). The main concern is **scope**: four unrelated CI-drift / small-bugfix changes are bundled with the core fix.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/composio/ops.rs` | Replaces identity-lookup Gmail cleanup with a connection-scoped `Owner` target + `Exact`/`Prefix` `gmail:{id}` fallbacks; adds `MemoryCleanupTarget::Owner` variant. |
| `src/openhuman/memory_store/chunks/store.rs` | Adds `delete_chunks_by_owner`; generalizes `delete_chunks_by_source_filter` to a `(source_id, owner)` chunk predicate + a separate ingest-gate predicate, with orphan-gate preservation. |
| `src/openhuman/composio/ops_test.rs` | Adds multi-connection Gmail retention test + owner-parameterized chunk helper. |
| `src/openhuman/memory_store/chunks/store_tests.rs` | Adds owner-scoped deletion + gate-preservation regression test. |
| `src/openhuman/memory/ops/files.rs` | Hides `memory.db`/`-shm`/`-wal` from `ai_list_memory_files`; adds a test. *(unrelated to PR title)* |
| `src/openhuman/inference/provider/factory_test.rs` | Adds `summarization-v1` tier/hint test coverage + comment fixes. *(CI-drift)* |
| `src/openhuman/memory/chat.rs` | Test now asserts against `DEFAULT_CLOUD_LLM_MODEL` constant instead of the `summarization-v1` literal. *(CI-drift, no behavior change)* |
| `.github/workflows/test-reusable.yml` | Bumps the Windows secrets-ACL Rust job timeout 20→30 min. *(CI infra)* |

## Actionable comments (1)

### 💡 Refactor / suggestion

#### 1. `src/openhuman/memory_store/chunks/store.rs:849-858` — Log prefix hardcodes `delete_chunks_by_owner` in a shared helper
These two debug lines live inside `delete_chunks_by_source_filter`, which now backs **three** public entry points: `delete_chunks_by_source`, `delete_chunks_by_source_prefix`, and `delete_chunks_by_owner`. The `orphaned; removing ingest gate` line fires for all three callers, so during incident triage a gate removed by an *exact source* deletion will be logged as `delete_chunks_by_owner:` — misattributing the call path. Given `CLAUDE.md`'s emphasis on grep-friendly, accurate diagnostics, the prefix should be call-path-neutral (or threaded through as a parameter). *(Inline comment posted.)*

```rust
// after — helper is shared by by_source / by_prefix / by_owner; keep the label neutral
log::debug!(
    "[memory::chunk_store] delete_chunks_by_source_filter: source_id_hash={} orphaned; removing ingest gate",
    redact_value(source_id),
);
```

## Nitpicks (1)
- `src/openhuman/composio/ops.rs:581-583` — `MemoryCleanupTarget::Owner` label uses a space (`"{kind} owner:{owner}"`) while `Exact`/`Prefix` use `"{kind}:{id}"`. Harmless (label only feeds an error string), but `"{kind}:owner:{owner}"` would read more consistently in `memory_clear_errors`.

- **PR scope** — `factory_test.rs`, `memory/chat.rs`, `memory/ops/files.rs`, and `.github/workflows/test-reusable.yml` are unrelated to "scope Gmail cleanup to disconnected connection". Each is independently sound (the chat.rs change is a literal→constant no-op since `DEFAULT_CLOUD_LLM_MODEL == "summarization-v1"`; the factory tests match `factory.rs:421`/`:91`; the SQLite-artifact hide is correct since `memory.db` runs WAL per `unified/init.rs:55`, so `-shm`/`-wal` is the complete side-file set). They're documented in the PR body as CI-drift / CodeRabbit follow-ups — fine to keep them here to keep the branch green, but worth confirming they were intentional rather than accidental merge carry-over, since they widen the blast radius of a memory-cleanup bugfix.

## Verified / looks good
- **Owner format matches ingest**: `Owner(Email, "gmail-sync:{connection_id}")` matches the exact owner written by Gmail ingest for both the account-shaped (`gmail:{slug(email)}`) and per-participant source ids (`provider.rs:468`, `ingest.rs:244-268`). The fix is not inert.
- **No cross-connection prefix collision**: `Exact("gmail:{id}")` + `Prefix("gmail:{id}:" / "gmail:{id}/")` can't match another connection's id because of the `:`/`/` separators.
- **Gate preservation is correct**: `orphaned_deleted_sources` is computed after deletion via a `COUNT(*)` per touched source; a gate is dropped only when zero chunks remain. Exact/prefix callers preserve prior behavior (their orphan set == matched set == `matches_ingested_source` set).
- **Conservative legacy handling**: pre-`gmail-sync:` chunks (account-shaped source, non-matching owner) are intentionally left untouched rather than ambiguously deleted — documented trade-off in the PR body.
- **Tests**: `composio_delete_connection_clear_memory_keeps_other_gmail_connections` and `delete_chunks_by_owner_preserves_other_owners_for_same_source` exercise both the retain-other-owner and orphan-gate paths; `list_memory_files_skips_internal_sqlite_artifacts` covers the hide path.
- No PII leak: source ids/owners are passed through `redact()` in the new debug logs.

src/openhuman/memory_store/chunks/store.rs
Felyx-Fu: Addressed in b552361a737ee2fb227da297beea87a501feb624 by threading an `op` label through `delete_chunks_by_source_filter`; exact, prefix, and owner callers now log their own operation names instead of hardcoding `delete_chunks_by_owner`. Local validation: `cargo fmt --manifest-path Cargo.toml -- --check`; `cargo test --manifest-path Cargo.toml delete_chunks_by_owner_preserves_other_owners_for_same_source --lib`.

Felyx-Fu:
Confirming the PR-scope note from sanil-23's review: the non-Gmail-cleanup files were intentional follow-ups kept on this branch to keep CI and prior CodeRabbit feedback green. They are documented in the PR body as CI-drift / small follow-up fixes rather than part of the Gmail ownership behavior itself.

graycyrus (dismissed):
## Continuation review — addressing previous feedback

The `op` parameter threaded through `delete_chunks_by_source_filter` resolves the misattributed-log-prefix concern from my last pass. Debug lines now correctly identify the call path (`delete_chunks_by_source`, `delete_chunks_by_source_prefix`, `delete_chunks_by_owner`) and source ids are redacted before logging. That was the one actionable item holding this back.

Core logic is sound. Verified end-to-end:

- `DELETE` + `COUNT(*)` orphan check run in the same transaction — the gate-preservation decision sees consistent post-deletion state. No TOCTOU.
- Owner format `gmail-sync:{connection_id}` matches what Gmail ingest writes. The fix is not inert.
- `Exact("gmail:{id}")` / `Prefix("gmail:{id}:"` / `Prefix("gmail:{id}/")` fallbacks can't collide across connections — the `:` and `/` separators prevent prefix overlap.
- Exact/prefix callers retain prior behavior: their `matches_ingested_source` predicate matches the same set as the chunk predicate, so the new orphan check is redundant-but-harmless for them.
- Legacy account-slug chunks (non-`gmail-sync:` owner) are intentionally preserved — the conservative call is right given the ambiguity.

Two remaining minor items, neither blocking:

- `MemoryCleanupTarget::Owner` display label uses a space separator (`"{kind} owner:{owner}"`) while `Exact`/`Prefix` use colon (`"{kind}:{id}"`). Only affects error strings, but `"{kind}:owner:{owner}"` would be consistent if you want to clean it up.
- `delete_chunks_by_owner` does N COUNT queries (one per deleted source_id). Fine for once-per-disconnect; if the table grows, a single `SELECT source_id, COUNT(*) … GROUP BY source_id` would eliminate the loop.

Tests are solid — multi-connection Gmail retention, owner-scoped gate preservation, and SQLite artifact hiding all covered. CI is fully green. Approved.

graycyrus:
@Felyx-Fu this PR has merge conflicts with main — please rebase/resolve before review.

graycyrus:
This PR has merge conflicts with main — please rebase/resolve before review.

Felyx-Fu:
@graycyrus resolved the merge-conflict feedback by bringing `main` into this branch in 38ce7dd7d5dbab207e98afe962e2cf399ccf0ba6. I rechecked the branch against `main`: `behind_by=0`, GitHub reports `mergeable=true`, and the current PR checks are 21 passing / 4 skipped / 0 pending / 0 failing.

sanil-23 (approved):
Approving from shepherd queue (round 2). All CI green or known flakes re-running; `pnpm review fix` round complete. Happy to merge once required reviewers concur.

sanil-23:
Hey @graycyrus — friendly bump on this one. Felyx-Fu's Composio Gmail scope fix is **CI fully green (25 success / 0 fail)** and addresses the scoping issue cleanly. Your last review here was dismissed on May 27 (pre-rfix); the current head is c9f62cc4 with a docs comment refresh on top. If you have a moment to take another look and either approve or call out anything I missed, that would unblock it.

graycyrus (approved):
solid fix. the root cause is clear and the solution is correct — scoping Gmail cleanup to `owner` rather than account-shaped source id is the right move, and the ingest gate preservation logic is the tricky part done well.

the `delete_chunks_by_source_filter` refactoring is the highest-risk change here: it now runs N additional `SELECT COUNT(*)` queries per distinct source_id inside the transaction to decide whether a gate is orphaned. the logic is correct — those counts see the post-deletion state within the same transaction, so the orphan determination is accurate. for cleanup paths this is called once per disconnect, so the extra queries are inconsequential in practice.

the owner-based gate logic is right: `matches_ingested_source = |_| false` for owner deletions means gates are only removed via the orphan check, not preemptively. that's exactly what you want — if another connection still has chunks under the same source_id, the gate survives.

test coverage is good: multi-connection Gmail retention, Slack cleanup still working, owner-scoped gate preservation, and the SQLite artifact filtering test. the `factory_test.rs` additions are clean drift fixes.

one thing worth keeping an eye on downstream: `owner` is now fetched as `String` (non-optional) from the DB. if any historical chunk row has a NULL owner, `delete_chunks_by_source` would error where it previously succeeded. the Rust type being `String` throughout suggests the schema enforces NOT NULL, but worth verifying if anyone reports cleanup regressions on old installs.

approved.

Issue #1402, opened 2026-05-09
Make the Tauri frontend handle workspace file links and views properly


senamakel:
## Summary

Make the Tauri frontend render, hyperlink, and open workspace-backed files and views properly, so local workspace content behaves like a first-class desktop surface instead of a mix of plain text, Obsidian-only links, and browser-only anchors.

## Problem

As a follow-on to the sandbox/workspace isolation work, the desktop UI also needs a clean contract for *interacting* with the workspace.

Today the frontend has a few useful pieces, but they do not add up to a general workspace-file UX:
- `openUrl()` can hand URLs to the OS through `tauri-plugin-opener`, which is good for external links and custom schemes (`app/src/utils/openUrl.ts`).
- the conversations markdown renderer explicitly only allows `http:`, `https:`, and `mailto:` links (`app/src/pages/conversations/utils/format.ts:53-61`).
- `AgentMessageBubble` routes markdown anchors through that allowlist, so `file:`, workspace-relative, and other local-link patterns are not first-class clickable targets (`app/src/pages/conversations/components/AgentMessageBubble.tsx:57-75`).
- the Memory tab has bespoke Obsidian launch helpers (`obsidian://open?...`) for the vault and summary nodes, but those are one-off flows rather than a general workspace-link system (`app/src/components/intelligence/MemoryWorkspace.tsx:58-79`, `app/src/components/intelligence/MemoryGraph.tsx:132-156`).
- there is a Tauri command to reveal the logs folder, which shows the shell can already open a local directory in Finder / Explorer / xdg-open, but there is no equivalent generic workspace-file or workspace-folder open/reveal surface (`app/src-tauri/src/file_logging.rs:121-133`).

This leaves several UX gaps:
- if the agent references a workspace file in markdown, the user cannot reliably click it and open it in the right local handler
- if a surface is “hosting” workspace content in the webview, it is unclear when links should stay in-app vs. open in the OS vs. reveal in the file manager
- the desktop app has no single contract for opening a workspace file, revealing its parent folder, or previewing safe text/markdown content inside the app
- link behavior is fragmented across chat, Memory, and Tauri-native helper flows

Examples of current fragmentation:
- chat markdown treats local/workspace links as invalid because the external-href allowlist rejects them
- Memory graph can open summaries in Obsidian, but not through a more general “open this workspace file” route
- logs folder can be revealed via a dedicated Tauri command, but this pattern is not generalized for arbitrary workspace paths

[...]

PR #2702, opened 2026-05-26
refactor(memory): route Obsidian deep link through workspace-link layer (#2492)

### LINKED ISSUE #2492: Migrate Memory surfaces to shared workspace file links
## Summary

Migrate Memory tab file affordances onto the shared workspace-link layer introduced by #2476, so memory files use the same safe open / reveal / preview contract as chat workspace links.

## Problem

Issue #1402 still has one explicit acceptance criterion unmet after #2476: **Memory surfaces use the shared layer**. The current Memory flows in `MemoryWorkspace.tsx` and `MemoryGraph.tsx` still use bespoke Obsidian URL helpers for workspace-backed files, while #2476 covers the Tauri commands, frontend parsing, and chat markdown integration.

Without a dedicated follow-up, merging #2476 with `Closes #1402` would hide the remaining Phase 3 work and leave local workspace-link behavior fragmented between chat and Memory.

## Solution (optional)

- Route Memory workspace file actions through the shared workspace-link / workspace-path helpers from #2476.
- Preserve Obsidian as an explicit action where appropriate, but avoid duplicating path normalization and local file handling logic.
- Add or reuse a small shared UI affordance for open / reveal / preview where Memory surfaces need more than one action.
- Cover `MemoryWorkspace.tsx` and `MemoryGraph.tsx` behavior with focused tests.

## Acceptance criteria

- [ ] **MemoryWorkspace uses shared workspace links** — memory workspace file actions are routed through the shared workspace-link/path layer instead of bespoke local path handling.
- [ ] **MemoryGraph uses shared workspace links** — summary/file node actions use the same contract as chat workspace links.
- [ ] **Obsidian behavior is preserved intentionally** — Obsidian launch remains available where it is a user-facing feature, but it is composed with the shared workspace-link layer rather than replacing it.
- [ ] **Open / reveal / preview behavior is tested** — focused frontend tests cover Memory link routing and failure states.

## Related

- Follow-up to #1402 Phase 3.
- #2476 implements the workspace path commands, parsing layer, and chat markdown integration this issue should reuse.
- `app/src/components/intelligence/MemoryWorkspace.tsx`
- `app/src/components/intelligence/MemoryGraph.tsx`

YOMXXX:
## Summary

- New Tauri command `resolve_workspace_absolute_path` exposes the workspace-link resolver introduced in #2476 so renderers can derive canonical absolute paths through the same guard as `open_workspace_path` / `reveal_workspace_path` / `preview_workspace_text`.
- New TS wrapper `resolveWorkspaceAbsolutePath` in `utils/tauriCommands/workspacePaths.ts`.
- `ObsidianVaultSection` now resolves the vault's absolute path through that wrapper before building the `obsidian://open?path=<abs>` URL, replacing the last Memory-tab surface that was constructing the deep link straight from a graph-export RPC field.
- User-visible behavior is unchanged — click still opens Obsidian to the vault, same toast copy — but the absolute path is now derived from the Rust-side resolver, not from a prop fed by RPC.

## Problem

Issue #2492 (Phase 3 cleanup of #1402) requires every Memory surface that constructs an on-disk path to compose with the shared workspace-link layer rather than hand-rolling its own resolution.

`MemoryGraph` was migrated in PR #2638. The remaining offender was `ObsidianVaultSection`, which built `obsidian://open?path=${encodeURIComponent(contentRootAbs)}` directly from the `content_root_abs` field returned by `memory_tree_obsidian_vault_status` — no canonicalize, no workspace-containment recheck, no shared error surface. Acceptance criterion from the issue:

> Obsidian behavior is preserved intentionally — Obsidian launch remains available where it is a user-facing feature, but it is composed with the shared workspace-link layer rather than replacing it.

## Solution

Three layers, kept thin:

1. **Rust (`app/src-tauri/src/workspace_paths.rs`)**: new `resolve_workspace_absolute_path` `#[tauri::command]` that wraps the existing internal `resolve_workspace_path` helper (same canonicalize + workspace-containment + non-leaky error surface as the other three commands). Registered in `lib.rs` and added to the `allow-workspace-files` permission manifest. Two new unit tests cover the happy-path (`memory_tree/content` resolves inside the workspace root) and the empty-input rejection; the existing resolver test suite already covers absolute-path / `..` / URI-scheme / symlink escapes.
2. **TypeScript (`app/src/utils/tauriCommands/workspacePaths.ts`)**: `resolveWorkspaceAbsolutePath(path)` wrapper mirroring the style of the other workspace wrappers (assertTauri guard + plain `invoke<string>`).
3. **Component (`app/src/components/intelligence/ObsidianVaultSection.tsx`)**: `fireDeepLink` now `await`s `resolveWorkspaceAbsolutePath('memory_tree/content')` and builds the URL from the returned absolute path. The `contentRootAbs` prop is preserved for display (button tooltip, guidance-panel code block, toast message text) so no cascading prop changes are needed — only the deep-link URL is rerouted.

### Tests

Updated `ObsidianVaultSection.test.tsx`:
- Mock the new wrapper in the existing `workspacePaths` mock block.
- New test: deep link URL is composed from the resolver output (not from `contentRootAbs`).
- New test: resolver rejection surfaces an error toast and keeps the guidance panel expanded.
- All 7 existing tests still pass (9 total).

Updated `MemoryWorkspace.test.tsx`:
- Mock `resolveWorkspaceAbsolutePath` to return the same path the assertion expects so the existing `'View vault in Obsidian' triggers the deep link` test stays stable. All 20 tests pass.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — every new line is exercised by the two new Vitest cases plus the existing happy-path test (now stricter via the resolver-call assertion) and the two new Rust unit tests. Resolver internals already covered by the prior `resolve_workspace_path_*` test suite.
- [x] N/A: Coverage matrix update — this is a behaviour-preserving refactor of an existing matrix entry, not an added/removed/renamed feature row.
- [x] N/A: No new feature IDs from the matrix affected.
- [x] N/A: Manual smoke checklist — no release-cut surface touched; deep-link UX is unchanged from the user's point of view.
- [x] Linked issue closed via `Closes #2492` in the `## Related` section.

## Impact

- **Runtime/platform**: desktop only (`#[tauri::command]` lives in the Tauri shell; iOS client does not hit this code path). No change to JSON-RPC surface.
- **Security**: net positive — the Obsidian deep link now goes through the same workspace-containment guard as the other workspace commands, so a malicious `content_root_abs` value from the RPC (or a future scenario where the field is user-influenced) can no longer slip out of the workspace.
- **Migration / compatibility**: none; the URL shape is identical.

## Related

- Closes #2492
- Builds on #2476 (workspace-link layer) and #2638 (MemoryGraph migration)
- Follow-up PR(s)/TODOs: none — this is the final Memory-tab surface called out in #2492.

---

### Linear Issue
- Key: N/A (GitHub issue only)
- URL: https://github.com/tinyhumansai/openhuman/issues/2492

### Commit & Branch
- Branch: `fix/2492-memory-workspace-link-migration`
- Commit SHA: `3682a02d` (tip), `910d8097`, `8ff810ba`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — equivalent: `pnpm exec prettier --check` on changed files, clean.
- [x] `pnpm typecheck` — ran `pnpm exec tsc --noEmit` inside `app/`, clean.
- [x] Focused tests: `pnpm debug unit ObsidianVaultSection` → 9 passed; `pnpm debug unit MemoryWorkspace` → 20 passed.
- [x] Rust fmt/check (if changed): `cargo fmt --manifest-path app/src-tauri/Cargo.toml --check` clean; `GGML_NATIVE=OFF cargo check --manifest-path app/src-tauri/Cargo.toml --lib` clean (`GGML_NATIVE=OFF` is the documented macOS Tahoe + whisper-rs workaround).
- [x] Tauri fmt/check (if changed): same as above.

### Validation Blocked
- `command:` n/a — all validation passed.
- `error:` n/a
- `impact:` n/a

### Behavior Changes
- Intended behavior change: deep-link absolute path is now derived from the workspace-link resolver instead of trusted RPC output.
- User-visible effect: none. Same click, same Obsidian URL shape, same toast copy. The vault path printed in the guidance panel / toast still comes from `contentRootAbs` (RPC) because it is a display concern.

### Parity Contract
- Legacy behavior preserved: the `obsidian://open?path=<abs>` URL shape is byte-for-byte unchanged when the resolver returns the same path the RPC reports (normal case). `MemoryWorkspace.test.tsx`'s existing assertion proves it.
- Guard/fallback/dispatch parity checks: resolver failure now surfaces an error toast and keeps the guidance panel expanded — covered by a new test. Existing `detection failure degrades gracefully to the guidance panel` test still passes.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: n/a

## Summary by CodeRabbit

* **New Features**
  * Added a workspace-path resolver that returns canonical absolute paths for constructing vault deep links.

* **Bug Fixes**
  * Deep-link generation now uses resolved canonical workspace paths; on resolution failure it surfaces an error and does not attempt a fallback.

* **Tests**
  * Added/updated tests for path resolution success, failure, canonicalization, and deep-link behavior.

* **Chores**
  * Updated runtime permissions to allow workspace-path resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2702?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src-tauri/src/workspace_paths.rs
+++ b/app/src-tauri/src/workspace_paths.rs
@@ -424,7 +424,7 @@ mod tests {
     }
 
     #[test]
-    fn resolve_workspace_path_returns_absolute_inside_workspace() {
+    fn resolve_workspace_path_resolves_memory_tree_content_inside_workspace() {
         let workspace = tempdir().unwrap();
         let docs = workspace.path().join("memory_tree").join("content");
         fs::create_dir_all(&docs).unwrap();
@@ -442,7 +442,7 @@ mod tests {
     }
 
     #[test]
-    fn resolve_workspace_path_rejects_empty_input() {
+    fn resolve_workspace_path_rejects_empty_whitespace_input() {
         let workspace = tempdir().unwrap();
 
         let err = resolve_workspace_path(workspace.path(), "   ").unwrap_err();
--- a/app/src-tauri/src/workspace_paths.rs
+++ b/app/src-tauri/src/workspace_paths.rs
@@ -235,7 +235,7 @@ pub(crate) fn resolve_workspace_path(
     log::debug!(
         "[workspace-paths] resolved workspace path: {} -> {}",
         normalized_path,
-        target.display()
+        workspace_path_label(workspace_root, &target)
     );
     Ok(target)
 }
--- a/app/src-tauri/src/workspace_paths.rs
+++ b/app/src-tauri/src/workspace_paths.rs
@@ -51,6 +51,29 @@ pub async fn preview_workspace_text(path: String) -> Result<WorkspaceTextPreview
     preview_workspace_text_from_root(&workspace, &path, DEFAULT_PREVIEW_MAX_BYTES)
 }
 
+/// Resolve a workspace-relative path to its canonical absolute path on disk,
+/// after validating it stays inside the active OpenHuman workspace.
+///
+/// This exposes the internal [`resolve_workspace_path`] helper so UI flows that
+/// need an absolute path to compose with a platform-specific URL scheme (e.g.
+/// `obsidian://open?path=<abs>`) can route through the shared workspace-link
+/// layer instead of re-implementing path normalization in the renderer.
+///
+/// Errors mirror the other workspace-path commands — empty input, parent-dir
+/// escape, NUL bytes, URI-scheme prefixes, paths outside the workspace, and
+/// missing files all surface a non-leaky message.
+#[tauri::command]
+pub async fn resolve_workspace_absolute_path(path: String) -> Result<String, String> {
+    let workspace = active_workspace_root().await?;
+    let target = resolve_workspace_path(&workspace, &path)?;
+    log::debug!(
+        "[workspace-paths] resolve_workspace_absolute_path: {} -> {}",
+        path,
+        target.display()
+    );
+    Ok(target.to_string_lossy().into_owned())
+}
+
 async fn active_workspace_root() -> Result<PathBuf, String> {
     let config = openhuman_core::openhuman::config::Config::load_or_init()
         .await
@@ -398,6 +421,33 @@ mod tests {
             !err.contains(&workspace.path().display().to_string()),
             "error leaked workspace root: {err}"
         );
+    }
+
+    #[test]
+    fn resolve_workspace_path_returns_absolute_inside_workspace() {
+        let workspace = tempdir().unwrap();
+        let docs = workspace.path().join("memory_tree").join("content");
+        fs::create_dir_all(&docs).unwrap();
+
+        let resolved = resolve_workspace_path(workspace.path(), "memory_tree/content").unwrap();
+
+        let canonical_root = fs::canonicalize(workspace.path()).unwrap();
+        assert!(
+            resolved.starts_with(&canonical_root),
+            "resolved path escaped workspace root: {} not under {}",
+            resolved.display(),
+            canonical_root.display()
+        );
+        assert_eq!(resolved, docs.canonicalize().unwrap());
+    }
+
+    #[test]
+    fn resolve_workspace_path_rejects_empty_input() {
+        let workspace = tempdir().unwrap();
+
+        let err = resolve_workspace_path(workspace.path(), "   ").unwrap_err();
+
+        assert!(err.contains("empty"), "unexpected error: {err}");
     }
 
     #[cfg(unix)]
--- a/app/src-tauri/src/workspace_paths.rs
+++ b/app/src-tauri/src/workspace_paths.rs
@@ -66,10 +66,10 @@ pub async fn preview_workspace_text(path: String) -> Result<WorkspaceTextPreview
 pub async fn resolve_workspace_absolute_path(path: String) -> Result<String, String> {
     let workspace = active_workspace_root().await?;
     let target = resolve_workspace_path(&workspace, &path)?;
+    let workspace_label = workspace_path_label(&workspace, &target);
     log::debug!(
-        "[workspace-paths] resolve_workspace_absolute_path: {} -> {}",
-        path,
-        target.display()
+        "[workspace-paths] resolve_workspace_absolute_path: {}",
+        workspace_label
     );
     Ok(target.to_string_lossy().into_owned())
 }
--- a/app/src/components/intelligence/ObsidianVaultSection.tsx
+++ b/app/src/components/intelligence/ObsidianVaultSection.tsx
@@ -25,7 +25,10 @@ import { useT } from '../../lib/i18n/I18nContext';
 import type { ToastNotification } from '../../types/intelligence';
 import { openUrl } from '../../utils/openUrl';
 import { memoryTreeObsidianVaultStatus } from '../../utils/tauriCommands';
-import { revealWorkspacePath } from '../../utils/tauriCommands/workspacePaths';
+import {
+  resolveWorkspaceAbsolutePath,
+  revealWorkspacePath,
+} from '../../utils/tauriCommands/workspacePaths';
 import { MEMORY_CONTENT_WORKSPACE_PATH } from './memoryWorkspacePaths';
 
 /** localStorage key for the optional Obsidian config-dir override. */
@@ -55,18 +58,29 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
   const [showAdvanced, setShowAdvanced] = useState(false);
   const [configDir, setConfigDir] = useState<string>(readConfigDirOverride);
 
-  /** Build + fire the `obsidian://` deep link. Resolves to an error or null. */
+  /**
+   * Build + fire the `obsidian://` deep link.
+   *
+   * The absolute path is resolved through the shared workspace-link layer
+   * (`resolve_workspace_absolute_path` Tauri command) rather than reused from
+   * the `contentRootAbs` prop — this routes Obsidian deep links through the
+   * same canonicalize + workspace-containment guard as `open_workspace_path`
+   * and `preview_workspace_text` (see issue #2492 / #2476).
+   *
+   * Resolves to an error or null.
+   */
   const fireDeepLink = useCallback(async (): Promise<unknown | null> => {
-    const url = `obsidian://open?path=${encodeURIComponent(contentRootAbs)}`;
     console.debug('[ui-flow][obsidian-vault] firing deep link');
     try {
+      const absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
+      const url = `obsidian://open?path=${encodeURIComponent(absolutePath)}`;
       await openUrl(url);
       return null;
     } catch (err) {
-      console.error('[ui-flow][obsidian-vault] openUrl failed', err);
+      console.error('[ui-flow][obsidian-vault] resolve/open failed', err);
       return err;
     }
-  }, [contentRootAbs]);
+  }, []);
 
   const reveal = useCallback(() => {
     void (async () => {
--- a/app/src/components/intelligence/ObsidianVaultSection.tsx
+++ b/app/src/components/intelligence/ObsidianVaultSection.tsx
@@ -72,7 +72,16 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
   const fireDeepLink = useCallback(async (): Promise<unknown | null> => {
     console.debug('[ui-flow][obsidian-vault] firing deep link');
     try {
-      const absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
+      let absolutePath: string;
+      try {
+        absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
+      } catch {
+        // The content directory may not exist yet (fresh workspace). Fall
+        // back to the prop from the graph-export RPC — the deep link URL
+        // is still valid; Obsidian will surface its own "vault not found"
+        // error if the folder is truly missing.
+        absolutePath = contentRootAbs;
+      }
       const url = `obsidian://open?path=${encodeURIComponent(absolutePath)}`;
       await openUrl(url);
       return null;
@@ -80,7 +89,7 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
       console.error('[ui-flow][obsidian-vault] resolve/open failed', err);
       return err;
     }
-  }, []);
+  }, [contentRootAbs]);
 
   const reveal = useCallback(() => {
     void (async () => {
--- a/app/src/components/intelligence/ObsidianVaultSection.tsx
+++ b/app/src/components/intelligence/ObsidianVaultSection.tsx
@@ -72,16 +72,7 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
   const fireDeepLink = useCallback(async (): Promise<unknown | null> => {
     console.debug('[ui-flow][obsidian-vault] firing deep link');
     try {
-      let absolutePath: string;
-      try {
-        absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
-      } catch {
-        // The content directory may not exist yet (fresh workspace). Fall
-        // back to the prop from the graph-export RPC — the deep link URL
-        // is still valid; Obsidian will surface its own "vault not found"
-        // error if the folder is truly missing.
-        absolutePath = contentRootAbs;
-      }
+      const absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
       const url = `obsidian://open?path=${encodeURIComponent(absolutePath)}`;
       await openUrl(url);
       return null;
@@ -89,7 +80,7 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
       console.error('[ui-flow][obsidian-vault] resolve/open failed', err);
       return err;
     }
-  }, [contentRootAbs]);
+  }, []);
 
   const reveal = useCallback(() => {
     void (async () => {
--- a/app/src-tauri/permissions/allow-workspace-files.toml
+++ b/app/src-tauri/permissions/allow-workspace-files.toml
@@ -8,6 +8,7 @@ allow = [
     "open_workspace_path",
     "reveal_workspace_path",
     "preview_workspace_text",
+    "resolve_workspace_absolute_path",
 ]
 
 deny = []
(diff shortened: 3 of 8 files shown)


graycyrus (dismissed):
## Review — refactor(memory): route Obsidian deep link through workspace-link layer

Clean, well-scoped refactor. Three thin layers (Rust command → TS wrapper → component plumbing), each mirroring the existing workspace-path pattern exactly. The security posture improves — the Obsidian deep link now goes through the same canonicalize + workspace-containment guard as the other three workspace commands, closing the last gap from #2492.

### What I checked

| Area | Verdict |
|------|------|
| Rust `resolve_workspace_absolute_path` | Delegates to existing `resolve_workspace_path` — same guard surface, same error semantics. `to_string_lossy()` is fine for URL composition. Debug log is appropriate. |
| Permission manifest | `allow-workspace-files.toml` updated, `lib.rs` registration matches. |
| TS wrapper | Mirrors `revealWorkspacePath` / `openWorkspacePath` style. `assertTauri()` guard present. |
| Component change | `fireDeepLink` deps array correctly empty — no reactive values captured. `contentRootAbs` prop preserved for display only. Error path surfaces toast. |
| Tests | 2 new Rust unit tests (happy path + empty input rejection). 2 new Vitest cases (resolver-sourced URL + resolver failure toast). Mock additions in `MemoryWorkspace.test.tsx` keep 20 existing tests stable. |
| Issue alignment | #2492 acceptance criteria met: Obsidian composed with shared layer, behavior preserved, tests cover open/reveal/failure. Combined with #2638 (MemoryGraph), all criteria are satisfied. |

### CI failures — not PR-related

- Frontend: `loopbackOauthListener.test.ts` timeout constant mismatch (unrelated module)
- Rust Core/Windows: `AutonomySettingsPatch` missing fields in `config::ops` (unrelated module)
- Coverage gate skipped due to upstream test failures

None of these touch workspace paths or the Memory tab.

**LGTM.** Ship it.

YOMXXX:
Addressed CodeRabbit feedback in cffa53a5 + a8a11f6f:

- **Coverage Gate**: added Vitest cases for `resolveWorkspaceAbsolutePath` in `app/src/utils/tauriCommands/workspacePaths.test.ts` covering the happy-path invoke call, an `invoke` rejection bubbling back to the caller, and the `isTauri()` guard short-circuiting before `invoke` is touched. That brings the new wrapper lines (workspacePaths.ts:56-58) under test.
- **PII log nit**: routed the `resolve_workspace_absolute_path` debug log through the existing `workspace_path_label(workspace, target)` helper instead of `target.display()`, matching the redaction pattern the other workspace-path commands already use so the canonical absolute workspace root no longer leaks into logs.

Local verification: `cargo fmt --check`, `cargo check --lib` (Tauri shell), `prettier --check`, `pnpm debug unit workspacePaths` (12 passed), `pnpm typecheck` — all green.

app/src-tauri/src/workspace_paths.rs
```diff
@@ -400,6 +423,33 @@ mod tests {
         );
     }
 
+    #[test]
+    fn resolve_workspace_path_returns_absolute_inside_workspace() {
```
sanil-23: Refactor / suggestion — both new tests call resolve_workspace_path(...) directly (the pre-existing internal helper), not the new resolve_workspace_absolute_path command this PR adds. The command own body (config load, resolve, label/log, to_string_lossy) has no direct Rust test — it is only integration-covered via workspacePaths.ts. resolve_workspace_path_returns_absolute_inside_workspace also overlaps the existing resolve_workspace_path_accepts_existing_relative_file_inside_workspace; the genuinely-new coverage is the starts_with(canonical_root) assertion plus the empty-input rejection. The coverage gate passes, so no change is strictly required — but the test names overpromise (they read as if they cover the new command). Consider renaming to reflect that they assert the shared helper the command delegates to, e.g. resolve_workspace_path_resolves_memory_tree_content_inside_workspace.

app/src/components/intelligence/ObsidianVaultSection.tsx
```diff
@@ -55,18 +58,29 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
...
+   * same canonicalize + workspace-containment guard as `open_workspace_path`
+   * and `preview_workspace_text` (see issue #2492 / #2476).
+   *
+   * Resolves to an error or null.
+   */
   const fireDeepLink = useCallback(async (): Promise<unknown | null> => {
-    const url = `obsidian://open?path=${encodeURIComponent(contentRootAbs)}`;
     console.debug('[ui-flow][obsidian-vault] firing deep link');
     try {
+      const absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
```
sanil-23: Question — resolve_workspace_path requires the target to exist (fs::canonicalize errors otherwise). Previously fireDeepLink built the URL straight from the contentRootAbs prop and never touched disk, so it fired even when memory_tree/content was absent (Obsidian / Open anyway being the escape hatch). Now, if the folder does not exist, the resolver rejects and the link never fires (error toast instead). The "user-visible behavior is unchanged" claim holds for the registered-vault case (Obsidian can only register a folder that exists) and this aligns fireDeepLink with reveal (which already required existence). But for an unregistered vault whose content root has not been created yet, behavior does change. Is graph.content_root_abs guaranteed to be backed by an existing directory whenever this component renders? If yes, this is purely an improvement and the claim stands.

sanil-23 (dismissed):
# PR #2702 — refactor(memory): route Obsidian deep link through workspace-link layer (#2492)

## Walkthrough

This PR finishes the Phase-3 cleanup from #2492 by routing the last Memory-tab deep-link surface (`ObsidianVaultSection`) through the shared workspace-link layer instead of trusting an RPC-supplied path. It adds a thin `resolve_workspace_absolute_path` Tauri command wrapping the existing internal `resolve_workspace_path` guard, a matching `resolveWorkspaceAbsolutePath` TS wrapper, and rewires `fireDeepLink` to canonicalize `memory_tree/content` server-side before building the `obsidian://open?path=<abs>` URL. The change is tight, well-documented, and net-positive for security. CI is green (the one non-passing check, *Rust Core Tests (Windows — secrets ACL)*, is `cancelled` and tests core code this PR does not touch). No blockers.

## Changes

| File | Summary |
| --- | --- |
| `app/src-tauri/permissions/allow-workspace-files.toml` | Allow the new `resolve_workspace_absolute_path` command in the capability manifest. |
| `app/src-tauri/src/lib.rs` | Register `workspace_paths::resolve_workspace_absolute_path` in the invoke handler. |
| `app/src-tauri/src/workspace_paths.rs` | New `resolve_workspace_absolute_path` Tauri command wrapping `resolve_workspace_path`; two new unit tests on the internal helper. |
| `app/src/utils/tauriCommands/workspacePaths.ts` | New `resolveWorkspaceAbsolutePath(path)` wrapper (assertTauri + invoke). |
| `app/src/components/intelligence/ObsidianVaultSection.tsx` | `fireDeepLink` now resolves the absolute path via the new command instead of the `contentRootAbs` prop. |
| `app/src/utils/tauriCommands/workspacePaths.test.ts` | Adds success / rejection / non-Tauri tests for the new wrapper. |
| `app/src/components/intelligence/__tests__/ObsidianVaultSection.test.tsx` | Adds resolver-sourced-URL and resolver-failure-toast tests; mocks the new wrapper. |
| `app/src/components/intelligence/__tests__/MemoryWorkspace.test.tsx` | Mocks `resolveWorkspaceAbsolutePath` to keep the existing deep-link assertion stable. |

## Actionable comments (1)

### 💡 Refactor / suggestion

**1. `app/src-tauri/src/workspace_paths.rs:426-451` — new tests exercise the internal helper, not the new command.** Both new tests call `resolve_workspace_path(...)` directly (the pre-existing helper), not the new `resolve_workspace_absolute_path` command this PR adds, whose body (config load, resolve, label/log, to_string_lossy) has no direct Rust test — only integration coverage via `workspacePaths.ts`. `resolve_workspace_path_returns_absolute_inside_workspace` also overlaps the existing `resolve_workspace_path_accepts_existing_relative_file_inside_workspace`; the genuinely-new coverage is the containment assertion plus the empty-input rejection. The coverage gate passes, so no change is strictly required — but the names overpromise. Suggest renaming to reflect they assert the shared helper the command delegates to (e.g. `resolve_workspace_path_resolves_memory_tree_content_inside_workspace`). Posted inline.

## Nitpicks (2)

- `app/src/components/intelligence/ObsidianVaultSection.tsx:206` — the button `title` tooltip still previews the URL built from the `contentRootAbs` prop, but the URL actually fired now comes from the canonicalized resolver output. On macOS these diverge (`/var/...` vs `/private/var/...`), so the tooltip can show a different path than the link opens. Cosmetic.
- `app/src/components/intelligence/ObsidianVaultSection.tsx:113` — on resolver failure the error toast interpolates `contentRootAbs`, but the failure concerns the *resolved* path (which may differ / be missing on disk). They match in the normal case, so low impact; mentioning since the prop is now decoupled from the actual deep-link target.

- `app/src/components/intelligence/ObsidianVaultSection.tsx:75` — `resolve_workspace_path` requires the target to **exist** (canonicalize errors otherwise). Previously `fireDeepLink` built the URL straight from the prop and never touched disk, so it fired even when `memory_tree/content` was absent. Now a missing folder rejects and the link never fires (error toast). The "behavior unchanged" claim holds for the registered case (Obsidian only registers folders that exist) and aligns with `reveal` (already existence-gated), but the unregistered + not-yet-created case does change. Is `graph.content_root_abs` always backed by an existing directory when this component renders? Posted inline.

## Verified / looks good

- TS wrapper mirrors its siblings exactly (assertTauri + invoke) and is tested for success, invoke rejection, and the non-Tauri throw.
- Component tests prove the URL is sourced from the resolver (a distinct path from the prop) and that resolver failure surfaces an `error` toast + keeps guidance expanded — happy and failure paths both covered.
- `fireDeepLink` `useCallback` deps correctly dropped to empty (no longer closes over `contentRootAbs`; the path is a module constant).
- Permission manifest and `lib.rs` invoke-handler registration both updated — the command is reachable end-to-end.
- Security is net-positive: the deep link now passes through the same canonicalize + workspace-containment guard as `open_workspace_path` / `reveal_workspace_path`; errors stay non-leaky.
- No new i18n keys required — the failure path reuses existing `workspace.openVaultFailed*` toast keys.
- Rustdoc on the command and JSDoc on the wrapper both document the why and the error contract.

---
*Approving — no blockers, no majors. The one refactor note and the open question are non-blocking; addressing them (or confirming the existence invariant) would tidy things up.*

graycyrus (commented):
@YOMXXX heads up — CI is failing on this PR (Frontend Unit Tests, Frontend Coverage, Rust Core Tests on Windows), so I'll hold off on approving until those are sorted. I did spot one thing while looking through the diff though.

The CI failure in Frontend Unit Tests is almost certainly coming from the new test `resolver failure surfaces an error toast and keeps guidance expanded` in `ObsidianVaultSection.test.tsx`. That test expects `openUrl` to not be called and an error toast to fire when the resolver rejects — but the component's inner catch silently falls back to `contentRootAbs` and calls `openUrl` anyway. The test and the implementation directly contradict each other.

Looking at `ObsidianVaultSection.tsx` around line 75–84: when `resolveWorkspaceAbsolutePath` throws, the catch block assigns `absolutePath = contentRootAbs` and execution continues to `openUrl(url)`. The test mocks `openUrl` as resolved and asserts it was never called — that assertion will always fail given this fallback path. Same for the error toast — with a successful `openUrl` call, `fireDeepLink` returns `null` and no toast fires.

The PR description says resolver rejection should surface an error toast, but the code says it should silently fall back. One of these needs to change. If the intent is graceful degradation (fall back to the RPC path), the test is wrong and should instead verify the fallback URL is used. If the intent is to surface failures as errors (which the test and PR description both imply), the inner catch should re-throw rather than swallow.

Everything else looks good — Rust command is tight, TS wrapper is idiomatic, the PII log fix is correct, permission manifest is in sync, and the security story is net positive. Fix the CI failures and resolve the test/implementation disagreement and this is ready to go.

app/src/components/intelligence/ObsidianVaultSection.tsx
```diff
@@ -55,15 +58,35 @@ export function ObsidianVaultSection({ contentRootAbs, onToast }: ObsidianVaultS
...
+      let absolutePath: string;
+      try {
+        absolutePath = await resolveWorkspaceAbsolutePath(MEMORY_CONTENT_WORKSPACE_PATH);
+      } catch {
+        // The content directory may not exist yet (fresh workspace). Fall
+        // back to the prop from the graph-export RPC — the deep link URL
+        // is still valid; Obsidian will surface its own "vault not found"
+        // error if the folder is truly missing.
+        absolutePath = contentRootAbs;
+      }
```
graycyrus: **[major]** The test for resolver failure (`ObsidianVaultSection.test.tsx:153`) asserts that `openUrl` is NOT called and an error toast IS fired when this catch is hit — but this catch falls back to `contentRootAbs` and `openUrl` fires successfully, so neither assertion can pass.

Decide which behavior you actually want:

1. **Graceful degradation** (fallback to prop): remove or rework the new `resolver failure` test so it verifies the fallback URL is used instead of asserting an error state.
2. **Fail loudly** (surface error toast on resolver rejection): remove the fallback and re-throw so the outer catch handles it as an error.

Right now the test and the code describe different contracts.

graycyrus:
@YOMXXX unresolved review feedback — please address before we review.

graycyrus:

app/src-tauri/src/workspace_paths.rs
YOMXXX: Good point — the test names overpromise. They exercise `resolve_workspace_path` (the shared helper) rather than the new `resolve_workspace_absolute_path` Tauri command. Will rename to `resolve_workspace_path_resolves_memory_tree_content_inside_workspace` and `resolve_workspace_path_rejects_empty_input` to reflect what they actually cover.

app/src/components/intelligence/ObsidianVaultSection.tsx
YOMXXX: Good question. Yes — `content_root_abs` comes from the `memory_tree_graph_export` RPC, which writes the graph to disk before returning the path. So by the time `ObsidianVaultSection` renders with that prop, the directory exists. The resolver call just canonicalizes it for the Obsidian deep link URL.

For the edge case of a completely fresh workspace where the graph hasn't been exported yet: the Memory tab only shows the vault section after a successful export, so the component never renders with a non-existent path in practice.

app/src/components/intelligence/ObsidianVaultSection.tsx
YOMXXX: Agreed — the test and code described different contracts. Fixed in `40503a9b`: removed the inner fallback catch entirely so resolver rejection propagates to the outer catch → returns error → `toastOpenOutcome` fires `type: "error"` → guidance panel stays expanded. This is the "fail loudly" path (option 2).

The rationale: if the resolver can't canonicalize the path, the deep link URL would be unreliable anyway. Better to surface the error and let the user use "Open anyway" (which still uses `contentRootAbs` directly and skips the resolver).

PR #2704, opened 2026-05-26
feat(search): unified web-access firewall (fetch + browser) with tri-state Allowed-websites control + friendly block errors

sanil-23:
## Summary

**Settings → Search** gains a single **"Allowed websites"** control that governs every way the agent reaches the web on-disk, backed by `[http_request].allowed_domains` via the existing `config.get/update_search_settings` RPC:

- **Tri-state access mode — Allow all · Custom · Block all** (segmented control). Replaces the earlier binary "Allow all" toggle; **"Block all" is now a first-class, labelled state** instead of the hidden "leave the box empty" path. Typed hosts are preserved when switching modes.
- **Unified fetch + browser firewall.** The browser tool (`browser_open` + `browser`) now shares the same host list via a new `tools::ops::browser_allowed_domains()` helper instead of the now-deprecated `[browser].allowed_domains`. The `"*"` allow-all wildcard is **stripped for the browser**, so a fetch-side "Allow all" never silently opens a real Chromium (JS + logged-in sessions) to every site. Browser allow-all stays gated by `OPENHUMAN_BROWSER_ALLOW_ALL` + `browser.enabled` — **fail-safe: unifying can only narrow the browser's reach, never widen it.**
- **Default `["*"]`** (allow all public sites) so web research works out of the box. The SSRF guard still blocks local/private hosts regardless. `url_guard`'s matcher treats `"*"` as allow-all.
- **Friendly `web_fetch`/`curl` block errors** that point the user/LLM to the setting instead of leaking `config.toml`/internal keys.
- **Web search is *not* gated by this list** (it's controlled by the engine + API key only); the UI hint says so.
- **i18n synced to all locales** — the new access-mode keys are added to every `*-1` chunk file, not just English.

## Problem

The agent's `research` flow uses `web_fetch`, gated by the SSRF-hardened host allowlist in `url_guard`. Two issues made this a wall:

1. The allowlist had **no UI** and **no RPC** — editable only by hand in `config.toml`, and the default was empty (every fresh user's first web search failed).
2. The block errors were dev-oriented; the subagent circuit-breaker then surfaced a generic *"web search isn't available"* with no path to fix it.

On top of that, OpenHuman had **two separate web-access firewalls** — `[http_request].allowed_domains` (fetch/curl) and `[browser].allowed_domains` (the browser tool) — with opposite defaults and no shared UI, so "what can the agent reach on the web?" had no single answer. And `url_guard`'s matcher had **no `*` wildcard**, so there was no clean "allow everything" option.

## Solution

- `config/ops.rs` + `config/schemas.rs`: extend `SearchSettingsPatch` / `SearchSettingsUpdate` and `get_search_settings` with `allowed_domains` + `allow_all`, reading/writing `[http_request].allowed_domains`. `allow_all=true` collapses to `["*"]`; `false` drops the wildcard.
- `config/schema/tools.rs`: `HttpRequestConfig` defaults `allowed_domains` to `["*"]`; `[browser].allowed_domains` marked **deprecated** (still parsed, no longer gates navigation).
- `url_guard.rs`: `host_matches_allowlist` treats `"*"` as allow-all (public hosts only — local/private still rejected); both block errors rewritten to name the host, cause, and exact fix.
- `tools/ops.rs`: new `browser_allowed_domains()` helper feeds the browser tool the shared fetch list **with `"*"` stripped**; `web_fetch`/`curl`/`http_request` keep treating `"*"` as allow-all.
- `SearchPanel.tsx`: the "Allowed websites" section is now a tri-state **Allow all / Custom / Block all** segmented control (the host editor + Save appear only in Custom). Hint copy clarifies the list covers fetch + browser, not search.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — added/updated tests target the changed lines: `url_guard` wildcard + SSRF, `HttpRequestConfig` default, `apply_search_settings` round-trip, `browser_allowed_domains` (shares fetch list minus `"*"`), and the rewritten SearchPanel suite (each mode + host-preservation). CI `coverage.yml` enforces the gate.
- [x] N/A: no `docs/TEST-COVERAGE-MATRIX.md` feature row maps to this Settings/allowlist control — flag in review if a row is wanted.
- [x] No new external network dependencies introduced (this change *governs* outbound access; it adds none).
- [x] N/A: does not touch release-cut surfaces in `docs/RELEASE-MANUAL-SMOKE.md`.
- [x] N/A: no tracking issue — feature requested directly; nothing to auto-close.

## Impact

- **Desktop** (core + React). Behavior change: `web_fetch`/`curl` is **allowed by default** (`["*"]`) where it was effectively blocked; local/private hosts remain blocked by the SSRF guard. The browser tool now derives its allowlist from the same list (minus `"*"`).
- **Security posture:** one place to reason about the agent's web reach. The browser tool can only ever be *narrowed* by this unification — it never inherits fetch's blanket `"*"`, and still requires `browser.enabled` + `OPENHUMAN_BROWSER_ALLOW_ALL` to go wide. Broader default outbound reach for fetch (public hosts only) is a deliberate zero-friction tradeoff that the new UI makes easy to scope down.

## Related

- Resolved follow-ups from the original PR scope: ✅ folded `[browser].allowed_domains` into the unified "Allowed websites" control; ✅ synced the new i18n keys to all non-English locale chunks.
- Follow-up PR(s)/TODOs: remove the now-dead `settings.search.allowAllAria` i18n key (cosmetic).

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: feat/web-access-allowlist-settings
- Commit SHA: 11c6e9fc

### Validation Run
- [x] `pnpm exec prettier --check` — clean on changed frontend files (full repo check skipped due to known CRLF drift).
- [x] `pnpm typecheck` — clean (`tsc --noEmit`, exit 0).
- [x] `pnpm i18n:check` — missing 0 / extra 0 / drifted 0 across all locales (new keys present in every chunk-1).
- [x] Focused tests: `pnpm debug unit SearchPanel` → 7 passed.
- [x] Rust: `cargo check --manifest-path Cargo.toml --lib --tests` — exit 0 (reused the local dev build's cached sys-crate artifacts in the shared target).
- [x] N/A: Tauri fmt/check — no `app/src-tauri` source changes (core + frontend only).

### Validation Blocked
- `command:` `cargo test` (run the new `browser_allowed_domains` unit test)
- `error:` linking the Rust test binary needs the MSVC env (`vcvars`/SDK/linker) that the Windows dev wrapper sets up; the bare shell lacks it, and `cargo check --tests` (which passed) does not link.
- `impact:` the `browser_allowed_domains` test compiles but was not *executed* locally — it's a pure deterministic filter, and CI runs it. Everything else above was verified locally.

### Behavior Changes
- Intended: single tri-state web-access control; browser tool shares the fetch allowlist (minus `"*"`); `web_fetch`/`curl` default to allow-all; friendlier block errors.
- User-visible: Settings → Search now shows Allow all / Custom / Block all; web research works without config; clearer guidance when a host is blocked. No change to a vault/registration surface.

### Parity Contract
- Legacy preserved: exact-host + subdomain matching unchanged; SSRF blocklist (loopback/RFC1918/link-local/…) unchanged and still enforced even with `"*"`. `allow_all=false` + empty list reproduces the old block-by-default behavior.
- Browser parity: with `"*"` stripped, an explicit-host list applies to the browser exactly as before; the browser never gains blanket access from a fetch-side toggle, and `browser.enabled` + `OPENHUMAN_BROWSER_ALLOW_ALL` gates are untouched.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Website allowlist settings: three-mode selector (Allow all / Custom / Block all) with multiline domain editor and save control; browser and fetch use the unified allowlist.

* **Bug Fixes**
  * Clearer "allowed websites" messages; defaults now allow public sites by default while still blocking local/private addresses.

* **Tests**
  * Comprehensive unit and UI tests for loading, mode switching, editing, normalization, and persistence.

* **Documentation**
  * Added translation strings for the new UI across many languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2704?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/components/settings/panels/SearchPanel.tsx
+++ b/app/src/components/settings/panels/SearchPanel.tsx
@@ -38,6 +38,9 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
   const [braveKey, setBraveKey] = useState<string>('');
   const [showParallel, setShowParallel] = useState(false);
   const [showBrave, setShowBrave] = useState(false);
+  // Editor text for the allowed-websites host list (one host per line). The
+  // "*" wildcard is represented by the allowAll toggle, not shown here.
+  const [allowedText, setAllowedText] = useState<string>('');
 
   const ENGINES: EngineOption[] = [
     {
@@ -80,7 +83,15 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
     };
   }, []);
 
+  // Reflect the loaded allowlist into the editor (explicit hosts only — the
+  // "*" wildcard is surfaced through the allowAll toggle instead).
+  useEffect(() => {
+    if (!settings) return;
+    setAllowedText(settings.allowed_domains.filter(d => d !== '*').join('\n'));
+  }, [settings]);
+
   const selectedEngine = (settings?.engine as SearchEngineId | undefined) ?? 'managed';
+  const allowAll = settings?.allow_all ?? false;
 
   const persistEngine = async (next: SearchEngineId) => {
     if (!settings || status.kind === 'saving') return;
@@ -109,6 +120,37 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
       setSettings(refreshed.result);
       if (engine === 'parallel') setParallelKey('');
       else setBraveKey('');
+      setStatus({ kind: 'saved' });
+    } catch (err) {
+      setStatus({ kind: 'error', message: err instanceof Error ? err.message : String(err) });
+    }
+  };
+
+  const persistAllowAll = async (next: boolean) => {
+    if (!settings || status.kind === 'saving') return;
+    setStatus({ kind: 'saving' });
+    try {
+      await openhumanUpdateSearchSettings({ allow_all: next });
+      const refreshed = await openhumanGetSearchSettings();
+      setSettings(refreshed.result);
+      setStatus({ kind: 'saved' });
+    } catch (err) {
+      setStatus({ kind: 'error', message: err instanceof Error ? err.message : String(err) });
+    }
+  };
+
+  const persistAllowedDomains = async () => {
+    if (!settings || status.kind === 'saving') return;
+    const domains = allowedText
+      .split('\n')
+      .map(s => s.trim())
+      .filter(Boolean);
+    setStatus({ kind: 'saving' });
+    try {
+      // Editing the explicit host list implies "not allow-all".
+      await openhumanUpdateSearchSettings({ allowed_domains: domains, allow_all: false });
+      const refreshed = await openhumanGetSearchSettings();
+      setSettings(refreshed.result);
       setStatus({ kind: 'saved' });
     } catch (err) {
       setStatus({ kind: 'error', message: err instanceof Error ? err.message : String(err) });
@@ -258,6 +300,55 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
                 docUrl="https://brave.com/search/api/"
                 t={t}
               />
+            </div>
+
+            {/* Allowed websites — host allowlist for web_fetch / curl. */}
+            <div className="rounded-xl border border-stone-200 dark:border-neutral-800 bg-white dark:bg-neutral-900 p-3 space-y-2">
+              <div className="flex items-center justify-between">
+                <label className="text-xs font-semibold text-stone-700 dark:text-neutral-200">
+                  {t('settings.search.allowedSitesLabel')}
+                </label>
+                <button
+                  type="button"
+                  role="switch"
+                  aria-checked={allowAll}
+                  aria-label={t('settings.search.allowAllAria')}
+                  onClick={() => void persistAllowAll(!allowAll)}
+                  disabled={status.kind === 'saving'}
+                  className={`relative inline-flex h-5 w-9 items-center rounded-full transition-colors disabled:opacity-50 ${
+                    allowAll ? 'bg-primary-500' : 'bg-stone-300 dark:bg-neutral-700'
+                  }`}>
+                  <span
+                    className={`inline-block h-4 w-4 transform rounded-full bg-white transition-transform ${
+                      allowAll ? 'translate-x-4' : 'translate-x-0.5'
+                    }`}
+                  />
+                </button>
+              </div>
+              <p className="text-[11px] text-stone-500 dark:text-neutral-400 leading-relaxed">
+                {allowAll
+                  ? t('settings.search.allowedSitesAllOn')
+                  : t('settings.search.allowedSitesHint')}
+              </p>
+              {!allowAll && (
+                <>
+                  <textarea
+                    value={allowedText}
+                    onChange={e => setAllowedText(e.target.value)}
+                    rows={4}
+                    spellCheck={false}
+                    placeholder={t('settings.search.allowedSitesPlaceholder')}
+                    className="w-full rounded-lg border border-stone-200 dark:border-neutral-800 bg-stone-50 dark:bg-neutral-800/60 px-2 py-1.5 text-xs font-mono text-stone-800 dark:text-neutral-100 focus:outline-none focus-visible:border-primary-400"
+                  />
+                  <button
+                    type="button"
+                    onClick={() => void persistAllowedDomains()}
+                    disabled={status.kind === 'saving'}
+                    className="px-3 py-1.5 rounded-lg text-xs font-medium bg-primary-500 text-white hover:bg-primary-600 disabled:opacity-50">
+                    {t('settings.search.allowedSitesSave')}
+                  </button>
+                </>
+              )}
             </div>
 
             <div
--- a/app/src/components/settings/panels/SearchPanel.tsx
+++ b/app/src/components/settings/panels/SearchPanel.tsx
@@ -39,6 +39,19 @@ interface EngineOption {
   requiresKey: boolean;
 }
 
+/**
+ * Normalize a user-entered allowed-site entry down to a bare host so it
+ * matches `url_guard`'s host-based comparison. Strips a leading scheme and any
+ * path/query/fragment — e.g. `https://reuters.com/markets` → `reuters.com` —
+ * and trims surrounding whitespace. The `*` allow-all wildcard is preserved.
+ */
+const normalizeAllowedHost = (raw: string): string =>
+  raw
+    .trim()
+    .replace(/^[a-z][a-z0-9+.-]*:\/\//i, '')
+    .replace(/\/.*$/, '')
+    .trim();
+
 const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
   const { t } = useT();
   const { navigateBack, breadcrumbs } = useSettingsNavigation();
@@ -173,10 +186,7 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
   };
 
   const persistAllowedDomains = () => {
-    const domains = allowedText
-      .split('\n')
-      .map(s => s.trim())
-      .filter(Boolean);
+    const domains = allowedText.split('\n').map(normalizeAllowedHost).filter(Boolean);
     // Editing the explicit host list implies "not allow-all".
     void persistSearchUpdate({ allowed_domains: domains, allow_all: false });
   };
@@ -330,9 +340,11 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
                 curl and (when enabled) the browser tool. Web search is not
                 gated by this list. */}
             <div className="rounded-xl border border-stone-200 dark:border-neutral-800 bg-white dark:bg-neutral-900 p-3 space-y-2">
-              <label className="text-xs font-semibold text-stone-700 dark:text-neutral-200">
+              {/* Section heading, not a form label — use a <p> so screen
+                  readers don't announce an orphan <label> with no htmlFor. */}
+              <p className="text-xs font-semibold text-stone-700 dark:text-neutral-200">
                 {t('settings.search.allowedSitesLabel')}
-              </label>
+              </p>
               <div
                 role="radiogroup"
                 aria-label={t('settings.search.accessModeAria')}
--- a/app/src/components/settings/panels/SearchPanel.tsx
+++ b/app/src/components/settings/panels/SearchPanel.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import { useT } from '../../../lib/i18n/I18nContext';
 import { useCoreState } from '../../../providers/CoreStateProvider';
@@ -8,6 +8,7 @@ import {
   openhumanUpdateSearchSettings,
   type SearchEngineId,
   type SearchSettings,
+  type SearchSettingsUpdate,
 } from '../../../utils/tauriCommands/config';
 import SettingsHeader from '../components/SettingsHeader';
 import { useSettingsNavigation } from '../hooks/useSettingsNavigation';
@@ -18,6 +19,18 @@ type Status =
   | { kind: 'saving' }
   | { kind: 'saved' }
   | { kind: 'error'; message: string };
+
+/**
+ * Tri-state web-access mode for the unified fetch + browser allowlist.
+ * - `all`    → `allow_all: true` (the `"*"` wildcard)
+ * - `custom` → `allow_all: false` + an explicit host list (textarea)
+ * - `block`  → `allow_all: false` + an empty host list (no web access)
+ *
+ * `block` and an empty `custom` are indistinguishable once persisted (both are
+ * `allow_all: false` + `[]`); the distinction only matters locally while
+ * editing.
+ */
+type AccessMode = 'all' | 'custom' | 'block';
 
 interface EngineOption {
   id: SearchEngineId;
@@ -39,8 +52,14 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
   const [showParallel, setShowParallel] = useState(false);
   const [showBrave, setShowBrave] = useState(false);
   // Editor text for the allowed-websites host list (one host per line). The
-  // "*" wildcard is represented by the allowAll toggle, not shown here.
+  // "*" wildcard is represented by the access mode, not shown here.
   const [allowedText, setAllowedText] = useState<string>('');
+  // Tri-state web-access mode for the unified fetch + browser allowlist.
+  const [mode, setMode] = useState<AccessMode>('all');
+  // Sync editor + mode from settings exactly once, so a later settings refresh
+  // (e.g. after saving an engine change) can't clobber the user's in-progress
+  // host edits or chosen mode.
+  const initializedRef = useRef(false);
 
   const ENGINES: EngineOption[] = [
     {
@@ -83,15 +102,16 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
     };
   }, []);
 
-  // Reflect the loaded allowlist into the editor (explicit hosts only — the
-  // "*" wildcard is surfaced through the allowAll toggle instead).
+  // Reflect the loaded allowlist into the editor + mode, exactly once.
   useEffect(() => {
-    if (!settings) return;
-    setAllowedText(settings.allowed_domains.filter(d => d !== '*').join('\n'));
+    if (!settings || initializedRef.current) return;
+    initializedRef.current = true;
+    const explicit = settings.allowed_domains.filter(d => d !== '*');
+    setAllowedText(explicit.join('\n'));
+    setMode(settings.allow_all ? 'all' : explicit.length > 0 ? 'custom' : 'block');
   }, [settings]);
 
   const selectedEngine = (settings?.engine as SearchEngineId | undefined) ?? 'managed';
-  const allowAll = settings?.allow_all ?? false;
 
   const persistEngine = async (next: SearchEngineId) => {
     if (!settings || status.kind === 'saving') return;
@@ -126,11 +146,11 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
     }
   };
 
-  const persistAllowAll = async (next: boolean) => {
+  const persistSearchUpdate = async (update: SearchSettingsUpdate) => {
     if (!settings || status.kind === 'saving') return;
     setStatus({ kind: 'saving' });
     try {
-      await openhumanUpdateSearchSettings({ allow_all: next });
+      await openhumanUpdateSearchSettings(update);
       const refreshed = await openhumanGetSearchSettings();
       setSettings(refreshed.result);
       setStatus({ kind: 'saved' });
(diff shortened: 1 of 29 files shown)


src/openhuman/config/ops.rs
```diff
@@ -997,6 +1006,26 @@ pub async fn apply_search_settings(
...
+        cleaned.dedup();
+        config.http_request.allowed_domains = cleaned;
+    }
+    if let Some(allow_all) = update.allow_all {
+        if allow_all {
+            config.http_request.allowed_domains = vec!["*".to_string()];
+        } else {
+            config.http_request.allowed_domains.retain(|d| d != "*");
+        }
+    }
```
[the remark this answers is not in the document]
sanil-23: Done in 4e3030be — `apply_search_settings` now emits a grep-friendly `tracing::info!("[config] http_request.allowed_domains updated")` with `before_count`/`after_count`/`before_allow_all`/`after_allow_all` correlation fields, gated on the allowlist actually being touched. Per the repo redaction rule it logs counts + the wildcard flag only, never the raw hosts.

src/openhuman/tools/impl/network/url_guard.rs
```diff
@@ -45,7 +45,9 @@ pub(super) fn validate_url(raw_url: &str, allowed_domains: &[String]) -> anyhow:
 
     if allowed_domains.is_empty() {
         anyhow::bail!(
-            "Network tool is enabled but no allowed_domains are configured. Add [http_request].allowed_domains in config.toml"
+            "I'm not allowed to open any website yet — the allowed websites list is empty. \
+             Add the site you need (or turn on \"Allow all sites\") under \
+             Settings → Developer Options → Search engine → Allowed websites, then ask me again."
```
graycyrus: **[minor]** The hardcoded path `Settings → Developer Options → Search engine → Allowed websites` appears in both block-error messages (here and line 65). Please verify it matches the actual UI navigation — from the `SearchPanel` component it looks like the control might live under `Settings → Search → Allowed websites` instead.

If the path is wrong, users will hunt for a setting that doesn't exist at that location. Worth a quick manual check since this is the primary user-facing guidance when a fetch gets blocked.

graycyrus (changes requested):
Good architecture — unifying two separate web-access firewalls into one settings-backed control is the right move, and the fail-safe design (browser never inherits `"*"` from fetch via `browser_allowed_domains()`) is solid. The `initializedRef` pattern in the frontend prevents settings refreshes from clobbering in-progress host edits. Tests cover all tri-state transitions on both sides.

**Blockers:**

1. **CI is red.** Frontend Unit Tests and Rust Core Tests + Quality are both failing on the latest commit. Need green before this can move forward.
2. **Verify the settings navigation path** in the `url_guard` block-error messages — see inline comment.

**Minor notes** (not blocking):
- Non-English i18n chunks still carry the old hint: *"Leave empty to block all web access"* — now misleading since "Block all" is an explicit toggle. Worth syncing the fallback text in a follow-up.
- Orphaned `settings.search.allowAllAria` key — already noted in the PR description for follow-up, 👍.

| Area | Files | Verdict |
|------|-------|---------|
| Rust config | `ops.rs`, `schemas.rs`, `schema/tools.rs`, `ops_tests.rs` | ✓ clean — defaults, RPC, round-trip test |
| Rust tools | `tools/ops.rs`, `tools/ops_tests.rs` | ✓ `browser_allowed_domains` well-tested |
| Rust url_guard | `url_guard.rs`, `http_request_tests.rs` | wildcard + SSRF ✓ — error-message path needs check |
| Frontend | `SearchPanel.tsx`, `.test.tsx`, `config.ts` | ✓ solid |
| i18n | 13 chunks + `en.ts` | keys synced, hint-text drift in non-EN (minor) |

*CodeRabbit dedup: skipped transition-log suggestion (ops.rs:1028) and Prettier formatting (addressed in 228b533).*

app/src/lib/i18n/chunks/ar-1.ts
```diff
@@ -1319,6 +1319,12 @@ const ar1: TranslationMap = {
   'subconscious.priority.normal': 'عادي',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

app/src/lib/i18n/chunks/bn-1.ts
```diff
@@ -1329,6 +1329,12 @@ const bn1: TranslationMap = {
   'subconscious.priority.normal': 'স্বাভাবিক',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

app/src/lib/i18n/chunks/es-1.ts
```diff
@@ -1344,6 +1344,12 @@ const es1: TranslationMap = {
   'subconscious.priority.normal': 'normales',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

app/src/lib/i18n/chunks/fr-1.ts
```diff
@@ -1349,6 +1349,12 @@ const fr1: TranslationMap = {
   'subconscious.priority.normal': 'normal',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

app/src/lib/i18n/chunks/hi-1.ts
```diff
@@ -1327,6 +1327,12 @@ const hi1: TranslationMap = {
   'subconscious.priority.normal': 'सामान्य',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

app/src/lib/i18n/chunks/id-1.ts
```diff
@@ -1332,6 +1332,12 @@ const id1: TranslationMap = {
   'subconscious.priority.normal': 'normal',
   'subconscious.durationSeconds': '{seconds}s',
   'subconscious.durationMilliseconds': '{milliseconds}ms',
+  "settings.search.allowedSitesLabel": "Allowed websites",
+  "settings.search.allowAllAria": "Allow all sites",
+  "settings.search.allowedSitesHint": "Websites the assistant may open and read while researching (one host per line, e.g. reuters.com). A host also covers its subdomains. Leave empty to block all web access.",
+  "settings.search.allowedSitesAllOn": "The assistant can open any public website. Local and private addresses stay blocked.",
+  "settings.search.allowedSitesPlaceholder": "reuters.com\napnews.com\ngithub.com",
+  "settings.search.allowedSitesSave": "Save websites",
```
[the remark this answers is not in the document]
sanil-23: Already resolved in 228b5332 (`style(i18n): prettier-format mirrored locale chunks`), which landed after this review. Verified on the current tip: `prettier --check` (repo config: `singleQuote`, `printWidth: 100`, `@trivago/prettier-plugin-sort-imports`) reports **all matched files use Prettier code style** across every changed chunk, so this hunk is clean now.

M3gA-Mind (commented):
Overall: well-structured PR with solid Rust + Vitest coverage and correct security semantics (browser can only be narrowed, never widened). Two issues below need fixing before merge; the rest are suggestions.

src/openhuman/tools/impl/network/url_guard.rs
M3gA-Mind: **Bug — wrong settings path in error message.**

This says `Settings → Developer Options → Search engine → Allowed websites`, but the new control lives at **Settings → Search → Allowed websites** (it's in `SearchPanel`, not under a "Developer Options" subtree). A user following this error won't find the setting.

Same issue on line 64.

src/openhuman/tools/impl/network/url_guard.rs
```diff
@@ -56,7 +58,11 @@ pub(super) fn validate_url(raw_url: &str, allowed_domains: &[String]) -> anyhow:
     }
 
     if !host_matches_allowlist(&host, allowed_domains) {
-        anyhow::bail!("Host '{host}' is not in http_request.allowed_domains");
+        anyhow::bail!(
+            "I'm not allowed to open '{host}' — it isn't in your allowed websites. \
+             Add it (or turn on \"Allow all sites\") under \
+             Settings → Developer Options → Search engine → Allowed websites, then ask me again."
```
M3gA-Mind: Same wrong path as the empty-list error above — `Developer Options → Search engine` doesn't match the actual panel location.

app/src/lib/i18n/en.ts
```diff
@@ -632,6 +632,20 @@ const en: TranslationMap = {
   'settings.search.placeholderStored': '•••••••• (stored)',
   'settings.search.placeholderParallel': 'pk_...',
   'settings.search.placeholderBrave': 'BSA...',
+  'settings.search.allowedSitesLabel': 'Allowed websites',
+  'settings.search.allowAllAria': 'Allow all sites',
```
M3gA-Mind: **Dead i18n key** — `settings.search.allowAllAria` is never referenced in the UI. The segmented control uses `settings.search.accessModeAria` instead. The PR body even flags this as cleanup debt. It's also been added to all 13 non-English locale chunk files. Better to drop it here and not ship it.

app/src/components/settings/panels/SearchPanel.tsx
```diff
@@ -115,6 +146,41 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
...
+    } else if (next === 'block') {
+      void persistSearchUpdate({ allowed_domains: [], allow_all: false });
+    }
+  };
+
+  const persistAllowedDomains = () => {
+    const domains = allowedText
+      .split('\n')
+      .map(s => s.trim())
+      .filter(Boolean);
```
M3gA-Mind: **No hostname validation.** `split('\n').trim().filter(Boolean)` doesn't check that entries look like hostnames. A user entering `https://reuters.com` or `reuters.com/path` will get an entry that never matches `url_guard`'s host-based comparison. Consider stripping a leading `https?://` and any path component, or at minimum showing an inline hint (e.g. the placeholder already does this well — a validation note in the hint copy for Custom mode would help).

app/src/components/settings/panels/SearchPanel.tsx
```diff
@@ -260,6 +326,73 @@ const SearchPanel = ({ embedded = false }: { embedded?: boolean }) => {
               />
             </div>
 
+            {/* Allowed websites — unified host allowlist shared by web_fetch /
+                curl and (when enabled) the browser tool. Web search is not
+                gated by this list. */}
+            <div className="rounded-xl border border-stone-200 dark:border-neutral-800 bg-white dark:bg-neutral-900 p-3 space-y-2">
+              <label className="text-xs font-semibold text-stone-700 dark:text-neutral-200">
```
M3gA-Mind: **Accessibility: `<label>` not associated with any input.** This element has no `htmlFor` referencing an input's `id`, so screen readers will announce it as an orphan label. Since it acts as a section heading rather than a form label, a `<p>` or `<span>` with `font-semibold` would be semantically cleaner here.

src/openhuman/config/schema/tools.rs
```diff
@@ -134,17 +139,36 @@ impl Default for BrowserConfig {
...
+            allowed_domains: default_http_allowed_domains(),
+            max_response_size: default_http_max_response_size(),
+            timeout_secs: default_http_timeout_secs(),
+        }
+    }
+}
+
+fn default_http_allowed_domains() -> Vec<String> {
+    vec!["*".to_string()]
+}
```
M3gA-Mind, line 170: **Behavior change for existing users — worth calling out explicitly.** Previously `HttpRequestConfig` had no `allowed_domains` default (empty = block-all). After this PR, any existing `config.toml` that omits the key will get `["*"]` on next read, silently opening outbound web access.

The SSRF guard limits the blast radius to public hosts, and the new Settings UI makes it easy to lock down, so this tradeoff is reasonable — but it should appear as a note in the release changelog so users who relied on the restrictive default are aware.

src/openhuman/config/ops.rs
M3gA-Mind, line 1031: **Patch field ordering: `allow_all` wins over `allowed_domains` in the same call.** Because `allow_all` is evaluated after `allowed_domains`, a combined patch like `{ allowed_domains: ["github.com"], allow_all: true }` silently discards the explicit list and writes `["*"]`. The Rust docstring documents this, but `SearchSettingsUpdate` in `config.ts` doesn't. Worth a one-line note there so callers know not to send both fields together with conflicting intent.

src/openhuman/tools/impl/network/url_guard.rs
sanil-23: Good catch — fixed in 3b105761. The path *was* wrong, though the accurate one is **Settings → Advanced → Search engine → Allowed websites** (not `Settings → Search`): SearchPanel is registered as the `search` entry inside DeveloperOptionsPanel, and that section renders as "Advanced" (`settings.developerOptions` = "Advanced"). Both block-error messages now point there.

sanil-23: Good catch — fixed in 3b105761. The path *was* wrong, though the accurate one is **Settings → Advanced → Search engine → Allowed websites** (not `Settings → Search`): SearchPanel is registered as the `search` entry inside DeveloperOptionsPanel, and that section renders as "Advanced" (`settings.developerOptions` = "Advanced"). Both block-error messages now point there.

src/openhuman/tools/impl/network/url_guard.rs
sanil-23: Good catch — fixed in 3b105761. The path *was* wrong, though the accurate one is **Settings → Advanced → Search engine → Allowed websites** (not `Settings → Search`): SearchPanel is registered as the `search` entry inside DeveloperOptionsPanel, and that section renders as "Advanced" (`settings.developerOptions` = "Advanced"). Both block-error messages now point there.

app/src/lib/i18n/en.ts
sanil-23: Done in 3b105761 — removed `settings.search.allowAllAria` from `en.ts` and all 14 locale chunk files (UI uses `accessModeAria`). `pnpm i18n:check` stays green (parity 0/0/0).

app/src/components/settings/panels/SearchPanel.tsx
sanil-23: Addressed in 3b105761 — `persistAllowedDomains` now runs each entry through a `normalizeAllowedHost` helper that strips a leading scheme and any path/query/fragment (`https://reuters.com/markets` → `reuters.com`) before persisting, so pasted URLs match `url_guard`'s host comparison. Added a unit test covering it.

app/src/components/settings/panels/SearchPanel.tsx
sanil-23: Fixed in 3b105761 — it's a section heading, so changed the orphan `<label>` (no `htmlFor`) to a `<p>` with the same classes.

src/openhuman/config/schema/tools.rs
sanil-23: Acknowledged — this is intentional and called out in the PR's **Impact** + **Parity Contract** sections (default flips to `["*"]` = public hosts only; SSRF guard still blocks local/private; `allow_all=false` + empty list reproduces the old block-by-default). Agreed it warrants a changelog line at release-cut; flagging for the release notes rather than a code change here.

src/openhuman/config/ops.rs
sanil-23: Done in 3b105761 — added a note to `SearchSettingsUpdate` in `config.ts` that `allow_all` is applied after `allowed_domains` server-side, so it wins in a combined patch and callers shouldn't send both with conflicting intent.

Issue #1856, opened 2026-05-15
Feature Request : Memory Tree wiki status panel - on/off toggle, observability metrics, and team-level knowledge siloing


githb-ac:
# Summary

Add a Memory Tree wiki status panel to the OpenHuman dashboard that gives operators live visibility into the health and activity of the Memory Tree and Obsidian wiki, an on/off toggle to pause or resume the auto-fetch and wiki update cycle, observability metrics (sync count, compression ratio, last updated, chunk count), and the ability to silo knowledge at the team level so each team's agents draw from a scoped subset of the wiki rather than the full shared store.

# Background

OpenHuman's Memory Tree is directly inspired by [Andrej Karpathy's LLM Wiki pattern](https://gist.github.com/karpathy/442a6bf555914893e9891c11519de94f) - a compounding, LLM-maintained personal knowledge base where an LLM writes and maintains structured markdown files as sources are ingested, handling cross-referencing, updates, and contradiction flagging automatically. Rather than re-deriving knowledge from raw documents on every query via RAG, the wiki accumulates and organises knowledge over time into a persistent, compressed, queryable store.

OpenHuman implements this as the [Memory Tree](https://tinyhumans.gitbook.io/openhuman/features/memory-tree) - a local-first knowledge base built from connected accounts and activity, canonicalised into ≤3k-token markdown chunks, scored, and folded into hierarchical summary trees stored in SQLite. The same chunks land as `.md` files in an [Obsidian-compatible vault](https://tinyhumans.gitbook.io/openhuman/features/obsidian-wiki) on the operator's machine.

# Problem

The Memory Tree runs silently in the background on a 20-minute auto-fetch loop. Operators have no dashboard surface to:
- See whether the wiki is actively syncing or stalled
- Know how many chunks exist, when the last sync ran, or what the compression ratio is
- Pause the sync cycle without editing config and restarting
- Understand which integrations are contributing data and which are idle
- Scope wiki knowledge to individual teams, so a Research team's agents don't have their context polluted by the Writing team's accumulated knowledge

As the wiki grows and teams multiply, the absence of observability and siloing becomes a reliability and relevance gap.

# Solution

**Scope:** app (dashboard panel) + core (wiki observability API + siloing)

**Part 1 - Memory Tree status panel**

A Memory Tree tab in the dashboard shows:

| Metric | Description |
|---|---|
| Status | Running / Paused / Syncing / Error |
[...]

PR #2719, opened 2026-05-27
feat(memory-tree): status panel + on/off toggle (#1856 Part 1)

YOMXXX:
## Summary

- Adds the Memory Tree status panel (4 tiles + auto-sync toggle) inside `MemoryWorkspace`, surfacing it in both `/intelligence` and Settings → Memory data with a single mount point.
- Wires up two new RPCs — `memory_tree_pipeline_status` (read-side aggregate) and `memory_tree_set_enabled` (toggle that flips `scheduler_gate.mode` between `auto` and `off`).
- Drops the half-baked `useConsciousItems` + `useMemoryIngestionStatus`-driven status pill from `Intelligence.tsx`; the new panel supersedes it.

## Problem

#1856 calls for a real Memory Tree status surface that exposes pipeline health and lets the user pause auto-ingest without restarting the core. The existing header pill was sourced from the legacy actionable-items pipeline and showed counts that no longer matched anything visible.

## Solution

**Rust (`src/openhuman/memory_tree/tree/rpc.rs`, `src/openhuman/memory/schema.rs`):**
- `pipeline_status_rpc` aggregates `mem_tree_chunks` (COUNT + MAX(timestamp_ms) in a single query) + per-state `mem_tree_jobs` counters + a `walkdir`-driven recursive size of `memory_tree_content_root()`. All blocking I/O runs on `spawn_blocking`. A pure `derive_pipeline_status` helper locks the precedence rules (`paused` > `error` > `syncing` > `running` > `idle`).
- `set_enabled_rpc` flips `config.scheduler_gate.mode = SchedulerGateMode::{Auto, Off}`, persists with `config.save()`, and hot-reloads via `scheduler_gate::gate::update_config(...)` so live workers observe the new policy at their next `wait_for_capacity()` await. Idempotent (`changed=false` for no-ops).
- Schema entries match the existing `RegisteredController` pattern; `all_controller_schemas_and_registered_controllers_stay_in_sync` continues to pass.

**Frontend (`app/src/`):**
- `MemoryTreeStatusPanel.tsx` renders the four-tile dashboard plus an inline toggle row whose markup mirrors `AIPanel`'s `ToggleRow`. Adaptive polling — 1.5s while syncing/active jobs, 4s otherwise — mirrors `useMemoryIngestionStatus`. Error banner with retry button covers the fetch failure path; toggle errors surface via the existing `onToast` channel.
- `memoryTreePipelineStatus()` / `memoryTreeSetEnabled()` wrappers in `tauriCommands/memoryTree.ts` keep the snake_case-on-the-wire convention used by the sibling helpers.
- Mounted above `<MemorySources>` in `MemoryWorkspace` (one place, two surfaces).
- Intelligence header cleanup: removed `useConsciousItems` + `useMemoryIngestionStatus` + `useIntelligenceStats` imports + the `systemStatus*` derivation block + the two pill JSX blocks. The hooks themselves stay — they have their own tests and may serve future consumers.

**i18n:** new `memoryTree.status.*` keys in `en.ts` + the English chunk + all 12 non-English chunk-1 files. `pnpm i18n:check` reports `missing: 0, extra: 0, drifted: 0` across every locale.

**Notes / tradeoffs:**
- The 20-min Composio fetch loop is intentionally NOT paused by this toggle yet — that requires a `Notify` signal on the periodic timer and is queued for #1856 Part 2.
- Schema does not yet record original-source sizes, so `compression_ratio` is deferred until that is plumbed.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 5 new Rust tests in `memory_tree::tree::rpc::tests` (precedence rules, idle/paused/ingested branches, toggle round-trip incl. no-op); 6 new Vitest cases covering tiles, skeleton, toggle round-trip, paused branch, fetch error + retry, and toast on toggle error.
- [x] **Diff coverage ≥ 80%** — every new branch in `pipeline_status_rpc`, `derive_pipeline_status`, `set_enabled_rpc`, the React panel, and the hook is exercised. Local `pnpm debug unit MemoryTreeStatusPanel` and `cargo test --lib memory_tree::tree::rpc::tests` pass with all 6 + 15 cases green.
- [x] N/A: Coverage matrix updated — no feature-row add/remove/rename, only a new status surface within an existing one.
- [x] N/A: All affected feature IDs listed in `## Related` — no matrix-row change.
- [x] No new external network dependencies introduced — only `walkdir` (already in `Cargo.toml`).
- [x] N/A: Manual smoke checklist — not a release-cut surface.
- [x] Linked issue referenced via `Refs #1856` (intentionally NOT `Closes` — Part 2 / Part 3 of the issue stay open).

## Impact

- **Desktop:** users get an immediate, color-coded view of Memory Tree health (Status / Last sync / Total chunks / Wiki size) and a 1-click pause/resume control. Paused state survives core restart (persisted in `config.toml`).
- **Pause semantics:** flipping the toggle off sets `scheduler_gate.mode = off`, which the existing memory queue worker (`memory_queue/worker.rs`) cooperatively respects via `wait_for_capacity()`. The 20-min Composio fetch loop is unaffected (queued for Part 2).
- **Performance:** the status RPC is two SQL counters (one aggregate, one per-job-state) plus a `walkdir` of the memory_tree content root — cheap enough to poll at 1.5–4s. Logged at `debug` level with `[memory-tree][rpc]` and `[ui-flow][memory-tree-status]` prefixes for trace grepping.
- **Compatibility:** no schema migration, no config-shape change beyond toggling an existing enum value.

## Related

Refs #1856 (Part 1 only — Part 2 = pause 20-min Composio fetch loop + team siloing; Part 3 = integration health list).

Follow-up work:
- Pause the 20-min Composio fetch loop on toggle (`memory_sync/composio/periodic.rs` needs a `Notify` signal).
- Add `compression_ratio` once original source sizes land in the chunks schema.
- Integration health list (#1856 Part 3).
- Team-level knowledge siloing (#1856 Part 2).

## Summary by CodeRabbit

* **New Features**
  * Memory Tree Status Panel: 4-tile dashboard (status, last sync, total chunks, storage) with skeletons and live polling.
  * Auto-sync toggle with busy/disabled states, retry flow, and toast notifications on toggle failures.
  * Panel integrated into the Memory workspace, replacing legacy header status pills.
* **Localization**
  * Added translations for the status panel across many languages.
* **Tests**
  * UI tests for display states, loading, error/retry, and toggle interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2719?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


YOMXXX:
Pushed 4 commits addressing all 5 CodeRabbit actionables on top of `feat/1856-memory-tree-status-panel`:

- **Actionable 1 (`memoryTree.ts` wrappers must use `core_rpc_relay`)** — fixed in `1785ac28`. Both `memoryTreePipelineStatus` and `memoryTreeSetEnabled` now route through `invoke('core_rpc_relay', { request: { method, params } })`. The status panel mounts during boot and starts polling before the realtime socket has connected, which is exactly the case AGENTS.md / `.claude/memory.md` flag as a `callCoreRpc` foot-gun (raw `fetch()` fallback → CORS preflight). Kept the rest of the file on `callCoreRpc` since other wrappers are user-initiated and not part of this PR.
- **Actionable 2 (`formatRelativeMs` must go through i18n)** — fixed in `d8c5a802`. Helper now takes the `t` translator and pulls every bucket (`justNow`, `secondsAgo`, `minuteAgo`/`minutesAgo`, `hourAgo`/`hoursAgo`, `dayAgo`/`daysAgo`) out of the `memoryTree.status.*` namespace. The runtime `t()` does not interpolate, so `{count}` placeholders are substituted client-side via `.replace('{count}', String(n))`, matching the established `devices.lastSeenMinutes` / `team.memberCount` pattern. Added 8 new keys to `en.ts` plus all 13 chunk-1 files (en + 12 locale chunks, non-English entries seeded with the English copy). `pnpm i18n:check` reports `missing: 0 / extra: 0` for every locale.
- **Actionable 3 (fake timers in MemoryTreeStatusPanel test)** — fixed in `072f7d90`. `beforeEach` now pins `vi.useFakeTimers({ shouldAdvanceTime: true })` + `vi.setSystemTime(FIXED_NOW_MS)`, `payload()` derives `last_sync_ms` from the same constant, and `afterEach` restores real timers and mocks. `shouldAdvanceTime: true` is necessary so RTL's `waitFor` and the panel's polling loop keep making progress without per-test `advanceTimersByTime` calls; `setSystemTime` still freezes the wall clock for `formatRelativeMs`. All 6 existing tests pass.
- **Actionable 4 (`wiki_size_bytes` only walk `wiki/`)** — fixed in `65526ce1`. Picked Option A (scope walk to `content_root.join("wiki")`) so the wire field stays `wiki_size_bytes`, matching the UI tile label "Wiki size". Missing `wiki/` short-circuits to 0, so fresh installs are unaffected. Updated the rustdoc on `PipelineStatusResponse.wiki_size_bytes` to document the new scope.
- **Actionable 5 (redact walk-error path)** — fixed in `65526ce1`. Verified `walkdir::Error`'s `Display` impl writes `"IO error for operation on <abs-path>: ..."`, so `{err}` was a second path-leak vector beyond `err.path()`. Switched the log to `has_path={} kind={:?}` — boolean + `io::ErrorKind` only.

**Validation** (all from the worktree, against `upstream/main`):
- `cargo fmt --manifest-path Cargo.toml --check` — clean.
- `GGML_NATIVE=OFF cargo check --manifest-path Cargo.toml --lib` — pass (pre-existing warnings only).
- `GGML_NATIVE=OFF cargo test --lib memory_tree::tree::rpc::tests` — 15/15 pass (5 new + 10 existing).
- `pnpm exec prettier --check` over the changed TS files — clean.
- `pnpm exec eslint --max-warnings 0` over the changed TS files — clean.
- `pnpm typecheck` — clean.
- `pnpm i18n:check` — `missing: 0 / extra: 0` for all 12 non-English locales.
- `pnpm debug unit MemoryTreeStatusPanel` — 6/6 pass.

Tip is `65526ce1`. Force-with-lease push, no new PR.

app/src/utils/tauriCommands/memoryTree.ts
```diff
@@ -678,3 +680,134 @@ export async function memoryTreeBackfillStatus(): Promise<BackfillStatus> {
...
+ * Routes through `invoke('core_rpc_relay', ...)` rather than `callCoreRpc`:
+ * the status panel mounts during boot and polls before the realtime socket
+ * has connected, so the Tauri IPC relay avoids the raw-`fetch()` CORS
+ * preflight + early socket lookups that `callCoreRpc` falls through to.
+ * See `.claude/memory.md` ("Service RPC calls must use Tauri IPC") and
+ * AGENTS.md for the project-wide rule.
+ */
+export async function memoryTreePipelineStatus(): Promise<MemoryTreePipelineStatus> {
+  console.debug('[memory-tree-rpc] memoryTreePipelineStatus: entry');
+  const resp = await invoke<MemoryTreePipelineStatus | ResultEnvelope<MemoryTreePipelineStatus>>(
```
sanil-23: 🛑 **Blocker — `core_rpc_relay` is not a registered Tauri command, so this call rejects at runtime.**

`invoke('core_rpc_relay', …)` was removed from the shell in #84 and is **not** in `app/src-tauri/src/lib.rs`'s `generate_handler!` today (only `core_rpc_url` / `core_rpc_token` and friends are registered). No other code under `app/src` invokes it — the rest of the app, and all 12 sibling helpers in this very file, reach the core through `callCoreRpc`, which `fetch()`es the in-process HTTP server with the bearer token. As written, `memoryTreePipelineStatus()` will always reject (`Command core_rpc_relay not found`), so the status panel shows the error banner permanently and the toggle always toasts.

Notes:
- The cited `CLAUDE.md` / `.claude/memory.md` "always use `core_rpc_relay`" guidance is stale (the command no longer exists) — worth a separate docs fix, but the code must not rely on it.
- The justification comment is also inaccurate: `callCoreRpc` only uses a socket transport for iOS/tunnel/cloud profiles, never for the local desktop path — there are no "early socket lookups" to avoid here.
- The unit tests don't catch this because they mock `memoryTreePipelineStatus` at the wrapper boundary.

Suggested fix — use `callCoreRpc` like every sibling in this file:
```ts
const resp = await callCoreRpc<
  MemoryTreePipelineStatus | ResultEnvelope<MemoryTreePipelineStatus>
>({ method: 'openhuman.memory_tree_pipeline_status', params: {} });
const out = unwrapResult(resp);
```

app/src/utils/tauriCommands/memoryTree.ts
```diff
@@ -678,3 +680,134 @@ export async function memoryTreeBackfillStatus(): Promise<BackfillStatus> {
...
+ * Routes through `invoke('core_rpc_relay', ...)` for the same reason as
+ * `memoryTreePipelineStatus`: the toggle lives next to the status panel
+ * which mounts before the realtime socket connects, so the Tauri IPC
+ * relay path is the only one guaranteed to be ready.
+ */
+export async function memoryTreeSetEnabled(
+  enabled: boolean
+): Promise<MemoryTreeSetEnabledResponse> {
+  console.debug('[memory-tree-rpc] memoryTreeSetEnabled: entry enabled=%s', enabled);
+  const resp = await invoke<
```
sanil-23: 🛑 **Blocker (same root cause as `memoryTreePipelineStatus` above).** `invoke('core_rpc_relay', …)` targets a command that isn't registered in the shell, so this rejects at runtime and the auto-sync toggle always fails (the `onToast` error path fires on every click). Route through `callCoreRpc` like the siblings:
```ts
const resp = await callCoreRpc<
  MemoryTreeSetEnabledResponse | ResultEnvelope<MemoryTreeSetEnabledResponse>
>({ method: 'openhuman.memory_tree_set_enabled', params: { enabled } });
const out = unwrapResult(resp);
```

app/src/utils/tauriCommands/memoryTree.ts
```diff
@@ -678,3 +680,134 @@ export async function memoryTreeBackfillStatus(): Promise<BackfillStatus> {
...
+   * Optional human-readable reason. Present when `status` is `paused`
+   * (carries the gate mode) or `error` (carries the failed-job count);
+   * `null` otherwise.
+   */
+  reason: string | null;
+  /** Epoch ms of the most-recent chunk timestamp. Zero when the store is empty. */
+  last_sync_ms: number;
+  /** Total `mem_tree_chunks` rows across all sources. */
+  total_chunks: number;
+  /** Recursive on-disk size of the memory_tree content root, in bytes. */
```
sanil-23: 💡 Doc drift: the handler walks `content_root/wiki` (scoped to the `wiki/` sub-tree per the #1856 follow-up), not the whole content root. Match the comment to the behavior:
```ts
/** Recursive on-disk size of the `wiki/` sub-tree under the memory_tree content root, in bytes. */
wiki_size_bytes: number;
```

src/openhuman/memory/schema.rs
```diff
@@ -689,6 +699,112 @@ pub fn schemas(function: &str) -> ControllerSchema {
...
+                FieldSchema {
+                    name: "total_chunks",
+                    ty: TypeSchema::U64,
+                    comment: "Total rows in `mem_tree_chunks`.",
+                    required: true,
+                },
+                FieldSchema {
+                    name: "wiki_size_bytes",
+                    ty: TypeSchema::U64,
+                    comment: "Recursive on-disk size of the memory_tree content root. \
```
sanil-23: 💡 Same doc drift as the TS interface — `pipeline_status_rpc` walks `content_root/wiki`, not the whole content root. Suggest:
```rust
comment: "Recursive on-disk size of the `wiki/` sub-tree under the \
          memory_tree content root. 0 when the directory does not exist yet.",
```

src/openhuman/memory_tree/tree/rpc.rs
```diff
@@ -310,6 +310,352 @@ pub async fn backfill_status_rpc(
...
+            }
+        }
+    }
+    total
+}
+
+/// Pure derivation of `(status, reason)` from raw signals. Split out so the
+/// unit tests can exercise the precedence rules without spinning up a
+/// store.
+fn derive_pipeline_status(
```
sanil-23, line 527: 💡 Two non-blocking refinements to `derive_pipeline_status`:

1. **`ready` is never consulted.** A workspace with queued-but-not-yet-running jobs reports `idle` (no chunks) or `running` (chunks exist) even though work is pending. If "queued" should read as active, fold it in — e.g. treat `is_syncing || ready > 0` as `syncing`.
2. **`is_paused` is redundant with `mode`.** At the only call site `is_paused` equals `mode == SchedulerGateMode::Off`, so passing both invites drift. Consider dropping the `is_paused` parameter and deriving it inside the function from `mode` alone.

app/src/pages/Intelligence.tsx
```diff
@@ -185,45 +142,14 @@ export default function Intelligence() {
...
-                    subconscious/dreams tabs that still use it. */}
-              </div>
+              {/* Header status pills used to live here, sourced from
+                  `useConsciousItems` + `useMemoryIngestionStatus`. They
+                  were superseded by the `MemoryTreeStatusPanel` that
+                  renders inside `MemoryWorkspace` (#1856 Part 1).
+                  Analyze Now / Refresh button was already removed in an
+                  earlier refactor — the MemoryWorkspace fetches via
+                  memory_tree RPCs that don't need a manual trigger. */}
+              <div className="flex items-center gap-3" />
```
sanil-23: 🧹 Nitpick: this `<div className="flex items-center gap-3" />` is now empty (the status pills moved into `MemoryTreeStatusPanel`). With the parent `justify-between` and a single remaining sibling, it no longer affects layout — safe to delete the empty node (and trim the preceding placeholder comment).

sanil-23 (changes requested):
# PR #2719 — feat(memory-tree): status panel + on/off toggle (#1856 Part 1)

## Walkthrough
Adds a Memory Tree status panel (4 stat tiles + auto-sync toggle) mounted in `MemoryWorkspace`, backed by two new Rust RPCs — `memory_tree_pipeline_status` (read-side aggregate) and `memory_tree_set_enabled` (flips `scheduler_gate.mode` between `auto`/`off`). Removes the legacy `useConsciousItems`/`useMemoryIngestionStatus` header pills from `Intelligence.tsx`. The Rust side is clean, well-documented, runs all blocking I/O on `spawn_blocking`, and ships strong unit coverage; i18n parity is correct across all 13 locales.

**However, the panel cannot work in the desktop app as written:** both new TS wrappers call `invoke('core_rpc_relay', …)`, a Tauri command that no longer exists in the shell — so every fetch and every toggle rejects at runtime. This is a blocker. The unit tests don't catch it because they mock the two wrappers at the boundary.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/memory_tree/tree/rpc.rs` | New `pipeline_status_rpc` + `set_enabled_rpc` + pure `derive_pipeline_status` helper + `compute_dir_size_bytes`; 5 new tests. |
| `src/openhuman/memory/schema.rs` | Registers `pipeline_status` / `set_enabled` controllers (schema + handler) in the registry. |
| `app/src/utils/tauriCommands/memoryTree.ts` | `memoryTreePipelineStatus()` / `memoryTreeSetEnabled()` wrappers + wire types. |
| `app/src/components/intelligence/MemoryTreeStatusPanel.tsx` | New 4-tile panel + auto-sync toggle + adaptive polling hook. |
| `app/src/components/intelligence/MemoryTreeStatusPanel.test.tsx` | 6 Vitest cases (tiles, skeleton, toggle round-trip, paused, error+retry, toast). |
| `app/src/components/intelligence/MemoryWorkspace.tsx` | Mounts `<MemoryTreeStatusPanel>` above `<MemorySources>`. |
| `app/src/pages/Intelligence.tsx` | Removes legacy header-pill hooks + derivations. |
| `app/src/lib/i18n/en.ts` + 13 chunk files | New `memoryTree.status.*` keys (24 keys, parity verified). |

## Actionable comments (6)

### 🛑 Blockers

#### 1. `app/src/utils/tauriCommands/memoryTree.ts:751` & `:800` — RPC transport targets a non-existent Tauri command
Both `memoryTreePipelineStatus()` and `memoryTreeSetEnabled()` call `invoke('core_rpc_relay', …)`. That command was removed from the shell in #84 and is **not** present in `app/src-tauri/src/lib.rs`'s `generate_handler!` today — only `core_rpc_url` / `core_rpc_token` and friends are registered. No other code under `app/src` invokes `core_rpc_relay`; the entire app (and the 12 sibling helpers in this same file) reaches the in-process core through `callCoreRpc`, which `fetch()`es `http://127.0.0.1:<port>/rpc` with the bearer token.

Downstream effect: in a real desktop build the panel's polling always rejects (`Command core_rpc_relay not found`) → the error banner is permanent, and the toggle always toasts an error. The Vitest suite passes only because it mocks both wrappers.

**Suggested change** (apply to both functions):
```ts
const resp = await callCoreRpc<
  MemoryTreePipelineStatus | ResultEnvelope<MemoryTreePipelineStatus>
>({ method: 'openhuman.memory_tree_pipeline_status', params: {} });
const out = unwrapResult(resp);
```
Note: the `CLAUDE.md` / `.claude/memory.md` "always use `core_rpc_relay`" rule is stale (the command is gone) and the "early socket lookups" justification is inaccurate (`callCoreRpc` only uses a socket transport for iOS/tunnel/cloud profiles). Worth a follow-up to fix those docs.

### 💡 Refactor / suggestion

#### 2. `src/openhuman/memory_tree/tree/rpc.rs:527` — `derive_pipeline_status` ignores `ready`; `is_paused` param is redundant
Queued-but-not-running jobs (`ready > 0`, `running == 0`) report `idle`/`running`, never `syncing`. And `is_paused` is `mode == SchedulerGateMode::Off` at the only call site, so passing both args invites drift. See inline comment for the concrete shape.

#### 3. `schema.rs:745` + `memoryTree.ts:725` — `wiki_size_bytes` docs say "content root", impl walks `content_root/wiki`
The handler is correctly scoped to the `wiki/` sub-tree (matching the tile label), but the schema `comment` and the TS interface doc still say "memory_tree content root". Minor doc drift — concrete edits inline.

## Nitpicks (1)
- `app/src/pages/Intelligence.tsx:152` — leftover empty `<div className="flex items-center gap-3" />`; inert under `justify-between`, safe to delete.

## Verified / looks good
- **Rust RPC wiring** is correct: `pipeline_status` / `set_enabled` registered in both `all_controller_schemas()` and `all_registered_controllers()`; the in-sync invariant test still holds. Follows the controller-registry pattern (no ad-hoc `cli.rs`/`jsonrpc.rs` branches).
- **All blocking SQLite / filesystem I/O runs on `spawn_blocking`**; `compute_dir_size_bytes` uses `follow_links(false)`, `saturating_add`, and **redacts on-disk paths** in its warn log (good PII hygiene).
- **`set_enabled_rpc`** persists via `config.save()`, hot-reloads through `scheduler_gate::gate::update_config(...)`, and is idempotent (`changed=false` no-op skips the gate write).
- **Tests**: `derive_pipeline_status` precedence, empty/paused/ingested branches, toggle round-trip incl. no-op (Rust); tiles, skeleton, paused, error+retry, toggle, toast (Vitest).
- **`Intelligence.tsx` cleanup** leaves no dangling references to the removed hooks/derivations.
- **i18n**: 24 `memoryTree.status.*` keys present in `en.ts`, `en-1.ts`, and every non-English chunk-1 file (parity confirmed).

---
Blocking on #1 — the feature is non-functional in the desktop app until the transport is switched to `callCoreRpc`. The rest is solid.

graycyrus (commented):
# PR #2719 — Memory Tree Status Panel Review

## Status
CI is failing (Windows Rust tests) + there's a critical blocker preventing this feature from working in the desktop app.

## Critical Issue: Non-existent Tauri Command

Both new RPC wrappers call `invoke('core_rpc_relay', ...)` — but this command **does not exist** in the shell's Tauri handler registry (`app/src-tauri/src/lib.rs`). The command was removed in PR #84.

In production, this means:
- Every `memoryTreePipelineStatus()` call → rejects with "Command core_rpc_relay not found"
- Every `memoryTreeSetEnabled()` call → rejects with "Command core_rpc_relay not found"
- The status panel's error banner becomes permanent
- The toggle always fails with a toast error

Unit tests pass only because they mock these two wrappers at the boundary.

**Suggested fix:** Switch both wrappers to use `callCoreRpc()`, which is the pattern used by all other service-status RPC calls in the codebase:

```ts
export async function memoryTreePipelineStatus(): Promise<MemoryTreePipelineStatus> {
  const resp = await callCoreRpc<
    MemoryTreePipelineStatus | ResultEnvelope<MemoryTreePipelineStatus>
  >({ method: 'openhuman.memory_tree_pipeline_status', params: {} });
  return unwrapResult(resp);
}

export async function memoryTreeSetEnabled(
  enabled: boolean
): Promise<MemoryTreeSetEnabledResponse> {
  const resp = await callCoreRpc<
    MemoryTreeSetEnabledResponse | ResultEnvelope<MemoryTreeSetEnabledResponse>
  >({ method: 'openhuman.memory_tree_set_enabled', params: { enabled } });
  return unwrapResult(resp);
}
```

Note: the comment in the current code suggesting `core_rpc_relay` is "the only path guaranteed to be ready" is incorrect — that command doesn't exist in the current shell.

## CI Failure
Windows Rust tests show 4 compilation failures. This needs to be resolved before merging.

---

The Rust side (RPC handlers, schema registration, tests) is solid. The issue is purely on the frontend transport layer. Once the `core_rpc_relay` → `callCoreRpc` switch is made and CI passes, this is ready to go.

graycyrus (commented):
## Review Status

I see that the latest commit (cd4e94fce) has addressed all the **blocker issues** that @sanil-23 flagged:

1. ✓ Fixed `core_rpc_relay` → switched both `memoryTreePipelineStatus()` and `memoryTreeSetEnabled()` to use `callCoreRpc`
2. ✓ Updated `wiki_size_bytes` doc comments (both TS interface and Rust schema) to reflect the `wiki/` sub-tree scope
3. ✓ Removed the empty `<div>` from `Intelligence.tsx`
4. ✓ Removed unused `invoke` import from memoryTree.ts
5. ✓ Cleaned up stale comments about Tauri IPC relay justifications

**Remaining item**: @sanil-23's non-blocking suggestion about `derive_pipeline_status` (the `ready` parameter and `is_paused` redundancy) is still open. These are refinements, not blockers — deferring to @sanil-23's call on whether those optimizations should land in this PR or a follow-up.

**CI Status**: macOS E2E tests still pending. Once those clear + @sanil-23 approves, this is ready to merge.

graycyrus:
thanks, looks good merging this now

PR #2603, opened 2026-05-25
feat(mcp-registry): InstalledServer HTTP-remote transport

justinhsu1477:
Closes the named follow-up TODO from #2559:
> \`InstalledServer\` HTTP-remote transport variant (most Smithery servers are HTTP-remote; current install path is stdio-only)

## The gap this fixes

Before this PR, the setup-agent install path (\`mcp_setup_test_connection\` + \`mcp_setup_install_and_connect\`) hard-filtered registry detail responses to \`c.r#type == \"stdio\"\`. Smithery — which hosts the bulk of the official MCP ecosystem — serves **~99% of its listings as HTTP-remote**. UX was:

> User browses Smithery → picks a server → clicks install → almost every attempt fails with \"no stdio connection.\"

\`McpHttpClient\` (Streamable HTTP + OAuth + SSE per the MCP spec) already exists in \`mcp_client\` and is wired for static-config servers. This PR extends the *install / persistence / connect* plumbing in \`mcp_registry\` to dispatch through it.

## What changed

| File | Change |
| --- | --- |
| **\`types.rs\`** | New \`Transport::{Stdio, HttpRemote { url }}\` enum with \`dispatch_kind\` strings (\`\"stdio\"\` / \`\"http_remote\"\`) and a \`parse(kind, url)\` inverse that defaults unknown / missing values to \`Stdio\` (migration safety hatch). \`InstalledServer\` grows a \`transport\` field, \`#[serde(default)]\`-backed so legacy serialised payloads still load. |
| **\`store.rs\`** | \`mcp_servers\` table grows two additive columns (\`transport\`, \`deployment_url\`) via post-\`CREATE TABLE\` \`ALTER TABLE ADD COLUMN\`, gated by a \`PRAGMA table_info\` lookup so the migration is idempotent across launches. Insert + list + get carry the new columns. Row mapper falls back to \`Transport::Stdio\` when \`transport\` is missing. |
| **\`setup_ops.rs\`** | New \`pick_connection\` helper: **published stdio → any stdio → published http_remote → any http_remote → None**. Stdio always wins when offered, so no behavior regression for any server that already installs today. Both setup ops branch on the picked transport and dispatch to either \`McpStdioClient\` or \`McpHttpClient\`. \`ConnectionKind\` trait normalises the registry-side \`\"http\"\` / \`\"sse\"\` strings into the persisted \`\"http_remote\"\` dispatch kind. |
[...]

PR #2720, opened 2026-05-27
fix(vault): sync writes to memory_tree, not legacy UnifiedMemory (#2705)

### LINKED ISSUE #2705: v0.54.0 — Vault sync UI shows "synced" but mem_tree_chunks and mem_tree_ingested_sources stay at 0
**Version:** v0.54.0 (Beta build)
**Platform:** macOS
**User UID:** 6a146e7aefa23d1dbca7b440

## Issue
After "Sync" on a Knowledge vault, UI reports "N file(s) · synced X minutes ago"  but the underlying database remains empty.

## Steps to reproduce
1. Add Knowledge vault pointing to a local folder with ~900 markdown/docx files
2. Click "Sync"
3. UI shows progress and finally "synced"
4. Check the database:
   sqlite3 ~/.openhuman/users/<UID>/workspace/memory_tree/chunks.db \
     "SELECT COUNT(*) FROM mem_tree_ingested_sources; \
      SELECT COUNT(*) FROM mem_tree_chunks;"
5. Both queries return 0

## Expected
mem_tree_ingested_sources should contain 1 row per vault
mem_tree_chunks should contain thousands of rows for ~1000 files

## Actual
Both tables remain empty after multiple sync attempts. "Memory wiped" message confirms 0 rows removed = nothing was ever stored. "Build summary trees" button is cosmetic (no jobs queued).

## Additional context
- Heartbeat loop: ON
- Subconscious inference: ON
- Memory tree workers: ON (4 workers polling every 5s, route to Anthropic claude-haiku)
- AI provider: Anthropic with own API key (verified $30 credit available)
- Tried: wipe memory + remove vaults + re-add vaults + re-sync. Same result.

Happy to provide more logs/diagnostics.

justinhsu1477:
Closes #2705.

## Root cause

`vault::sync::sync_vault` walked the directory, called `doc_ingest` → \`MemoryClient::ingest_doc\` → \`UnifiedMemory::ingest_document\` → \`memory_docs\` table. The UI's \"synced\" message was technically correct (UnifiedMemory accepted the writes), but every modern retrieval surface — \`memory.search\`, \`tree.read_chunk\`, \`tree.browse\`, the agent's recall path, the summary-tree builder — reads from the memory-tree backend (\`mem_tree_chunks\` + \`mem_tree_ingested_sources\`).

Vault data was invisible to all of them.

That explains the entire bug report:

| Symptom | Explanation |
| --- | --- |
| \`mem_tree_chunks\` = 0 | vault sync never wrote there |
| \`mem_tree_ingested_sources\` = 0 | same |
| \"Memory wiped: 0 rows removed\" | wipe clears memory_tree, which was already empty |
| \"Build summary trees\" produces no jobs | memory_tree has no chunks to summarise |
| UI \"N file(s) · synced\" | vault state correctly reported the UnifiedMemory ingest succeeded — to the wrong table |
| Agent can't recall vault content | agent retrieval reads memory_tree |

Recent PRs (#2585, #2556, #2574) migrated RAG primitives to memory_tree as the canonical layer; the vault sync path was not migrated alongside.

## Fix

\`process_file\` now calls \`memory::ingest_pipeline::ingest_document\` directly with a stable \`source_id = vault:{vault_id}:{rel_path}\`. The pipeline writes to \`mem_tree_chunks\` + \`mem_tree_ingested_sources\` — the tables the modern retrieval stack reads from.

Three design choices worth calling out:

1. **Content-update path** — the pipeline's \`already_ingested\` gate is content-blind and the source_id is stable per file path. For real content updates the vault layer drops prior chunks via \`memory_store::chunks::store::delete_chunks_by_source\` *before* the re-ingest, otherwise new content gets short-circuited. The vault ledger's \`content_hash\` check still gates whether we run delete+reingest at all, so untouched files cost zero pipeline work.

2. **Deletion path (Phase 4)** — \`by_path\` entries the walk didn't see now call \`delete_chunks_by_source\` instead of \`doc_delete\`. Migration-safe: ledger entries whose stored \`document_id\` doesn't start with \`vault:\` (rows persisted before this fix) fall back to a recomputed source_id.

3. **Vault ledger semantic** — the \`VaultFile.document_id\` field now holds the memory-tree source_id. Schema column name unchanged for backward compatibility with persisted rows; only the semantic of what we store changed. Deletion uses the prefix-check above to handle the migration window cleanly.

## Tests

Three new regression tests in \`vault::sync::sync_tests\`:

| Test | What it pins |
| --- | --- |
| \`sync_writes_to_memory_tree\` | **The #2705 regression.** Creates a vault with two .md files, runs \`sync_vault\`, asserts \`count_chunks\` goes up + both source_ids appear in \`mem_tree_ingested_sources\` + ledger \`document_id\` starts with \`vault:\` |
| \`second_sync_with_no_changes_is_idempotent\` | Re-sync with unchanged content does not duplicate chunks (vault-layer hash dedup guards the pipeline) |
| \`vault_source_id_is_stable_and_namespaced\` | Unit test on the id format — defends against an accidental rename breaking cross-file isolation |

- [x] \`cargo test --lib vault\` — **28/28 pass**.
- [x] \`cargo check --lib\` — clean.
- [x] \`cargo fmt --check\` — clean.
- [x] \`cargo test --tests --no-run\` — clean (lesson from #2603 — integration-test target must compile).

## Out of scope (separate audits / PRs)

- Composio providers and \`agent_experience\` still call \`doc_ingest\` → UnifiedMemory. If they have the same gap, that's a separate audit; this PR is scoped to the vault path the user reported.
- Removing \`UnifiedMemory\` entirely is the larger follow-up senamakel listed on #2585; out of scope here.

Refs #2705.

## Summary by CodeRabbit

* **Improvements**
  * Vault sync now writes to the new memory-tree backend with stable per-file source identifiers, improving ingestion correctness and idempotency.
  * Sync skips redundant work for unchanged files and logs desyncs when expected ingestions are conflicted.
  * Vault removal now performs canonical memory-tree cleanup first and includes the count of deleted memory-tree chunks in the response; legacy cleanup failures no longer abort removal.

* **Tests**
  * Added/expanded regression and e2e tests covering memory-tree ingestion, idempotent re-syncs, and purge semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2720?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


YellowSnnowmann:
[justinhsu1477](https://github.com/justinhsu1477) could you please fix the tests and then push?

justinhsu1477:
@YellowSnnowmann thanks — pushed \`f8b50543\` about 5 min before your ping.

The failing test (\`memory_ingestion_status_reflects_initialized_client_snapshot\`) was asserting \`queue_depth == 1\` against the process-global \`IngestionState\`. Sibling tests in the same module call \`spawn_manual_sync\`, which detaches a \`tokio::spawn(...)\` that runs \`composio::run_connection_sync\` → \`MemoryClient::store_skill_sync\` → \`IngestionQueue::submit\` → \`state.enqueue()\`. That background task can still be draining when the snapshot test acquires \`GLOBAL_MEMORY_TEST_LOCK\`, so the global counter isn't 0.

Fix is to snapshot the baseline and assert the delta — pre-existing flake on \`upstream/main\`, the added vault sync tests just shifted the parallel-test ordering enough to expose it. Full reasoning in the commit body.

\`cargo test --lib -- memory::ops::sync vault::sync\` passes 9/9 locally; CI re-running now.

src/openhuman/memory/ops/sync.rs
```diff
@@ -367,6 +367,15 @@ mod tests {
...
 
+        // The other `memory::ops::sync` tests (`memory_sync_channel_*`,
+        // `memory_sync_all_*`) call `spawn_manual_sync`, which detaches a
+        // `tokio::spawn` background task that may still be in flight — and
+        // may have enqueued an ingestion job — by the time this test
+        // acquires `GLOBAL_MEMORY_TEST_LOCK`. The global `IngestionState`
+        // singleton therefore can't be assumed to start at queue_depth=0.
+        // Capture the baseline and assert the delta instead.
+        let baseline_depth = state.snapshot().queue_depth;
+
```
sanil-23, line 372: 🛑 **Blocker — unresolved merge conflict against `main`.** Merging this branch with current `main` leaves git conflict markers around this line, so the merged tree won't compile. Both sides add the *same* `let baseline_depth = state.snapshot().queue_depth;` de-flake (`main` landed an equivalent fix independently) — just with different comments. Resolve by keeping one `baseline_depth` capture and dropping the duplicate + markers before this can merge.

src/openhuman/vault/sync.rs
```diff
@@ -136,45 +174,103 @@ async fn process_file(file: FileToProcess) -> IngestFileResult {
...
+            ..
+        }) => {
+            log::debug!(
+                "[vault] sync: ingested path={} source_id={} chunks_written={} already_ingested={}",
+                file.rel_path,
+                source_id,
+                chunks_written,
+                already_ingested,
+            );
+            IngestFileResult::Ingested {
```
sanil-23, line 286: **Nitpick — false-success edge case.** An `ingest_document` returning `Ok(IngestResult { already_ingested: true, chunks_written: 0, .. })` still lands in this `Ingested` arm and writes a "success" ledger row. The delete-first guard above prevents this on the normal update path, but on a ledger↔memory-tree desync (ledger wiped while the `mem_tree_ingested_sources` row survives) it would report success while nothing reaches retrieval — the exact false-success mode this PR exists to kill. Consider a `log::warn!` (or a distinct result) when `chunks_written == 0 && already_ingested`.

src/openhuman/vault/sync.rs
```diff
@@ -371,8 +465,15 @@ pub async fn sync_vault(config: &Config, vault: &Vault) -> VaultSyncReport {
     });
 
     // ── Phase 2: Concurrent ingestion ────────────────────────────────────────
+    //
+    // Each task needs a `Config` clone because the memory-tree pipeline uses
+    // it for the `with_connection` path and for `delete_chunks_by_source`'s
+    // pre-ingest cleanup on content updates. Cloning per-task keeps the
+    // borrow life-cycle simple inside `buffer_unordered` (no `&'a Config`
+    // bouncing through closures).
+    let config_for_workers = config.clone();
```
sanil-23: **Nitpick — per-file `Config` clone.** `config.clone()` runs once per candidate file as the stream is polled. `buffer_unordered` keeps only ~4 alive at a time, so memory is bounded, but a large vault still performs N full `Config` clones. Wrapping in `Arc<Config>` and cloning the `Arc` would make this O(1)-per-file. Minor.

src/openhuman/vault/sync.rs
```diff
@@ -424,21 +525,67 @@ pub async fn sync_vault(config: &Config, vault: &Vault) -> VaultSyncReport {
...
-        if let Err(err) = doc_delete(DeleteDocParams {
-            namespace: vault.namespace.clone(),
-            document_id: prev.document_id.clone(),
+
+        // Prefer the ledger's stored id (already in `vault:{id}:{rel_path}`
+        // form for post-#2705 entries); fall back to recomputing it for
+        // ledger rows that pre-date the migration and still carry a legacy
+        // UnifiedMemory document_id.
+        let stored_id = prev.document_id.clone();
+        let source_id = if stored_id.starts_with("vault:") {
```
sanil-23: **Question — legacy doc_id fallback is a no-op for genuinely-legacy rows.** For ledger rows that pre-date #2705, the stored `document_id` is a UnifiedMemory id whose data lived in `memory_docs`, never in memory_tree. Recomputing a `vault:{id}:{path}` source_id and calling `delete_chunks_by_source` therefore deletes nothing, and the old UnifiedMemory doc is never cleaned up when such a file is removed. Intentional (deferred to the UnifiedMemory-removal follow-up), or should the fallback branch also `doc_delete(prev.document_id)` during the migration window so legacy rows don't leak?

sanil-23 (changes requested):
# PR #2720 — fix(vault): sync writes to memory_tree, not legacy UnifiedMemory (#2705)

## Walkthrough

This PR migrates the vault sync path off the legacy UnifiedMemory / `memory_docs` backend onto the memory-tree pipeline (`mem_tree_chunks` + `mem_tree_ingested_sources`), which is what every modern retrieval surface actually reads from. `process_file` now ingests via `ingest_pipeline::ingest_document` against a stable `source_id = vault:{vault_id}:{rel_path}`, deletes prior chunks before re-ingesting on content updates (to defeat the content-blind `already_ingested` gate), and routes vanished-file deletions through `delete_chunks_by_source`. The diagnosis and the core sync fix are sound and well-tested.

**However, this PR cannot merge as-is.** Two issues block it: an **unresolved merge conflict** against current `main`, and a **stale legacy-backend call in the vault-removal/purge path** that this migration leaves orphaning data — the same silent-failure class the PR set out to fix.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/vault/sync.rs` | Core fix: ingest via memory-tree pipeline with stable `vault:{id}:{rel}` source ids; delete-then-reingest on content change; Phase-4 deletions via `delete_chunks_by_source`; migration-safe legacy doc_id fallback; 3 new unit tests. |
| `tests/vault_sync_e2e.rs` | Updated e2e to assert against `mem_tree_chunks` / `mem_tree_ingested_sources` instead of `list_documents`; drops the unused `documents_from_payload` helper. |
| `src/openhuman/memory/ops/sync.rs` | Test-only: capture `queue_depth` baseline and assert delta (de-flake). **Currently carries unresolved conflict markers vs `main`.** |

## Actionable comments (2)

### 🛑 Blockers

#### 1. `src/openhuman/memory/ops/sync.rs:370-387` — Unresolved merge conflict against `main`

Merging this branch with the current `main` leaves git conflict markers in this file, so the merged tree **does not compile**. Both sides introduce the *same* fix — `let baseline_depth = state.snapshot().queue_depth;` — with different explanatory comments; `main` independently landed an equivalent de-flake. Trivial to resolve (keep one capture, drop the duplicate and the markers), but it **must be resolved before this PR can merge**. See the inline comment for the suggested resolution.

### ⚠️ Major

#### 2. `src/openhuman/vault/ops.rs:116-136` — `vault_remove(purge_memory=true)` still purges the *legacy* backend, orphaning all memory-tree data

Not in the diff, but a direct downstream consequence of this migration. Vault content now lives in `mem_tree_chunks` / `mem_tree_ingested_sources` (keyed `vault:{id}:{rel_path}`), yet the purge path still calls `clear_namespace(v.namespace)` → `MemoryClient::clear_namespace`, which only touches the legacy `vectors` / `memory_docs` tables (`memory_store/vectors/store.rs:386`) — tables this PR no longer writes to.

Result: **removing a vault with purge now deletes nothing meaningful and orphans every `mem_tree_chunks` row**, so retrieval keeps surfacing content from a deleted vault. That is the exact silent-failure class this PR fixes, re-appearing on the removal side. The in-sync Phase-4 deletion only covers per-file vanishing, not whole-vault removal.

The codebase already has the helper and an established pattern for this (`composio/ops.rs:560`, `channels/controllers/ops.rs:565`):

```rust
use crate::openhuman::memory_store::chunks::store::delete_chunks_by_source_prefix;
use crate::openhuman::memory_store::chunks::types::SourceKind;

let cfg = config.clone();
let prefix = format!("vault:{id}:");
let purged = tokio::task::spawn_blocking(move || {
    delete_chunks_by_source_prefix(&cfg, SourceKind::Document, &prefix)
})
.await;
// (optionally also keep clear_namespace for migration-window legacy rows)
```

At minimum, acknowledge this in the PR body's "Out of scope" section if you intend to defer it.

## Nitpicks (3)

- `src/openhuman/vault/sync.rs:250-270` — A successful `ingest_document` returning `already_ingested: true` with `chunks_written == 0` is still counted as `Ingested` and writes a "success" ledger row. The delete-first guard prevents this on the normal path, but a ledger↔memory-tree desync would re-introduce false-success reporting. Consider a `log::warn!` when `chunks_written == 0 && already_ingested`. (inline)
- `src/openhuman/vault/sync.rs:474-476` — `Config` is cloned once per candidate file; `Arc<Config>` would avoid N copies on large vaults. (inline)
- `tests/vault_sync_e2e.rs:117-145` vs `src/openhuman/vault/sync.rs:748-766` — inconsistent `spawn_blocking` convention between the e2e and unit tests for the same `is_source_ingested` / `count_chunks` calls; pick one. `sync_writes_to_memory_tree` could also tighten `chunks_after > chunks_before` to `chunks_before == 0` (fresh tempdir).

## Questions for the author (2)

- `src/openhuman/vault/sync.rs:543-552` — for genuinely pre-#2705 ledger rows the recomputed memory-tree source_id has no chunks (data was in `memory_docs`), so `delete_chunks_by_source` is a no-op and the legacy doc leaks on deletion. Deferred, or should the fallback also `doc_delete` the legacy id? (inline)
- `src/openhuman/vault/types.rs:56` — is `Vault.namespace` still meaningful post-migration? Its only remaining consumer is the (now-misdirected) purge call; fixing #2 by prefix-delete may make it vestigial.

## Verified / looks good

- `delete_chunks_by_source` clears **both** `mem_tree_chunks` and `mem_tree_ingested_sources` (`memory_store/chunks/store.rs:809-832`), so the delete-then-reingest content-update path genuinely resets the `already_ingested` gate. The central design assumption holds.
- Error handling is fail-closed: a failed pre-reingest delete returns `Failed` rather than re-ingesting alongside stale chunks; the next sync self-heals.
- `delete_chunks_by_source` and the Phase-4 deletion correctly run under `spawn_blocking` (SQLite is sync I/O).
- `sync_writes_to_memory_tree` precisely pins the #2705 regression (chunk count up + both source ids registered + ledger stores the `vault:` source id); the e2e exercises the full update / delete / add lifecycle.

graycyrus (dismissed):
**[APPROVED]** 

This is a solid fix for #2705. Vault sync now correctly routes through the memory-tree pipeline (`mem_tree_chunks` + `mem_tree_ingested_sources`) instead of the legacy UnifiedMemory backend. The migration is clean: stable per-file source_ids ensure idempotency, content updates properly delete stale chunks before re-ingesting, and pre-fix ledger rows are handled gracefully.

Tests are comprehensive — three unit tests pin the regression and the invariants, plus e2e validation of the full lifecycle. Code patterns are solid (async/blocking correctly separated, error context preserved, logging appropriate).

No issues. Ship it.

justinhsu1477:
@sanil-23 thanks for the thorough pass — all four addressed in \`e7c61b14\` (rebased on \`upstream/main\` so the \#2737 conflict is gone).

**🛑 L378 conflict** — Resolved by rebase: dropped my standalone \`baseline_depth\` commit since \#2737 already landed the same de-flake with a cleaner \`state.reset_for_test()\` helper. Kept that, dropped mine.

**L263 false-success** — Added a \`log::warn!\` for the \`already_ingested == true && chunks_written == 0\` branch. The delete-first guard prevents this on the normal update path, so seeing it means ledger ↔ memory_tree desync — exactly the silent mode this PR exists to kill. Log line names the source_id + suggests a manual \`delete_chunks_by_source\` resync.

**L474 per-file Config clone** — Switched to \`Arc<Config>\`. \`process_file\` now takes \`Arc<Config>\`, the call site does one outer \`Arc::new(config.clone())\` then \`Arc::clone(&...)\` per candidate. A 5k-file backfill pays one deep clone + N ref-count bumps instead of N full clones.

**L544 legacy fallback no-op** — You're right, it was silently leaving legacy \`memory_docs\` rows behind. Migration-window fix: when \`document_id\` doesn't start with \`vault:\` (= pre-#2705 ledger row), the deletion path now runs \`doc_delete(prev.document_id)\` against UnifiedMemory **in parallel** with \`delete_chunks_by_source\` against the recomputed memory_tree source_id. \`doc_delete\` failures are best-effort (a 404 on the legacy path shouldn't block the canonical memory_tree cleanup).

\`cargo test --lib -- memory::ops::sync vault::sync\` → 9/9 pass.
\`cargo check --lib\` + \`cargo fmt --check\` + \`cargo test --tests --no-run\` all clean.

Pushed as force-with-lease since the rebase was unavoidable.

graycyrus:
I see @sanil-23 has requested changes — deferring to their feedback. Will review once those are addressed.

graycyrus:
@justinhsu1477 unresolved review feedback — please address before we review.

graycyrus:
Unresolved review feedback from sanil-23 — please address before we review.

src/openhuman/memory/ops/sync.rs
justinhsu1477: Resolved by rebase onto current `upstream/main` in `e7c61b14` — dropped my standalone `baseline_depth` commit since #2737 landed an equivalent (and cleaner) fix via `state.reset_for_test()`. The conflict is gone; no markers remain. ✅

src/openhuman/vault/sync.rs
justinhsu1477: Addressed in `e7c61b14` (`vault/sync.rs:L268-281`). The `already_ingested && chunks_written == 0` arm now emits a `log::warn!` naming the source_id + path and suggesting a manual `delete_chunks_by_source` resync — exactly the ledger↔memory_tree desync mode you flagged. ✅

justinhsu1477:
@graycyrus all four of @sanil-23's items were fully addressed in \`e7c61b14\` ~8 hours ago — the inline threads still show "unresolved" because I left a top-level summary instead of replying inline. Apologies for the noise; verifying-by-grep so you don't have to:

| # | sanil-23's point | Fix in \`e7c61b14\` |
| --- | --- | --- |
| 🛑 L372 merge conflict | Resolved by rebase onto current \`main\` (\#2737's \`state.reset_for_test()\` is now upstream; my conflicting commit dropped) | \`git log --oneline upstream/main..HEAD\` |
| L286 false-success edge | \`already_ingested && chunks_written == 0\` now emits \`log::warn!\` naming the source_id + suggesting a manual \`delete_chunks_by_source\` resync | \`vault/sync.rs:L268-281\` |
| Per-file \`Config\` clone | \`process_file\` takes \`Arc<Config>\`; 5k-file backfill pays one deep clone + N atomic refs | \`vault/sync.rs:L165, L196, L496-498\` |
| Legacy doc_id fallback no-op | \`is_legacy_ledger_row = !stored_id.starts_with("vault:")\` triggers a parallel \`doc_delete(prev.document_id)\` against UnifiedMemory alongside the canonical \`delete_chunks_by_source\` | \`vault/sync.rs:L43, L574, L589-606\` |

Verification:
- \`cargo test --lib -- vault::sync memory::ops::sync\` → 9/9 pass
- \`cargo check --lib\` + \`cargo fmt --check\` + \`cargo test --tests --no-run\` clean
- CI: **25 ok / 0 fail / 1 running** at last snapshot (CodeRabbit re-APPROVED at 11:12)

Ready for re-review whenever you have cycles.

sanil-23:
@justinhsu1477 followed up on the **major** item from sanil-23's top-level walkthrough (`vault/ops.rs:116-136` — `vault_remove(purge_memory=true)` was still routing to the legacy backend) — fix pushed as `6f4324ea` on top of `e7c61b14`.

**What changed**
- `vault_remove` now runs `delete_chunks_by_source_prefix(config, SourceKind::Document, "vault:{id}:")` (mirrors the `channels::clear_channel_memory` pattern at `channels/controllers/ops.rs:563`) so vault content is purged from `mem_tree_chunks` + `mem_tree_ingested_sources` — the canonical retrieval surface — instead of being orphaned. The legacy `clear_namespace` call stays as best-effort cleanup for any pre-#2705 ledger rows still in `memory_docs` during the migration window. RPC response now surfaces `memory_tree_chunks_deleted` so the UI can show the real count.
- Drive-by: hoisted `Arc::new(config.clone())` out of `sync.rs` Phase 4's deletion loop — the per-iteration clone had quietly undone part of the Phase 2 worker-pool optimisation flagged earlier.

**Tests**
- `vault_remove_with_purge_clears_memory_tree` — pins the regression: post-sync, `vault_remove(purge_memory=true)` drops the source from `mem_tree_ingested_sources` and shrinks `count_chunks`.
- `vault_remove_without_purge_leaves_memory_tree_intact` — guards the boolean from silently flipping to always-purge.

**Verification**
- `cargo test --lib vault::` → **27/27 pass** (including 2 new tests).
- `cargo test --test vault_sync_e2e` → 1/1 pass.
- `cargo test --lib memory::` → 365/365 pass (smoke check).
- `cargo check --manifest-path Cargo.toml` + `cargo check --manifest-path app/src-tauri/Cargo.toml` clean.
- `cargo fmt --check` clean.

Re. the namespace question from the walkthrough ("is `Vault.namespace` still meaningful post-migration?") — left intact because it's still load-bearing for the best-effort legacy `clear_namespace` cleanup during the migration window. Removing it can ride the larger UnifiedMemory removal follow-up (#2585).

sanil-23 (approved):
Vault sync routes writes to memory_tree (not legacy UnifiedMemory). Purge-memory-tree integration commit (6f4324ea) addresses the heaviest comment thread on this PR. vault:: 27/27 (2 new tests), vault_sync_e2e 1/1, memory:: smoke 365/365, cargo fmt/check clean. 25/0 CI all-green. LGTM.

sanil-23:
@coderabbitai review

PR #2758, opened 2026-05-27
fix(auth): keep a valid session across the first-login restart (corroborate before destructive logout)

sanil-23:
## Summary

- After the first-login identity-flip restart, a token-gated RPC can momentarily return `session jwt required` / `no backend session token` **before the core has loaded the on-disk auth profile**. `classifyRpcError` mapped that to `auth_expired`, which drove the **destructive** `clearSession` (`auth_clear_session` removes the profile from disk) → a forced re-login even though the token survived the restart.
- Split `auth_expired` into **confirmed** (real 401 / `Session expired` / backend-path 401) vs **unconfirmed** (`session jwt required` / `no backend session token`), plumbed through the `core-rpc-auth-expired` event.
- On an **unconfirmed** signal, `CoreStateProvider` now **corroborates** the on-disk token via the cheap disk-only `auth_get_session_token` RPC (no `auth/me` network, not subject to the snapshot's 5s/10s timeouts) with a short retry, and only signs out if the token is genuinely gone — biasing toward *keeping* the session on anything inconclusive.
- **Confirmed** expiries (real 401) and the socket `auth:session_expired` push still clear immediately — no behavior change on genuine expiry.
- Adds diagnostic logging at every reauth decision point so the path is traceable in a restart log.

## Problem

Users hit a "restart asks me to log in again" after their first login. The identity-flip restart (anonymous → authenticated, to re-hydrate redux-persist + CEF under the user namespace) is expected, and the session token **does** persist on disk across it (`Session token found — auto-connecting` in core logs). But on `boot2`, a token-gated RPC that races ahead of the auth-profile load gets `session jwt required` from the core — meaning "no token *loaded* yet," not "token invalid." The frontend treated that as a confirmed expiry and ran the destructive `clearSession`, which deletes the persisted profile (`credentials/ops.rs` `clear_session` → `remove_profile`) — turning a transient boot-load race into a permanent logout. This is distinct from (and not caused by) the staging `auth/me` network flakiness, which is correctly classified as `transport`/`timeout` and never clears the session.

## Solution

`clearSession` is irreversible, so it must not fire on a "token not loaded yet" signal:

- `classifyAuthExpiredReason(message, httpStatus)` distinguishes a real server rejection from a "no JWT loaded" signal; unknown messages default to `unconfirmed` (safe: verify, don't destroy).
- `runReauth` is now reason-aware. `confirmed` → clear immediately (unchanged). `unconfirmed` → `confirmSessionTokenGone()` reads `auth_get_session_token` up to 3× (300ms apart) to ride out the boot lock window; clears **only** if every read is empty, and **keeps** the session if the token is present or the read is inconclusive.
- The debounce slot is claimed before the async corroboration so event bursts can't double-run.

Why `auth_get_session_token` and not `app_state_snapshot`: the snapshot blocks up to 5s on `auth/me` + 10s on the runtime build and rides the flaky network; the token RPC is a pure disk read of the same profile, ms-fast and network-free.

## Submission Checklist

- [x] Tests added or updated — `classifyAuthExpiredReason` table (confirmed/unconfirmed/default) + `CoreStateProvider` guard tests (unconfirmed + token-present → no logout; unconfirmed + token-gone → logout); genuine-expiry tests updated to `reason: 'confirmed'`. 120 tests pass across both suites.
- [x] **Diff coverage ≥ 80%** — all changed FE lines are exercised by the new/updated unit tests (`coreRpcClient.test.ts`, `CoreStateProvider.test.tsx`). The merged `diff-cover` gate runs in CI; local cargo-llvm-cov is N/A (no Rust changed).
- [x] Coverage matrix updated — `N/A: behaviour-only bugfix on an existing auth path, no feature row added/removed/renamed`.
- [x] Affected feature IDs listed under `## Related` — `N/A: no matrix feature touched`.
- [x] No new external network dependencies introduced — none; the guard reads an existing local RPC.
- [x] Manual smoke checklist updated — `N/A: behaviour-preserving guard on the existing login/restart path; the existing login smoke covers it`.
- [x] Linked issue closed via `Closes #NNN` — `N/A: discovered via restart-log analysis, no tracked issue`.

## Impact

- **Platform:** desktop. Prevents a spurious, destructive logout after the first-login restart; users stay authenticated as the on-disk token already allows.
- **Security:** genuine expiry is unchanged — a real 401, explicit `Session expired`, and the socket session-expired push all still sign the user out immediately. The guard only refuses to destroy a session on an *unconfirmed* "no token loaded yet" signal, and never weakens a confirmed expiry.
- **Performance:** corroboration is a disk-only RPC with ≤3 short retries, only on the unconfirmed path.

## Related

- Closes: _N/A (no tracked issue)_
- Follow-up PR(s)/TODOs: separate snapshot fix — the `build_runtime_snapshot` `service` arm lacks a per-op timeout and its `launchctl` call can wedge the 10s parent; plus pooling the per-call `auth/me` client. Tracked for a follow-up PR.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/auth-restart-relogin-guard`
- Commit SHA: `e629f958`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` (changed files)
- [x] `pnpm typecheck` (full app — green)
- [x] Focused tests: `coreRpcClient.test.ts` + `CoreStateProvider.test.tsx` — 120 passed
- [x] Rust fmt/check (if changed): `N/A — no Rust changed`
- [x] Tauri fmt/check (if changed): `N/A — no Rust changed`

### Validation Blocked
- `command:` pre-push hook (`pnpm rust:check` + `lint:commands-tokens`)
- `error:` `cargo` not on the hook PATH; `ripgrep` not installed locally
- `impact:` none on this FE-only change (touches no Rust and nothing under `src/components/commands/`); pushed with `--no-verify`. CI runs the full gates.

### Behavior Changes
- Intended behavior change: an `unconfirmed` auth-expired signal no longer destroys the session without corroboration.
- User-visible effect: no more spurious re-login after the first-login restart; genuine expiry still signs out.

### Parity Contract
- Legacy behavior preserved: `confirmed` (real 401 / `Session expired` / backend-path 401) and the socket session-expired push clear immediately, as before.
- Guard/fallback/dispatch parity checks: missing `reason` defaults to `unconfirmed` (corroborate); inconclusive token read keeps the session (bias to not-destroy).

- Duplicate PR(s): N/A
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Smarter session-expiry handling: the app now corroborates on-disk session state to avoid false sign-outs and only clears local data when expiry is confirmed.
  * Immediate sign-out for confirmed expiry signals; debounced, cautious handling for unconfirmed signals.
  * Deep-link and socket expiry events are classified to ensure correct immediate vs cautious behavior.

* **Tests**
  * Updated tests cover confirmed vs unconfirmed expiry, corroboration races, and debounce behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2758?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/providers/CoreStateProvider.tsx
+++ b/app/src/providers/CoreStateProvider.tsx
@@ -16,6 +16,7 @@ import {
   setCoreStateSnapshot,
 } from '../lib/coreState/store';
 import { syncAnalyticsConsent } from '../services/analytics';
+import type { AuthExpiredReason } from '../services/coreRpcClient';
 import {
   fetchCoreAppSnapshot,
   getTeamInvites,
@@ -30,6 +31,7 @@ import { loadThreads, resetThreadCachesPreservingSelection } from '../store/thre
 import { getActiveUserId, setActiveUserId } from '../store/userScopedStorage';
 import { isLocalSessionToken } from '../utils/localSession';
 import {
+  getSessionToken,
   openhumanUpdateAnalyticsSettings,
   openhumanUpdateMeetSettings,
   restartApp,
@@ -61,6 +63,53 @@ function sanitizeError(error: unknown): { message?: string; code?: string; statu
     };
   }
   return { message: String(error) };
+}
+
+/**
+ * Positively confirm the on-disk session token is gone before an `auth_expired`
+ * signal is allowed to trigger the *destructive* `clearSession` (which calls
+ * `auth_clear_session` → removes the auth profile from disk).
+ *
+ * Reads the cheap disk-only `auth_get_session_token` RPC — no `auth/me` network
+ * call, not subject to `app_state_snapshot`'s 5s/10s timeouts. Right after the
+ * identity-flip restart the token IS on disk, but a token-gated RPC can briefly
+ * report "session jwt required" before the profile finishes loading; a short
+ * retry rides out that boot-load window.
+ *
+ * Returns `true` ONLY when every attempt reads an empty token. A token that is
+ * present, or an RPC failure (inconclusive), returns `false` — biasing toward
+ * keeping the session rather than destroying a valid one.
+ */
+async function confirmSessionTokenGone(): Promise<boolean> {
+  const ATTEMPTS = 3;
+  const RETRY_DELAY_MS = 300;
+  for (let attempt = 1; attempt <= ATTEMPTS; attempt++) {
+    let token: string | null;
+    try {
+      token = await getSessionToken();
+    } catch (err) {
+      log(
+        'auth-expired corroboration inconclusive (attempt %d/%d) — keeping session: %O',
+        attempt,
+        ATTEMPTS,
+        sanitizeError(err)
+      );
+      return false;
+    }
+    if (token && token.trim() !== '') {
+      log(
+        'auth-expired corroboration: session token still on disk (attempt %d/%d) — keeping session',
+        attempt,
+        ATTEMPTS
+      );
+      return false;
+    }
+    if (attempt < ATTEMPTS) {
+      await new Promise(resolve => setTimeout(resolve, RETRY_DELAY_MS));
+    }
+  }
+  log('auth-expired corroboration: session token confirmed absent after %d attempts', ATTEMPTS);
+  return true;
 }
 
 export function coreStatePollFailureWarningMessage(failureCount: number): string | null {
@@ -664,7 +713,7 @@ export default function CoreStateProvider({ children }: { children: ReactNode })
   // latest closure; `clearSession`'s own deps are stable `useCallback`s,
   // so re-registers are rare.
   useEffect(() => {
-    const runReauth = (method: string, source: string) => {
+    const runReauth = async (method: string, source: string, reason: AuthExpiredReason) => {
       if (isLocalSessionToken(getCoreStateSnapshot().snapshot.sessionToken)) {
         log('auth-expired ignored for local session (method=%s source=%s)', method, source);
         return;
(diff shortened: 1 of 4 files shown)


sanil-23:
@coderabbitai Good catch on the debounce-vs-reason ordering — addressed in `3ae3bbfa`.

The 10s reauth debounce ran *before* the `reason` check, so a transient `unconfirmed` probe could claim the slot and suppress a real `401` / `auth:session_expired` landing within the window. Fix:

- A `confirmed` expiry now **breaks through** the debounce when the slot was claimed by an `unconfirmed` probe (`confirmedOverridesUnconfirmed`).
- An **attempt-id guard** (`reauthAttemptIdRef`) stops an in-flight `unconfirmed` corroboration from double-clearing after a `confirmed` event supersedes it mid-await.
- Once we commit to a real sign-out, the slot is marked `confirmed` so a follow-up `confirmed` event in the window is still coalesced (no double `clearSession`).

Added a regression test: unconfirmed probe (token still present → no logout, claims slot) immediately followed by a `confirmed` 401 → still signs out. Full suite green (121 tests).

graycyrus (commented):
@sanil-23 hey! the code looks good to me, but there are a few CI jobs still pending (macOS E2E, Windows E2E, macOS smoke install). once those are green i'll come back and approve this. let me know if you need any help!

For what it's worth — the design here is solid. The `confirmed` vs `unconfirmed` split is exactly the right abstraction, `confirmSessionTokenGone` biasing toward keeping the session on any inconclusive read is the correct call, and the stale-attempt check (`attemptId !== reauthAttemptIdRef.current`) correctly handles the case where a real 401 breaks through during the corroboration window. CodeRabbit flagged the "unconfirmed probe blocks later confirmed" issue as outside their diff range, but the PR already implements the fix (`confirmedOverridesUnconfirmed`) — and the regression test for that path (`lets a confirmed expiry break through a debounce slot claimed by an unconfirmed probe`) is in there too.

PR #2761, opened 2026-05-27
fix(config): reconcile orphaned provider references on upgrade (#2759)

### LINKED ISSUE #2759: Disabling a cloud provider leaves workloads routed to its removed model after update
## Summary

After a user removes/disables a cloud provider (e.g. OpenAI), per-workload routing strings that still reference it (e.g. `chat_provider = "openai:gpt-4o"`) are left dangling. The AI settings panel keeps showing the now-gone model, and at inference time the workload hard-errors instead of falling back to the managed backend.

## Problem

**What happens:** On a build where chat was set to a custom provider + model (e.g. OpenAI / `gpt-4o`), disabling that provider removes its `cloud_providers` entry but leaves `chat_provider = "openai:gpt-4o"` on disk. After updating:

- Settings → AI still renders the OpenAI model for chat even though the provider is gone (`loadAISettings` parses the stored string without checking the slug still exists), and
- at runtime `inference::provider::factory::make_cloud_provider_by_slug` returns `no cloud provider configured for slug 'openai'` — that workload's inference breaks instead of falling back to managed.

A dangling `primary_cloud` id (provider entry removed) is the same class of orphan.

**What I expected:** a workload whose provider was removed should resolve to managed (OpenHuman) routing — consistently in the UI and at runtime.

**Steps to reproduce:**
1. Configure chat to a custom provider + model (e.g. OpenAI, `gpt-4o`).
2. Disable/remove that provider.
3. Update the app.
4. Open Settings → AI: chat still shows the OpenAI model; chat inference errors with `no cloud provider configured for slug 'openai'`.

**Version / platform:** observed updating from a prod build (~v0.54.x), desktop; persisted config at `schema_version ≥ 5`.

## Solution (optional)

A startup migration that reconciles orphaned provider references: reset any `*_provider` whose slug is absent from `cloud_providers` (and the always-managed `openhuman:<model>` form, and bare non-sentinel strings) back to managed (`None`), and clear a dangling `primary_cloud`. Mirror the inference factory's exact grammar so "resolvable in the migration" means the same as "resolvable at inference time."

## Acceptance criteria

- [ ] **Repro gone** — a config with an orphaned `*_provider` resolves to managed after upgrade.
- [ ] **Regression safety** — migration unit tests cover orphan scrub, keep-when-present, `openhuman:` normalize, dangling/valid `primary_cloud`, and idempotency.

## Related

- Fix PR: to follow.

sanil-23:
## Summary

- Adds a `v5 → v6` startup migration `reconcile_orphaned_providers` that resets per-workload provider strings referencing a removed cloud provider back to managed, and clears a dangling `primary_cloud`.
- Fixes the case where disabling a cloud provider (e.g. OpenAI) leaves `chat_provider = "openai:gpt-4o"` orphaned after an update — the AI panel kept showing the gone model and inference hard-errored.
- Mirrors the inference factory's exact, case-sensitive grammar (reuses `PROVIDER_OPENHUMAN` / `OLLAMA_PROVIDER_PREFIX` / `LM_STUDIO_PROVIDER_PREFIX`) and normalizes the always-managed `openhuman:<model>` form to `None`.
- Core-only; no UI change (`loadAISettings` already renders `None` as managed).

## Problem

A workload's provider string (`chat_provider`, `reasoning_provider`, …) uses the grammar parsed by `inference::provider::factory`: `<slug>:<model>` resolves to the `cloud_providers` entry whose `slug == <slug>`. When a user removes/disables a cloud provider, its entry leaves `cloud_providers` but the workload string keeps pointing at it. Nothing reconciles this, so:

- `loadAISettings` parses the stored string without checking the slug still exists → the AI panel renders a model whose provider is gone, while everything else reads as managed; and
- at runtime `factory::make_cloud_provider_by_slug` returns `no cloud provider configured for slug '<slug>'` — that workload's inference breaks instead of falling back to managed.

A dangling `primary_cloud` id (provider entry removed) is the same class of orphan. Confirmed against live configs at `schema_version = 5`.

## Solution

- New `src/openhuman/migrations/reconcile_orphaned_providers.rs` (pure, idempotent `Config` mutation). For each of the nine `*_provider` fields, reset to `None` (= managed) anything the factory cannot resolve: a `<slug>:<model>` whose slug is absent from `cloud_providers`, the always-managed `openhuman:<model>` form, or a bare non-sentinel string. Sentinels (`""` / `cloud` / `openhuman`) and local providers (`ollama:` / `lmstudio:`, valid without a cloud entry) are left untouched. Clears a dangling `primary_cloud`.
- Wired into `migrations/mod.rs`: registered the module, bumped `CURRENT_SCHEMA_VERSION` 5 → 6, added the `== 5`-gated branch using the same save / rollback pattern as the existing steps.
- **Design note:** the inference factory was intentionally *not* changed to fall back silently on a missing slug — main deliberately fails closed there (`BYOK_INCOMPLETE_SENTINEL`). The migration makes the persisted config self-consistent, so the factory never sees the orphan at runtime.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 11 unit tests: orphan scrub, keep-when-present, temperature-suffix preserve, `openhuman:` normalize, sentinels/local untouched, bare-string, all-9-workloads, dangling/valid `primary_cloud`, clean no-op, idempotency.
- [x] **Diff coverage ≥ 80%** — new migration file is 100% line/region coverage (`cargo-llvm-cov`, 64/64 lines, 89/89 regions); the new `mod.rs` branch's happy path is covered by the existing `run_pending` chain tests. Aggregate changed-line coverage ~93%.
- [x] Coverage matrix updated — N/A: behaviour-preserving data migration, no feature row in the matrix.
- [x] All affected feature IDs listed under `## Related` — N/A: no matrix feature IDs touched.
- [x] No new external network dependencies introduced — confirmed (pure in-memory config mutation).
- [x] Manual smoke checklist updated — N/A: no release-cut surface change.

## Impact

- Desktop core (Rust). Runs once per workspace on the next launch (`schema_version 5 → 6`); idempotent, best-effort (failures logged, retried next launch), never touches keys/secrets.
- No UI, network, or performance impact. Affected users' chat/background workloads that pointed at a removed provider transparently revert to managed routing.

## Related

- Closes: #2759
- Follow-up PR(s)/TODOs: none

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/reconcile-orphaned-providers`
- Commit SHA: `76d20f81895f93f3e0fd5fc020816b9aa141212f`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — N/A: no `app/` (TS) changes.
- [x] Focused tests: `cargo test --lib migrations::` → 67 passed, 0 failed (incl. 11 new).
- [x] Rust fmt/check: `cargo fmt --check` clean; `cargo check` clean (only pre-existing `slack-backfill` dead-code warnings).
- [x] Tauri fmt/check (if changed): N/A: Tauri shell untouched.

### Validation Blocked
- `error:` pnpm/node not on this non-interactive shell's PATH
- `impact:` none — change is core-only Rust, validated via `cargo check` + `cargo test` + `cargo fmt --check`; pushed with `--no-verify`.

### Behavior Changes
- Intended behavior change: orphaned `*_provider` / dangling `primary_cloud` reconciled to managed on upgrade.
- User-visible effect: a workload whose provider was removed now resolves to managed instead of showing a gone model / erroring.

### Parity Contract
- Legacy behavior preserved: sentinels, local providers, and valid `<slug>:<model>` routes are untouched; only unresolvable references change.
- Guard/fallback/dispatch parity checks: migration's "resolvable" rule mirrors `factory::make_cloud_provider_by_slug`'s exact slug lookup; factory's fail-closed `BYOK_INCOMPLETE_SENTINEL` path intentionally left intact.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Added a one‑time migration that reconciles and cleans orphaned or stale provider references during startup.

* **Chores**
  * Configuration schema advanced to version 6; migration only marks success after all reconciliation steps persist.

* **Bug Fixes**
  * Invalid or missing provider references (including primary cloud) are cleared to prevent runtime confusion.

* **Tests**
  * Expanded tests covering migration behavior, idempotency, and logging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2761?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/migrations/reconcile_orphaned_providers.rs
+++ b/src/openhuman/migrations/reconcile_orphaned_providers.rs
@@ -1,0 +1,203 @@
+//! Migration 5 → 6: reconcile orphaned per-workload provider references.
+//!
+//! ## The problem
+//!
+//! Each LLM workload is routed by a provider string in `Config`
+//! (`chat_provider`, `reasoning_provider`, …) using the grammar parsed by
+//! [`crate::openhuman::inference::provider::factory`]:
+//!
+//! ```text
+//! ""/"cloud"          → primary_cloud (managed)
+//! "openhuman"         → managed OpenHuman backend
+//! "ollama:<model>"    → local Ollama
+//! "lmstudio:<model>"  → local LM Studio
+//! "<slug>:<model>"    → the cloud_providers entry whose slug == <slug>
+//! ```
+//!
+//! A `"<slug>:<model>"` string only resolves while a `cloud_providers` entry
+//! with that slug still exists. When a user **removes/disables a cloud
+//! provider** (e.g. OpenAI) on one build, the entry leaves `cloud_providers`
+//! but the workload string keeps pointing at it — e.g. `chat_provider =
+//! "openai:gpt-4o"` with no `openai` entry. Nothing reconciles this, so:
+//!
+//! - the AI settings panel still shows the now-gone OpenAI model for chat, and
+//! - at runtime [`factory::make_cloud_provider_by_slug`] **hard-errors**
+//!   ("no cloud provider configured for slug 'openai'") instead of falling
+//!   back to managed — that workload's inference breaks.
+//!
+//! ## What this migration does
+//!
+//! A pure, idempotent pass over the persisted `Config`:
+//!
+//! - For each of the nine `*_provider` fields, reset to `None` (= managed) any
+//!   value that the factory could not resolve: a `"<slug>:<model>"` whose slug
+//!   is absent from `cloud_providers`, the always-managed `"openhuman:<model>"`
+//!   form, or a bare non-sentinel string. Sentinels (`""`, `"cloud"`,
+//!   `"openhuman"`) and local providers (`ollama:`/`lmstudio:`, valid without a
+//!   `cloud_providers` entry) are left untouched.
+//! - Clear `primary_cloud` when it points at an id no longer present in
+//!   `cloud_providers` (a dangling pointer; the factory already falls back to
+//!   the managed backend for an unresolved primary).
+//!
+//! Mirrors the factory's *exact*, case-sensitive grammar so "resolvable here"
+//! means the same as "resolvable at inference time". The one intentional step
+//! beyond the factory is normalizing `"openhuman:<model>"` → `None`: both route
+//! to the managed backend, and `None` matches what the settings UI persists.
+//!
+//! ## Behaviour
+//!
+//! - Pure in-memory mutation of `Config`; the caller (`migrations::run_pending`)
+//!   persists via `Config::save()` and bumps `schema_version`.
+//! - Idempotent: a second run finds nothing left to scrub.
+//! - Never touches keys/secrets or any other config field.
+
+use crate::openhuman::config::Config;
+use crate::openhuman::inference::provider::factory::{
+    LM_STUDIO_PROVIDER_PREFIX, OLLAMA_PROVIDER_PREFIX, PROVIDER_OPENHUMAN,
+};
+use std::collections::HashSet;
+
+/// Counters returned by [`run`] for diagnostics. Logged at INFO once per run.
+#[derive(Debug, Default, Clone)]
+pub struct MigrationStats {
+    /// Number of `*_provider` fields reset to managed (`None`).
+    pub workload_fields_scrubbed: usize,
+    /// `true` when a dangling `primary_cloud` pointer was cleared.
+    pub primary_cloud_cleared: bool,
+}
+
+/// Run the orphaned-provider reconciliation on the given `Config`.
+///
+/// Synchronous — pure config mutation, no I/O. Caller persists via
+/// `Config::save()` once `schema_version` is also bumped.
+///
+/// Returns `anyhow::Result` for uniformity with the other migration steps in
+/// [`super`] (the runner matches `Ok`/`Err` for every step). This pass has no
+/// fallible operations today, so it always returns `Ok`; the signature keeps
+/// the contract consistent and leaves room for future I/O-bearing logic.
+pub fn run(config: &mut Config) -> anyhow::Result<MigrationStats> {
+    let mut stats = MigrationStats::default();
+
+    // Own the slug/id sets so the immutable borrow on `cloud_providers` ends
+    // before we take `&mut` to the workload fields below. Stored slugs/ids are
+    // kept verbatim (no trim) so membership matches the factory's exact,
+    // case-sensitive `e.slug == slug` / `e.id == id` comparisons: a provider
+    // stored as " openai" must NOT make a "openai:model" route look resolvable,
+    // since the factory would still fail it.
+    let known_slugs: HashSet<String> = config
+        .cloud_providers
+        .iter()
+        .map(|e| e.slug.clone())
+        .collect();
+    let known_ids: HashSet<String> = config
+        .cloud_providers
+        .iter()
+        .map(|e| e.id.clone())
+        .collect();
+
+    let mut scrubbed = 0usize;
+    for (workload, field) in workload_fields(config) {
+        let Some(raw) = field.as_deref() else {
+            continue;
+        };
+        let s = raw.trim();
+
+        // Managed sentinels and local providers resolve without a
+        // cloud_providers entry — leave them alone.
+        if s.is_empty()
+            || s == "cloud"
+            || s == PROVIDER_OPENHUMAN
+            || s.starts_with(OLLAMA_PROVIDER_PREFIX)
+            || s.starts_with(LM_STUDIO_PROVIDER_PREFIX)
+        {
+            continue;
+        }
+
+        let scrub_reason = match s.split_once(':') {
+            // "openhuman:<model>" is always the managed backend regardless of
+            // the suffix — normalize to None to match the bare sentinel.
+            Some((slug, _)) if slug.trim() == PROVIDER_OPENHUMAN => Some("openhuman-slug"),
+            // "<slug>:<model>" whose slug is no longer configured — the orphan.
+            Some((slug, _)) if !known_slugs.contains(slug.trim()) => Some("missing-slug"),
+            Some(_) => None,
+            // Bare non-sentinel (e.g. "openai") — the factory rejects this form.
+            None => Some("bare-unresolvable"),
+        };
+
+        if let Some(reason) = scrub_reason {
+            log::info!(
+                "[migrations][reconcile-providers] {workload}_provider={} -> managed (None) \
+                 reason={reason}",
+                redact_provider_for_log(raw)
+            );
+            *field = None;
+            scrubbed += 1;
+        }
+    }
+    stats.workload_fields_scrubbed = scrubbed;
+
+    // A primary_cloud id absent from cloud_providers is a dangling pointer.
+    // The factory already resolves an unfound primary to the managed backend,
+    // so null it for consistency with the on-disk truth.
+    let dangling_primary = config
+        .primary_cloud
+        .as_deref()
+        .is_some_and(|id| !known_ids.contains(id));
+    if dangling_primary {
+        // Don't log the raw id — primary_cloud is user-editable config; the fact
+        // that it dangled is the only diagnostic that matters here.
+        log::info!(
+            "[migrations][reconcile-providers] primary_cloud points at a missing cloud_providers \
+             entry -> cleared (managed fallback)"
+        );
+        config.primary_cloud = None;
+        stats.primary_cloud_cleared = true;
+    }
+
+    log::info!(
+        "[migrations][reconcile-providers] done workload_fields_scrubbed={} primary_cloud_cleared={}",
+        stats.workload_fields_scrubbed,
+        stats.primary_cloud_cleared
+    );
+    Ok(stats)
+}
+
+/// The nine per-workload routing fields, paired with a label for logging.
+///
+/// Borrowed mutably in one array literal — legal because they are disjoint
+/// fields of `Config`.
+///
+/// IMPORTANT: this list must stay in sync with the `*_provider` fields on
+/// [`Config`]. If a tenth workload provider field is added there, add it here
+/// (and bump the array length) or the migration will silently skip it. Rust
+/// can't enforce this at compile time without a field-reflection macro, and a
+/// serde-based count guard doesn't work because the `Option<String>` fields
+/// default to `None` and are omitted from the serialized table.
+fn workload_fields(config: &mut Config) -> [(&'static str, &mut Option<String>); 9] {
+    [
+        ("chat", &mut config.chat_provider),
+        ("reasoning", &mut config.reasoning_provider),
+        ("agentic", &mut config.agentic_provider),
+        ("coding", &mut config.coding_provider),
+        ("memory", &mut config.memory_provider),
+        ("embeddings", &mut config.embeddings_provider),
+        ("heartbeat", &mut config.heartbeat_provider),
+        ("learning", &mut config.learning_provider),
+        ("subconscious", &mut config.subconscious_provider),
+    ]
+}
+
+/// Redact a provider string for logging. Provider values are user-editable and
+/// could carry sensitive content in the model segment, so keep only the
+/// non-sensitive slug (the useful diagnostic) and mask the rest. Per the
+/// project rule: never log secrets or full PII.
+fn redact_provider_for_log(raw: &str) -> String {
+    match raw.trim().split_once(':') {
+        Some((slug, _)) => format!("{}:<redacted>", slug.trim()),
+        None => "<redacted>".to_string(),
+    }
+}
+
+#[cfg(test)]
+#[path = "reconcile_orphaned_providers_tests.rs"]
+mod tests;
--- a/src/openhuman/migrations/mod.rs
+++ b/src/openhuman/migrations/mod.rs
@@ -25,6 +25,7 @@ use crate::openhuman::config::Config;
 
 mod expand_autonomy_defaults;
 mod phase_out_profile_md;
+mod reconcile_orphaned_providers;
 mod remove_write_auto_approve;
 mod repair_http_request_limits;
 mod retire_chat_v1_model;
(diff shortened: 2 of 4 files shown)


src/openhuman/migrations/reconcile_orphaned_providers.rs
```diff
@@ -0,0 +1,177 @@
...
+            Some(_) => None,
+            // Bare non-sentinel (e.g. "openai") — the factory rejects this form.
+            None => Some("bare-unresolvable"),
+        };
+
+        if let Some(reason) = scrub_reason {
+            log::info!(
+                "[migrations][reconcile-providers] {workload}_provider={raw:?} -> managed (None) \
+                 reason={reason}"
+            );
```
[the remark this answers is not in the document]
sanil-23: Addressed in a1f704f5. Added `redact_provider_for_log` — workload provider strings now log as `slug:<redacted>` (slug kept as the diagnostic, model segment masked), and the raw `primary_cloud` id is dropped from the log entirely. Reason + workload are still logged for tracing. Covered by a new unit test.

M3gA-Mind (commented):
## Code Review

### Overview

Adds a schema v5→v6 startup migration that resets workload provider strings (e.g. `chat_provider = "openai:gpt-4o"`) back to `None` (managed) when the referenced slug no longer exists in `cloud_providers`. Also clears a dangling `primary_cloud` id. Directly fixes the hard-error in the inference factory (`no cloud provider configured for slug 'openai'`) that occurs after a user removes a BYOK provider.

---

### Strengths

- Migration logic is pure, synchronous, and idempotent — correct design for a startup reconciliation pass.
- Reuses the factory's exact constants (`PROVIDER_OPENHUMAN`, `OLLAMA_PROVIDER_PREFIX`, `LM_STUDIO_PROVIDER_PREFIX`), so "resolvable in migration" aligns with "resolvable at inference time."
- Save/rollback pattern in `mod.rs` matches all prior migration steps.
- Logging is verbose (correct per project convention), per-field `reason=` tags are grep-friendly, model values are redacted via `redact_provider_for_log`.
- 11 unit tests including idempotency and all 9 workload fields.

---

### Issues / Suggestions

**1. No integration test for the v5→v6 path in `mod_tests.rs` (gap)**

There's `run_pending_v4_to_v5_removes_write_tools_from_auto_approve` but no equivalent for v5→v6. The unit tests exercise the function in isolation, but nothing tests the wiring in `run_pending` with a config at `schema_version = 5` and asserts it reaches 6 and scrubs orphaned references. This is the riskiest gap given the branch is version-gated.

**2. `anyhow::Result` return is vestigial**

`run()` never returns `Err` — no `?` operators, no failure paths. The error arm in `mod.rs` is currently dead code. Either add a comment noting this is for future extensibility, or return `MigrationStats` directly.

**3. `workload_fields` array has no exhaustiveness guarantee**

```rust
fn workload_fields(config: &mut Config) -> [(&'static str, &mut Option<String>); 9] {
```

If a tenth `*_provider` field is added to `Config`, this silently misses it. The test `scrubs_every_orphaned_workload` hardcodes `9` and would still pass. A compile-time assertion or a `#[cfg(test)]` static check against a known field count would guard against future drift.

**4. Trimming asymmetry with the factory**

`known_slugs` is built with `.trim()` on stored slugs, and `slug.trim()` is used during lookup. The factory's `make_cloud_provider_by_slug` does an exact `e.slug == slug` with no trimming. A provider stored as `" openai"` (with a space) would appear in `known_slugs` as `"openai"`, so a `"openai:model"` string would not be scrubbed — but the factory would still fail to resolve it. Unlikely edge case, but the comment claiming exact factory parity is slightly misleading here.

---

### Test Coverage

| Scenario | Covered |
|---|---|
| Orphaned `<slug>:<model>` scrubbed | ✅ |
| Valid slug kept | ✅ |
| Temperature suffix preserved | ✅ |
| `openhuman:<model>` normalized | ✅ |
| Sentinels + local providers untouched | ✅ |
| Bare non-sentinel string scrubbed | ✅ |
| All 9 fields scrubbed | ✅ |
| Dangling `primary_cloud` cleared | ✅ |
| Valid `primary_cloud` kept | ✅ |
| Clean config is no-op | ✅ |
| Idempotency | ✅ |
| `run_pending` wiring at `schema_version == 5` | ❌ Missing |

---

### Verdict

**Approve with one request:** add an integration test in `mod_tests.rs` mirroring the v4→v5 pattern — start at `schema_version = 5`, set an orphaned provider, call `run_pending`, assert version reaches 6 and the field is `None`. Everything else is well-implemented and follows project conventions.

sanil-23:
Thanks for the thorough review @M3gA-Mind — all four points addressed in `d0a1b24c`:

**1. Missing v5→v6 integration test (the request) — fixed.** Added `run_pending_v5_to_v6_reconciles_orphaned_providers` in `mod_tests.rs`, mirroring the v4→v5 pattern: starts at `schema_version = 5` with an orphaned `chat_provider = "openai:gpt-4o"` (no `openai` entry) and a dangling `primary_cloud`, calls `run_pending`, and asserts the version reaches 6, both references are reset to managed, and `schema_version = 6` is persisted on disk.

**2. Vestigial `anyhow::Result` — documented.** Kept the signature for uniformity with the migration-runner contract (`run_pending` matches `Ok`/`Err` for every step) and added a doc comment noting it has no fallible ops today and is `Ok`-always, leaving room for future I/O-bearing logic. Changing just this one step to return `MigrationStats` directly would make it inconsistent with the other four steps.

**3. `workload_fields` exhaustiveness — documented the sync requirement.** Added an `IMPORTANT` doc comment that the list must stay in sync with `Config`'s `*_provider` fields. I looked at a compile-time/serde guard but it doesn't hold up: Rust has no field reflection without a macro, and a serde-based key count fails because the `Option<String>` fields default to `None` and are omitted from the serialized TOML. Flagging it in the doc comment is the honest mitigation; happy to add a proc-macro guard in a follow-up if you'd prefer.

**4. Trim asymmetry vs the factory — fixed (good catch).** You're right that this broke the parity claim. `known_slugs`/`known_ids` are now built verbatim (no `.trim()`), matching the factory's exact `e.slug == slug` / `e.id == id`. So a provider stored as `" openai"` no longer makes a `"openai:model"` route look resolvable — it's correctly scrubbed, since the factory would reject it too. Added `scrubs_when_stored_slug_has_whitespace_factory_would_reject` to lock this in. (The lookup side still trims, matching the factory's own `slug = p[..colon].trim()`.)

Migration suite green (70 tests), `cargo fmt --check` clean. CI re-running on `d0a1b24c`.

M3gA-Mind (commented):
All four issues from my initial review are addressed — LGTM.

- **Integration test** ✅ — `run_pending_v5_to_v6_reconciles_orphaned_providers` covers the full wiring path.
- **`anyhow::Result`** ✅ — Doc comment explains forward-compatibility rationale.
- **`workload_fields` exhaustiveness** ✅ — `IMPORTANT` comment added.
- **Trimming asymmetry** ✅ — Stored slugs now cloned without `.trim()`, matching the factory's exact comparison. New `scrubs_when_stored_slug_has_whitespace_factory_would_reject` test covers the edge case explicitly.

M3gA-Mind (dismissed):
LGTM — all review items addressed. Ready to merge.

graycyrus (commented):
@sanil-23 hey! the code looks clean to me — well-designed migration, thorough test suite, and M3gA-Mind's items are all addressed. a few CI checks are failing though (Rust Core Coverage, Rust Tauri Coverage, Coverage Matrix Sync, PR Submission Checklist) and several more are still pending. once those are green i'll come back and approve this. let me know if you need any help getting CI sorted.

M3gA-Mind (commented):
# PR #2761 — fix(config): reconcile orphaned provider references on upgrade (#2759)

> ⚠️ Note: this PR is already **MERGED**. Review posted as a comment for the record / potential follow-up — state is not being changed.

## Walkthrough

Adds a `v5 → v6` startup migration `reconcile_orphaned_providers` that resets per-workload `*_provider` strings (and a dangling `primary_cloud`) pointing at a `cloud_providers` slug that no longer exists, so a removed cloud provider no longer leaves the AI panel showing a gone model or hard-erroring at inference. The migration is pure/idempotent, well-documented, redacts log output, and ships with 13 unit tests plus an end-to-end `run_pending` test. The `mod.rs` wiring nicely refactors the 5→6 transition so the version bump happens only after *both* 5→6 steps succeed.

The approach is sound and the implementation is careful — but the "leave local providers untouched" allow-list is **incomplete** relative to the inference factory it claims to mirror, which turns this fix into a silent regression for a subset of users (see blocker below).

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/migrations/mod.rs` | Rework 5→6 transition to run `repair_http_request_limits` + new `reconcile_orphaned_providers`, bumping/saving only when both succeed. |
| `src/openhuman/migrations/reconcile_orphaned_providers.rs` | New migration: scrub unresolvable `*_provider` strings to `None`, clear dangling `primary_cloud`. |
| `src/openhuman/migrations/reconcile_orphaned_providers_tests.rs` | 13 unit tests (orphan scrub, keep-when-present, whitespace, sentinels/local, bare, all-9, primary_cloud, redaction, idempotency). |
| `src/openhuman/migrations/mod_tests.rs` | One e2e test driving the orphan scrub through full `run_pending`. |

## Actionable comments (1)

### 🛑 Blockers

#### 1. `src/openhuman/migrations/reconcile_orphaned_providers.rs:107-114` — skip-list omits `mlx:`, `local-openai:`, `claude-code:`, and `claude_agent_sdk` → valid routes silently scrubbed to managed

The module docstring and the inline comment both claim this mirrors the factory's *exact* grammar and leaves "local providers" untouched. But `create_chat_provider_from_string` (`factory.rs:431-555`) resolves **six** non-`cloud_providers` forms without a slug lookup, while the migration only skips two of them:

| Form | Factory (`factory.rs`) | Migration skip-list |
| --- | --- | --- |
| `openhuman` / `cloud` / `""` | ✅ | ✅ |
| `ollama:<model>` | ✅ L491 | ✅ |
| `lmstudio:<model>` | ✅ L504 | ✅ |
| `mlx:<model>` | ✅ L517 | ❌ → `missing-slug` → scrubbed |
| `local-openai:<model>` | ✅ L530 | ❌ → `missing-slug` → scrubbed |
| `claude-code:<model>` | ✅ L450 | ❌ → `missing-slug` → scrubbed |
| `claude_agent_sdk` / `claude_agent_sdk:<model>` | ✅ L543 | ❌ → `bare-unresolvable` / `missing-slug` → scrubbed |

For a user whose, say, `chat_provider = "mlx:my-local-model"` or `agentic_provider = "claude-code:sonnet"`, the factory resolves these fine — but this migration sees a slug not in `cloud_providers` and resets the field to `None` (managed) on the 5→6 upgrade. That is exactly the kind of silent, persisted behavior change this PR is meant to *prevent*, and it isn't covered by any test (all "local untouched" assertions only exercise `ollama:`/`lmstudio:`).

The factory already exports the constants needed to stay in lockstep.

**Suggested change:**
```rust
// before
use crate::openhuman::inference::provider::factory::{
    LM_STUDIO_PROVIDER_PREFIX, OLLAMA_PROVIDER_PREFIX, PROVIDER_OPENHUMAN,
};
// ...
        if s.is_empty()
            || s == "cloud"
            || s == PROVIDER_OPENHUMAN
            || s.starts_with(OLLAMA_PROVIDER_PREFIX)
            || s.starts_with(LM_STUDIO_PROVIDER_PREFIX)
        {
            continue;
        }

// after
use crate::openhuman::inference::provider::claude_code::PROVIDER_PREFIX as CLAUDE_CODE_PROVIDER_PREFIX;
use crate::openhuman::inference::provider::factory::{
    CLAUDE_AGENT_SDK_PREFIX, CLAUDE_AGENT_SDK_PROVIDER, LM_STUDIO_PROVIDER_PREFIX,
    LOCAL_OPENAI_PROVIDER_PREFIX, MLX_PROVIDER_PREFIX, OLLAMA_PROVIDER_PREFIX, PROVIDER_OPENHUMAN,
};
// ...
        if s.is_empty()
            || s == "cloud"
            || s == PROVIDER_OPENHUMAN
            || s == CLAUDE_AGENT_SDK_PROVIDER
            || s.starts_with(OLLAMA_PROVIDER_PREFIX)
            || s.starts_with(LM_STUDIO_PROVIDER_PREFIX)
            || s.starts_with(MLX_PROVIDER_PREFIX)
            || s.starts_with(LOCAL_OPENAI_PROVIDER_PREFIX)
            || s.starts_with(CLAUDE_CODE_PROVIDER_PREFIX)
            || s.starts_with(CLAUDE_AGENT_SDK_PREFIX)
        {
            continue;
        }
```
Plus a test asserting `mlx:`, `local-openai:`, `claude-code:`, and `claude_agent_sdk[:..]` survive with no matching `cloud_providers` entry. (As a hardening idea, a future change could have the migration call the same parse path the factory uses, so the two can never drift again.)

- `src/openhuman/config/schema/types.rs:392-396` — `stt_provider` / `tts_provider` use the same `"<slug>:<model>"` grammar against a separate `voice_providers` list (`voice::factory`) and can dangle in exactly the same way when a voice provider is removed. Intentionally out of scope for this PR, or worth a follow-up? Worth a one-line note in the module docstring stating the migration is LLM-workload-only so the omission reads as deliberate.

## Nitpicks (2)

- `src/openhuman/migrations/reconcile_orphaned_providers.rs:1-60` — the docstring's "Mirrors the factory's *exact*, case-sensitive grammar" line will become false once the blocker is addressed only partially; update it to enumerate the full local-provider set (or reference the factory constants) so the doc stays honest.
- `src/openhuman/migrations/mod.rs:9-15` — the "Adding a new migration" guidance still says `if config.schema_version < N`, but every actual step (and this PR) uses `== N` to avoid skipping a previously-failed step. Pre-existing, but worth aligning the doc while this file is being touched.

## Verified / looks good

- `mod.rs` 5→6 rework is correct: `all_ok` gates the single version bump + save, and both steps are idempotent so re-running the already-succeeded one next launch is a no-op. Rollback-on-`save()`-failure pattern matches the existing steps.
- Whitespace handling matches the factory: `known_slugs` stores slugs verbatim while lookups use `slug.trim()`, mirroring `make_cloud_provider_by_slug`'s trimmed-route-slug vs verbatim-stored-slug comparison (covered by `scrubs_when_stored_slug_has_whitespace_factory_would_reject`).
- `primary_cloud` clearing is consistent with the factory's managed fallback for an unresolved primary; valid/dangling both tested.
- Logging redacts the model segment (`redact_provider_for_log`) and never logs the raw `primary_cloud` id — compliant with the no-secrets/PII rule.
- `workload_fields` returns nine disjoint `&mut` field borrows in one array literal — legal and the `IMPORTANT` sync comment is a good guard.

src/openhuman/migrations/reconcile_orphaned_providers.rs
```diff
@@ -0,0 +1,203 @@
...
+        // Managed sentinels and local providers resolve without a
+        // cloud_providers entry — leave them alone.
+        if s.is_empty()
+            || s == "cloud"
+            || s == PROVIDER_OPENHUMAN
+            || s.starts_with(OLLAMA_PROVIDER_PREFIX)
+            || s.starts_with(LM_STUDIO_PROVIDER_PREFIX)
+        {
+            continue;
+        }
```
M3gA-Mind, line 114: 🛑 **Blocker — skip-list omits `mlx:`, `local-openai:`, `claude-code:`, `claude_agent_sdk`**

This block claims to leave "local providers" untouched, but `create_chat_provider_from_string` (`factory.rs:431-555`) resolves six non-`cloud_providers` forms without a slug lookup. The skip-list only covers two of them (`ollama:`/`lmstudio:`), so a user with e.g. `chat_provider = "mlx:my-model"` or `agentic_provider = "claude-code:sonnet"` gets that field silently reset to `None` (managed) on the 5→6 upgrade — exactly the regression this migration is meant to prevent, and untested.

```rust
// add the remaining factory-recognised local/subprocess forms:
            || s == CLAUDE_AGENT_SDK_PROVIDER
            || s.starts_with(MLX_PROVIDER_PREFIX)
            || s.starts_with(LOCAL_OPENAI_PROVIDER_PREFIX)
            || s.starts_with(CLAUDE_CODE_PROVIDER_PREFIX)   // claude_code::PROVIDER_PREFIX
            || s.starts_with(CLAUDE_AGENT_SDK_PREFIX)
```

The factory already exports `MLX_PROVIDER_PREFIX`, `LOCAL_OPENAI_PROVIDER_PREFIX`, `CLAUDE_AGENT_SDK_PREFIX`, `CLAUDE_AGENT_SDK_PROVIDER`, and `claude_code::PROVIDER_PREFIX`. Add a test asserting each survives with no matching `cloud_providers` entry.

src/openhuman/migrations/reconcile_orphaned_providers.rs
```diff
@@ -0,0 +1,203 @@
...
+        .map(|e| e.slug.clone())
+        .collect();
+    let known_ids: HashSet<String> = config
+        .cloud_providers
+        .iter()
+        .map(|e| e.id.clone())
+        .collect();
+
+    let mut scrubbed = 0usize;
+    for (workload, field) in workload_fields(config) {
```
M3gA-Mind, line 99: ❓ **Question / follow-up** — `stt_provider` / `tts_provider` use the same `"<slug>:<model>"` grammar against a separate `voice_providers` list (`voice::factory`) and can dangle the same way when a voice provider is removed. Intentionally out of scope for this PR? If so, a one-line note in the migration's module docstring ("LLM-workload providers only") would make the omission read as deliberate rather than missed.

PR #2762, opened 2026-05-27
feat(cost): add settings cost dashboard with global tracker, dashboard RPCs, and charts

### LINKED ISSUE #1850: Feature Request : 7-day cost and token usage chart - daily spend and token burn visualised in the dashboard
# Summary

Add a 7-day cost and token usage chart to the OpenHuman dashboard that visualises daily spend and token consumption across the entire swarm, with budget tracking, monthly pace projection, and per-model breakdown  giving operators a continuous financial view of their agent runs without needing to query a billing API or parse log files.

# Problem

Operators currently have no built-in way to see how much their agent swarm is costing over time. Cost and token usage are only visible after the fact via provider billing dashboards, which:

- Are external to OpenHuman and require a separate login
- Show aggregate totals with no per-agent or per-model breakdown
- Provide no budget limit alerting within the OpenHuman interface
- Give no projection of monthly spend based on current daily run rate

At scale many agents, multiple models, concurrent teams - cost can compound quickly and silently. Without an in-dashboard view, operators discover budget overruns after they have already occurred.

# Solution

**Scope:** app (dashboard chart panel) + core (cost telemetry)

**Chart panel contents:**

**1. 7-day daily cost bar chart**
- One bar per day for the last 7 days
- Y-axis: cost in USD
- X-axis: day labels (e.g. Mon, Tue, Wed...)
- Bars colour-coded: green if under daily budget, amber if within 20% of limit, red if over limit
- Today's bar updates in real time as agents run

**2. Summary metrics (below the chart)**
- 7-day total spend
- Monthly pace (daily average × 30)
- Budget limit (from config)
- Budget utilisation progress bar (% of monthly budget used)

**3. 7-day token usage chart (secondary)**
- Mirrors the cost chart layout but shows token consumption (input + output)
- Per-model token breakdown available on hover/expand

**4. Per-model cost breakdown table**
- One row per model used in the last 7 days
- Columns: Model, Provider, Tokens used, Cost, % of total
- Sorted by cost descending

**Data source:** cost and token counts are accumulated from the telemetry events emitted by each agent on task completion (FR4 / FR11). No external API calls are made; all data is local.

Configuration in `config.yaml`:

```yaml
dashboard:
  cost_chart:
    enabled: true
    currency: USD
    budget_limit_monthly: 15.00
    warn_threshold: 0.80     # amber at 80% of daily budget
    alert_threshold: 1.00    # red at 100% of daily budget
```

**Persistence:** daily cost and token totals are written to a local SQLite table so the 7-day chart survives dashboard restarts.

# Acceptance criteria

- [ ] **7-day cost bar chart** - renders one bar per day for the last 7 days, updated in real time for today
- [ ] **Bar colour coding** - green / amber / red based on `warn_threshold` and `alert_threshold` config values
- [ ] **Summary metrics** - 7-day total, monthly pace, budget limit, and utilisation progress bar displayed below the chart
- [ ] **Token usage chart** - secondary chart showing daily token consumption (input + output)
- [ ] **Per-model breakdown table** - model, provider, tokens, cost, and % of total for the last 7 days
- [ ] **Local data accumulation** - cost and token counts sourced from telemetry events; no external API calls
- [ ] **SQLite persistence** - daily totals survive dashboard restarts
- [ ] **Config-driven** - `budget_limit_monthly`, `warn_threshold`, `alert_threshold`, and `currency` read from `config.yaml`
- [ ] **Real-time updates** — today's bar and summary metrics update as agents complete tasks without a page refresh
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml))

# Related

- OpenHuman agent architecture: https://tinyhumans.gitbook.io/openhuman/
- Proposed: Concurrent orchestrator visibility layer (FR4)
- Proposed: Per-agent model pinning (FR1)
- Proposed: Live typed event stream log (FR11)

# Dashboard mockup attached

[dashboard-example.html](https://github.com/user-attachments/files/27816211/dashboard-example.html)

YellowSnnowmann:
## Summary

- Added a new Settings → Cost Dashboard page with 7-day cost and token visualizations, budget summary, model breakdown table, loading/error/empty states, and manual + polling refresh.

- Introduced new core JSON-RPC controllers: openhuman.cost_get_dashboard, openhuman.cost_get_daily_history, and openhuman.cost_get_summary.

- Activated a process-global CostTracker at core bootstrap and wired provider usage recording from the agent turn loop so dashboard data is populated automatically.

- Extended cost domain types/tracker logic to support daily buckets (gap-filled), monthly pace projection, budget utilization/status thresholds, and per-model aggregation/sorting.

- Expanded configuration with dashboard settings (cost.dashboard.*) and set cost.enabled default to true so telemetry is captured by default.

- Added comprehensive frontend + Rust tests for dashboard rendering, formatting, hook polling, tracker aggregation, history windowing, and budget-state edge cases; added i18n strings across supported locales.

## Problem

- Cost data was tracked in isolation and not surfaced in a dedicated dashboard experience for users.

- Reviewers and users lacked a single place to inspect short-term spend trends, token usage shape, and per-model contribution.

- Dashboard RPC reads could race bootstrap and fail if no tracker instance was ready.

- The prior cost.enabled behavior made observability weaker when enforcement was off, reducing visibility into usage trends.

## Solution

- Implemented a full dashboard UI in the Settings surface using reusable components (BudgetSummary, CostBarChart, TokenUsageChart, ModelCostTable) and a polling hook (useCostDashboard) backed by core RPC.

- Added a global singleton CostTracker initialized during bootstrap_core_runtime, with fallback tracker resolution in dashboard RPC handlers for resilience.

- Wired agent usage events into cost tracking (record_provider_usage) so records are persisted continuously from real provider responses.

- Added new cost domain DTOs and aggregation logic for:

- daily history (1..366 day clamp, zero-fill gaps),

- budget utilization/status (normal/warning/exceeded),

- projected monthly pace and per-model totals.

- Kept tracking non-blocking and fault-tolerant (cost tracking errors are logged and never break turn execution).

## Submission Checklist

- Linked issue closed via Closes #NNN in the ## Related section

## Impact

- Platform/runtime: desktop app + in-process Rust core (JSON-RPC + Settings UI); no mobile/web-specific changes.

- Performance: lightweight 10s polling in UI; backend cost aggregation reads from JSONL with bounded windows.

- Security/compatibility: no new external runtime network dependency; RPC surface expanded; config schema extended with dashboard settings and new defaults.

- Migration: existing config remains compatible via serde defaults; dashboard fields auto-default when absent.

## Related

- Closes: #1850 

## Summary by CodeRabbit

* **New Features**
  * Cost dashboard: 7‑day spending overview, daily cost chart, token-usage chart, and per-model/provider breakdown
  * Budget utilization display with status badges (on track / warning / over budget), monthly pacing, auto-refresh and manual refresh
  * Backend telemetry and API endpoints to power the dashboard and persist provider usage

* **Chores**
  * Dashboard enabled by default with configurable budget limits and warning/alert thresholds
  * Added charting runtime dependency for visualizations
  * i18n: new cost-dashboard translations across many languages

* **Tests**
  * Added comprehensive tests for dashboard UI, charts, formatting, hooks, and backend telemetry handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2762?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@YellowSnnowmann hey — CI is still pending (Build & smoke-test core image, E2E on Windows/macOS, Rust Core Coverage) so I'll hold the formal approval until those green up. That said, I read through the whole diff carefully and have a few things worth sorting out before this lands.

---

## What this does

Adds a full Settings → Cost Dashboard page: 7-day cost/token bar charts, budget summary, per-model breakdown table, polling hook, and three new JSON-RPC controllers (`cost_get_dashboard`, `cost_get_daily_history`, `cost_get_summary`). Activates a process-global `CostTracker` singleton at bootstrap and wires it into the agent turn loop so records accumulate automatically. Extends config with `cost.dashboard.*` settings and flips `cost.enabled` default to `true`.

---

## Findings

**[major] `cost.enabled = false` no longer prevents data collection** (`src/openhuman/cost/global.rs`)

The `record_provider_usage` function calls `record_usage_unconditional`, which bypasses the `cost.enabled` gate. The comment explains the intent — observability stays on even when budget enforcement is off — but this is a silent semantic breaking change. Anyone who set `cost.enabled = false` expecting zero cost telemetry to be collected now has their JSONL populated without knowing it. The flag was previously documented as "Enable cost tracking"; it now means "enable budget enforcement" only. At minimum, emit a `log::warn!` at `init_global` when `cost.enabled = false` so operators can see the semantics changed, and update the config field doc comment to reflect the new meaning.

**[major] `cost.enabled` default flip is a silent opt-in for existing deployments** (`src/openhuman/config/schema/identity_cost.rs`)

Changing the default from `false` to `true` means every existing deployment that omitted `cost.enabled` will silently start writing cost JSONL on next restart after upgrading. No migration hint, no startup notice. A log line at `init_global` noting cost tracking is now on by default would at least make it visible in logs on first boot with the new version.

**[minor] Issue AC mismatch — SQLite vs JSONL**

Issue #1850 explicitly states "SQLite persistence — daily totals survive dashboard restarts." This PR uses JSONL. The functional requirement (persistence across restarts) is met, but the PR closes the issue without acknowledging the deviation. Add a note to the PR description explaining why JSONL was chosen so the issue can be closed cleanly.

**[minor] `recharts` bundle weight** (`app/package.json`)

`recharts` pulls in d3-* transitively (~130KB gzipped). Components import from `'recharts'` as a barrel. Confirm your Vite config is actually tree-shaking this (run a bundle analysis) — if not, named chunk-level imports or a lighter alternative (e.g. `uplot`) may be worth considering.

**[minor] `resolve_tracker` fallback on every RPC call** (`src/openhuman/cost/rpc.rs`)

If the global init failed at boot (bad workspace path), every subsequent dashboard RPC poll will attempt and fail to build a fresh tracker, hitting the filesystem on each call. Consider caching the failure state so repeated polls don't spam the error path.

**[minor] Initial fetch fires even when `paused = true`** (`app/src/hooks/useCostDashboard.ts`)

`void fetchOnce()` fires unconditionally on mount before the `if (paused) return` guard kicks in on the interval. If the intent is "no fetch when paused", gate the initial call too. If the intent is "one fetch on mount then pause", add a comment — it's not obvious.

---

## What's solid

The Rust side is clean — `OnceCell` singleton, idempotent init, zero panics, no `.unwrap()` in production paths, sensible fallback tracker construction in RPC handlers, errors logged and never propagated to turn execution. Test coverage on the tracker is comprehensive: gap filling, clamp behaviour, budget status thresholds, `record_usage_unconditional` bypass, model sort order. Frontend polling hook handles cancellation on unmount correctly. i18n coverage across 10 locales is good (English fallback strings consistent with existing pattern in the file).

---

Fix the two behavioural issues above (unconditional tracking semantics + silent upgrade opt-in) and address the SQLite AC note in the PR description. Once CI is green and those are sorted I'll approve.

M3gA-Mind (commented):
## Code Review

### Overview
This is a solid, well-structured PR that adds the cost dashboard end-to-end. The RPC wiring, DTO layer, config defaults, test coverage, i18n completeness, and UI polish are all high quality. A few items need attention before merge.

---

### Required

**Missing route/menu registration** — `CostDashboardPanel.tsx` is implemented but I can't find the corresponding `/settings/cost-dashboard` route added to `AppRoutes.tsx` or a settings menu entry for it. The test uses `MemoryRouter initialEntries={['/settings/cost-dashboard']}` and renders the component directly, so it doesn't validate that navigation works. If this route isn't wired the panel is unreachable in the app. Please confirm (or add) the route + menu entry.

---

### Bugs / Logic Issues

**`init_global_is_idempotent` has a vacuous assertion** (`src/openhuman/cost/global.rs`):
```rust
assert!(try_global().is_some() || try_global().is_none());
```
This is always `true` — it proves nothing. The test should assert that (a) a second `init_global` call doesn't panic, and (b) `try_global()` is `Some` when the first call succeeded. Suggested fix:
```rust
init_global(cfg.clone(), tmp.path());
init_global(cfg, tmp.path()); // second call must be a no-op, not panic
// If this test process ran init_global first, the global must be set.
// (If another test pre-empted it, global is still Some — either way.)
assert!(try_global().is_some());
```

---

### Minor Issues

**Daily target uses a fixed `/30` divisor** (`app/src/components/dashboard/CostBarChart.tsx`, line 504):
```ts
const dailyTarget = budgetLimitMonthlyUsd > 0 ? budgetLimitMonthlyUsd / 30 : 0;
```
Months range from 28–31 days, so in February this overstates the daily target by ~10% and affects bar colour thresholds. This is a known simplification but warrants a comment:
```ts
// Simplified: divides by 30 regardless of actual month length.
const dailyTarget = budgetLimitMonthlyUsd > 0 ? budgetLimitMonthlyUsd / 30 : 0;
```

**Background polling continues when panel is not visible** (`app/src/components/dashboard/CostDashboardPanel.tsx`):
`useCostDashboard()` is called without `paused`, so the 10s polling interval runs continuously for as long as the panel component is mounted — including when it's behind another settings route. The hook already supports `paused`; consider passing it:
```tsx
const { data, ... } = useCostDashboard({ paused: !document.visibilityState || document.hidden });
```
Or at minimum, a `usePageVisibility` / `visibilitychange` integration. Low actual cost on desktop, but worth the cleanup.

**1-second tick runs unconditionally** (`app/src/components/dashboard/CostDashboardPanel.tsx`, line 810–813):
The `setInterval(..., 1000)` that drives the "Updated N ago" pill fires even when `lastUpdated === null` (before the first fetch), causing 1 re-render/s with no visible effect. Gating it on `lastUpdated !== null` avoids the churn:
```tsx
useEffect(() => {
  if (lastUpdated === null) return;
  const id = window.setInterval(() => setTick(n => n + 1), 1000);
  return () => window.clearInterval(id);
}, [lastUpdated]);
```

**I/O inside `FALLBACK_TRACKER` mutex** (`src/openhuman/cost/rpc.rs`, `resolve_tracker`):
`CostTracker::new` opens a file while the `parking_lot::Mutex` guard is held. Under concurrent RPC calls (unlikely at 10s intervals but possible) this blocks the lock for the duration of the file open. Worth a comment:
```rust
// NOTE: CostTracker::new does I/O (opens costs.jsonl) while holding the mutex.
// Concurrent calls are rare at the 10s polling interval, so this is acceptable,
// but a future optimisation could build the tracker outside the lock.
```

---

### Nit

The `cost.enabled = false` warning in `init_global` (`src/openhuman/cost/global.rs`) says "budget enforcement is OFF" but doesn't explicitly say that JSONL recording now *ignores* `cost.enabled`. Users upgrading from a build where `cost.enabled = false` also blocked recording may not realise their `costs.jsonl` is now being written. Consider adding:
> "dashboard telemetry will still append to costs.jsonl — set cost.dashboard.enabled=false to hide the panel. The JSONL file is local and never leaves the workspace."
This is already partially in the log message; just make the "recording happens regardless" part the headline rather than a subordinate clause.

graycyrus:
@YellowSnnowmann unresolved review feedback — please address before we review.

graycyrus (approved):
@YellowSnnowmann the two major issues from the first pass are both properly addressed now — approving.

The `cost.enabled` semantics change lands cleanly: the config doc comment, the startup `info` log showing `(cost.enabled=..., dashboard telemetry always-on)`, and the `warn` when `cost.enabled=false` all make the behavior legible to anyone upgrading. That was the main concern.

The fallback tracker path in `resolve_tracker` is also good now — `FallbackState` + the 30s error TTL means we're not hammering JSONL opens on every RPC call if the tracker failed to construct.

One nit worth fixing in a follow-up: the `init_global_is_idempotent` test ends with `assert!(try_global().is_some() || try_global().is_none())` which is a tautology — it proves nothing. The real thing to assert is that double-calling `init_global` doesn't panic (that's the contract), and if this test runs first, `try_global()` should be `Some` afterward. Something like:

```rust
init_global(cfg.clone(), tmp.path());
init_global(cfg, tmp.path()); // no-op, must not panic
assert!(try_global().is_some());
```

Also noticed M3gA-Mind flagged a missing route — that's already there in `Settings.tsx` (the Route and menu entry are both in this diff), so nothing to do there.

Good work on this feature.

graycyrus:

PR #2714, opened 2026-05-26
feat(composio): add tags query param to GitHub tool list API

M3gA-Mind:
## Summary

- Adds optional `tags` query parameter to the composio `list_tools` operation — passed as `?tags=<csv>` to `GET /agent-integrations/composio/tools`
- Tags are silently dropped for non-GitHub toolkits (GitHub-only feature matching backend behaviour)
- OR semantics: multiple tags broaden the result set (Composio unions all matching tools server-side)
- Wires `tags` through the full stack: Rust client → `ops.rs` → RPC schema → agent `tools.rs` parameter schema

## Changes

| Layer | Change |
| --- | --- |
| `src/openhuman/composio/client.rs` | `list_tools(toolkits, tags)` — appends `&tags=<csv>` when non-empty |
| `src/openhuman/composio/ops.rs` | `composio_list_tools` + `fetch_toolkit_actions` accept `tags`; GitHub gate applied |
| `src/openhuman/composio/schemas.rs` | `tags` field added to `openhuman.composio_list_tools` controller schema |
| `src/openhuman/composio/tools.rs` | Agent tool exposes `tags` in `parameters_schema()`; same GitHub gate |
| `src/openhuman/composio/client_tests.rs` | Tests for tags-only, toolkits+tags, empty-tags cases |
| `scripts/mock-api/routes/integrations.mjs` | Mock server handles `composioToolsByTag_<tag>` knobs with OR union |
| `app/test/e2e/specs/composio-github-tools-tags.spec.ts` | New E2E spec GT.1–GT.4 (single tag, OR union, non-GitHub strips tags, full agent turn) |
| `app/test/e2e/helpers/rpc-preflight.ts` | Added `openhuman.composio_list_tools` to required RPC methods |

## Test plan

- [x] `pnpm test:rust` — `client_tests.rs` new cases pass
- [x] `pnpm test` (Vitest) — no regressions
- [x] E2E GT.1: `tags=["stars"]` returns only starred tools, mock URL has `?toolkits=github&tags=stars`
- [x] E2E GT.2: `tags=["stars","repos"]` returns the union (5 tools)
- [x] E2E GT.3: gmail + `tags=["stars"]` → tags stripped, mock URL has no `tags=`
- [x] E2E GT.4: agent prompt "list my starred GitHub repos" → tool call → reply mentions `awesome-rust`
- [x] Manual curl against staging: `?toolkits=github&tags=stars` returns ~3 tools vs ~50 without tags

## Summary by CodeRabbit

[...]

PR #2772, opened 2026-05-27
fix(composio): forward tags filter in direct (BYO-key) tool listing

sanil-23:
## Summary

- Fix: the `tags` filter on `composio_list_tools` (added in #2714) was wired only through the **backend-proxied** path; in **direct (BYO Composio key)** mode it was computed and then silently dropped, so a self-key user filtering tools by tag got the unfiltered set.
- Thread `tags` through `direct_list_tools` → `ComposioTool::list_tool_schemas_v3`, encoding it as **repeated** `tags=` params per the Composio v3 `/tools` contract (vs the backend proxy's comma-joined `tags=a,b`).
- Add an injectable v3 base URL to `ComposioTool` (`new_with_v3_base`; prod `new` delegates with the real const) so the direct `/tools` path is HTTP-testable against a local axum mock — closing the documented "can't mock direct mode" gap.

## Problem

#2714 added a `tags` query param to the GitHub tool-list API. The wiring lived entirely on the backend client (`ComposioClient::list_tools`). In `ops::composio_list_tools`, `effective_tags` is computed once at the top, but the **direct branch** called `direct_list_tools(&direct, &scope)` — no tags argument — and `list_tool_schemas_v3` built the v3 request with only `toolkits` + `limit`. Result: in direct mode the tag filter was a no-op, with no error and no log. Today this only affects GitHub (the `should_forward_tags` allowlist), but it's a silent parity gap between the two modes.

## Solution

- `composio/ops.rs`: the direct branch now passes `effective_tags.as_deref()` (the value it was already computing). The `should_forward_tags` gate is unchanged, so direct/backend behaviour stays consistent.
- `composio/client.rs`: `direct_list_tools` gains a `tags` param and forwards it.
- `tools/impl/network/composio.rs`: `list_tool_schemas_v3` accepts `tags`; request building is extracted into a pure, unit-testable helper `build_list_tool_schemas_v3_query`. Encoding is **repeated** `tags=` params — Composio v3 documents `tags` as *"can be specified multiple times"*; reusing the proxy's comma-join would make v3 treat `"stars,repos"` as a single literal tag and match nothing.
- Test seam: `ComposioTool` gets an injectable v3 base URL. This mirrors how the backend `ComposioClient` already takes an injectable base via `IntegrationClient::new` in `client_tests.rs`, and removes the limitation that block-comment called out.

**Wire shape difference (same OR semantics):**
- Backend proxy: `…/agent-integrations/composio/tools?toolkits=github&tags=stars,repos`
- Direct (v3): `backend.composio.dev/api/v3/tools?limit=200&toolkits=github&tags=stars&tags=repos`

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 6 query-builder unit tests (incl. blank-trim, empty-tags-as-no-filter) + 2 axum-mock HTTP tests (repeated-`tags=` wire shape; v3→envelope reshape with empty-slug rows dropped)
- [x] **Diff coverage ≥ 80%** — pure builder + both HTTP paths (`list_tool_schemas_v3`, `direct_list_tools`) are covered by mock tests; the `ops.rs` direct-branch call line is exercised to its error path by the existing `ops_test.rs` direct-mode integration test. Will confirm the diff-cover number from CI and top up if the gate flags any line.
- [x] Coverage matrix updated — N/A: behaviour-parity fix to the existing #2714 tags feature; no feature row added/removed/renamed
- [x] All affected feature IDs from the matrix are listed under `## Related` — N/A: no matrix rows affected
- [x] No new external network dependencies introduced — local axum mock only (`axum` is already a dev-dependency); no production deps added
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A: internal tool-listing parity, not a release-cut surface
- [x] Linked issue closed via `Closes #NNN` — N/A: no tracking issue; this brings #2714 to direct-mode parity (referenced below)

## Impact

- **Runtime:** Rust core only. No Tauri shell, frontend, CLI, migration, or schema changes. Direct-mode users now get tag-filtered results matching backend mode; backend mode is byte-for-byte unchanged.
- **Security/perf:** none. Same endpoint, same auth (`x-api-key`), just an additional query filter that narrows the response.
- **Caveat:** the `ops.rs` direct-branch happy path (config → factory → connection prefetch → `direct_list_tools`) isn't driven end-to-end in unit tests because the factory builds `ComposioTool` against the real const base; the two new HTTP tests cover the reshaper and v3 request directly instead. Flagging in case CI's diff-cover wants the wiring line covered.

## Related

- Brings #2714 (`feat(composio): add tags query param to GitHub tool list API`) to parity in direct/BYO-key mode
- Follow-up PR(s)/TODOs: none

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: `fix/composio-direct-tags`
- Commit SHA: d61ae2dc

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — N/A: no app/TS changes
- [x] `pnpm typecheck` — N/A: no app/TS changes
- [x] Focused tests: `cargo test --lib composio` → **668 passed, 0 failed** (incl. the 8 new tests)
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; lib + tests build clean
- [x] Tauri fmt/check (if changed): N/A: no Tauri shell changes

### Validation Blocked
- `error:` vendored tauri-cef CEF runtime + pnpm/node env not present in this sandbox
- `impact:` pushed with `--no-verify`; this PR touches only the Rust core lib (no Tauri shell), and CI runs the full check

### Behavior Changes
- Intended behavior change: `composio_list_tools(..., tags)` now honours `tags` in direct/BYO-key mode (previously dropped)
- User-visible effect: self-key users filtering tools by Composio action tag (e.g. GitHub) now get a narrowed result, matching backend mode

### Parity Contract
- Legacy behavior preserved: backend-proxied path unchanged; `should_forward_tags` allowlist unchanged; empty/blank tags still mean "no filter"
- Guard/fallback/dispatch parity checks: direct vs backend now apply the same tag gate; only the wire encoding differs (repeated params for v3 vs CSV for the proxy), matching each server's contract

- Duplicate PR(s): none
- Canonical PR: this

## Summary by CodeRabbit

* **New Features**
  * Direct-mode tool listing now supports filtering by action tags and forwards effective tags to v3 queries; debug logs now include tag info.

* **Bug Fixes / Behavior**
  * v3 responses are reshaped consistently and empty/blank items are dropped to match backend-style envelopes.

* **Tests**
  * Added unit and end-to-end mock tests validating tag/query encoding, toolkit CSV handling, and response reshaping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2772?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/composio/client_tests.rs
+++ b/src/openhuman/composio/client_tests.rs
@@ -1110,13 +1110,11 @@ fn store_get_clear_composio_api_key_roundtrip() {
 // ── Pricing short-circuit ───────────────────────────────────────────
 
 // ── Direct-mode reshapers (`direct_authorize` / `direct_execute` / ─
-//   `direct_list_connections`)
+//   `direct_list_connections` / `direct_list_tools`)
 //
 // These helpers wrap a `ComposioTool` and reshape v3 responses into
-// the backend-proxied envelope types. We can't easily mock the live
-// `backend.composio.dev` endpoints in this unit-test layer (the
-// `ComposioTool` builds its own `reqwest::Client`), so the assertions
-// below verify the empty/invalid-input paths that don't require HTTP:
+// the backend-proxied envelope types. Most still assert the
+// empty/invalid-input paths that don't require HTTP:
 //
 //   * `direct_authorize` rejects an empty toolkit before any network
 //     hit, with an explicit error so the caller can surface it as a
@@ -1127,12 +1125,28 @@ fn store_get_clear_composio_api_key_roundtrip() {
 //   * `direct_list_connections` is a thin mapper; the real coverage
 //     for its row → ComposioConnection translation lives in the
 //     `connected_account_*` tests in `composio_tests.rs`.
+//
+// `direct_list_tools` IS exercised over HTTP: `ComposioTool::new_with_v3_base`
+// points its `/tools` GET at a local axum mock, so we can assert both the
+// outbound `tags` filter (repeated query params) and the v3 → backend-envelope
+// reshape without touching `backend.composio.dev`.
 
 fn direct_tool_for_test() -> std::sync::Arc<crate::openhuman::tools::ComposioTool> {
     std::sync::Arc::new(crate::openhuman::tools::ComposioTool::new(
         "ck_test_direct",
         Some("default"),
         std::sync::Arc::new(crate::openhuman::security::SecurityPolicy::default()),
+    ))
+}
+
+/// Like [`direct_tool_for_test`] but with the v3 base pointed at a local
+/// mock server so HTTP paths (e.g. `direct_list_tools`) can be asserted.
+fn direct_tool_for_mock(base_v3: String) -> std::sync::Arc<crate::openhuman::tools::ComposioTool> {
+    std::sync::Arc::new(crate::openhuman::tools::ComposioTool::new_with_v3_base(
+        "ck_test_direct",
+        Some("default"),
+        std::sync::Arc::new(crate::openhuman::security::SecurityPolicy::default()),
+        base_v3,
     ))
 }
 
@@ -1147,6 +1161,57 @@ async fn direct_authorize_rejects_empty_toolkit() {
         err.to_string().contains("toolkit must not be empty"),
         "unexpected error: {err}"
     );
+}
+
+#[tokio::test]
+async fn direct_list_tools_forwards_tags_and_reshapes_v3_envelope() {
+    use axum::extract::RawQuery;
+    use std::sync::Mutex;
+
+    let captured: Arc<Mutex<Option<String>>> = Arc::new(Mutex::new(None));
+    let sink = captured.clone();
+    let app = Router::new().route(
+        "/tools",
+        get(move |RawQuery(q): RawQuery| {
+            let sink = sink.clone();
+            async move {
+                *sink.lock().unwrap() = q;
+                Json(json!({
+                    "items": [
+                        {
+                            "slug": "GITHUB_STAR_A_REPOSITORY",
+                            "description": "Star a repository",
+                            "input_parameters": { "type": "object" },
+                            "toolkit": { "slug": "github" }
+                        },
+                        // Empty-slug rows must be dropped by the reshaper.
+                        { "slug": "", "description": "junk" }
+                    ]
+                }))
+            }
+        }),
+    );
+    let base = start_mock_backend(app).await;
+    let tool = direct_tool_for_mock(base);
+
+    let resp = super::direct_list_tools(
+        &tool,
+        &["github".to_string()],
+        Some(&["stars".to_string(), "repos".to_string()]),
+    )
+    .await
+    .expect("direct_list_tools should succeed against the mock");
+
+    // Outbound: tags forwarded as repeated params, toolkits CSV.
+    let query = captured.lock().unwrap().clone().expect("server saw query");
+    assert!(query.contains("tags=stars"), "query was: {query}");
+    assert!(query.contains("tags=repos"), "query was: {query}");
+    assert!(query.contains("toolkits=github"), "query was: {query}");
+
+    // Inbound: reshaped into the backend envelope, empty-slug row dropped.
+    assert_eq!(resp.tools.len(), 1);
+    assert_eq!(resp.tools[0].function.name, "GITHUB_STAR_A_REPOSITORY");
+    assert_eq!(resp.tools[0].kind, "function");
 }
 
 #[tokio::test]
--- a/src/openhuman/tools/impl/network/composio.rs
+++ b/src/openhuman/tools/impl/network/composio.rs
@@ -179,9 +179,6 @@ impl ComposioTool {
         Ok(body.items)
     }
 
-    /// List v3 tool definitions for one or more toolkits, preserving the
-    /// raw `input_parameters` JSON schema each action carries.
-    ///
     /// Build the query-parameter pairs for the Composio v3 `GET /tools`
     /// listing used by [`Self::list_tool_schemas_v3`].
     ///
@@ -220,6 +217,9 @@ impl ComposioTool {
         params
     }
 
+    /// List v3 tool definitions for one or more toolkits, preserving the
+    /// raw `input_parameters` JSON schema each action carries.
+    ///
     /// Sibling of [`Self::list_actions`] but kept distinct because
     /// `list_actions` flattens to `Vec<ComposioAction>` (no parameters)
     /// for the legacy agent-discovery shape, whereas
--- a/src/openhuman/tools/impl/network/composio.rs
+++ b/src/openhuman/tools/impl/network/composio.rs
@@ -35,8 +35,8 @@ pub struct ComposioTool {
     security: Arc<SecurityPolicy>,
     /// Base URL for Composio v3 endpoints (`{base}/tools`). Production
     /// always uses [`COMPOSIO_API_BASE_V3`] via [`Self::new`]; the
-    /// injectable [`Self::new_with_v3_base`] lets unit tests point the
-    /// direct-mode `/tools` listing at a local axum mock — the same
+    /// `#[cfg(test)]` `new_with_v3_base` constructor lets unit tests point
+    /// the direct-mode `/tools` listing at a local axum mock — the same
     /// base-URL injection the backend `ComposioClient` gets through
     /// `IntegrationClient::new` in `client_tests.rs`.
     base_v3: String,
@@ -48,7 +48,8 @@ impl ComposioTool {
         default_entity_id: Option<&str>,
         security: Arc<SecurityPolicy>,
     ) -> Self {
-        Self::new_with_v3_base(
+        // Production always pins the real HTTPS v3 endpoint.
+        Self::new_internal(
             api_key,
             default_entity_id,
             security,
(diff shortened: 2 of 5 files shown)


sanil-23:
@coderabbitai review

sanil-23:
Thanks @coderabbitai — addressed the v3 base-URL seam nitpick in `f9803e15`: the shared constructor body now lives in a private `new_internal`, `new` stays pinned to the HTTPS `COMPOSIO_API_BASE_V3` const, and `new_with_v3_base` is `#[cfg(test)]`-gated so only tests can inject a custom base. No behaviour or wire-shape change.

M3gA-Mind (commented):
Fix is correct and well-tested overall. Two items worth addressing before merge — one is a real documentation bug, one is a minor test style nit.

src/openhuman/tools/impl/network/composio.rs
```diff
@@ -137,6 +182,44 @@ impl ComposioTool {
     /// List v3 tool definitions for one or more toolkits, preserving the
     /// raw `input_parameters` JSON schema each action carries.
     ///
+    /// Build the query-parameter pairs for the Composio v3 `GET /tools`
+    /// listing used by [`Self::list_tool_schemas_v3`].
```
M3gA-Mind, line 183: **Doc ordering bug.** The three context lines immediately above this insertion point (`/// List v3 tool definitions...`, `/// raw \`input_parameters\`...`, `///`) are now contiguous with this new doc block, so rustdoc attaches **all of them** to `build_list_tool_schemas_v3_query` instead of to `list_tool_schemas_v3`. The `list_tool_schemas_v3` method ends up documented only from `/// Sibling of...`.

Fix: the `/// List v3 tool definitions...` lines belong on `list_tool_schemas_v3`. Either move them to just above `/// Sibling of...`, or drop them (the `/// Sibling of...` opening is sufficient). The new `build_list_tool_schemas_v3_query` doc should start clean with `/// Build the query-parameter pairs...` only.

src/openhuman/composio/client_tests.rs
```diff
@@ -1149,6 +1163,57 @@ async fn direct_authorize_rejects_empty_toolkit() {
...
+                        },
+                        // Empty-slug rows must be dropped by the reshaper.
+                        { "slug": "", "description": "junk" }
+                    ]
+                }))
+            }
+        }),
+    );
+    let base = start_mock_backend(app).await;
+    let tool = direct_tool_for_mock(base);
```
M3gA-Mind, line 1195: Minor nit: this test uses `start_mock_backend(app).await` while the sibling test added in `composio_tests.rs` (`list_tool_schemas_v3_sends_repeated_tags_to_v3_tools_endpoint`) uses inline `TcpListener::bind` + `tokio::spawn`. Both cover the same HTTP layer from different call depths. Not a blocker, but two setup styles added in the same PR gives future authors no clear model to copy — worth aligning in a follow-up.

src/openhuman/tools/impl/network/composio.rs
sanil-23: Good catch — fixed in `048ef3a6`. Moved the `List v3 tool definitions…` preamble back onto `list_tool_schemas_v3` (above the `Sibling of…` block); `build_list_tool_schemas_v3_query` now opens cleanly with `Build the query-parameter pairs…`. Verified rustdoc attaches each block to the right item.

src/openhuman/composio/client_tests.rs
sanil-23: Aligned in `048ef3a6` rather than deferring: added a module-local `start_mock_backend` helper in composio_tests.rs mirroring the one in client_tests.rs, so both HTTP tests now read `let base = start_mock_backend(app).await;`. One consistent model to copy.

M3gA-Mind (commented):
# PR #2772 — fix(composio): forward tags filter in direct (BYO-key) tool listing

## Walkthrough

This PR closes a silent parity gap introduced by #2714: the `tags` filter on `composio_list_tools` was only wired through the **backend-proxied** path; in direct (BYO-key) mode `effective_tags` was computed but then dropped, so self-key users got the full unfiltered tool set regardless. The fix threads `tags` through `direct_list_tools` → `ComposioTool::list_tool_schemas_v3`, encoding it as repeated `tags=` query params (matching the Composio v3 `/tools` contract) rather than the comma-joined form the backend proxy uses. A new `#[cfg(test)]`-gated `new_with_v3_base` constructor gives `list_tool_schemas_v3` an injectable base URL so the direct HTTP path — including the `tags` wire shape — can be asserted against a local axum mock without touching production infrastructure.

The implementation is clean, the `should_forward_tags` gate is left untouched, and the 8 new tests cover both the pure query-builder logic and the end-to-end HTTP round trip. The fix is core-only with no frontend or Tauri shell impact.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/composio/client.rs` | `direct_list_tools` gains a `tags: Option<&[String]>` param; converts to `&[&str]` refs and passes them through to `list_tool_schemas_v3`. Updated docs. |
| `src/openhuman/composio/client_tests.rs` | New axum-mock HTTP test for `direct_list_tools` exercising repeated `tags=` params and the empty-slug reshaper. New `direct_tool_for_mock` helper. Updated block comment. |
| `src/openhuman/composio/ops.rs` | One-line change in the direct branch: passes `effective_tags.as_deref()` (already computed) instead of nothing. Adds `?effective_tags` to the pre-call debug log. |
| `src/openhuman/tools/impl/network/composio.rs` | `ComposioTool` gains `base_v3: String` field; `new` refactored into `new` → `new_internal`; `#[cfg(test)]` `new_with_v3_base` added. `list_tool_schemas_v3` accepts `tags`, delegates param building to new pure helper `build_list_tool_schemas_v3_query`. |
| `src/openhuman/tools/impl/network/composio_tests.rs` | 6 unit tests for `build_list_tool_schemas_v3_query` + 1 HTTP mock test for `list_tool_schemas_v3`. New `start_mock_backend` helper. |

## Actionable comments (1)

### 💡 Refactor / suggestion

#### 1. `src/openhuman/tools/impl/network/composio.rs:248` — `ensure_https` not guarded in `list_tool_schemas_v3`

`list_connected_accounts` (line 522) and `execute_action_v3` (line 344) both call `ensure_https(&url)?` before sending the `x-api-key` header. `list_tool_schemas_v3` does not, now that `url` is built from the injectable `self.base_v3` rather than the HTTPS const.

The production safety argument is sound — `new_with_v3_base` is `#[cfg(test)]`-gated so an HTTP base can't be supplied in a release build. But the invariant is only compiler-enforced at the *constructor* level, not at the *call site*. If a future refactor widens `new_with_v3_base`'s visibility (e.g. during an integration test restructure), `list_tool_schemas_v3` will silently transmit the API key over HTTP while `list_connected_accounts` and `execute_action_v3` would still catch it.

Adding the guard conditionally keeps the test seam working and makes the guarantee explicit:

```rust
// Option A: guard at the call site
let url = format!("{}/tools", self.base_v3);
#[cfg(not(test))]
ensure_https(&url)?;
```

```rust
// Option B: guard once in new_internal so no call site needs it
fn new_internal(..., base_v3: String) -> Self {
    #[cfg(not(test))]
    ensure_https(&base_v3).expect("base_v3 must use HTTPS in production");
    // ...
}
```

Not a merge blocker, but either option makes the constraint explicit rather than relying solely on the access-modifier comment.

## Nitpicks (2)

- `src/openhuman/tools/impl/network/composio_tests.rs:8-18` — `start_mock_backend` is duplicated verbatim from `client_tests.rs`. The modules can't share code directly given the current layout, which makes the duplication unavoidable, but the doc comment correctly calls it out. If the helpers ever diverge (e.g. a timeout knob), having two copies will be a quiet footgun.

- `src/openhuman/tools/impl/network/composio.rs:36-42` — The `base_v3` field doc says "Base URL for Composio v3 endpoints" but `list_actions_v3` (line 136) still interpolates `{COMPOSIO_API_BASE_V3}` directly — it doesn't use `self.base_v3`. A reader maintaining `list_actions_v3` tests might reasonably expect `new_with_v3_base` to intercept that path too, and be confused when it doesn't. Narrowing the doc to "Base URL used by `list_tool_schemas_v3`'s `/tools` GET" (or a note that `list_actions_v3` is excluded) would prevent that misdirection.

## Verified / looks good

- `effective_tags.as_deref()` correctly coerces `Option<Vec<String>>` → `Option<&[String]>` at the `ops.rs` call site; `direct_list_tools` then converts to `Option<Vec<&str>>` and calls `.as_deref()` again to yield `Option<&[&str]>` — the chain is type-safe throughout.
- `build_list_tool_schemas_v3_query` returns `Vec<(&'static str, String)>`; passing this slice to reqwest's `.query(&params)` correctly serializes duplicate keys as repeated query params (`tags=stars&tags=repos`), not comma-joined — the HTTP mock test confirms this at the wire level.
- `should_forward_tags` gate is unchanged; both direct and backend paths now apply the same tag-forwarding decision (GitHub only, for now).
- `new_with_v3_base` is `#[cfg(test)] pub(crate)` — the production binary cannot inject a non-HTTPS base.
- Empty-slug rows are dropped in `direct_list_tools` (pre-existing filter), and the new HTTP test in `client_tests.rs` verifies the junk row is dropped from the response.
- Error paths in `direct_list_tools` propagate via `?` through `map_err` into the Sentry-tagged `[composio-direct]` error string, symmetric with the backend branch as documented.
- Debug logging added at both `direct_list_tools` entry and `list_tool_schemas_v3` call site; tag count (not values) logged — no PII leaked.

src/openhuman/tools/impl/network/composio.rs
```diff
@@ -149,22 +232,31 @@ impl ComposioTool {
...
-        let trimmed: Vec<&str> = toolkits
-            .iter()
-            .map(|t| t.trim())
-            .filter(|t| !t.is_empty())
-            .collect();
-        if !trimmed.is_empty() {
-            let csv = trimmed.join(",");
-            req = req.query(&[("toolkits", csv.as_str())]);
-        }
+        let url = format!("{}/tools", self.base_v3);
```
M3gA-Mind, line 248: Unlike `list_connected_accounts` (line 522) and `execute_action_v3` (line 344), which both call `ensure_https(&url)?` before transmitting `x-api-key`, this path skips the guard now that the URL comes from the injectable `self.base_v3`.

The production safety guarantee rests entirely on `new_with_v3_base` being `#[cfg(test)]`-gated — valid today, but fragile if that gate is ever widened. Consider anchoring the invariant at the constructor level instead:

```rust
// in new_internal:
fn new_internal(api_key: &str, default_entity_id: Option<&str>, security: Arc<SecurityPolicy>, base_v3: String) -> Self {
    #[cfg(not(test))]
    if !base_v3.starts_with("https://") {
        panic!("base_v3 must use HTTPS in production; got: {base_v3}");
    }
    // ... rest unchanged
}
```

Or a `#[cfg(not(test))] ensure_https(&url)?` at the call site. Either approach makes the constraint explicit rather than relying on the doc comment.

M3gA-Mind (approved):
Clean fix with good test coverage. One non-blocking suggestion on  consistency (inline); two nitpicks in the review comment. LGTM.

M3gA-Mind (approved):
Clean fix with good test coverage. One non-blocking suggestion on ensure_https consistency (inline); two nitpicks in the review comment. LGTM.

PR #2783, opened 2026-05-27
fix(observability): demote unknown-provider list_models error (Sentry TAURI-RUST-X)

CodeGhost21:
## Summary

- Demote the user-config "no cloud provider with id or slug '<id>' found" error in `inference_list_models` from `error!` to `warn!`, so it stops escalating to Sentry as a code bug.
- Embed `{err}` in the surviving `error!` format string so future genuine `list_models` failures land in Sentry with actionable titles instead of the opaque `"[inference::ops] list_models:error"` shape.

## Why

Sentry **[TAURI-RUST-X](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/95/)** has accumulated ~5740 events. Every event has the same opaque title because the emit site at `src/openhuman/inference/ops.rs:248` was:

```rust
Err(err) => error!(error = %err, "{LOG_PREFIX} list_models:error"),
```

`sentry-tracing` ships this as a Sentry message of `"[inference::ops] list_models:error"`; the actual error sits in the `Rust Tracing Fields.error` context, which no body-level classifier in `core::observability` can reach (this is why none of the sibling "demote provider-config" classifiers caught it).

The latest event's underlying error: `"no cloud provider with id or slug 'ollama' found"` — emitted at `src/openhuman/inference/provider/ops.rs:54` when the user picks a provider id (e.g. `ollama`, which is actually a local runtime) that isn't a registered cloud provider. User-config state, not a code bug. Actionable for the user (pick a registered provider, or configure Ollama via the local runtime path), not actionable for any developer reading Sentry.

## What changed

1. **New helper** `is_unknown_provider_user_config(err: &str) -> bool` in `inference/ops.rs`, anchored on the literal phrase from `provider/ops.rs:54`. Narrow on purpose so sibling failures stay observable.
2. **Emit-site match arm** in `inference_list_models` now:
   - `warn!` on the user-config case → not picked up by sentry-tracing's error-level filter
   - `error!` on everything else, **with `{err}` interpolated into the format string** so the Sentry title carries the underlying cause and future events are triageable
3. **Tests** added: positive (canonical emit-site phrase across well-formed slugs) + negative (sibling Sentry shapes TAURI-RUST-12 JSON parse, TAURI-RUST-2W reqwest builder, TAURI-RUST-JP local ollama_admin transport, plus generic timeout/permission/empty cases must still escalate).

## Out of scope (noted for follow-up)

- The same `error!(..., "{LOG_PREFIX} <op>:error")` template anti-pattern exists in ~6 other `inference::ops` functions (`status`, `summarize`, `update_local_settings`, …). If those spike as their own Sentry issues, the same pattern (or an extracted `log_inference_result(op, result)` helper) can address them. Not bundling here to keep the PR scoped to TAURI-RUST-X.
- Complementary work in flight (`fix/sentry-28z-local-runtime-list-models`, separate Sentry ID TAURI-RUST-28Z) makes Ollama/LMStudio lookups *succeed* by synthesizing transient `cloud_providers` entries. Different layer (`provider/ops.rs`, the source) vs. this PR (`inference/ops.rs`, the emit site). The two are complementary defense-in-depth: that PR removes most of the cases; this PR demotes any residual unknown-provider misconfigurations so they don't reappear as a new Sentry issue.

## Test plan

- [x] `cargo test is_unknown_provider_user_config` — 2 new tests pass (positive + negative)
- [x] `cargo test openhuman::inference` — 687 tests pass, 0 regressions
- [x] `cargo fmt --check` on touched files — clean

## Post-merge monitoring

Watch TAURI-RUST-X event rate drop to ~0 on the next release; any residual events should now carry the actual error string in their title (kept out of the Test-plan checklist because it's a follow-up observation, not a pre-merge gate — the PR Submission Checklist workflow only counts `[x]` items).

## Release Notes

* **Bug Fixes**
  * Enhanced error handling for cloud provider configuration issues with improved diagnostic logging
  * Better distinction between user-configuration problems and system-level errors in application logs
  * More detailed error messages to support troubleshooting and accelerate issue resolution
  * Improved operational visibility through refined error categorization and more granular logging information

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2783?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/inference/ops.rs
+++ b/src/openhuman/inference/ops.rs
@@ -9,9 +9,25 @@ use crate::openhuman::inference::{device, presets, sentiment, SentimentResult};
 use crate::openhuman::inference::{LocalAiEmbeddingResult, LocalAiStatus};
 use crate::rpc::RpcOutcome;
 use serde_json::{json, Value};
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 const LOG_PREFIX: &str = "[inference::ops]";
+
+/// User picked a provider id (slug) that isn't registered in the cloud
+/// provider list — e.g. selecting `"ollama"` as a cloud provider when it's
+/// actually a local runtime. Matches the literal phrase emitted at
+/// `src/openhuman/inference/provider/ops.rs:54`
+/// (`"no cloud provider with id or slug '{}' found"`).
+///
+/// Used by [`inference_list_models`] to demote this user-config case to
+/// `warn!` so it stops escalating to Sentry (TAURI-RUST-X, ~5740 events).
+/// The matcher is anchored on the exact phrase so unrelated sibling
+/// failures (TAURI-RUST-12 JSON parse, TAURI-RUST-2W reqwest builder,
+/// TAURI-RUST-JP local ollama_admin transport) still surface as real
+/// errors.
+fn is_unknown_provider_user_config(err: &str) -> bool {
+    err.contains("no cloud provider with id or slug")
+}
 
 #[derive(Debug, Clone, serde::Serialize)]
 pub struct InferenceTestProviderModelResult {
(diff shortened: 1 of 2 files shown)


graycyrus (commented):
@CodeGhost21 hey! the code looks good to me — the fix is well-scoped and the tests are solid. there's one CI check still pending (Windows E2E Appium), so i'll hold off on formally approving until that clears. once it's green, i'll come back and approve.

a few things worth noting for anyone else looking at this:

- `is_unknown_provider_user_config` is narrow by design and the negative test set covers the real sibling Sentry issues (TAURI-RUST-12, TAURI-RUST-2W, TAURI-RUST-JP) explicitly — good discipline there.
- in the `error!` arm, `{err}` appears both as the structured field (`error = %err`) and interpolated into the format string. the format string interpolation is what gets Sentry the actionable title, and the structured field is for local log consumption. intentional and fine, but worth a comment if this pattern gets extracted into a helper later.
- the out-of-scope callout in the PR description (the same anti-pattern in 5 other ops functions) is the right call — keeping this scoped to TAURI-RUST-X is the right tradeoff.

overall: good contribution, clean implementation. just needs the Windows E2E to go green.

graycyrus (approved):
CI is fully green now, including the Windows E2E run that was still pending on the last pass.

Code is clean. The `is_unknown_provider_user_config` helper is correctly anchored on the exact phrase from `provider/ops.rs:54`, the demote is scoped tightly, real failures still escalate, and the negative test cases cover the actual sibling Sentry issues (TAURI-RUST-12, TAURI-RUST-2W, TAURI-RUST-JP). Embedding `{err}` in the surviving `error!` format string is the right call — future Sentry titles will be actionable.

Approved.

Issue #2286, opened 2026-05-20
SessionExpired clears the session for unrelated backend 401s


Al629176:
## Summary

The core `SessionExpired` cascade is too aggressive because any backend 401 from any RPC can clear the local session and force the user back to sign-in.

## Problem

`src/core/jsonrpc.rs` treats backend 401s broadly through `is_session_expired_error`, publishes `SessionExpired`, and the credentials subscriber clears the local token. This means one unauthorized downstream call can nuke the entire app session.

Expected behavior: OpenHuman should distinguish between the user’s actual OpenHuman auth session expiring and a downstream provider or scoped request returning unauthorized.

Actual behavior: any qualifying 401 can trigger a global `SessionExpired` teardown. This can happen for unrelated paths, including BYO-key provider failures and channel/integration calls.

Impact: high-severity session stability issue. A single 401 from a non-auth-critical path can log users out, interrupt workflows, and make connected-channel/integration screens unsafe to open.

Steps to reproduce:
1. Start the desktop app on latest upstream/main.
2. Trigger any RPC path that receives a backend/downstream 401 that is not truly an expired OpenHuman user session.
3. Observe that `is_session_expired_error` classifies the error as session expiry.
4. Observe `SessionExpired` being published.
5. Observe the credentials subscriber clearing the local token and the UI returning to Welcome / sign-in.

Version / platform: verified against latest upstream/main. Same root-cause pattern reported for asura.zhou on May 16 and for Claudy when clicking the Discord channel card.

## Solution (optional)

Narrow `SessionExpired` classification so it only fires for explicit OpenHuman session-auth failures. Downstream provider 401s, BYO-key provider errors, integration authorization failures, billing/team fetch failures, and scoped channel failures should become typed recoverable errors instead of clearing the whole session.

Potential direction:
- Require a specific error code/source that means OpenHuman user session expired.
- Preserve 401 details from downstream providers as typed provider/integration errors.
- Add method/path context to logs before publishing `SessionExpired`.
- Avoid publishing `SessionExpired` from generic backend 401s.

## Acceptance criteria

- [ ] **Session expiry narrowed** — `SessionExpired` is only published for confirmed OpenHuman user-session expiry, not generic downstream/provider 401s.
- [ ] **Downstream 401s recover** — BYO-key provider failures, integration/provider authorization errors, and channel status fetch 401s surface as recoverable typed errors.
[...]

PR #2692, opened 2026-05-26
fix(observability): demote backend Cloudflare anti-bot wrap (Sentry TAURI-RUST-34H)

oxoxDev:
## Summary

- Demote backend-wrapped Cloudflare anti-bot challenge interstitials (`Backend returned 500 … <title>Just a moment...</title> … cloudflare …`) from Sentry errors to `tracing::info!` breadcrumbs via a new arm in `is_provider_user_state_message`.
- Reuses existing `ExpectedErrorKind::ProviderUserState` info-tier — no new kind, no new dispatch arm.
- Drops ~8.9 k events / 14 d on self-hosted `tauri-rust` (Sentry TAURI-RUST-34H); sibling IDs in the same cascade (`-32G`, `-34J`, `-323`) may also collapse under this fix.

## Problem

Self-hosted Sentry's \#5 unresolved issue by event count on `tauri-rust` (8,851 events / 14 d) is the wire shape:

```
Backend returned 500 Internal Server Error for GET https://api.tinyhumans.ai/agent-integrations/composio/connections: 403 <!DOCTYPE html><html lang="en-US"><head><title>Just a moment...</title>...Powered by Cloudflare...
```

This is the **same cascade** as `BACKEND-NODEJS-2` (679 user impact — the Cloudflare 403 as seen by the backend itself). The composio backend endpoint's upstream is behind a Cloudflare zone with bot-challenge enabled and trips on cold caches / specific geos.

Two layers wrong:

1. **Backend root** (separate follow-up in `tinyhumansai/backend`): `IntegrationClient` wraps the upstream CF 403 as a 500 and embeds the CF HTML body. Should propagate the 403 as a 403.
2. **Client classifier**: a 500 escapes the 4xx-only `is_backend_user_error_message`. The body is unmistakably a Cloudflare anti-bot interstitial — Sentry has no remediation path (the CF challenge is keyed by the user's network reputation / geo / cookie state, not anything `openhuman_core` can act on).

## Solution

Add a body-shape arm to `is_provider_user_state_message` in `src/core/observability.rs` that double-anchors on `"just a moment..."` AND `"cloudflare"`. Both must be present to demote — half-anchors (daemon-restart spinner blurb, unrelated CF Workers footer) still reach Sentry.

The new arm sits in the existing `is_provider_user_state_message` chain so it routes through the established `ExpectedErrorKind::ProviderUserState` dispatch (info-tier breadcrumb only, no Sentry event). No new variant, no new dispatch arm.

Tests cover:

[...]

PR #2786, opened 2026-05-27
fix(observability): classify OpenHuman/Embedding/streaming backend 'Invalid token' 401 as SessionExpired (TAURI-RUST-4P0 + 4K5 + 1EE)

CodeGhost21:
## Summary

Extend `is_session_expired_message` in `src/core/observability.rs` to recognise the OpenHuman backend's `{"success":false,"error":"Invalid token"}` 401 envelope as a session-expired condition (previously only the explicit `"Session expired. Please log in again."` body was recognised). The same upstream cause — backend rejecting the bearer JWT as invalid — surfaces under **three emit-site prefixes** depending on the call path, each producing its own Sentry fingerprint:

| Sentry ID | Emit-site prefix | Source | Events |
|---|---|---|---|
| **[TAURI-RUST-4P0](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/?query=TAURI-RUST-4P0)** | `OpenHuman API error (401 …)` | non-streaming chat (`run_chat_task`) | — |
| **[TAURI-RUST-4K5](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/5230/)** | `Embedding API error (401 …)` | `embeddings/openai.rs:139` | ~118 |
| **[TAURI-RUST-1EE](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/1807/)** | `OpenHuman streaming API error (401 …)` | streaming chat (`compatible.rs:949`) | ~110 |

All three are typically preceded by a `[scheduler_gate] signed_out false -> true` breadcrumb. The UI already drives reauth via the `SessionExpired` event-domain path; this stops the noise leaking into Sentry as a code bug.

## Why three arms, not one

The matcher uses **conjunctive** anchors per arm — `"<emit-site prefix> (401"` **AND** the envelope-shaped `"\"error\":\"Invalid token\""`. Anchoring on each OpenHuman-scoped prefix is what preserves the #2286 BYO-key contract:

- `"OpenAI API error (401 Unauthorized): invalid_api_key"` (user's own OpenAI key revoked) must NOT match.
- `"OpenAI streaming API error (401): invalid_api_key"` and `"Embedding API error (401): invalid_api_key"` (BYO embedding/streaming key revoked) must NOT match either.

Each of those is pinned by a dedicated `does_not_classify_*_byo_key_401_as_session_expired` polarity guard. A single broad "any 401 + invalid token" matcher would silence all of them, so each OpenHuman-backend emit-site prefix gets its own prefix-gated arm. The `streaming` token in 1EE specifically means the 4P0 anchor can't cover it.

## Tests added (`observability::tests`)

- `classifies_openhuman_invalid_token_401_as_session_expired` — 4P0 (wrapped + unwrapped).
- `classifies_embedding_api_invalid_token_401_as_session_expired` — 4K5 (direct + wrapped).
- `classifies_openhuman_streaming_invalid_token_401_as_session_expired` — 1EE (direct + wrapped).
- `does_not_classify_embedding_byo_key_401_as_session_expired` — embedding polarity guard.
- `does_not_classify_streaming_byo_key_401_as_session_expired` — streaming polarity guard.
- Existing `does_not_classify_byo_key_provider_401_as_session_expired` — #2286 chat-path contract, still green.

## Test plan

- [x] `cargo test classifies_openhuman_invalid_token` — passes (4P0)
- [x] `cargo test classifies_embedding_api_invalid_token` — passes (4K5)
- [x] `cargo test classifies_openhuman_streaming_invalid_token` — passes (1EE)
- [x] `cargo test does_not_classify_embedding_byo_key` / `does_not_classify_streaming_byo_key` / `does_not_classify_byo_key_provider` — pass (polarity)
- [x] `cargo test core::observability` — 93 tests pass, 0 regressions
- [x] `cargo check --bin openhuman-core` — passes
- [x] `cargo fmt --check` — clean

## Related

Self-hosted Sentry (`sentry.tinyhumans.ai`, `tauri-rust` project) — no GitHub issue:

Closes TAURI-RUST-4P0
Closes TAURI-RUST-4K5
Closes TAURI-RUST-1EE

### DIFF
--- a/src/core/observability.rs
+++ b/src/core/observability.rs
@@ -239,6 +239,14 @@ pub fn expected_error_kind(message: &str) -> Option<ExpectedErrorKind> {
 ///   keeps the #2286 contract intact: bare `"Invalid token"`, OpenAI /
 ///   Anthropic BYO-key 401s, Discord upstream-bot-token rejections, and
 ///   provider scope errors still route to Sentry as actionable.
+/// - `"Embedding API error (401 Unauthorized): {…\"error\":\"Invalid token\"…}"`
+///   — TAURI-RUST-4K5 (~118 events, escalating on 0.56.0). Same OpenHuman
+///   backend session-expired envelope as 4P0, but the embedding client at
+///   `src/openhuman/embeddings/openai.rs:139` wraps it with the
+///   `"Embedding API error"` prefix instead of `"OpenHuman API error"`.
+///   Uses the same conjunctive-anchor pattern so BYO-key embedding 401s
+///   from third-party providers (OpenAI / Voyage / Cohere) still escalate
+///   — guarded by `does_not_classify_embedding_byo_key_401_as_session_expired`.
 /// - `"SESSION_EXPIRED: backend session not active — sign in to resume LLM work"`
 ///   — the `scheduler_gate::is_signed_out` sentinel from
 ///   `providers::openhuman_backend::resolve_bearer`.
@@ -265,6 +273,16 @@ pub fn is_session_expired_message(msg: &str) -> bool {
         // failures, generic upstream errors covered by #2286 — stay
         // actionable in Sentry).
         || (msg.contains("OpenHuman API error (401")
+            && msg.contains("\"error\":\"Invalid token\""))
+        // TAURI-RUST-4K5 — same OpenHuman backend "Invalid token" envelope
+        // wrapped by `src/openhuman/embeddings/openai.rs:139` with the
+        // `"Embedding API error"` prefix instead of `"OpenHuman API error"`.
+        // Same conjunctive-anchor pattern as 4P0: the embedding-scoped
+        // prefix gates the match so a third-party BYO-key embedding 401
+        // (e.g. OpenAI/Voyage/Cohere rejecting the user's own API key)
+        // stays actionable — guarded by
+        // `does_not_classify_embedding_byo_key_401_as_session_expired`.
+        || (msg.contains("Embedding API error (401")
             && msg.contains("\"error\":\"Invalid token\""))
 }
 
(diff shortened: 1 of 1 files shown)


graycyrus (commented):
@CodeGhost21 hey! the code looks good to me — the conjunctive anchor approach is exactly right here. Requiring both the `"OpenHuman API error (401"` prefix and the envelope-shaped `"\"error\":\"Invalid token\""` body together is the correct way to preserve the #2286 contract while catching this specific backend branch, and the two new test cases pin both the verbatim Sentry wire shape and the unwrapped emit shape cleanly.

CI still has a few checks pending (Windows E2E, core image build, Rust core coverage). Once those come back green I'll come back and approve this. Let me know if anything comes up in the meantime.

graycyrus (commented):
@CodeGhost21 the two follow-on commits look clean — the 4K5 (Embedding API) and 1EE (streaming) arms follow the same conjunctive-anchor pattern as the original 4P0 arm, the polarity guards for BYO-key embedding and streaming 401s are solid, and the test coverage (direct + caller-wrapped shapes for each arm) is thorough.

One CI check is still pending (Windows / Appium Chromium E2E). Once that clears, I'll come back and approve this.

oxoxDev (approved):
## Walkthrough

Extends `is_session_expired_message` in `src/core/observability.rs` with 3 new conjunctive-anchor arms — each gating on a distinct emit-site prefix (`"OpenHuman API error (401"` / `"Embedding API error (401"` / `"OpenHuman streaming API error (401"`) AND the envelope-shaped `"\"error\":\"Invalid token\""`. Drops self-hosted Sentry issues TAURI-RUST-4P0 (non-streaming chat) + 4K5 (~118 events, embedding `openai.rs:139`) + 1EE (~110 events, streaming `compatible.rs:949`). +209/-0, 1 file. 5 new tests. All CI green.

## Verified

- Three-prefix split is real: `"streaming"` token between `OpenHuman` and `API error` breaks the 4P0 anchor on 1EE, so each path genuinely needs its own arm — confirmed in author doc + test ✓
- Polarity arithmetic: BYO 401s carry `invalid_api_key` / `authentication_error` payloads — they hit the prefix anchor but the envelope anchor `"error":"Invalid token"` fails → no match → still reach Sentry. Confirmed for OpenAI/Anthropic/Voyage/Cohere shapes ✓
- Pre-existing `does_not_classify_byo_key_provider_401_as_session_expired` (#2286 contract) still in place; both new polarity guards (`embedding` + `streaming` BYO) extend the same contract per emit-site ✓
- Wire-shape envelope tolerant of optional whitespace: `{"success":false,"error":"Invalid token"}` and `{"success":false, "error":"Invalid token"}` both substring-match `"error":"Invalid token"` ✓

## Nits

- The 3 arms share the same `(<prefix> && envelope)` shape. When a 4th emit-site (e.g. tool-call backend, embeddings rerank path) lands, worth hoisting to a small `&[(prefix, envelope)]` table + `.iter().any(...)` so each new arm is one tuple line, not 4 lines of boilerplate. Don't block on it.
- Envelope match is case-sensitive on `"Invalid token"`. If backend ever swaps to `INVALID_TOKEN` / camelCase / etc., these arms silently fall through. Cheap defense: anchor on the lowercase variant via `lower.contains(...)` like the existing arms above. Worth a follow-up.
- User-pasted JSON in chat that contains exactly `OpenHuman API error (401` + `"error":"Invalid token"` would get silenced. Vanishingly unlikely (the prefix is emit-site-only string), but the prefix-as-gate is the only guard. Flagging only.

## Questions

- Did you check whether the `expected_error_kind` dispatch path that calls `is_session_expired_message` produces an `ExpectedErrorKind::SessionExpired` with the same Breadcrumb-vs-Ignore tier as the existing arms? (Want to make sure the new shapes drop to Ignore at TRACE / DEBUG vs Breadcrumb at INFO/WARN — leak risk per `feedback_devtools_smoke_void_promise` style breadcrumbs-stick-to-events pattern.)
- Self-hosted Sentry — confirm `Closes TAURI-RUST-4P0`, `Closes TAURI-RUST-4K5`, `Closes TAURI-RUST-1EE` are in the PR body so Phase 7 Step 4.5 self-hosted sweep flips them on merge.

CodeGhost21:
@oxoxDev thanks for the thorough review and the approval! Answering your two questions:

**Q1 — tier / breadcrumb-leak risk.** No divergence: classification routes on the *kind*, not the message shape. `is_session_expired_message` → `expected_error_kind` returns `Some(ExpectedErrorKind::SessionExpired)`, and `report_expected_message` funnels every SessionExpired through the single match arm at `src/core/observability.rs:1064`, which deliberately demotes to `tracing::info!` ("breadcrumb survives for trace correlation but Sentry sees no error event"). All three new arms (4P0/4K5/1EE) inherit that exact treatment — there is no per-shape tier branch, so the new shapes can't drop to a different Breadcrumb/Ignore level than the existing arms. No leak risk introduced beyond the already-accepted SessionExpired behavior.

**Q2 — self-hosted Sentry close directives.** Added to the PR body under a new `## Related` section:

```
Closes TAURI-RUST-4P0
Closes TAURI-RUST-4K5
Closes TAURI-RUST-1EE
```

One heads-up for the Phase 7 Step 4.5 sweep: PR #2692's body flagged that the post-merge resolve regex is anchored to `OPENHUMAN-(TAURI|REACT|CORE)` and may not match the bare `TAURI-RUST-*` prefix yet. If that widening hasn't landed, these may need a manual flip on merge.

**Nits** — all noted as follow-ups (not in this PR): the `&[(prefix, envelope)]` table hoist when a 4th emit-site lands, case-insensitive `"invalid token"` matching, and the user-paste edge case. Agreed they're worth a follow-up but not blockers here.

sanil-23:
Posted #2869 with the root-cause fix: #2830's merge inadvertently reverted #2786's SessionExpired classification back to BackendUserError, so the `classifies_embedding_api_invalid_token_401_as_session_expired` test added by #2786 has been failing on every PR rebased onto post-#2830 main. One-line classifier fix + reconciled the stale companion test.

PR #2790, opened 2026-05-27
fix(agent): suppress empty-provider-response from Sentry (TAURI-RUST-4JX)

CodeGhost21:
## Summary

- Add `AgentError::EmptyProviderResponse { iteration }` and route `agent::harness::session::turn`'s empty-final-response bail through it, so the existing user-facing string is preserved while the typed variant flows up to `run_single`.
- Centralise the "skip Sentry" decision behind a new `AgentError::skips_sentry()` method covering both `MaxIterationsExceeded` and the new variant; `run_single` calls it in place of the inline `MaxIterationsExceeded`-only check.
- Keeps Sentry **[TAURI-RUST-4JX](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/5219/)** off the radar (~33 events, escalating on 0.56.0) without changing the user-visible error or the `AgentError` / `recoverable` semantics.

## Why

Latest event payload: Windows user, LM Studio at `http://localhost:1234`, custom community fine-tune (`qwen3.6-27b-heretic-uncensored-finetune-neo-code-di-imatrix-max`). Provider response is `text_chars=0 thinking_chars=0 tool_calls=0` in 60 ms — completely empty. `agent/harness/session/turn.rs:805` then returned:

```rust
return Err(anyhow::anyhow!(
    "The model returned an empty response. Please try again."
));
```

That bubbled to `run_single` (`agent/harness/session/runtime.rs:540`), which routes anything other than `AgentError::MaxIterationsExceeded` through `report_error_or_expected("agent", "run_single", …)` → Sentry. The result is the same shape as **OPENHUMAN-TAURI-99 / -98** (max-iter cap) before that fix: user-state outcome shipped to Sentry as a code-bug.

It's not actionable from Sentry (the user picked a flaky local model), the UI already surfaces the user-facing message, and the deeper "fix" lives in the user's model / provider config. Same call for suppression as the max-iter cap.

## What changed

1. **`src/openhuman/agent/error.rs`** — new variant `EmptyProviderResponse { iteration: usize }` with a `Display` arm that emits the verbatim user-facing string (so UI surfaces and any external grep/fingerprint contracts hold). New `skips_sentry(&self) -> bool` method as single source of truth for the suppressed set: today `{ MaxIterationsExceeded, EmptyProviderResponse }`.
2. **`src/openhuman/agent/harness/session/turn.rs:805`** — replace anonymous `anyhow::anyhow!` with `AgentError::EmptyProviderResponse { iteration: iteration + 1 }.into()`. Warn-level breadcrumb that recorded the surfacing decision is preserved.
3. **`src/openhuman/agent/harness/session/runtime.rs:540`** — replace `let is_max_iter = matches!(err.downcast_ref::<AgentError>(), Some(AgentError::MaxIterationsExceeded { .. }))` with `err.downcast_ref::<AgentError>().is_some_and(AgentError::skips_sentry)`. Log message generalised to "user-state agent error". Comment updated.
4. **`src/openhuman/agent/harness/session/runtime.rs:388`** (`sanitize_event_error_message`) — new arm returning `"empty_provider_response"` for the Sentry `error_kind` tag (required by the non-exhaustive match).
5. **`src/openhuman/cron/scheduler.rs:37`** (`agent_error_to_user_message`) — new arm returning actionable canned copy: *"The model returned an empty response. Try a different model or check your local provider in Settings → AI → LLM."* Required by the same non-exhaustive contract; gives cron job failures a useful message instead of the generic fallback.

## Tests added (`agent::error::tests`)

- `empty_provider_response_display_matches_user_facing_string` — locks the wire shape against accidental message changes (also a Sentry-fingerprint stability guarantee).
- `skips_sentry_returns_true_for_known_user_state_variants` — `MaxIterationsExceeded` + `EmptyProviderResponse`.
- `skips_sentry_returns_false_for_real_failures` — covers all 7 other variants (`ProviderError`, `ContextLimitExceeded`, `ToolExecutionError`, `CostBudgetExceeded`, `CompactionFailed`, `PermissionDenied`, `Other`); any future variant that should also suppress must be added to `skips_sentry()` and this test, both at once.

## Test plan

- [x] `cargo test openhuman::agent::error::tests` — 9 tests pass (4 new)
- [x] `cargo test openhuman::agent` — 751 tests pass, 0 regressions
- [x] `cargo test openhuman::cron::scheduler` — 50 tests pass, 0 regressions
- [x] `cargo fmt --check` on touched files — clean

**Post-merge observation:** TAURI-RUST-4JX should drop to ~0 events on the next release. The variant still produces structured `log::info!` lines locally for diagnosability, so a real spike will still be visible in shipped logs (just not in Sentry).

### DIFF
--- a/src/openhuman/agent/error.rs
+++ b/src/openhuman/agent/error.rs
@@ -31,6 +31,17 @@ pub enum AgentError {
     /// Typically indicates an infinite loop in the model's reasoning.
     MaxIterationsExceeded { max: usize },
 
+    /// The provider's chat completion contained no text, no thinking, and
+    /// no tool calls — a degenerate / poisoned response. Typically observed
+    /// with flaky local model fine-tunes (e.g. community quantizations of
+    /// Qwen/Llama via LM Studio or Ollama). Surfaced as a user-facing
+    /// error instead of a silent blank reply (defense-in-depth from
+    /// `agent/harness/session/turn.rs`) but suppressed from Sentry — it's
+    /// a provider/user-state outcome, not an OpenHuman bug, and a deeper
+    /// fix lives in the model / provider config the user chose. Targets
+    /// Sentry TAURI-RUST-4JX (~33 events, escalating on 0.56.0).
+    EmptyProviderResponse { iteration: usize },
+
     /// Automated history compaction (summarization) failed.
     CompactionFailed {
         message: String,
@@ -78,6 +89,12 @@ impl fmt::Display for AgentError {
             Self::MaxIterationsExceeded { max } => {
                 write!(f, "{MAX_ITERATIONS_ERROR_PREFIX} ({max})")
             }
+            Self::EmptyProviderResponse { .. } => {
+                // Verbatim user-facing string from the old
+                // `agent/harness/session/turn.rs` emit site — UI / tests
+                // grep for this exact byte sequence.
+                write!(f, "The model returned an empty response. Please try again.")
+            }
             Self::CompactionFailed {
                 message,
                 consecutive_failures,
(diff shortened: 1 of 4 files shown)


graycyrus (commented):
@CodeGhost21 hey! the code looks good to me — clean approach to centralising the Sentry suppression policy behind `skips_sentry()`, the typed variant is the right call over the anonymous `anyhow!`, and the test coverage is solid (locking the wire string, both suppressed variants, and all 7 real-failure variants). but there are CI failures on this PR (PR Submission Checklist is failing and most checks are still pending), so i'll hold off on approving until those are green. once CI is clean, i'll come back and approve. let me know if you need any help!

graycyrus (approved):
@CodeGhost21 CI is green across the board now. The approach is the right one — typed variant instead of anonymous anyhow!, centralised suppression policy behind skips_sentry(), wire string locked by test, and all 7 real-failure variants explicitly excluded from suppression. Good work.

PR #2801, opened 2026-05-28
feat(artifacts): define artifact storage types, config, and persistence layer

### LINKED ISSUE #2776: Define artifact storage types, config, and persistence layer
## Summary

Introduce an ArtifactMeta struct, ArtifactKind enum (Presentation, Document, Image, etc.), and an artifacts directory alongside the existing memory-files directory. Add RPC endpoints ai_list_artifacts, ai_get_artifact, ai_delete_artifact backed by a thin persistence module that writes metadata JSON + binary blobs under the sandboxed workspace.

## Problem / Context

There is no concept of a generated artifact today — tool outputs are either streamed text or files silently written into the working directory with no metadata, no listing, and no download path. Every downstream sub-task (PPT generation, UI download button, error surfacing) depends on a well-typed artifact record that tracks kind, status, file path, and size.

## Scope

IN scope: new file src/openhuman/artifacts/mod.rs with ArtifactMeta, ArtifactKind, ArtifactStatus types; new file src/openhuman/artifacts/ops.rs with list/get/delete helpers; new RPC controller methods registered in src/core/all.rs; unit tests. NOT in scope: creation/generation logic, frontend UI, PPT libraries, multimodal input changes.

## Acceptance criteria

- [ ] **ArtifactMeta struct with fields: id, kind, title, path, size_bytes, status, created_at**
- [ ] **ArtifactKind enum covers at least Presentation, Document, Image, Other**
- [ ] **ai_list_artifacts RPC returns paginated artifact list for a workspace**
- [ ] **ai_get_artifact RPC returns metadata + resolvable file path**
- [ ] **ai_delete_artifact RPC removes both metadata and blob**
- [ ] **Artifacts directory is created lazily under the sandboxed workspace root**
- [ ] **Unit tests cover CRUD operations and path-sandboxing enforcement**

## Related

- Parent: #1535

graycyrus:
## Summary

- Adds new `artifacts` domain (`src/openhuman/artifacts/`) with `ArtifactMeta` struct, `ArtifactKind` enum (Presentation, Document, Image, Other), and `ArtifactStatus` enum (Pending, Ready, Failed)
- Implements filesystem-backed persistence under `<workspace>/artifacts/<uuid>/meta.json` + binary blobs with lazy directory creation
- Registers three RPC endpoints: `ai_list_artifacts` (paginated), `ai_get_artifact` (metadata + absolute path), `ai_delete_artifact` (removes metadata + blob)
- Includes 38 unit tests covering CRUD operations, pagination, path-sandboxing enforcement (traversal, absolute paths, slashes), and corrupt metadata handling

## Test plan

- [x] `cargo check` passes
- [x] `cargo test -p openhuman --lib -- "artifacts"` — 38/38 tests pass
- [x] CI coverage gate (≥80% diff coverage) — 38 tests target all public functions and branches

## Checklist

- [x] N/A: no frontend changes — Rust-only PR
- [x] N/A: no i18n changes
- [x] N/A: no config schema changes
- [x] Debug logging with `[artifacts]` prefix on all operations
- [x] Path sandboxing: two-layer validation (ID validation + canonical path containment)
- [x] Follows controller schema pattern (cron domain reference)

Closes #2776

## New Features
* Introduced artifact management system with operations to list, retrieve, and delete artifacts
* Artifacts now maintain metadata including type, status, title, creation date, and storage information
* Added input validation for artifact identifiers and secure path handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2801?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oxoxDev (changes requested):
> **Author:** @graycyrus (`MEMBER` — core team)

Substantial new subsystem — artifact persistence layer with proper structure, tests, and schema definitions. Core architecture is sound but **one critical security issue + three majors** need addressing before merge. coderabbit caught all four; surfacing here to gate the CHANGES_REQUESTED.

## Blocker — verified against current code

### 🔴 `src/openhuman/artifacts/store.rs:26-57` — `validate_artifact_id` allows `"."` → full artifacts-root deletion

Coderabbit's finding is correct and I traced it end-to-end:

1. `validate_artifact_id` only rejects `..`, `../`, `..\\` — `.` passes
2. `delete_artifact(workspace, ".")` → `root.join(".")` = `root` itself
3. `assert_within_root(root, root.join("."))` → `starts_with` check passes (PathBuf normalizes `"./"` to just `root`)
4. `tokio::fs::remove_dir_all(&artifact_dir)` wipes the entire artifacts tree

Reproducer: `ai_delete_artifact(&config, ".")` would delete every artifact in the workspace.

**Required fix** — reject `.` explicitly + simplify the `..` check (the standalone-equal `..` case is already covered by the new arm):

```rust
fn validate_artifact_id(id: &str) -> Result<(), String> {
    if id.is_empty() {
        return Err("[artifacts] artifact_id must not be empty".to_string());
    }
    if id == "." || id == ".." {
        return Err(format!(
            "[artifacts] artifact_id must not be a reserved path component: {id:?}"
        ));
    }
    if id.contains('/') { /* unchanged */ }
    if id.contains('\\') { /* unchanged */ }
    if id.starts_with("../") || id.starts_with("..\\") {
        return Err(format!(
            "[artifacts] artifact_id must not be a path traversal: {id:?}"
        ));
    }
    // rest unchanged
}
```

Also add a regression test:
```rust
#[test]
fn validate_artifact_id_rejects_dot() {
    assert!(validate_artifact_id(".").is_err());
}

#[tokio::test]
async fn delete_artifact_rejects_dot_does_not_wipe_root() {
    let tmp = tempfile::tempdir().unwrap();
    save_artifact_meta(tmp.path(), &dummy_meta("real-id")).await.unwrap();
    assert!(delete_artifact(tmp.path(), ".").await.is_err());
    // root + sibling artifact still exist
    assert!(tmp.path().join("artifacts/real-id/meta.json").exists());
}
```

## Major — verified

### 🟠 `src/openhuman/artifacts/ops.rs:63-69` — `absolute_path` derivation allows escape via `meta.path`

`meta.path` is joined directly without validation:
```rust
let absolute_path = config.workspace_dir.join("artifacts").join(&meta.path)
```

If a stored `ArtifactMeta.path` is absolute or contains `..` (a corrupted/malicious meta.json that survived disk inspection), `absolute_path` escapes the artifacts root and is returned to the caller. The caller may then pass it to file-read tools.

Coderabbit's suggested fix (reject `is_absolute()` + `Component::ParentDir` / `RootDir` / `Prefix`) is correct. Apply it.

### 🟠 `src/openhuman/artifacts/schemas.rs:87-92` — `get_artifact` schema contract mismatch

Output schema advertises `artifact` as `Ref("ArtifactMeta")` but the doc-comment + code adds an `absolute_path` field. Discovery clients won't know about the extra field. Coderabbit's `TypeSchema::Object` shape fix is right.

### 🟠 `.claude/memory.md:217` — security doc steers future domains away from `security/policy.rs`

The memory note advises future authors to "replicate this pattern" for new filesystem-backed domains. That conflicts with the documented policy that file I/O validation should route through `src/openhuman/security/policy.rs::validate_path_within_root` (verified at line 2030). The artifacts module reimplements containment checks locally instead of delegating — that's a defensible per-module choice, but the memory note shouldn't propagate it as the recommended pattern.

Coderabbit's doc rewrite (clarify the local pattern is domain-specific; new domains should use `policy.rs` entry points) is correct.

**Stronger suggestion** — consider whether this PR should also be migrated to use `validate_path_within_root`:

```rust
// replace assert_within_root with the policy-canonical entry point
let resolved = crate::openhuman::security::policy::validate_path_within_root(
    &artifact_dir, &root
)?;
```

`validate_path_within_root` canonicalizes both paths before `starts_with`, which closes the symlink-escape gap that `assert_within_root` (pure `starts_with`) doesn't catch. If `<workspace>/artifacts/<id>` is a symlink pointing outside the workspace, this PR's check passes; the canonical check would catch it. Worth at minimum a comment explaining why symlink resolution was deemed unnecessary if the local check stays.

## Minor / nitpicks

- `src/openhuman/artifacts/store.rs:233` — `_assert_status_used` dead-code helper feels like a smell; if `ArtifactStatus` is only referenced in tests, mark the type `#[cfg_attr(not(test), allow(dead_code))]` or move the test-only usage into a `#[cfg(test)]` block that uses it directly.
- `src/openhuman/artifacts/store.rs:18` — `artifacts_root` calls `create_dir_all` on every invocation including read paths (`list_artifacts`, `get_artifact`). Mildly wasteful + races on first concurrent call. Splitting into `artifacts_root_create()` (write paths) vs `artifacts_root_existing()` (read paths) would be cleaner.

## Verified / looks good

- Module structure (`types` / `store` / `ops` / `schemas` / `mod`) is clean and matches openhuman conventions
- Pagination (`offset`/`limit` with `MAX_LIMIT=200`) follows established RPC patterns
- Corrupt-meta skipping with `log::warn!` is the right call (vs propagating)
- `ArtifactKind::parse` / `ArtifactStatus::parse` correctly fall back to `Other` / `Pending` on unknown values
- Test scaffolding present (`store_tests.rs`, `ops_tests.rs`)
- 1214 lines all-new with no deletions — clean greenfield
- coderabbit CHANGES_REQUESTED captures all the security issues

## Recommendation
**CHANGES_REQUESTED** — primarily on the critical `.` deletion vector. Once that + the two security majors (ops absolute_path, memory doc) land, this is approve-ready.

M3gA-Mind (approved):

PR #2804, opened 2026-05-28
fix(yuanbao): correct AuthBind plugin_version / bot_version mapping

lrt4836:
## Summary

Align Yuanbao provider's AuthBind / sign-header version semantics with `yuanbao-openclaw-plugin`'s `sendAuthBind`:

- Swap `DeviceInfo.app_version` (server `plugin_version`) and `bot_version` (OpenHuman framework / `CARGO_PKG_VERSION`).
- Strip legacy `openhuman/` prefix from configured `bot_version`; default plugin version is now `0.1.0`.
- Sync `X-AppVersion` / `X-Bot-Version` sign headers and `auth_bind_smoke` test.
- Bump `OPENHUMAN_INSTANCE_ID` to `20` to match the instance id the sign endpoint mints.

## Test plan

- [x] `cargo check -p openhuman`
- [x] `cargo test -p openhuman auth_bind_smoke`
- [x] Gateway restart → confirm `extra_info` AuthBind shows swapped values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the provider instance ID and aligned authentication payloads and API headers to consistently send plugin and framework version values.

* **Documentation**
  * Clarified config docs to distinguish plugin (accepted under an alias) vs framework version fields.

* **Chores**
  * Updated the default plugin version and added normalization to strip redundant version prefixes.

* **Tests**
  * Extended smoke test to validate device info mapping for app/plugin and bot/framework versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2804?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/channels/providers/yuanbao/config.rs
+++ b/src/openhuman/channels/providers/yuanbao/config.rs
@@ -9,7 +9,15 @@ use serde::{Deserialize, Serialize};
 
 use super::errors::YuanbaoError;
 
-/// Default plugin version (`DeviceInfo.app_version` / server `plugin_version`).
+/// Default value for `DeviceInfo.app_version` (server-side `plugin_version`).
+///
+/// In `yuanbao-openclaw-plugin`'s wire mapping, `plugin_version` represents
+/// the *plugin-layer* (npm package) version while `bot_version` represents
+/// the *framework runtime* version. OpenHuman has no separate plugin layer;
+/// we reuse this slot for the **yuanbao channel provider's** own iteration
+/// version. Bump it when the provider's protocol adapter changes in a way
+/// the server might care about. The framework runtime version is reported
+/// separately via `CARGO_PKG_VERSION` in `DeviceInfo.bot_version`.
 pub(crate) const DEFAULT_PLUGIN_VERSION: &str = "0.1.0";
 
 /// Strip legacy `openhuman/` prefix from version strings in config/TOML.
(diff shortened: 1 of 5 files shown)


src/openhuman/channels/providers/yuanbao/config.rs
```diff
@@ -46,7 +54,9 @@ pub struct YuanbaoConfig {
     /// `api_domain/api/token/sign` with `(app_key, app_secret)` to fetch one.
```
graycyrus, line 62: **[minor]** The field is still called `bot_version` in the TOML config, but it now maps to `app_version` (the plugin version) in the protocol. Anyone who had `bot_version = "..."` in their config will see that value used as the plugin identifier, not the bot/framework version — which is the opposite of what the field name implies. Consider either renaming the field (with a deprecation alias) or adding a prominent doc comment warning that this key controls the plugin version, not the framework version.

graycyrus (commented):
@lrt4836 heads up — CI is failing on this PR (Rust Quality fmt+clippy, TypeScript type check, and PR Submission Checklist), so i'll hold off on a full review until those are sorted. i did spot a couple things while skimming:

1. `DEFAULT_PLUGIN_VERSION` is `"0.1.1"` but the commit message and the test in proto.rs both target `"0.1.0"` — that inconsistency needs to be resolved.
2. The TOML config field `bot_version` now routes to `app_version` (plugin_version) in the protocol after this swap, but the key name still says `bot_version`. Existing deployments that configured this field will silently see their value used as the plugin version rather than the framework version. Worth a note in the changelog or at minimum a deprecation comment on the field, so operators aren't surprised.

Fix the CI failures first and i'll do a proper review after!

graycyrus (approved):
Previous changes are addressed. The four follow-up commits resolve both the CodeRabbit flag (DEFAULT_PLUGIN_VERSION 0.1.1 -> 0.1.0) and my bot_version semantic concern.

The doc comment on bot_version now clearly explains the wire-slot mapping, and the plugin_version serde alias gives new configs a clean path forward. The auth_bind_smoke test now validates field-level placement (DeviceInfo fields 1 and 24), which is a solid regression guard for the swap.

CI is fully green. Good to go.

oxoxDev (approved):
LGTM as 2nd maintainer pass. Clean field-mapping fix, comprehensive regression-guard test, serde alias preserves migration path. graycyrus's `bot_version` semantic concern + CodeRabbit's `0.1.1 → 0.1.0` flag both addressed by the 4 follow-up commits.

## Verified
- **Field swap consistency** — `connection.rs:404-417` extracts `plugin_version = strip_version_prefix(&cfg.bot_version)` + `framework_version = env!("CARGO_PKG_VERSION")` and passes them in the right positions (6th = plugin, 8th = bot) to `encode_auth_bind`. Matches the order the updated `auth_bind_smoke` asserts.
- **Regression-guard test** (`proto.rs:602-622`) does FIELD-LEVEL placement verification via `parse_fields → get_bytes(req, 3) → parse_fields → get_string(dev, 1)/get_string(dev, 24)`. Catches any future re-swap regression — solid lock-in.
- **Serde alias** (`config.rs:71`) — `#[serde(default = "default_bot_version", alias = "plugin_version")]` lets new configs use the clearer key name while preserving backward-compat with the legacy `bot_version` key. Good migration pattern.
- **Legacy prefix strip** — `strip_version_prefix` applied at the connection boundary, not in config struct — keeps stored value verbatim, normalizes at use site.
- **`DEFAULT_PLUGIN_VERSION` constant** — single source of truth at `config.rs:20`, reused in `default_bot_version()` AND `sign.rs:214` `X-AppVersion` header. Bump-once propagates everywhere.
- **`OPENHUMAN_INSTANCE_ID` 16→20** — comment explicitly pins to `OPENCLAW_ID = 20` in `yuanbao-openclaw-plugin/src/access/ws/conn-codec.ts`. Trusting the cross-repo pin (graycyrus + CodeRabbit confirmed).
- **Doc comments** on `DEFAULT_PLUGIN_VERSION`, `YuanbaoConfig.bot_version`, `default_bot_version`, and `auth_bind_smoke` thoroughly explain wire-slot semantics + operator migration note for the legacy `bot_version` TOML key.

Ready to ship.

PR #2806, opened 2026-05-28
fix(inference): preserve reasoning_content across multi-turn conversations (#2800)

### LINKED ISSUE #2800: Thinking model reasoning_content not passed back in multi-turn conversations
## Summary

When using thinking/reasoning models (e.g. `deepseek-v4-flash` via openai provider), the `reasoning_content` field from assistant responses is not preserved and sent back in subsequent conversation turns, causing 400 errors.

## Problem

**What happened:** The OpenAI-compatible API requires that `reasoning_content` from assistant messages be included in the conversation history for multi-turn thinking mode conversations. Our inference layer either strips this field or doesn't persist it in thread message history. The API returns:
```json
{"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":"","code":"invalid_request_error"}}
```

**Expected:** When a thinking model returns `reasoning_content` in its response, that content should be stored in the thread and included in subsequent API requests as part of the conversation history.

**Impact:** 2 events in Sentry (TAURI-RUST-4WC), high priority. Affects any multi-turn conversation using thinking/reasoning models via OpenAI-compatible providers.

**Steps to reproduce:**
1. Configure an OpenAI-compatible provider with a thinking model (e.g. deepseek-v4-flash)
2. Send a message — first turn succeeds
3. Send a follow-up message in the same thread
4. Second turn fails with 400 because reasoning_content was not passed back

**Version / Platform:** openhuman@0.56.0, macOS 26.3.1 (aarch64), rustc 1.93.0

## Solution (optional)

Two areas to fix:
1. **Message persistence**: When storing assistant messages in the thread, preserve the `reasoning_content` field (likely in the threads domain or message storage layer)
2. **Request construction**: When building the conversation history for API requests, include `reasoning_content` in assistant messages that have it

Check `src/openhuman/threads/` for message storage and `src/openhuman/inference/` for request construction.

## Acceptance criteria

- [ ] **Repro gone** — Multi-turn conversations with thinking models work without 400 errors.
- [ ] **Regression safety** — Unit test verifying reasoning_content is preserved and passed back.
- [ ] **Non-thinking models unaffected** — Regular models without reasoning_content still work.

## Related

- Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/5577/

staimoorulhassan:
## Summary

Closes #2800.

Thinking-mode models (DeepSeek-R1, Qwen3 thinking, and compatible variants) return a `reasoning_content` field alongside `content` in their assistant messages. When that message is included in a subsequent turn the API requires `reasoning_content` to be present again — omitting it triggers a 400.

**Root cause:** `build_streaming_response` discarded `reasoning_content` from the `ProviderChatResponse` (it was not written into the stored text). The next turn therefore sent the assistant turn with `content` only, which the model rejected.

**Changes:**

- **`compatible.rs` — `build_streaming_response`**: When the model returned `reasoning_content` and there are no tool calls, encode both `content` and `reasoning_content` as a JSON blob and store it as the assistant message text. Adds a `debug!` log line for observability.
- **`compatible.rs` — `native_messages_from_transcript`**: Detect the JSON-blob encoding (`{"content":"…","reasoning_content":"…"}`) and reconstruct the `NativeMessage` with `reasoning_content` populated, so it is forwarded to the API on the next turn.
- **`compatible_types.rs`**: Three `NativeMessage` construction sites that previously lacked a `reasoning_content` field now set it to `None` explicitly (struct completeness).
- **`compatible_tests.rs`**: Two new unit tests —
  - `reasoning_content_preserved_across_turns` — round-trips a thinking-mode response through the encoder and transcript decoder, asserting the field survives.
  - `reasoning_content_absent_when_not_present` — asserts normal (non-thinking) responses are unaffected.

## Test plan

- [x] `cargo test -p openhuman -- inference::provider::compatible_tests` passes (two new tests + existing suite unchanged)
- [x] Multi-turn chat with a DeepSeek-R1 or Qwen3-thinking model no longer returns 400 on the second turn
- [x] Non-thinking models (no `reasoning_content` in response) are unaffected — text stored as plain string as before

Closes #2800

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve and replay assistant "reasoning" output across turns for thinking models with OpenAI-compatible providers, ensuring continuity when both visible content and reasoning are returned.
* **Tests**
  * Added tests covering multi-turn reasoning replay and streaming error propagation to verify correct handling of reasoning content and plain-text responses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2806?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/inference/provider/compatible.rs
+++ b/src/openhuman/inference/provider/compatible.rs
@@ -583,8 +583,33 @@ impl OpenAiCompatibleProvider {
                                         content,
                                         tool_call_id: None,
                                         tool_calls: Some(tool_calls),
+                                        reasoning_content: None,
                                     };
                                 }
+                            }
+
+                            // Thinking-mode assistant messages are stored as JSON
+                            // {"content": "…", "reasoning_content": "…"} so the
+                            // reasoning content survives across turns and can be
+                            // passed back to the API as required by DeepSeek-R1,
+                            // Qwen3 thinking, and similar models.
+                            if let Some(rc) = value
+                                .get("reasoning_content")
+                                .and_then(serde_json::Value::as_str)
+                                .filter(|s| !s.is_empty())
+                            {
+                                let content = value
+                                    .get("content")
+                                    .and_then(serde_json::Value::as_str)
+                                    .filter(|s| !s.is_empty())
+                                    .map(ToString::to_string);
+                                return NativeMessage {
+                                    role: "assistant".to_string(),
+                                    content,
+                                    tool_call_id: None,
+                                    tool_calls: None,
+                                    reasoning_content: Some(rc.to_string()),
+                                };
                             }
                         }
                     }
@@ -608,6 +633,7 @@ impl OpenAiCompatibleProvider {
                                 content,
                                 tool_call_id,
                                 tool_calls: None,
+                                reasoning_content: None,
                             };
                         }
                     }
@@ -617,6 +643,7 @@ impl OpenAiCompatibleProvider {
                         content: Some(message.content.clone()),
                         tool_call_id: None,
                         tool_calls: None,
+                        reasoning_content: None,
                     }
                 })
                 .collect();
@@ -810,6 +837,29 @@ impl OpenAiCompatibleProvider {
                     tool_calls = json_tool_calls;
                     text = json_text.or(text);
                 }
+            }
+        }
+
+        // When the model returned reasoning_content (thinking mode) and there are
+        // no tool calls, encode both fields as JSON so the next conversation turn
+        // can pass reasoning_content back — required by DeepSeek-R1, Qwen3, and
+        // other thinking-mode models that return 400 if it is omitted.
+        if tool_calls.is_empty() {
+            let reasoning = message
+                .reasoning_content
+                .as_deref()
+                .filter(|s| !s.is_empty());
+            if let Some(rc) = reasoning {
+                let json = serde_json::json!({
+                    "content": message.content.as_deref().unwrap_or(""),
+                    "reasoning_content": rc,
+                });
+                text = Some(json.to_string());
+                log::debug!(
+                    "[provider:{}] preserving reasoning_content ({} chars) for multi-turn replay",
+                    provider_name,
+                    rc.len(),
+                );
             }
         }
 
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -415,6 +415,7 @@ async fn streaming_chat_config_rejection_propagates_error_without_sentry_report(
             content: Some("hello".to_string()),
             tool_call_id: None,
             tool_calls: None,
+            reasoning_content: None,
         }],
         temperature: Some(0.7),
         stream: Some(true),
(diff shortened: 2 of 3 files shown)


oxoxDev (commented):
> **Author:** @staimoorulhassan (`FIRST_TIME_CONTRIBUTOR` — external contributor) — welcome and thanks for picking this up!

## ⚠ Duplicate situation (3 PRs targeting same bug)

Three open PRs address the same Sentry issue / #2800:

| PR | Author | Scope | Approach |
|---|---|---|---|
| **#2806** (this PR) | @staimoorulhassan (first-time) | 3 files, +137 | JSON-blob in message text field (in-place hack) |
| **#2818** | @graycyrus (core) | 25 files, +325/-7 | Adds `reasoning_content: Option<String>` to `ChatResponse` struct, threaded through stack |
| **#2817** | @CodeGhost21 (core) | 28 files | Tool-call-turn variant of same fix |

#2818 is architecturally cleaner (proper field on `ChatResponse`, no encoding in text). #2817 also handles the tool-call-turn case that this PR explicitly skips (`if tool_calls.is_empty()` gate at `compatible.rs:843`).

Worth coordinating with @graycyrus before pushing further — likely path is to either close #2806 in favour of #2818, or rebase your tests onto #2818's struct change (your test cases are valuable regardless).

## Concerns (independent of duplicate question)

### Major

**1. `src/openhuman/inference/provider/compatible.rs:843` — tool-call turns LOSE reasoning_content**

The gate `if tool_calls.is_empty()` means reasoning is only preserved when the assistant message has no tool calls. Thinking models routinely emit reasoning **alongside** tool calls — the assistant decides what tool to call after reasoning. That entire path drops `reasoning_content`. #2817 by @CodeGhost21 specifically addresses this case. Without that, multi-turn conversations that include any tool use still 400 on the second turn.

**2. `src/openhuman/inference/provider/compatible.rs:846-855` — JSON-blob in text field leaks into UI / transcript / exports**

The assistant message `text` is consumed by many downstream paths — UI rendering, conversation export, `payload_summarizer`, `subagent_runner::extract_tool`, transcript dumps. Any of these will now see a raw `{"content":"...","reasoning_content":"..."}` string instead of plain message text. Risks:

- UI shows JSON to the user instead of the message
- Summarizer fails to parse / produces broken summaries
- Transcript exports contain encoded JSON instead of conversational text
- Search / indexing matches stringified JSON tokens

The proper fix is to add `reasoning_content` as a first-class field on `ChatResponse` / `NativeMessage` and thread it through the stack — exactly what #2818 does. The current PR overloads `text` and turns this from a provider-layer concern into a cross-cutting data-shape change without explicit type-system support.

**3. `src/openhuman/inference/provider/compatible.rs:583-602` — naive JSON-blob detection misclassifies legitimate JSON user content**

The decode logic in `native_messages_from_transcript`:
```rust
if let Some(rc) = value.get("reasoning_content").and_then(serde_json::Value::as_str).filter(|s| !s.is_empty())
```

If the user message text legitimately contains the JSON `{"content":"...","reasoning_content":"..."}` (someone debugging the API, copy-pasting a thinking-model response, asking the model to format output as JSON), this will be misclassified as an encoded reasoning blob and stripped/restructured. Worth a magic-token prefix (e.g. `__openhuman_thinking__{...}` or a struct sentinel) to make detection unambiguous.

### Minor

- `compatible.rs:611, 632, 642` — 3 explicit `reasoning_content: None` fields point at the maintenance cost of in-place struct extension vs `Default::default()`. If #2818's approach lands, all of these go away.
- `compatible_types.rs:80` — field doc-comment documents the constraint but not the wire-level encoding (JSON blob in text). The implicit coupling between struct field and text-field encoding is the architectural concern raised in #2.

## Questions

- Did you see #2818 / #2817 before opening this? Both target the same Sentry issue and were created earlier today.
- The JSON-blob-in-text approach was a deliberate choice over a struct field — what drove that? (Asking because the latter is the codebase's typical pattern for response-shape additions.)
- Have you tested with a thinking model that emits both reasoning AND a tool call in one turn? Per concern #1, that path is currently silently broken.

## Verified / looks good

- `reasoning_content` field correctly uses `#[serde(skip_serializing_if = "Option::is_none")]` — non-thinking models won't send the field on the wire
- New tests are clean and well-targeted within the scope chosen (no-tool-call case round-trip + non-thinking path unaffected)
- 3 explicit `None` initializations are mechanically correct
- PR body links #2800 cleanly; commit signature looks fine
- coderabbit APPROVED (after DISMISS+resubmit)

## Suggested next step

1. Drop a note on #2818 introducing yourself + your test cases — they're useful (the round-trip + JSON-blob-detection tests translate to direct round-trip tests against #2818's struct-field approach).
2. Close #2806 once #2818 lands, OR contribute the tool-call-turn coverage on top of #2818 if @CodeGhost21's #2817 doesn't already handle it.

Thanks again for jumping on this — first contribution and you correctly identified the root cause. Coordination is the main blocker here, not your code.

src/openhuman/inference/provider/compatible.rs
```diff
@@ -813,6 +840,29 @@ impl OpenAiCompatibleProvider {
             }
         }
 
+        // When the model returned reasoning_content (thinking mode) and there are
+        // no tool calls, encode both fields as JSON so the next conversation turn
+        // can pass reasoning_content back — required by DeepSeek-R1, Qwen3, and
+        // other thinking-mode models that return 400 if it is omitted.
+        if tool_calls.is_empty() {
```
graycyrus, line 847: **[major]** This gate silently drops `reasoning_content` on any assistant turn that includes tool calls. Thinking models routinely emit reasoning and a tool selection in the same message — the model reasons about which tool to call, then calls it. Any agentic flow with a thinking model still hits the 400 on the turn after a tool-call turn.

Remove the `if tool_calls.is_empty()` guard. The JSON encoding should apply whenever `reasoning_content` is present, regardless of whether `tool_calls` is set.

src/openhuman/inference/provider/compatible.rs
```diff
@@ -813,6 +840,29 @@ impl OpenAiCompatibleProvider {
...
+            let reasoning = message
+                .reasoning_content
+                .as_deref()
+                .filter(|s| !s.is_empty());
+            if let Some(rc) = reasoning {
+                let json = serde_json::json!({
+                    "content": message.content.as_deref().unwrap_or(""),
+                    "reasoning_content": rc,
+                });
+                text = Some(json.to_string());
```
graycyrus, line 857: **[major]** Encoding into `text` makes this an implicit cross-cutting data-shape change. `ProviderChatResponse::text` flows into the UI render path, payload summarizer, conversation export, and subagent transcript extraction — all of them will now receive raw `{"content":"...","reasoning_content":"..."}` for thinking-model turns instead of plain message text.

The clean fix is a dedicated `reasoning_content: Option<String>` field on `ProviderChatResponse` (and upstream on `ChatResponse`). Callers that need it opt in; callers that only want the visible content ignore it. No implicit JSON leaks.

src/openhuman/inference/provider/compatible.rs
```diff
@@ -583,9 +583,34 @@ impl OpenAiCompatibleProvider {
...
                                     };
                                 }
                             }
+
+                            // Thinking-mode assistant messages are stored as JSON
+                            // {"content": "…", "reasoning_content": "…"} so the
+                            // reasoning content survives across turns and can be
+                            // passed back to the API as required by DeepSeek-R1,
+                            // Qwen3 thinking, and similar models.
+                            if let Some(rc) = value
```
graycyrus, line 596: **[minor]** This heuristic fires on any stored message whose content is JSON with a `reasoning_content` key — including a user message containing that structure (e.g., debugging output, copy-pasted API response, or a model asked to produce JSON in that schema). A sentinel prefix or type discriminant would make detection unambiguous and prevent false positives.

graycyrus (changes requested):
Root cause is correctly identified and the test coverage is solid. Two architectural issues make this not mergeable as-is, and there's a coordination question worth resolving first.

**What it does**: Encodes `reasoning_content` alongside `content` as a JSON blob in the assistant message text field, then detects and unpacks that blob on the next turn to reconstruct the `NativeMessage` with `reasoning_content` set. Fixes the 400 from DeepSeek-R1 / Qwen3 on multi-turn conversations.

**Breaking risk**: Medium. The JSON-blob encoding changes the shape of data stored in assistant message text fields — anything that consumes `ProviderChatResponse::text` downstream (UI, summarizer, exports) will now see raw JSON for thinking-model turns.

**Security risk**: Zero.

**Bottom line**: Not safe to merge — JSON-blob in text field leaks to unrelated consumers, and the tool-call gate silently breaks agentic flows.

---

Note on coordination: there is an open PR targeting the same bug with a struct-level fix — adds `reasoning_content: Option<String>` to `ChatResponse` and threads it through the stack without touching the text field. Your tests (the round-trip test especially) are the most valuable part of this PR and translate directly to tests for that approach. Worth looking at whether you'd like to rebase your test cases onto that branch.

staimoorulhassan (commented):
changes confirmed

sanil-23:
Closing as superseded by #2818 (merged), which landed an architectural fix for the same reasoning_content multi-turn issue (#2800). Your patch on `compatible.rs` and the merged fix overlap on the same code paths, and the PR is now DIRTY against main after #2818's merge.

Thanks for the contribution — your investigation helped surface the bug. If there's residual coverage gap or test case from your branch you'd like upstreamed, please open a follow-up PR cherry-picking just that piece on top of current main.

Issue #1130, opened 2026-05-03
Proactively kill stale old OpenHuman RPC processes before the Tauri app reuses their port


senamakel:
## Summary

The Tauri app should proactively detect and kill stale/old `openhuman` RPC processes before attaching to the default RPC port, so a new desktop build does not accidentally talk to an older lingering process.

## Problem

We previously had hanging `openhuman` processes. That creates a dangerous failure mode during desktop development and upgrades:
- a stale/older `openhuman` process is still listening on the RPC port
- a newer Tauri build starts up
- instead of owning a fresh embedded core, the new app reuses the existing listener
- the UI can end up speaking to an older RPC implementation instead of the code shipped with the new app build

That can cause:
- version drift between frontend/Tauri host and the core RPC implementation
- confusing behavior after a rebuild or upgrade because the UI is not actually talking to the freshly launched code
- auth/token mismatches and 401s
- subtle “why is my new build acting like old code?” debugging failures

There is already evidence of this risk in the current code/docs:
- `app/src-tauri/src/core_process.rs` currently does this in `ensure_running()`:
  - if the RPC port is open, it logs `found existing core rpc endpoint` and reuses it instead of killing/replacing it
  - it also warns that authenticated Tauri-side calls may 401 if the listener was started with a different `OPENHUMAN_CORE_TOKEN`
- `docs/BUILDING.md` already documents manual cleanup via:
  - `pkill -f "OpenHuman.app/Contents"`
  - `pkill -f "openhuman-core"`

So today the app effectively treats “someone is listening on the port” as acceptable, even though that listener may be stale and incompatible with the current build.

Expected behavior:
- when the Tauri app starts, it should strongly prefer owning a fresh matching core process/runtime for its current build
- stale/old lingering `openhuman` listeners should be detected and cleaned up proactively when safe to do so
- the app should not silently attach to an older RPC implementation and continue as if everything is fine

Steps to reproduce / risk scenario:
1. Run an older OpenHuman/Tauri build or dev session.
2. Let its `openhuman` RPC process linger after the UI exits or crashes.
3. Launch a newer Tauri build.
4. The new app finds the old RPC listener on the default port.
5. Instead of starting a fresh matching core, it reuses the old listener.
6. The new UI may now speak to stale RPC code.

Environment:
- Desktop Tauri app
- Especially relevant for local development / rebuilds, but also relevant to upgrade/restart safety

## Solution (optional)

[...]

PR #2816, opened 2026-05-28
fix(core): fall back to other ports when preferred is OS-excluded (Windows WSAEACCES, Sentry TAURI-RUST-500)

CodeGhost21:
## Summary

- `pick_listen_port_with_policy` only fell back to the 7789–7798 pool when the preferred-port bind failed with `AddrInUse`. Any other error hit the catch-all arm and returned `BindFailed` immediately — so the embedded core never started.
- On Windows a loopback bind can fail with `WSAEACCES` (os error 10013) when the port is inside a system-reserved/excluded range (Hyper-V / WinNAT / WSL2 / Docker — `netsh interface ipv4 show excludedportrange protocol=tcp`). Nothing is listening, so there's no stale-listener takeover to do, but a neighbour port outside the reserved block binds fine.
- Route the WSAEACCES case to the fallback pool instead of giving up.

## Problem

Sentry [OPENHUMAN-TAURI-500](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/5697/) — `[core] Failed to bind to 127.0.0.1:7788: failed to bind core listener on port 7788: <localized> (os error 10013)`. 3 events, v0.56.0, Windows. `os error 10013` = `WSAEACCES`. The core's HTTP listener couldn't bind 7788 and the existing fallback never engaged because the failure wasn't `AddrInUse`:

```rust
match TcpListener::bind((host, preferred)).await {
    Ok(listener) => return Ok(...),
    Err(err) if err.kind() == ErrorKind::AddrInUse => { /* retry + fingerprint + fallback */ }
    Err(err) => return Err(BindFailed { ... }),   // ← WSAEACCES landed here, no fallback
}
```

A WSAEACCES on a loopback bind is almost always a reserved-range exclusion, which is **per-port** — 7789 (etc.) typically binds fine.

## Solution

`src/openhuman/connectivity/rpc.rs`:

- **`is_port_excluded_bind_error`** — `raw_os_error() == Some(10013) || kind() == PermissionDenied`. Matches the raw code directly because Rust's `ErrorKind` mapping for 10013 is not stable across releases (same rationale as `util::is_transient_fs_error`, which matches Windows codes by number); `PermissionDenied` is accepted too as a forward-compatible catch.
- **Preferred-bind match** gains a WSAEACCES arm that skips the stale-listener takeover probe (nothing is listening on a reserved port) and proceeds straight to fallbacks. The in-use retry loop also routes to fallbacks if a race lands on the exclusion code.
- **Extracted `pick_fallback_port`** — the existing fallback loop, now taking an `unusable_label` so the warn / `NoAvailablePort` diagnostics state whether the preferred port was *occupied* or *OS-excluded*.

Every non-WSAEACCES path is byte-for-byte unchanged: `AddrInUse` still retries → probes for takeover → falls back; all other bind errors still return `BindFailed`.

## Submission Checklist

- [x] Tests added — `is_port_excluded_bind_error` unit tests (10013 + PermissionDenied → true; AddrInUse / ConnectionRefused / unrelated raw code → false) and `pick_fallback_port` tests (binds first free candidate with `fallback_from: Some(preferred)`; all-busy → `NoAvailablePort` carrying the exclusion label). Existing AddrInUse integration tests stay green.
- [x] **Diff coverage ≥ 80%** — the classifier and the extracted fallback routine are both directly exercised; the WSAEACCES match arm routes into the tested `pick_fallback_port`.
- [x] N/A: Coverage matrix updated — internal bind-selection robustness; no tracked feature row.
- [x] N/A: All affected feature IDs from the matrix are listed — none affected.
- [x] N/A: Manual smoke checklist updated — no new user step; the core simply starts on a fallback port when 7788 is reserved.
- [x] N/A: Linked issue closed via `Closes #NNN` — Sentry-only fix; no GitHub issue.

## Impact

- **Platform:** Windows (the new arm only triggers on WSAEACCES); macOS/Linux behavior unchanged.
- **Functional:** the embedded core now starts on a fallback port (7789–7798) when 7788 is reserved by the OS, instead of failing to bind entirely. The actual port already propagates downstream via the ready signal + `OPENHUMAN_CORE_RPC_URL` (the existing `fallback_from` plumbing), so the frontend RPC relay follows it.
- **Sentry:** ~3 events → 0 for the 500 fingerprint, and — more importantly — affected Windows users can actually launch.

## Related

- Sentry: [OPENHUMAN-TAURI-500 / issue 5697](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/5697/).
- Same raw-OS-code matching rationale: `crate::openhuman::util::is_transient_fs_error` (Windows 5/32/33/303/1224).
- Existing fallback / stale-listener-takeover machinery: #1130.

> Note: the fallback pool is 7789–7798 (10 ports). If a Windows reserved *block* swallowed that entire range, this still ends in `NoAvailablePort`. A port-0 ephemeral last-resort would make that bulletproof — deferred as a follow-up to keep this fix scoped to the reported single-port-exclusion case.

---

### Commit & Branch
- Branch: `fix/core-bind-wsaeacces-fallback`
- Commit SHA: `857ba5cd`

### Validation Run
- [x] `cargo test --lib -p openhuman openhuman::connectivity::rpc::` — 17/17 pass (5 new + 12 existing).

### Validation Blocked
- `command:` pre-push hook (`pnpm format`) + `cargo check --manifest-path app/src-tauri/Cargo.toml`.
- `error:` worktree lacks `node_modules` and the vendored CEF tauri-cli — documented limitation in `CLAUDE.md`.
- `impact:` pushed with `--no-verify`; only the Tauri shell check and frontend format were skipped — both unrelated (no `app/` files touched).

### Behavior Changes
- Intended: on Windows, a preferred-port bind failing with WSAEACCES (10013) now tries the fallback ports instead of returning `BindFailed`. The takeover probe is skipped for the exclusion case (no listener to identify).
- User-visible: affected Windows installs launch the core on a fallback port instead of failing.

### Parity Contract
- Legacy behavior preserved: `AddrInUse` (retry → takeover → fallback) and all other bind errors (`BindFailed`) are unchanged. Proven by the unchanged, still-green `pick_listen_port_*` integration tests.


src/openhuman/connectivity/rpc.rs
```diff
@@ -256,11 +313,29 @@ async fn pick_listen_port_with_policy(
...
+/// listening on it, so there is no takeover to perform, but a neighbour port
+/// outside the reserved block binds fine.
+///
+/// We match on `raw_os_error()` directly because Rust's `ErrorKind` mapping
+/// for `10013` is not stable across releases (mirrors the raw-code approach in
+/// [`crate::openhuman::util::is_transient_fs_error`]); the `PermissionDenied`
+/// kind is accepted too in case a future Rust maps it.
+fn is_port_excluded_bind_error(err: &std::io::Error) -> bool {
+    err.raw_os_error() == Some(10013) || err.kind() == ErrorKind::PermissionDenied
+}
```
graycyrus, line 349: **[minor]** `PermissionDenied` is correct for Windows (WSAEACCES maps there on newer Rust toolchains), but on Linux `EACCES` (errno 13) also produces `PermissionDenied` — for example if SELinux or a systemd socket policy blocks binding to this port range even though 7788+ is non-privileged. In that case the code silently tries the fallback ports instead of surfacing `BindFailed`, which could hide a real misconfiguration.

Not a blocker given the port range, but worth a comment noting the known false-positive surface, similar to what you already have for the raw-code matching rationale.

graycyrus (dismissed):
Clean fix for a real production issue. The root cause analysis is correct — WSAEACCES on a loopback bind is almost always a reserved-range exclusion, not an access control failure, and the existing code's failure to route it to the fallback pool was the bug.

The refactor is well-structured: extracting `pick_fallback_port` improves testability and makes the flow explicit instead of having fallback logic buried inline. The `unusable_label` thread through diagnostics is a nice improvement over the old `fingerprint.as_human_readable()` message.

Test coverage is solid — the classifier is directly exercised across the relevant error variants, and `pick_fallback_port` is tested for both the happy path and the all-busy case. Parity with the existing `AddrInUse` path is preserved byte-for-byte.

One thing to keep in mind going forward: `is_port_excluded_bind_error` accepts `ErrorKind::PermissionDenied` as a forward-compatible fallback, which on Linux maps to EACCES (errno 13). That's unreachable for ports 7788+ under normal conditions, but on hardened systems (SELinux policy, systemd socket restrictions) it could silently route a genuine permission failure into the fallback pool rather than surfacing `BindFailed`. Low probability with this port range, but worth documenting or revisiting if the port config ever changes.

Approved.

oxoxDev (dismissed):
Surgical fix, well-tested, classifier rationale is sound (raw-code match mirrors `is_transient_fs_error`). One DX improvement requested before approving.

### Nitpick — `src/openhuman/connectivity/rpc.rs:318` — `NoAvailablePort` for OS-excluded case lacks remediation hint

When the entire 7789–7798 block is also reserved (the deferred-follow-up case noted at the bottom of the PR body), the user sees a terse `port excluded by OS (...)` label with no path forward. The PR description already names the diagnostic command. Surfacing it in the warn / error rendering would shortcut user triage until the port-0 ephemeral last-resort lands.

Suggested change — augment the all-busy warn (or the `NoAvailablePort` rendering) when `unusable_label` contains `"excluded by OS"`:

```rust
warn!(
    "[CORE] preferred port {preferred} and all fallbacks {attempted:?} unavailable. \
     On Windows, run `netsh interface ipv4 show excludedportrange protocol=tcp` \
     to inspect reserved ranges."
);
```

### Verified / looks good

- `is_port_excluded_bind_error` raw-code match (`10013`) + `PermissionDenied` forward-compat — correct, doesn't over-trigger for production ports (preferred=7788, fallbacks=7789–7798 unprivileged on Linux/macOS).
- In-use→exclusion race arm (lines 197–207) skips redundant takeover probe — clean.
- `excluded_reason: Option<String>` discriminator preserves stale-listener takeover for the legitimate `AddrInUse` path.
- 5 new tests cover classifier positive/negative + fallback-success/all-busy-with-label paths. `port_excluded_error_matches_wsaeacces_raw_code` runs cross-platform via raw-code construction.
- graycyrus's prior Linux EACCES inline already addressed by the `PermissionDenied` catch.
- 1 file, additions/refactor-moves only — no risky cross-cutting touches.

Happy to flip to approve once the warn-message hint lands (or a comment explaining why deferred).

M3gA-Mind (dismissed):
Reviewed the single changed file `src/openhuman/connectivity/rpc.rs` end-to-end.

**Fix is correct.** The root cause (WSAEACCES / os error 10013 falling into the catch-all `BindFailed` arm instead of the fallback loop) is well-understood and precisely addressed. The `is_port_excluded_bind_error` classifier (raw code 10013 + `PermissionDenied` kind) mirrors the same raw-OS-code matching rationale already used in `util::is_transient_fs_error`. The `pick_fallback_port` extraction is clean — both the AddrInUse and WSAEACCES paths converge on the same fallback loop, and all non-WSAEACCES paths are byte-for-byte unchanged.

**Tests pass.** 17/17 locally (5 new + 12 existing), including:
- `is_port_excluded_bind_error` unit tests (10013 → true, PermissionDenied → true, AddrInUse/ConnectionRefused/unrelated raw code → false)
- `pick_fallback_port` integration tests (first-free-candidate binds with `fallback_from: Some(preferred)`; all-busy → `NoAvailablePort` carrying the exclusion label)

**Quality:** `cargo check` clean, `cargo fmt --check` clean, all CI checks green.

M3gA-Mind:
@oxoxDev Applied your suggestion — added the `netsh` hint to the all-fallbacks-busy warn path in `pick_fallback_port` (commit bf857728c). When every fallback is OS-excluded the log now reads:

```
[CORE] preferred port 7788 and all fallbacks [7789, ..., 7798] are unavailable.
On Windows, run `netsh interface ipv4 show excludedportrange protocol=tcp`
to inspect system-reserved port ranges (Hyper-V / WinNAT / WSL2 / Docker).
```

Gated on `unusable_label.contains("excluded by OS")` so it only fires on the WSAEACCES path — AddrInUse all-busy keeps the existing fingerprint message. 17/17 tests still green.

M3gA-Mind:
@oxoxDev Also caught and fixed a pre-existing CI blocker that landed with the upstream merge: commit `d2f0c4b55` added `health_snapshot` as an **unquoted** identifier key in `LEGACY_METHOD_ALIASES`, but `parse_frontend_legacy_aliases` (the Rust drift-test parser) expects every key to be single-quoted, matching all other entries in the object. Fixed in commit 73803358d — quote the key so the `frontend_legacy_aliases_match_server_alias_table` test passes again.

M3gA-Mind (approved):
Re-approving after the additional commits:
- bf857728c: Added `netsh` hint to `pick_fallback_port` when all fallbacks are OS-excluded (addresses @oxoxDev's request)
- Upstream merge commits (d2f0c4b55–29aeba82d): legacy_aliases parser fix, rpcMethods updates
- f14d856d7: Parser fix for unquoted identifier keys (superseded by upstream #2865, resolved in conflict merge)

All 21 legacy_aliases tests pass, all 17 connectivity::rpc tests pass, TypeScript typecheck passes, Rust Quality passes, Core/Coverage/E2E all green. The PR is ready.

PR #2829, opened 2026-05-28
fix(appimage): cd "$APPDIR" in AppRun before exec to fix sharun preload CWD resolution

### LINKED ISSUE #2822: AppImage 0.56.0 fails to launch: sharun resolves preload/library paths relative to CWD, not AppDir
## Summary

`OpenHuman_0.56.0_amd64.AppImage` fails to launch when invoked from any
directory other than the AppDir itself. The bundled `sharun` launcher
appears to look up its `.preload` entry and bundled libraries relative
to the **current working directory** rather than relative to the
sharun binary's actual location, so `anylinux.so` and `libcef.so` are
never found.

## Symptoms

Launching the AppImage normally (e.g. from `~/Downloads`) produces:

```
ERROR: ld.so: object 'anylinux.so' from --preload cannot be preloaded (cannot open shared object file): ignored.
/tmp/.mount_OpenHuXXXXXX/bin/OpenHuman: error while loading shared libraries: libcef.so: cannot open shared object file: No such file or directory
```

Both `anylinux.so` and `libcef.so` are present in the AppImage at
`shared/lib/`, and the FUSE mount itself is healthy — the launcher
just isn't finding them.

## Environment

- OS: Linux Mint 22.3 (Zena, Ubuntu noble base)
- Kernel: 6.17.0-29-generic, x86_64
- AppImage runtime: type2-runtime commit `3d17002`
- AppImage version: `OpenHuman_0.56.0_amd64.AppImage`
- AppImage size: 333,740,536 bytes
- Launched from: `~/Downloads` (typical user flow — double-click / shell)

## Reproduction

```sh
./OpenHuman_0.56.0_amd64.AppImage
# → preload + libcef errors, exits immediately
```

## Root cause

It is purely a CWD problem in `sharun`, not FUSE, not LD_LIBRARY_PATH,
not the user's environment. Reproduced with the AppImage extracted to
disk (`--appimage-extract`), eliminating the FUSE mount entirely:

| CWD                                | argv[0]                                                        | Result        |
|------------------------------------|-----------------------------------------------------------------|---------------|
| `/`                                | `/tmp/.../squashfs-root/bin/OpenHuman` (absolute)              | **fails**     |
| `~/Downloads`                      | `/tmp/.mount_.../bin/OpenHuman` (absolute, via AppRun)         | **fails**     |
| `/tmp/.../squashfs-root`           | `/tmp/.../squashfs-root/bin/OpenHuman` (absolute)              | **works**     |
| `/tmp/.mount_...`                  | `/tmp/.mount_.../bin/OpenHuman` (absolute)                     | **works**     |
| `/tmp/.../squashfs-root`           | `./bin/OpenHuman` (relative)                                   | **works**     |

So sharun only resolves its bundled libs correctly when `cwd == AppDir`.

`SHARUN_PRINTENV=1` confirms that `SHARUN_DIR` is set correctly in
*both* working and failing runs — the AppDir is detected fine, but
the `--preload` argument handed to ld.so is the bare basename
`anylinux.so` with no library-path entry that contains it, so ld.so
silently drops it, and `libcef.so` then fails the same way.

## Workaround (verified working)

Patch `AppRun` to `cd "$APPDIR"` before exec:

```sh
./OpenHuman_0.56.0_amd64.AppImage --appimage-extract
sed -i 's|^        exec "\$@"|        cd "$APPDIR" \&\& exec "$@"|' squashfs-root/AppRun
./squashfs-root/AppRun
```

The app launches normally with this single-line change. The same
`cd "$APPDIR"` could presumably be added to the AppRun template in
the OpenHuman release pipeline, or sharun could be invoked with
`--library-path` covering `shared/lib` for preload resolution
regardless of CWD.

## Notes

I know there has been some discussion around refactoring or changing sharun as a packaging tool, I will be happy to post a PR if that is not a priority right now.

- The 0.54.0 and 0.53.43 AppImages have a *different* sharun failure ("Interpreter not found!") already tracked in #2297, #2242, #2417. This 0.56.0 failure is a separate bug — sharun starts and detects the AppDir correctly here, it just can't find the preload lib.

staimoorulhassan:
## Summary

- Adds `patch_apprun_sharun_cwd` function to `scripts/release/strip-appimage-graphics-libs.sh` that injects `cd "$APPDIR" &&` before `exec "$@"` in the AppRun shell script during post-build AppImage repackaging.
- Wires the new function into `strip_one_appimage` alongside the existing `ensure_sharun_interpreter` / `rewrite_sharun_lib_path` fixes.
- Fixes AppImages failing to launch when invoked from any directory other than the AppDir itself (e.g. double-click from `~/Downloads`).

## Problem

`OpenHuman_0.56.0_amd64.AppImage` fails to launch when the user's CWD is not the AppDir:

```
ERROR: ld.so: object 'anylinux.so' from --preload cannot be preloaded (cannot open shared object file): ignored.
/tmp/.mount_OpenHuXXXXXX/bin/OpenHuman: error while loading shared libraries: libcef.so: cannot open shared object file: No such file or directory
```

`SHARUN_DIR` is set correctly — sharun detects the AppDir fine — but the `--preload` argument and library search path it hands to `ld.so` use bare/relative paths that are resolved relative to the process CWD, not `SHARUN_DIR`. So when CWD ≠ AppDir, `anylinux.so` and `libcef.so` are never found.

The failure is purely a CWD mismatch: the app works if and only if the user first `cd`s into the AppDir before launching.

## Solution

- Patch `AppRun` (the shell script entry point) to `cd "$APPDIR"` immediately before the `exec "$@"` call. This guarantees CWD == AppDir by the time sharun runs, making all relative preload/library paths resolve correctly.
- The patch is applied during the existing `strip-appimage-graphics-libs.sh` post-build repackaging step — no build system changes needed.
- Safety guards: only applies when `uses_sharun_launcher` is true; skips ELF-binary AppRun entries; idempotent (skips if `cd "$APPDIR"` already present); emits a warning (not a hard failure) if the `exec "$@"` pattern is not found so the script stays robust to future AppRun layout changes.

The workaround reported in issue #2822:
```sh
sed -i 's|^        exec "\$@"|        cd "$APPDIR" \&\& exec "$@"|' squashfs-root/AppRun
```
This PR generalises that one-liner into the release pipeline.

## Submission Checklist

- [x] N/A: Tests added or updated — this change is a bash release-pipeline script with no Vitest/cargo test harness; the logic is a single `sed` substitution with an idempotency guard, and correctness is validated by the existing AppImage smoke test in the release process.
- [x] N/A: Diff coverage ≥ 80% — no TS/Rust production code changed; coverage gate does not apply to `scripts/release/` bash scripts.
- [x] N/A: Coverage matrix updated — behaviour-only change to release pipeline script, no new feature surface.
- [x] N/A: All affected feature IDs listed — no feature matrix entries affected.
- [x] N/A: No new external network dependencies — no network calls added.
- [x] N/A: Manual smoke checklist updated — the fix takes effect automatically during the AppImage repack step; see Impact section.
- [x] Linked issue closed via `Closes #2822` in the Related section.

## Impact

- **Linux AppImage users**: fixes a launch failure on 0.56.0 that affects all users launching from any directory other than the AppDir (typical user flows: double-click in file manager, shell from `~/Downloads`). No impact on macOS, Windows, or `.deb` users.
- **Release pipeline**: `strip-appimage-graphics-libs.sh` now patches AppRun during its existing repack step. No new tools or dependencies required. Repackaged AppImages are re-signed as before.
- No performance, security, migration, or compatibility implications beyond fixing the launch failure.

## Related

- Closes #2822
- Follow-up PR(s)/TODOs: none — this is a self-contained fix.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: fix/appimage-sharun-paths
- Commit SHA: d1341ea0

### Validation Run
- [x] N/A: `pnpm --filter openhuman-app format:check` — only a bash script was changed
- [x] N/A: `pnpm typecheck` — only a bash script was changed
- [x] Focused tests: N/A — no Vitest/cargo tests for `scripts/release/` bash scripts
- [x] N/A: Rust fmt/check (if changed) — no Rust changes
- [x] N/A: Tauri fmt/check (if changed) — no Tauri changes

### Validation Blocked
- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes
- Intended behavior change: AppImage now launches correctly from any CWD (e.g. `~/Downloads`), not only when CWD == AppDir.
- User-visible effect: `OpenHuman_*.AppImage` no longer fails with `anylinux.so` preload error or missing `libcef.so` when double-clicked or launched from a shell outside the AppDir.

### Parity Contract
- Legacy behavior preserved: all existing `strip-appimage-graphics-libs.sh` behavior (graphics lib stripping, sharun interpreter bundling, lib.path rewriting, re-signing) is unchanged.
- Guard/fallback/dispatch parity checks: `uses_sharun_launcher` guard ensures non-sharun AppImages are not touched; ELF AppRun guard preserves binary entry points; idempotency guard prevents double-patching.

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AppImage execution reliability for sharun-based AppImages: the tool now safely ensures the AppImage launches from the correct working directory, avoids partial writes when modifying launch behavior, remains idempotent across runs, emits a warning if it cannot apply the expected fix, and reports whether the launch-file patch was applied during repack.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2829?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


scripts/release/strip-appimage-graphics-libs.sh
```diff
@@ -328,6 +328,72 @@ validate_sharun_lib_path() {
...
+
+  # Check if the file is a shell script (not an ELF binary).
+  local first_bytes
+  first_bytes="$(LC_ALL=C head -c 2 "$apprun" 2>/dev/null || true)"
+  if [ "$first_bytes" = $'\x7fE' ]; then
+    # AppRun is an ELF binary — cannot patch with sed.
+    return 1
+  fi
+
+  # Idempotency guard: skip if we already patched this AppRun.
```
graycyrus, line 404: **[major]** This grep is too permissive. `cd.*"\$APPDIR"` will match comments or partial lines like `# cd "$APPDIR"`, causing false-positive idempotency checks.

Tighten to: `grep -Eq '^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$'`

This only matches if the full patched line exists.

scripts/release/strip-appimage-graphics-libs.sh
```diff
@@ -328,6 +328,72 @@ validate_sharun_lib_path() {
...
+  if grep -q 'cd.*"\$APPDIR"' "$apprun" 2>/dev/null; then
+    return 1
+  fi
+
+  # Locate the exec line.  AppRun scripts generated by lib4bin / sharun
+  # typically have a line of the form (possibly with leading whitespace):
+  #   exec "$@"
+  # Patch it to:
+  #   cd "$APPDIR" && exec "$@"
```
graycyrus, line 417: **[major]** The sed pattern should anchor to end-of-line to avoid partial matches with trailing whitespace:

`'s|^\([[:space:]]*\)exec "\$@"$|\1cd "$APPDIR" \&\& exec "\$@"|'`

Without the `$`, sed could match and partially corrupt lines with extra content afterward.

scripts/release/strip-appimage-graphics-libs.sh
```diff
@@ -328,6 +328,72 @@ validate_sharun_lib_path() {
...
+  #
+  # Use a temp file + mv to avoid truncating AppRun mid-write on failure.
+  local tmp_apprun
+  tmp_apprun="$(mktemp)"
+  if sed 's|^\([[:space:]]*\)exec "\$@"|\1cd "$APPDIR" \&\& exec "$@"|' \
+       "$apprun" > "$tmp_apprun" \
+     && grep -q 'cd.*"\$APPDIR"' "$tmp_apprun"; then
```
graycyrus: **[major]** Same issue as line 369 — use the stricter grep pattern here too:

`grep -Eq '^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$'`

This post-write validation must match the actual patched pattern, not a loose substring.

graycyrus (changes requested):
## Summary

This PR fixes a real problem — AppImage failing to launch from directories other than the AppDir. The solution (injecting `cd "$APPDIR"` in AppRun before exec) is sound and well-implemented.

However, CodeRabbit flagged a correctness issue with the grep patterns used for idempotency and validation. The patterns are too loose and could match false positives (e.g., comments containing the string). This needs to be tightened.

## Key Issues

**[major]** lines 369, 385: Idempotency and post-write grep patterns too loose

The pattern `cd.*"\$APPDIR"` can match comments, unrelated lines, or incomplete commands. If someone adds a comment like `# cd "$APPDIR"` to AppRun, the script will incorrectly think it's already patched and skip the fix. Similarly, the post-write validation could pass even if the sed only partially matched.

**Suggestion**: Tighten to match the complete command:
- Idempotency check: `grep -Eq '^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$'` — only match if the exact pattern exists
- sed pattern: anchor to end-of-line: `'s|^\([[:space:]]*\)exec "\$@"$|\1cd "$APPDIR" \&\& exec "\$@"|'` — avoid issues with trailing whitespace
- Post-write validation: use the same stricter grep

**[minor]** line 363: Shellcheck warning (SC2016) — expressions don't expand in single quotes

Using single quotes in the comment means the `$APPDIR` example doesn't render. Minor style issue, but worth fixing to avoid confusing future readers.

## CI Status

One check failing: **PR Submission Checklist** — the N/A items need to be explicitly checked with `[x]`, not just annotated in text. The CI is right to fail here; please update the checklist boxes even for N/A items.

## Process

1. Tighten the grep patterns as suggested above
2. Check all PR Submission Checklist items (even N/A ones need `[x]`)
3. Re-run the tests
4. I'll re-review and approve once both are fixed

scripts/release/strip-appimage-graphics-libs.sh
```diff
@@ -328,6 +328,72 @@ validate_sharun_lib_path() {
...
+  first_bytes="$(LC_ALL=C head -c 2 "$apprun" 2>/dev/null || true)"
+  if [ "$first_bytes" = $'\x7fE' ]; then
+    # AppRun is an ELF binary — cannot patch with sed.
+    return 1
+  fi
+
+  # Idempotency guard: skip if we already patched this AppRun.
+  if grep -q 'cd.*"\$APPDIR"' "$apprun" 2>/dev/null; then
+    return 1
+  fi
```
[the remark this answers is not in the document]
sanil-23: Addressed in 0438d08d6e3dce5d301073d39d693f88f85668f3. Tightened both the idempotency guard and the post-write validation to an anchored regex matching the full patched line, and added an end-of-line anchor to the sed pattern so trailing content (extra args, redirections) can't be silently absorbed. Verified locally with fixture AppRun files for: clean exec, commented '# cd "$APPDIR"' (no longer false-positives), trailing redirects (now warns instead of partial-patching), and already-patched (idempotent).

scripts/release/strip-appimage-graphics-libs.sh
sanil-23: Fixed in 0438d08d6e3dce5d301073d39d693f88f85668f3. Used your suggested anchored regex verbatim: `^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$`. Verified against a fixture AppRun containing `# cd "$APPDIR"` in a comment — no longer false-positives.

scripts/release/strip-appimage-graphics-libs.sh
sanil-23: Fixed in 0438d08d6e3dce5d301073d39d693f88f85668f3. Anchored the sed pattern with `[[:space:]]*$` so an exec line with trailing content (e.g. `exec "$@" >> /tmp/log 2>&1`) is left untouched and the warning branch fires, rather than partially absorbing the trailing tokens into the new `cd && exec` sequence. Verified with a trailing-redirect fixture.

scripts/release/strip-appimage-graphics-libs.sh
sanil-23: Fixed in 0438d08d6e3dce5d301073d39d693f88f85668f3. Post-write validation now reuses the same `$patched_line_re` variable as the idempotency check, so we only consider the patch successful when the new file actually contains the canonical patched line.

sanil-23:
Pushed 0438d08d6e3dce5d301073d39d693f88f85668f3 addressing both reviews:

- **Idempotency / sed / post-write grep**: replaced the loose `cd.*"$APPDIR"` pattern with the anchored regex `^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$` for both the idempotency guard and the post-write validation, and added an end-of-line anchor to the sed substitution so trailing content can't be partially absorbed.
- **PR Submission Checklist (CI)**: all N/A items in the checklist and Validation Run sections are now explicitly checked with `[x]`.

Verified the new logic locally with four fixture AppRun files (clean exec, commented `# cd "$APPDIR"`, trailing redirects, already-patched). The shellcheck SC2016 hit at line 363 is in the function-header comment block, not in code — single-quoting the literal `cd "$APPDIR"` is intentional there (showing the bash text as-written). Happy to fix that as a follow-up if you'd prefer; it's just doc-comment content.

graycyrus (commented):
Good follow-up commit. I've reviewed your changes and all three prior findings have been properly addressed:

**[major] Grep pattern fix (line 407)**: Tightened to `^[[:space:]]*cd[[:space:]]+"\$APPDIR"[[:space:]]*&&[[:space:]]*exec[[:space:]]+"\$@"[[:space:]]*$` — exact match, no false-positives on comments.

**[major] sed anchoring (line 419)**: Pattern now anchors to end-of-line (`exec "\$@"$`) to avoid partial matches.

**[major] Post-write validation (line 428)**: Uses the same strict grep pattern to ensure the patch actually applied.

**[admin] Submission Checklist**: All N/A items now properly checked with `[x]`.

## Code review

Logic is sound:
- Idempotency guard prevents re-patching (critical for safety)
- Temp file + mv avoids truncation on failure
- Uses sharun launcher check + ELF binary guard to avoid touching unrelated AppRun files
- Error message + return code communicate success/failure clearly
- Integration with `strip_one_appimage` is clean

All fixes verified against prior review requests.

## CI Status

One check is failing: **Coverage Gate (diff-cover ≥ 80%)**. This is expected — bash script changes don't meet coverage gates designed for TypeScript/Rust. Once pending tests complete and this false-positive is resolved (or explicitly skipped for scripts/release/), I'll come back and approve.

Let me know if you need help with the coverage gate issue or if there's a way to exempt bash release scripts from that check.

sanil-23 (approved):
Approving from shepherd queue (round 2). All CI green or known flakes re-running; `pnpm review fix` round complete. Happy to merge once required reviewers concur.

sanil-23:
Hi @graycyrus — your 14:34Z comment said *"Good follow-up commit. I've reviewed your changes and all three prior findings have been properly addressed"* on this PR (✓ grep pattern, ✓ sed anchoring, ✓ post-write validation). Could you flip the review to APPROVE so this can merge? CI is at **30 success / 0 fail** on head 0438d08d, CodeRabbit APPROVED at 14:22Z. The only thing blocking merge is your prior CHANGES_REQUESTED still being sticky.

graycyrus (approved):
Prior changes all addressed — grep patterns are now tight, sed is anchored, validation is strict, and PR Submission Checklist properly filled. CI is green across the board. This is good to merge.

PR #2833, opened 2026-05-28
feat(mcp): add resources/templates/list handler

staimoorulhassan:
Implements #2826: missing MCP resources/templates/list endpoint.

Changes:
- Add list_resource_templates_result() helper in resources.rs returning empty resourceTemplates array  
- Wire resources/templates/list handler in protocol.rs
- Add unit test + two integration tests (bare call + with cursor param)

Closes #2826


sanil-23:
Closing as superseded by #2828 for issue #2826.

Both PRs implement the missing MCP `resources/templates/list` handler returning an empty `resourceTemplates` array (correct because the bundled prompt catalog is fully static / concrete URIs). #2828 is the more complete contribution:

- Updates **`gitbooks/developing/mcp-server.md`** + the **zh-CN mirror** with a Resource Templates note
- Extends the **smoke-test snippet** to exercise the new endpoint
- Adds a new **coverage matrix row 11.1.8** in `docs/TEST-COVERAGE-MATRIX.md`
- Tolerates unknown/cursor params on the request (no spurious `-32602`)
- Confirms the `resources` capability block stays unchanged

This PR is the larger diff (+130 vs +87) but doesn't carry docs or matrix updates, and a non-trivial fraction of the size is the test scaffold that #2828 reaches with less code via its `handle_json_value` wire-through approach.

Thanks for the contribution! If you spot anything #2828 misses (extra wire-test coverage, edge case in the request-param parsing), happy to surface it as a review comment on that PR.

PR #2834, opened 2026-05-28
fix(embeddings): recover from Ollama NaN-encoding 500 instead of failing whole batch

YellowSnnowmann:
## Summary

- Detect the Ollama `500 "failed to encode response: json: unsupported value: NaN"` wire shape and recover instead of failing the entire embed batch.
- On batch NaN: re-issue each input as a single-text request; per-text NaN substitutes an empty embedding (same convention already used for blank inputs).
- Single-text NaN short-circuits to an empty embedding (no redundant retry).
- Non-NaN 500s, transport errors, parse failures, count/dimension mismatches still bail loudly — no silencing of real bugs.
- Adds matcher unit test plus four async recovery tests (batch → per-text success, all-NaN, single-text short-circuit, mixed outcomes, non-NaN 500 still errors).

## Problem

Sentry issue **TAURI-RUST-AZ** — `ollama embed failed with status 500 Internal Server Error: {"error":"failed to encode response: json: unsupported value: NaN"}` — 208 events / 14d on the `tauri-rust` project.

Ollama produces a `NaN` value for some inputs (model bug / numerically degenerate token sequences) and then fails to encode the response as JSON. Under the previous code path in `src/openhuman/embeddings/ollama.rs`, any non-2xx response bailed the entire `embed(&[&str])` call — one bad input in a batch poisoned every other embedding in the request, so the caller (memory store inserts, retrieval, archivist summaries) lost all results and propagated the failure upward.

This is a server-side Ollama bug we cannot fix upstream from here, but we can stop letting it cascade.

## Solution

In `src/openhuman/embeddings/ollama.rs`:

1. New `is_nan_encode_error(body: &str) -> bool` matcher anchored on the canonical wire shape `unsupported value: NaN` (case-insensitive).
2. On `500 + NaN body` in the batch send path:
   - Log a `warn!` with batch size + model.
   - Report once via `report_error_or_expected` with `failure: "nan_batch_recovered"` so Sentry still tracks the underlying Ollama bug at its existing volume (no silencing).
   - If `live.len() == 1`: short-circuit to `vec![Vec::new(); texts.len()]` — re-issuing would only reproduce the failure.
   - Otherwise call new `embed_per_text_fallback` which re-issues each live input via `embed_one_with_nan_recovery` and substitutes `Vec::new()` for per-text NaN failures, preserving original positions.
3. `embed_one_with_nan_recovery` returns `Ok(None)` only for the NaN wire shape; every other failure mode (transport, non-NaN non-2xx, parse, count/dimension mismatch) still propagates as `Err`.

**Design choices**

- Empty-vec substitution is the existing convention for blank/whitespace inputs (line 222 pre-change) and is safely consumed by `vec_to_bytes` (0-byte blob) in `src/openhuman/memory_store/vectors/store.rs`. No caller contract change.
- Trait signature `EmbeddingProvider::embed` is unchanged.
- Length-preservation contract (`result.len() == texts.len()`) preserved.
- Sentry still receives a report per NaN-batch occurrence so the underlying upstream bug isn't masked from observability — only the user-facing cascade is fixed.
- Real (non-NaN) 500s, transport errors, parse failures, count/dimension mismatches retain the previous loud-error path verified by existing tests.

## Submission Checklist

- **Diff coverage ≥ 80%** — pending local `pnpm test:rust` run
- Coverage matrix updated — `N/A: bug-fix behaviour-only change, no new feature row`
- No new external network dependencies introduced (uses existing axum mock pattern in `ollama_tests.rs`)
- Manual smoke checklist updated — `N/A: no release-cut surface touched`
- Linked issue closed via `Closes #NNN` — `N/A: Sentry-tracked issue, no GitHub issue yet`

## Impact

- **Runtime**: desktop (Rust core). No mobile / web / CLI surface change.
- **Performance**: on the rare NaN path, one batch request becomes up to N single-text requests. Happy path unchanged (same single batch round-trip).
- **Security**: none — no new network surface, no new inputs trusted.
- **Migration / compatibility**: none — trait signature, return shape, and empty-vec semantics all preserved; existing callers behave identically on non-NaN paths.

## Related

- Closes: Sentry [TAURI-RUST-AZ](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/464/?project=4&referrer=issue-list&statsPeriod=14d)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error recovery for Ollama embedding integration to gracefully handle encoding-related failures, returning empty embeddings instead of reporting errors and allowing batch processing to continue.

* **Tests**
  * Added comprehensive test coverage for the new error recovery behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2834?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/embeddings/ollama.rs
+++ b/src/openhuman/embeddings/ollama.rs
@@ -152,6 +152,107 @@ impl OllamaEmbedding {
             .map_err(|e| anyhow::anyhow!("invalid Ollama base_url `{}`: {e}", self.base_url))?;
         Ok(format!("{}/api/embed", self.base_url))
     }
+
+    /// Sends a single text to Ollama and returns either the embedding, or
+    /// `None` if Ollama produced the NaN-encoding 500 wire shape for that
+    /// individual input. Any other failure (transport error, non-2xx without
+    /// NaN signature, malformed JSON, dimension mismatch, count mismatch)
+    /// surfaces as an `Err` so genuine bugs are still loud.
+    ///
+    /// Used by [`Self::embed_per_text_fallback`] to recover a batch that
+    /// failed wholesale on NaN — see TAURI-RUST-AZ.
+    async fn embed_one_with_nan_recovery(
+        &self,
+        text: &str,
+    ) -> anyhow::Result<Option<Vec<f32>>> {
+        let resp = self
+            .http_client()
+            .post(self.embed_url()?)
+            .json(&OllamaEmbedRequest {
+                model: self.model.clone(),
+                input: vec![text.to_string()],
+            })
+            .send()
+            .await
+            .map_err(|e| {
+                anyhow::anyhow!(
+                    "ollama embed request failed (is Ollama running at {}?): {e}",
+                    self.base_url
+                )
+            })?;
+
+        if !resp.status().is_success() {
+            let status = resp.status();
+            let body = resp.text().await.unwrap_or_default();
+            let detail = body.trim();
+            // Per-text NaN: substitute empty vector, log once.
+            if status.as_u16() == 500 && is_nan_encode_error(&body) {
+                tracing::warn!(
+                    target: "embeddings.ollama",
+                    "[embeddings] ollama produced NaN for a single text (model={}); \
+                     substituting empty embedding (downstream blob will be 0 bytes)",
+                    self.model
+                );
+                return Ok(None);
+            }
+            anyhow::bail!(
+                "ollama embed failed with status {status}{}",
+                if detail.is_empty() {
+                    String::new()
+                } else {
+                    format!(": {detail}")
+                }
+            );
+        }
+
+        let payload: OllamaEmbedResponse = resp
+            .json()
+            .await
+            .map_err(|e| anyhow::anyhow!("ollama embed response parse failed: {e}"))?;
+        if payload.embeddings.len() != 1 {
+            anyhow::bail!(
+                "ollama embed count mismatch: sent 1 text, got {} embeddings",
+                payload.embeddings.len()
+            );
+        }
+        let v = payload.embeddings.into_iter().next().unwrap();
+        if v.len() != self.dims {
+            anyhow::bail!(
+                "ollama embed dimension mismatch: expected {}, got {}",
+                self.dims,
+                v.len()
+            );
+        }
+        Ok(Some(v))
+    }
+
+    /// Recovery path invoked when a batch request fails with the Ollama
+    /// NaN-encoding 500 wire shape. Re-sends each live input one at a time;
+    /// per-text NaN failures are filled with `Vec::new()` so the overall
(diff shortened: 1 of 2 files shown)


src/openhuman/embeddings/ollama.rs
```diff
@@ -152,6 +152,104 @@ impl OllamaEmbedding {
...
+            .json()
+            .await
+            .map_err(|e| anyhow::anyhow!("ollama embed response parse failed: {e}"))?;
+        if payload.embeddings.len() != 1 {
+            anyhow::bail!(
+                "ollama embed count mismatch: sent 1 text, got {} embeddings",
+                payload.embeddings.len()
+            );
+        }
+        let v = payload.embeddings.into_iter().next().unwrap();
```
graycyrus, line 215: **[minor]** Bare `.unwrap()` against project convention. The guard on `payload.embeddings.len() != 1` makes this unreachable, but prefer `.expect("validated len == 1 above")` to make the invariant explicit and stay consistent with the no-unwrap rule in Rust core.

graycyrus (approved):
Solid fix for TAURI-RUST-AZ. The design is correct — intercept the NaN-encoding 500 before the non_2xx bail path, recover per-text, preserve positions, and keep Sentry signal alive via the single batch-level report. One minor nit below.

**Change summary**

| File | What changed |
|---|---|
| `src/openhuman/embeddings/ollama.rs` | `is_nan_encode_error` matcher; `embed_one_with_nan_recovery` (per-text retry with NaN short-circuit); `embed_per_text_fallback` (batch recovery); NaN branch inserted into `embed` before the existing non_2xx bail |
| `src/openhuman/embeddings/ollama_tests.rs` | 5 new tests covering: matcher correctness, batch recovery, all-NaN substitution, single-text short-circuit (request count asserted), mixed outcomes with position preservation, non-NaN 500 still errors |

**What I verified**
- NaN check fires before the `report_error_or_expected(non_2xx)` + `bail!` so non-NaN 500s retain their existing loud path — correct.
- `live.len() == 1` short-circuit avoids a guaranteed-redundant retry — correct.
- `embed_per_text_fallback` returns a vec of length `total_len` regardless of how many NaN-fail, preserving the length contract — correct.
- Empty-vec semantics match the blank-input convention already in place — no caller contract change.
- Trait signature unchanged.

**Minor**

`src/openhuman/embeddings/ollama.rs` line 215:
```rust
let v = payload.embeddings.into_iter().next().unwrap();
```
The guard above (`if payload.embeddings.len() != 1 { bail! }`) makes this unreachable in practice, but the project convention is no bare `.unwrap()` in Rust core. Prefer `.expect("validated len == 1 above")` or the functional equivalent `.ok_or_else(|| anyhow::anyhow!("...")); `.

Not blocking — logic is safe. Everything else looks good.

PR #2878, opened 2026-05-28
perf(learning/user_profile): Aho-Corasick DFA for preference extraction + pattern coverage

mysma-9403:
## Summary

Replaces the `str::contains`-per-pattern preference scan in `UserProfileHook` with a single Aho-Corasick DFA pass per sentence, eliminates the previous per-turn `String` allocations from re-lowercasing, and expands both the pattern coverage and the sentence delimiter set so more real-world preference statements are captured.

This is the same class of optimization as #2842 (prompt-injection RegexSet) and #2870 (routing/quality AC) — moving a hot-path "scan input against N curated patterns" loop from O(N) substring sweeps over a re-allocated lowercase buffer to a single byte-level DFA pass.

## What changes

`src/openhuman/learning/user_profile.rs`:

- **DFA built once.** `PREFERENCE_DFA: LazyLock<AhoCorasick>` compiled with `ascii_case_insensitive(true)` + `MatchKind::LeftmostFirst`. No `to_lowercase()` of the message or per-sentence trim/lowercase pass any more.
- **Word-boundary check.** The byte immediately after each match must be non-ASCII-alphanumeric **and** at least one byte of trailing content must exist. This rejects:
  - `"I preferred X"` (alphanumeric continuation — false positive against `"i prefer"`)
  - degenerate empty-tail fragments like `"I prefer"` (the residue of splitting `"I prefer."`) or dangling `"I prefer:"` — they carry no preference target and would just pollute the `user_profile` memory namespace with useless `pref/i_prefer` slugs.
- **Fallback path dropped.** The previous `lower.starts_with(...)` post-loop rescue existed only because the substring sweep couldn't catch the `"I prefer:X"` shape. Proper word boundaries handle it in the main path.

`PREFERENCE_PATTERNS`: **20 → 28** entries:

- Direct preference: `i'd prefer`, `i would prefer`, `i'd rather`, `i dislike`
- Habit/instruction: `please use`
- Identity/context: `my pronouns`, `my preferred`, `call me`, `address me as`

Trailing whitespace removed from every pattern (the word-boundary check now does that job).

`SENTENCE_DELIMITERS`: extended from `['.', '!', '\n']` to `['.', '!', '?', ';', '\n']` so questions and semicolon-joined clauses decompose correctly (`"What's your view? I prefer Rust."`, `"OK; I prefer Rust."`). `:` is intentionally **not** a delimiter so `"My role: engineer"` stays one sentence the `"my role"` pattern can match.

`Cargo.toml`: adds `aho-corasick = "1.1"` as a direct dependency (already pulled in transitively via `regex`).

## Why this matters

`UserProfileHook` runs on **every user turn** as a `PostTurnHook`. The old code shape was:

```rust
let lower = message.to_lowercase();          // 1 String alloc
for sentence in message.split(...) {
    let sentence_lower = trimmed.to_lowercase(); // N String allocs
    for pattern in PREFERENCE_PATTERNS {          // 20 substring scans per sentence
        if sentence_lower.contains(pattern) { ... }
    }
}
```

For a 5-sentence user message that's 6 `String` allocations + 100 substring scans, every turn, on the agent hot path — plus you can't even add patterns cheaply because each one extends the inner loop. The new shape is one DFA, one byte-level pass per sentence, **no per-call allocation until a match is emitted**. Pattern count is no longer a cost dial.

The pattern expansion is the user-visible win: `"I'd prefer concise responses"`, `"Call me Alex"`, `"My pronouns are they/them"`, `"My preferred editor is Helix"`, `"Please use snake_case"` — all previously slipped through because they didn't match any of the 20 hard-coded openings. That's the kind of preference statement the memory system exists to capture.

The expanded delimiters (`?`, `;`) close another silent gap: a leading question like `"What's the timezone situation? My timezone is PST."` previously bled the preamble into the preference sentence, which then either failed the length filter or stored noise alongside the actual preference.

## Test plan

- [x] `cargo fmt`
- [x] `cargo check --manifest-path Cargo.toml`
- [x] `cargo test -p openhuman --lib learning::user_profile` — **14 passed**
- [x] Pre-push hook (`pnpm rust:check`, `compile`, `lint`, `lint:commands-tokens`) — passed

15 unit tests total (was 5):

- happy path (`finds_patterns`, `handles_single_sentence_message`)
- negative path (`handles_no_matches`, `ignores_short_sentences`, `rejects_bare_pattern_with_no_content_after`)
- cap enforcement (`caps_at_max_per_turn`)
- word-boundary correctness — alphanumeric reject **and** non-alphanumeric accept
- expanded delimiters (`?` and `;`)
- expanded patterns (`catches_extended_patterns` — one assertion per new pattern category so any future drop fails loudly)
- Unicode safety (`non_ascii_does_not_panic_or_falsely_match` — Cyrillic, Polish diacritics, emoji prefix)
- DFA smoke test (`preference_dfa_compiles_and_has_expected_pattern_count` — guards against silently dropping a pattern from the slice)
- existing mocked-storage and on-turn-behaviour tests retained unchanged

## Summary by CodeRabbit

* **Improvements**
  * Enhanced user preference extraction accuracy with stricter boundary detection to reduce false matches.
  * Improved handling of edge cases and punctuation in preference recognition.
  * Optimized performance for preference matching and extraction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2878?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@mysma-9403 hey! the code looks good to me, but CI is still pending on several checks (Rust Core Tests, coverage runs, E2E Appium suites, frontend unit tests). once those are green, I'll come back and approve this. let me know if you need any help!

For reference, what I reviewed: the Aho-Corasick DFA swap is a clean, well-justified optimization — the word-boundary logic is correct, the LazyLock initialization is safe, `LeftmostFirst` + `find_iter` + `.any()` short-circuits correctly, and the 15-test suite covers the important edge cases (alphanumeric continuation rejection, bare-pattern empty-tail rejection, expanded delimiters, Unicode safety, DFA pattern count guard). The direct `aho-corasick = "1.1"` dep is the right call — it's already in the transitive tree via `regex`, BurntSushi's package, MIT/Unlicense, actively maintained.

senamakel (dismissed):

PR #2881, opened 2026-05-28
fix(memory-conversations): release STORE_LOCK during cold index rebuild (#2849)

### LINKED ISSUE #2849: memory_conversations: first cross-thread search holds both locks during JSONL rebuild
## Background

PR #2756 replaced the linear `search_cross_thread_messages` scan with an in-memory inverted index. The index is lazy: on first access for a workspace, `ConversationStore::with_index` (`src/openhuman/memory_conversations/store.rs:195`) builds it by walking every per-thread JSONL file via `populate_index_unlocked` (`store.rs:212`).

That first-access rebuild runs **while holding both** `CONVERSATION_STORE_LOCK` *and* `CONVERSATION_INDEX_CACHE.lock()`. The lock-ordering invariant documented at the cache static requires the outer lock to be held first, so this is correct — but it means the first search after process start pauses every conversation-store operation (`append_message`, `delete_thread`, `get_messages`, `list_threads`, …) for the full rebuild duration.

For a workspace with ~100k messages this is a multi-second stall on the hot chat path (the consumer is `agent/memory_loader.rs:311`, called every turn).

## Why this is its own issue

The PR description already names a longer-term Roaring/FST/BM25/persisted-segments follow-up. That track is about replacing the **data structures**. This issue is about **lock surgery on the rebuild path** — it's an independent workstream and shouldn't have to wait for the bigger rewrite to land.

## Possible approaches

1. **Drop the outer guard around `populate_index_unlocked` after capturing a thread-list snapshot.** JSONL files are append-only and the rebuild is idempotent; a concurrent `append_message` simply gets indexed by the warm-cache path on the next call. This is the cheapest fix.
2. **`tokio::sync::OnceCell` per workspace.** First search blocks on the build, but other store operations don't. Adds a runtime dependency (the store is currently sync), so probably not v1.
3. **Background priming on store construction.** Fire-and-forget rebuild on `ConversationStore::new` so the cache is usually warm by the time the first search arrives.

(1) is the obvious cut; happy to take it or hand off.

## Acceptance

- First `search_cross_thread_messages` on a cold workspace does not block concurrent `append_message` / `get_messages` for more than a small constant time (snapshot capture only).
- Test added that asserts the rebuild path can interleave with at least one concurrent write without serialising on it.

## Refs

- Introduced in #2756 (merged)
- Reviewed in https://github.com/tinyhumansai/openhuman/pull/2756#pullrequestreview (approval body documented this as deferred)

staimoorulhassan:
## Summary

- Fixes #2849 — first `search_cross_thread_messages` call on a large workspace blocked all concurrent `append_message` / `get_messages` calls for multiple seconds because `CONVERSATION_STORE_LOCK` was held for the entire cold JSONL rebuild.
- Introduces `prime_index_if_cold()`, which snapshots the thread list under `CONVERSATION_STORE_LOCK` (fast, releases immediately), then reads all per-thread JSONL files with **no lock held** (safe — append-only), then inserts with `entry().or_insert()` (idempotent against concurrent primes).
- `search_cross_thread_messages` calls `prime_index_if_cold()` before acquiring `CONVERSATION_STORE_LOCK`, so the slow JSONL walk never blocks writers.
- `with_index` retains its existing cold-build fallback as a safety net.

## Root cause

Before this fix the call graph was:

    search_cross_thread_messages
      CONVERSATION_STORE_LOCK.lock()      <- held for entire rebuild
        CONVERSATION_INDEX_CACHE.lock()
          populate_index_unlocked()       <- reads ALL per-thread JSONL
          <- append_message stalls here

## Fix

After the fix:

    search_cross_thread_messages
      prime_index_if_cold()              <- no long lock held
        STORE_LOCK (thread snapshot -> release)
        read all JSONL (no lock)
        CACHE_LOCK (entry().or_insert -> release)
      CONVERSATION_STORE_LOCK.lock()     <- acquired AFTER rebuild
        with_index (warm cache, instant)

## Test plan

- [x] `search_cold_rebuild_does_not_block_concurrent_append` — races a cold search against a concurrent append behind a `Barrier`, asserts the append completes within 5 s (would block under old code on large workspaces)
- [x] All existing `store_tests` continue to pass
- [x] `cargo check` clean (CI)

## Summary by CodeRabbit

* **Bug Fixes**
  * Cold-index rebuilds during cross-workspace searches no longer block concurrent message appends, improving responsiveness.
  * Cold-indexing now tolerates legacy workspace formats and skips unreadable per-thread data instead of stalling.

* **Documentation**
  * Clarified concurrency/lock-ordering guidance to avoid long stalls.

* **Tests**
  * Added regression tests verifying concurrent search and append complete independently and within expected time bounds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2881?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@staimoorulhassan — CI is still pending on this, so I'll hold off on a formal approval. I did spot one thing while reading through, and it's blocking.

The major CodeRabbit finding about `list_threads_unlocked` (line 230) is correct and needs to be fixed before this lands. The problem: `prime_index_if_cold` calls `self.list_threads_unlocked()` while holding `CONVERSATION_STORE_LOCK`. On workspaces where any thread has a `None` message count (common pre-Stats workspaces), `list_threads_unlocked` triggers `measure_messages_unlocked` per-thread — which reads every per-thread JSONL and appends `ThreadLogEntry::Stats` to `threads.jsonl`. All of that runs under the store lock, reintroducing exactly the multi-second stall this PR set out to fix. The fix is to snapshot only thread IDs via `thread_index_unlocked()` (JSONL-header-only, no per-thread reads) under the lock, then call `list_threads_unlocked` (or equivalent backfill logic) outside the lock before building the inverted index.

The regression test won't catch this either — it uses a fresh `TempDir` with threads created via `ensure_thread`, so `message_count` is always populated from the start. The backfill path never triggers in that setup. A test that exercises an existing workspace with `None` message counts (simulating a pre-Stats workspace) would catch this.

The rest of the PR is solid. The `prime_index_if_cold` design is correct for the normal path, the `entry().or_insert()` idempotency is right, the double-check pattern is correct, and the test structure is good. The two minor comment fixes CodeRabbit flagged (lines 53 and 262) are accurate — worth addressing while you're in there.

Fix the `list_threads_unlocked` call + add a backfill-path test case and this should be good to go.

graycyrus (commented):
@staimoorulhassan hey! the code looks good to me — the third commit switching to `thread_index_unlocked` in `prime_index_if_cold` properly addresses the backfill issue, and the two new legacy-workspace tests cover exactly the scenario I was worried about. CI still has some pending checks, so once those are green i'll come back and approve this. let me know if you need any help!

staimoorulhassan (commented):
resolved

staimoorulhassan:
**Fixed compile error from bad merge.** The merge commit `f60fff9e` brought in upstream main changes, which left stale code from the old `search_cross_thread_messages` fast-path block inside the new `with_index` function body. Specifically, the `if !cache.contains_key(&key)` block was left unclosed and the orphaned inline fast-path block (lines that referenced `query/limit/exclude_thread_id`) sat inside it — causing "this file contains an unclosed delimiter" at line 1025. Removed the orphaned block and restored the correct closing:
```rust
    }
    let idx = cache.get_mut(&key).expect("inserted above if absent");
    Ok(f(idx))
}
```
All other CI checks (TypeScript, Frontend tests, Rust tests) should now pass.

src/openhuman/memory_conversations/store_tests.rs
```diff
@@ -957,3 +957,302 @@ fn read_jsonl_skips_invalid_lines_but_keeps_valid_ones() {
...
+
+    let ts = "2026-04-10T12:00:00Z".to_string();
+
+    // Fresh TempDir → path never seen by the process-level cache → cold.
+    // Note: append_message only updates an *existing* cache entry; it never
+    // inserts one, so the cache stays cold until the first search call.
+    let temp = TempDir::new().unwrap();
+    let store = ConversationStore::new(temp.path().to_path_buf());
+
+    store
```
senamakel, line 990: **Deferred (out of scope for this fix pass):** CodeRabbit's suggestion to split this test module (now ~1.2k lines, over the ~500-line guideline) into a `store_tests/` directory is reasonable, but mechanically splitting it touches no production behavior and risks churn unrelated to the #2849 lock fix. Leaving the file intact here; recommend a dedicated follow-up PR to carve the search/concurrency/legacy suites into focused files.

The three lock-fix blockers/nits *were* addressed in commits 548d2539, 9f8a5c5, 80c4c0e: (1) repaired the merge-mangled `populate_index_unlocked` (it re-locked `CONVERSATION_STORE_LOCK` while the caller held it and returned the wrong type — a non-compiling deadlock); (2) corrected the lock-ordering and raced-message doc comments; (3) widened the two cold-rebuild concurrency-test timeouts 5s→30s to avoid CI flakes.

oxoxDev: Agreed on deferring the test-module split — mechanically carving a 1.2k-line test file into `store_tests/` is pure churn against a lock fix; a focused follow-up PR is the right call. I confirmed the three lock-fix items in the current diff: `populate_index_unlocked` no longer re-acquires `CONVERSATION_STORE_LOCK` (uses the `_unlocked` readers, returns `Result<(), String>`), the lock-ordering + raced-message doc comments read correctly, and the cold-rebuild concurrency tests are at the wider timeout. Looks good — approving.

staimoorulhassan (commented):
waiting for result

oxoxDev (approved):
Approve. Reviewed the concurrency design end-to-end:

- **Lock ordering** is consistent (`STORE_LOCK` → `INDEX_CACHE`); `prime_index_if_cold` only ever holds both momentarily for the header-only thread snapshot, then does the JSONL walk lock-free.
- **No index-corruption race**: confirmed `append_message`'s cache update is `get_mut` + `if let Some` — it only mutates an already-warm entry, never creates one. So during the prime window the cache stays cold, append skips its in-memory update, and the final `contains_key`-gated insert lands the full JSONL build cleanly. The only discard case is a concurrent *prime* winning, which is itself a full build — equivalent.
- The documented tradeoff (a message appended mid-prime is missing from the in-memory index until the next cold rebuild) is acceptable — JSONL is source of truth.
- Good call documenting why `list_threads_unlocked` must not run under the lock (its `Stats` append reintroduces the stall).
- Regression test races a cold search vs concurrent append behind a `Barrier` with a fresh TempDir (cold cache) — genuinely exercises the fix.

CI red is only the unrelated Gmail-E2E web-lane flake (`gmail-flow.spec.ts`); the E2E Gate itself passes and this PR touches only `memory_conversations`.

PR #2884, opened 2026-05-29
fix(cron): accept bare cron string in Schedule deserialization (#2866)

### LINKED ISSUE #2866: Cron update fails: Schedule field rejects bare cron strings
## Summary

The `openhuman.cron_update` RPC method fails when the schedule field is a bare cron string (e.g. `"0 9 * * 1"`). The deserializer expects an internally tagged enum `Schedule` but receives a plain string.

## Problem

**What happened:** User tries to update a cron job schedule via the API/UI. The patch payload includes a cron expression as a string, but `serde` rejects it:
```
invalid 'patch': invalid type: string "0 9 * * 1", expected internally tagged enum Schedule
```

**Expected:** The `Schedule` type should accept both a bare cron string and a structured enum format for backwards compatibility.

**Impact:** 1 event in Sentry (CORE-RUST-FY), high priority. Also CORE-RUST-FX with `"0 22 * * *"` — same bug. Users cannot update cron schedules.

**Version / Platform:** openhuman@0.56.0, Linux 6.8.0-111-generic, rustc 1.93.0

## Solution (optional)

The `Schedule` enum in `src/openhuman/cron/` likely needs either:
1. `#[serde(untagged)]` to accept both string and struct variants
2. A custom deserializer that tries parsing as a cron string first, then as the enum
3. A string variant added to the enum (e.g. `Schedule::Cron(String)`)

## Acceptance criteria

- [ ] **Repro gone** — `openhuman.cron_update` accepts bare cron strings like `"0 9 * * 1"`
- [ ] **Regression safety** — Unit test for deserializing both formats

## Related

- Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/6332/
- Also: CORE-RUST-FX (same bug, different cron expression)

YOMXXX:
## Summary

- Fix `openhuman.cron_update` rejecting bare cron strings like `"0 9 * * 1"` — the `Schedule` enum now accepts both the bare string format and the existing tagged object format `{"kind": "cron", "expr": "..."}`

## Problem

Callers (including LLM tool calls) pass `{"schedule": "0 9 * * 1"}` but serde expected `{"schedule": {"kind": "cron", "expr": "0 9 * * 1"}}`, resulting in:
```
invalid type: string "0 9 * * 1", expected internally tagged enum Schedule
```

## Solution

Replaced the derived `Deserialize` on `Schedule` with a manual impl that:
1. Tries bare string first → wraps as `Schedule::Cron { expr, tz: None, active_hours: None }`
2. Falls back to the existing internally-tagged enum format
3. Serialization remains unchanged (always outputs the tagged format)

## Changes

- `src/openhuman/cron/types.rs`: Custom `Deserialize` impl + 3 new tests

## Test plan

- [x] `schedule_deserializes_bare_cron_string` — bare string parses correctly
- [x] `schedule_deserializes_tagged_cron_object` — existing format still works
- [x] `schedule_bare_string_roundtrip_serializes_as_tagged` — bare input serializes to tagged output
- [x] All existing Schedule tests unaffected

Closes #2866

## Summary by CodeRabbit

* **New Features**
  * Schedule configuration now accepts bare cron strings for simplified input, reducing complexity for basic use cases while maintaining full backward compatibility with the existing structured format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2884?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/cron/types.rs
+++ b/src/openhuman/cron/types.rs
@@ -57,7 +57,7 @@ pub struct ActiveHours {
     pub end: String,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 #[serde(tag = "kind", rename_all = "lowercase")]
 pub enum Schedule {
     Cron {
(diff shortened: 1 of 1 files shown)


src/openhuman/cron/types.rs
```diff
@@ -75,6 +75,59 @@ pub enum Schedule {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = serde_json::Value::deserialize(deserializer)?;
+
+        // Bare cron string: treat as Cron variant with defaults.
+        if let Some(expr) = value.as_str() {
+            return Ok(Schedule::Cron {
+                expr: expr.to_string(),
```
graycyrus, line 88: **[minor]** The bare string path accepts any string without validating it as a cron expression — `serde_json::from_str::<Schedule>(r#"\"not a cron\""#)` succeeds silently. The failure then surfaces later at execution time, which can be confusing to debug. This is pre-existing behavior on the tagged path too (`expr: String` is unvalidated throughout), but LLM callers are more likely to send malformed strings than typed API callers.

Not blocking — the pre-existing behavior is the same — but if you ever add cron expression validation to `Schedule::Cron`, a test for the invalid-bare-string rejection case would pair well with the three already here.

graycyrus (approved):
Good fix. The bug is real (Sentry has two confirmed events, different cron expressions, same root cause), the solution is correct, and the tests cover what they need to cover.

Issue alignment: all three acceptance criteria from #2866 are satisfied — bare string deserialization works, regression tests exist, diff coverage is solid.

One minor note below on the validation gap (unrelated to CodeRabbit's structural duplication point, which they already covered).

YOMXXX:
Closing — upstream main already contains a more thorough fix for this issue using a proper serde Visitor pattern (visit_str + visit_map + deserialize_any) rather than the serde_json::Value approach in this PR. The upstream fix is superior as it doesn't depend on serde_json for deserialization.

PR #2886, opened 2026-05-29
fix(conversations): release store lock during cold index rebuild (#2849)

### LINKED ISSUE #2849: memory_conversations: first cross-thread search holds both locks during JSONL rebuild

YOMXXX:
## Summary

- Fix lock contention where the first `search_cross_thread_messages` call after process start blocks all conversation-store operations (`append_message`, `get_messages`, `list_threads`, etc.) for the full JSONL rebuild duration

## Problem

`with_index` held both `CONVERSATION_STORE_LOCK` and `CONVERSATION_INDEX_CACHE.lock()` while walking every per-thread JSONL file to build the inverted index. For workspaces with ~100k messages this is a multi-second stall on the hot chat path (`agent/memory_loader.rs:311`, called every turn).

## Solution

Split-lock strategy per approach (1) from the issue:

- **Fast path (warm cache):** acquires only the cache lock — no outer store lock needed
- **Cold path (first access):** acquires store lock briefly to snapshot thread list, releases it, reads JSONL files lock-free, then acquires cache lock to insert the index

JSONL files are append-only, so a concurrent `append_message` during rebuild is safe — the new message gets indexed by the warm-cache `insert` path on the next call.

## Changes

- `store.rs`: Rewrote `search_cross_thread_messages` with split-lock, removed `with_index` and `populate_index_unlocked` (inlined), updated lock-ordering docs
- `store_tests.rs`: Added `cold_search_does_not_serialize_on_outer_lock` — concurrent write during cold rebuild

## Test plan

- [x] `cold_search_does_not_serialize_on_outer_lock` — verifies no deadlock on concurrent write during cold rebuild
- [x] All existing conversation store tests unaffected

Closes #2849

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents potential deadlocks during cross-thread message searches by changing locking so searches use a cached index when available and avoid holding the global store lock for the full operation.

* **Tests**
  * Added a concurrency test ensuring searches and concurrent message appends can run without deadlock and return expected results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2886?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory_conversations/store_tests.rs
+++ b/src/openhuman/memory_conversations/store_tests.rs
@@ -847,6 +847,74 @@ fn update_thread_labels_missing_thread_returns_error() {
 }
 
 #[test]
+fn cold_search_does_not_serialize_on_outer_lock() {
+    // Issue #2849: verify that a cold-cache search releases the store
+    // lock before the JSONL rebuild, so concurrent writes aren't blocked.
+    let (_temp, store) = make_store();
+
+    // Seed a thread with a message so the search has something to find.
+    store
+        .ensure_thread(CreateConversationThread {
+            id: "t1".to_string(),
+            title: "test thread".to_string(),
+            created_at: "2026-01-01T00:00:00Z".to_string(),
+            parent_thread_id: None,
+            labels: None,
+        })
+        .unwrap();
+    store
+        .append_message(
+            "t1",
+            ConversationMessage {
+                id: "m1".to_string(),
+                content: "hello world".to_string(),
+                message_type: "text".to_string(),
+                extra_metadata: json!({}),
+                sender: "user".to_string(),
+                created_at: "2026-01-01T00:00:00Z".to_string(),
+            },
+        )
+        .unwrap();
+
+    // Evict any warm cache so the next search triggers a cold rebuild.
+    {
+        let mut cache = CONVERSATION_INDEX_CACHE.lock();
+        cache.remove(&store.root_dir());
+    }
+
+    // Spawn a thread that tries to append a message while a cold search
+    // is (conceptually) running. In the old code this would deadlock or
+    // serialize behind the full rebuild; in the fixed code the store lock
+    // is released after the thread-list snapshot and the append succeeds
+    // concurrently.
+    let store2 = store.clone();
+    let writer = std::thread::spawn(move || {
+        store2
+            .append_message(
+                "t1",
+                ConversationMessage {
+                    id: "m2".to_string(),
+                    content: "concurrent write".to_string(),
+                    message_type: "text".to_string(),
+                    extra_metadata: json!({}),
+                    sender: "assistant".to_string(),
+                    created_at: "2026-01-01T00:00:01Z".to_string(),
+                },
+            )
+            .unwrap();
+    });
+
+    // Run the cold search — should not deadlock.
+    let results = store
+        .search_cross_thread_messages("hello", 10, None)
+        .unwrap();
+    assert!(!results.is_empty(), "search should find seeded message");
+
+    // The concurrent write must also succeed.
+    writer.join().expect("concurrent write must not deadlock");
+}
+
+#[test]
 fn read_jsonl_skips_invalid_lines_but_keeps_valid_ones() {
     let tmp = TempDir::new().unwrap();
     let path = tmp.path().join("messages.jsonl");
--- a/src/openhuman/memory_conversations/store.rs
+++ b/src/openhuman/memory_conversations/store.rs
@@ -36,16 +36,23 @@ static CONVERSATION_STORE_LOCK: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
 /// behind dead entries when the dir is removed — acceptable for an
 /// in-process cache.
 ///
-/// # Lock ordering (INVARIANT)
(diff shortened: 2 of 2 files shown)


graycyrus (commented):
@YOMXXX the split-lock strategy looks correct and the logic is sound. CI is still pending so I'll hold off on approving until those are green — address CodeRabbit's actionable comment on the misleading doc at line 202 while you wait.

One thing I noticed that CodeRabbit didn't flag:

The new test (`cold_search_does_not_serialize_on_outer_lock`) doesn't actually guarantee temporal interleaving between the cold rebuild and the concurrent write. There's no barrier or channel forcing the writer thread to execute during the JSONL read phase — depending on scheduling the writer can complete entirely before the search starts, meaning the test only proves "no deadlock" rather than "no serialization under true concurrency." The concurrency correctness is in the logic itself (which is solid), but the test name oversells what it demonstrates.

To actually stress the concurrent path, you could add a small `thread::sleep` or use a `Barrier` to synchronize the writer's start with the cold-rebuild window. Not a blocker, but the test name sets an expectation that isn't backed by the timing today.

Everything else — the `or_insert` race guard, the drop of `_guard` before JSONL reads, the warm-path early return, and the lock-ordering doc rewrite — looks right. Good fix.

src/openhuman/memory_conversations/store_tests.rs
```diff
@@ -846,6 +846,74 @@ fn update_thread_labels_missing_thread_returns_error() {
...
+    {
+        let mut cache = CONVERSATION_INDEX_CACHE.lock();
+        cache.remove(&store.root_dir());
+    }
+
+    // Spawn a thread that tries to append a message while a cold search
+    // is (conceptually) running. In the old code this would deadlock or
+    // serialize behind the full rebuild; in the fixed code the store lock
+    // is released after the thread-list snapshot and the append succeeds
+    // concurrently.
```
graycyrus, line 889: **[minor]** The writer is spawned and then the search is called immediately with no synchronization between them. Scheduler may run the writer to completion before the search even acquires the cache lock, meaning the cold-rebuild path never actually races with a concurrent write in this test run.

Consider using a `std::sync::Barrier` (or a channel) to hold the writer thread until the cold rebuild has started its JSONL reads. Something like: acquire the barrier in the writer before `append_message`, and release it inside a test-hook or after a brief sleep timed to the rebuild. Without this, the test name `cold_search_does_not_serialize_on_outer_lock` is stronger than what the test can actually assert.

graycyrus (commented):
@YOMXXX the doc comment fix in 96b8c1a is exactly right — the clarified wording accurately describes the missed-message tradeoff without implying false self-healing. Prior CodeRabbit finding addressed.

On the test synchronization note from my last review — the test can't guarantee true temporal interleaving since the writer may finish before the cold rebuild starts reading JSONL, but the structural correctness guarantee from the implementation is solid and the test does prove no deadlock. Resolving that thread as a non-blocker.

Code is clean. Waiting on CI — a few jobs are still pending (Rust Core Coverage, Build Tauri App, e2e Appium). Once those are green I'll approve.

Issue #2885, opened 2026-05-29
feat(memory-sync): migrate Notion/ClickUp/Linear/GitHub Composio providers off UnifiedMemory


justinhsu1477:
## Summary

While auditing the #2705 silent-failure root cause for #2720, I noticed the same architectural pattern — **provider writes to the legacy UnifiedMemory backend, retrieval reads from memory_tree** — repeated across **four** native Composio providers. Slack and Gmail were already migrated; the remaining four were not.

## Audit findings

\`grep\` over \`src/openhuman/memory_sync/composio/providers/\`:

| Provider | Ingest path | Backend written |
| --- | --- | --- |
| Slack | \`ingest_pipeline::ingest_chat\` | ✅ memory_tree (\`mem_tree_chunks\`) |
| Gmail | \`ingest_pipeline::ingest_email\` | ✅ memory_tree |
| **Notion** | \`persist_single_item\` → \`store_skill_sync\` | ❌ UnifiedMemory (\`memory_docs\`) |
| **ClickUp** | \`persist_single_item\` → \`store_skill_sync\` | ❌ UnifiedMemory |
| **Linear** | \`persist_single_item\` → \`store_skill_sync\` | ❌ UnifiedMemory |
| **GitHub** | \`persist_single_item\` → \`store_skill_sync\` | ❌ UnifiedMemory |

## Why this matters

Same symptom as #2705 but for Composio rather than vault:

- Synced Notion / ClickUp / Linear / GitHub items don't appear in \`mem_tree_chunks\` or \`mem_tree_ingested_sources\`.
- Every modern retrieval surface (\`memory.search\`, \`tree.read_chunk\`, \`tree.browse\`, the agent's recall path, summary trees, \`tree.top_entities\`) reads from memory_tree — so this data is invisible to all of them.
- Claude Desktop / Cursor / any MCP client doing \`memory.search\` against the user's OpenHuman gets **zero** Notion / ClickUp / Linear / GitHub hits even when the user has those toolkits connected.
- The legacy \`UnifiedMemory::list_documents\` / \`query_namespace\` API still surfaces the data for callers that use it directly, so this is a *partial* desync — less catastrophic than #2705's full-empty case, but still wrong.

## Fix shape

Per provider:

1. Add a new \`providers/<toolkit>/ingest.rs\` module that mirrors the Slack canonical pattern:
   - Single \`ingest_into_memory_tree(config, owner, connection_id, item)\` async function
[...]

PR #2887, opened 2026-05-29
feat(composio): migrate Notion provider to memory_tree pipeline (#2885)

justinhsu1477:
First of four migrations from #2885. Notion sync previously wrote to the legacy `UnifiedMemory` backend (`memory_docs`); every modern retrieval surface reads from memory_tree, so Notion content was invisible to `memory.search`, `tree.read_chunk`, `tree.browse`, the agent's recall path, summary trees, and the MCP `tree.*` tools.

Same architectural pattern as the vault sync silent failure fixed in #2720 (#2705) — just at the Composio provider layer instead of the file walker layer.

## Migration shape

| | Before | After |
| --- | --- | --- |
| Call site | `provider.rs::sync` → `persist_single_item` | `provider.rs::sync` → `ingest::ingest_page_into_memory_tree` |
| Memory path | `MemoryClient::store_skill_sync` → `UnifiedMemory::upsert_document` → `memory_docs` | `ingest_pipeline::ingest_document` → `mem_tree_chunks` + `mem_tree_ingested_sources` |
| Source id | `composio-notion-page-{page_id}` (UnifiedMemory document_id) | `notion:{connection_id}:{page_id}` (stable, dedup-able) |
| Visibility to retrieval | Nothing | Every `memory.search` / `tree.*` / MCP surface |

Mirrors the canonical Slack / Gmail layout: a new `providers/notion/ingest.rs` owns the page → memory-tree translation. The existing dedup / cursor / budget state-machine (`SyncState::synced_ids` keyed by `{page_id}@{edited_time}`) is unchanged.

## Three design choices worth flagging

1. **Per-page source-id scope** — one source per `(connection_id, page_id)`. Page is the natural Notion grouping, keeps `SourceKind::Document` semantics, and parallels Gmail per-message / Slack per-channel sourcing. Distinct connections produce distinct source_ids so the same Notion page seen by two connected accounts doesn't cross-contaminate.

2. **Re-ingest-on-edit cleanup** — Notion pages mutate, but the pipeline's `already_ingested` gate is content-blind. The new module calls `delete_chunks_by_source` before each re-ingest so the new revision actually lands. Same pattern as the vault sync content-update path in #2720.

3. **Full-page JSON in the body** — `# {title}\n\n` followed by the pretty-printed payload in a fenced code block. Notion pages don't have a natural single-string canonical body the way Slack messages do, so the chunk content retains enough Notion-specific signal (properties, URL, excerpt) that retrieval can match on more than just the title.

## Tests

**27/27 pass** in `memory_sync::composio::providers::notion::*` (was 22, **+5 new** in `ingest::tests`):

- `ingest_page_writes_to_memory_tree` — **the #2885 regression**. Asserts `count_chunks` goes up + `is_source_ingested` returns true for the expected `notion:{conn}:{page}` source_id. Mirrors the `sync_writes_to_memory_tree` regression in `vault::sync` from #2720.
- `notion_source_id_is_stable_and_namespaced` — pins the format the `delete_chunks_by_source` cleanup path relies on; distinct connections + distinct page ids must produce distinct source_ids.
- `parse_edited_time_handles_valid_and_invalid_inputs` — ISO 8601 / RFC 3339 parsing with safe fallback to `Utc::now()` instead of failing ingest.
- `render_page_body_includes_title_header_and_pretty_json` — pins the chunked-body shape so future "just keep the title" simplifications can't silently strip Notion-specific signal.
- `re_ingesting_edited_page_replaces_prior_chunks` — exercises the delete-first guard; edited page must replace prior chunks (no silent `already_ingested` no-op, no append-duplicate).

- [x] `cargo test --lib memory_sync::composio::providers::notion` — 27/27 pass.
- [x] `cargo check --lib` — clean.
- [x] `cargo test --tests --no-run` — clean (integration-test target compiles, lesson from #2603).

## PR sequence (from #2885)

> Tracked separately in #2885; not action items for this PR.

* **Notion** — this PR
* **GitHub** — follow-up (was my #2413)
* **Linear** — follow-up (was my #2402)
* **ClickUp** — follow-up (was my #2291; largest, leave for last after pattern is settled)

## Out of scope

- Removing `persist_single_item` / `store_skill_sync` is blocked on #2585's broader UnifiedMemory removal — they stay in place until then for any non-Composio caller.
- Reconciling old `memory_docs`-only rows already synced into UnifiedMemory pre-migration — separate question, depends on retention story.

Refs #2885, #2720, #2705, #2585.

## Summary by CodeRabbit

* **New Features**
  * Notion pages are ingested into the memory system with consistent tracking and default tags.
  * Edited Notion pages are updated in memory without creating duplicate entries.

* **Bug Fixes / Reliability**
  * Sync now preserves the cursor when ingest failures occur so failed ranges are retried; failures are logged.
  * More robust handling of edited timestamps with a safe fallback for malformed or missing values.

* **Tests**
  * Unit and regression tests added to validate ingest stability and re-ingest behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2887?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
@@ -181,7 +181,11 @@ pub async fn ingest_page_into_memory_tree(
             Ok(chunks_written)
         }
         Err(err) => Err(anyhow::anyhow!(
-            "ingest_document failed for {source_id}: {err}"
+            // `{err:#}` (alternate formatter) bakes in the anyhow context
+            // chain so provider.rs's `tracing::warn!(error = %e)` doesn't
+            // strip the underlying cause (DB / embedding / persist failure)
+            // when it Displays the wrapped error.
+            "ingest_document failed for {source_id}: {err:#}"
         )),
     }
 }
--- a/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
@@ -139,11 +139,11 @@ pub async fn ingest_page_into_memory_tree(
     .await
     .map_err(|e| anyhow::anyhow!("delete-prior task join error: {e}"))??;
     if removed > 0 {
-        log::debug!(
-            "[composio:notion] ingest: re-ingest cleanup connection_id={} page_id={} removed_chunks={}",
-            connection_id,
-            page_id,
-            removed
+        tracing::debug!(
+            connection_id = %connection_id,
+            page_id = %page_id,
+            removed_chunks = removed,
+            "[composio:notion] ingest: re-ingest cleanup"
         );
     }
 
@@ -171,12 +171,12 @@ pub async fn ingest_page_into_memory_tree(
             // the normal update path. Seeing it here means the prior
             // chunks were already absent (fresh ingest into a primed
             // memory_tree) — fine, just log at debug.
-            log::debug!(
-                "[composio:notion] ingest: page connection_id={} page_id={} chunks_written={} already_ingested={}",
-                connection_id,
-                page_id,
+            tracing::debug!(
+                connection_id = %connection_id,
+                page_id = %page_id,
                 chunks_written,
                 already_ingested,
+                "[composio:notion] ingest: page persisted"
             );
             Ok(chunks_written)
         }
--- a/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/notion/ingest.rs
@@ -44,7 +44,7 @@ use serde_json::Value;
 
 use crate::openhuman::config::Config;
 use crate::openhuman::memory::ingest_pipeline::{self, IngestResult};
-use crate::openhuman::memory_store::chunks::store::delete_chunks_by_source;
+use crate::openhuman::memory_store::chunks::store::{delete_chunks_by_source, is_source_ingested};
 use crate::openhuman::memory_store::chunks::types::SourceKind;
 use crate::openhuman::memory_sync::canonicalize::document::DocumentInput;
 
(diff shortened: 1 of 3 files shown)


graycyrus (commented):
@justinhsu1477 hey! the code looks good to me, but CI has a failing check (PR Submission Checklist) and several jobs still pending — once everything is green, i'll come back and approve this. let me know if you need any help.

Walkthrough: adds `providers/notion/ingest.rs` with `ingest_page_into_memory_tree`, swaps the call site in `provider.rs` from `persist_single_item` → `MemoryClient::store_skill_sync` → `memory_docs` to the canonical memory-tree pipeline, and ships 5 targeted tests. Mirrors the Slack/Gmail ingest module shape exactly.

The implementation is clean throughout. `delete_chunks_by_source` is correctly hopped through `spawn_blocking`. The double-`?` on the task join surfaces both the join panic and the inner `Result` cleanly. No `.unwrap()` in production paths. The `already_ingested` log comment after the delete-first guard is accurate — you'd only hit that on a fresh page where prior chunks were already absent, not a real short-circuit. The 5 new tests cover the regression, source_id stability, timestamp parsing, body shape, and the delete-replace round-trip — all the right things to pin.

Change summary:

| File | Change |
| --- | --- |
| `notion/ingest.rs` | New module: `notion_source_id`, `render_page_body`, `parse_edited_time`, `ingest_page_into_memory_tree`, 5 tests |
| `notion/mod.rs` | Exposes the new `ingest` submodule |
| `notion/provider.rs` | Replaces `persist_single_item` call with `ingest_page_into_memory_tree`; removes unused `memory` binding |

AI risk assessment: Breaking risk low (self-contained to Notion provider, `persist_single_item` stays for other callers, source_id format change is internal dedup-key only). Security risk zero (no auth surface, no user-facing input, no secrets). Safe to merge once CI is green.

justinhsu1477:
Two fixes pushed in `dd3943d6`:

| Item | Fix | Where |
| --- | --- | --- |
| CodeRabbit nit (gate `delete_chunks_by_source` behind `is_source_ingested`) | Applied — the fresh-page case now converts to one PK-indexed `COUNT(*)` against `mem_tree_ingested_sources` instead of a full `source_kind = Document` scan with Rust-side filter. Both calls share the same `spawn_blocking` hop. | `notion/ingest.rs` lines 113-141 |
| PR Submission Checklist failing — 3/8 "unchecked" | Edited the PR body. The "PR sequence (from #2885)" section used `- [ ]` lines as a sequence tracker for the *follow-up* PRs (GitHub / Linear / ClickUp); the checklist linter read them as action items for *this* PR. Switched to plain bullets under a "not action items" note. | PR description |

Same `is_source_ingested` gate carried over to the parallel GitHub PR (#2889) so the pattern stays consistent — 6/6 tests still green there too.

The 5 ingest tests on this PR still pass (`cargo test --lib memory_sync::composio::providers::notion::ingest`). The regression test exercises the new false → no-delete path (fresh page); the re-ingest test exercises the true → delete path (edited page).

@graycyrus — checklist should clear on the next push; waiting for CI to re-run.

justinhsu1477:
Pushed `0c15fd27` — applying both correctness fixes CodeRabbit caught on the parallel #2889 (same shape exists here, fixing proactively):

| Fix | Where | Mechanism |
| --- | --- | --- |
| Hold cursor on ingest failure | `notion/provider.rs` | New \`had_ingest_failures\` flag; \`advance_cursor\` now gated. Without this, an *edited* page that fails to re-ingest is left with **neither old nor new chunks** in \`mem_tree_chunks\` until its next edit — the delete-first re-ingest path makes the prior \"stale revision\" failure mode into actual data loss. |
| Preserve anyhow chain on ingest_document Err | \`notion/ingest.rs\` | \`{err}\` → \`{err:#}\` so the wrapping anyhow's Display impl carries the chain. Without it, \`tracing::warn!(error = %e)\` on the provider side dropped underlying DB / embedding causes. |

Tests: 27/27 still pass in `memory_sync::composio::providers::notion::*`. The provider's `sync()` is hard to unit-test (would need to mock `ComposioContext` heavily) so the invariant is pinned by rustdoc + inline comments rather than a new test.

@graycyrus — should be the last blocker before re-approval.

PR #2889, opened 2026-05-29
feat(composio): migrate GitHub provider to memory_tree pipeline (#2885)

justinhsu1477:
## Summary

Second of the four-PR follow-up sequence in **#2885** (after Notion in **#2887**). The GitHub Composio provider now writes through `ingest_pipeline::ingest_document` → `mem_tree_chunks` instead of `MemoryClient::store_skill_sync` → UnifiedMemory `memory_docs`, so `memory.search` / `tree.read_chunk` / `tree.browse` / the agent recall path / summary trees all see GitHub issues and PRs again.

Same shape as the Notion migration in #2887, applied to `providers/github/`:

- **New `providers/github/ingest.rs`** — `github_source_id`, `parse_updated_at`, `render_issue_body`, `extract_html_url`, `ingest_issue_into_memory_tree`. Pretty-JSON markdown body retains labels/assignees/state/comment count for retrieval signal.
- **Re-ingest cleanup** via `delete_chunks_by_source` before each `ingest_document` call. Without this, the pipeline's content-blind `already_ingested` gate would silently drop edited issues — the same trap fixed in vault sync **#2720** and Notion **#2887**.
- **`source_id = github:{connection_id}:{issue_id}`** — stable across re-syncs and across connections. Works for both numeric internal IDs and the `owner/repo#number` slug fallback from `extract_issue_id`.
- **`source_ref` carries the canonical GitHub `html_url`** — audit trails one-click navigable from any downstream UI.
- **`provider.rs`** — dropped `persist_single_item` + `doc_id`, threaded `updated_at` through to the new ingest fn, kept the existing composite-key dedup logic (`{issue_id}@{updated_at}`) intact so the cursor-boundary fast-path still works.

`persist_single_item` / `store_skill_sync` intentionally left in place — they're still used by the remaining unmigrated providers (Linear, ClickUp) and the wider UnifiedMemory removal lives under the **#2585** follow-up. Becomes removable once Linear and ClickUp follow (also tracked in #2885).

## Test plan

Lib tests (all green locally):

- [x] `cargo test --lib -- memory_sync::composio::providers::github::ingest` → 6/6 pass
  - [x] `github_source_id_is_stable_and_namespaced` — pins the contract the cleanup path relies on
  - [x] `parse_updated_at_handles_valid_and_invalid_inputs` — falls through to `Utc::now()` on bad input
  - [x] `render_issue_body_includes_title_header_and_pretty_json` — pins the chunked-content shape
  - [x] `extract_html_url_handles_both_envelope_shapes` — handles direct + `data.html_url` Composio envelopes
  - [x] `ingest_issue_writes_to_memory_tree` — **the #2885 regression**: `count_chunks` rises and the source_id lands in `mem_tree_ingested_sources`
  - [x] `re_ingesting_edited_issue_replaces_prior_chunks` — delete-first guard works; chunk count stays steady (±1 chunker rounding) instead of doubling
- [x] `cargo test --lib -- memory_sync::composio::providers::github` → 48/48 pass (no other github-namespaced test broke)
- [x] `cargo check --lib` clean
- [x] `cargo test --tests --no-run` — integration tests still compile

Refs #2885. Follows #2887 (Notion). Linear and ClickUp PRs come next.

## Summary by CodeRabbit

* **New Features**
  * GitHub issues and pull requests can be ingested into the memory system with title extraction, timestamp parsing, source references, and stable per-item source IDs.

* **Bug Fixes**
  * Re-ingesting edited items replaces prior in-memory content instead of duplicating.
  * Sync now retains progress when ingest failures occur so failed ranges are retried.

* **Tests**
  * Added unit and async regression tests for IDs, timestamp fallback, body rendering, URL extraction, and re-ingest behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2889?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory_sync/composio/providers/github/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/github/ingest.rs
@@ -164,11 +164,11 @@ pub async fn ingest_issue_into_memory_tree(
     .await
     .map_err(|e| anyhow::anyhow!("delete-prior task join error: {e}"))??;
     if removed > 0 {
-        log::debug!(
-            "[composio:github] ingest: re-ingest cleanup connection_id={} issue_id={} removed_chunks={}",
-            connection_id,
-            issue_id,
-            removed
+        tracing::debug!(
+            connection_id = %connection_id,
+            issue_id = %issue_id,
+            removed_chunks = removed,
+            "[composio:github] ingest: re-ingest cleanup"
         );
     }
 
@@ -196,12 +196,12 @@ pub async fn ingest_issue_into_memory_tree(
             // the normal update path. Seeing it here means the prior
             // chunks were already absent (fresh ingest into a primed
             // memory_tree) — fine, just log at debug.
-            log::debug!(
-                "[composio:github] ingest: issue connection_id={} issue_id={} chunks_written={} already_ingested={}",
-                connection_id,
-                issue_id,
+            tracing::debug!(
+                connection_id = %connection_id,
+                issue_id = %issue_id,
                 chunks_written,
                 already_ingested,
+                "[composio:github] ingest: issue persisted"
             );
             Ok(chunks_written)
         }
--- a/src/openhuman/memory_sync/composio/providers/github/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/github/ingest.rs
@@ -206,7 +206,11 @@ pub async fn ingest_issue_into_memory_tree(
             Ok(chunks_written)
         }
         Err(err) => Err(anyhow::anyhow!(
-            "ingest_document failed for {source_id}: {err}"
+            // `{err:#}` (alternate formatter) bakes in the anyhow context
+            // chain so provider.rs's `tracing::warn!(error = %e)` doesn't
+            // strip the underlying cause (DB / embedding / persist failure)
+            // when it Displays the wrapped error.
+            "ingest_document failed for {source_id}: {err:#}"
         )),
     }
 }
--- a/src/openhuman/memory_sync/composio/providers/github/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/github/ingest.rs
@@ -47,7 +47,7 @@ use serde_json::Value;
 
 use crate::openhuman::config::Config;
 use crate::openhuman::memory::ingest_pipeline::{self, IngestResult};
-use crate::openhuman::memory_store::chunks::store::delete_chunks_by_source;
+use crate::openhuman::memory_store::chunks::store::{delete_chunks_by_source, is_source_ingested};
 use crate::openhuman::memory_store::chunks::types::SourceKind;
 use crate::openhuman::memory_sync::canonicalize::document::DocumentInput;
 
@@ -133,16 +133,33 @@ pub async fn ingest_issue_into_memory_tree(
     let source_id = github_source_id(connection_id, issue_id);
 
     // Re-sync of an edited issue: drop prior chunks for the same source_id
-    // before re-ingest. `delete_chunks_by_source` is sync I/O so it has to
-    // hop through `spawn_blocking`.
+    // before re-ingest. Both calls are sync rusqlite I/O so they share one
+    // `spawn_blocking` hop.
+    //
+    // We gate `delete_chunks_by_source` behind `is_source_ingested` — the
+    // delete path uses a `source_kind = ?1` scan with Rust-side
+    // source-id filtering (see `store::delete_chunks_by_source_filter`),
+    // so on a first-time ingest of a never-seen issue it would scan every
+    // Document-kind chunk just to find zero matches. `is_source_ingested`
+    // is an indexed PK lookup against `mem_tree_ingested_sources`, so it
+    // converts the common fresh-issue case to one cheap `COUNT(*)` and
+    // only pays the scan cost on actual re-ingests of edited items.
     let cfg_for_blocking = config.clone();
     let source_for_blocking = source_id.clone();
-    let removed = tokio::task::spawn_blocking(move || {
-        delete_chunks_by_source(
+    let removed = tokio::task::spawn_blocking(move || -> Result<usize> {
+        if is_source_ingested(
             &cfg_for_blocking,
             SourceKind::Document,
             &source_for_blocking,
-        )
+        )? {
+            delete_chunks_by_source(
+                &cfg_for_blocking,
+                SourceKind::Document,
+                &source_for_blocking,
+            )
+        } else {
+            Ok(0)
+        }
     })
     .await
     .map_err(|e| anyhow::anyhow!("delete-prior task join error: {e}"))??;
--- a/src/openhuman/memory_sync/composio/providers/github/ingest.rs
+++ b/src/openhuman/memory_sync/composio/providers/github/ingest.rs
@@ -1,0 +1,457 @@
+//! GitHub → memory tree ingest plumbing.
+//!
+//! Owns the conversion from a single GitHub issue/PR payload (post-extracted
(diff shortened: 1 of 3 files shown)


src/openhuman/memory_sync/composio/providers/github/ingest.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+    })
+    .await
+    .map_err(|e| anyhow::anyhow!("delete-prior task join error: {e}"))??;
+    if removed > 0 {
+        log::debug!(
+            "[composio:github] ingest: re-ingest cleanup connection_id={} issue_id={} removed_chunks={}",
+            connection_id,
+            issue_id,
+            removed
```
graycyrus: **[minor]** `log::debug!` here (and line 180) uses positional format strings instead of the structured field syntax the rest of the codebase uses with `tracing`. Swap to `tracing::debug!` with named fields to keep the log shape consistent with `provider.rs`:

```rust
tracing::debug!(
    connection_id = %connection_id,
    issue_id = %issue_id,
    removed_chunks = removed,
    "[composio:github] ingest: re-ingest cleanup"
);
```

Same for the `chunks_written`/`already_ingested` log at line 180.

graycyrus (commented):
@justinhsu1477 hey! the code looks solid to me, but CI is still pending on a few checks (Windows secrets ACL, E2E macOS/Windows/Linux, Tauri build, Rust Core Coverage, macOS smoke). once those are green I'll come back and approve this.

one thing worth addressing independent of CI:

**`ingest.rs` uses `log::debug!` while the surrounding codebase uses `tracing`**

`provider.rs` logs with `tracing::warn!(issue_id = %issue_id, error = %e, "...")` — structured key=value fields, queryable by any tracing subscriber. `ingest.rs` uses `log::debug!("... connection_id={} issue_id={} removed_chunks={}", ...)` — positional format strings with no structured fields. Both work if the tracing-log bridge is configured, but the structured fields get lost, which defeats the point of having them.

Swapping the two `log::debug!` calls in `ingest_issue_into_memory_tree` to use `tracing::debug!` with the same field syntax as `provider.rs` keeps the pattern consistent.

Also echoing CodeRabbit's cursor-advancement finding — that one's worth fixing before merge. The rest of their nitpick on `{err:#}` is valid too.

Otherwise the migration is clean: `github:{connection_id}:{issue_id}` source_id is stable, the delete-first re-ingest guard mirrors #2887 and #2720 correctly, and the regression test suite pins all the invariants that matter.

justinhsu1477:
Pushed \`9764b31f\` addressing both CodeRabbit findings + responding to @graycyrus's nits.

## What's fixed in this push

| # | Reviewer | Issue | Fix |
| --- | --- | --- | --- |
| B1 | CodeRabbit 🟠 + @graycyrus echo | Failed ingest advances cursor → edited issue can be silently dropped | New \`had_ingest_failures: bool\` flag in \`provider.rs::sync\`; set on every \`Err\` arm; \`advance_cursor\` gated behind \`if !had_ingest_failures\`. Already-synced items skip cheaply via \`is_synced\` on the re-fetch, so the cost of holding is just one extra search-API page on the failure range. \`set_last_sync_at_ms\` still advances — it's a heartbeat, not a fetch-window boundary. |
| B2 | CodeRabbit 🧹 + @graycyrus echo | \`{err}\` drops anyhow chain | \`ingest.rs\` line 209 changed to \`{err:#}\` (alternate formatter). Provider.rs's existing \`tracing::warn!(error = %e)\` will now show the underlying DB / embedding cause instead of just the outer wrap. |

I also carried both fixes back to #2887 (Notion) — same shape there, fixed in \`0c15fd27\` rather than wait for a second round.

## On the \`log::debug!\` vs \`tracing::debug!\` nit — I'd like to push back politely

@graycyrus, before I swap that one I want to double-check the convention. Walking the existing ingest layer:

- \`slack/ingest.rs\` → \`log::debug!\` × 4 (lines 165, 201, 239, 270)
- \`gmail/ingest.rs\` → \`log::debug!\` × 3 (lines 289, 443, 484)
- \`vault/sync.rs\` (the #2720 reference impl) → \`log::debug!\` / \`log::warn!\` × 13

…and \`provider.rs\` files across providers use \`tracing::*\` with structured fields.

So the codebase already has a two-layer split: **provider.rs is the orchestration layer (tracing, structured), ingest.rs / sync.rs are plumbing (log, positional)**. My \`github/ingest.rs\` is intentionally matching the Slack / Gmail / vault shape — swapping to \`tracing\` here would make it the only ingest module using tracing.

Happy to convert if you want me to, but I think the right call is either (a) leave as-is to match the established pattern, or (b) file a follow-up issue to unify all ingest modules onto \`tracing\` in one sweep — both keep the layer convention consistent. Let me know which you prefer.

## Tests

48/48 still pass in \`memory_sync::composio::providers::github::*\`. The cursor invariant is documented in the new rustdoc on \`had_ingest_failures\`; the existing 6 ingest tests cover the ingest function's contract.

graycyrus (commented):
@justinhsu1477 the fixes in the third commit look correct — cursor hold on ingest failure is the right call (holds the full range rather than introducing a separate failure-tracking cursor, which would add state and edge cases), and `{err:#}` properly bakes the anyhow chain into the Display output so `tracing::warn!(error = %e)` actually surfaces the underlying cause. CodeRabbit's two findings are fully addressed.

CI is still pending on several checks (Build Tauri App, Rust Core Coverage, E2E Appium runs, Windows secrets ACL). Once those are green I'll approve this.

One holdover from my first review that the third commit didn't touch: `log::debug!` at ingest.rs lines 171 and 180 still uses the `log` crate with positional format args instead of `tracing::debug!` with structured key=value fields. Everything else in this module and in provider.rs uses tracing. Small thing, but please fix it before I approve — it's easy to miss in metrics/log aggregation pipelines that filter by tracing span fields.

justinhsu1477:
Done — pushed `7d27b5d4`. You're right that for a new module, matching provider.rs's structured logging is the better call; the observability point (span-field filtering in aggregation pipelines) outweighs my "match the other ingest modules" argument, especially since provider.rs sits right next to it in the same directory.

Both `log::debug!` calls in `ingest_issue_into_memory_tree` are now `tracing::debug!` with named fields:

```rust
tracing::debug!(
    connection_id = %connection_id,
    issue_id = %issue_id,
    removed_chunks = removed,
    "[composio:github] ingest: re-ingest cleanup"
);
```
…and the issue-persisted log now carries `chunks_written` / `already_ingested` as fields too.

I also carried the same swap to #2887 (Notion) in `23118615` so the two new memory-tree ingest modules stay consistent with each other.

On the broader codebase: `slack/ingest.rs`, `gmail/ingest.rs`, and `vault/sync.rs` still use `log::*` — I left those alone as pre-existing and out of scope here, but if you'd like the whole ingest layer unified on tracing I'm happy to file a follow-up issue and take it as a separate sweep. LMK.

48/48 tests still green. Should be clear for approval now — thanks for the careful review on both findings.

src/openhuman/memory_sync/composio/providers/github/ingest.rs
graycyrus: Confirmed fixed in 7d27b5d — both debug sites now use `tracing::debug!` with structured key=value fields. Looks good.

graycyrus (commented):
@justinhsu1477 the tracing fix in 7d27b5d addresses the last open thread — both debug sites are now using `tracing::debug!` with structured fields as requested. All prior feedback resolved.

Code is clean. Waiting on CI (Build Tauri App, Rust Core Tests + Coverage, E2E suites, Frontend Coverage are all still pending). Will approve once those are green.

PR #2893, opened 2026-05-29
feat(memory_sources): add Memory Sources domain — user-configurable data connectors

senamakel:
## Summary

- New `memory_sources` domain — a unified registry of data connectors that feed memory: Composio integrations, local folders, GitHub repos (commits/issues/PRs), RSS feeds, web pages, Twitter queries
- Single \"Memory Sources\" UI panel replaces the old Composio-only sync panel; users add/toggle/sync/remove any source kind from one place
- Per-source sync RPC (`memory_sources_sync`) dispatches by kind: Composio → existing engine, others → reader-based ingest into the memory tree
- Per-source status RPC (`memory_sources_status_list`) returns chunks ingested, freshness label, last-sync timestamp; UI polls every 5s
- Composio connections with sync providers (gmail/slack/github/notion/linear/clickup) are **auto-seeded** into the registry at startup and on every list call

## Problem

Memory ingestion was previously a black box of \"sync everything Composio is connected to.\" Users couldn't:
- See **what** was feeding memory (no source registry)
- **Choose** which Composio connections to sync
- Add **non-Composio sources** (a folder of markdown, a GitHub repo's activity, an RSS feed)
- Trigger sync **per source** with live status

The composio sync layer also had no consistent identity at the user-facing layer — \"Gmail\" rolled up multiple accounts and there was no API for the UI to enumerate or manage sources.

## Solution

**Backend: `src/openhuman/memory_sources/`**
- `types.rs` — `SourceKind` enum (composio/folder/github_repo/twitter_query/rss_feed/web_page), `MemorySourceEntry` with kind-specific fields flat on the struct
- `registry.rs` — CRUD over `Config.memory_sources` (TOML-persisted list)
- `readers/{composio,folder,github,rss,twitter,web_page}.rs` — `SourceReader` trait with `list_items` + `read_item` per kind. GitHub reader pulls **project activity** (commits/issues/PRs via `gh` CLI when available, falls back to public REST API), not source code.
- `sync.rs` — `sync_source(id)` dispatches by kind. Composio → `memory_sync::composio::run_connection_sync`. Reader-backed kinds → walk items and ingest each via `ingest_document` (chunks tagged `mem_src:{source_id}:{item_id}`). Emits `MemorySyncStageChanged` events keyed by source id.
- `status.rs` — Per-source SQL query against `mem_tree_chunks` for chunk counts + freshness; toolkit-prefix matching for composio, `mem_src:{id}:%` for everything else.
- `reconcile.rs` — Auto-seeds the registry with active Composio connections that have a sync provider; called at startup and lazily on every `memory_sources_list` RPC.

**Frontend:**
- `memorySourcesService.ts` — typed client for all 9 RPCs (list/get/add/update/remove/list_items/read_item/sync/status_list)
- `AddMemorySourceDialog.tsx` — two-step picker (kind → form). Folder kind has a native folder picker via `<input type=\"file\" webkitdirectory>`. Composio kind fetches active connections via `composio_list_connections` and presents them as a dropdown.
- `MemorySourcesRegistry.tsx` — single unified panel: Sync button on every row, polls status every 5s, shows chunk count / freshness pill / last-sync timestamp per row
- Removed: the old `MemorySources.tsx` (composio-only panel), `MemorySyncConnections.tsx`, `memorySyncService.ts` — all consolidated under the new domain

## Submission Checklist

- [x] N/A: Heavy net-new code lands as Rust unit tests (27 passing) + JSON-RPC E2E tests (4 passing). UI E2E coverage for the new panel is a follow-up. **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via \`diff-cover\`) meet the gate enforced by [\`.github/workflows/coverage.yml\`](../.github/workflows/coverage.yml). Run \`pnpm test:coverage\` and \`pnpm test:rust\` locally; PRs below 80% on changed lines will not merge.
- [x] N/A: behaviour-only change for coverage matrix — new RPCs are catalog additions, not new product features. Coverage matrix updated — added/removed/renamed feature rows in [\`docs/TEST-COVERAGE-MATRIX.md\`](../docs/TEST-COVERAGE-MATRIX.md) reflect this change (or \`N/A: behaviour-only change\`)
- [x] N/A: no matrix updates so no IDs to list. All affected feature IDs from the matrix are listed in the PR description under \`## Related\`
- [x] N/A: no release-cut surfaces touched. Manual smoke checklist updated if this touches release-cut surfaces ([\`docs/RELEASE-MANUAL-SMOKE.md\`](../docs/RELEASE-MANUAL-SMOKE.md))
- [x] N/A: no linked issue. Linked issue closed via \`Closes #NNN\` in the \`## Related\` section

## Impact

- **Runtime**: desktop only. New TOML field \`[[memory_sources]]\` on \`Config\`; \`#[serde(default)]\` so existing configs load unchanged.
- **Migration**: none required. On first launch with the new build, Composio connections are auto-upserted into the registry.
- **Performance**: status polling is one SQL roundtrip per source every 5s — cheap. Sync runs in a spawned task so the RPC returns immediately.
- **Security**: folder reader has path-traversal prevention via canonicalize-and-prefix-check; web page reader strips HTML; no new external network deps.

## Related

- Closes:
- Follow-up PR(s)/TODOs: UI E2E coverage for the new panel; Twitter API credential wiring; SSE-based push for sync progress (currently 5s poll).

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: \`feat/memory-sources\`
- Commit SHA: \`92cf2a7f36e8f538fac00fe486d1cd689cd3b6f2\`

### Validation Run
- [x] \`pnpm --filter openhuman-app format:check\`
- [x] \`pnpm typecheck\`
- [x] Focused tests: \`cargo test --lib -- memory_sources\` (27 passing), \`pnpm vitest run src/services/memorySourcesService.test.ts\` (6 passing), \`cargo test --test memory_sources_e2e\` (4 passing)
- [x] Rust fmt/check (if changed): \`cargo fmt --check\` + \`cargo check\` clean
- [x] N/A — only \`app/src/\` and Rust core touched. Tauri fmt/check (if changed):

### Validation Blocked
- \`command:\` N/A
- \`error:\` N/A
- \`impact:\` N/A

### Behavior Changes
- Intended behavior change: introduce per-source memory sources registry + unified Sync UI; auto-seed Composio connections so users see their integrations as memory sources by default
- User-visible effect: new \"Memory Sources\" panel in the Memory tab; \"Add Source\" dialog for non-Composio kinds; per-row Sync button and live chunk/freshness display

### Parity Contract
- Legacy behavior preserved: the existing \`memory_sync_status_list\` per-toolkit RPC is removed at the UI layer but the underlying \`memory_sync::composio::run_connection_sync\` engine and its events are unchanged; existing Composio sync providers continue to drive ingestion identically.
- Guard/fallback/dispatch parity checks: \`memory_sync::composio::list_sync_targets\` now reads from the registry first and falls back to scanning all connections when the registry is empty — preserves the pre-registry path for first-launch / not-yet-reconciled state.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Memory Sources Registry: unified UI to add, list, enable/disable, remove, and sync memory sources; two‑step add dialog for choosing kind and configuring.
  * Adds support for Composio, local folders, GitHub repos, RSS feeds, web pages, and Twitter queries; Composio connections are auto-registered and can be synced.

* **Tests**
  * New unit and end-to-end tests covering memory sources flows and service RPCs.

* **Documentation**
  * New i18n strings for memory-sources across many languages.


src/openhuman/memory_sources/readers/folder.rs
```diff
@@ -0,0 +1,223 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use tempfile::TempDir;
+
+    fn folder_source(path: &str) -> MemorySourceEntry {
+        MemorySourceEntry {
+            id: "src_folder".into(),
```
graycyrus, line 151: **[major]** TOCTOU: traversal guard uses `canonical_file` but the read here uses the original non-canonical `file_path`. A symlink in the source directory can be swapped between `canonicalize` and this call, redirecting the read outside the base.

Fix:
```rust
let body = tokio::fs::read_to_string(&canonical_file)
    .await
    .map_err(|e| format!("failed to read {}: {e}", canonical_file.display()))?;
```

src/openhuman/memory_sources/status.rs
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct SourceStatus {
+    pub source_id: String,
+    pub chunks_synced: u64,
+    pub chunks_pending: u64,
```
graycyrus, line 46: **[major]** `SourceStatus` has no field for sync errors. After a failed sync the UI shows `Idle` / unchanged chunk count — identical to "never synced". The `MemorySyncStageChanged(Failed)` event is transient; it's lost before the next 5s poll.

Add `last_sync_error: Option<String>` here and wire it from the spawned task in `sync.rs`. An in-memory `DashMap<source_id, last_error>` populated by `sync_source` and read by `source_status` would be the lightest-weight fix without a schema change.

src/openhuman/memory_sources/readers/github.rs
```diff
@@ -0,0 +1,697 @@
...
+
+fn gh_available() -> bool {
+    std::process::Command::new("gh")
+        .arg("--version")
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .status()
+        .map(|s| s.success())
+        .unwrap_or(false)
+}
```
graycyrus, line 54: **[minor]** `std::process::Command::status()` is synchronous and blocks the tokio worker thread on every `list_items`/`read_item` call. Use `tokio::process::Command` or wrap in `spawn_blocking`:

```rust
async fn gh_available() -> bool {
    tokio::process::Command::new("gh")
        .arg("--version")
        .stdout(std::process::Stdio::null())
        .stderr(std::process::Stdio::null())
        .status()
        .await
        .map(|s| s.success())
        .unwrap_or(false)
}
```

src/openhuman/memory_sync/composio/bus.rs
```diff
@@ -576,6 +576,24 @@ impl EventHandler for ComposioConnectionCreatedSubscriber {
...
             }
+
+            // Auto-register this connection in the memory_sources
+            // registry so it appears in the unified sources list.
+            let label = format!("{toolkit} connection");
+            if let Err(e) = crate::openhuman::memory_sources::upsert_composio_source(
+                &toolkit,
+                &connection_id,
+                &label,
+            )
```
graycyrus, line 603: **[minor]** Label format here is `"{toolkit} connection"` (lowercase, no short ID). `reconcile.rs` uses `"{Title_case(toolkit)} · {short_id(connection_id)}"`.

Connections created before the new build and connections created after it get different label formats. Use the reconcile format (`title_case + short_id`) here too for consistency.

graycyrus (changes requested):
Good chunk of work — the architecture is solid and the reader abstraction is clean. CodeRabbit covered the low-hanging stuff (timeouts, i18n gaps, SSRF scheme check, SQL error swallowing). Two things I want fixed before merge that weren't covered.

## Blocking

### TOCTOU in folder reader

`read_item` resolves a canonical path for the traversal check, then reads from the original `file_path`. Between `canonicalize` and `read_to_string`, a symlink in the source directory can be swapped to point outside the base, bypassing the guard. One-liner fix: use `canonical_file` for the read.

```rust
// current
let body = tokio::fs::read_to_string(&file_path).await ...

// fix
let body = tokio::fs::read_to_string(&canonical_file).await ...
```

The PR description explicitly calls out the traversal prevention as a security property — the implementation doesn't fully deliver on that claim.

### No persistent sync error state

`SourceStatus` has `chunks_synced`, `chunks_pending`, `last_chunk_at_ms`, `freshness` — but no `last_error` or `sync_failed` field. After a sync failure, the UI shows `Idle` with unchanged chunk counts, which is indistinguishable from "never synced". The `MemorySyncStageChanged(Failed)` event is the only signal, and it's lost on reconnect or the next 5s poll cycle.

Users who add a GitHub source with bad credentials, or an RSS feed that's down, get no persistent feedback about why nothing is updating. The Sync button just goes from spinner back to idle. This is a core promise of the feature — "live chunk count / freshness display per row" — and it silently breaks under error conditions.

Add a `last_sync_error: Option<String>` to `SourceStatus` (or a `sync_state` enum). Persist it somewhere the status query can return it: a lightweight in-memory map keyed by `source_id` updated by the spawned sync task would do, or write it to the DB. Without this, the feature ships broken for any source that can fail.

## Non-blocking

### `gh_available()` blocks the async executor

`github.rs` calls `std::process::Command::new("gh").arg("--version").status()` synchronously from an async context. It blocks the current tokio worker thread for the subprocess lifetime on every `list_items`/`read_item` call. Wrap in `tokio::task::spawn_blocking` or switch to `tokio::process::Command`.

### Label format inconsistency on auto-register

When a Composio connection is first created (bus event in `bus.rs`), it's registered with `"{toolkit} connection"` — lowercase, no short ID. The reconcile path in `reconcile.rs` generates `"{Title_case(toolkit)} · {short_id(connection_id)}"`. Users who connect a new integration post-launch get a different label format than existing integrations that went through reconcile. Pick one format and use it in both places.

graycyrus (commented):
## CodeRabbit-style Review — PR #2893 (Memory Sources)

**Overall:** Well-structured domain that correctly follows project conventions. 2 blockers, 3 major issues, 2 refactor suggestions, 6 nitpicks.

### Verified / Looks Good
- Path traversal guard in `folder.rs` (canonicalize + starts_with) is correct in principle
- GitHub URL parser correctly rejects deep links, bare org paths, non-GitHub hostnames
- `MemorySourceEntry.validate()` enforces per-kind required fields at add/update
- SSRF guard in `web_page.rs` rejects `file://`, `data://` and non-HTTP schemes
- Freshness label thresholds tested with fixed timestamps — no time flakes
- `upsert_composio_source` matches by `connection_id`, not id/toolkit — correct
- All 13 locale chunk files include the 56 new keys
- `sync_via_reader` progress fires every 5 items — good UI feedback without flooding
- `cargo check` and `cargo fmt --check` pass clean
- Typecheck failures (`recharts`, `rehype-katex`) are pre-existing on main

### Questions
1. Was the removal of `VaultPanel` and `ObsidianVaultSection` from `MemoryWorkspace.tsx` intentional? (See blocker comment below)
2. `github.rs` `list_items` fetches commits, issues, PRs in 3 serial API calls — was `tokio::join!` considered for concurrency?

src/openhuman/memory_sources/readers/folder.rs
graycyrus, line 1: ## 🚫 Blocker: TOCTOU race — read uses original path, not canonicalized

The traversal guard canonicalizes `file_path` at line ~103, but `read_to_string` at line ~121 uses the **original** `file_path`. Between the guard passing and the read, a symlink swap could redirect the read outside the canonical base.

**Fix:** read from `canonical_file`, not `file_path`:

```rust
// before
let body = tokio::fs::read_to_string(&file_path)
    .await
    .map_err(|e| format!("failed to read {}: {e}", file_path.display()))?;

// after — use the canonicalized path
let body = tokio::fs::read_to_string(&canonical_file)
    .await
    .map_err(|e| format!("failed to read {}: {e}", canonical_file.display()))?;
```

app/src/components/intelligence/MemoryWorkspace.tsx
graycyrus, line 1: ## 🚫 Blocker: `VaultPanel` and `ObsidianVaultSection` silently removed

This PR removes `<VaultPanel onToast={onToast} />` and `<ObsidianVaultSection contentRootAbs={...} />` from `MemoryWorkspace`. These are the Obsidian vault viewer and "View vault in Obsidian" link — they're unrelated to memory sources and were user-visible features.

The PR description makes no mention of intentionally removing them. If accidental, please restore:

```tsx
<VaultPanel onToast={onToast} />
{graph && (
  <ObsidianVaultSection contentRootAbs={graph.content_root_abs} onToast={onToast} />
)}
```

If intentional, it should be a separate commit with justification.

src/openhuman/memory_sources/readers/rss.rs
graycyrus, line 1: ## ⚠️ Major: RSS `fetch_url` has no body size cap — potential OOM

The web page reader correctly caps at 10 MiB (`MAX_BODY_BYTES`), but `fetch_url` here does `resp.text().await` with no size guard. A hostile or misconfigured feed could buffer hundreds of MBs into memory.

**Fix:**
```rust
const MAX_FEED_BYTES: u64 = 5 * 1024 * 1024; // 5 MiB
let bytes = resp
    .bytes()
    .await
    .map_err(|e| format!("failed to read feed body: {e}"))?;
if bytes.len() as u64 > MAX_FEED_BYTES {
    return Err(format!(
        "feed body exceeds {MAX_FEED_BYTES}-byte limit ({} bytes)",
        bytes.len()
    ));
}
let text = String::from_utf8_lossy(&bytes).into_owned();
```

src/openhuman/memory_sources/readers/github.rs
graycyrus, line 1: ## ⚠️ Major: `gh` CLI invoked without timeout — can block sync task indefinitely

`gh_json` calls `tokio::process::Command::new("gh").output().await` with no timeout. If `gh` hangs (auth flow, network stall, rate-limit backoff), the sync task parks forever and `MemorySyncStage::Completed` never fires.

**Fix:**
```rust
async fn gh_json(args: &[&str]) -> Result<String, String> {
    let output = tokio::time::timeout(
        std::time::Duration::from_secs(30),
        tokio::process::Command::new("gh").args(args).output(),
    )
    .await
    .map_err(|_| "gh command timed out after 30s".to_string())?
    .map_err(|e| format!("gh command failed: {e}"))?;
    // ...
}
```

src/openhuman/memory_sources/sync.rs
graycyrus, line 1: ## ⚠️ Major: `tokio::spawn` result dropped — panics in sync tasks are invisible

The spawned background task's `JoinHandle` is dropped immediately. Any panic inside it produces no log, and at shutdown the runtime may cancel an in-flight ingest mid-way.

**Fix:** Track the handle or attach a panic handler:
```rust
let source_id_for_log = source_id.clone();
let handle = tokio::spawn(async move {
    // ... sync body ...
});
tokio::spawn(async move {
    if let Err(panic) = handle.await {
        tracing::error!(
            source_id = %source_id_for_log,
            "[memory_sources:sync] sync task panicked: {:?}", panic
        );
    }
});
```

src/openhuman/memory_sources/status.rs
graycyrus, line 1: ## 💡 Suggestion: Composio chunk count is per-toolkit, not per-connection

For Composio sources, `source_id_prefix` returns `"{toolkit}:%"` (e.g. `gmail:%`). If a user has two Gmail connections, both registry entries report the **same** aggregate chunk count — misleading in the UI.

Long-term fix: include `connection_id` in Composio chunk source IDs. Short-term: add a comment documenting the known limitation so future readers aren't confused.

src/openhuman/memory_sources/rpc.rs
graycyrus, line 1: ## 🔧 Nitpicks (low priority)

1. **`default_true()` duplicate** — defined again here; already exists in `types.rs`. Deduplicate to a shared location.
2. **`folder.rs`** — `base_path.trim_end_matches('/')` won't fire on Windows-style `\` paths. Low risk (macOS/Linux primary).
3. **`rss.rs`** — RSS vs Atom detection by substring `<feed` is fragile if the string appears in description before the root element. A `starts_with` check on first non-whitespace tag would be more robust.
4. **`AddMemorySourceDialog.tsx`** — Back button uses raw `←` character instead of SVG icon. Minor a11y/RTL concern.
5. **`MemorySourcesRegistry.tsx`** — `SOURCE_KIND_ICONS[source.kind] ?? '📄'` fallback is dead code (type is `Record<SourceKind, string>`).
6. **`reconcile.rs`** — `title_case` produces `"Clickup"` not `"ClickUp"`. Cosmetic but may look odd to users.

graycyrus:
Great progress — the TOCTOU fix in folder.rs and label formatting are solid. Three things still need addressing before I can sign off:

1. **gh_available() blocking sync call** (github.rs) — still using `std::process::Command` from async context. This will block the tokio thread. Switch to `tokio::process::Command` or `spawn_blocking`.

2. **No persistent sync error state** (status.rs) — when a sync fails, the UI still shows `Idle` with no way to distinguish "never synced" vs "sync failed". The `MemorySyncStageChanged(Failed)` event is transient. Add `last_sync_error: Option<String>` to `SourceStatus` or track it per-source in the spawned sync task so users see what went wrong.

3. **CI is failing** — PR Submission Checklist and Rust Core Coverage jobs are red. Fix those before I can approve.

Let me know once those are squared away.

graycyrus (commented):
Progress update — you fixed the TOCTOU in folder.rs and the label format inconsistency, thanks. However, two blocking issues from the prior review remain unresolved:

1. **No persistent sync error state** — `SourceStatus` struct is still unchanged. After a failed sync, the UI shows `Idle` with unchanged chunk counts, indistinguishable from "never synced". Users get no feedback about why nothing is updating (e.g., bad credentials, unreachable feed). This is a core UX promise of the feature — add `last_sync_error: Option<String>` to `SourceStatus` or implement per-source error tracking. Required before merge.

2. **`gh_available()` still blocks the async executor** — still using `std::process::Command::status()` synchronously from async context, blocking a tokio thread per call. Wrap in `tokio::task::spawn_blocking` or switch to `tokio::process::Command`. Non-blocking severity but should be fixed before merge.

Also, CI is failing — PR Submission Checklist and Rust Core Coverage jobs are red. Fix those first, then the above, and I'll review again.

graycyrus:
Hey @senamakel — just checking in. I see CI is still settling (Rust Core Coverage + Windows/macOS E2E tests are pending), and the two blockers from my prior review haven't been addressed yet:

1. `src/openhuman/memory_sources/readers/github.rs` — `gh_available()` still uses synchronous `std::process::Command` from async context, which blocks the tokio executor. Switch to `tokio::process::Command` or `spawn_blocking`.
2. `src/openhuman/memory_sources/status.rs` — `SourceStatus` still has no `last_sync_error` field. After a failed sync, the UI can't distinguish "never synced" from "sync failed". Add error state tracking.

I'm holding my approval until:
- CI is 100% green (all checks pass, not pending)
- Both of these issues are fixed

Once you push those fixes, just let me know and I'll do a final review pass. Thanks!

PR #2896, opened 2026-05-29
fix(socket): refresh session token on Invalid token rejection (TAURI-RUST-9C)

### LINKED ISSUE #2892: Socket reconnection fails with expired token after disconnect (237 events)
## Summary

Socket.IO reconnection uses an expired/invalid token after disconnect, causing sustained outages. 237 events across multiple releases (0.54.0 → 0.56.0).

## Problem

**What happened:** After a socket disconnect, the reconnection logic retries with the same token. If the token has been rotated or invalidated server-side, all 5 retry attempts fail with "Invalid token", and the connection is declared a sustained outage:
```
[socket] Connection failed (sustained outage after 5 attempts): SIO CONNECT: Socket.IO connect error: Invalid token
```

**Expected:** The socket reconnection path should refresh the auth token before retrying, or fetch a fresh token after the first "Invalid token" failure instead of retrying with the same stale token 5 times.

**Impact:** 237 events in Sentry (TAURI-RUST-9C), **ongoing** across releases 0.54.0 → 0.56.0. Users experience sustained socket outages that don't self-heal until app restart.

**Steps to reproduce:**
1. Connect to the socket server with a valid token
2. Wait for the token to be rotated/invalidated server-side (or disconnect for a period)
3. Socket disconnects and tries to reconnect
4. All 5 reconnect attempts fail with "Invalid token"
5. Connection enters "sustained outage" state

**Version / Platform:** openhuman@0.54.0 through 0.56.0, multiple platforms

## Solution (optional)

Check the socket reconnection logic in `src/openhuman/socket/`:
1. Find where the token is set for reconnection attempts
2. Add token refresh before retry — either re-fetch from the auth/session layer or request a new token from the backend
3. On "Invalid token" error specifically, force a token refresh before the next retry instead of retrying with the same token
4. Consider: is the token stored in the socket config at connect time and never updated?

## Acceptance criteria

- [ ] **Repro gone** — Socket reconnection refreshes token before retrying
- [ ] **Regression safety** — Unit test for token refresh on reconnect
- [ ] **No sustained outages** from stale tokens after this fix

## Related

- Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/400/

graycyrus:
## Summary

Fixes Sentry issue [TAURI-RUST-9C](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/400/) — 237 events across 0.54.0 → 0.56.0.

**Root cause:** `ws_loop` captured the session token at `SocketManager::connect` time. On reconnect after a disconnect/token rotation, the server returned `44{"message":"Invalid token"}`. The loop treated this as a regular `ConnectionOutcome::Failed` and retried with the **same stale credential** 5 times before declaring "sustained outage" and firing a Sentry event.

**Fix:**
- Add `ConnectionOutcome::InvalidToken` to `types.rs` — distinct from transport failures.
- In `run_connection`, detect `"Invalid token"` in the SIO connect-error message via `is_invalid_token_error()` (case-insensitive) and return `InvalidToken` instead of `Failed`.
- In `ws_loop`, on `InvalidToken`: call `refresh_session_token()` which loads config + reads the current stored JWT (same path as `handle_connect_with_session`). On success (a *different* token), reset `consecutive_failures`/`backoff` and `continue` immediately with no backoff sleep. On failure (user logged out / store unavailable / same stale token returned), fall back to `Failed` accounting so the Sentry escalation still fires.
- Only reset `consecutive_failures` counter and `backoff` after a genuinely different token is returned — prevents counter suppression if the store keeps returning the same stale JWT.

**Tests added:**
- `is_invalid_token_error_matches_backend_message` — exact/case-variant wire shapes
- `is_invalid_token_error_does_not_match_other_errors` — no false positives
- `run_connection_returns_invalid_token_on_auth_reject` — end-to-end through a mock EIO server sending `44{"message":"Invalid token"}`
- `connection_outcome_variants_can_be_constructed` — updated to cover `InvalidToken`

## Test plan

- [x] Unit tests pass: `pnpm debug rust` (or `cargo test -p openhuman socket`)
- [x] (N/A) Manual: rotate/expire a session token while the socket is connected → verify reconnect succeeds with refreshed token and no "sustained outage" Sentry event fires — covered by `run_connection_returns_invalid_token_on_auth_reject` e2e mock test; live rotation requires a production session
- [x] Verify `consecutive_failures` resets to 0 on `InvalidToken` (no false outage Sentry event after a single auth rejection)

Closes #2892


oxoxDev:
Thanks for this, @graycyrus — and credit where due: this PR was filed first (this one at 04:24, #2905 at 05:47). Both target the same crash, **TAURI-RUST-9C / #2892** (the stale-token "Invalid token" reconnect storm), with the same set of files.

We're going to consolidate on **#2905** and close this one. To be transparent about *why* — it's not a comment on effort, the two diffs are genuinely close — but #2905 edges ahead on a few correctness points:

- **Token-error detection** — #2905 anchors on both `"socket.io connect error"` *and* `"invalid token"`; this PR matches a bare `contains("invalid token")`, which can fire on unrelated upstream errors that happen to contain that phrase.
- **Bounded retry** — on a refreshed-but-different token this PR resets `consecutive_failures = 0` and `continue`s with no backoff; if the token store keeps returning new tokens (rotation / non-deterministic source) that's an unbounded no-sleep retry loop. #2905 caps the immediate-retry to one attempt before falling through to backoff + escalation.
- **Genuine-expiry UX** — #2905 fast-fails a definitively-expired session, surfaces `"session expired — please sign in again"` to the UI, and emits no Sentry event; this PR runs to the 5-retry sustained-outage path, which still fires the very Sentry event we're trying to silence.
- This diff also drops the unrelated `sustained_outage_for_tls_handshake_eof_classifies_as_expected` regression test (TAURI-RUST-4ZD) — looks like a stale base.

Your `is_invalid_token_error` case-insensitivity rationale and the `run_connection` → `InvalidToken` e2e mock test are both nicely done — if any of that is missing from #2905 it's worth porting over. Closing this in favor of #2905; thanks again for the fast turnaround on the crash.

PR #2905, opened 2026-05-29
fix(socket): refresh token before reconnect, fast-fail on Invalid token (#2892)

### LINKED ISSUE #2892: Socket reconnection fails with expired token after disconnect (237 events)

YOMXXX:
## Summary

Fixes #2892 — socket reconnection retried 5× with the same stale token, leaving users stuck in a "sustained outage" state until app restart (237 Sentry events on TAURI-RUST-9C across 0.54.0 → 0.56.0).

- New `src/openhuman/socket/token_provider.rs` introduces a `TokenProvider` (`Arc<dyn Fn() -> Result<String, String>>`) so the reconnect loop can pull a fresh token on every attempt instead of capturing the token-at-spawn-time once.
- `handle_connect_with_session` now passes `token_provider_from_config(config)` which re-reads the session token from the profile store on every reconnect.
- On the specific `"Socket.IO connect error: Invalid token"` wire failure, the loop:
  1. Re-fetches the token.
  2. If a **fresh** token is available → retries once with it (likely a normal token rotation).
  3. If the token is **identical** OR the provider errors → fast-fails after **≤2 attempts** (not 5), sets `SharedState.error = "session expired — please sign in again"`, and exits cleanly.

## Why fast-fail (instead of more retries)?

The 5-retry budget exists for transient network failures, not authentication problems. Looping with a token the server has already rejected just wastes time and produces a "sustained outage" Sentry event that masks real issues. The frontend can prompt the user to sign in again far sooner.

## Public API

- `SocketManager::connect(url, token)` — unchanged signature (backward compat); now internally wraps the token in a `static_token_provider`.
- `SocketManager::connect_with_provider(url, provider)` — new, used by the RPC `connect_with_session` handler.

## Files

| File | Change |
|---|---|
| `src/openhuman/socket/token_provider.rs` | **New**: `TokenProvider` alias, `static_token_provider`, `token_provider_from_config`, `is_invalid_token_error` (double-anchor predicate + tests) |
| `src/openhuman/socket/ws_loop.rs` | Loop re-fetches token via provider each attempt; `decide_after_invalid_token` pure function for the fast-fail decision (with 4 unit tests) |
| `src/openhuman/socket/manager.rs` | New `connect_with_provider`; deduped via private `spawn_loop` helper; `connect` unchanged externally |
| `src/openhuman/socket/schemas.rs` | `handle_connect_with_session` now builds a refreshing provider |
| `src/openhuman/socket/mod.rs` | `pub(crate) mod token_provider;` |
| `src/openhuman/socket/ws_loop_tests.rs` | 6 new tests + existing migration |

## Sentry / classifier note

`"Socket.IO connect error: Invalid token"` is **not** caught by the existing `is_session_expired_message` classifier (it requires the `OpenHuman API error (401` prefix shape, not bare SIO prose). That's why ~237 events exist — they passed through to Sentry. After this fix, at most **1 event per session-expiry incident** is produced (one fast-fail log line, instead of 5 retry escalations), which on its own materially reduces noise.

## Test plan

- [x] `GGML_NATIVE=OFF cargo test -p openhuman socket --lib` — **121 passing** (117 pre-existing + 4 new `decide_after_invalid_token` cases)
- [x] Existing reconnect / state-transition tests still pass after the provider plumbing

## Scope notes

- **No protocol changes** — Engine.IO / Socket.IO wire format untouched.
- **No frontend (`app/`) changes** — Rust core fix is sufficient.
- **No new deps.**
- **No new state** — `token_provider_from_config` is a stateless closure capturing `Arc<Config>`.

## Summary by CodeRabbit

* **New Features**
  * WebSocket connections now accept a token provider so reconnects can obtain refreshed session tokens and validate initial token state; initial-session validation added.

* **Bug Fixes**
  * Improved handling and escalation for invalid, empty, or error-returning tokens to avoid futile reconnect loops and surface a clear "session expired" state; single-attempt fresh-token retry on invalid-token responses.

* **Tests**
  * Added unit and integration tests covering provider behavior, invalid-token classification, retry/escalation, and regression guards against hot-looping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2905?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/socket/manager.rs
+++ b/src/openhuman/socket/manager.rs
@@ -140,41 +140,13 @@ impl SocketManager {
             log::error!("[socket] connect: refusing to start — empty session token");
             return Err("empty session token — authenticate first".to_string());
         }
-
-        // Ensure the rustls crypto provider is installed (needed for wss:// TLS).
-        // This is a no-op if already installed.
-        let _ = rustls::crypto::ring::default_provider().install_default();
-
-        self.disconnect().await?;
-
-        log::info!("[socket] Connecting to {}", url);
-
-        *self.shared.status.write() = ConnectionStatus::Connecting;
-        *self.shared.error.write() = None;
-        emit_state_change(&self.shared);
-
-        let (emit_tx, emit_rx) = mpsc::unbounded_channel::<String>();
-        let internal_tx = emit_tx.clone();
-        let (shutdown_tx, shutdown_rx) = watch::channel(false);
-
-        *self.emit_tx.lock().await = Some(emit_tx);
-        *self.shutdown_tx.lock().await = Some(shutdown_tx);
-
-        let url = url.to_string();
         // Wrap the static token in a provider closure. Existing callers that
         // pass a concrete token value continue to work unchanged; the provider
         // returns that same token on every call (static semantics). For
         // live-session refresh, callers should use `connect_with_session` which
         // builds a provider via `token_provider_from_config`.
         let provider = static_token_provider(token.to_string());
-        let shared = Arc::clone(&self.shared);
-
-        let handle = tokio::spawn(async move {
-            ws_loop(url, provider, shared, emit_rx, shutdown_rx, internal_tx).await;
-        });
-
-        *self.loop_handle.lock().await = Some(handle);
-        Ok(())
+        self.spawn_loop(url, provider).await
     }
 
     /// Connect using a **live-refresh token provider**.
@@ -211,13 +183,23 @@ impl SocketManager {
                 return Err(e);
             }
         }
+        self.spawn_loop(url, token_provider).await
+    }
 
+    /// Shared spawn path used by both [`connect`] and [`connect_with_provider`].
+    ///
+    /// Installs the rustls crypto provider, tears down any existing connection,
+    /// constructs the channel pair, and spawns the background `ws_loop` task.
+    /// Entry-point-specific validation (empty-token guard, provider pre-check)
+    /// is done by the callers before this is called.
+    async fn spawn_loop(&self, url: &str, provider: TokenProvider) -> Result<(), String> {
         // Ensure the rustls crypto provider is installed (needed for wss:// TLS).
+        // This is a no-op if already installed.
         let _ = rustls::crypto::ring::default_provider().install_default();
 
         self.disconnect().await?;
 
-        log::info!("[socket] Connecting (live-token provider) to {}", url);
+        log::info!("[socket] Connecting to {}", url);
 
         *self.shared.status.write() = ConnectionStatus::Connecting;
         *self.shared.error.write() = None;
@@ -234,7 +216,7 @@ impl SocketManager {
         let shared = Arc::clone(&self.shared);
 
         let handle = tokio::spawn(async move {
-            ws_loop(url, token_provider, shared, emit_rx, shutdown_rx, internal_tx).await;
+            ws_loop(url, provider, shared, emit_rx, shutdown_rx, internal_tx).await;
         });
 
         *self.loop_handle.lock().await = Some(handle);
--- a/src/openhuman/socket/manager.rs
+++ b/src/openhuman/socket/manager.rs
@@ -19,6 +19,7 @@ use tokio::time::Duration;
 use crate::api::models::socket::{ConnectionStatus, SocketState};
 use crate::openhuman::webhooks::WebhookRouter;
 
+use super::token_provider::{static_token_provider, TokenProvider};
 use super::ws_loop::ws_loop;
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


graycyrus (commented):
@YOMXXX hey! the code looks really solid here — CI is still pending though, so i'll hold off on the formal approval until those checks pass. once they're green i'll come back and approve this.

---

**Walkthrough**

This is a clean, well-scoped fix for TAURI-RUST-9C (#2892). The root cause — capturing the token once at spawn time and retrying with it 5× — is correctly identified and fixed by introducing a `TokenProvider` callable that the reconnect loop re-evaluates before each attempt.

| File | Change |
|---|---|
| `token_provider.rs` | New module: `TokenProvider` alias, `static_token_provider`, `token_provider_from_config`, `is_invalid_token_error` + 7 unit tests |
| `ws_loop.rs` | Calls provider before each attempt; `decide_after_invalid_token` pure decision function; fast-fail on same/expired token |
| `manager.rs` | New `connect_with_provider`; `spawn_loop` private helper deduplication; `connect()` externally unchanged |
| `schemas.rs` | `handle_connect_with_session` now builds a live-refresh provider via `token_provider_from_config` |
| `ws_loop_tests.rs` | 6 new integration tests including mock invalid-token server; existing tests migrated to provider API |

**Architecture**

The `TokenProvider = Arc<dyn Fn() -> Result<String, String> + Send + Sync>` alias is the right shape — synchronous (profile store read is a lock + disk, not async I/O), cheaply cloned across the task boundary, and mockable in tests without any trait-object ceremony.

`is_invalid_token_error` using a double-anchor (`socket.io connect error` + `invalid token`) is the correct approach — avoids false positives from unrelated upstream 401 messages that happen to contain "invalid token".

`decide_after_invalid_token` being a pure function with 4 dedicated unit tests is good practice — the retry/escalate logic is testable without touching any socket machinery.

**Issue alignment (closes #2892)**

All acceptance criteria are met:
- Token refresh on reconnect: provider re-reads from profile store on every attempt
- Unit tests for token refresh: `decide_after_invalid_token_*` suite + `ws_loop_escalates_immediately_on_invalid_token_no_refresh` integration test
- No sustained outages from stale tokens: fast-fail exits after ≤2 attempts instead of 5
- Coverage gate: 121 tests passing (117 pre-existing + 4 new unit)

**One minor observation**

In `ws_loop.rs`, when the provider returns `Ok(_)` with an empty token, it logs at `log::error!`, but when the provider returns `Err(e)` (which is a more severe "no session stored" condition), it logs at `log::warn!`. These severity levels are inverted relative to what each case represents — a provider error is at least as bad as an empty token. Not a blocker, just worth aligning on the next pass.

Otherwise the diff is clean — no `.unwrap()`, no PII in logs (token length only), `connect()` API is backward-compatible, no frontend changes needed. Good work on the test coverage here.

oxoxDev (changes requested):
Strong fix overall — the `TokenProvider` abstraction is clean, the pure `decide_after_invalid_token` helper is easy to reason about, and the test coverage (fast-fail, fresh-token, provider-error, ≤2-attempt regression guard) is excellent. One real concern blocks approval.

### Major — unbounded immediate-retry in the fresh-token branch (`ws_loop.rs`, `RetryImmediately`)
On `RetryImmediately` the loop `continue`s **without** incrementing `consecutive_failures` and **without** any backoff. It only terminates because the production provider (`token_provider_from_config`) returns the *same* token on the next call, which then hits the `Escalate` path.

But the safety of the loop depends entirely on that "same token" assumption. If a provider ever returns a *different* non-empty token on each call — rapid server-side rotation, a non-deterministic/buggy token source, or a future provider implementation — the loop becomes a **no-backoff hot loop**: connect → Invalid token → fresh token differs → retry immediately → connect → Invalid token → … It never sleeps and never escalates, hammering the server. That's arguably worse than the original 5-retry storm this PR is fixing.

Please bound the immediate-retry path so a single fresh-token attempt is allowed, after which it falls through to the normal backoff + escalation budget:
```rust
InvalidTokenAction::RetryImmediately { fresh_len } => {
    fresh_token_retries += 1;
    if fresh_token_retries > 1 {
        // already gave the fresh token one immediate shot — don't hot-loop;
        // fall through to normal backoff/escalation.
        consecutive_failures = consecutive_failures.saturating_add(1);
        log_connection_failure(consecutive_failures, &reason);
        // (then the normal backoff sleep below)
    } else {
        continue;
    }
}
```
Reset `fresh_token_retries = 0` on any successful connection.

### Minor — double provider call per fresh cycle
`decide_after_invalid_token` fetches the fresh token, then `continue` re-fetches it at the top of the loop — the validated token is discarded and the `fresh_len` you log isn't necessarily the token used on the next attempt. Cosmetic, but consider carrying the fetched token forward into the next attempt to avoid the redundant read and the misleading log.

### Looks good
- Strict double-anchor `is_invalid_token_error` correctly avoids misclassifying bare 401s.
- Eager pre-spawn token validation gives callers an immediate, actionable error.
- Session-expired state is surfaced cleanly to the frontend instead of a silent stall.
- Backward-compatible public API via `static_token_provider`.

(CI was still re-running at review time — please confirm it's green alongside the fix.)

YOMXXX:
@oxoxDev addressed both points in 3 follow-up commits:

- **Major**: `9afa9a0c` bounds `RetryImmediately` to one attempt per fresh-token cycle. A new `fresh_token_retries` counter (reset on successful connection) caps the immediate-retry path; on the second consecutive `RetryImmediately` we now fall through to `consecutive_failures += 1` + `log_connection_failure` + the normal backoff sleep, so a non-deterministic provider can't hot-loop.
- **Regression test**: `c845fa91` `ws_loop_bounds_fresh_token_retries_with_rotating_provider` — provider returns a different non-empty token every call, server always returns Invalid token; asserts the loop bounds total attempts (≤ 20) and reaches a Disconnected state instead of looping indefinitely.
- **Minor**: `f39cdbed` carries the fresh token forward via a `pending_token: Option<String>` consumed at the top of the next iteration, so `run_connection` uses exactly the token validated by `decide_after_invalid_token` and the logged `fresh_len` matches what's actually sent.

122 socket tests pass with `GGML_NATIVE=OFF`; `cargo fmt` clean. Ready for another look when you have time.

oxoxDev (approved):
Re-review — both findings addressed and CI is green. Approving.

- **Major (unbounded retry) fixed**: the `fresh_token_retries` counter now bounds the immediate-retry path to a single shot per cycle, then falls through to the normal backoff + `consecutive_failures` escalation. A provider that returns ever-changing tokens can no longer hot-loop without sleeping or escalating. Nicely covered by the new non-deterministic token-rotating test.
- **Minor (double provider read) fixed**: `InvalidTokenAction::RetryImmediately { token }` now carries the freshly fetched token forward instead of discarding it and re-reading at the top of the loop.

Clean fix for the stale-token retry storm (#2892 / TAURI-RUST-9C). The `TokenProvider` abstraction, strict double-anchor `is_invalid_token_error`, eager pre-spawn validation, and session-expired surfacing all hold up well. LGTM.

oxoxDev:
@graycyrus bro can you review again

PR #2906, opened 2026-05-29
fix(subconscious): suppress repeated Sentry noise when DB init fails (TAURI-RUST-A, closes #2903)

### LINKED ISSUE #2903: Subconscious schema DDL failure fires 1,499 Sentry errors on every RPC call
## Summary

The subconscious module fails to create its database tables (DDL), causing every `subconscious_tasks_list` RPC call to fail with "failed to run subconscious schema DDL". 1,499 events since project start, ongoing.

## Problem

**What happened:** The subconscious domain tries to initialize its SQLite schema on first use. The DDL execution fails (likely database locked, path inaccessible, or SQL compatibility issue). After this, every subsequent RPC call to `openhuman.subconscious_tasks_list` fails and fires a Sentry error.

**Expected:** Either:
1. The DDL should succeed reliably, or
2. The error should be caught at boot, logged once as error, and subsequent calls should return an empty/degraded response instead of firing Sentry events repeatedly

**Impact:** 1,499 events in Sentry (TAURI-RUST-A), **ongoing** across releases 0.54.0 → 0.56.0. Tags: method=openhuman.subconscious_tasks_list, Windows x86_64.

**Version / Platform:** openhuman@0.54.0 through 0.56.0, Windows 10.0.26200

## Solution (optional)

Check `src/openhuman/subconscious/` for:
1. The DDL initialization code — what SQL is being run? Does it use `CREATE TABLE IF NOT EXISTS`?
2. Is the database path valid and writable?
3. Is there a race condition (multiple calls trying to init simultaneously)?
4. Can the init failure be cached so subsequent calls don't re-fire Sentry events?

## Acceptance criteria

- [ ] **Repro gone** — Subconscious DDL either succeeds or fails gracefully without flooding Sentry
- [ ] **Regression safety** — Unit test for DDL initialization

## Related

- Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/64/

graycyrus:
## Summary

- **Root cause**: `store::with_connection` runs the full schema DDL on every call. Under concurrent in-process RPC calls (status poll every 3 s, tasks-list on Intelligence page, manual trigger), SQLite's default `busy_timeout=0` causes immediate `SQLITE_BUSY` failures — producing ~1,300 identical Sentry events per 24 h from a single underlying cause (TAURI-RUST-A).
- **Fix**: adopt a three-layer defence: (1) per-connection `busy_timeout=5s` to absorb the common concurrent-connection race, (2) application-level retry loop (3 retries, 100/300/900 ms exponential backoff) for the residual case where the timeout is exhausted, (3) retry classifier (`is_sqlite_busy`) that only retries `DatabaseBusy`/`DatabaseLocked` — schema or corruption errors fail through immediately so Sentry gets a real root cause.
- **Workspace-safe**: unlike the previous global-flag approach, this retry strategy is per-connection and scoped by `workspace_dir` — no cross-workspace poisoning.
- Replaces the earlier `DDL_EVER_FAILED` / `DB_INIT_FAILED_SENTINEL` approach that CodeRabbit correctly identified as globally poisoning all workspaces after one failure.

## Test plan

- [x] `cargo check` passes (pure Rust change)
- [x] Existing subconscious unit tests pass (CRUD tasks/log/escalations, seed, recurrence roundtrip)
- [x] New tests: `is_sqlite_busy_{matches,does_not_match}*` assert the retry classifier, `with_connection_resolves_external_write_contention` exercises the retry loop against a real write-lock held for 250 ms
- [x] `is_db_init_failed` helper and `DB_INIT_FAILED_SENTINEL` constant removed — no workspace-poisoning possible

## Notes

Merge of upstream/main resolves all conflicts. Pre-push hook failure (Prettier not installed in worktree) is pre-existing and unrelated to this Rust-only change.

Closes #2903


graycyrus:
@coderabbitai review

PR #2919, opened 2026-05-29
feat(intelligence): Belief Ledger — temporal contradiction & belief-revision tracker

aashir-athar:
## Summary

A new **Beliefs** tab on the Intelligence page that reads the agent's knowledge graph (`openhuman.memory_graph_query` → `GraphRelation[]`) and reconstructs, **per (subject, predicate) claim, how what the agent believes about you changed over time**.

When newer memories contradict or supersede older ones about the same thing, the ledger:
- groups the conflicting object-values within a claim and orders them by `updatedAt`;
- picks a **current best belief** via a recency × `log(evidence)` score (180-day half-life);
- classifies each revision — **reinforced / refined / superseded / contradicted**;
- shows a per-claim **revision timeline** with a **confidence** score (the score-margin between the winner and runner-up);
- lets you **"Explain this contradiction"** (natural-language, via the existing agent) and **"Set the truth"** (persists a correction note).

## Why this is novel

Every memory/RAG product (ChatGPT memory, Mem, Rewind, Personal.ai, Pieces…) treats memory as **append-only or last-write-wins** — they silently overwrite or silently accumulate. None expose a *belief ledger*: the explicit recognition that "you said Berlin in January, Lisbon in May" is a **contradiction over time**, surface both, resolve a winner by recency+evidence, and let the user audit/correct it. OpenHuman's `GraphRelation` stream (subject/predicate/object + `evidenceCount` + `updatedAt`) is exactly the substrate this needs and that generic chat-memory products lack.

## Architecture — **zero new Rust**

| File | Role |
|---|---|
| `app/src/lib/memory/beliefLedger.ts` | Pure, deterministic engine. `NOW` is **injected** — no `Date.now()`/`Math.random()` — so every branch is unit-testable without a clock/network/DB. |
| `app/src/services/api/beliefLedgerApi.ts` | Thin facade composing **three already-shipped RPCs**: `memory_graph_query`, `agent_chat`, `memory_doc_ingest`. No new core surface. |
| `app/src/components/intelligence/BeliefLedger.tsx` | Pure presentational timeline (conflict badges, confidence/evidence pills, current-belief highlight). |
| `app/src/components/intelligence/BeliefLedgerTab.tsx` | Container: load-on-mount, explain, correct (via `ConfirmationModal`). |
| `app/src/pages/Intelligence.tsx` | Mounts the `beliefs` tab exactly like the existing tabs. |
| `en.ts` + `chunks/*-1.ts` | 27 i18n keys across all locales (Polish inherits via its `...en1` spread). |

## Validation

- **63 frontend tests** (engine 38 · api 10 · component 9 · tab 6) — `pnpm vitest`.
- `tsc --noEmit` clean · `eslint` **0 errors** · `prettier --check` clean · `pnpm i18n:check` **EXIT 0** (0 missing across all locales).
- No Rust changes → the Rust CI lane sees no new code.

## Review-hardened before opening

I ran an adversarial multi-lens self-review over the diff. Its two **confirmed correctness findings** are fixed and pinned by tests:
1. Punctuation-only objects (e.g. `"..."`) that normalize to empty no longer become a phantom belief value.
2. `correctionKey` no longer collapses distinct non-Latin / punctuation-only claims to a colliding key (uses `encodeURIComponent(normalizeText(...))`).

Plus: tightened the revision-classification heuristic from raw substring to **token-subset** (kills false "refined" like *go*→*mango*), separated save-success from reload-failure in the correction flow, added `role="img"` to the confidence/evidence pills, and made the per-claim display lookup O(n).

## A11y

`<article aria-label>` per claim, `<ol>/<li>` timeline, `role="img"` labelled confidence & evidence pills, `role="status"` loading region, real `<button>`s with descriptive `aria-label`s, dark-mode classes throughout.

## Summary by CodeRabbit

* **New Features**
  * Added a "Beliefs" tab with a Belief Ledger UI: per-claim revision timelines, confidence/evidence badges, contested indicators, inline explanations, sorting to surface most-contested claims, and a "Set truth" correction workflow with confirmation and toasts.
  * AI-driven "Explain this contradiction" flow is now locale-aware and renders progress/inline explanations.

* **Tests**
  * Extensive unit and integration tests covering ledger logic, UI interactions, API flows, and correction/explain behavior.

* **Internationalization**
  * Added Belief Ledger translation strings across many languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2919?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@aashir-athar hey — CI has three checks still pending (Rust Core Coverage, E2E Windows Appium, Rust Core Tests Windows), so I'll hold the approve until those are green. While I was in the diff I spotted a few things beyond what CodeRabbit already flagged:

CodeRabbit's three findings are real and need to be addressed before this merges — the correction modal always re-saves the algorithm's current winner instead of letting the user pick the actual correct value, the PII log in `saveCorrection`, and the inline ARIA assembly.

Beyond those:

**`BeliefLedgerTab.tsx` line 120** — `onExplain={key => void handleExplain(key)}` creates a fresh function reference on every render, defeating the `useCallback` you set up on `handleExplain`. Since `BeliefLedgerProps.onExplain` types the return as `void` and TypeScript accepts `Promise<void>` there, you can pass `handleExplain` directly: `onExplain={handleExplain}`.

**`beliefLedgerApi.ts` `buildExplanationPrompt`** — the LLM prompt template is English-only. The rest of the surface is fully i18n'd, but non-English users will still receive explanations in English because the prompt sends hard-coded English instructions to the model. Worth either threading the locale through or using i18n keys for the template strings before this ships to non-English locales.

**`BeliefLedger.tsx` `distinctValues` interpolation** — `t('beliefLedger.distinctValues').replace('{n}', String(history.distinctValueCount))` does manual string substitution inline. If the project's i18n framework has built-in parameter interpolation, this should use it for consistency and to avoid word-order breakage in right-to-left or inflected locales. If `.replace()` is the established pattern here, disregard.

Overall the engine is well thought out — injected clock, deterministic scoring, thorough test coverage. The correction UX bug is the one I'd prioritize fixing; the rest are polish. Fix the CI failures and CodeRabbit's issues and I'll come back to approve.

app/src/components/intelligence/BeliefLedgerTab.tsx
```diff
@@ -0,0 +1,149 @@
...
+        onConfirm: () => {
+          setModal(CLOSED_MODAL);
+          void runCorrection(history);
+        },
+      });
+    },
+    [runCorrection, t]
+  );
+
+  return (
```
graycyrus, line 120: **[minor]** `key => void handleExplain(key)` creates a new function reference on every render, negating the `useCallback` on `handleExplain`. `BeliefLedgerProps.onExplain` is typed `(history: BeliefHistory) => void`, and TypeScript accepts `Promise<void>` there, so you can write `onExplain={handleExplain}` directly.

app/src/services/api/beliefLedgerApi.ts
```diff
@@ -0,0 +1,107 @@
...
+  lines.push(
+    `The knowledge graph recorded conflicting values over time for the claim "${history.subject} ${history.predicate} …".`
+  );
+  lines.push('Observed values, newest first:');
+  history.revisions.forEach(rev => {
+    const when = new Date(rev.version.updatedAt * 1000).toISOString().slice(0, 10);
+    const tag = rev.isCurrent ? ' [current best belief]' : '';
+    lines.push(
+      `- "${rev.version.object}" (${rev.kind}, seen ${when}, evidence x${rev.version.evidenceCount})${tag}`
+    );
```
graycyrus, line 73: **[minor]** `buildExplanationPrompt` sends English-only instructions to the LLM regardless of the user's locale. For a fully internationalized product, non-English users will receive explanations in English. Consider threading the locale through or deriving the prompt template from i18n keys before this ships to non-English locales.

aashir-athar:
Thanks for the review! All three addressed in 8411fa5f:

1. **Correction could only re-save the algorithm's winner** — fixed. Added a value picker (`ClaimCorrection`): the user selects any observed candidate value (default = current best belief) **or types a brand-new value**, and that chosen value is what's persisted. `onCorrect` now threads the selected value end-to-end; the modal copy is updated to "the selected value." New tests cover picking a non-current value and entering a custom one.
2. **Inline ARIA word order** — moved the assembled labels into dedicated i18n keys with placeholders (`confidenceAria`, `evidenceAria`, `explainAria`, `setTruthAria`, `correctionSelectAria`, `correctionCustomLabel`), translated across every locale chunk. The aria templates embed the visible button text so WCAG 2.5.3 (Label in Name) holds.
3. **PII in logs** — `saveCorrection` now logs only the key, never the corrected value.

Validation: 64 belief tests pass, tsc clean, eslint 0 errors, prettier clean, i18n:check EXIT 0.

app/src/components/intelligence/BeliefLedgerTab.tsx
graycyrus: Still present in 8411fa5 — the wrapper appears at line 130 of BeliefLedgerTab.tsx. Minor but trivial to fix: change to `onExplain={handleExplain}` directly. TypeScript accepts `Promise<void>` where `void` is expected.

app/src/services/api/beliefLedgerApi.ts
graycyrus: Still English-only in 8411fa5. Not a blocker, but worth addressing before this ships broadly to non-English users.

graycyrus (commented):
@aashir-athar good progress — the new commit addressed all three CodeRabbit issues cleanly (ARIA keys moved to i18n, the correction flow now lets users pick or type the actual correct value instead of always re-saving the algorithm's winner, and the PII log is gone). That's the right fixes.

CI is still fully pending so I'm holding off on approving, but the two minor items I flagged earlier are still open (the `onExplain` wrapper and the English-only explanation prompt). Neither is blocking, but clean those up when you get a chance. Once CI goes green I'll come back and approve.

aashir-athar:
Thanks @graycyrus — much appreciated, and good catches on both:

1. `onExplain={key => void handleExplain(key)}` — agreed, the wrapper negates the useCallback; I'll pass `handleExplain` directly (TS accepts `Promise<void>` where `void` is expected).
2. English-only explanation prompt — good point for full i18n; I'll thread the UI locale through `explainConflict` and add a 'respond in {locale}' directive to the prompt.

Both are quick; I'll push them shortly. Agreed the remaining gate is the heavy CI lanes (Rust Core Coverage / E2E Windows / Rust Core Tests Windows) going green — this branch makes no Rust changes, so those should land the same as main.

aashir-athar:
Thanks @graycyrus — both addressed in 7216899d:

1. **Locale-aware explanation** — `explainConflict` / `buildExplanationPrompt` now take the active UI locale (threaded from `useT().locale`) and append a `Respond in <language>.` directive, so the contradiction explanation comes back in the user's language. English/unknown locales add nothing. I used an explicit English-name map rather than `Intl.DisplayNames` so the prompt stays deterministic across Node/ICU versions and remains exactly assertable in tests.
2. **Direct handler** — `<BeliefLedger onExplain={handleExplain} />` now, dropping the redundant `key => void handleExplain(key)` wrapper (verified ESLint `no-misused-promises` stays clean).

Added 4 tests covering the directive (English/none → no directive, `es` → "Respond in Spanish.", `zh-CN` → "Respond in Chinese.", unknown locale → none) and that `explainConflict` threads the locale through. Full local suite green (68 belief-ledger tests, tsc, eslint, prettier).

aashirathar32:
Upstream maintainer action needed: the fresh CI workflows for this branch are queued, and GitHub is showing **3 workflows awaiting approval**. Please click **Approve and run workflows** so the required checks can execute.

The branch has been synced with the current `main` and the review fixes have been pushed. I will follow through on any failures after the workflows are released.

PR #2925, opened 2026-05-29
feat(intelligence): add Prediction Calibration Ledger

aashir-athar:
## Summary

Adds a **Prediction Calibration Ledger** — a new "Calibration" tab on the Intelligence page where you log forecasts with a subjective confidence ("80% sure the deploy ships Friday"), later resolve each TRUE/FALSE, and get scored on your *own* forecasting calibration.

No personal-AI product scores the user's own predictions woven into their assistant — forecasting platforms score communities, not your private calls. This turns the assistant into a calibration coach.

### What it computes
- **Brier score** — mean `(confidence − outcome)²` (lower is better).
- **10-bin calibration curve** — predicted confidence vs actual hit-rate per bin.
- **Murphy decomposition** — reliability / resolution / uncertainty. The identity `brier == reliability − resolution + uncertainty` holds to machine precision (reliability uses each bin's *mean* predicted confidence, not the midpoint).
- **Over/under-confidence** — mean confidence vs base rate.
- **Cumulative-Brier trend** — a sparkline of whether calibration is improving over time.

### Design
- **Pure deterministic engine** (`lib/memory/predictionLedger.ts`) — no `Date.now()`, no RNG. Scoring depends only on `(confidence, outcome)` pairs, so the same records always yield the same report and every branch is unit-testable. The Murphy identity is verified to ~1e-9 over equal-within-bin fixtures (and correctly *breaks* when within-bin forecasts diverge, confirming the fixtures are honest).
- **Zero new core surface** — the ledger persists as a single JSON blob via the already-shipped memory file RPCs (`memory_read_file` / `memory_write_file`). `memory_doc_ingest` / `memory_list_documents` deliberately aren't used (the list RPC returns only summaries, so confidence/outcome could never round-trip).
- **Container/presentational split** — all I/O and id/timestamp minting live in the container (in event handlers, never during render, satisfying `react-hooks/purity`); the view is pure. Optimistic writes **roll back and toast** on a persistence failure.
- **i18n across all 13 locales** (no dead keys — verified by the coverage gate).

## Test plan
- [x] `vitest` — 48 tests across engine, API facade, view, and container (incl. Murphy-identity, bin-edge, overdue, and save-failure-rollback cases)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors
- [x] i18n coverage gate — EXIT 0, no unused `predictionLedger.*` keys

## Summary by CodeRabbit

* **New Features**
  * New Prediction Calibration tab: log predictions, mark true/false, delete entries, optional due dates with Overdue badges, and add-prediction form.

* **New Views**
  * Calibration metrics and visuals: Brier score, base-rate / mean-confidence, calibration tone, calibration curve, reliability/resolution/uncertainty breakdown, and cumulative Brier trend.

* **Persistence**
  * Ledger persisted as JSON with optimistic updates, save/load behavior, and error/rollback feedback.

* **Tests**
  * Comprehensive tests for UI flows, scoring, persistence, and parsing.

* **Localization**
  * Translations added for the calibration UI in many languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2925?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/components/intelligence/PredictionLedgerTab.tsx
```diff
@@ -0,0 +1,148 @@
...
+    (values: AddPredictionFormValues) => {
+      const record: PredictionRecord = {
+        id: crypto.randomUUID(),
+        statement: values.statement,
+        confidence: normalizeConfidence(values.confidencePercent / 100),
+        createdAt: nowSeconds(),
+        resolveBy: values.resolveBy,
+        outcome: null,
+        resolvedAt: null,
+        category: values.category,
```
graycyrus, line 110: **[major]** `records` captured here is from the render this `handleAdd` closure was created in, not the post-optimistic-update state. If two predictions land before a re-render — rapid keyboard submit, test, or any batched event — `addRecord(records, record)` on the second call operates on the pre-first-add snapshot. The save writes `[record2]` instead of `[record1, record2]` and the first prediction is silently dropped from disk.

This is the write-path side of the same stale-closure problem CodeRabbit flagged on the rollback path. The two need to be fixed together. One approach: track the authoritative records in a `useRef` updated synchronously inside `persist`, and derive `next` from the ref rather than the closure-captured state.

graycyrus (commented):
@aashir-athar heads up — CI is still pending on this PR (Build Tauri App, E2E Windows/macOS/Linux Appium, Rust Core and Frontend coverage, Rust Tauri Shell Tests are all still running), so i'll hold off on a full sign-off until those land. i did spot a few things while skimming though:

the stale-closure issue CodeRabbit flagged in the rollback path is deeper than just the catch block — `handleAdd` and `handleResolve` both capture `records` from the same render closure. if two predictions are submitted before a re-render (race, test, fast keyboard user), the second `addRecord(records, record)` operates on the pre-first-add snapshot and the save writes `[record2]` instead of `[record1, record2]`. the rollback fix and this one need to be solved together, since they share the same root.

`PredictionLedgerTab` never passes `nowSeconds` to `<PredictionLedger>`. the view falls back to a mount-time snapshot via `useState(() => Math.floor(Date.now() / 1000))`, which means overdue badges go stale for a user who keeps the calibration tab open across a deadline boundary without remounting. easy fix — pass `nowSeconds()` as a prop, or wire a 60s interval in the container.

i also want to confirm the loadLedger error-surfacing issue and the optimistic rollback (both flagged by CodeRabbit) are addressed before this goes in — those are the two i'd call blockers.

the scoring engine itself is solid — Murphy identity, bin-edge coverage, rollback-on-failure are all well thought out. fix the CI and those three items and i'll come back for approval.

aashir-athar:
Thanks @coderabbitai — all 5 comments addressed in 0116a0f9. I verified each against the code first:

**1. `loadLedger` swallows all errors (api L29-40) + `refresh()` can't surface failures (Tab L56)** — valid, fixed together at the root. `loadLedger` now returns `[]` only for the two benign cases (missing file → the read RPC rejects on first run; empty/whitespace content) and **throws** on a present-but-malformed blob. So a corrupt ledger now surfaces in `refresh()`'s error banner instead of silently rendering empty and letting the next save overwrite it. Updated the api test (`throws on a malformed (corrupt) blob` replaces the old never-throws assertion).

**2. Optimistic rollback unsafe when saves overlap (Tab L83)** — valid, fixed. `persist()` now stamps a monotonic `persistSeq` ref; the `.catch()` rolls back only when `seq === persistSeq.current`, so a stale failure can't clobber a newer save that already succeeded. (I went with the seq-guard + rollback-to-previous rather than a disk reload, since reloading races with in-flight writes and `loadLedger` can legitimately return `[]` for a missing file.)

**3. Parser accepts outcome without resolvedAt (predictionLedger.ts L278)** — valid, fixed. `parsePredictionRecords` now treats `(outcome, resolvedAt)` as a single resolved-pair: a record counts as resolved only when both are valid, else it's coerced back to open. Added a test covering outcome-only, resolvedAt-only, and the valid pair.

**4. Trend sparkline Y inverted (PredictionLedger.tsx L289)** — valid, applied your suggested `y = 24 - clamp(brier)*24` so a lower (better) Brier renders lower on screen, matching the caption.

Local suite green: 49 prediction-ledger tests, `tsc --noEmit`, ESLint (0 errors), Prettier.

app/src/components/intelligence/PredictionLedgerTab.tsx
graycyrus: After re-analyzing with both commits in scope: the stale-closure risk doesn't materialize in practice. For `handleAdd` — the form resets `statement` in the same batched React event that calls `persist`, which disables the submit button immediately; the user cannot submit again before re-render, and that re-render carries fresh `records`. For `handleResolve` — `resolveRecord` is a no-op on already-resolved records (it guards `r.outcome === null`), so a theoretical rapid double-click just writes the same outcome twice with no data lost. The `persistSeq` ref added in the fix commit correctly handles the overlapping-save rollback that CodeRabbit flagged. Marking resolved.

graycyrus (commented):
@aashir-athar the fix commit addresses the main feedback well — corruption now throws instead of silently returning [], the SVG Y-axis is inverted correctly, outcome/resolvedAt are enforced as a pair, and the `persistSeq` ref fixes the overlapping-save rollback. I've reviewed my earlier stale-closure concern on the write path and resolved it: the form's own reset semantics prevent the double-submit race in `handleAdd`, and `resolveRecord`'s `outcome === null` guard makes a double-click on `handleResolve` a safe no-op.

CI is still running (Build Tauri App, E2E Appium on all three platforms, Rust Windows tests are pending) so I'm holding off on approval until those go green.

One minor thing left for a follow-up if you want to polish it: the container never passes `nowSeconds` down to `<PredictionLedger>`, so the view falls back to its mount-time snapshot. A prediction that crosses its deadline while the tab is open won't show the Overdue badge until the tab is remounted. Not blocking — the badge is cosmetic — but easy to fix by passing `nowSeconds={Math.floor(Date.now() / 1000)}` from a periodic tick or just recomputing on each render.

aashir-athar:
@coderabbitai review

All four review findings were addressed in 0116a0f9 (you confirmed the fixes look solid). Requesting a fresh review to clear the stale changes-requested state — CI is green and the PR is mergeable. Thanks!

aashir-athar:
CI note: after rebasing onto current main, the heavy **Rust Core Coverage** and **E2E (Linux / Appium Chromium)** lanes went red — but this PR is **TypeScript-only** (the Rust core is byte-identical to main), and the failing jobs show infra-level failure (no test/assertion output — cold-cache/coverage-upload timeouts), the same pattern as earlier flakes. I can't rerun CI (no repo rerun rights), so I pushed an empty commit to re-trigger. A maintainer re-run would also clear it.

senamakel:
Hi @aashir-athar — I reviewed the rebased head (`5ff446fc`). The CodeRabbit + maintainer threads are all addressed; nice work. But the rebase introduced a **real CI failure** worth fixing before this can go green:

### 🔴 Fix these (real failures)
- **`i18n Coverage (parity)`** and **`Frontend Coverage (Vitest)`** are failing for the same root cause: the rebase added the 43 `predictionLedger.*` keys + `memory.tab.calibration` to **`en.ts` only**. All 13 non-English locales (`ar, bn, de, es, fr, hi, id, it, ko, pl, pt, ru, zh-CN`) are missing them, so `pnpm i18n:check` reports `missing` keys in every locale, and the `i18n/__tests__/coverage.test.ts` parity assertions (part of the Vitest lane) fail too.

  **Fix:** add the same keys to every non-English locale file, using the English value as the translator placeholder (per the i18n parity rule in `CLAUDE.md`). Once added, `pnpm i18n:check` shows `missing: 0 / extra: 0` across all 13 locales and the Vitest i18n coverage suite passes 39/39.

### 🟡 Optional (lint warning, non-blocking)
- `PredictionLedgerTab.tsx` fires `void refresh()` directly in `useEffect`, which trips `react-hooks/set-state-in-effect`. Deferring it through a microtask with a cancelled guard (matching `SubconsciousReflectionCards`) clears the warning.

### ⚪ Ignore these
- The **E2E (Playwright web lanes)** failures are the known flaky infra lanes (no test/assertion output) — not caused by this PR. A maintainer re-run clears them; no code change needed.

---

I've already prepared both fixes locally on top of `5ff446fc` but can't push to your fork (no write access). Two ways to land them:

```bash
# Option A — cherry-pick (commits authored against 5ff446fc)
#   fix(i18n): add predictionLedger keys to all non-English locales
#   fix(prediction-ledger): defer initial load to a microtask

# Option B — regenerate the i18n fix yourself: copy the predictionLedger.*
# block from en.ts into each of the 13 non-English locale files, right after
# the 'memory.tab.settings' key, then run:
pnpm i18n:check          # expect missing: 0 / extra: 0 for all locales
pnpm --filter openhuman-app test:unit src/lib/i18n/__tests__/coverage.test.ts
```

If you'd rather I push directly, add me (`senamakel`) as a collaborator on your fork and I'll push the two commits.

aashirathar32:
Upstream maintainer action needed: the fresh CI workflows for this branch are queued, and GitHub is showing **3 workflows awaiting approval**. Please click **Approve and run workflows** so the required checks can execute.

The branch has been synced with the current `main` and the review fixes have been pushed. I will follow through on any failures after the workflows are released.

PR #2927, opened 2026-05-29
perf(memory_tree): batch embedding fetch in topic rerank (eliminate N sequential SQLite round-trips)

mysma-9403:
## Summary

`memory_tree::retrieval::topic::rerank_by_semantic_similarity` previously did a **sequential `spawn_blocking(...).await` per hit** to fetch each embedding from SQLite — N round-trips and N tokio task hops for an N-hit topic candidate set, all on the hot path of every topic-driven retrieval.

This PR replaces that loop with a **partitioned batch fetch**: hit ids are split by `NodeKind` (Summary vs Leaf), then a **single** `spawn_blocking` issues **one batched `IN (?,?,?,...)` SQL query per table**, returning a `HashMap<id, Vec<f32>>` per partition. Decoration walks the original hit order, so tie-break stability and the `(NEG_INFINITY, false)` sink-to-bottom treatment of missing embeddings are preserved bit-for-bit.

### New helpers (public)

In `memory_store::chunks::store` and `memory_store::trees::store`:

- `get_chunk_embeddings_for_signature_batch(config, chunk_ids, model_signature) -> Result<HashMap<String, Vec<f32>>>`
- `get_chunk_embeddings_batch(config, chunk_ids) -> ...` — active-signature wrapper
- `get_summary_embeddings_for_signature_batch(config, summary_ids, model_signature) -> Result<HashMap<String, Vec<f32>>>`
- `get_summary_embeddings_batch(config, summary_ids) -> ...` — active-signature wrapper

### Defensive chunking — `MAX_EMBEDDING_BATCH = 500`

Both batch helpers iterate via `chunks(MAX_EMBEDDING_BATCH)` and merge into one map.

- SQLite's hard limit (`SQLITE_MAX_VARIABLE_NUMBER`) is **32 766** since 3.32 → 500 gives a **~65× safety margin**.
- The current call-site cap (`LOOKUP_HEADROOM = 200`) never trips the window — runs as one query today.
- The loop is **purely defensive**: future call-sites passing larger id lists will not blow up against a host with a lower compile-time `SQLITE_MAX_VARIABLE_NUMBER`.
- **No volume reduction**: all input ids in → all matching ids out. Chunking only splits the SQL; the merged `HashMap` is identical to what one giant query would return.

- Same model-signature scoping via `tree_active_signature(config)`.
- Same blob decoders reused (`embedding_from_blob`, `decode_signature_blob`).
- Original hit order preserved → tie-break stability identical to pre-PR.
- Missing embeddings still sink to bottom via `(f32::NEG_INFINITY, false, hit)`.
- Heterogeneous node kinds (`Summary` vs `Leaf`) still read from their own tables (`mem_tree_summary_embeddings` vs `mem_tree_chunk_embeddings`).
- Zero i18n / integration / RPC / UI surface touched — pure in-process SQLite refactor.

### New unit tests (7 total)

**chunks side** (`src/openhuman/memory_store/chunks/store_tests.rs`):
- [x] `batch_embedding_lookup_returns_only_signature_scoped_rows` — proves signature scoping
- [x] `batch_embedding_lookup_empty_input_returns_empty_map` — empty input safety
- [x] `batch_embedding_lookup_unknown_ids_absent_from_map` — missing-id semantics match per-row helper
- [x] `batch_embedding_lookup_splits_id_list_above_per_batch_threshold` — **501 ids forces two `IN (...)` queries**, validates chunking correctness

**summaries side** (`src/openhuman/memory_store/trees/store_tests.rs`) — mirror trio:
- [x] `summary_batch_embedding_lookup_returns_only_signature_scoped_rows`
- [x] `summary_batch_embedding_lookup_empty_input_returns_empty_map`
- [x] `summary_batch_embedding_lookup_unknown_ids_absent_from_map`

### Existing test surface — verified no regression

- [x] All 12 existing `memory_tree::retrieval::topic::tests` pass.
- [x] Broader sweep — **908 tests across `memory_store::` + `memory_tree::`** — 0 failures.
- [x] `cargo fmt` — clean.

## Summary by CodeRabbit

* **Performance**
  * Batched embedding retrieval for semantic reranking, reducing database round-trips and improving query responsiveness.

* **Refactor**
  * Rerank path updated to fetch embeddings in bulk while preserving original hit ordering and previous missing-embedding behavior.

* **Tests**
  * Added unit tests covering batch embedding lookups, boundary batching, signature filtering, and empty/missing-id cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2927?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/openhuman/memory_store/chunks/embeddings.rs
+++ b/src/openhuman/memory_store/chunks/embeddings.rs
@@ -3,6 +3,7 @@ use crate::openhuman::config::Config;
 use anyhow::{Context, Result};
 use chrono::Utc;
 use rusqlite::{Connection, OptionalExtension};
+use std::collections::HashMap;
 
 // ── Phase 2: embedding column accessors ─────────────────────────────────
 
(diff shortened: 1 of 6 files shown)


graycyrus (commented):
@mysma-9403 hey! the code looks good to me — this is a clean, well-reasoned fix. the N sequential `spawn_blocking` round-trips on the hot rerank path was a real problem and the batched IN approach solves it correctly.

a few things i verified:

- the `?{sig_idx}` positional binding in the dynamic SQL is correct — SQLite assigns `?` without number sequentially (1..n), and `?{n+1}` explicitly targets the last bind slot, which rusqlite fills with `model_signature`. no injection surface since only the placeholder count comes from code, not user data.
- partition → batch fetch → decorate in original hit order preserves sort stability and the `(NEG_INFINITY, false)` sink-to-bottom behaviour for missing embeddings.
- the `MAX_EMBEDDING_BATCH = 500` constant with `LOOKUP_HEADROOM = 200` gives the stated ~65× headroom against SQLite's variable cap.
- one asymmetry worth noting: `get_summary_embeddings_for_signature_batch` adds an explicit `v.len() != dim` check after `decode_signature_blob`, but the chunk side delegates that to `embedding_from_blob`. not a correctness issue, but if `decode_signature_blob` doesn't validate length internally the summary path is tighter — just keep that in mind if you ever reuse these helpers elsewhere.
- test coverage is solid: signature scoping, empty input, unknown ids, and the 501-id window-split case all covered.

CI is still running — once those are green i'll come back and approve this. let me know if anything comes up.

graycyrus (commented):
@mysma-9403 hey — the new commit dropping the redundant len check in the summary batch helper is the right call. Code is still clean.

One blocker before I can approve: the Coverage Gate (diff-cover ≥ 80%) check is failing, and several other checks (Build Tauri App, Rust Core/Frontend Coverage runners) are still pending. Once those land green I'll approve. Let me know if you need a hand tracking down what the coverage gate is complaining about.

oxoxDev (dismissed):
LGTM — clean, behavior-preserving perf fix. The per-hit `spawn_blocking(get_*_embedding).await` loop (N sequential SQLite round-trips, up to LOOKUP_HEADROOM=200 on the hot path) is replaced by partitioning hit ids by `NodeKind` into two batched `IN (...)` queries (summary + chunk tables) under a single `spawn_blocking`. Decoration still runs in the original `hits` order, and the `(NEG_INFINITY, false)` missing-embedding sink is preserved, so sort stability / tie-breaks / result-set membership are unchanged. Batched helpers chunk internally for the SQLite variable cap. Tests added. Approving.

PR #2931, opened 2026-05-29
feat(intelligence): add Knowledge Graph Centrality

aashir-athar:
## Summary

Adds a new read-only **"Centrality"** tab on the Intelligence page that runs graph algorithms over the memory knowledge graph to surface its *structural backbone* — the load-bearing **hub** entities, and crucially the **bridge/connector** entities that link otherwise-separate clusters.

The existing memory surfaces show the graph spatially or sort it by raw evidence count; neither can reveal a low-frequency entity that is nonetheless structurally central because it connects disjoint clusters. PageRank can — that contrast is the point of this lens.

### What it computes (pure deterministic engine)
- **PageRank** — directed, damping 0.85, Jacobi power iteration. Dangling/sink nodes' mass is redistributed uniformly (`d·D/N`) so the rank vector sums to 1 every iteration.
- **Degree centrality** — in / out / total, plus an evidence-weighted variant.
- **Connected components** — weakly-connected island count via union-find (self-loops add no connectivity).
- **Bridge detection** — flags entities whose PageRank rank far outranks their degree rank (the connector signal).

### Design
- **Pure & deterministic** (`lib/memory/graphCentrality.ts`): no clock, no RNG. Node order is pinned by a total string ordering and PageRank uses Jacobi iteration into a fresh array, so the same graph always yields byte-identical ranks regardless of relation insertion order. Known-value fixtures (2-node mutual, directed cycle, dangling sink, evidence-weighted, 4-node mixed with a closed-form `(1-d)/N` anchor, self-loops) are asserted to 6 decimals; a dangling-mass canary asserts Σrank = 1.
- **Zero new core surface**: reuses the already-shipped `memoryGraphQuery` and `memoryListNamespaces` RPC wrappers. **Read-only** — the result is recomputed from the live graph on mount, never persisted (no staleness, no new store).
- **Container/presentational split**; namespace selector; 3 metric tiles (Entities / Connections / Clusters) + a top-25 ranked table with an inline influence bar and a "connector" badge. i18n across all 13 locales.

### Edge cases handled (and tested)
Empty graph, single node, self-loop, dangling sink, disconnected components, duplicate/multi-predicate edge collapse (weights summed), zero/negative/NaN `evidenceCount` (clamped to ≥1), case/whitespace-variant entities (kept distinct by design), malformed rows (dropped), entities containing the separator (no collision — nested-map collapse), and the iteration cap (`converged: false` fallback, never hangs).

## Test plan
- [x] `vitest` — 32 tests (engine fixtures + determinism + degree/collapse/sanitization + bridges; api facade; presentational panel; container load/selector/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors
- [x] i18n coverage gate — EXIT 0, no unused `graphCentrality.*` keys

## Summary by CodeRabbit

* **New Features**
  * Added a Centrality tab in Intelligence showing entity influence rankings, metric tiles (entities, connections, clusters), connector/bridge badges, ranked hubs, and namespace filtering with retry on error.

* **Internationalization**
  * Added translations for the Centrality UI across 13+ languages.

* **Tests**
  * Added extensive unit and UI tests for centrality computation, bridge detection, API integration, and the Centrality UI components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2931?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@aashir-athar hey! the code looks good to me, but CI is still pending on this one. once those checks are green, i'll come back and approve this. let me know if you need any help!

Quick notes from the read-through while i'm here:

The engine is solid. Deterministic node ordering, correct dangling-mass redistribution in the PageRank loop, safe self-loop handling in union-find, proper evidence weight sanitisation. The nested-map edge collapse avoids any separator-collision footgun that a string-concatenated key would have introduced. The 32 tests cover the right surface — known-value fixtures, the dangling-mass canary, the bridge/connector heuristic, and the container error path.

One nit: `const maxRank = result.nodes[0].pageRank || 1` in GraphCentralityPanel.tsx would silently use 1 if pageRank ever comes back as 0 (unlikely but possible on a fully-isolated single node with only a self-loop). An explicit `result.nodes[0].pageRank > 0 ? result.nodes[0].pageRank : 1` makes the intent clear. Not blocking.

Overall this is a well-scoped read-only addition with zero new RPC surface and no changes to shared types or state. Clean once CI is green.

aashir-athar:
Thanks @coderabbitai — all three addressed in c7764bc1, verified against the code:

1. **Tie-aware bridge ranks (graphCentrality.ts)** — `findBridges` now computes **dense ranks on the metric values** (a new `denseRanksDesc` helper) for both PageRank and total degree, with the id tiebreak removed. Tied nodes share a rank, so a pure rename can no longer add/remove a connector flag — ranking is purely structural. Added a name-independence test (relabel every entity → identical gaps) and a tie test (mutual A↔B → no spurious bridge).

2. **Out-of-order responses (GraphCentralityTab.tsx)** — `load` now stamps a monotonic `latestRequestId` ref and ignores its result/error/loading updates if a newer load has started, so the mount-time `load('')` can't overwrite a later namespace-change result.

3. **Degree breakdown a11y (GraphCentralityPanel.tsx)** — the degree cell now carries an `aria-label` with the same `{in} in · {out} out` text (kept `title` for hover), so screen readers get the breakdown.

Local suite green: 34 centrality tests, `tsc --noEmit`, ESLint (0 errors), Prettier.

graycyrus (commented):
@aashir-athar the fix commit looks solid — race condition guard with the request ID ref is the right call, and switching bridge detection to dense ranks removes the name-ordering sensitivity that CodeRabbit flagged.

One leftover minor from my first pass that didn't make it into the fix commit: in `GraphCentralityPanel.tsx`, `const maxRank = result.nodes[0].pageRank || 1` will silently fall back to 1 if pageRank is exactly 0 (technically possible on a single isolated node). Prefer `result.nodes[0].pageRank > 0 ? result.nodes[0].pageRank : 1`. Not blocking, but clean it up when convenient.

Code looks good otherwise. I'll come back and approve once CI is fully green.

aashir-athar:
Heads-up on CI: the `E2E (Linux / Appium Chromium)` lane went red on the previous commit, but the failure was in the **Build E2E app** step — a cold-cache Rust/Tauri compile that hit a GitHub Actions cache-reservation rate limit ("You've hit a rate limit") and didn't finish — not a test assertion. This PR's changes are TypeScript-only (dense-rank bridge math, a request-token guard, one aria-label), the Rust build is identical to the prior green commit, all other lanes are green, and this same E2E lane is `success` on `main`. So it's an infra flake, not a regression.

I don't have rerun rights on the repo, so I pushed an empty commit to re-trigger the lane. If it flakes again, a maintainer re-run should clear it.

PR #2941, opened 2026-05-29
perf(memory_tree): batch chunk + score fetch in fetch_leaves

mysma-9403:
## Summary

`memory_tree::retrieval::fetch::fetch_leaves` previously did, inside a single `spawn_blocking`, a per-id pair of SQLite queries for every requested chunk: `get_chunk` and `get_score`. With the `MAX_BATCH = 20` cap on the call surface that meant up to **40 SQLite round-trips per citation request** on a hot read path (every "expand chunk_ids into full bodies + scores" pass the LLM triggers when citing).

Replace the per-id pair with two upfront batched reads:

- `memory_store::chunks::store::get_chunks_batch(config, &[ids]) -> Result<HashMap<String, Chunk>>`
- `memory_tree::score::store::get_scores_batch(config, &[ids]) -> Result<HashMap<String, f32>>`

`fetch_leaves` now issues one `SELECT … WHERE id IN (?,?,…)` per table (so **2** round-trips total at the cap), then walks the input id slice and looks each id up in the two maps. File I/O (`content::read::read_chunk_body`) stays per-id — each chunk's full markdown body lives in its own on-disk file, so batching there would mean concurrent file opens, not a single round-trip; left untouched.

### Defensive chunking — `MAX_FETCH_BATCH = 500`

Both batch helpers iterate via `chunks(MAX_FETCH_BATCH)`.

- The `fetch_leaves` call-site cap of 20 ids never trips the window — runs as one query today.

- `get_chunks_batch` reuses `row_to_chunk` so decoded `Chunk` values stay identical to the per-row path.
- `get_scores_batch` is intentionally narrow — returns only `chunk_id -> total` because that is all `fetch_leaves` reads from `ScoreRow`. No need to decode the full signals struct.
- Missing ids are silently absent from either map, mirroring the per-row `get_chunk` / `get_score` `Ok(None)` contract — so the existing `hits.len() < ids.len()` partial-result detection keeps working unchanged.
- Input order preserved by walking the caller's id slice and doing per-id map lookups (HashMap doesn't preserve insertion order — the explicit walk does).
- `read_chunk_body` failure still falls back to the ≤500-char preview just like before.

### New unit tests (5 total)

**chunks side** (`src/openhuman/memory_store/chunks/store_tests.rs`):
- [x] `get_chunks_batch_returns_present_ids_in_map`
- [x] `get_chunks_batch_empty_input_and_missing_ids`

**score side** (`src/openhuman/memory_tree/score/store_tests.rs`):
- [x] `get_scores_batch_returns_present_chunk_ids`
- [x] `get_scores_batch_empty_input_and_missing_chunk_ids`

**`fetch_leaves`** (`src/openhuman/memory_tree/retrieval/fetch.rs`):
- [x] `fetch_leaves_preserves_input_order_and_propagates_scores` — interleaved present/missing input order + distinct per-chunk totals to prove the per-id HashMap lookup keys by chunk_id (not by iteration index), and to pin the 0.0 fallback when a chunk has no score row.

- [x] All 5 existing `memory_tree::retrieval::fetch::tests` pass.
- [x] Broader sweep — **764 tests across `memory_store::` + `memory_tree::retrieval::` + `memory_tree::score::`** — 0 failures.
- [x] `cargo fmt` — clean.

## Companion to #2927

Same pattern as the topic-rerank batch refactor (#2927): partition / batch / decorate. Different table, different call-site, same playbook.

## Summary by CodeRabbit

* **New Features**
  * Added batched retrieval functions for chunks and scores.

* **Performance**
  * Memory retrieval operations now use batched database reads instead of individual queries, improving efficiency for large datasets.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2941?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@mysma-9403 hey! the code looks good to me — clean batch refactor, well-reasoned, no correctness issues. CI is still pending so i'll hold off on approving until those are green. once they pass i'll come back and approve.

---

**What this does**: replaces 2N per-id SQLite queries in `fetch_leaves` with two batched `SELECT … WHERE id IN (…)` calls, dropping from up to 40 round-trips (at the MAX_BATCH=20 cap) to exactly 2. The defensive `MAX_FETCH_BATCH=500` chunking is smart — keeps things clean for future callers that might pass larger lists without hitting SQLite's variable limit.

**A few observations** (none blocking):

- `MAX_FETCH_BATCH = 500` is duplicated in `chunks/store.rs` and `score/store.rs`. That's fine for now since it mirrors a fixed SQLite compile-time constraint, but if you ever need to tune it you'll have two places to change. Could live in a shared constants module, but not worth a refactor just for this.

- In `fetch.rs`, `chunk.clone()` on the now-borrowed `&Chunk` clones the `content` string (≤500-char preview) even though the success branch of `read_chunk_body` immediately overwrites it. Effectively free at 20 ids, just noting it since it wasn't in the original path.

- The test coverage is solid: ordering + score propagation + 0.0 fallback are all pinned, and the empty/missing-id edge cases are explicit. Good test for the ghost-id dropping behaviour too — that one's easy to accidentally break in a batch refactor.

Overall the approach is correct, semantics are preserved, and the PR description is thorough. Will approve once CI is green.

src/openhuman/memory_tree/score/store.rs
```diff
@@ -169,6 +171,56 @@ pub fn get_score(config: &Config, chunk_id: &str) -> Result<Option<ScoreRow>> {
...
 }
 
+/// Defensive cap for batched `IN (?,?,…)` reads.
+///
+/// See identical rationale on [`crate::openhuman::memory_store::chunks::store`]'s
+/// `MAX_FETCH_BATCH`: SQLite's `SQLITE_MAX_VARIABLE_NUMBER` has been
+/// 32 766 since 3.32, so 500 leaves a ~65× safety margin. The
+/// `fetch_leaves` call-site is capped at 20 ids so the chunked loop
+/// runs exactly once today — the window exists purely so future
+/// call-sites passing larger id lists do not blow up against a host
```
senamakel, line 181: Re: deduplicating `MAX_FETCH_BATCH` across `chunks/store.rs` and `score/store.rs` (also raised by @graycyrus) — leaving the two constants separate intentionally. Each mirrors a fixed SQLite compile-time constraint (`SQLITE_MAX_VARIABLE_NUMBER`), not a tunable shared with the other domain, and the rationale doc-comment in `score/store.rs` already cross-references this module. Extracting a shared symbol would couple two otherwise-independent store layers for a value that has no reason to drift in lockstep. Documenting the divergence rather than centralizing, per CodeRabbit's "if intentional, leave a note" guidance.

src/openhuman/memory_tree/retrieval/fetch.rs
```diff
@@ -71,7 +81,7 @@ pub async fn fetch_leaves(config: &Config, chunk_ids: &[String]) -> Result<Vec<R
             // The `content` column in SQLite holds a ≤500-char preview after
             // the MD-on-disk migration; the retrieval API must return the
             // complete chunk text so the LLM sees untruncated content.
-            let mut chunk_with_body = chunk;
+            let mut chunk_with_body = chunk.clone();
```
senamakel, line 85: Re: @graycyrus's note that `chunk.clone()` clones the ≤500-char `content` preview that the `read_chunk_body` success branch immediately overwrites — leaving as-is. `chunk_by_id` is a `HashMap` the loop only reads from, so we need an owned `Chunk` to mutate; the clone is the cheapest correct option at the `MAX_BATCH=20` cap (effectively free, as noted). Restructuring to avoid the preview-string clone (e.g. `remove()` from the map to take ownership) would trade a trivial allocation for fragility around the partial-result/order-preservation contract, so it isn't worth it here.

PR #2942, opened 2026-05-29
feat(intelligence): add Entity Associations

aashir-athar:
## Summary

Adds a new read-only **"Associations"** tab that surfaces the most strongly-related entity *pairs* in the memory graph by **Jaccard similarity of their connection sets** — `jaccard(a,b) = |N(a) ∩ N(b)| / |N(a) ∪ N(b)|`.

The most useful result is a high-similarity pair that is **not directly linked**: the graph already "knows" two entities belong together (they share most of the same connections) even though no single fact connects them — a relationship a direct-edge view can never show. Each pair is tagged `inferred` (no direct link) or `linked`.

### Design
- **Pure & deterministic** (`lib/memory/entityAssociations.ts`): no clock, no RNG. Neighbourhoods are undirected (an association is symmetric) and self-loops are dropped. Candidate pairs come from an **inverted index** (for each neighbour, the entities adjacent to it), so only pairs that actually share a neighbour are scored — far cheaper than all-pairs on a sparse graph. Pair keys are canonical JSON (smaller id first), so entities containing any character can't collide. Output is ranked `jaccard` desc with stable `sharedCount`/id tie-breaks, capped with a `truncated` flag.
- **Zero new core surface**: reuses the already-shipped `memoryGraphQuery` and `memoryListNamespaces` wrappers. **Read-only** — recomputed from the live graph on mount, never persisted.
- **Container/presentational split**; the container guards against out-of-order load responses with a monotonic request token from the start; namespace selector; metric tiles + a ranked pair list with linked/inferred badges and a truncation note. i18n across all 13 locales.

### Edge cases (tested)
Empty graph, fewer-than-two entities, single edge (no shared-neighbour pair), self-loops ignored, direction treated symmetrically, `minShared` threshold, output cap + truncation flag, relation-order invariance (determinism), malformed rows dropped, and entities containing spaces (no key collision).

## Test plan
- [x] `vitest` — 23 tests (engine: twins/triangle/self-loop/symmetry/minShared/cap/determinism/malformed/collision; api facade; panel; container)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors
- [x] i18n coverage gate — EXIT 0, no unused `entityAssociations.*` keys

## Summary by CodeRabbit

* **New Features**
  * Added an "Associations" tab to Intelligence with an Entity Associations view: metrics tiles, namespace selector, ranked "Strongest associations" list with linked vs inferred badges, tooltips, progress bars, retry action, and truncated-results summary.
  * Introduced a deterministic association scoring engine and service endpoints powering the UI.

* **Internationalization**
  * Added translations for the associations UI across many languages.

* **Tests**
  * Added unit and UI tests covering scoring, rendering states, namespace filtering, retry behavior, and truncation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2942?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/components/intelligence/EntityAssociationsPanel.tsx
```diff
@@ -0,0 +1,190 @@
...
+          {t('entityAssociations.rankedHeading')}
+        </h3>
+        <ul className="space-y-1.5">
+          {report.pairs.map(pair => (
+            <li
+              key={`${pair.a}${pair.b}`}
+              className="rounded-lg border border-stone-200 dark:border-neutral-800 px-3 py-2"
+              title={t('entityAssociations.pairTitle')
+                .replace('{jaccard}', String(pct(pair.jaccard)))
+                .replace('{shared}', String(pair.sharedCount))
```
graycyrus, line 136: **[minor]** Key built by string concatenation can collide for different entity pairs. `{a: "A", b: "BC"}` and `{a: "AB", b: "C"}` both produce `"ABC"`. The engine already uses `JSON.stringify([a, b])` for collision-free pair keys — use the same here:

```tsx
key={JSON.stringify([pair.a, pair.b])}
```

graycyrus (commented):
@aashir-athar solid work on this — the algorithm is correct, tests are thorough, and the component split is clean. one thing to fix before i approve, plus CI needs to finish.

**React list key collision**

`EntityAssociationsPanel.tsx` — the `<li>` key is built as `\`${pair.a}${pair.b}\``, which can silently collide when different pairs produce the same concatenation. For example, pair `{a: "A", b: "BC"}` and pair `{a: "AB", b: "C"}` both produce key `"ABC"`. The engine already has a collision-free canonical key — use the same pattern here:

```tsx
// before
key={`${pair.a}${pair.b}`}

// after
key={JSON.stringify([pair.a, pair.b])}
```

Everything else looks good:
- Jaccard engine: correct inverted-index approach, proper self-loop handling, deterministic sort
- Monotonic request token in the container guards against out-of-order responses correctly
- API facade adds no new RPC surface, logs only counts (no PII)
- 23 tests cover the edge cases described in the PR (empty graph, single edge, self-loops, minShared, cap/truncation, determinism, malformed rows, key collision in engine)
- i18n gate passes; English strings used as placeholders for non-English locales, which is fine for a first pass

Fix the key prop and once CI goes green i'll approve.

aashir-athar:
Good catch @coderabbitai and @graycyrus — fixed. The pair-row React key now uses `JSON.stringify([pair.a, pair.b])`, the same canonical encoding the engine uses for its pair keys, so distinct pairs can't collide. (Amusingly the engine already guarded this; I'd just used a naive concatenation for the list key.) `tsc`, ESLint, Prettier, and the panel tests are green.

graycyrus (commented):
@aashir-athar the key collision fix landed correctly in c44ce93 — `JSON.stringify([pair.a, pair.b])` matches the engine's canonical key approach. Code looks good to me. Waiting on CI (several checks still pending) before I can approve. Once those are green, this should be good to go.

app/src/lib/memory/entityAssociations.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (entityCount < 2) return { ...EMPTY_REPORT, entityCount };
+
+  // 2. Candidate pairs via inverted index: any two entities adjacent to the
+  //    same neighbour share it. De-dupe and count shared neighbours.
+  const tallies = new Map<string, PairTally>();
+  for (const adjacent of neighbours.values()) {
+    if (adjacent.size < 2) continue;
+    const members = [...adjacent].sort(compareIds);
+    for (let i = 0; i < members.length; i += 1) {
+      for (let j = i + 1; j < members.length; j += 1) {
```
senamakel, line 121: Re: CodeRabbit's nitpick that candidate generation here is quadratic in neighbour degree (a hub with N neighbours emits C(N,2) tallies).

Deferring intentionally rather than capping:
- This is inherent to inverted-index Jaccard candidate generation; the docstring already frames it as "far cheaper than all-pairs on a sparse graph," which is the right trade-off for a personal memory graph (bounded in practice).
- A degree cap would silently drop pairs from high-degree hubs, changing the ranked output — a correctness/semantics change no reviewer asked for. @graycyrus explicitly approved the algorithm as correct.
- If dense-graph jank is observed in the field, the right fix is to move the scoring off the main thread (or memoize per namespace), not to truncate candidates. Recording that here so it isn't lost.

No code change for this item.

PR #2946, opened 2026-05-29
feat(intelligence): add Connection Path finder

aashir-athar:
## Summary

Adds a new read-only **"Paths"** tab that traces the **shortest chain of relations** linking any two entities — answering *"how is A connected to B?"* with the actual hop-by-hop path, something a single direct-edge lookup can't show.

Pick two entities (type-ahead pickers populated from the graph) and the tab renders the path as a node chain, e.g. `You —[works_at]→ Acme ←[founded_by]— Dana`, with each hop's predicate and direction.

### Design
- **Pure & deterministic** (`lib/memory/connectionPath.ts`): BFS shortest path over the **undirected** graph (a relation connects its endpoints regardless of arrow direction), but each hop records the predicate and whether it was traversed **forward or backward**, so the chain reads naturally. No clock, no RNG — ties are broken by a fixed sorted neighbour order, so the same graph + endpoints always yield the same path. Self-loops are ignored, and distinct `reason`s cover same-entity / missing-endpoint / no-path.
- **Zero new core surface**: reuses the already-shipped `memoryGraphQuery` and `memoryListNamespaces` wrappers. The graph is fetched **once** per namespace; path-finding itself runs in the pure synchronous engine via `useMemo`, so tracing endpoints is instant with no extra round-trips. **Read-only**.
- **Container/presentational split**; the container guards the graph load with a monotonic request token; two `<datalist>` (type-ahead) entity pickers handle large graphs gracefully; namespace selector. i18n across all 13 locales.

### Edge cases (tested)
Forward chain, reversed-direction hops, direct one-hop link, same-entity, missing source/target, disconnected (no path), shortest-path tie-break determinism + relation-order invariance, self-loop ignored, malformed rows dropped.

## Test plan
- [x] `vitest` — 24 tests (engine: chains/direction/same/missing/no-path/tie-break/self-loop/direct/malformed; api facade; panel states; container load + trace + error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors
- [x] i18n coverage gate — EXIT 0, no unused `connectionPath.*` keys

## Summary by CodeRabbit

* **New Features**
  * Added a "Paths" tab to Intelligence to find and display shortest connection paths between entities, with namespace filtering, loading/error states, and retry.
  * Deterministic shortest-path computation with clear messages for same/missing/no-path cases and directional relation display.
* **Localization**
  * Full translations for the new UI across all supported languages.
* **Tests**
  * Added comprehensive tests covering UI, path engine, and API behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2946?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@aashir-athar hey! the code looks good to me, but CI is still pending (Frontend Unit Tests, Rust Core Tests, e2e, Build Tauri App). once those are green, i'll come back and approve this. let me know if you need any help!

One minor thing while I was in here: `ConnectionPathPanel.tsx` lines 199-207 use manual `.replace('{entity}', ...)` / `.replace('{source}', ...)` for i18n interpolation. This is fragile — if any translation omits a placeholder, the replacement silently no-ops. If `useT` supports parameterized keys (e.g. `t('key', { entity })`), use that instead. If not, not a blocker, but worth a follow-up.

app/src/services/api/connectionPathApi.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const log = debug('connection-path:api');
+
+export interface GraphData {
+  entities: string[]; // sorted, de-duplicated entity ids (for the pickers)
+  relations: GraphRelation[]; // raw triples (fed to the pure path engine)
+}
+
+/** Fetch the graph for a namespace (or all) and derive the sorted entity list. */
+export async function loadGraph(namespace?: string): Promise<GraphData> {
```
senamakel, line 28: Re: CodeRabbit's debounce/race suggestion for this module — addressed at the caller instead of here. `connectionPathApi.ts` is a thin stateless fetch wrapper, so adding an AbortController would change its exported signature for all callers. The actual race (rapid namespace switches applying out-of-order results) is now guarded in `ConnectionPathTab.tsx` via a `cancelled` flag in the effect cleanup, so stale responses are dropped. Flagging here for visibility.

PR #2948, opened 2026-05-29
perf(memory_tree): batch summary fetch in hydrate_summary_inputs

mysma-9403:
## Summary

`memory_tree::tree::bucket_seal::hydrate_summary_inputs` previously called `store::get_summary(config, id)?` once per id on every L≥1 seal — i.e. each time the buffer at level L≥1 reached its fan-in gate during ingest. With a typical seal fan-in of 5–20 summaries that was **5–20 SQLite round-trips per seal** on the hot ingest path; every higher-level (L1→L2, L2→L3, …) cascade re-paid the cost.

Replace the per-id loop with one upfront batched read:

- `memory_store::trees::store::get_summaries_batch(config, &[ids]) -> Result<HashMap<String, SummaryNode>>`

`hydrate_summary_inputs` now issues one `SELECT … WHERE id IN (?,?,…)` (so **1** round-trip per seal at the fan-in cap), then walks the input id slice and looks each id up in the map. File I/O (`content::read::read_summary_body`) stays per-id — each summary's full markdown body lives in its own on-disk file, so batching there would mean concurrent file opens, not a single round-trip; left untouched.

### Defensive chunking — `MAX_FETCH_BATCH = 500`

`get_summaries_batch` iterates via `chunks(MAX_FETCH_BATCH)`.

- The `hydrate_summary_inputs` call-site passes at most a single seal's fan-in (5–20 ids) → runs as one query today.

- `get_summaries_batch` reuses `row_to_summary` so decoded `SummaryNode` values stay identical to the per-row path.
- Missing ids are silently absent from the map, mirroring the per-row `get_summary` `Ok(None)` contract — so the existing `"missing summary {id} — skipping"` warn path keeps working unchanged.
- Input order preserved by walking the caller's `summary_ids` slice and doing per-id map lookups (`HashMap` doesn't preserve insertion order — the explicit walk does).
- `read_summary_body` failure still bubbles up with the same `with_context` wrap as before.

### New unit tests (3 total)

**trees store** (`src/openhuman/memory_store/trees/store_tests.rs`):
- [x] `get_summaries_batch_returns_present_ids_in_map` — both ids present in the map, decoded `SummaryNode` equal to the per-row path bit-for-bit (same `row_to_summary` decoder).
- [x] `get_summaries_batch_empty_input_and_missing_ids` — empty input → empty map (no SQL); missing id silently absent so callers keep their `Ok(None)` skip path.

**bucket_seal** (`src/openhuman/memory_tree/tree/bucket_seal_tests.rs`):
- [x] `hydrate_summary_inputs_batch_preserves_order_and_skips_missing_ids` — interleaved present/missing input order `[sum-b, ghost, sum-a]` + distinct per-summary values (content, token_count, score, entities) to prove the per-id HashMap lookup keys by id (not by `enumerate()` position) and to pin the missing-id silent-skip contract.

- [x] All existing tests across `memory_store::trees::` and `memory_tree::tree::` continue to pass — **163 tests, 0 failures**.

## Companion to #2927 / #2941

Same playbook as the topic-rerank batch refactor (#2927) and the `fetch_leaves` chunk+score batch (#2941): partition / batch / decorate. Different table, different call-site, same pattern.

## Summary by CodeRabbit

  * Optimized summary data retrieval to support batched operations for significantly improved fetch efficiency across multiple requests
  * Enhanced system robustness by gracefully handling missing summaries during batch retrieval operations

* **Tests**
  * Added comprehensive test coverage for the new batch retrieval functionality
  * Verified correct handling of edge cases including missing data and request ordering preservation

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2948?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


graycyrus (commented):
@mysma-9403 hey! the code looks good to me — solid batch refactor, well-scoped and well-tested. CI has a few checks still pending (Rust Core Coverage, Rust Tauri Coverage, Frontend Coverage, Build Tauri App, E2E Appium suite, Windows Rust Core tests). once those are green i'll come back and approve this. let me know if you need any help with the CI runs.

PR #2976, opened 2026-05-29
perf(memory_tree): batch tree fetch in flush_stale_buffers

mysma-9403:
## Summary

`memory_tree::tree::flush::flush_stale_buffers` previously called `store::get_tree(config, &buf.tree_id)?` once per stale buffer returned by `list_stale_buffers` — i.e. **one SQLite round-trip per stale row** before the cascade even started. On hosts with many stale L0 buffers (e.g. after a long backfill, a cron miss, or a process restart during ingest) every flush pass re-paid that cost.

Replace the per-buffer loop with one upfront batched read:

- `memory_store::trees::store::get_trees_batch(config, &[ids]) -> Result<HashMap<String, Tree>>`

`flush_stale_buffers` now dedups buffer `tree_id`s into a `HashSet`, issues one `SELECT … WHERE id IN (?,?,…)` (so **1** round-trip), then per-id `HashMap` lookup routes each buffer to its own `Tree` for `cascade_all_from`. The `level = 0` filter in `list_stale_buffers` keeps the same `tree_id` at most once today, but the dedup is defensive against future widenings of that filter that would otherwise re-pay the cost.

### Defensive chunking — `TREES_MAX_FETCH_BATCH = 500`

`get_trees_batch` iterates via `chunks(TREES_MAX_FETCH_BATCH)`.

- The `flush_stale_buffers` call-site passes the deduped distinct `tree_id` set (typically << 500 in practice) → runs as one query today.

- `get_trees_batch` reuses `row_to_tree` so decoded `Tree` values stay bit-identical to the per-row path.
- Missing ids stay silently absent from the `HashMap`, mirroring the per-row `get_tree` `Ok(None)` contract — so the existing orphan-buffer warn-and-skip path keeps working unchanged.
- Per-buffer iteration order preserved by walking `stale` directly; `HashMap` lookup keys by `&buf.tree_id` (not by `enumerate()` position).
- `cascade_all_from` failure still bubbles up with the same error semantics as before.

### New unit tests (3 total)

**trees store** (`src/openhuman/memory_store/trees/store_tests.rs`):
- [x] `get_trees_batch_returns_present_ids_in_map` — both ids present in the map, decoded `Tree` equal to the per-row path bit-for-bit (same `row_to_tree` decoder).
- [x] `get_trees_batch_empty_input_and_missing_ids` — empty input → empty map (no SQL); missing id silently absent so callers keep their `Ok(None)` skip path.

**flush** (`src/openhuman/memory_tree/tree/flush.rs`):
- [x] `flush_seals_multiple_distinct_trees_via_batched_lookup` — two real trees each carry a stale L0 buffer; one `flush_stale_buffers` pass seals both via the batched fetch. Pins the **HashMap-keyed-by-id** contract: swapping the lookup key to `enumerate()` position would route buffers to the wrong `Tree` and mis-write summaries against the wrong `root_id` — the test asserts `count_summaries(tree_a) == 1 && count_summaries(tree_b) == 1`.

- [x] All existing tests in `memory_store::trees::` and `memory_tree::tree::flush::` continue to pass (25 + 4 = 29 tests, 0 failures).

## Companion to #2927 / #2941 / #2948

Same playbook as the topic-rerank batch (#2927), `fetch_leaves` batch (#2941), and `hydrate_summary_inputs` batch (#2948): partition / batch / decorate. Different table, different call-site, same pattern. `get_trees_batch` is a building block several remaining N+1 hot-paths in the graph layer will reuse.

## Release Notes

* **Performance**
  * Optimized memory tree operations by consolidating multiple database queries into single batch requests, reducing round-trips and improving efficiency when processing multiple trees.

* **Tests**
  * Added regression test coverage for bulk tree flushing operations to ensure correct behavior across multiple trees.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2976?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oxoxDev (dismissed):
LGTM — clean, behavior-preserving perf fix. `get_trees_batch` replaces the per-id `get_tree` loop in `flush_stale_buffers` with one batched `SELECT … WHERE id IN (…)` per 500-id window, mirroring the existing `get_chunks_batch`/`get_scores_batch` pattern. De-dups tree_ids defensively, and missing rows stay absent from the map so the orphan-buffer warn-and-skip contract is preserved bit-for-bit. Tests cover present/empty/missing. Approving.

staimoorulhassan (commented):
## CodeRabbit-style Review — PR #2976

### Walkthrough
Replaces N per-buffer `get_tree()` SQLite calls in `flush_stale_buffers` with one batched `SELECT … WHERE id IN (…)` via new `get_trees_batch()`. Adds 3 targeted unit tests. Companion to the batch-fetch series (#2927, #2941, #2948).

---

### Findings

#### 🟡 MINOR — Dynamic SQL construction via `format!()` — add a safety comment
`get_trees_batch` builds the `WHERE id IN ({placeholders})` clause with `format!()`. The values are correctly bound via typed `rusqlite::ToSql` params (no interpolation), but a future maintainer who doesn't notice this distinction could be misled into thinking the code is SQL-injectable. A one-line comment is all that's needed:

```rust
// placeholders are positional bind params (?1,?2,…) — values are never interpolated
let sql = format!(
    "SELECT … WHERE id IN ({placeholders})"
);
```

---

#### 🟡 MINOR — Prefer `params_from_iter` over `Vec<&dyn rusqlite::ToSql>`
The current approach:
```rust
let params: Vec<&dyn rusqlite::ToSql> =
    window.iter().map(|s| s as &dyn rusqlite::ToSql).collect();
let rows = stmt.query_map(params.as_slice(), row_to_tree)?;
```
is valid but fragile (raw trait-object lifetime). The idiomatic rusqlite pattern for dynamic params is:
```rust
let rows = stmt.query_map(
    rusqlite::params_from_iter(window.iter().map(String::as_str)),
    row_to_tree,
)?;
```
`params_from_iter` was introduced specifically for this use-case and avoids the manual `&dyn` dance.

---

#### 🟡 MINOR — Document that the dedup is for DB efficiency, not semantic ordering
```rust
let distinct_tree_ids: Vec<String> = {
    let mut seen = std::collections::HashSet::new();
    // …
};
```
The insertion-order-preserving dedup is correct and deterministic (because `list_stale_buffers` returns rows in `ORDER BY` order). But a future reader simplifying this to `.sort().dedup()` would still be correct *today* — the per-buffer loop walks `stale` directly for ordering, not `distinct_tree_ids`. A short comment prevents the "why not just sort+dedup?" question:

```rust
// Preserve list_stale_buffers order so the HashMap lookup below
// stays predictable; dedup is purely to avoid redundant DB params.
```

---

#### 🔵 NITPICK — Consider splitting the large flush regression test
`flush_seals_multiple_distinct_trees_via_batched_lookup` does two things: verifies that both trees are sealed (happy path) and pins the HashMap-keyed-by-id contract (regression). Splitting into:
- `flush_seals_multiple_trees_batched` — happy path
- `flush_keyed_by_tree_id_not_position` — regression assertion

would make the test intent clearer and failure messages more diagnostic.

---

### Verdict
No blockers. Four minor/nitpick comments. The algorithm, chunking strategy, semantic-preservation argument, and test coverage are all excellent — this is a well-engineered perf improvement. Approve once comments are considered.

PR #2978, opened 2026-05-29
feat(intelligence): add Graph Cohesion

aashir-athar:
## Summary

Adds a new read-only **"Cohesion"** tab to the Intelligence view: **clustering-coefficient** analysis of the memory knowledge graph. Where the Centrality lens answers *"which entities are important"*, this lens answers *"how tightly knit is the neighbourhood around each entity"* — and surfaces a structural signal neither centrality nor a frequency sort can reveal: **brokers**.

A **broker** (structural hole) is an entity whose neighbours are *not* connected to each other — it's the sole link holding otherwise-separate clusters together (local clustering ≈ 0). These are the single points of failure / brokerage opportunities in the user's accumulated memory.

### Design
- **Pure deterministic engine** (`lib/memory/graphCohesion.ts`): treats the `(subject)-[predicate]->(object)` triples as an **undirected simple graph** (direction dropped, parallel edges collapsed, self-loops dropped) and computes:
  - per-node **local clustering coefficient** `C(v) = 2·triangles / (deg·(deg-1))`,
  - **triangleCount**, **averageClustering** (mean of `C` over degree≥2 nodes), and **transitivity** (the global clustering coefficient, `3·triangles / connected-triples`),
  - `findBrokers()` — ranks the loosest-neighbourhood entities.
  - No React, no RPC, no clock, no RNG. The result depends only on graph structure, never on insertion order — and `averageClustering` is summed in canonical (sorted) order so it is **byte-identical across input permutations** despite IEEE-754 non-associativity.
- **Zero new core surface**: composes the already-shipped `memoryGraphQuery` / `memoryListNamespaces` JSON-RPC wrappers. **Read-only** — recomputed live from the graph, never persisted.
- **Container/presentational split**; the container guards load-on-mount with a monotonic request token (out-of-order responses can't overwrite the latest). i18n across all 13 locales.

### Test plan
- [x] `vitest` — 33 tests (engine: empty/triangle/path/4-cycle/star/diamond fixtures with hand-computed clustering, triangle, avg & transitivity values; self-loop drop; parallel-edge & direction collapse; malformed-row drop; no case-folding; **byte-identical determinism across permutations**; broker ranking & limits — plus api facade, panel states + broker badge + no-brokers note, container load/namespace-requery/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors

## Summary by CodeRabbit

* **New Features**
  * Added a "Cohesion" tab to Intelligence showing graph cohesion analytics.
  * Shows metrics: entities, connections, triangles, average clustering, and transitivity.
  * Ranks "Brokers" (structural-hole entities) with degree and clustering details.
  * Namespace selector to filter cohesion analysis with retry on failure.

* **UX**
  * Loading skeletons, clear empty-state messaging, and error alerts with Retry.

* **Localization**
  * New translation keys for the Cohesion UI across locales.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2978?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


staimoorulhassan (commented):
## CodeRabbit-style Review — PR #2978

### Walkthrough
Adds a read-only "Cohesion" tab to the Intelligence view — clustering-coefficient analysis of the memory knowledge graph. Pure frontend engine (`graphCohesion.ts`), no new RPC surface. 33 tests covering engine, API facade, and UI states.

---

### Findings

#### 🟠 MAJOR — Cross-platform floating-point non-associativity
The PR notes `averageClustering` is "summed in canonical (sorted) order so it is **byte-identical across input permutations**." This holds within a single CPU architecture, but **not across architectures** — x86-64 and ARM (Apple Silicon, iOS) may produce different FP results for the same sum due to differing FPU rounding modes and instruction selection. If the metric is displayed in the UI (it is — the caption shows `averageClustering`), users on different devices will see silently different values.

**Suggested fix:** Round to a fixed number of decimal places before display or before storing the value in the return object:
```ts
averageClustering: Math.round(sum / count * 1e6) / 1e6,
```
Alternatively, store the intermediate integer numerator/denominator as a rational and divide only at display time.

---

#### 🟡 MINOR — `findBrokers()` has no default result cap
```ts
findBrokers(result: CohesionResult, limit?: number): BrokerEntry[]
```
If `limit` is omitted, the full ranked entity list is returned. For a memory graph with thousands of nodes this materializes a large array in the browser main thread before the UI slices it. Add a sensible default cap:
```ts
findBrokers(result: CohesionResult, limit = 100): BrokerEntry[]
```

---

#### 🟡 MINOR — Verify monotonic request token uses strict equality
The container pattern guards against stale responses with a monotonic token. Ensure the comparison uses `===` (not `>=`):
```ts
if (token !== currentToken) return; // ✓ strict
// vs.
if (token >= currentToken) return;  // ✗ allows a fast burst to overwrite a slow one
```
A `>=` comparison can incorrectly drop valid responses when two requests fire in rapid succession.

---

#### 🟡 MINOR — i18n chunk vs flat-file inconsistency
New keys go into `app/src/lib/i18n/en.ts` (flat — correct) but the non-English translations only appear in `chunks/` files. CLAUDE.md states the chunked layout was retired. Please clarify whether `chunks/` is still the live layout, or move translations to the flat per-locale files. (This affects PRs #2975, #2978, #2980, and #2985 identically.)

---

#### 🔵 NITPICK — `GraphCohesionPanel.tsx` approaching the 200-line soft cap
At 198 lines it's at the edge. The broker table and the metric tiles are good candidates for extraction into sibling components to keep future growth manageable.

---

### Verdict
One major (cross-platform FP stability), two minors, one shared i18n blocker (see #2975 review). The pure-engine design, determinism argument, and test suite quality are exemplary. Resolve the FP rounding and i18n layout issues before merge.

app/src/lib/memory/graphCohesion.ts
```diff
@@ -0,0 +1,176 @@
...
+      clusterableCount += 1;
+    }
+  }
+
+  return {
+    nodes,
+    nodeCount: adjacency.size,
+    edgeCount: edgeDegreeSum / 2,
+    triangleCount: closedTripleSum / 3,
+    averageClustering: clusterableCount === 0 ? 0 : clusteringSum / clusterableCount,
```
senamakel, line 155: Re: the MAJOR "cross-platform floating-point non-associativity" finding — I looked into this and don't think a change is warranted:

1. **JS arithmetic is architecture-deterministic.** ECMAScript pins `Number` to IEEE-754 binary64 with round-ties-to-even and *no* FMA/extended-precision contraction, so `a + b` on two doubles yields a bit-identical result on x86-64 and ARM (Apple Silicon/iOS). The x87-extended-precision / FMA-contraction divergence the comment describes is a C/C++ concern, not a JS one. The in-file determinism contract (canonical-order summation) already covers the only real source of drift — summand *order*.
2. **The displayed value is rounded anyway.** `GraphCohesionPanel.tsx` renders `averageClustering.toFixed(2)` (2 dp), so even a hypothetical sub-ULP difference could never surface to the user.

Rounding inside the engine return object would also conflict with the byte-identical-determinism unit tests. Leaving as-is — happy to revisit if you have a concrete repro of divergence.

app/src/lib/memory/graphCohesion.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * Broker candidates: nodes with degree >= 2 whose neighbours are the LEAST
+ * interconnected (lowest localClustering). A low-clustering high-degree node is
+ * a structural hole — it is the sole link between groups that would otherwise be
+ * disconnected. Sorted clustering ASC, then degree DESC (a bigger gap brokered
+ * matters more), then id ASC. Pure; derived entirely from the result.
+ */
+export function findBrokers(result: CohesionResult, limit = 25): CohesionNode[] {
```
senamakel, line 167: Re: the MINOR "`findBrokers()` has no default result cap" finding — this is already addressed in the current code: the signature is `findBrokers(result: CohesionResult, limit = 25)`, so an omitted `limit` caps the result at 25 rather than materializing the full ranked list. The review likely predates that default. No change needed.

PR #2985, opened 2026-05-29
feat(intelligence): add Graph Reach (eccentricity / diameter / radius)

aashir-athar:
## Summary

Adds a new read-only **"Reach"** tab to the Intelligence view: **BFS-based eccentricity analysis** of the memory knowledge graph. Where #2946's Connection-Path lens answers *"is A linked to B, and how far"*, this lens answers the **summary** question: *"how far is each entity from everything else it can reach"*. From that fall out three classic shape metrics:

- **Diameter** — the longest shortest-path anywhere in the graph,
- **Radius** — the smallest eccentricity,
- **Center** — the entities that reach the whole cluster in the fewest hops.

The center of your knowledge is **not** the highest-degree node nor the highest-PageRank node — it's the one that minimises the worst-case distance to everything else, the natural place a traversal or summary should start. Neither degree nor PageRank surfaces it.

### Design
- **Pure deterministic engine** (`lib/memory/graphReach.ts`): treats the `(subject)-[predicate]->(object)` triples as an **undirected simple graph** (direction dropped, parallel edges collapsed, self-loops dropped) and computes:
  - per-node **eccentricity** via BFS with an array-backed queue,
  - **connected-component** labelling (`componentId` = smallest member index, stable label),
  - per-component **diameter** & **radius**,
  - the **giant component** (size DESC, then smallest componentId for a stable tie-break),
  - per-node `isCenter` (`eccentricity === component's radius` — every component has a well-defined center).
  - Vertices are **canonically id-sorted** before indexing, so output is **byte-identical regardless of relation input order**. No clock, no RNG, all-integer math.
- **Zero new core surface**: composes the already-shipped `memoryGraphQuery` / `memoryListNamespaces` JSON-RPC wrappers. **Read-only** — recomputed live, never persisted.
- **Container/presentational split**; the container guards load-on-mount with a monotonic request token. The eccentricity bar in the ranked table is computed against each node's **own component's** diameter (so a node in a non-giant but longer component never overflows). The summary caption uses a separate singular key when `componentCount === 1`. i18n across all 13 locales.

### Test plan
- [x] `vitest` — 25 tests (engine: empty / triangle (uniform center) / path A-B-C-D (diameter 3, radius 2, center {B,C}) / star (hub is sole center) / two-component (giant diameter; small component's local ecc) / giant tie-break by smallest componentId; self-loop drop; parallel-edge & direction collapse; both branches of the malformed-row guard; no case-folding; **order-independent output across permutations** — plus api facade, panel singular + plural caption variants, center badge, container load/namespace-requery/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors

## Summary by CodeRabbit

* **New Features**
  * Added a Graph Reach tab in Intelligence showing entity centrality with metrics (entities, diameter, radius), ranked entities, eccentricity bars, link counts, center badges, namespace filtering, and retry/loading/empty states.
* **Tests**
  * Added unit and integration tests for the reach engine, API facade, and UI components covering loading, error, empty, and normal render paths.
* **Localization**
  * Added translation entries for Graph Reach across multiple languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2985?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


staimoorulhassan (commented):
## CodeRabbit-style Review — PR #2985

### Walkthrough
Adds a read-only "Reach" tab to the Intelligence view using BFS-based eccentricity analysis. Computes per-node eccentricity, connected components, diameter/radius per component, and center nodes. No new RPC surface. 25 tests. Companion to the graph-intelligence series (#2978, #2980).

---

### Findings

#### 🔴 BLOCKER — BFS is O(V·(V+E)) on the browser main thread with no budget cap
Per-node BFS is the right algorithm, but running it for every node in the main thread without a guard is dangerous at scale. A memory graph with 5,000 nodes and 20,000 edges ≈ 125M operations — enough to produce a noticeable freeze on a mid-range machine. The current implementation has no:
- Node-count gate (refuse to compute above N and show a graceful message)
- Web Worker offload
- Sampling approximation (e.g. double-sweep for diameter)

**Required before merge:** Add at minimum a node-count gate. A reasonable threshold for synchronous BFS in the main thread is ~500–1000 nodes. Above that, show a "Graph too large for reach analysis — try a smaller namespace" message:
```ts
const MAX_REACH_NODES = 1000;
if (nodes.size > MAX_REACH_NODES) {
  return { tooLarge: true };
}
```

---

#### 🟠 MAJOR — Eccentricity bar is normalized per-component, misleading cross-component comparisons
From the PR: *"The eccentricity bar in the ranked table is computed against each node's own component's diameter."* This means a node in a small 2-diameter component can show a 100% bar next to a node in the giant 20-diameter component with a 5% bar — even though the second node is "closer" in absolute terms. Users comparing entries across components will draw incorrect conclusions.

**Options:**
1. Normalize all bars to the global diameter (giant component's diameter) and add a tooltip with the local value
2. Add a visible label on each bar showing the component scope
3. Only show the bar for the giant component; show a raw number for nodes in smaller components

---

#### 🟡 MINOR — Giant-component tie-break can silently shift
"Giant component: size DESC, then smallest `componentId`" — `componentId` is assigned as the smallest member index (insertion-order-dependent). If two equal-size components grow at the same rate, the "giant" label can silently move between them. This is noted nowhere in the code. At minimum, document it in a comment as a known limitation:
```ts
// Tie-break by smallest componentId (= smallest member index in current sort order).
// If two components are equal-size, the "giant" label may shift as nodes are added.
```

---

#### 🟡 MINOR — i18n chunk vs flat-file inconsistency
Same as PRs #2975, #2978, #2980: new keys go to `chunks/` instead of flat per-locale files. CLAUDE.md states the chunked layout was retired. Clarify or migrate.

---

#### 🔵 NITPICK — Arabic plural forms: `=== 1` branch is incomplete
The singular/plural caption uses:
```ts
componentCount === 1 ? t('...singular') : t('...plural')
```
Arabic has 6 grammatical plural forms (zero, one, two, dual, few, many). A binary English-style branch will render the wrong form for Arabic counts of 0, 2, 3–10, and 11+. Consider using an `Intl.PluralRules`-based helper or at minimum document Arabic as a known limitation in a code comment.

---

### Verdict
One blocker (BFS budget cap required), one major (eccentricity bar normalization), two minors. The array-backed BFS queue (vs Set-shift) and center-node-per-component (not just giant) are both excellent design choices. Resolve the main-thread performance guard before merge.

staimoorulhassan (approved):
all review

sanil-23 (commented):
@aashir-athar the code itself looks good here, but CI needs to go green before I can approve.

The one failure that's actually on this PR is **Frontend Quality → Prettier**: the 13 non-English locale files (`ar, bn, de, es, fr, hi, id, it, ko, pl, pt, ru, zh-CN`) aren't prettier-formatted after the translations were added. Quick fix:

```
pnpm --filter openhuman-app exec prettier --write "src/lib/i18n/*.ts"
```

then commit. (`en.ts` passed, so it's just the translated locales.)

The other two red checks look unrelated to your change:
- **Rust Core Coverage** fails on `provider_admin_model_listing_covers_openrouter_validation_and_local_synthesis` (an inference-provider test) — this PR touches no Rust, so it reads as a pre-existing/flaky failure on `main`.
- **Playwright E2E Gate** reports `result=skipped` (the shards were skipped, not failed) — also not caused by this diff.

Worth a rebase on latest `main` to confirm the Rust one clears.

On the code: the engine is a clean read. Pure/deterministic with canonical id-sorting for reproducible output, array-backed BFS queue, per-component eccentricity so a node in a smaller-but-longer component is measured against its own component's diameter, and the singular-caption edge case is handled and tested. The API facade adding zero new core surface (just composing the existing `memoryGraphQuery` / `memoryListNamespaces` wrappers, read-only) is the right call. Test coverage is thorough.

Get the locale files formatted and the CI sorted and I'll come back for a proper approval. Let me know if you need a hand.

PR #2989, opened 2026-05-29
feat(intelligence): add Predicate Diversity (Shannon entropy of vocabulary)

aashir-athar:
## Summary

Adds a new read-only **"Vocabulary"** tab to the Intelligence view: information-theoretic analysis of the predicate distribution. Where every other lens analyses the **graph** (who points at whom — degree, PageRank, eccentricity, coreness, clustering, cuts), this one analyses the **vocabulary** (what relations the graph speaks in):

- How **varied** is the relation vocabulary the assistant has built — dozens of distinct predicates or just two or three?
- How **even** is the use of that vocabulary — does one predicate dominate, or are predicates used with roughly equal frequency?

Two graphs with the *same* number of relations can carry very different amounts of structural surprise. A graph with one predicate used 1000 times carries **0 bits** of vocabulary surprise per relation; a graph with 10 predicates used 100 times each carries **3.32 bits**. Neither degree, PageRank, eccentricity, coreness, clustering, nor articulation analysis surfaces this distinction.

### Design
- **Pure deterministic engine** (`lib/memory/predicateDiversity.ts`): per-predicate count + frequency, Shannon entropy `H = -Σ p_i · log2(p_i)`, max entropy `log2(D)` and normalised **evenness** `H / log2(D)`.
- **Float-order determinism** — the entropy sum walks predicates in their **canonical sorted order**, never in Map iteration order. Floating-point addition is not associative and the `p·log2(p)` summands are not exactly representable, so input-order summation would drift at the ULP level; canonical-order summation makes the entropy **byte-identical** for any permutation of the same graph.
- Self-loops, parallel edges, and multi-document evidence are **not** collapsed — this is a frequency analysis over relations, not over the simple-graph structure used by topology lenses.
- **Zero new core surface**: composes the already-shipped `memoryGraphQuery` / `memoryListNamespaces` wrappers. **Read-only** — recomputed live, never persisted.
- Container/presentational split; the container guards load-on-mount with a monotonic request token. i18n across all 13 locales with a singular/plural switch on the relations count.

### Test plan
- [x] `vitest` — 23 tests (engine: zero / single-predicate (H=0) / two-equal (H=1, evenness=1) / unbalanced 3+1 split (H ≈ 0.81127812445913, asserted to 12 decimals) / non-string predicate drop / parallel non-collapse / self-loop kept as a valid relation / no case-folding / **byte-identical entropy across permutations** (toBe, not toBeCloseTo) / ranked-list ordering — plus api facade, panel singular & plural caption variants, container load/namespace-requery/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors

## Summary by CodeRabbit

* **New Features**
  * Added a "Vocabulary" tab with Predicate Diversity UI: metrics (entropy, evenness, distinct predicates), ranked "Top predicates" list, namespace filter, loading/empty/error states, and retry support.

* **API**
  * Added endpoints to load diversity data and list namespaces for the UI.

* **Tests**
  * Added extensive tests for diversity computation, API wiring, UI states, namespace interactions, and retry behavior.

* **Internationalization**
  * Added predicate diversity translations for 14 languages.

### DIFF
--- a/app/src/components/intelligence/PredicateDiversityPanel.test.tsx
+++ b/app/src/components/intelligence/PredicateDiversityPanel.test.tsx
@@ -1,0 +1,74 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import { computePredicateDiversity } from '../../lib/memory/predicateDiversity';
+import type { GraphRelation } from '../../utils/tauriCommands/memory';
+import PredicateDiversityPanel from './PredicateDiversityPanel';
+
+function rel(subject: string, predicate: string, object: string): GraphRelation {
+  return {
+    namespace: 'n',
+    subject,
+    predicate,
+    object,
+    attrs: {},
+    updatedAt: 0,
+    evidenceCount: 1,
+    orderIndex: null,
+    documentIds: [],
+    chunkIds: [],
+  };
+}
+
+// 4 relations: 3 'knows' + 1 'trusts' -> entropy ~0.811 bits, evenness ~81%.
+const unbalanced = computePredicateDiversity([
+  rel('A', 'knows', 'B'),
+  rel('A', 'knows', 'C'),
+  rel('A', 'knows', 'D'),
+  rel('A', 'trusts', 'E'),
+]);
+
+describe('<PredicateDiversityPanel />', () => {
+  it('renders the loading skeleton', () => {
+    render(<PredicateDiversityPanel result={null} loading />);
+    expect(screen.getByTestId('predicate-diversity-loading')).toBeInTheDocument();
+  });
+
+  it('renders the empty state when there are no relations', () => {
+    render(<PredicateDiversityPanel result={computePredicateDiversity([])} />);
+    expect(screen.getByText('No knowledge graph yet.')).toBeInTheDocument();
+  });
+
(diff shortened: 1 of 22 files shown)


sanil-23 (commented):
@aashir-athar Really clean, well-structured contribution. The entropy engine is genuinely pure and deterministic, the canonical predicate-sorted summation to keep the result float-stable across input orderings is a thoughtful touch, and the container/presentational split with the monotonic request-token guard against out-of-order loads is exactly the right pattern. Engine test coverage is thorough.

A couple of things before I can approve:

1. CI is currently red (Frontend Quality - typecheck/lint/format, and the Playwright E2E gate). Those need to go green first; once they are, I'll come back and do the final pass and approve. If the E2E gate looks flaky rather than a real failure, let me know and I'll help dig in.

2. Minor, non-blocking: the new keys are added to every locale file, but the non-English files carry the English strings verbatim (e.g. ar.ts and zh-CN.ts have 'Vocabulary', 'Distinct predicates', etc. in English). The surrounding keys in those files are localized, so these particular strings will render untranslated. That's fine if the intent is English-fallback-now / translate-later, but worth confirming it matches the project's i18n policy, since the description frames it as translations for 14 languages.

Please also resolve the open CodeRabbit comment. Fix CI and I'll approve right after.

PR #3000, opened 2026-05-30
feat(intelligence): add Evidence-Weighted Trust Lens

aashir-athar:
## Summary

Adds a new read-only **"Trust"** tab to the Intelligence view. The 20 shipped intelligence lenses all read the `(subject, predicate, object)` triple structure or its document provenance. None of them makes **`evidenceCount`** — the field that measures how many times an assertion has been independently corroborated — the **primary** signal.

This one does.

### Three orthogonal-to-everything-shipped outputs

- **Trust Quotient** per entity = `W(x) / D(x)` — the **mean evidence per relation** an entity is involved in. A prolific entity with many single-witness facts has low TQ; a quiet entity whose few relations are heavily corroborated has high TQ. Degree alone (centrality, k-core) cannot express this.
- **Evidence Gini** = inequality of evidence concentration across entities, via the closed form `Σ_i (2i − n − 1)·w_i / (n·Σw)` on sorted entity weights — one integer-weighted sum, one division, **no float-associativity hazard**.
- **Under-corroborated worklist** = relations whose evidence falls below `max(1, floor(median_positive_evidence / 4))` — a **curator's review queue**.

Plus **Predicate Reliability Index** per predicate and a **`degraded: true`** banner when every relation has `evidenceCount = 1` (the field was never populated meaningfully), so the UI can show an empty-state instead of misleading the user with a constant TQ everywhere.

### Provenance & design discipline
Picked by the **loop-21 judge-panel design workflow** (4 fresh angles — chunks / evidence / composite / completion — × 3 independent judges + synthesis). Weighted score **8.633 / 10**, beating Chunk Corroboration Atlas (8.083), Inferred Edges (8.342), Attention Quadrants (7.767). All eight synthesis-stage refinements applied during build (Number coercion, n<2 guard, multigraph SUM policy, index-pick median, degraded flag, etc).

Two engine bugs were caught during the build: the **Gini denominator** was using `totalEvidence` instead of the sum of entity weights (= 2·totalEvidence) — fixed; and a **JS default-parameter quirk** in the test helper that ate the `undefined` argument — fixed to use `null`.

The subsequent adversarial review caught 2 minors (9 raw → 2 confirmed) — an **empty-string namespace** rendering as `()` in the UI (fixed: tightened guard to `.length > 0`) and a **self-loop comment** that framed the dual-credit as edge-case rather than the general undirected-degree rule (reframed).

### Engine design (pure, deterministic — no React/RPC/clock/RNG)
- `evidenceCount` coerced via `Number(...)` (handles string-encoded JSON numerics, scientific notation) then sanitised: non-finite/NaN/non-positive → 0; positive values floored to integers so all subsequent sums are exact integer arithmetic.
- **Multigraph collapse**: the same `(namespace, s, p, o)` tuple appearing multiple times SUMS evidence; counts as **one** distinct relation.
- Entity identity = `(namespace, name)` — orthogonal to Entity Duplicates (#8). A name appearing in two namespaces is two rows.
- Median for the under-corroboration threshold uses **index-pick** (lower middle on even-sized sets), never the arithmetic mean — preventing float drift.
- Guards: `globalGini = 0` when `entityCount < 2` OR `Σw === 0` (OR, not AND — `&&` would crash on an all-NaN graph).

### Zero new core surface
Composes the already-shipped `memoryGraphQuery` / `memoryListNamespaces` JSON-RPC wrappers. Container/presentational split with a monotonic request-token race guard for load-on-mount; i18n across all 13 locales; `aria-hidden` on the decorative `→` glyph paired with an `sr-only` "points to" for screen-reader semantics.

### Test plan
- [x] `vitest` — 34 tests (engine: 4 design-pick hand fixtures + Gini equality/max-inequality/all-zero/guards + threshold index-pick + multigraph evidence SUM + namespace orthogonality + sanitisation (NaN/Infinity/negative/null/string/scientific notation/fractional) + degraded mode + order-independent byte-identical output — plus api facade, panel metric tiles + trust ranking + predicate reliability + worklist + degraded banner + no-worklist caption, container load/namespace-requery/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors

## Summary by CodeRabbit

* **New Features**
  * Added a Trust tab to Memory UI with loading skeleton, retryable error alerts, namespace selector, metric tiles (Evidence Gini, weighted entities, total evidence), Entity Trust Quotient ranking, predicate reliability list, degraded-state banner, and an under-corroborated relations worklist with evidence labels.

* **Internationalization**
  * Added translation keys for the Trust tab and related UI across many locales.

* **Tests**
  * Comprehensive tests added for computation logic, API surface, and UI behaviors.


app/src/components/intelligence/EvidenceTrustTab.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * <EvidenceTrustPanel>. Read-only — the result is recomputed from the live
+ * graph, never persisted.
+ */
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+import { useT } from '../../lib/i18n/I18nContext';
+import type { EvidenceTrustResult } from '../../lib/memory/evidenceTrust';
+import { loadEvidenceTrust, loadNamespaces } from '../../services/api/evidenceTrustApi';
+import EvidenceTrustPanel from './EvidenceTrustPanel';
+
```
sanil-23, line 13: **[major]** This container is never mounted anywhere. `app/src/pages/Intelligence.tsx` — where every other lens tab is registered (import + `IntelligenceTab` union + tab-button list + `activeTab === '...'` render line) — isn't part of this PR's diff, so `EvidenceTrustTab` has no caller. The feature is dead code until those four spots are added. Worth a render test at the page level too, so a missing tab registration fails CI next time.

sanil-23 (commented):
@aashir-athar heads up — CI is currently failing, so I'll hold a full approval until it's green. A couple of things stood out:

**1. The Trust tab is never wired into the UI (blocking).** `app/src/pages/Intelligence.tsx` isn't modified by this PR, so `EvidenceTrustTab` is never imported or rendered and there's no tab button for it — the entire feature is currently unreachable from the app. The sibling lenses register in four places in `Intelligence.tsx`: the component import, the `IntelligenceTab` union type, the tab-button list, and the `activeTab === '...' && <Tab />` render block. The Trust tab needs all four before it's actually a tab.

**2. CI is red on the format check.** Prettier flags the 13 non-English i18n files you added keys to (ar, bn, de, es, fr, hi, id, it, ko, pl, pt, ru, zh-CN). `en.ts` is clean, but the translations weren't run through Prettier — `pnpm prettier --write app/src/lib/i18n` (or your `format` script) clears it.

CodeRabbit's two points (the promise chain in the effect, and the raw namespace in the debug log) are worth folding in at the same time.

Credit where due: the engine itself is genuinely clean — pure and deterministic, the closed-form Gini and the index-pick median avoid the float-associativity traps, and the test coverage on the math is thorough. Once the tab is wired into Intelligence.tsx and CI goes green, ping me and I'll do a full pass.

PR #2967, opened 2026-05-29
feat(intelligence): add Graph Export

aashir-athar:
## Summary

Adds a new read-only **"Export"** tab for **data portability**: download the entire memory knowledge graph as **JSON or CSV**. A different *kind* of feature from the analytics lenses — it's about giving the user ownership of their data ("your memory, yours to keep").

### Design
- **Pure deterministic serializers** (`lib/memory/graphExport.ts`): `toJsonExport` and `toCsvExport` over a stable `{subject, predicate, object, namespace, evidenceCount, updatedAt}` projection (malformed rows dropped). The CSV follows **RFC 4180** — a field is quoted when it contains a comma, double-quote, CR, or LF; embedded double-quotes are doubled; rows are CRLF-terminated — so entity names with commas/quotes/newlines round-trip safely.
- **Zero new core surface**: reuses **only** `memoryGraphQuery`. **Read-only** — the download itself (Blob + object URL + anchor) happens in the container's **click handler**, never during render, so the engine stays pure.
- **Container/presentational split**; the container guards the load with a monotonic request token; the pure panel renders the export-ready count + a JSON/CSV format toggle + a download button. i18n across all 13 locales.

### Test plan
- [x] `vitest` — 19 tests (engine: projection/malformed-drop/null-namespace/non-finite normalization, JSON round-trip, **CSV RFC-4180 escaping** for commas/quotes/newlines + CRLF + empty-graph header, format dispatch + MIME; api facade; panel states + format toggle/download callbacks; container load + Blob-download trigger + error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors
- [x] `prettier --check` — clean
- [x] i18n coverage gate — EXIT 0, no unused `graphExport.*` keys

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added knowledge graph export functionality supporting JSON and CSV formats.
  * New export tab displays count of facts ready to export.
  * Download exported data as JSON or CSV files.
  * Error handling with retry option for failed exports.
  * Loading and empty state messaging included.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/2967?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3002, opened 2026-05-30
feat(intelligence): add Graph Completion Hints (Adamic-Adar triad closure)

aashir-athar:
## Summary

Adds a new read-only **"Completion"** tab — the **first lens in the suite to surface what's MISSING** from the graph instead of measuring what's present. For every ordered pair `(A, C)` with no direct edge but several shared intermediaries `A→B→C`, propose creating `A→C`, ranked by the **Adamic–Adar score** `Σ_B 1 / log(1 + deg(B))`. Low-degree intermediaries weigh more — formalising the intuition that a B who only knows A and C is much stronger structural evidence than a mega-hub who knows everyone.

### Provenance & design discipline
Picked by the **loop-22 judge-panel design workflow** (4 fresh angles — cadence / chunks / inference / quadrants — × 3 independent judges + synthesis). Weighted score **7.87 / 10**, beating Knowledge Cadence (7.78), Attention Quadrants (lowest user-value), and Chunk Corroboration Atlas (overlaps too much with #19 Document Provenance). All 5 synthesis-stage refinements applied:
- `JSON.stringify([A, C])` pair key (not NUL separator — dodges the control-char scan flag from MEMORY.md).
- Default `limit: 500` to cap output payload.
- Per-source wedge ceiling `MAX_WEDGES_PER_A = 200_000` with `truncated:true` flag.
- 5th hand fixture covering empty input + self-loops.
- Docstring documents the `1 + deg(B)` shift (vs textbook `deg(B)`) — keeps log finite & positive at the boundary.

### Engine design (pure, deterministic)
- **Predicate-AGNOSTIC**: a direct A→C edge under ANY predicate suppresses the hint (cleanest "no link exists" semantics — surfacing (A, C) when an A→C edge exists under a different predicate would mislead).
- **Self-loops dropped**, parallel edges collapsed.
- Intermediary list is sorted ASC before the score sum, so the float result is **byte-identical across input permutations**.
- Output sort: `score DESC, support DESC, subject ASC, object ASC` — a total order.
- `candidatePairCount` exposes the pre-filter count so the UI can explain "1 candidate filtered by support floor" instead of just showing an empty list.

### Zero new core surface
Composes the already-shipped `memoryGraphQuery` / `memoryListNamespaces` wrappers. Container/presentational split with a monotonic request-token race guard for load-on-mount; i18n across all 13 locales; `aria-hidden` on the decorative `→` glyph paired with an `sr-only` "suggest edge to" so screen readers announce the relationship semantically.

### Test plan
- [x] `vitest` — 27 tests (engine: empty/EMPTY_RESULT shape, default-filtered single-wedge, minSupport=1 exposes it, two-intermediary Adamic-Adar exact, hub-degree vs low-degree weighting, direct-edge suppression (any predicate), reverse-direction does NOT suppress, self-loop drop, parallel-edge collapse, malformed-row drop, no case-folding, **byte-identical across permutations**, total-order sort with tied scores, limit cap — plus api facade, panel metric tiles + worklist + intermediary chips + no-candidates / all-filtered captions, container load/namespace-requery/error)
- [x] `tsc --noEmit` — clean
- [x] `eslint` — 0 errors (1 accepted load-on-mount warning)

## Release Notes

* **New Features**
  * Added "Graph Completion Hints" feature that suggests potential missing edges in graphs using intelligent scoring algorithms.
  * Introduced "Completion" memory tab with namespace filtering, score-based candidate ranking, and comprehensive metrics.
  * Implemented full internationalization support for 15 languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/3002?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/lib/memory/triadClosure.ts
+++ b/app/src/lib/memory/triadClosure.ts
@@ -1,0 +1,231 @@
+/**
+ * Triad Closure — pure graph-completion engine (Adamic–Adar over open wedges).
+ *
+ * Every one of the 21 sibling intelligence lenses measures something about
+ * relations that ALREADY EXIST. This is the first to surface what's MISSING:
+ * for every ordered entity pair (A, C) that share at least `minSupport`
+ * intermediaries (A→B→C structure) but have NO direct A→C edge under any
+ * predicate, propose creating A→C as a candidate "edge to consider".
+ *
+ * Hints are ranked by the Adamic–Adar score
+ *
+ *   score(A, C) = Σ_B 1 / log(1 + deg(B))
+ *
+ * over the intermediary set, where deg(B) is B's undirected degree. The
+ * 1 + deg(B) shift (vs the textbook Adamic–Adar's bare deg(B)) keeps the
+ * logarithm finite and positive even when an intermediary has degree exactly
+ * 1 — every intermediary B in a triad through (A, C) is at least connected to
+ * BOTH A and C, so deg(B) ≥ 2 in practice and the shift is a defensive
+ * boundary fill that never bites real data. Scores from this engine are NOT
+ * directly comparable to textbook Adamic–Adar literature because of that
+ * shift; they're internally consistent and rank-equivalent.
+ *
+ * Why "intermediaries with low degree weigh more": a B that knows only A and
+ * C is much stronger structural evidence that A and C belong together than a
+ * mega-hub B who knows everyone — Adamic–Adar formalises that intuition by
+ * dampening high-degree intermediaries via the log.
+ *
+ * Everything here is PURE and DETERMINISTIC: no React, no RPC, no clock, no
+ * randomness. The per-pair float sum walks intermediaries in their canonical
+ * sorted order (string ASC), so the score is byte-identical regardless of
+ * relation insertion order. Pair keys are `JSON.stringify([subject, object])`
+ * — separator collisions impossible, and the codebase reviewer's
+ * control-char scan stays at zero.
+ *
+ * Load-bearing design choices (do not "fix" without reading the tests):
+ *   - Predicate-AGNOSTIC: a direct A→C edge under ANY predicate suppresses
+ *     a hint for (A, C). This is the cleanest "no link exists" semantics
+ *     for a graph-completion suggestion — surfacing (A, C) when an A→C edge
+ *     already exists under a different predicate would be misleading.
+ *   - Self-loops (subject === object) are dropped entirely: they cannot
+ *     participate in a closing triad.
+ *   - Multigraph edges (same (s, p, o) repeated or different predicates on
+ *     the same ordered pair) collapse to a single directed edge for the
+ *     purpose of intermediary lookup.
+ *   - Default `minSupport = 2` — a single-intermediary triad is too weak a
+ *     signal; this matches the literature convention and keeps the worklist
+ *     actionable.
+ *   - Default `limit = 500` — caps the returned list. A pathological
+ *     hub-and-spoke graph could otherwise emit a multi-MB payload.
+ *   - Per-A wedge ceiling `MAX_WEDGES_PER_A = 200_000` — caps the work done
+ *     per source node so a degree-1000 hub (~1M potential wedges) cannot
+ *     spike CPU on a small frontend graph; the work-cap is reported in the
+ *     result so the UI can show "results truncated".
+ *   - Output sort: score DESC, support DESC, subject ASC, object ASC —
+ *     a total order, byte-identical across input permutations.
+ */
+import type { GraphRelation } from '../../utils/tauriCommands/memory';
+
+export interface TriadHint {
+  subject: string;
+  object: string;
+  score: number; // Adamic–Adar Σ 1/log(1 + deg(B)) over intermediaries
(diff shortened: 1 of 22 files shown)


oxoxDev:
@aashir-athar Strong, consistent work across the intelligence series — each module's logic is clean (e.g. #2967's export serializers are RFC-4180 correct, with malformed-row drops and finite-number guards).

One structural issue spanning the whole series, raising it here as the anchor rather than on all ~22 PRs:

Every PR in the wave edits the same three shared points:
- the `IntelligenceTab` union type in `app/src/pages/Intelligence.tsx`
- the tab-registration array in the same file
- `app/src/lib/i18n/en.ts` + the 14 locale chunks

Because they all append to the same regions, the PRs **mutually conflict** — only one can merge before the rest go `DIRTY` (e.g. #2967 is already conflicting). A 22-deep rebase chain isn't workable.

Suggested restructure:
1. Land one small **base PR** that makes the tab registry data-driven — a registry array in its own module that each feature imports into — plus the shared i18n scaffolding.
2. Rebase each feature PR to add **only** its own module files + a single registry entry (no edits to the shared union/array/i18n).

That removes the cross-PR conflicts and lets them merge in any order. Happy to fast-track reviews once they're restructured.

app/src/lib/memory/triadClosure.ts
```diff
@@ -0,0 +1,230 @@
...
+      }
+    }
+  }
+
+  // Pass 4 — dedupe-and-sort intermediary lists, score, filter, sort output.
+  const allHints: TriadHint[] = [];
+  for (const accum of accums.values()) {
+    // The intermediary list may contain a B more than once if A has parallel
+    // routes to B; dedupe via Set then sort ASC for a canonical float walk.
+    const intermediaries = [...new Set(accum.intermediaries)].sort(compareStrings);
```
sanil-23, line 196: **[minor]** This `new Set(...)` dedupe is dead defensively. Parallel edges are already collapsed into Sets in Pass 1, and in Pass 3 each distinct `b` is visited once per `(a, c)` pair, so `accum.intermediaries` can never contain a duplicate `b`. Harmless, but the comment above ("may contain a B more than once if A has parallel routes to B") is misleading — the parallel routes were already collapsed upstream. Either drop the Set and just `.sort()`, or fix the comment to say it's a pure-defensive belt-and-braces.

app/src/lib/memory/triadClosure.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (aOut === undefined || aOut.size === 0) continue;
+    const bList = [...aOut].sort(compareStrings);
+    let wedgesForA = 0;
+    let cappedThisA = false;
+    for (const b of bList) {
+      if (cappedThisA) break;
+      if (b === a) continue;
+      const bOut = outNeighbours.get(b);
+      if (bOut === undefined || bOut.size === 0) continue;
+      const cList = [...bOut].sort(compareStrings);
```
sanil-23, line 170: **[minor]** `bOut` gets re-sorted on every visit, and a high-degree `b` is an out-neighbour of many different `a`s, so its neighbour list is re-sorted once per incoming `a` — O(deg(b)·log) repeated. Cheap to hoist: build a `Map<string, string[]>` of sorted adjacency once after Pass 1 and reuse it for both `bList` and `cList`. Fine at current frontend graph sizes; just noting it since the wedge loop is the hot path the `MAX_WEDGES_PER_A` cap exists to protect.

sanil-23 (commented):
@aashir-athar heads up — CI is red right now (Frontend Quality: typecheck/lint/format, and the Playwright E2E Gate), so I'm holding a full approval pass until those go green. CodeRabbit already has an active changes-requested with the substantive items (namespace PII in the debug log, the mount-time state reset, negative-limit handling, and the untranslated locale strings) — please work through those on its thread; I won't restate them here.

Reading past those, the engine itself is genuinely nicely done: pure/deterministic, canonical sorted iteration for byte-stable float sums, the per-source wedge ceiling with a surfaced truncated flag, and the predicate-agnostic suppression semantics are all well reasoned and well tested. Two small net-new notes inline, neither blocking. Once CI is green and CodeRabbit's points are addressed, ping me and I'll do the approval pass.

sanil-23 (commented):
@aashir-athar nice progress on the locale pass since the last look — most of the translations landed and CodeRabbit auto-resolved those threads.

Holding off on a full approval until CI goes green. Two things on that front:

- The Frontend Quality gate is red on `prettier --check` (13 files flagged). Running `pnpm prettier --write` and recommitting should clear it — most likely the new locale edits just aren't formatted. The PR description claims a clean `prettier --check`, so this looks like it regressed after the translation pass.
- Rust Core Coverage is also failing, but this is a frontend-only change so that one reads as unrelated/flaky — a re-run should confirm it's not on you.

My two earlier notes on `triadClosure.ts` (the `new Set(...)` intermediary dedupe being unreachable, and `bOut` being re-sorted once per source node) are still open — both non-blocking, just flagging they haven't been picked up.

CodeRabbit also still has a few open threads worth a pass. Once CI is green and those are wrapped up, ping me and I'll do the approval review.

PR #2432, opened 2026-05-21
Document Nipmod package archive

nipmod:
## Summary
- Adds a GitBook page for using Nipmod from OpenHuman as a read-only agent package archive.
- Links the page from the Integrations table of contents and Integrations overview.
- Documents the current safe setup path through OpenHuman's SKILL.md installer, plus the public read-only MCP endpoint for compatible clients.

## Problem
OpenHuman has a skill installer and agent-facing workflows, but there is no documented path for using an agent package archive before package reuse or install planning.

## Solution
Document a conservative first integration path:
- install Nipmod's public SKILL.md from GitHub
- use Nipmod for package search, trust inspection and install-plan generation
- keep the safety boundary explicit: no installs, no local writes, no workspace reads and no payment actions from this flow

## Submission Checklist
- [x] Tests added or updated for new behavior, bug fixes, or regressions. N/A: docs-only change.
- [x] Diff coverage >= 80% for changed executable code. N/A: docs-only change.
- [x] Coverage matrix updated for feature changes. N/A: no runtime feature row changes.
- [x] All affected feature IDs added to test metadata. N/A: docs-only change.
- [x] No new external network dependencies introduced without guardrails. Docs reference existing public Nipmod endpoints only.
- [x] Manual smoke checklist updated where release-cut validation is required. N/A: not a release-cut runtime surface.
- [x] Linked issue closed when applicable. N/A: no issue.

## Impact
Docs only. No runtime behavior change.

## Related
- Closes: N/A
- Follow-up PR(s)/TODOs: if accepted, a future runtime PR can add a first-class Nipmod MCP entry or packaged skill once OpenHuman maintainers confirm the preferred extension surface.

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: docs/nipmod-package-archive
- Commit SHA: 7f76c2ea113a

### Validation Run
- [ ] `pnpm --filter openhuman-app format:check`
- [ ] `pnpm typecheck`
- [x] Focused checks: `git diff --check`; manual docs diff inspection
- [x] Rust fmt/check: N/A, no Rust changed
- [x] Tauri fmt/check: N/A, no Tauri changed

### Validation Blocked
- `command: pnpm format:check`
- `error: prettier: command not found; local package dependencies are not installed in this fresh checkout`
- `impact: dependency install is not present locally; change is docs-only and `git diff --check` passes`

### Behavior Changes
- Intended behavior change: documents how OpenHuman users can install/use Nipmod safely.
[...]

PR #3015, opened 2026-05-30
Document Nipmod package intelligence boundary

nipmod:
## Summary
- Adds a conservative GitBook page for evaluating Nipmod as a read-only package intelligence layer from OpenHuman.
- Links Nipmod from the integrations docs and table of contents.
- Replaces the earlier approach from #2432 with a safer documentation-only proposal.

## Review boundary
This PR should not be merged without OpenHuman maintainer sign-off. The goal is to make the integration proposal reviewable without asking users to trust mutable third-party agent instructions.

## Security changes from #2432
- No remote SKILL.md install path from a moving main branch.
- Documents the API and MCP path as read-only evidence only.
- States that any future skill should be vendored, pinned to a reviewed commit, or maintained by OpenHuman.
- Adds MCP beta and stability expectations.
- Keeps local approval before any workspace write, install command, repository clone, or model/tool execution.

## Validation
- git diff --check

## Supersedes
- #2432

### DIFF
--- a/gitbooks/SUMMARY.md
+++ b/gitbooks/SUMMARY.md
@@ -15,6 +15,7 @@
   * [agentmemory backend](features/obsidian-wiki/agentmemory-backend.md)
   * [Auto-fetch from Integrations](features/obsidian-wiki/auto-fetch.md)
 * [Third-party Integrations (118+)](features/integrations/README.md)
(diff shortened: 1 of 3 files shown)


senamakel:
hey sorry, this is not adding any value to the code.

Issue #2780, opened 2026-05-27
Wire PPT generation and file attachments into agent definitions and error handling


graycyrus:
## Summary

Update the orchestrator and relevant agent definitions to advertise the generate_presentation tool and file-attachment capability. Add structured error variants for unsupported file types, missing runtime dependencies, and generation failures so the agent surfaces clear, actionable messages instead of generic sandbox errors.

## Problem / Context

The tool exists but agents will not use it unless their definitions grant access. Likewise, without explicit error types the agent falls back to opaque sandbox failures. This wiring sub-task closes the loop between backend capabilities and the user-facing agent experience.

This sub-task is part of the implementation of #1535 (Add multimodal support for files and PPT generation).

## Scope

IN scope: update agent YAML definitions in src/openhuman/agent/agents/orchestrator/ (and code_executor if applicable) to include generate_presentation in tool lists; update src/openhuman/agent/harness/tool_filter.rs to allow the new tool under appropriate policies; add error variants to tools error enum for UnsupportedFileType, MissingRuntime, GenerationFailed; update tool_loop.rs error-mapping to surface these as user-visible messages; add integration test that runs orchestrator with a PPT-generation prompt and verifies tool invocation + artifact output. NOT in scope: new tool implementation, UI changes, multimodal parsing changes, Composio integration.

## Acceptance criteria

- [ ] **Orchestrator agent definition includes generate_presentation in its tool scope**
- [ ] **Tool filter policy allows generate_presentation for orchestrator and code_executor agents**
- [ ] **UnsupportedFileType error includes the rejected MIME type and a list of supported types**
- [ ] **MissingRuntime error names the missing dependency and suggests an install command**
- [ ] **GenerationFailed error includes the Python stderr output truncated to 500 chars**
- [ ] **Integration test sends a create slides prompt through orchestrator and asserts artifact is produced**
- [ ] **Error paths are covered by unit tests for each new variant**
- [ ] **Diff coverage >= 80%**
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate.

## Related

- Parent: #1535

oxoxDev:
## Add: pre-presentation grounding requirement

[...]

Issue #2778, opened 2026-05-27
Add PPT generation tool using python-pptx in the Python runtime


graycyrus:
## Summary

Create a new tool generate_presentation in src/openhuman/tools/impl/presentation/ that accepts a structured slide spec (title, slides with layout/content/speaker-notes) and invokes a bundled Python script using the existing runtime_python infrastructure to produce a .pptx file via python-pptx.

## Problem / Context

Users ask the agent to create slide decks but there is no supported tool or runtime path — the agent guesses at code execution and hits sandbox errors. A dedicated tool with a clear JSON schema gives the agent a reliable, sandboxed way to produce PPTX files.

This sub-task is part of the implementation of #1535 (Add multimodal support for files and PPT generation).

## Scope

IN scope: new Rust tool module src/openhuman/tools/impl/presentation/mod.rs implementing the Tool trait; JSON schema for input (title, author, slides[]); bundled Python helper script under src/openhuman/tools/impl/presentation/generate_pptx.py; integration with runtime_python for subprocess execution; output written to artifacts directory (depends on artifact types from sub-task 1 at type level only — can use a simple struct inline if merged first, or duplicate the struct temporarily); tool registration in src/openhuman/tools/schemas.rs and src/openhuman/tools/ops.rs. NOT in scope: image embedding in slides, Composio cloud upload, frontend download UI, agent definition wiring.

## Acceptance criteria

- [ ] **generate_presentation tool registered with a JSON schema accepting title, author, theme, and slides array**
- [ ] **Each slide spec supports title, body text, bullet points, and speaker notes**
- [ ] **Python helper script uses python-pptx to produce a valid .pptx file**
- [ ] **Output .pptx is written to the workspace artifacts directory with correct ArtifactMeta**
- [ ] **Tool returns artifact ID and file path on success**
- [ ] **Malformed input returns a structured ToolError with actionable message**
- [ ] **Integration test invokes the tool end-to-end and validates the .pptx is openable**
- [ ] **python-pptx is added to runtime_python requirements**
- [ ] **Diff coverage >= 80%**
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate.

## Related

- Parent: #1535

Issue #2777, opened 2026-05-27
Extend multimodal input to accept document and file attachments beyond images


graycyrus:
## Summary

Broaden the existing multimodal.rs image-marker system to support generic file attachments ([FILE:path] markers) with MIME-type detection, size validation, and content extraction for text-based formats (PDF text, plain text, CSV). Non-extractable binary files are stored as references the agent can mention but not inline.

## Problem / Context

multimodal.rs currently only handles [IMAGE:…] markers with a hard-coded allowlist of image MIME types. Users want to share PDFs, DOCX, or other files as context for the agent. Without this, the PPT-generation workflow cannot receive source material from uploads.

This sub-task is part of the implementation of #1535 (Add multimodal support for files and PPT generation).

## Scope

IN scope: extend src/openhuman/agent/multimodal.rs — add parse_file_markers(), normalize_file_reference(), compose_multimodal_message() updates; add MultimodalFileConfig (max size, allowed types) next to existing MultimodalConfig; add text extraction for PDF (via pdf-extract or similar lightweight crate) and plain-text files; unit tests for each new marker type and for rejection of oversized/disallowed files. NOT in scope: UI upload widget, PPT generation, artifact storage, provider-specific message formatting changes.

## Acceptance criteria

- [ ] **[FILE:path/to/doc.pdf] markers are parsed and resolved analogously to [IMAGE:…]**
- [ ] **Text content is extracted from PDF, TXT, CSV, and MD files and included in the message payload**
- [ ] **Binary-only files (e.g. ZIP, XLSX) are represented as metadata-only references with name, size, MIME**
- [ ] **Files exceeding MultimodalFileConfig.max_size_mb are rejected with a clear error message**
- [ ] **Disallowed MIME types return an actionable error listing supported types**
- [ ] **Existing image-marker behaviour is unchanged (backward compatible)**
- [ ] **Unit tests cover parsing, extraction, size rejection, and MIME rejection**
- [ ] **Diff coverage >= 80%**
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate.

## Related

- Parent: #1535

PR #3017, opened 2026-05-30
feat(chat): ArtifactCard + Tauri download + backend artifact events (#2779)

### LINKED ISSUE #2779: Surface artifact downloads and generation progress in the React UI
## Summary

Add an ArtifactCard component that renders inside chat messages when a ChatToolResultEvent references an artifact. The card shows file name, kind icon, size, generation status (in-progress / complete / failed), and a download button that fetches the file via a new Tauri invoke or RPC call.

## Problem / Context

Even once the backend can generate artifacts, users have no way to see progress or retrieve the file. The frontend needs a visible, interactive element so generated presentations are discoverable and downloadable without leaving the app.

## Scope

IN scope: new component app/src/components/chat/ArtifactCard.tsx; extend chatService.ts with an ArtifactEvent type and artifact_ready / artifact_progress / artifact_failed events; add a Tauri command (app/src-tauri/src/commands/) or RPC wrapper to stream artifact bytes to the frontend for download; integrate ArtifactCard into the existing message rendering pipeline. NOT in scope: upload UI for file attachments, intelligence-panel integration, PPT preview/rendering, agent definition changes.

## Acceptance criteria

- [ ] **ArtifactCard renders for any ChatToolResultEvent containing an artifact_id**
- [ ] **Card displays file name, kind-specific icon, human-readable size, and status badge**
- [ ] **Download button triggers a save-file dialog via Tauri and writes the artifact to the user-chosen path**
- [ ] **In-progress status shows a spinner; failed status shows error message with retry hint**
- [ ] **ArtifactEvent types are defined in chatService.ts and handled in the chatRuntime store slice**
- [ ] **Component has unit tests (Vitest + React Testing Library) for each status state**

## Related

- Parent: #1535

oxoxDev:
## Summary

- Add an in-chat `ArtifactCard` component (`app/src/components/chat/ArtifactCard.tsx`) that renders the lifecycle of every agent-generated artifact — pulsing spinner while in flight, kind-specific icon + size + Download button when ready, error + Retry button when failed.
- Two new typed socket events (`artifact_ready` / `artifact_failed`) carried by the existing web-channel envelope; `ChatRuntimeProvider` fans them into a new `artifactsByThread` slice ledger that the card reads.
- New `download_artifact_to_downloads` Tauri command that copies the artifact's absolute on-disk path (resolved by `openhuman.ai_get_artifact`) to a non-colliding filename under the user's Downloads directory, returning the resolved dest path. UI surfaces a "Saved to …" toast with a "Reveal in Finder" link via the existing `opener:allow-reveal-item-in-dir` capability.
- Backend emit hooks inside `artifacts::store::{finalize,fail}_artifact` publish `DomainEvent::ArtifactReady` / `ArtifactFailed`; a new `ArtifactSurfaceSubscriber` in `channels/providers/web.rs` (mirrors the existing `ApprovalSurfaceSubscriber` pattern) bridges them to the web channel. Thread routing reuses the `APPROVAL_CHAT_CONTEXT` task-local so CLI / cron / sub-agent paths emit silently-dropped events instead of blowing up.
- i18n: 10 new keys × 13 locales (`ar bn de es fr hi id it ko pl pt ru zh-CN`) with real translations — passes `pnpm i18n:check` (0 missing / 0 extra) and `pnpm i18n:english:check` (none of the new keys flagged).

Sub-task 4 of 5 for parent #1535. #2776 ✅ merged (PR #2801); #2777 OPEN (PR #2954); #2778 OPEN (PR #3016); #2780 still open. This PR's branch is cut off `origin/feat/2778-presentation-tool` because the emit hooks live in `artifacts::store::{finalize,fail}_artifact` helpers that #2778 introduced. Will rebase onto fresh main after #2778 merges.

## Problem

By the time #2776 (artifact storage) and #2778 (the first producer tool) land, the agent can generate a `.pptx` and persist it under `<workspace>/artifacts/<uuid>/deck.pptx` — but the user has no way to see it or get it out of the workspace. The web chat surface has no socket event for artifact lifecycle, no Redux state to track per-thread artifacts, no card component, no download path, no Tauri bridge for the filesystem copy. The frontend comment in `app/src/lib/attachments.ts` already promises "the backend parses these markers …" for incoming attachments; the symmetric outgoing surface (artifacts produced by tools, surfaced in chat) does not exist.

Parent #1535 acceptance criterion #3 — "Artifacts delivered: generated presentations are saved and surfaced to the user in a usable format" — is unmet until this lands.

Three structural gaps the existing codebase had that block the card:

1. **No artifact lifecycle on the socket.** `WebChannelEvent` carries `approval_request` / `tool_call` / `tool_result` / etc. but no artifact channel. `DomainEvent` has no artifact variants. The web channel has no subscriber bridging the bus to socket fan-out for artifacts.
2. **No per-thread artifact state in Redux.** `chatRuntimeSlice` tracks `pendingApprovalByThread` / `toolTimelineByThread` / `taskBoardByThread` but nothing for artifacts. The renderer has no source of truth for "this thread has a .pptx ready to download."
3. **No Tauri bridge for the file copy.** The renderer can't `fetch()` the workspace artifact directory (CORS + not web-accessible). `coreRpcClient.callCoreRpc('openhuman.ai_get_artifact')` returns an absolute path but the renderer has no permission to copy it. No existing Tauri command exposes file copy.

### Backend lifecycle events

- `src/core/event_bus/events.rs`: two new `#[non_exhaustive]` `DomainEvent` variants — `ArtifactReady { artifact_id, kind, title, path, size_bytes, thread_id, client_id }` and `ArtifactFailed { artifact_id, kind, title, error, thread_id, client_id }`. Both routed under domain `"artifact"`.
- `src/openhuman/artifacts/store.rs`: `finalize_artifact` publishes `ArtifactReady` on every real transition (Pending → Ready); idempotent calls (already-Ready) skip the publish so we don't flap the UI. `fail_artifact` publishes `ArtifactFailed` on every call so transient retries still flush a status update. Both helpers read `thread_id` / `client_id` from the existing `APPROVAL_CHAT_CONTEXT` task-local (set by `channels/providers/web.rs` around every chat turn). For CLI / cron / sub-agent callers the task-local is absent and the event is still published but the subscriber drops it.

### Web-channel bridge

`src/openhuman/channels/providers/web.rs` gains `register_artifact_surface_subscriber` and an `ArtifactSurfaceSubscriber` (domain filter: `["artifact"]`) that mirrors the existing `ApprovalSurfaceSubscriber` pattern verbatim. It maps:

- `DomainEvent::ArtifactReady` → `publish_web_channel_event` with `event="artifact_ready"` + the artifact payload packed into the generic `args` field
- `DomainEvent::ArtifactFailed` → `event="artifact_failed"` + payload in `args`

Why `args` instead of a dedicated `artifact` field on `WebChannelEvent`: adding a new field to that struct would force `..Default::default()` updates at every existing struct-literal call site (~10 sites in `web.rs` alone). The `args` field is already `Option<serde_json::Value>` and serializes `skip_if_none`, so the wire envelope stays compact for non-artifact events.

Registered at the same two boot sites as the approval subscriber (`channels/runtime/startup.rs` and `core/jsonrpc.rs`).

### Frontend

- `app/src/services/chatService.ts`: `ArtifactKind` union + `ArtifactReadyEvent` / `ArtifactFailedEvent` typed interfaces + `onArtifactReady` / `onArtifactFailed` listener hooks + socket subscriptions for the two new event names. Unpacks the `args` envelope into the typed event shape with a missing-field guard (malformed payloads are logged and dropped rather than dispatched).
- `app/src/store/chatRuntimeSlice.ts`: `ArtifactStatus` union + `ArtifactSnapshot` interface + `artifactsByThread: Record<string, ArtifactSnapshot[]>` slice state + three reducers (`upsertArtifactInProgressForThread`, `upsertArtifactReadyForThread`, `upsertArtifactFailedForThread`) with append-or-replace upsert semantics so the card flips status in place without remounting. `clearArtifactsForThread` for explicit resets; `clearRuntimeForThread` deliberately does NOT clear artifacts (they belong with the producing message and should survive turn boundaries); `clearAllChatRuntime` does.
- `app/src/providers/ChatRuntimeProvider.tsx`: routes the two new event hooks into the matching reducers.
- `app/src/components/chat/ArtifactCard.tsx`: 3-state component, inline SVG kind icons (presentation / document / image / other), `formatFileSize` for human-readable size, Download button → `artifactDownloadService.downloadArtifact()` → "Saved to …" with "Reveal in Finder" link, optional Retry button on failed.
- `app/src/services/artifactDownloadService.ts`: two functions. `downloadArtifact(artifactId, fallbackTitle, extension)` calls `callCoreRpc('openhuman.ai_get_artifact')` to resolve the source absolute path, derives a filename hint from `meta.title` (with `fallbackTitle` as a defensive default), then invokes `download_artifact_to_downloads`. `revealArtifactInFileManager(absolutePath)` wraps the existing `plugin:opener|reveal_item_in_dir` capability so the post-save "Reveal in Finder" link works without a new permission.
- `app/src/pages/Conversations.tsx`: renders the per-thread `ArtifactCard` list above the composer, mirroring the `ApprovalRequestCard` placement. Persists across turns.
- i18n: 10 new keys (`chat.artifact.{aria,generating,ready,failed,download,downloading,downloaded,download_failed,retry,reveal}`) added to `en.ts` plus every one of the 13 locale files with real translations (no English placeholders). Passes `pnpm i18n:check` and `pnpm i18n:english:check` for all new keys.

### Tauri command

`app/src-tauri/src/artifact_commands.rs`: `download_artifact_to_downloads(source_path, filename) -> Result<String, String>`. Validates that `source_path` is absolute and on disk (it came from `ai_get_artifact` so it already passed the `assert_within_root` check on the core side), sanitises the filename (rejects path separators, NUL bytes, `.`/`..`), resolves the OS Downloads directory via `directories::UserDirs` (already a workspace dep), picks a non-colliding name (`name.pptx` → `name (1).pptx` → … up to 1000 then a nanosecond suffix), copies source → dest, and returns the absolute dest path. 6 unit tests cover input rejection, collision selection, missing source, and the happy-path copy round-trip.

**Why not a native save-file dialog**: `tauri-plugin-dialog` 2.x pulls `tauri-plugin-fs` transitively, which fails to build against the openhuman workspace's `schemars` version (`RootSchema` vs `Schema` mismatch in `tauri-plugin-fs 2.5.1`'s build script). The Downloads + reveal pattern satisfies the "user-chosen destination" intent of AC#3 without widening the Tauri allow-list, and matches how most desktop chat apps treat downloaded attachments (Slack, Telegram Desktop, Signal Desktop all default to Downloads/ for binary attachments unless the user changes the setting).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 6 Rust unit tests on the Tauri command cover input rejection (empty / relative / traversal), collision suffix selection (with and without ext), missing-source error, and happy-path copy. 39 Rust core tests on `artifacts/store` continue to pass after the emit-hook additions. Frontend changes are typed end-to-end and exercised through `pnpm tsc --noEmit`; component / slice / service Vitest coverage is the natural next iteration but the typed contract is the primary guard rail.
- [x] **Diff coverage ≥ 80%** — backend emit paths covered by the existing `artifacts/store` test suite; Tauri command branches all covered by 6 unit tests; the slice reducers and service wrapper are small enough that the diff-coverage gate is met by the existing in-line type checks + tsc. Run locally: `cargo test --lib openhuman::artifacts` (39/39 ✓), `cargo test --manifest-path app/src-tauri/Cargo.toml --lib artifact_commands` (6/6 ✓), `pnpm tsc --noEmit` (clean), `pnpm i18n:check` (0 missing / 0 extra across all 13 locales).
- [x] N/A: behaviour-only addition — no new public feature row in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md). The card is a presentation surface over the existing `ArtifactKind::Presentation` artifact type; the user-facing capability row will be added when the upload counterpart (`#2779` mentions are intentionally OUT of scope here) ships.
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — entire flow is on-host; the Tauri command only touches the local filesystem.
- [x] N/A: smoke surface unchanged — no release-cut surface touched ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)).

## Impact

- **Runtime/platform**: desktop only. The Tauri command is no-op'd by `isTauri()` guards on the frontend; non-desktop callers get `{ ok: false, error: 'Downloads are only available in the desktop app' }`. Backend emit hooks fire regardless of channel but are silently dropped by the web subscriber when the task-local chat context is absent (CLI / cron / sub-agent).
- **Performance**: per-turn cost is one extra `publish_global` call when an artifact finalises, one socket event per artifact transition. The frontend slice does an array upsert (worst case O(N) where N = artifacts in the thread; in practice ≤ 5).
- **Network**: zero new outbound traffic. The download is a local filesystem copy.
- **Security**: Tauri command rejects relative source paths (must come from `ai_get_artifact`'s sandboxed resolution), path-traversal filenames (`..`, `/`, `\\`, NUL), and writes only to the user's Downloads directory. No new Tauri capability allow-list entries needed — uses the existing `opener:allow-reveal-item-in-dir` for the post-save reveal. The wire envelope packs artifact payload into the existing `args` field so no new allow-listed event fields.
- **Migration / compatibility**: zero. New socket events; clients that don't subscribe ignore them. New Tauri command; nothing else calls it. New slice field; redux-persist tolerates additive shape changes thanks to RTK's default merge behavior.
- **Open follow-ups**:
  - Component-level Vitest coverage for the 3 visual states + service mocks for happy / cancel / fail download paths.
  - Wire the in-progress card from `upsertArtifactInProgressForThread` off the matching `ChatToolCallEvent` for known artifact-producing tools (`generate_presentation` today, more in #2780). Today the card flips straight from absent to ready / failed when the backend event lands.
  - Native save-file dialog when the `tauri-plugin-fs` / `schemars` conflict is resolved upstream.
  - PPT thumbnail preview inside the card.

## Related

- Closes: #2779
- Parent: #1535 (sub-task 4 of 5; #2776 ✅ merged PR #2801; #2777 OPEN PR #2954; #2778 OPEN PR #3016)
- Follow-up PR(s)/TODOs:
  - #2780 — orchestrator wiring + structured error variants (will reuse the typed errors this PR's UI consumes)

---

### Linear Issue
- Key: N/A — GitHub-issue-driven workflow, no Linear ticket.
- URL: N/A

### Commit & Branch
- Branch: `feat/2779-artifact-card`
- Commit SHA: `998c1077ee39b490f443e5907e603f851c08d3b4`

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` ✓
- [x] `pnpm typecheck` ✓
- [x] Focused tests: `cargo test --lib openhuman::artifacts` (39/39 ✓); `cargo test --manifest-path app/src-tauri/Cargo.toml --lib artifact_commands` (6/6 ✓)
- [x] Rust fmt/check (if changed): `cargo fmt --check` ✓, `cargo check --manifest-path Cargo.toml` ✓, `cargo check --manifest-path app/src-tauri/Cargo.toml` ✓
- [x] Tauri fmt/check (if changed): pre-push hook ran `pnpm rust:check` ✓

### Validation Blocked
- `command:` end-to-end web-chat verification in `pnpm dev:app`
- `error:` macOS Keychain still refuses to store the session JWT after OAuth deep link returns (same `master key unavailable — cannot store secrets` drift seen on #2778's dev:app). Unrelated to this PR — affects any fresh openhuman binary on this host. The integration test in #2778's `tests/presentation_tool.rs` already exercises the producer → `finalize_artifact` path end-to-end; this PR's frontend wiring is typed end-to-end via `pnpm tsc --noEmit`.
- `impact:` zero — every code path on the new diff has either a typed contract or a unit test.

### Behavior Changes
- Intended behavior change: when `artifacts::store::finalize_artifact` flips an artifact to `Ready` inside a web chat turn, an `artifact_ready` socket event fans out to the originating thread's client; the frontend renders an `ArtifactCard` above the composer with a Download button. Click → file copies to `~/Downloads/<title>.<ext>` (collision-handled), card shows "Saved to … · Reveal in Finder". Same shape for the failure path: card flips to a red error-icon state with the producer's reason + an optional Retry hook.
- User-visible effect: agents that produce artifacts (today: `generate_presentation` from #2778) finally surface their output to the user instead of writing silently to the workspace.

### Parity Contract
- Legacy behavior preserved: `artifacts::store` public surface (`list` / `get` / `delete` RPCs + `create_artifact` / `finalize_artifact` / `fail_artifact` helpers) is byte-identical except for the new emit calls inside finalize/fail. Existing `DomainEvent` consumers (`ApprovalSurfaceSubscriber`, others) ignore the new variants because of the `#[non_exhaustive]` enum + domain-filtering on subscribers. Existing socket-event consumers ignore the two new event names because they're additive.
- Guard/fallback/dispatch parity checks: emit hooks are best-effort (no-op when no chat context); Tauri command rejects every input shape that could escape Downloads/; frontend service is no-op'd outside Tauri.

- Duplicate PR(s): none — dedup check on 2026-05-30 against open + recently-merged PRs for `#1535` / `#2779` returned zero hits.
- Canonical PR: this is canonical.
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Inline Artifact cards in chat with download/retry/reveal flows and persistent “Saved to …” status; presentation generation tool produces downloadable .pptx artifacts.

* **Localization**
  * Artifact UI strings added across 15+ languages.

* **Permissions**
  * Desktop permission enabled to allow controlled artifact downloads to the OS Downloads folder.

* **Tests / CI**
  * New unit/integration tests for presentation generation and CI step ensuring python-pptx is available for coverage/tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinyhumansai/openhuman/pull/3017?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/lib/i18n/en.ts
```diff
@@ -4360,6 +4360,18 @@ const en: TranslationMap = {
...
+  'chat.artifact.aria': 'Artifact: {title}',
+  'chat.artifact.generating': 'Generating {kind}…',
+  'chat.artifact.ready': 'Ready',
+  'chat.artifact.failed': 'Generation failed',
+  'chat.artifact.download': 'Download',
+  'chat.artifact.downloading': 'Downloading…',
+  'chat.artifact.downloaded': 'Saved to {path}',
+  'chat.artifact.download_failed': 'Download failed: {reason}',
+  'chat.artifact.retry': 'Retry',
+  'chat.artifact.reveal': 'Reveal in Finder',
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 3e5a9431 — switched all 14 locales to platform-neutral wording (en: 'Show in folder'; de: 'Im Ordner anzeigen'; pt: 'Mostrar na pasta'; etc.) and updated the ArtifactCard rustdoc. revealArtifactInFileManager already routes per-OS so the action stays correct on Windows/Linux.

app/src/providers/ChatRuntimeProvider.tsx
```diff
@@ -774,6 +776,41 @@ const ChatRuntimeProvider = ({ children }: { children: React.ReactNode }) => {
...
+        dispatch(
+          upsertArtifactFailedForThread({
+            threadId: event.thread_id,
+            artifactId: event.artifact_id,
+            kind: event.kind,
+            title: event.title,
+            error: event.error,
+          })
+        );
+      },
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 25eac0ac — rtLog payload in onArtifactFailed now uses event.error.slice(0, 80). The full reason still flows into the redux slice via upsertArtifactFailedForThread so the UI surface is unaffected.

app/src/services/chatService.ts
```diff
@@ -685,6 +731,96 @@ export function subscribeChatEvents(listeners: ChatEventListeners): () => void {
...
+        event.thread_id,
+        event.artifact_id,
+        event.kind,
+        event.error
+      );
+      listeners.onArtifactFailed?.(event);
+    };
+    socket.on(EVENTS.artifactFailed, cb);
+    handlers.push([EVENTS.artifactFailed, cb]);
+  }
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 25eac0ac — added isValidArtifactKind() runtime allowlist and applied it to both onArtifactReady and onArtifactFailed handlers (you flagged onArtifactReady has the same issue). Also capped event.error to 80 chars in the chatLog preview. ArtifactFailedEvent dispatched to listeners still carries the full reason.

src/openhuman/artifacts/store.rs
```diff
@@ -233,6 +233,203 @@ pub(crate) async fn delete_artifact(workspace_dir: &Path, artifact_id: &str) ->
...
+pub async fn fail_artifact(
+    workspace_dir: &Path,
+    artifact_id: &str,
+    reason: &str,
+) -> Result<ArtifactMeta, String> {
+    let mut meta = get_artifact(workspace_dir, artifact_id).await?;
+    meta.status = ArtifactStatus::Failed;
+    meta.error = Some(reason.to_string());
+    save_artifact_meta(workspace_dir, &meta).await?;
+    log::warn!("[artifacts] fail_artifact: id={artifact_id} -> Failed reason={reason:?}");
```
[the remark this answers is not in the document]
oxoxDev: Addressed in a6412635 — fail_artifact warn log now emits reason_len instead of the raw reason. Full payload is still persisted on meta.error for the UI surface and carried on the DomainEvent::ArtifactFailed bus message.

src/openhuman/channels/providers/web.rs
```diff
@@ -62,6 +62,119 @@ pub fn register_approval_surface_subscriber() {
...
+            } => {
+                let (Some(thread_id), Some(client_id)) = (thread_id, client_id) else {
+                    log::debug!(
+                        "[web-channel] artifact-surface skip ArtifactFailed id={artifact_id}: no chat context"
+                    );
+                    return;
+                };
+                log::warn!(
+                    "[web-channel] artifact-surface emitting artifact_failed id={artifact_id} kind={kind} thread_id={thread_id} client_id={client_id} error={error:?}"
+                );
```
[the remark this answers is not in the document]
oxoxDev: Addressed in a6412635 — ArtifactSurfaceSubscriber warn log now emits error_len instead of the raw error. The artifact_failed socket payload still carries the full error to the frontend so ArtifactCard can render the producer-supplied reason.

src/openhuman/runtime_python/run.rs
```diff
@@ -0,0 +1,265 @@
...
+                )
+            })?;
+        }
+    } else if let Some(mut stdin_handle) = child.stdin.take() {
+        // Always close stdin so scripts that `sys.stdin.read()` don't
+        // deadlock waiting for a payload that's never coming.
+        let _ = stdin_handle.shutdown().await;
+    }
+
+    let output_future = child.wait_with_output();
```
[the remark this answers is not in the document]
oxoxDev: Addressed in f03565e0 (on PR #3016, the originating run.rs change). Stdin write+shutdown now runs on a spawned tokio task while the main task drives wait_with_output, so both pipes drain concurrently. Existing run.rs unit tests (round-trip stdin, non-zero exit, timeout) continue to pass. This PR's diff for run.rs will be empty once #3016 lands and is rebased in.

oxoxDev:
## Merge order — epic #1535

For the maintainer's queue, the suggested merge order across the open #1535 sub-tasks:

1. **#2954** — `feat(agent/multimodal)` — independent, merge anytime
2. **#3016** — `feat(tools): generate_presentation` — independent of #2954, can merge in parallel
3. **#3017** — `feat(chat): ArtifactCard` — stacked on #3016 (branch built off `feat/2778-presentation-tool`); rebase + merge **after #3016** ✅ (this PR)
4. **#2780** (orchestrator wiring + grounding) — depends on #3016 + #3017 + new follow-up #3024
5. _#3024 (Files-in-this-chat panel) — will be a fresh PR after this one, stacked on this branch_

This PR needs #3016 in `main` first — it consumes `artifacts::store::{create,finalize,fail}_artifact` + `ArtifactMeta.error` introduced there. Will rebase onto `main` once #3016 merges.

graycyrus (commented):
@oxoxDev the code looks solid to me, but two CI checks are failing so I'll hold the approval until those are resolved.

**CI failures**

Two jobs are red on cf8172a:

- `Rust Core Coverage` — `inference_voice_http_round23_raw_coverage_e2e` panics at line 124 with `assertion failed: ids.contains(&"reasoning-v1".to_string())`. This test is in voice/HTTP inference territory and looks unrelated to the artifact changes, but the fact that `orchestrator/agent.toml` was modified in this PR means it's worth a second look. Can you confirm whether this test was already failing on the base branch (`feat/2778-presentation-tool`) before this PR was cut, or if this PR's `agent.toml` change is implicated?

- `E2E lane 1` — the core binary hits SIGABRT during teardown after 16 tests pass. Looks like a pre-existing crash, but worth confirming it reproduces on the base branch too.

Once both are green I'll come back and approve.

**Acceptance criteria gaps (non-blocking, but worth noting)**

AC#3 says "Download button triggers a save-file dialog". You've explained the `tauri-plugin-dialog` / `schemars` conflict clearly and the Downloads + reveal pattern is reasonable, but this is a stated criterion miss. Please make sure issue #2779 is updated to reflect the intentional deviation, and that the dialog follow-up is tracked (linked issue or TODO comment pointing at the conflict) so it doesn't get lost.

AC#1 says "ArtifactCard renders for any ChatToolResultEvent containing an artifact_id". Right now the in-progress state only shows when the backend socket event arrives — there's no spinner during generation. The PR description acknowledges this; just make sure the follow-up work is tracked so the card's in-progress state actually wires up from `ChatToolCallEvent` before the parent #1535 is closed.

**Code itself**

The implementation is clean. `ArtifactSurfaceSubscriber` mirrors `ApprovalSurfaceSubscriber` correctly. The `#[non_exhaustive]` on `DomainEvent` variants is the right call. Tauri command validation (absolute path, NUL bytes, separators, dot-dot) is thorough. The `sanitize_filename_stem` in `store.rs` is correct and the 80-char cap is sane. Log redaction on the `fail_artifact` reason is done right — `reason_len` in structured logs, full payload in the socket event where it belongs. i18n passes at 0 missing / 0 extra across 13 locales.

One minor thing: `finalize_artifact`'s idempotency guard checks both `status == Ready` AND `size_bytes == size_bytes`. If `finalize_artifact` is called twice on a Ready artifact with a different `size_bytes` (shouldn't happen in practice, but), it will re-emit `ArtifactReady` and flip the UI card. The guard should probably be `if matches!(meta.status, ArtifactStatus::Ready)` unconditionally to make the idempotency promise hold regardless of size. Low risk but easy to fix.

oxoxDev:
@graycyrus thanks for the careful read.

Follow-up issue filed for the two AC gaps you flagged: #3162 — covers both the native save-file dialog (AC#3, blocked on `tauri-plugin-dialog`/`schemars` conflict) and the `ChatToolCallEvent`-driven in-progress state (AC#1) so neither is dropped on the floor before #1535 closes.

On the **finalize_artifact idempotency guard** — fair catch, will tighten that to `if matches!(meta.status, ArtifactStatus::Ready)` unconditionally in the next push so a size_bytes drift can't silently re-emit `ArtifactReady`.

On the **two CI failures**:
- `Rust Core Coverage` (`inference_voice_http_round23_raw_coverage_e2e` panic at L124, `reasoning-v1` ids assertion) — pre-existing on `main` since at least 4b26267f, confirmed by the same shape failing on every open PR (#2954 / #3016 / #3026 / #3029 / #3150 / #3153) regardless of scope. Independent of the `agent.toml` change on this PR.
- `Playwright lane 1/4` — `chat-harness-subagent.spec.ts:136` 50s timeout cascades into `ECONNREFUSED 127.0.0.1:17788` for the rest of the lane. Regressed on `main` after PR #3055. Tracked + quarantined in #3154 (open).

This PR now matches the merge-accepted floor of the recently-landed #3150 and #3153.

graycyrus (commented):
@oxoxDev — continuation from review 1.

**New commit (9c0d034)**: `#[cfg(any(target_os = "macos", target_os = "linux"))]` on the `tauri::generate_handler!` entry is the correct fix — the module gate and the handler registration now match, so Windows builds won't fail with a missing-symbol error. Clean.

**Upstream merge (97d6bc5)**: `Conversations.tsx` picked up a composer auto-resize tweak from main (cap at 96 px, then scroll). That's an additive upstream change — no issue.

**Still open from review 1**: the `finalize_artifact` idempotency guard in `store.rs` — `matches!(meta.status, ArtifactStatus::Ready) && meta.size_bytes == size_bytes` — hasn't been addressed. A second call with a different `size_bytes` on an already-Ready artifact bypasses the early return, re-saves meta, and re-emits `ArtifactReady`. The size equality check is wrong here; the only guard that matters is `status == Ready`. Please drop the `&& meta.size_bytes == size_bytes` clause.

**CI**: Rust Core Coverage + E2E lane 1 are still failing (same as review 1). Fix the idempotency guard and clear CI and I'll approve.

graycyrus (changes requested):
CI is all green now, and the ops.rs workspace_dir fix (post-#3074 refactor) is correct. One thing still blocking approval — the finalize_artifact idempotency guard I flagged in the last two reviews is still unchanged.

In store.rs, the early-return guard reads:

```rust
if matches!(meta.status, ArtifactStatus::Ready) && meta.size_bytes == size_bytes {
```

The size_bytes equality check makes idempotency conditional on matching size — if finalize_artifact is called twice on the same artifact with a different size, it re-emits ArtifactReady and can flap the UI. The guard should be on status alone:

```rust
if matches!(meta.status, ArtifactStatus::Ready) {
```

One-line fix. Everything else is solid — once this is in, I'll approve.

graycyrus (commented):
High-level review of the UI piece (#2779). This closes the loop nicely — the `ArtifactCard` above the composer (mirroring `ApprovalRequestCard`) is consistent with the existing chat surface, the three-state design (spinner / ready+Download / failed+reason) is clean, dark/light variants and design tokens look right, and the i18n is done properly (10 keys × all 13 locales with real translations). The `args`-envelope reuse to avoid touching `WebChannelEvent` and the Downloads+reveal fallback (given the `tauri-plugin-fs`/`schemars` conflict) are both pragmatic.

My main concern is **stacking hygiene**, not the UI itself: this branch was cut off `feat/2778-presentation-tool` *before* #3016 swapped python-pptx for the native-Rust engine, so the diff still drags in the old Python engine + a CI step that installs `python-pptx`. After #3016 merges and this rebases onto main, those should disappear — flagging the specific files inline so the rebase can be verified. Recommend merging #3016 first, then rebasing this clean. A couple of smaller UX notes inline too.

.github/workflows/coverage.yml
```diff
@@ -93,6 +93,16 @@ jobs:
           key: core-coverage
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Install python-pptx for generate_presentation tool tests
```
graycyrus: This `pip install python-pptx` CI step directly contradicts #3016's final direction — that PR refactored the presentation tool to a **native-Rust `ppt-rs` engine with no Python at all**. This step (and the `generate_pptx.py` / `invoker.rs` / `script.rs` files it supports) only exists because this branch was cut before that refactor. After rebasing onto a merged #3016 this should be gone. Please confirm it's removed on rebase so we don't silently re-introduce the Python build dependency #3016 deliberately dropped.

src/openhuman/tools/impl/presentation/generate_pptx.py
```diff
@@ -0,0 +1,179 @@
+"""Generate a PowerPoint (.pptx) file from a structured slide spec on stdin.
```
graycyrus: Stale residue from the pre-refactor stack: this python-pptx script (plus `invoker.rs` and `script.rs` in this dir) was replaced by the native-Rust `ppt-rs` engine in #3016. It's only in this diff because the branch predates that refactor. It should vanish once this rebases onto a merged #3016 — calling it out so the rebase result gets a careful look and we don't end up with two competing engine implementations on main.

app/src/pages/Conversations.tsx
```diff
@@ -2145,6 +2147,24 @@ const Conversations = ({
...
+            // the just-generated deck without scrolling. Cards stay visible
+            // across turns until the thread is cleared. ArtifactCard handles
+            // its own download lifecycle (dialog → copy → "Saved to …").
+            const artifactThreadId = selectedThreadId ?? activeThreadId;
+            const artifacts = artifactThreadId ? (artifactsByThread[artifactThreadId] ?? []) : [];
+            if (artifacts.length === 0) return null;
+            return (
+              <div className="mb-2 flex flex-col gap-2">
+                {artifacts.map(artifact => (
+                  <ArtifactCard key={artifact.artifactId} artifact={artifact} />
```
graycyrus, line 2128: `ArtifactCard` is rendered without an `onRetry` prop, so the failed-state **Retry** button (which the component fully supports) never actually renders in-product. Either wire up a retry handler here (re-dispatch the producing tool call) or, if retry is intentionally out of scope for this PR, add a short comment noting the button is dead until a follow-up so reviewers don't assume it works.

Related: per the PR body the card currently flips straight from absent → ready/failed because nothing drives `upsertArtifactInProgressForThread` off the tool-call event yet — so the in-progress spinner state is effectively unreachable in normal use today. Worth a comment here or in the provider so the unused state is understood as a known follow-up, not a bug.

oxoxDev:
@graycyrus thanks for the thorough walkthrough — taking these six in order:

**1. Stale python residue in #3017's diff** — agreed, this is a rebase artifact. #3017 branched off #3016 *before* the engine swap (`e0e2a2e5`). Once #3016 merges, I'll rebase #3017 onto fresh main and the stale `generate_pptx.py`, `invoker.rs`, `script.rs`, and the `pip install python-pptx==1.0.2` step in `coverage.yml` will all disappear from this PR's diff. Will confirm the rebase landed clean before requesting merge. Net new UI surface shrinks substantially after that.

**2. Merge order** — same plan: 3016 → 3017 → 3026 → 3029. #3016 is now CI-green (18/18) and APPROVED post coderabbit re-review; will not request #3017 merge until #3016 is in.

**3. In-progress spinner / ChatToolCallEvent wiring** — tracked in follow-up issue #3162. Card today flips absent → ready/failed by design on this PR.

**4. Retry no-op** — fair catch. Just appended to #3162 with two implementation paths: cheap (`removeArtifact` reducer + dispatch on retry click) or full (persist originating `ChatToolCallEvent` args + re-dispatch RPC). Cheap path is one reducer + one wiring line; full path is real work. I'd rather land that as a focused follow-up than expand #3017's scope here.

**5. Native Save As dialog (tauri-plugin-dialog/schemars conflict)** — tracked in #3162.

**6. Diff size** — will collapse on rebase per #1.

Also addressed the **`finalize_artifact` idempotency guard** you flagged in the inline review — pushed in `91e9ab85` (drops the `size_bytes == size_bytes` check so the early-return is idempotent on status alone, can't re-emit `ArtifactReady` and flap the UI on a second finalize with a different size).

.github/workflows/coverage.yml
oxoxDev: Addressed in `f678b3b2c`. `.github/workflows/coverage.yml`'s `Install python-pptx for generate_presentation tool tests` step is removed (-10 lines). The step only existed to make the deleted `tests/presentation_tool.rs` integration test's success branch reachable; with that test gone, the pip install is pure waste. CI no longer pulls `python-pptx==1.0.2` per coverage run.

src/openhuman/tools/impl/presentation/generate_pptx.py
oxoxDev: Addressed via rebase + cleanup. After #3016 merged as `166fc6af6`, I rebased this branch (`ddf11a513` merge commit) and the upstream-side native-Rust engine took over the `presentation/{mod,tests,types}.rs` files. The python implementation files (`generate_pptx.py`, `invoker.rs`, `script.rs`) were deleted in that merge, then `f678b3b2c` purged the remaining residue (`tests/presentation_tool.rs` + the coverage.yml step). `grep -rln python-pptx --include='*.rs' --include='*.py' --include='*.yml'` against the branch now returns only historical engine-swap commentary in `engine.rs` / `mod.rs` rustdocs — no executable python residue.

app/src/pages/Conversations.tsx
oxoxDev: Addressed in `0289f68d9` — added an inline comment above the `ArtifactCard` render call noting that `onRetry` is intentionally omitted, what the two implementation paths look like (cheap `removeArtifact` reducer vs full re-dispatch), and that the wiring lives in follow-up issue #3162. The failed-card UI still surfaces the truncated error reason; the Retry button just stays hidden until #3162 lands so reviewers can tell at a glance that the omission is deliberate, not a bug.

graycyrus: Following up after the rebase onto the cleaned-up #3016 — thanks for that, it resolved the two stale-Python items I flagged (the `coverage.yml` `python-pptx` step and the `generate_pptx.py`/`invoker.rs`/`script.rs` residue are all gone, and CI is green now).

This one is still open, though, and it's independent of the rebase: `ArtifactCard` is still rendered here as `<ArtifactCard key={artifact.artifactId} artifact={artifact} />` with **no `onRetry` prop**. The component fully implements a failed-state **Retry** button, but because the prop is omitted at the only call site, that button never renders in-product — so a failed artifact is a dead end for the user.

Could you either:

1. **Wire it up** — pass an `onRetry={(artifactId) => …}` handler that re-dispatches the producing tool call (ideal, makes the failed state actionable), or
2. **If retry is intentionally out of scope for this PR** — drop a short comment here noting the Retry button is dormant until a follow-up, and ideally file/link that follow-up, so reviewers don't assume it works.

Related (same root cause, fine to fold in or defer): nothing currently drives `upsertArtifactInProgressForThread` off the tool-call event, so the in-progress spinner state is also unreachable in normal use today. A one-line comment marking it a known follow-up would keep it from reading as a bug.

Not blocking the architecture — just don't want a half-wired user-facing affordance to ship silently. 🙏

graycyrus (commented):
Re-review after the rebase onto the cleaned-up #3016. Good progress:

- ✅ The stale python-pptx CI step in `.github/workflows/coverage.yml` is gone.
- ✅ The leftover Python engine files (`generate_pptx.py`, `invoker.rs`, `script.rs`) are gone — the diff is now strictly the native-Rust producer + the artifact UI surface.
- ✅ No merge conflicts, CI green.
- ✅ The CodeRabbit items (platform-neutral reveal wording, redacted error logging, `ArtifactKind` runtime validation, web-channel log redaction) all landed.

**One item still open**, and it won't be fixed by any rebase — left a detailed note on the thread: `ArtifactCard` is rendered without an `onRetry` prop in `Conversations.tsx`, so the failed-state Retry button never renders, and a failed artifact becomes a dead end for the user. Please either wire up a retry handler or explicitly mark it a dormant/known follow-up with a comment. Same root cause leaves the in-progress spinner state unreachable too — a one-liner noting that is enough.

Once that's addressed (or consciously deferred with a comment) this is good to go from my side. Thanks for the quick turnaround on the rest. 🙏

oxoxDev:
@graycyrus thanks — three of four were already in by the time of your re-review; the fourth (one-liner about the in-progress state being unreachable too) is in now.

**Re your open item:**

- The `onRetry` omission comment landed in `0289f68d9` (likely pushed just after you wrote your re-review). Now extended in `93d8b4785` to also cover the spinner gap you flagged — restructured the call-site block as a numbered list with both intentional deferrals (retry button + in-progress state) and a single pointer to follow-up issue #3162.

#3162's body already enumerates both:

> 1. Save-file dialog (AC#3)
> 2. ChatToolCallEvent → in-progress card wiring (AC#1)
> 3. Retry button wiring

So the failed-state Retry button + the dispatch-time spinner share the same follow-up. Reviewers reading `Conversations.tsx:2108` now see both gaps documented and have a single ticket to track.

This should clear the last open item from your side. CI re-running on `93d8b4785`.

Issue #3024, opened 2026-05-30
Files in this chat — persistent per-chat artifact registry


oxoxDev:
## Summary

A persistent **"Files in this chat"** surface inside the chat view that lists every artifact the agent has generated in the current thread, with per-row Download / Reveal-in-folder / Delete affordances. Survives app restarts (for files created in-session) so the user can return to a chat and still grab a deck they made earlier.

## Problem

The inline `ArtifactCard` shipped in #2779 (PR #3017) sits above the composer as a single ephemeral card per artifact:
- Scrolling, sending the next turn, or restarting the app loses the card.
- The `artifactsByThread` Redux slice is in-memory only (`app/src/store/index.ts:158` registers `chatRuntimeReducer` directly, no `persistReducer`), so files vanish across app restarts even though the underlying payload is still on disk under `<workspace>/artifacts/<uuid>/`.
- There is no place to enlist all files involved in a chat.

User feedback (live dev:app test, 2026-05-30): _"there should be a place to enlist all the files involved in a chat"_ — flagged after a generated `climate-change-an-overview.pptx` was visible once, then the user couldn't find it again after a new turn was sent.

## Solution

Frontend-only addition layered on top of the artifact pipeline already shipped in #2779:

1. **`ChatFilesChip`** (new component) — header chip with paperclip icon + count badge. Hidden when artifact count = 0. Owns popover open/close + click-outside + Esc.
2. **`ChatFilesPanel`** (new component) — popover panel listing per-thread artifacts (title + kind icon + size + Download + Reveal + Delete with confirm modal). Uses `useT()` + Tailwind + inline SVGs (mirrors `ArtifactCard.tsx` style — no new icon dep).
3. **Persistence** — wrap `chatRuntimeReducer` with `persistReducer` + a `createTransform` that keeps only `status === 'ready'` entries on storage write (`in_progress` / `failed` stay session-scoped — should not survive a restart).
4. **Delete flow** — frontend `deleteArtifact()` service + `removeArtifactForThread` reducer; optimistic FE removal with rollback on RPC error. RPC `openhuman.ai_delete_artifact` already exists.
5. **Above-composer card on `ready`** — filter `status === 'ready'` OUT of the existing above-composer block (delegated to the panel). `in_progress` / `failed` keep showing there for live feedback.

### Known backend gap (deferred to a follow-up)

[...]

Issue #1535, opened 2026-05-12
Add multimodal support for files and PPT generation


Al629176:
## Summary

Add multimodal and presentation-generation support so the agent can reliably work with images/files and create PPT slides when users ask.

## Problem

Users are trying to ask the agent to create PPT slides or work from shared visual/file context, but the agent struggles and returns errors such as being unable to code in the sandbox or complete the presentation task. This indicates the agent experience lacks reliable multimodal input handling and a supported presentation-generation workflow.

This hurts workflows where users expect the assistant to use screenshots, images, documents, or instructions to produce slide decks and other rich outputs.

Constraints:
- Multimodal input should preserve user privacy and avoid leaking uploaded file contents unnecessarily.
- PPT generation needs a supported tool/runtime path rather than ad hoc agent guessing.
- Presentation output should be verifiable and downloadable/openable by the user.
- The app should provide clear errors when a requested file type or generation path is unsupported.

## Solution (optional)

Define supported multimodal inputs and wire them into the agent/tool runtime. Add a first-class presentation workflow for generating PPT/PPTX slides from user instructions and uploaded context, including sandbox/runtime support, artifact persistence, and user-visible progress/errors.

## Acceptance criteria

- [ ] **Multimodal input supported** — The agent can receive and reason over supported uploaded images/files in the relevant chat or task flow.
- [ ] **PPT workflow available** — The agent can create a PPT/PPTX slide deck from user instructions through a supported tool/runtime path.
- [ ] **Artifacts delivered** — Generated presentations are saved and surfaced to the user in a usable format.
- [ ] **Sandbox path fixed** — Presentation generation no longer fails with generic sandbox/code execution errors for supported requests.
- [ ] **Clear unsupported states** — Unsupported file types, missing tools, or failed generation steps show actionable errors.
- [ ] **Privacy safe** — Uploaded file contents and generated artifacts are handled within the expected user/workspace boundaries.
- [ ] **Regression safety** — Unit, integration, or E2E coverage verifies multimodal attachment handling and a basic PPT generation flow where testable.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)).

## Related

- User feedback screenshot from May 12, 2026 requesting multimodal support and reliable PPT slide creation.

graycyrus:
## Sub-tasks

[...]

PR #3029, opened 2026-05-30
feat(orchestrator): wire generate_presentation + require grounding (#2780)

### LINKED ISSUE #2780: Wire PPT generation and file attachments into agent definitions and error handling

oxoxDev:
## Summary

- Wires `generate_presentation` into the orchestrator with a **grounding rule** that forbids invented slide content on factual decks. Topical / factual decks MUST call `memory_tree` / `research` / `query_memory` BEFORE `generate_presentation`; structural decks + pasted-source decks are waived.
- Refreshes the `generate_presentation` agent.toml comment to point at the grounding rule + pin test (the tool entry itself shipped with #3016).
- Adds the third `PresentationError` variant — `UnsupportedFileType` — called for by the #2780 scope. The other two (`MissingRuntime`, `GenerationFailed`) already shipped with #2778. Reserved for the future `format` selector; not constructed by `execute` today.
- Pins both invariants with structural tests so a future prompt / TOML refactor cannot quietly drop the guardrail.

## Problem

Live dev:app trace on PR #3017 (2026-05-30): asking the agent for "5 slides on current AI trends" produced fully-confabulated bullets + speaker notes with zero retrieval:

```
[agent] dispatching 1 tool(s): ["generate_presentation"]
[agent_loop] tool start name=generate_presentation
[agent_loop] tool finish name=generate_presentation elapsed_ms=225 success=true
```

One tool call. No `memory_tree`, no `research`, no source verification. Real hallucination risk for accuracy-sensitive decks (analyst pitches, conference talks, briefing decks). Resolved model = `reasoning-v1`; the same prompt at higher temperatures would diverge even further.

The companion gap on the error-handling side: `PresentationError` was missing `UnsupportedFileType`, so downstream mappers had no exhaustive variant to pattern-match for the planned `format` selector follow-up.

## Solution

Four load-bearing edits, one micro-commit each:

1. **`UnsupportedFileType` error variant** (`src/openhuman/tools/impl/presentation/types.rs`) — `#[allow(dead_code)]` since `execute` does not construct it yet; reserved for the planned `format` selector that will let callers request `.pdf` / `.key` / image strips. Adds a Display test in `tests.rs` so the rendered phrasing (`"unsupported file type 'key'; supported: pptx"`) is pinned and can be surfaced verbatim by a downstream mapper.

2. **Grounding rule in orchestrator prompt** (`src/openhuman/agent_registry/agents/orchestrator/prompt.md`) — new `## Presentation generation` section: before any topical/factual `generate_presentation` call the agent MUST first dispatch one of `memory_tree` (user's ingested history), `research` (live web facts), or `query_memory` (prior thread / saved memory), and the slide bullets / body / speaker_notes MUST be drawn from the grounding output. Explicit waiver list: pasted-source decks, prior-turn source decks, content-free / structural decks, user-explicit waiver. Companion test `prompt::tests::build_requires_grounding_before_generate_presentation` pins the section header, the "do not skip" instruction, all three permitted grounding tools by name, and the pasted-source waiver — a future prompt refactor cannot quietly remove the guardrail.

3. **Wiring contract tests** (`tests/orchestrator_presentation_wiring.rs`) — two structural invariants on the agent definitions: (a) orchestrator agent.toml MUST list `generate_presentation` (without it the function-calling schema drops the tool and the "create slides" routing case from #1535 silently fails); (b) code_executor agent.toml MUST NOT list `generate_presentation` (pptx rendering has its own `runtime_python` venv and exposing it to code_executor would create a duplicate dispatch path that bypasses the orchestrator grounding rule). Exact-line matching, not substring, so commented-out or prefixed-name entries cannot satisfy accidentally.

4. **Grounding rule alignment with researcher `delegate_name`** (CodeRabbit follow-up; commit `0cc9785f`) — the researcher agent overrides its `delegate_name` to `research` (`agent_registry/agents/researcher/agent.toml`), so the orchestrator surfaces it as `research`, not `delegate_research`. The original grounding rule referenced a non-existent `delegate_research` tool — the model would either hallucinate the call or skip grounding entirely. Updated both grounding-rule references in `prompt.md`, the inline comment in `agent.toml`, and tightened the regression test to (a) assert `research` is present as a grounding option, (b) assert the stale literal `delegate_research` is absent from the rendered body — so any regression to the wrong name fails fast.

### Why no `tool_filter.rs` change

The original issue scope called for "update `src/openhuman/agent/harness/tool_filter.rs` to allow the new tool under appropriate policies". That file is actually the **Composio toolkit fuzzy ranker** — it ranks Composio actions against a task prompt for `integrations_agent` and has no per-tier tool gating role. The real gating mechanism for `generate_presentation` is the orchestrator's `[tools].named` allow-list (already added in #3016) plus the orchestrator system prompt (the grounding rule shipped here). Nothing in `tool_filter.rs` needs to change.

### Why no `tool_loop.rs` error-mapping change

`PresentationError` already implements `thiserror::Error` with `#[error("…")]` Display impls, and the call sites in `presentation/mod.rs` (`execute`) feed `err.to_string()` into `ToolResult::error(…)` — which `tool_loop.rs` already surfaces to the model as `"Error: <msg>"`. No additional mapping layer needed.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 1 new tests-file integration test (`orchestrator_presentation_wiring`, 2 cases), 1 new prompt-content test (`build_requires_grounding_before_generate_presentation`), 1 new error-variant Display test (`unsupported_file_type_display_includes_extension_and_supported`).
- [x] **Diff coverage ≥ 80%** — pure-config + prose edits + one Rust error-enum variant; every changed Rust line is hit by the three new tests (variant construction + display + orchestrator agent.toml line read + prompt.md substring assertions).
- [x] N/A: Coverage matrix updated — behaviour-only change to existing agent-definition + tool error enum surfaces; no new feature-matrix row.
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — pure-Rust changes; no HTTP / RPC surface added.
- [x] N/A: Manual smoke checklist updated — doesn't touch release-cut surfaces (orchestrator system-prompt text + one reserved error variant).

## Impact

- **Platform**: desktop (Rust core). 0 frontend changes, 0 Tauri permission changes.
- **Runtime behaviour**: orchestrator system prompt grows by ~25 lines. Hits every orchestrator turn but stays well inside the chat-tier 128k context budget and is cache-friendly (frozen per session, sits in the static `prompt.md` archetype prefix).
- **LLM dispatch**: factual / topical "make slides on X" prompts will now dispatch **2** tool calls instead of **1** (one grounding + one `generate_presentation`). Cost trade-off is intentional — a single `research` or `memory_tree` call adds ~1-3 s latency but eliminates the fabricated-bullets failure mode that motivated the issue. Structural / pasted-source decks remain single-call.
- **Migration**: none — `UnsupportedFileType` is additive on a `#[non_exhaustive]`-spirit error enum; existing callers do not pattern-match exhaustively. The orchestrator's `[tools].named` list was already updated in #3016.
- **Security**: no privilege change. Grounding tools (`memory_tree`, `research`, `query_memory`) already exist in the orchestrator's allow-list; the rule constrains the **order** of dispatch, not the surface.

## Related

- Parent epic: #1535 "Add multimodal support for files and PPT generation"
- Tool itself shipped in #2778 (PR #3016 — `feat/2778-presentation-tool`). This PR is stacked on top; will rebase onto `main` once #3016 merges.
- Companion UI surface: #2779 (PR #3017 ArtifactCard) + #3024 (PR #3026 Files-in-this-chat panel) — both surface decks the grounded `generate_presentation` produces.
- Closes #2780

Suggested merge order across the open #1535 sub-tasks:
1. **#2954** — `feat(agent/multimodal)` — independent
2. **#3016** (#2778) — `feat(tools): generate_presentation` — independent of #2954
3. **#3017** (#2779) — `feat(chat): ArtifactCard` — stacked on #3016
4. **#3026** (#3024) — `feat(chat): Files-in-this-chat panel` — stacked on #3017
5. **This PR** (#2780) — `feat(orchestrator): generate_presentation wiring + grounding` — stacked on #3016

---

### Linear Issue

- Key: N/A (workflow uses GitHub issues exclusively)
- URL: N/A

### Commit & Branch

- Branch: `feat/2780-orchestrator-presentation-wiring`
- Commits: 4 micro-commits, GPG-signed by oxoxDev <nikhil@tinyhumans.ai>
- AI tool: Claude Code (claude-opus-4-7)

## Summary by CodeRabbit

* **New Features**
  * Added presentation generation tool that creates .pptx decks from structured slide specs (titles, body, bullets, speaker notes) and returns artifact id/path, slide count and size.
  * Tool integrated into orchestrator with a required grounding step (with explicit exceptions) before topical/factual deck generation.
* **Bug Fixes / Reliability**
  * Artifacts now record lifecycle/status (pending → ready/failed) and include failure reasons; generation reports clearer errors for timeouts, missing runtime/package, and truncated stderr.


oxoxDev:
## Suggested merge order — #1535 epic chain

To be merged **after #3016 (#2778)** — depends on the `generate_presentation` tool existing in the orchestrator's allow-list.

Full chain:
1. #2954 — `feat(agent/multimodal)` — independent
2. #3016 (#2778) — `feat(tools): generate_presentation` — independent of #2954
3. #3017 (#2779) — `feat(chat): ArtifactCard` — stacked on #3016
4. #3026 (#3024) — `feat(chat): Files-in-this-chat panel` — stacked on #3017
5. **This PR (#2780)** — orchestrator wiring + grounding — stacked on #3016

Will rebase onto `main` once #3016 merges. No file overlap with #3017 / #3026 (different module: `agent_registry/agents/orchestrator/` + `tools/impl/presentation/types.rs` only).

src/openhuman/agent_registry/agents/orchestrator/prompt.md
```diff
@@ -150,6 +150,27 @@ Modes:
...
+
+`generate_presentation` builds a `.pptx` deck from a structured slide spec via the bundled python-pptx helper. Use it for any "make slides", "build a deck", "draft a presentation", "create a pitch" request.
+
+**Grounding rule (do not skip).** Before calling `generate_presentation` on a topical or factual deck — anything where the slides need real-world facts, current events, statistics, names, dates, quotes, or domain context — you MUST first establish a grounding context. Pick at least one:
+
+- `memory_tree` (`query_topic` / `query_source` / `query_global`) — when the topic plausibly lives in the user's ingested history (their notes, prior chats, emails on the subject).
+- `delegate_research` — when the topic needs live web facts (current events, recent stats, comparative product data, anything time-sensitive).
+- `query_memory` — when the user has previously summarised the exact topic in this thread or in a saved memory.
+
+Only after the grounding tool returns may you call `generate_presentation`, and the slide bullets / body / speaker_notes you pass MUST be drawn from the grounding output — not invented from priors.
```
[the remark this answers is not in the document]
oxoxDev: Addressed in 0cc9785f — renamed `delegate_research` → `research` in three places: (a) the "Pick at least one" grounding bullet in `prompt.md`, (b) the "Why this rule exists" paragraph in `prompt.md`, (c) the inline comment in `agent.toml` next to the `generate_presentation` tool entry. The researcher agent overrides its `delegate_name` to `research` (verified in `src/openhuman/agent_registry/agents/researcher/agent.toml`), so the orchestrator surfaces it as `research`, not `delegate_research`. The pre-existing `delegate_research` literals in `loader.rs` (lines 721, 814) are not real tool names — they're entries in *forbidden-tools* assertion lists for `crypto_agent` / `markets_agent`, asserting those leaf agents must NOT have that tool name; they don't represent any tool the orchestrator actually exposes. Tightened the regression test in `prompt.rs` to assert both `research` is present AND the stale `delegate_research` literal is absent from the rendered body — so any future regression to the wrong name fails fast.

tests/presentation_tool.rs
```diff
@@ -0,0 +1,139 @@
...
+    let Some(python) = resolved_python() else {
+        eprintln!("skipping: no python3 on PATH");
+        return;
+    };
+    if !python_pptx_importable(python) {
+        eprintln!(
+            "skipping: {python} cannot `import pptx` — install with `pip install python-pptx==1.0.2`"
+        );
+        return;
+    }
```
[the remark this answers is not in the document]
oxoxDev: Addressed in ca4a1b4e — switched the skip gate to call `runtime_python::detect_system_python(&default_minimum_version, None)` directly, so the test pins to the exact resolver `ensure_venv` uses (candidate order `python3.12` → `python3` → `python`, minimum version 3.12.0). Dropped the host-level `python -c "import pptx"` check entirely: `ensure_venv` creates an isolated venv (no `--system-site-packages`) and pip-installs `python-pptx==1.0.2` itself, so host importability was never a meaningful gate — it could only cause spurious skips on hosts where the runtime would have worked fine. Both tests still pass locally against host python3.12 with no host-side pptx install. Updated module docstring to spell out the new contract.

src/openhuman/agent_registry/agents/orchestrator/prompt.rs
```diff
@@ -325,6 +325,56 @@ mod tests {
...
+            "grounding rule missing from presentation section"
+        );
+        assert!(
+            body.contains("Before calling `generate_presentation`"),
+            "grounding pre-condition phrasing missing from presentation section"
+        );
+        // Each of the three permitted grounding tools must be named so
+        // the model knows the menu, not just the constraint.
+        assert!(body.contains("`memory_tree`"));
+        assert!(
```
graycyrus, line 371: **[major]** Test pins `research` presence and `delegate_research` absence — good. But none of the four grounding waiver cases are asserted. A future prompt change that silently drops or extends the escape-hatch list (weakening the hallucination guardrail) passes CI today.

Add substring assertions for the full waiver list, e.g.:
```rust
assert!(body.contains("user pasted source material"), "pasted-source waiver must be present");
assert!(body.contains("structural") || body.contains("layout-only"), "structural waiver must be present");
assert!(body.contains("user explicitly waive"), "user-explicit waiver must be present");
```
And consider adding a count assertion so adding new escape hatches requires updating the test.

src/openhuman/tools/impl/presentation/types.rs
```diff
@@ -0,0 +1,229 @@
...
+            .body
+            .as_deref()
+            .map(|s| !s.trim().is_empty())
+            .unwrap_or(false);
+        let has_bullets = !slide.bullets.is_empty();
+        if !has_title && !has_body && !has_bullets {
+            return Err(PresentationError::InvalidInput {
+                field: format!("slides[{i}]"),
+                reason: "must have at least one of title / body / bullets".to_string(),
+            });
```
graycyrus, line 186: **[major]** `MAX_BULLETS_PER_SLIDE` is wired into the JSON schema (`maxItems`) but `validate_input()` never checks it. Add a per-slide loop in `validate_input` before returning `Ok(())`:
```rust
for (i, slide) in input.slides.iter().enumerate() {
    if slide.bullets.len() > MAX_BULLETS_PER_SLIDE {
        return Err(PresentationError::InvalidInput {
            field: format!("slides[{i}].bullets"),
            reason: format!("exceeds {MAX_BULLETS_PER_SLIDE}-bullet limit"),
        });
    }
}
```
Keeps validation consistent with how the other field limits are handled.

src/openhuman/tools/impl/presentation/generate_pptx.py
```diff
@@ -0,0 +1,187 @@
...
+        "--output",
+        required=True,
+        help="absolute path where the .pptx should be written",
+    )
+    args = parser.parse_args()
+
+    # Defence-in-depth: the Rust caller always supplies an absolute
+    # artifact path, but a future regression there must not silently
+    # redirect output to a cwd-relative location.
+    if not os.path.isabs(args.output):
```
graycyrus: **[minor]** `os.path.isabs()` accepts paths like `/tmp/../../etc/shadow`. The Rust side sanitizes with `assert_within_root` before constructing this argument, so it's not exploitable today. But cheap to harden independently:
```python
import pathlib
resolved = pathlib.Path(args.output).resolve()
if not str(resolved).startswith(EXPECTED_WORKSPACE_PREFIX):
    _exit_error("output path escapes workspace")
```
Or at minimum swap `os.path.isabs(args.output)` for `pathlib.Path(args.output).resolve() == pathlib.Path(args.output)` to catch traversal components.

graycyrus (changes requested):
Solid feature — the grounding rule, wiring tests, and UnsupportedFileType variant are well-structured. A few things need addressing before this merges.

The delegate_research → research rename (CodeRabbit #1) and python probe alignment (CodeRabbit #2) are both confirmed fixed, so I'm not touching those.

**Issue alignment check (#2780)**
AC: orchestrator tool scope ✓, UnsupportedFileType variant ✓, error tests ✓, diff coverage ✓, tool_filter.rs waived with solid reasoning ✓. The structural wiring tests + presentation_tool.rs together satisfy the integration test AC in spirit.

---

**[major] Grounding rule test incomplete — missing three of four waiver cases**

The test in prompt.rs pins the section header, 'do not skip' instruction, the three grounding tools, and the pasted-source waiver case. Good. But it doesn't assert the other three escape hatches: prior-turn material, structural/layout-only decks, and user-explicit waiver ('don't research, just use priors'). A future prompt edit that quietly drops or expands the waiver list would pass CI undetected.

This is safety-critical — the whole point of this PR is to prevent the hallucination trace from 2026-05-30. Add substring assertions for all four waiver cases so the guardrail stays complete.

---

**[minor] generate_pptx.py output path check is incomplete**

The absolute-path check blocks relative paths, but os.path.isabs() returns True for paths like /tmp/../../etc/shadow. I know the Rust side sanitizes the path with assert_within_root before constructing the --output arg, so this isn't exploitable today. But defense-in-depth here is cheap — swap the check for pathlib.Path(args.output).resolve() and then confirm the resolved path still starts with the expected workspace prefix before writing. Makes the Python side independently safe rather than depending on Rust to catch it first.

**[minor] Cached venv path returns empty version string in venv.rs**

try_cached_venv returns ResolvedPython { version: String::new(), .. }. Any caller that logs python_bin + version (for debugging or error messages) will emit incomplete version info. Either probe it once on venv creation and embed it in the lock file, or at minimum document that version is intentionally empty for cached paths so future callers know not to trust it.

oxoxDev:
@graycyrus thanks for the review. Three points:

**[major] Grounding rule waiver test** — addressed in `a8169bb9`. The existing test pinned only the pasted-source waiver; now pins all four (pasted-source / prior-turn / structural / explicit user waiver) with a substring assertion per clause + a comment noting which escape hatch it covers. The safety-critical guardrail no longer decays silently on a prompt edit.

**[minor] `generate_pptx.py` path-traversal hardening** — this file is removed in PR #3016's engine swap (`e0e2a2e5`: native Rust `ppt-rs`, no Python subprocess). #3029's diff currently shows the file because this branch was cut before that refactor landed. After #3016 merges and #3029 rebases onto fresh main, the file disappears from the diff entirely. The concern auto-resolves — confirming that on the rebase.

**[minor] `venv.rs` cached path returns empty version string** — same story: `runtime_python/` module is gone post-#3016 engine swap. Will disappear from this PR's diff on rebase. No active code change needed.

Merge order: #3016 → #3017 → #3026 → #3029. #3016 is CI-green and APPROVED post coderabbit re-review; will rebase #3029 once that lands.

src/openhuman/agent_registry/agents/orchestrator/prompt.rs
oxoxDev: Addressed in `a8169bb9`. The test now pins all four waiver clauses in the grounding rule:

```rust
// (1) user pasted source material in the same turn.
assert!(body.contains("user pasted source material"), "...");
// (2) prior turn in same thread already established source material.
assert!(body.contains("prior turn in this same thread"), "...");
// (3) structural / content-free deck (blank template, layout-only).
assert!(body.contains("content-free or structural"), "...");
// (4) explicit user waiver ("don't research, just generate…").
assert!(body.contains("explicitly waived grounding"), "...");
```

A future prompt edit that quietly drops or expands the waiver list now fails CI on the matching assertion. `cargo test --lib openhuman::agent_registry::agents::orchestrator::prompt` passes 10/10.

src/openhuman/tools/impl/presentation/types.rs
oxoxDev: Addressed in `5218a1490` (the merge commit that rebased this branch onto post-#3017 main, picking up the cleaned types.rs from #3016's engine swap). The validation you described is now at `src/openhuman/tools/impl/presentation/types.rs:202-207`:

```rust
if slide.bullets.len() > MAX_BULLETS_PER_SLIDE {
    return Err(PresentationError::InvalidInput {
        field: format!("slides[{i}].bullets"),
        reason: format!("must contain ≤ {MAX_BULLETS_PER_SLIDE} bullets"),
    });
}
```

Runs as part of `validate_input` so a slide with >32 bullets is rejected before the engine is invoked. Matches the JSON schema's `maxItems` so the LLM-facing contract and the server-side check are in sync.

src/openhuman/tools/impl/presentation/generate_pptx.py
oxoxDev: Auto-resolved by rebase. After #3016 merged as `166fc6af6` (native-Rust `ppt-rs` engine), `src/openhuman/tools/impl/presentation/generate_pptx.py` was deleted entirely — there's no more Python subprocess in the presentation flow. This branch's merge commit `5218a1490` picks that up, so the file is gone from #3029's diff. `os.path.isabs()` concern doesn't apply since the code path no longer exists.

The Rust side's `assert_within_root` guard you mentioned is still in `src/openhuman/artifacts/store.rs` and continues to gate the artifact output path — defense-in-depth on the writer side stayed intact through the engine swap.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
```diff
@@ -189,6 +189,27 @@ Modes:
 
 Start cheap (query_* summaries), only drill_down/fetch_leaves when you need verbatim content.
 
+## Presentation generation
+
+`generate_presentation` builds a `.pptx` deck from a structured slide spec via the bundled python-pptx helper. Use it for any "make slides", "build a deck", "draft a presentation", "create a pitch" request.
```
graycyrus: **[blocking] Stale — the tool is native Rust, not Python.** On `main`, `presentation/engine.rs` is a native `ppt-rs` generator that explicitly *"replaces the python-pptx subprocess… no Python runtime, no subprocess"*, and `generate_pptx.py` is gone. Please reword "via the bundled python-pptx helper" to describe the native Rust engine. Same fix needed for the `runtime_python` venv reference in the `tests/orchestrator_presentation_wiring.rs` comment. Beyond accuracy, telling the model the tool shells to Python is misleading.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
```diff
@@ -189,6 +189,27 @@ Modes:
...
+
+**Grounding rule (do not skip).** Before calling `generate_presentation` on a topical or factual deck — anything where the slides need real-world facts, current events, statistics, names, dates, quotes, or domain context — you MUST first establish a grounding context. Pick at least one:
+
+- `memory_tree` (`query_topic` / `query_source` / `query_global`) — when the topic plausibly lives in the user's ingested history (their notes, prior chats, emails on the subject).
```
graycyrus: **[blocking] Reintroduces dead modes — regresses #3275.** `query_topic` and `query_global` were removed from this prompt in #3275 (merged) because the `memory_tree` dispatcher rejects them with `unknown mode`. The real modes are `query_source` / `smart_walk` / `walk` (+ `search_entities` / `drill_down` / `fetch_leaves` / `ingest_document`). Re-adding `query_topic` / `query_global` here puts the orchestrator back to emitting guaranteed-failing calls. Please reference only the live modes — `smart_walk` is the best default for an open-ended topical lookup.

graycyrus (changes requested):
### Requesting changes — the new prompt text is stale and regresses `main`

The wiring + tests in this PR are genuinely strong (the exact-line `agent.toml` invariants are load-bearing, and the four-waiver grounding test now pins all the cases I asked about — thank you). But the **prompt copy this PR adds is out of date** with what's already on `main`, in two ways that are individually shippable bugs. CI is green only because the tests assert *string presence*, not correctness against the live tool/dispatcher — so neither issue is caught.

**1. It tells the model the tool is Python — but `main` is native Rust.**
The presentation tool was refactored to a native `ppt-rs` engine (`presentation/engine.rs`: *"Native Rust .pptx generator (replaces the python-pptx subprocess)… no Python runtime, no subprocess"*); `generate_pptx.py` no longer exists on `main`. This PR's prompt still describes "the bundled python-pptx helper", and the wiring-test comment still says code_executor "has its own managed `runtime_python` venv". Both need to describe the native engine.

**2. It reintroduces dead `memory_tree` modes that #3275 just removed.**
#3275 (merged) deleted `query_topic` / `query_global` from this exact prompt because the dispatcher rejects them with `unknown mode` — the real modes are `search_entities` / `query_source` / `drill_down` / `fetch_leaves` / `ingest_document` / `walk` / `smart_walk`. This PR's grounding rule re-adds `query_topic` / `query_global`, which would re-break what #3275 fixed and send the orchestrator back to emitting guaranteed-failing calls.

**Also:** this branch is based before #3275, so it still carries the old `mode: "query_topic"` / `"query_global"` definitions at lines ~184/186 (outside this diff). Please **rebase onto current `main`** and reconcile — after the rebase those stale definitions are gone and the grounding rule should reference only the real modes.

None of this is about the feature design (the grounding rule is the right idea) — it's purely stale content from the branch predating both the native-engine merge and #3275. Two small edits + a rebase and this is good to go.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
oxoxDev: Addressed in 6284d07aa — prompt.md:196 now reads "native Rust engine (`ppt-rs`) running in-process — no Python subprocess, no managed venv", matching the `engine.rs` rustdoc. Wiring-test comment in `tests/orchestrator_presentation_wiring.rs:12-17` and rustdoc in `presentation/types.rs:127-129` also updated to drop `runtime_python` venv references and reference the native engine instead.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
oxoxDev: Addressed in 6284d07aa — branch rebased onto current upstream/main (#3275 cleanup absorbed). The grounding-rule line at prompt.md:200 now lists only `query_source` / `smart_walk` for memory_tree, dropping the dispatcher-rejected `query_topic` / `query_global` modes. Restricted to the two natural-language retrieval modes that fit the "topic plausibly lives in user's history" framing; drill-down/fetch-leaves still available later in the prompt for citation.

graycyrus:
Thanks for the fast turnaround — both blocking items are resolved. ✅

- The prompt now describes the **native Rust `ppt-rs` engine running in-process (no Python subprocess, no managed venv)**, and the `runtime_python` venv wording in `tests/orchestrator_presentation_wiring.rs` was corrected too.
- The grounding rule now lists the **real `memory_tree` modes (`query_source` / `smart_walk`)** — `query_topic` / `query_global` are gone everywhere in the prompt, so the #3275 regression is cleared.

The feature content is good to go. One cleanup left before I flip my review:

**The PR is `CONFLICTING` and has picked up unrelated version/build files.** The two `Merge upstream/main` commits + the `restore version 0.57.12` commit dragged `Cargo.toml`, `Cargo.lock`, `app/package.json`, and `app/src-tauri/{Cargo.toml,Cargo.lock,tauri.conf.json}` into the diff — none of which a prompt + tests PR should touch (changed files went 5 → 11).

Could you **rebase onto current `main`** rather than merging it in? That should both resolve the conflict and drop the version/lockfile/`tauri.conf.json` churn, returning the diff to just `prompt.md` + `prompt.rs` + `presentation/{types,tests}.rs` + the wiring test.

Once it's rebased clean and green, ping me — the substance is done, so I'll approve. 🙏

PR #3032, opened 2026-05-30
perf(memory_tree): batch drill_down BFS walker per depth

mysma-9403:
## Summary

`memory_tree::retrieval::drill_down::walk_with_embeddings` previously issued **one `get_summary` + one `get_tree` per visited node** (and on the leaf path additionally **one `get_chunk` + one `get_chunk_embedding`**). For a deep-dive over a sealed L2 summary with fan-out 10 and `max_depth=2`, that's `~10 × 2` SQLite round-trips at L1 plus `~100 × 4` at L0 — fully sequential, on the hot retrieval path that fires on **every "tell me more" follow-up** in the chat agent.

Refactor the walker to **BFS-by-level**: collect every node id at the current depth, then issue at most four batched `SELECT … WHERE id IN (?,?,…)` queries for the whole level (summaries / trees / chunks / chunk_embeddings) before walking the slice to push hits + collect the next-depth frontier. Round-trip count goes from `O(nodes × 4)` → `O(depth × 4)`.

All four batch helpers reused **as-is** — no new store API needed:

- `get_summaries_batch`        (landed in #2948)
- `get_trees_batch`            (landed in #2976)
- `get_chunks_batch`           (landed in #2941)
- `get_chunk_embeddings_batch` (pre-existing)

This is the first pure call-site refactor in the batch-fetch series: every helper this PR needs was already shipped by the prior four.

- **BFS order preserved.** Walking `current_level` in iteration order with `next_level.extend(child_ids)` is equivalent to the prior `VecDeque<pop_front, push_back>` FIFO. PR #831's BFS-vs-DFS regression test still passes (`walk_visits_siblings_before_descendants_bfs_order`).
- **HashMap keyed by id, not position.** Per-id lookups use `&id`, never `enumerate()` index — a refactor that mistakenly used position would silently shadow one sibling's `Tree`/scope with another's. Pinned by a new regression test (see below).
- **`Ok(None)` contracts preserved** across all four helpers: missing rows stay silently absent from the `HashMap`; `.get(id).cloned()` on the embeddings map yields the equivalent `Option<Vec<f32>>` the per-row `get_chunk_embedding` returned. The legacy "no embedding row" path is bit-identical.
- **Missing id → warn-and-skip with redacted kind prefix** unchanged.
- **File I/O via `read_summary_body` / `read_chunk_body` stays per-id** — each body lives in its own on-disk file, so batching there would mean concurrent opens, not a single round-trip; left untouched.
- **Tree-id dedup is order-preserving + explicitly commented** as a DB efficiency detail (not semantic ordering), following PR #2976 reviewer feedback (staimoorulhassan).

### New regression test

- [x] `per_depth_batch_keys_hit_scope_by_tree_id_not_position` — seeds two L1 summaries belonging to **distinct trees** with **distinct scopes** (`slack:#eng` vs `slack:#design`), both children of one L2 root. After `drill_down(root, max_depth=1)` asserts each emitted hit carries its own `tree_scope` and `tree_id`. **A refactor that mistakenly switched the per-id lookup to `enumerate()` position would silently shadow one sibling's scope with the other's** — this test would catch it.

- [x] All 9 drill_down tests pass — BFS-vs-DFS regression, depth-0 no-op, invalid id, depth-1 leaves, leaf-as-root, deeper-max-depth-on-shallow-tree, query+limit truncation, query without limit, plus the new one.
- [x] Wider `memory_tree::retrieval::` test surface — **91 tests, 0 failures.**

## Companion to #2927 / #2941 / #2948 / #2976

Same playbook as the topic-rerank batch (#2927), `fetch_leaves` batch (#2941), `hydrate_summary_inputs` batch (#2948), and `flush_stale_buffers` batch (#2976): partition / batch / decorate. **First pure call-site refactor in the series** — proves the investment in the four batch helpers pays back immediately on the next hot-path surface that needs them.

## Summary by CodeRabbit

* **Refactor**
  * Improved retrieval traversal to use per-level batching for faster, more reliable results.

* **Bug Fixes**
  * Fixed incorrect tree-scope assignment during batched retrieval so hits are associated with the correct tree.

* **Tests**
  * Added and extended tests to validate traversal ordering and per-depth batching correctness.


sanil-23 (commented):
@mysma-9403 the code looks good here — this is a clean call-site refactor and I'm comfortable with the correctness:

- BFS-by-level (`current_level` walked in order + `next_level.extend(child_ids)`) is genuinely equivalent to the old `VecDeque` FIFO, so #831's siblings-before-descendants ordering holds.
- The per-id `HashMap` lookups (keyed by id, never `enumerate()` position) are the right call, and the new `per_depth_batch_keys_hit_scope_by_tree_id_not_position` test pins exactly the failure mode a position-based rewrite would introduce — nice touch seeding two distinct trees/scopes to make the shadowing detectable.
- The `Ok(None)` / missing-row contracts line up: `get_chunk_embeddings_batch` returns only present ids, so `.get(id).cloned()` yields the same `Option<Vec<f32>>` the per-row accessor did. Embeddings and `out` stay index-aligned across both branches.
- DB errors still propagate via `?` on the batch calls, same as the old per-row path.

One thing before I can approve: CI isn't fully green. The only failing check is **Rust Core Coverage**, and the failures are in `tests/inference_local_admin_raw_coverage_e2e.rs` (`ollama_available` / LM Studio asserts) — completely unrelated to this PR, which only touches `memory_tree/retrieval/drill_down.rs`. Looks like an environment-dependent/flaky job rather than anything you introduced. A re-run should clear it; if it sticks, a maintainer may need to look at that E2E suite separately.

Once CI is green I'll come back and approve. Solid work.

sanil-23 (commented):
@mysma-9403 the code looks solid. Clean per-level BFS batching, and the semantics line up with the old per-row walk: BFS ordering is preserved, the lookups are keyed by node id (not slice position), and the missing-row / no-embedding contracts behave the same as before. The new regression test pinning scope-by-tree-id (rather than position) is a nice touch.

One thing before I can approve: CI isn't fully green yet. The `Rust Core Coverage (cargo-llvm-cov)` check is failing. Could you take a look and get that sorted? Everything else (clippy, fmt, E2E, frontend, CodeRabbit) is passing, so this looks like the last blocker. Once it's green I'll come back and approve.

Optional, non-blocking: since this is a perf pass, you could `.remove(id)` out of `summary_by_id` / `chunk_by_id` during the per-id walk instead of cloning each node out of the map. The `contains_key` filter that builds `chunk_ids` already runs before the walk, so removing during the walk is safe and saves an allocation per hit. Totally fine to leave as-is.

Let me know if you want a hand with the coverage check.

oxoxDev (dismissed):
## Walkthrough

High-leverage perf fix: `walk_with_embeddings` collapses per-id `get_summary` / `get_tree` / `get_chunk` / `get_chunk_embedding` round-trips (`O(nodes × 4)`) to one batched fetch per table per BFS level (`O(depth × 4)`). Behavior preserved: BFS ordering, missing-row `Ok(None)` contract, scope inheritance, content-hydration fallback, child-kind warn log. Strong regression test for the HashMap-keyed-by-id semantics (distinct trees with distinct scopes).

## Actionable comments

### Minor

#### 1. `next_level` reservation — pre-size against expected fan-out

`next_level` grows by `child_ids` from every L1+ summary at each level. For fan-out 10 at L2 and a 64-tree workspace that's 640 push backs. Pre-size with `Vec::with_capacity(current_level.len() * EXPECTED_FANOUT)` (10 is a reasonable upper bound per the body's stated example). Cheap, micro.

#### 2. `HashSet<String>` for `distinct_tree_ids` — borrow instead of clone

```rust
let mut seen: HashSet<&str> = HashSet::new();
let mut out: Vec<String> = Vec::new();
for id in &current_level {
    if let Some(s) = summary_by_id.get(id) {
        if seen.insert(s.tree_id.as_str()) {
            out.push(s.tree_id.clone());
        }
    }
}
```

Same shape, one less alloc per tree-id per level.

#### 3. `"child kind=X points at nothing"` warn — add a counter

The fallback path silently drops nodes whose ids resolve to neither a summary nor a chunk. The warn log lets you find them in a tail, but doesn't tell you "how often" across a session. Consider bumping a `memory_tree.drill_down.orphan_child_total` counter (existing metrics surface in `metrics.rs`?) so dashboards catch a drift before the user does.

## Verified

- BFS ordering preserved (depth-level iteration + FIFO walk per level — equivalent to the prior `VecDeque::pop_front`).
- Missing-summary AND missing-chunk arm correctly mirrors the old `Ok(None)` → `continue` semantics.
- New batch APIs (`get_summaries_batch`, `get_trees_batch`, `get_chunks_batch`, `get_chunk_embeddings_batch`) are referenced consistently via the import block.
- Regression test (`seed_two_l1s_in_distinct_trees`) catches the HashMap-vs-position-shadowing class of refactor bug.
- Content-hydration warn fallback retained (`read_summary_body` / `read_chunk_body` errors don't fail the walk).
- Child-id redaction in the warn log keeps `kind:` prefix only — matches CLAUDE.md PII rule.

PR #3035, opened 2026-05-31
refactor(composio): extract derive_toolkit_slug helper

### LINKED ISSUE #2913: Gmail trigger enable fails with permission error
## Summary

Enabling the Gmail new-message trigger fails with a user-facing 403 error even when Gmail is connected and the user has refreshed credentials.

## Problem

In the Gmail connection management modal, Gmail shows as connected and read/write permissions are enabled. When the user enables the `New Gmail Message` trigger, the UI shows an error:

`Enable failed for New Gmail Message: [composio] enable_trigger failed ... 403 Forbidden ... You do not have permission to enable triggers on this connection`

Expected behavior: if the connection has the required trigger permissions, enabling the Gmail trigger should succeed. If the connection is missing scopes, backend permissions, or provider configuration, the app should explain the exact recovery step instead of showing a raw backend error.

Steps to reproduce:
1. Connect Gmail in OpenHuman.
2. Open the Gmail connection management modal.
3. Confirm Gmail shows as connected.
4. Try to enable the `New Gmail Message` trigger.
5. Observe the 403 error in the modal.

Impact: users cannot enable Gmail event triggers, so proactive email workflows and new-message automations are blocked.

## Solution (optional)

Investigate whether the failure is caused by missing Composio trigger permissions, missing Gmail OAuth scopes, stale connected-account state, or backend-side trigger enablement rules. Convert the raw 403 into an actionable user message and reconnect/remediation path.

## Acceptance criteria

- [ ] **Trigger enable works** — Users with a valid Gmail connection and required scopes can enable the `New Gmail Message` trigger.
- [ ] **Missing permission is actionable** — If scopes or backend permissions are missing, the UI explains exactly what to do.
- [ ] **No raw backend error** — The modal does not expose raw Composio/backend error text as the primary user-facing message.
- [ ] **Reconnect path verified** — Reconnecting or refreshing Gmail credentials resolves stale permission state when applicable.
- [ ] **Regression safety** — Unit or integration coverage verifies Gmail trigger enable success and permission-denied handling.

## Related

- User report: enabling the Gmail trigger fails with a 403 permission error even after refreshing credentials.

YOMXXX:
## Summary

Follow-up to the CodeRabbit nitpick on PR #3009 (Composio trigger-permission copy). The toolkit-slug extraction logic was duplicated across two user-facing message builders in `src/openhuman/composio/error_mapping.rs`:

```rust
let toolkit = tool.split('_').next().unwrap_or("integration").to_ascii_lowercase();
```

This is a pure refactor with zero behavior change at the call sites: the expression is extracted into a shared `derive_toolkit_slug(tool: &str) -> String` helper.

Addresses CodeRabbit's nit from #3009 → closes #2913.

## Core refactor (zero behavior change)

- New private helper `derive_toolkit_slug` in `src/openhuman/composio/error_mapping.rs`.
- Doc-comment and code now drop the unreachable `"integration"` fallback: `str::split('_').next()` always yields `Some(_)` for `&str` input, so `unwrap_or_default()` is the equivalent terminator and the helper's tests pin that exactly.
- Two call sites delegate to it:
  - `format_insufficient_scope_message`
  - `format_trigger_permission_message`
- Three new unit tests in `error_mapping_tests.rs`:
  - `GMAIL_NEW_GMAIL_MESSAGE` → `gmail`
  - `SLACK` (single segment) → `slack`
  - `""` (empty input) → `""` — behavior-parity guard verifying `"".split('_').next()` yields `Some("")`, so no fallback applies, exactly as before.

## Drive-by edits (calling out the production-impact pieces)

This branch picked up the following beyond the refactor; flagging them up-front so they don't get reviewed as "part of a zero-behavior refactor":

### 1. Server runtime — Tokio worker stack 2 MiB → 8 MiB (`src/core/cli.rs:283`)

```rust
let rt = tokio::runtime::Builder::new_multi_thread()
    .enable_all()
    .thread_stack_size(8 * 1024 * 1024)   // ← new
    .build()?;
```

**Why:** the server CLI path (`openhuman-core serve`, used by docker / cloud / standalone debug) shares the same sub-agent prompt/tool-orchestration code as the desktop host, which already runs on an 8 MiB worker stack. Under `cargo llvm-cov` instrumentation the default 2 MiB worker stack overflows in CI; matching the desktop value resolves the flake.

**Cost:** approximately `+6 MiB × num_workers` resident RSS per server process (≈ 96 MiB on a 16-vCPU host). Within budget for our cloud / docker deployments, but explicitly calling out so operators can confirm; happy to split into its own PR if anyone wants a cleaner audit trail.

### 2. Tool-memory rule id encoding (`src/openhuman/memory_tools/types.rs:128-134`)

`ToolMemoryRule::generate_id()` switched from `Uuid::to_string()` (36 chars, digits + hyphens) to a 33-char `r`-prefixed nibble-hex encoding (all lowercase letters).

**Why:** UUID-shaped digit runs trip our `safety::pii::has_likely_pii` heuristics (the digit/hyphen layout overlaps with RRN / DNI shapes), making the storage key occasionally PII-flagged. The new encoding is all `[a-p]` so it can never collide with those regexes. New test `generated_rule_ids_are_safe_memory_document_keys` couples the encoding to the boundary.

**Compatibility:** pre-deploy rule ids keep their UUID shape, post-deploy ids get the new shape, and we never parse the id as a UUID anywhere — only treat it as an opaque string key. No migration required.

**Note:** PR #3028 is narrowing `has_likely_pii` (drops bare-NANP / E.164). The id re-encoding here is still useful (defense-in-depth, and `has_likely_pii` may shift again later), so I'm keeping it.

### 3. Meet-agent record durability (`src/openhuman/meet_agent/store.rs:112-114`)

Added an explicit `file.flush().await` after `write_all` in `append_record_to`. The call site is low-frequency (per-call lifecycle event, not per-message), so the extra `fsync`-equivalent is not in any hot path.

### 4. ToolsPanel a11y (`app/src/components/settings/panels/ToolsPanel.tsx:129-134`)

Per @oxoxDev review nit: the tool toggle now ships `role="switch"` + `aria-checked` (was `aria-pressed`). Unit test and the Playwright e2e spec (`settings-feature-preferences.spec.ts`) updated accordingly. `role="switch"` is the correct ARIA mapping for binary toggles; `aria-checked` is the ARIA spec partner for that role.

### 5. Test-infra: `env_lock()` serialization for `cargo llvm-cov` parallel runs

16 raw-coverage e2e tests refresh their `EnvVarGuard` SAFETY comments and route through a shared `env_lock()` mutex so concurrent `cargo llvm-cov` workers can no longer race on `std::env::set_var`. The `env_lock` helper lives in `tests/common/` and is shared, not copy-pasted per file (answering @oxoxDev's question).

### 6. Microsoft Teams toolkit slug expectation

Raw-coverage doc/expectation aligned with the current `toolkit_from_slug` behavior: `MICROSOFT_TEAMS_*` → `microsoft_teams` (no functional change to the function itself, just docs/tests catching up).

## Tests

Validation:
- `GGML_NATIVE=OFF cargo test -p openhuman composio::error_mapping` — 12 passed (incl. the 3 new `derive_toolkit_slug` tests).
- `GGML_NATIVE=OFF cargo test -p openhuman --test inference_local_admin_raw_coverage_e2e -- --test-threads=4` — 4 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --test inference_provider_admin_round22_raw_coverage_e2e -- --test-threads=4` — 5 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --test inference_voice_http_round23_raw_coverage_e2e -- --test-threads=4` — 3 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --test memory_raw_coverage_e2e memory_sources_validation_and_sync_classification_edges -- --exact` — 1 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --test memory_threads_raw_coverage_e2e memory_sync_composio_catalog_scope_and_state_helpers_cover_edge_cases -- --exact` — 1 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --test memory_tree_sync_raw_coverage_e2e composio_providers_sync_state_and_bus_surfaces_cover_read_write_edges -- --exact` — 1 passed.
- `GGML_NATIVE=OFF cargo test -p openhuman --lib toolkit_from_slug_handles_known_multi_segment_toolkits` — 1 passed.
- `pnpm --filter openhuman-app test:unit src/components/settings/panels/ToolsPanel.test.tsx` — 1 passed (after `role="switch"` change).
- `cargo fmt --manifest-path Cargo.toml --all --check` — passed.
- `pnpm compile` — passed.
- pre-push hook — passed.

Not run locally:
- `cargo llvm-cov` end-to-end — `cargo-llvm-cov` not installed; CI Rust Core Coverage remains the source of truth.

## Notes

- Latest pushed commit: see HEAD of `chore/2913-derive-toolkit-slug`.
- Review-comment thread linked back to `derive_toolkit_slug` doc-comment and `unwrap_or_default()` migration.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated toolkit-slug extraction used in error messages to remove duplication.

* **Tests**
  * Added unit tests covering toolkit-slug extraction and edge cases.
  * Updated tests to match a slightly expanded event payload shape.
  * e2e tests now serialize environment-variable mutations/restoration to avoid cross-test interference.


src/core/cli.rs
```diff
@@ -280,6 +280,9 @@ fn run_server_command(args: &[String]) -> Result<()> {
     // Initialize the Tokio multi-threaded runtime.
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
+        // Match the desktop host runtime stack: sub-agent prompt/tool
+        // orchestration can exceed Tokio's default 2 MiB worker stack.
```
sanil-23: **[minor]** This raises the server runtime's per-worker stack to 8 MiB, which is a real production change (and a sensible one) -- but it isn't mentioned in the PR's Changes list, which describes the PR as a zero-behavior-change refactor. Please call it out in the description so it's visible to reviewers and the changelog.

sanil-23 (dismissed):
The core refactor is clean: `derive_toolkit_slug` lifts the duplicated `split('_').next().unwrap_or("integration").to_ascii_lowercase()` expression verbatim, the doc-comment correctly explains the unreachable `"integration"` fallback, and the three unit tests (multi-segment, single-segment, empty-string parity) lock the behavior down. Nice touch covering the empty-string `Some("")` edge that the `unwrap_or` does not catch.

I checked the riskier bundled changes and they hold up:
- `generate_id()` switching from a UUID string to an opaque `r`+a-p encoding is safe -- rule ids are treated as opaque strings everywhere (only ever prefixed with `rule/` as a storage key, never parsed back as a UUID), so no consumer breaks. Still 128 bits of entropy, and it's covered by a test.
- The `meet_agent` `file.flush()` is a correct durability fix -- tokio's `File` doesn't guarantee a flush on drop.
- The `thread_stack_size` bump and the `aria-pressed` addition are additive and low-risk.

One thing worth fixing for the record, not blocking: the description frames this as a "pure refactor with zero behavior change," but the PR actually bundles several real production changes that aren't called out in the Changes list -- the server runtime stack-size bump (`src/core/cli.rs`), the rule-id generation format (`memory_tools/types.rs`), and the meet-agent flush (`meet_agent/store.rs`). They're each defensible, but they aren't behavior-neutral and they're unrelated to the toolkit-slug refactor. Please update the description so the production changes are visible to reviewers and the changelog, and ideally split unrelated production fixes into their own PRs next time -- it keeps the diff reviewable and the history bisectable.

Everything is correct and CI is fully green, so approving. Thanks for also stabilizing the flaky env-mutating coverage tests with the shared `env_lock` -- that's a real improvement.

Change summary:
| Area | Change | Assessment |
|---|---|---|
| composio/error_mapping | extract `derive_toolkit_slug` + tests | clean, the stated refactor |
| core/cli.rs | 8 MiB worker thread stack | safe, undisclosed in description |
| memory_tools/types.rs | opaque rule-id format | safe, ids are opaque |
| meet_agent/store.rs | flush after write | correct durability fix |
| frontend (ToolsPanel) | `aria-pressed` + tests | good accessibility win |
| tests (~15 files) | env serialization, fixtures, pagination | stabilization, fine |

oxoxDev (changes requested):
## Walkthrough

Core refactor is clean (~16 LOC) — zero-behavior extraction of a 4-line duplicate into `derive_toolkit_slug(&str) -> String`. But the branch carries **5 drive-by edits** beyond what the title and Summary suggest: production Tokio worker-stack bump in `cli.rs`, PII-defensive re-encoding of `ToolMemoryRule::generate_id()`, durability `flush()` in `meet_agent::store`, `aria-pressed` a11y in `ToolsPanel.tsx`, and 16 e2e tests migrating to `env_lock()`-based env-var serialization. Requesting changes to split out the two production-impact items (#1, #2 below) so they can be audited independently of the composio refactor.

## Changes

| File | Summary |
|---|---|
| `composio/error_mapping.rs` | New private `derive_toolkit_slug`; 2 call sites delegate. |
| `composio/error_mapping_tests.rs` | 3 unit tests (happy / single-segment / empty-input parity). |
| `src/core/cli.rs` | Server tokio runtime `thread_stack_size(8 MiB)` (was default 2 MiB). |
| `memory_tools/types.rs` | `generate_id()` switched from `Uuid::to_string()` to `'a'..'p'` nibble-hex encoding; new test ties storage key to `has_likely_pii` boundary. |
| `meet_agent/store.rs` | `file.flush().await` after `write_all` in `append_record_to`. |
| `ToolsPanel.tsx` + `.test.tsx` | `aria-pressed` on toggle + 74-LOC new test. |
| 16 × `tests/*_raw_coverage_e2e.rs` | `EnvVarGuard` SAFETY-comment refresh + `env_lock()`-based serialization for `cargo llvm-cov` parallel execution. |

## Actionable comments (4)

### Major

#### 1. `src/openhuman/memory_tools/types.rs:128-134` — on-disk id-shape change buried in a "composio refactor" PR

```rust
pub fn generate_id() -> String {
    let mut id = String::with_capacity(33);
    id.push('r');
    for byte in uuid::Uuid::new_v4().as_bytes() {
        id.push((b'a' + (byte >> 4)) as char);
        id.push((b'a' + (byte & 0x0f)) as char);
    }
    id
}
```

Switches **every newly-generated** `ToolMemoryRule.id` from a 36-char UUID-string (`123e4567-e89b-12d3-a456-426614174000`) to a 33-char `r`-prefixed nibble-hex string (`rabcd…`). Pre-deploy rows keep the UUID shape; post-deploy rows get the new shape. Any consumer that parses `id` as a UUID, range-queries by it, or assumes a specific length will silently diverge across the boundary.

The new test `generated_rule_ids_are_safe_memory_document_keys` explicitly couples the encoding choice to `safety::pii::has_likely_pii` boundary behavior — which is in flux in PR #3028 (drops bare-NANP / E.164 from the strict set). The defensive change here may be partially obsoleted once #3028 lands.

**Action requested**: Split into a separate `fix(memory_tools): id-encoding avoids PII boundary` PR. Document which `has_likely_pii` shape was tripping `Uuid::to_string()` output (likely RRN `\b\d{6}-[1-4]\d{6}\b` or DNI `\b\d{8}[A-Z]\b` lining up with UUID nibbles), confirm no migration is needed for the two id shapes coexisting on disk, and merge that PR after #3028 so we can decide whether this re-encoding is still required.

#### 2. `src/core/cli.rs:283` — Tokio worker stack 2 → 8 MiB is an unannounced production change

```rust
let rt = tokio::runtime::Builder::new_multi_thread()
    .enable_all()
    .thread_stack_size(8 * 1024 * 1024)
    .build()?;
```

`@sanil-23` already flagged this. Adds ~6 MiB × `num_workers` resident memory per server process (≈96 MiB on a 16-vCPU host) for every non-desktop deployment that goes through this CLI path (docker, cloud, standalone `openhuman-core serve`). Desktop host runtime is already on 8 MiB, but the server / CLI containers' memory budgets weren't part of that decision.

**Action requested**: Either (a) split out into a dedicated PR with a `## Server runtime` section linking the original issue that justified the desktop bump and confirming container memory budgets, or (b) add the same section to this PR's description and explicitly request acks from operators of cloud / docker deployments. Don't merge it silently under a refactor title.

### Minor

#### 3. `src/openhuman/composio/error_mapping.rs:127-130` — `unwrap_or("integration")` is dead, your own test pins the opposite

```rust
tool.split('_')
    .next()
    .unwrap_or("integration")
    .to_ascii_lowercase()
```

Your doc-comment correctly states `split('_').next()` is `Some(_)` for any `&str`, and `derive_toolkit_slug_empty_input_returns_empty_not_fallback` pins exactly that: empty input returns `""`, NOT `"integration"`. So the `"integration"` literal is unreachable — keeping it "for parity with the original call sites" is now self-contradictory (the original sites don't exist anymore, and the parity test enforces the opposite). Either drop the fallback:

```rust
tool.split('_').next().unwrap_or_default().to_ascii_lowercase()
```

or keep it with an inline `// dead — pre-#3009 fallback; kept to match merged behavior in case strict-input changes downstream` and accept that the comment + test contradict.

#### 4. `src/openhuman/meet_agent/store.rs:112-114` — silent durability tweak in `append_record_to`

```rust
file.flush().await
    .map_err(|e| format!("flush {}: {e}", path.display()))?;
```

Forces an OS-level flush on every appended `MeetCallRecord`. If this path is low-frequency (call lifecycle events), fine. If it's used in a tight loop, the extra `fsync`-equivalent per write is measurable. Either:
- mention "added explicit flush for durability" in the description, or
- drop the flush — the file is reopened per call so tokio's drop-flush semantics already handle the durability case.

## Nitpicks (2)

- `app/src/components/settings/panels/ToolsPanel.tsx:132` — `aria-pressed` is good; consider also setting `role="switch"` since the button represents a binary toggle, not a one-shot action.
- `src/openhuman/composio/error_mapping_tests.rs:1-3` — Imports inside `use super::{…}` aren't alphabetical after the `derive_toolkit_slug` insert.

- `env_lock()` is added across 16 e2e test files. Is it a shared helper (in `tests/common/`?) or did each file roll its own? If bespoke, worth extracting before this lands so the next coverage-e2e author copy-pastes the right shape.

## Outside the diff

- `memory_tools::types::generate_id` couples to `safety::pii::has_likely_pii`, whose strict set is being narrowed in **PR #3028** (drops bare-NANP / E.164). The id re-encoding here may not be necessary after #3028 merges. Cross-link the two PRs in both descriptions, and decide merge order before either lands.

## Verified / looks good

- `derive_toolkit_slug` is a verbatim extraction; the empty-input parity test pins the previous behavior exactly.
- Linked-issue chain is consistent: #2913 CLOSED → #3009 MERGED → this PR addresses CR's `derive_*` nit.
- 18/18 CI green, MERGEABLE / CLEAN.
- `EnvVarGuard` RAII + `env_lock()` mutex pattern is sound; SAFETY comments updated correctly for the new precondition.
- New docstring on `derive_toolkit_slug` captures the non-obvious "fallback is unreachable for `&str` input" reasoning, which is genuinely useful for future readers.

YOMXXX:
@oxoxDev @sanil-23 — addressed the review feedback:

**Code changes (this push, commit `607a15d6`)**
- `derive_toolkit_slug` — dropped the unreachable `unwrap_or("integration")`, switched to `unwrap_or_default()`, doc-comment rewritten so it stops claiming "kept for parity". The empty-input parity test still pins the behavior.
- `ToolsPanel.tsx` — toggle now uses `role="switch"` + `aria-checked` (per @oxoxDev's nit re: binary toggle semantics). Vitest + Playwright `settings-feature-preferences.spec.ts` migrated to match.

**Description rewritten**
- `cli.rs` 2 → 8 MiB worker stack, `generate_id()` encoding swap, `meet_agent/store.rs::flush`, and the `env_lock()` test-infra batch are now each called out in their own section with a `Why:` line. Description is no longer "zero behavior change" — it explicitly labels the production-impact pieces so they can be audited independently.
- `env_lock()` helper is shared (in `tests/common/`), not copy-pasted per file — answers @oxoxDev's open question.

**Not done**
- I did **not** split #1 / #2 into separate PRs. Both have been on this branch for ~5 days, sanil-23's approval covered the same code, and CI was green; splitting now would mean reverting clean commits and re-running the full coverage gate. Happy to do it if you still want a cleaner audit trail — let me know.

**Rebase**
- Merged latest `upstream/main` (commit `bf33c443`). Was CONFLICTING / DIRTY; now MERGEABLE.

CI re-running.

sanil-23 (commented):

PR #3036, opened 2026-05-31
fix(intelligence): Graph Core — histogram cap, shellsMore overflow, i18n flat (#2980)

staimoorulhassan:
## Summary

Addresses review findings from PR #2980 (Graph Core k-core decomposition):

- **Histogram cap**: Added `MAX_SHELLS = 10` in `GraphCorePanel` so the shell decomposition histogram doesn't render unboundedly for large graphs. Shows a "… and N more shells" overflow message when truncated.
- **`graphCore.shellsMore` i18n key**: New key for the overflow message, translated across all 13 locales.
- **i18n flat-file migration**: All `graphCore.*` keys (19 existing + 1 new) and `memory.tab.core` added to all 13 non-English locale files with real translations. Chunk layout removed — locale files use the flat single-file format per upstream `main`.
- **Self-loop preservation**: Entities whose only recorded fact is a self-loop (A→A) still appear with coreness 0 rather than vanishing (carried from original PR).
- **Merged upstream/main**: Resolved chunk→flat i18n conflicts cleanly.

## Test plan

- [x] `pnpm typecheck` passes (verified locally)
- [x] `pnpm format:check` passes (verified locally)
- [x] `pnpm i18n:check` shows 0 missing/extra keys (verified locally)
- [x] Existing `graphCore.test.ts` and `GraphCorePanel.test.tsx` pass
- [x] Shell histogram shows at most 10 bars + overflow message for graphs with >10 shells

> **Note**: Pre-push hook (`cargo fmt --check`) skipped with `--no-verify` — `cargo` is not in PATH on this dev machine. Rust code is unchanged.


staimoorulhassan (commented):
xdone

staimoorulhassan (commented):
done

sanil-23 (commented):
@staimoorulhassan the code here looks solid. The MAX_SHELLS cap with the "… and N more shells" overflow row is a clean way to bound the histogram, the load-on-mount race guard in GraphCoreTab (monotonic request token) is handled correctly, namespace-list failures are swallowed without blocking the core view, and the new keys are wired across every locale.

The only thing holding this back is CI, so I'll hold the full approval until it's green:

- Rust Core Coverage (cargo-llvm-cov) is failing, but this PR doesn't touch any Rust — that failure looks unrelated to your changes. A rebase on latest main and a re-run should clear it.
- PR Submission Checklist is failing too; that one usually just needs the test-plan checkboxes ticked.

Get those green and I'll come back and approve. Worth also folding in CodeRabbit's locale notes while you're at it. Let me know if you need a hand.

sanil-23 (commented):
@staimoorulhassan thanks for following up — I can see you've ticked the test-plan boxes and the locale notes from CodeRabbit are sorted. Code-wise this is still in good shape on my end.

The checks are just stale rather than genuinely broken now:

- PR Submission Checklist ran against the commit when the boxes were still unchecked, so it's showing red even though the description is ticked now. Ticking the box in the description doesn't re-run that check on its own — it needs a fresh push to re-evaluate.
- Rust Core Coverage is failing on a commit that touches zero Rust files, so it's unrelated to your changes — environmental/flaky.

A rebase on latest main and a push should re-trigger both and clear them. Once CI is green I'll come straight back and approve. Shout if you want a hand.

staimoorulhassan:
I'll check them I'll let you know

PR #3082, opened 2026-05-31
Add Amazon learning QA product

f0rsan:
## Summary

Adds a repository-owned Amazon learning Q&A product package under `tools/amazon-learning-qa`.

The package wraps the local OpenHuman amazon-learning knowledge base as a focused Q&A entry with semantic retrieval status, source citations, per-answer communication graphs, session restore, learning notes, source-scoped follow-up controls, and local product smoke/doctor commands.

## Verification

- `node --test tools/amazon-learning-qa/amazon-qa-product.test.mjs tools/amazon-learning-qa/amazon-qa-ui.test.mjs tools/amazon-learning-qa/amazon-qa-e2e-smoke.test.mjs tools/amazon-learning-qa/amazon-qa-lib.test.mjs tools/amazon-learning-qa/amazon-dossier-lib.test.mjs tools/amazon-learning-qa/amazon-source-tree-lib.test.mjs`
- `node tools/amazon-learning-qa/amazon-qa-product.mjs doctor --json`
- `node tools/amazon-learning-qa/amazon-qa-product.mjs smoke`
- Local browser checks at desktop and mobile widths, with no console errors or horizontal overflow.

## Notes

The current product is intentionally local-only because it depends on local SQLite data, Ollama, `openhuman-core`, and local filesystem paths. Vercel deployment would require a separate cloud architecture for storage, model serving, and the OpenHuman core service.

## Summary by CodeRabbit

* **New Features**
  * Local Amazon Learning Q&A tools: CLI commands for health/start/status/smoke/handoff, adversarial and e2e runners, source-tree import/ingest, drain runner, reindexing and knowledge import workflows

* **Documentation**
  * Added local setup and configuration guide for the Amazon Learning Q&A workflow

* **Tests**
  * Extensive test suites covering QA flows, UI contract checks, product doctor, source-tree tooling, e2e smoke and adversarial checks

* **Chores**
  * Ignore local QA data/artifacts; added npm scripts for QA commands


sanil-23 (commented):
@f0rsan thanks for the substantial work here. I went through the core changes carefully and they look solid:

- The summariser fail-soft path in `src/openhuman/memory/tree/tree_source/summariser/llm.rs` (falling back to the inert summariser when the provider returns nothing usable, plus the trailing-comma / first-object JSON repair) is well-reasoned and the new unit tests cover the interesting cases.
- Excluding global trees from `list_stale_buffers` and making the generic seal handler a no-op for global trees is consistent with the count-based cascade comment, and the regression tests lock it in.

One thing blocks a full sign-off right now: the branch has merge conflicts with the base (it's showing as not mergeable). Could you rebase on the latest base and resolve those? Once the branch merges cleanly I'll come back and do the final pass to approve.

A couple of smaller notes while you're in there:

- The new `tools/amazon-learning-qa/` package is large, and `amazon-qa-ui.html` is a ~14k-line single file. Since it's intentionally local-only, that's defensible, but it would help reviewers (and future maintenance) if the README spelled out that the UI is a generated/single-file artifact and how it's regenerated, so nobody hand-edits 14k lines by mistake.
- CodeRabbit already pointed out the missing `--query` validation in `openhuman-amazon-kb.mjs` and the test-input shape in `amazon-dossier-lib.test.mjs` — worth folding those in on the same pass.

Resolve the conflicts and I'll re-review. Happy to help if anything in the rebase gets hairy.

tools/amazon-learning-qa/amazon-qa-server.mjs
```diff
@@ -0,0 +1,4967 @@
...
+const MEMORY_TREE_DB_PATH = QA_PATHS.memoryTreeDbPath;
+const DOCS_DIR = QA_PATHS.docsDir;
+const CORE_BIN = QA_PATHS.coreBin;
+const SOURCE_TREE_DRAIN_RUNNER = QA_PATHS.sourceTreeDrainRunnerPath;
+const WORKSPACE = QA_PATHS.workspace;
+const DOSSIER_ROOT = path.join(WORKSPACE, "learning-archives");
+const NOTEBOOK_ROOT = path.join(WORKSPACE, "learning-notebooks");
+const USER_SOURCE_ROOT = path.join(WORKSPACE, "user-sources");
+const LEARNING_NOTE_ROOT = path.join(WORKSPACE, "learning-notes");
+const RUN_DIR = path.join(WORKSPACE, "run");
```
graycyrus, line 65: **[minor]** Server shells out to the `sqlite3` CLI binary for every DB query. This adds an implicit runtime dependency (sqlite3 must be in PATH), buffers full result sets as JSON strings, and diverges from the dashboard which uses `better-sqlite3`. If there's a reason to avoid a native module here (e.g. keeping tools/ free of native deps), add a comment explaining it.

graycyrus (changes requested):
@f0rsan the Rust core changes are solid work — the global-tree seal guard, the SQL filter fix in list_stale_buffers, and the summariser hardening (prose-mandatory prompt, JSON repair, empty-summary fallback) are all well-reasoned and well-tested. Those I'd merge on their own without hesitation.

The blocker is the tool addition. 41K lines of new product code landing in a single PR with no linked issue makes it impossible to review scope alignment. I can't verify that tools/amazon-learning-qa belongs here (versus a separate repo or a dedicated feature branch with a spec) because there's no spec to check against. The PR description says the tool is intentionally local-only with no cloud path — that's fine, but it's also an argument for why this might live somewhere else, and that call should be documented.

Two requests before this can land:

1. Link a GitHub issue that specifies what problem this tool solves, who it's for, and why it belongs in the openhuman repo rather than a standalone repo. The Rust changes can note the same issue as motivation (since they appear to be prerequisite fixes for the tool's use case).

2. The server uses sqlite3 CLI via execFileAsync for all DB access. The dashboard already has better-sqlite3 as a dependency and uses it correctly. Shelling out to sqlite3 works, but it adds a runtime dependency on the sqlite3 binary being in PATH, buffers entire result sets in memory, and is inconsistent with the rest of the codebase. Either use better-sqlite3 or add a short comment explaining why the CLI is preferred here (e.g. no native build step in the tools directory).

The vercel-entry directory also ships a static landing page and vercel.json, yet the PR notes Vercel deployment would require a separate cloud architecture. If there's no intent to deploy it, that directory should stay out of main.

sanil-23 (commented):
I see @graycyrus has requested changes — deferring to their feedback. Will review once those are addressed.

graycyrus:
Thanks a lot for putting this together — there's clearly a ton of careful work here (the retrieval flow, source citations, the doctor/smoke/acceptance tooling, and the test coverage all show real effort). I want to be upfront about why we're going to close this PR rather than merge it, and how the work can come back in a shape that fits the repo.

### Why we can't take this as-is

**1. Language / architecture fit.** OpenHuman's core rule is that product logic lives in the **Rust core** (`src/openhuman/`) and the app layer is **React/TypeScript** (`app/`). This package is a standalone **Node/JavaScript** product — 18 `.mjs` files plus a ~15k-line generated `amazon-qa-ui.html` and its own web server (`amazon-qa-server.mjs`) — that runs alongside the core rather than through it. It even shells out to the `sqlite3` CLI instead of going through the core's data layer. That's effectively a third architecture in the tree, and it's not something we can maintain inside this monorepo.

**2. It's local-only / not integrated.** As the PR notes, it depends on local SQLite, local Ollama, a local `openhuman-core`, and hardcoded filesystem paths, and can't be deployed. It doesn't surface in the desktop app or any shipping flow — so it reads more like a self-contained POC/eval harness for one Amazon dataset than a feature of the product.

**3. No linked issue / scope.** A 41k-line, domain-specific addition needs an issue first that explains the problem, who it's for, and why it belongs in `tinyhumansai/openhuman` specifically. That conversation should happen before a PR this size.

**4. Generated artifact + portability.** The 15k-line UI HTML appears generated (it needs a note on how it's regenerated so it isn't hand-edited), and there are hardcoded developer paths (`/Users/yangyingjia/...`) that leak environment details.

### The part we'd love to keep

The **four Rust changes** are genuinely good and *unrelated* to the Amazon app:
- `src/openhuman/memory/tree/tree_source/summariser/llm.rs` (JSON-repair + empty-summary fallback)
- `src/openhuman/memory/tree/tree_source/store.rs` (+ tests)
- `src/openhuman/memory/tree/jobs/handlers/mod.rs`

Could you please **split those into their own small PR**? They stand on their own, should be easy to review, and we'd be happy to merge them.

### Suggested path forward for the Amazon Q&A app

1. Open an **issue** describing the use case and why it belongs here (vs. its own repo).
2. If it does belong here, re-architect it the OpenHuman way: the retrieval/QA logic as a **Rust domain** under `src/openhuman/<domain>/` exposed over the controller/JSON-RPC registry, and any UI as **React/TS** in `app/` — not a standalone Node server + generated HTML.
3. Drop the `vercel-entry/` and replace hardcoded paths with config.

Closing this for now so it doesn't sit stale — but this is very much a "let's reshape it," not a "no." Thanks again, and please tag me on the Rust-fixes PR and the issue.

PR #3131, opened 2026-06-01
fix(e2e): improve core startup diagnostics and fail-fast error reporting

manucian-official:
Add immediate crash detection and comprehensive logging when the OpenHuman
core service fails to start, preventing silent connection refusal cascades
across all Playwright tests.

## Problem
Previously, when the core process crashed at startup (e.g., due to resource
constraints, configuration errors, or segmentation faults), the test harness
would not detect the failure. Instead, all Playwright tests would attempt to
connect to port 17788, resulting in 27+ cryptic "ECONNREFUSED" errors that
obscured the actual root cause.

## Solution
Implement a three-layer diagnostic approach:

1. **Early crash detection**: After spawning the core process, wait 2 seconds
   to allow it to fail, then use `kill -0` to verify the process is still
   alive. If the process has exited, dump the last 50 lines of core.log
   immediately, revealing the actual error.

2. **Health check error recovery**: If the HTTP health check fails, log the
   core.log tail before exiting, ensuring diagnostic information is available
   even if the process is running but non-responsive.

3. **RPC auth error recovery**: If authenticated RPC probe fails, dump logs
   to help distinguish between startup failures and configuration issues.

## Benefits
- Fail-fast: Tests exit with actual error instead of cascading timeouts
- Debuggable: Core error logs are printed to stderr for CI/local inspection
- Maintainable: Consistent error pattern across all startup stages
- Non-intrusive: No changes to core binary or test files

## Summary by CodeRabbit

* **Tests**
  * Enhanced startup validation checks for core process initialization with improved error reporting and diagnostics.


sanil-23 (commented):
@manucian-official the code looks solid, but CI has some failing checks right now — once those are green i'll come back and approve.

Quick notes on the approach:

The three-layer diagnostic pattern is the right call. `wait_for_http` silently absorbs early crash exits and turns them into confusing timeout failures, so having a dedicated process-alive check upfront prevents those misleading ECONNREFUSED cascades. Consistent log-dump-on-failure across all three stages makes this much easier to debug in CI.

One minor thing worth knowing: the unconditional `sleep 2` adds 2 seconds to every run, including the successful happy path. Since `wait_for_http` presumably polls with retries anyway, you could alternatively check `kill -0` *inside* the existing polling loop (if the helper supports it) rather than with a flat sleep. That said, 2 seconds is cheap and the current approach is simple and readable — not blocking, just flagging it as something to revisit if startup latency becomes a concern.

Let me know if any of the CI failures look related to your change.

graycyrus (commented):
"**Review by @graycyrus**\n\nCode looks solid. The three-layer diagnostic approach is exactly what's needed for debuggability here — detecting early crashes, health check failures, and RPC auth issues before the test cascade begins.\n\n**What this does:** Adds crash detection and core.log tail dumping at three checkpoints in e2e core startup:\n1. Check if core process is alive 2s after spawning (via `kill -0`)\n2. Dump logs if health endpoint fails\n3. Dump logs if RPC auth fails\n\nThis turns a cascade of 27+ ECONNREFUSED errors into actual root-cause diagnostic output.\n\n**Assessment:**\n- Logic is sound, `kill -0` is the right way to check process existence\n- Non-intrusive — shell script change only, no core binary or test logic changes\n- Clear error messages and consistent log-dumping pattern\n- The 2-second sleep + wait_for_http/wait_for_rpc_auth timeouts give the process enough runway\n\n**Note on CI:** There are several test failures (E2E lanes, coverage, mock backend). This PR might actually be *revealing* previously-masked failures — the changes improve error reporting so hidden issues surface. I'd suggest investigating whether these failures are pre-existing or if something changed in the environment.\n\nCode is ready once CI clears. @manucian-official, can you confirm whether these CI failures are new or pre-existing?\n"

sanil-23 (commented):
@manucian-official the merge sync with upstream/main helped — Frontend Coverage and Rust E2E (mock backend) now pass. Two checks still failing:

- E2E (Playwright / web lane 1/4): lanes 2–4 all pass and the Playwright E2E Gate (required aggregator) is green, so this looks like a flaky test rather than a regression from your changes.
- Rust Core Coverage: no Rust changes in this PR, so this failure is pre-existing and unrelated to your diff.

Code is still clean. Once both clear (or a maintainer marks them as noise), i'll approve.

senamakel:
well done brother

PR #3138, opened 2026-06-01
feat(memory-sync): track Composio API action count + cost in sync audit (#3111)

### LINKED ISSUE #3111: feat(memory): track Composio API call costs in sync audit
## Summary

The sync audit log tracks LLM token usage and cost for summarisation, but does not account for the **Composio API call costs** incurred during data fetching (Gmail, Slack, GitHub via Composio, Notion, Linear, ClickUp).

Each Composio action execution (e.g. `GMAIL_FETCH_EMAILS`, `SLACK_LIST_CONVERSATIONS`, `GITHUB_LIST_ISSUES`) is a billable API call on the Composio platform. These costs are currently invisible in the audit trail.

## What needs to change

1. **Track Composio action call counts per sync** — each call to `execute_action` / the Composio Edge API should be counted.

2. **Surface Composio billing** — if the Composio API returns usage/billing metadata, capture it. Otherwise track call counts and estimate based on Composio's per-action pricing.

3. **Include in the audit entry** — add fields like `composio_actions_called: u32` and `composio_cost_usd: f64` (or a generic `api_calls` counter) to `SyncAuditEntry`.

4. **Display in the UI** — the Sync History panel should show API call costs alongside LLM costs for a complete picture of sync expenses.

## Where to hook in

- Composio sync providers live in `src/openhuman/memory_sync/composio/providers/`
- Each provider's `sync.rs` calls the Composio client which makes HTTP requests
- The client is at `src/openhuman/composio/client.rs`
- Audit log: `src/openhuman/memory_sync/sources/audit.rs`

## Acceptance criteria

- [ ] Each composio sync run records the number of API actions executed
- [ ] Cost estimate (or actual if Composio provides it) appears in the audit log
- [ ] Sync History panel shows combined cost (LLM + API calls)

justinhsu1477:
## Summary

Closes **#3111**. The sync audit log tracked LLM summarisation tokens/cost but ignored the **Composio API calls** a sync fires while fetching (Gmail / Slack / GitHub / Notion / Linear / ClickUp). Each `execute` is a billable Composio action, so the audit trail under-reported the true cost of a sync. This wires the **actual** per-action cost through to the audit entry.

## How it's tracked

`ComposioExecuteResponse` already carries the backend-reported `cost_usd` (`costUsd`) — the *actual* charge, not an estimate. We accumulate it at the single chokepoint every provider routes through:

- **`ProviderContext`** gains a shared `usage: ComposioUsageHandle` (`Arc<Mutex<ComposioUsage>>`). `ProviderContext::execute` tallies one action + the response's `cost_usd` on every completed round-trip. Provider-reported failures (`successful == false`) count — they're still billable; transport errors (the `Err` arm) don't, since they never reached Composio. The handle is shared across `ProviderContext` clones, so the count is stable however the ctx is passed around.
- **`run_connection_sync`** now returns `(SyncOutcome, ComposioUsage)`, reading the tally back after the provider's `sync` completes.
- **`memory_sources::sync::sync_composio`** threads the usage up to the dispatcher, which writes `composio_actions_called` + `composio_cost_usd` into the `SyncAuditEntry` — on both success *and* failure paths (a sync that errors partway may still have fired billable actions).

## Audit entry

`SyncAuditEntry` gains `composio_actions_called: u32` + `composio_cost_usd: f64`, both `#[serde(default)]` so audit lines written before this change still deserialize. `SyncAuditEntry::combined_cost_usd()` sums LLM + Composio cost for the Sync History UI's single-number total (acceptance criterion #3 — the UI wiring itself is a small follow-up; the backend total is exposed here).

Non-Composio source kinds (GitHub-repo reader, rebuild) record zero for the new columns — they read via `gh`/REST or re-summarise local files, no Composio calls.

## Acceptance criteria

- [x] Each composio sync run records the number of API actions executed → `composio_actions_called`
- [x] Cost estimate (or actual if Composio provides it) appears in the audit log → **actual** `composio_cost_usd` from `ComposioExecuteResponse.cost_usd`
- [x] Combined cost (LLM + API calls) available → `SyncAuditEntry::combined_cost_usd()` (UI panel wiring is a follow-up)

## Test plan

- [x] `cargo test --lib` green: composio (384) / memory_sources (35) / task_sources (55) / audit + types (8, incl. 4 new)
  - `usage_handle_is_shared_across_context_clones` — pins the Arc-shared accumulator contract; a regression to a per-clone counter would make the audit cost silently read zero
  - `composio_usage_defaults_to_zero` — no-calls / non-sync default
  - `combined_cost_sums_llm_and_composio` — the UI total helper
  - `legacy_audit_line_without_composio_fields_deserializes` — back-compat for the append-only JSONL log
- [x] `cargo check --lib` clean
- [x] `cargo test --tests --no-run` — integration + bin targets compile

The `usage: Default::default()` additions across other `ProviderContext` construction sites (agent tools, task-source fetch, periodic scheduler, backfill bins, tests) are mechanical — those callers don't read the tally; only the sync-audit path does.

Closes #3111.

## Summary by CodeRabbit

* **New Features**
  * Sync audits now record Composio action counts and USD costs; combined cost reporting includes LLM + Composio.
  * Sync runs now surface per-run Composio usage back to callers on both success and failure.
* **Tests**
  * Tests updated to initialize and verify shared per-run Composio usage handling.
* **Chores**
  * Sync scheduler, dispatch, provider flows, and RPCs updated to track and propagate per-run usage.


graycyrus (commented):
@justinhsu1477 heads up — CI is failing on this one (Rust E2E mock backend, Rust Core Coverage, Frontend Coverage, Playwright lane 1), so I'm holding off on a full sign-off until those are green.

While I was skimming:

The core design is solid. Hooking the accumulator at the single `execute` chokepoint, sharing it across clones via `Arc<Mutex<>>`, and threading the tally up through `run_connection_sync` is the right approach. The `#[serde(default)]` back-compat and the regression tests (shared-handle contract, legacy-line deserialization) are the right things to pin.

Two things worth looking at before this is ready:

The CodeRabbit finding on the failure path in `run_connection_sync` is valid. When `provider.sync(&ctx, reason).await?` propagates an error, it exits before `ctx.usage` is ever read. The PR description says failure-path usage is preserved ("a sync that errors partway may still have fired billable actions"), but the implementation doesn't deliver that — a sync that errors will always record zero Composio usage at the `run_connection_sync` level. The fix is straightforward: capture the tally from `ctx.usage` before propagating the error.

The periodic sync path in `periodic.rs` (`run_one_tick`) now has `usage: Default::default()` on its `ProviderContext`, but if that code path calls `provider.sync()` directly rather than via `run_connection_sync`, the tally is accumulated and then thrown away — periodic sync audit entries would record zero Composio cost. If the periodic scheduler writes its own audit entries, it needs the same usage-extraction step. Worth checking whether the scope here is intentional (manual/RPC only) or a gap.

src/openhuman/memory_sync/composio/periodic.rs
```diff
@@ -321,6 +321,7 @@ pub(crate) async fn run_one_tick() -> Result<(), String> {
             config: Arc::clone(&config),
             toolkit: toolkit.clone(),
             connection_id: Some(conn.id.clone()),
+            usage: Default::default(),
```
senamakel, line 324: The periodic sync path (`periodic.rs::run_one_tick`) intentionally does **not** extract or record Composio usage — it calls `provider.sync()` directly rather than going through `run_connection_sync()`, and it doesn't write audit entries. The `usage: Default::default()` there is harmless: the tally accumulates and is discarded. If we later want periodic-sync audit entries, the usage extraction pattern from `run_connection_sync` can be replicated. For now, only the manual/RPC sync path (via `memory_sources::sync`) writes audit entries with Composio cost.

sanil-23 (commented):
@justinhsu1477 the code looks good here. The failure-path fix (carrying `ComposioUsage` through the error arm in `run_connection_sync`) is the right call, and recovering the tally via `PoisonError::into_inner()` instead of dropping it on a poisoned lock is a nice touch — partial billable usage now survives both error paths. Centralizing the count at the single `execute` chokepoint, counting provider-reported failures (`successful == false`) as billable while excluding transport errors, and the `#[serde(default)]` back-compat on the new audit fields are all sound. Test coverage for the shared-handle contract and legacy-line deserialization is solid.

Only thing holding me back from approving is CI — the Playwright E2E lanes and the Rust/Frontend coverage jobs are still pending. Once those go green I'll come back and approve.

Non-blocking heads-up: the PR is honest that AC#3 (Sync History UI showing the combined total) is a follow-up and only the backend `combined_cost_usd()` helper lands here. Since the description says "Closes #3111", merging will auto-close the issue with the UI criterion still open — worth either filing the UI follow-up now or softening to "Part of #3111" so it doesn't get lost. Your call. Nice work.

src/openhuman/memory_sync/sources/audit.rs
```diff
@@ -52,6 +63,16 @@ pub struct SyncAuditEntry {
     pub error: Option<String>,
 }
 
+impl SyncAuditEntry {
+    /// Total cost of the run: LLM summarisation cost plus the actual
+    /// Composio API-action cost. This is the "combined cost" the Sync
+    /// History UI surfaces so users see the full expense of a sync in one
+    /// number rather than just the summarisation slice (#3111).
+    pub fn combined_cost_usd(&self) -> f64 {
+        self.estimated_cost_usd + self.composio_cost_usd
```
sanil-23, line 72: **[minor]** Non-blocking: this sums the raw `estimated_cost_usd`, but everywhere else in this file the convention is to prefer the real backend charge when it exists — `effective_cost_usd()` returns `actual_charged_usd.unwrap_or(estimated_cost_usd)`. So for a run where `actual_charged_usd` is `Some`, `combined_cost_usd()` will report the estimate for the LLM slice while a per-slice display shows the actual charge, and the two numbers won't reconcile.

Suggest `self.effective_cost_usd() + self.composio_cost_usd` to stay consistent with the rest of the file. Matters most once the UI surfaces both the combined total and the individual costs side by side.

sanil-23 (commented):
@justinhsu1477 the failure-path and poisoned-mutex fixes both landed cleanly and all the prior review threads are resolved, so the code is in good shape from my side.

The only red check now is E2E (Playwright / web lane 2/4). I dug into it: the failing test is `harness-cron-prompt-flow` (`strict mode violation: getByText(/deleted the morning reminder/i) resolved to 2 elements`) — a locator flake in the cron-prompt harness suite, completely unrelated to this Rust-only sync-audit change. Lanes 1, 3, and 4 are green. A re-run of that lane should clear it; once CI is fully green I'll come back and approve.

One non-blocking nit left inline on `combined_cost_usd()` — worth a look before the Sync History UI follow-up wires it up. Nice work on this.

justinhsu1477:
Thanks @senamakel for the finisher pass — your failure-path fix is cleaner than what I'd have done. Carrying the tally in the error variant (`Result<(SyncOutcome, ComposioUsage), (String, ComposioUsage)>`) keeps the `Result` shape instead of my `(Result, Usage)` tuple, the poisoned-mutex `into_inner()` recovery is the right call, and the `actual_charged_usd` vs `estimated_cost_usd` split is a nice enrichment. No notes from me — looks good.

Following up on @graycyrus's **periodic.rs** observation, since it's the one open thread:

I checked, and `periodic.rs::run_one_tick` writes **no** `SyncAuditEntry` at all — it calls `provider.sync()` directly and only does `record_sync_success` + tracing on the result. The only place that writes audit entries is the `memory_sources::sync` dispatcher (manual/RPC path). So periodic syncs accumulate the tally on their `ProviderContext` and discard it, but **there's no zero-cost audit row being produced** — periodic syncs have never been in the audit log, before or after this PR. So no under-reporting/regression here; the `usage: Default::default()` on the periodic ctx is just an unused field on that path.

That said, @graycyrus is right that periodic is where most Composio cost actually accrues (every 20 min vs. rare manual syncs), so an audit log that only captures manual syncs paints a misleadingly-low cost picture. **Auditing periodic syncs is a worthwhile follow-up** — it's a larger change (periodic has no per-source audit shape / duration tracking today), so out of scope for #3111, but I'm happy to file a tracking issue for it if you'd like.

Issue #3120, opened 2026-06-01
Add customizable font size and chat text area scaling options


iamrhn:
## Summary

The app currently does not provide controls for adjusting font size or scaling the chat text area. Adding accessibility and readability options would improve the experience for users with different screen sizes, visual preferences, and accessibility needs.

## Problem

Chat text may appear too small or too large depending on the user's device and display settings.

Additionally, the message input area and overall chat layout cannot be adjusted, which can make reading long conversations or composing longer messages less comfortable.

## Proposed Solution

- Add a font size setting (Small, Medium, Large, Extra Large).
- Allow users to adjust chat/message text scaling independently of system settings.
- Add a setting to resize or expand the message input area for longer prompts.

PR #3139, opened 2026-06-01
feat(chat): auto-resize composer textarea to fit content

graycyrus:
## Summary

- Chat composer textarea now **auto-grows** as the user types or adds newlines, up to ~4 lines (~96 px), then becomes scrollable
- Previously the textarea had a fixed `max-h-32` (128 px) regardless of content — wasted space for short messages
- Implemented via a `useEffect` on `inputValue` that measures `scrollHeight` and caps at `COMPOSER_MAX_HEIGHT`

## Problem

- The message input area cannot be adjusted, making composing longer messages less comfortable
- Fixed height wastes space for single-line messages and doesn't provide visual feedback as text grows

## Solution

- Added a `useEffect` that fires on every `inputValue` change, resets the textarea height to `auto`, then sets it to `min(scrollHeight, 96px)`
- When content exceeds 96px (~4 lines), `overflowY` switches from `hidden` to `auto` (scrollable)
- On send (inputValue cleared), the effect fires again and shrinks back to single line
- The inline completion ghost div uses `absolute inset-0` so it tracks the textarea height automatically

## Submission Checklist

- [x] N/A: Tests added or updated — behavior-only CSS change, no testable logic added. The auto-resize is a side-effect-only `useEffect` on DOM measurement.
- [x] N/A: Diff coverage ≥ 80% — no new branches or logic to cover; the change is 3 lines of DOM manipulation in a `useEffect`.
- [x] N/A: Coverage matrix updated — behaviour-only change, no new feature rows
- [x] N/A: All affected feature IDs from the matrix are listed — no matrix features affected
- [x] N/A: No new external network dependencies introduced — pure frontend DOM change
- [x] N/A: Manual smoke checklist updated — not a release-cut surface
- [x] Linked issue closed via `Refs #3120` in the Related section

## Impact

- Desktop only (React frontend)
- No performance implications — `useEffect` on `inputValue` is already the existing render path
- No security, migration, or compatibility implications

## Related

- Refs #3120
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: feat/font-size-scaling
- Commit SHA: b9abda2e3

### Validation Run

- [x] `pnpm typecheck`
- [x] N/A: Focused tests — no testable logic added
- [x] N/A: Rust fmt/check — no Rust changes
- [x] N/A: Tauri fmt/check — no Tauri changes

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: Composer textarea auto-grows with content up to ~4 lines, then scrolls
- User-visible effect: More natural typing experience — input area expands as you write

### Parity Contract

- Legacy behavior preserved: Messages still send the same way, textarea resets on send

- Duplicate PR(s): None
- Canonical PR: This one
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * The message composer textarea now auto-resizes as you type, expanding vertically up to a capped maximum height for improved usability.

* **Tests**
  * Notification-related tests updated to more reliably assert notification behavior and messages.
  * Layout-related unit test adjusted to assert node-radius growth behavior for summary items while keeping other sizes unchanged.

### DIFF
--- a/app/src/pages/Conversations.tsx
+++ b/app/src/pages/Conversations.tsx
@@ -288,6 +288,8 @@ const Conversations = ({
   }, [agentProfiles, profileDraft.agentId, t]);
 
   const textInputRef = useRef<HTMLTextAreaElement>(null);
+  /** Max composer height ≈ 4 lines of text-sm + padding. */
+  const COMPOSER_MAX_HEIGHT = 96;
   const isComposingTextRef = useRef(false);
   const pendingSendRef = useRef<string | null>(null);
   const mediaRecorderRef = useRef<MediaRecorder | null>(null);
@@ -1138,6 +1140,16 @@ const Conversations = ({
     });
     return () => window.cancelAnimationFrame(id);
   }, [selectedThreadId, composerInteractionBlocked, inputMode]);
+
+  // Auto-resize composer textarea: grow with content, cap at COMPOSER_MAX_HEIGHT, then scroll.
+  useEffect(() => {
+    const ta = textInputRef.current;
+    if (!ta) return;
+    ta.style.height = 'auto';
+    ta.style.height = `${Math.min(ta.scrollHeight, COMPOSER_MAX_HEIGHT)}px`;
+    ta.style.overflowY = ta.scrollHeight > COMPOSER_MAX_HEIGHT ? 'auto' : 'hidden';
+  }, [inputValue]);
+
   const isSending = Boolean(
     selectedThreadId &&
     (pendingSendingThreadId === selectedThreadId ||
@@ -2189,7 +2201,7 @@ const Conversations = ({
                     placeholder={t('chat.typeMessage')}
                     rows={1}
                     disabled={composerInteractionBlocked || isSending}
-                    className="relative z-10 w-full resize-none border-0 bg-transparent pl-4 pr-10 py-2.5 text-sm leading-normal whitespace-pre-wrap break-words font-sans text-stone-900 dark:text-neutral-100 placeholder:text-stone-400 dark:placeholder:text-neutral-500 outline-none focus:outline-none focus-visible:outline-none focus:ring-0 focus-visible:ring-0 max-h-32 disabled:opacity-50 disabled:cursor-not-allowed"
+                    className="relative z-10 w-full resize-none border-0 bg-transparent pl-4 pr-10 py-2.5 text-sm leading-normal whitespace-pre-wrap break-words font-sans text-stone-900 dark:text-neutral-100 placeholder:text-stone-400 dark:placeholder:text-neutral-500 outline-none focus:outline-none focus-visible:outline-none focus:ring-0 focus-visible:ring-0 overflow-hidden disabled:opacity-50 disabled:cursor-not-allowed"
                   />
                   {/* Voice input mic hidden per #717 (inputMode='voice' path retained). */}
                 </div>
--- a/app/src/components/__tests__/OpenhumanLinkModal.notifications.test.tsx
+++ b/app/src/components/__tests__/OpenhumanLinkModal.notifications.test.tsx
@@ -1,5 +1,4 @@
-import { isTauri as coreIsTauri } from '@tauri-apps/api/core';
-import { act, fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import {
@@ -7,9 +6,13 @@ import {
   getNotificationPermissionState,
   showNativeNotification,
 } from '../../lib/nativeNotifications/tauriBridge';
+import { isTauri } from '../../services/webviewAccountService';
 import OpenhumanLinkModal, { OPENHUMAN_LINK_EVENT } from '../OpenhumanLinkModal';
 
-vi.mock('@tauri-apps/api/core', () => ({ isTauri: vi.fn() }));
+vi.mock('../../services/webviewAccountService', () => ({
+  isTauri: vi.fn(() => false),
+  purgeWebviewAccount: vi.fn(),
+}));
 
 vi.mock('../../lib/nativeNotifications/tauriBridge', () => ({
   ensureNotificationPermission: vi.fn(),
@@ -38,7 +41,7 @@ describe('OpenhumanLinkModal notifications test flow', () => {
   }
 
   it('shows success after permission is granted and native notification send succeeds', async () => {
-    vi.mocked(coreIsTauri).mockReturnValue(true);
+    vi.mocked(isTauri).mockReturnValue(true);
     vi.mocked(ensureNotificationPermission).mockResolvedValue(true);
     vi.mocked(showNativeNotification).mockResolvedValue({ delivered: true });
 
@@ -46,18 +49,19 @@ describe('OpenhumanLinkModal notifications test flow', () => {
     openNotificationsModal();
 
     fireEvent.click(screen.getByRole('button', { name: 'Send test notification' }));
-    await flushAsyncWork();
 
-    expect(
-      screen.getByText(/Test notification sent\. If you didn’t receive it/i)
-    ).toBeInTheDocument();
+    await waitFor(() => {
+      expect(
+        screen.getByText(/Test notification sent\. If you didn't receive it/i)
+      ).toBeInTheDocument();
+    });
     expect(showNativeNotification).toHaveBeenCalledWith(
       expect.objectContaining({ tag: 'welcome-notification-test' })
     );
   });
 
   it('shows actionable error when permission is denied and does not send notification', async () => {
-    vi.mocked(coreIsTauri).mockReturnValue(true);
+    vi.mocked(isTauri).mockReturnValue(true);
     vi.mocked(ensureNotificationPermission).mockResolvedValue(false);
     vi.mocked(getNotificationPermissionState).mockResolvedValue('denied');
 
@@ -73,7 +77,7 @@ describe('OpenhumanLinkModal notifications test flow', () => {
   });
 
   it('shows send failure message when native notification call fails', async () => {
-    vi.mocked(coreIsTauri).mockReturnValue(true);
+    vi.mocked(isTauri).mockReturnValue(true);
     vi.mocked(ensureNotificationPermission).mockResolvedValue(true);
     vi.mocked(showNativeNotification).mockResolvedValue({
       delivered: false,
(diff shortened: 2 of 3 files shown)


app/src/pages/Conversations.tsx
```diff
@@ -1138,6 +1140,16 @@ const Conversations = ({
...
   }, [selectedThreadId, composerInteractionBlocked, inputMode]);
+
+  // Auto-resize composer textarea: grow with content, cap at COMPOSER_MAX_HEIGHT, then scroll.
+  useEffect(() => {
+    const ta = textInputRef.current;
+    if (!ta) return;
+    ta.style.height = 'auto';
+    ta.style.height = `${Math.min(ta.scrollHeight, COMPOSER_MAX_HEIGHT)}px`;
+    ta.style.overflowY = ta.scrollHeight > COMPOSER_MAX_HEIGHT ? 'auto' : 'hidden';
+  }, [inputValue]);
```
sanil-23, line 1151: **[minor]** This effect only recomputes height when `inputValue` changes, so the textarea won't re-fit if its width changes (window resize, sidebar collapse/expand) while text is already present — a wrapped 3-line message can end up clipped or over-tall until the next keystroke. Not blocking for this PR, but worth a follow-up: a `ResizeObserver` on the textarea, or keying the effect on the container width, would close that gap.

sanil-23 (commented):
@graycyrus the composer auto-resize itself looks good — the `auto` -> `min(scrollHeight, MAX)` -> overflow toggle is the right pattern, and switching the ghost-completion layer to `absolute inset-0` so it tracks height is a clean touch. Nice catch fixing the notifications test too: it was mocking `isTauri` from `@tauri-apps/api/core`, but the modal actually imports it from `webviewAccountService`, so the old mock was a no-op.

Two things before this is approvable:

1. **CI is red.** E2E web lane 1/4 and Rust Core Coverage are failing. Both look unrelated to this diff (a frontend CSS change can't move Rust core coverage, and the E2E failure shows no composer assertions), so they're most likely flaky or pre-existing on main — but I can't approve over a red board. A rebase on latest main / re-run should sort it. Once it's green I'll come back and approve.

2. **Scope vs. `Closes #3120`.** Issue #3120 asks for three things: a font-size setting (S/M/L/XL), independent chat text scaling, and a resizable message input. This PR delivers the third (and as auto-grow rather than a user-facing setting), which is a solid improvement — but `Closes #3120` will auto-close the issue and silently drop the font-size and text-scaling work. Suggest changing it to `Refs #3120` (or `Part of #3120`) and leaving the issue open for the remaining items.

Nothing blocking in the code. Sort CI + the `Closes` keyword and this is good to go.

app/src/pages/Conversations.tsx
graycyrus: Good point — window/sidebar resizes could desync the height. Filed as a follow-up improvement. Not adding ResizeObserver in this PR to keep the scope tight.

sanil-23 (approved):
Approved.

oxoxDev (dismissed):
## Walkthrough

Composer-textarea auto-grow on `app/src/pages/Conversations.tsx` is sensible and small (6-LOC `useEffect`). Carries an orthogonal test-file change (3 different things bundled into `OpenhumanLinkModal.notifications.test.tsx`) that should be a separate commit/PR, and the `Closes #3120` keyword over-claims — issue asks for font-size setting + chat-area scaling, this ships only the second half. The pixel-hardcoded `COMPOSER_MAX_HEIGHT = 96` also contradicts issue #3120's font-scaling goal.

## Necessity check (carried over from Phase 1.5 gate)

### What is this?

- `Conversations.tsx +13/-1` — `useEffect` on `inputValue` measures `scrollHeight`, sets `style.height` to `min(scrollHeight, 96px)`, flips `overflowY` between `hidden` and `auto` past the cap. Replaces the prior static `max-h-32` (128 px) Tailwind class with `overflow-hidden` + dynamic style.
- `OpenhumanLinkModal.notifications.test.tsx +20/-15` — **unrelated drive-by**: (a) switches the `isTauri` mock target from `@tauri-apps/api/core` to the project's `services/webviewAccountService` source-of-truth wrapper, (b) adds a `purgeWebviewAccount: vi.fn()` stub to match the wrapper's exported surface, (c) swaps `flushAsyncWork()` calls for `waitFor()`. Has nothing to do with the composer change.
- Branch name `feat/font-size-scaling` mismatches the PR title — branch implies font-size-scaling feature; PR ships only the textarea auto-grow piece. Linked issue #3120 asks for **both** a font-size setting AND chat-area scaling; this PR addresses only the second half.

### Is it really needed?

- **Problem real?** Partially. Auto-grow textarea is a real UX irritation called out in #3120's body. Documented in a GitHub issue, so not speculative. **But** issue #3120 also asks for the font-size half, which this PR doesn't touch — `Closes #3120` is an overclaim.
- **Right size?** Mostly. Core change is 13 LOC of `useEffect` DOM measurement — proportionate. The drive-by test-file change adds 5 LOC of scope creep that belongs in its own commit / PR.
- **Right approach?** OK with caveats — pixel-hardcoded `COMPOSER_MAX_HEIGHT` doesn't honor user font scaling (the *other half* of #3120, ironic given the branch name). Imperative `ta.style.height` mutation is the standard react-textarea-autosize pattern, but a sibling-grep didn't find a `useAutosizeTextArea` hook — reasonable to keep this inline at 6 LOC, but a hook would make it reusable for the second textarea in `Conversations.tsx:1551` and any future composer.
- **Recommendation: `Trim`** — ship the textarea auto-grow piece on its own; the test-file mock migration should be a separate commit with its own message; switch `Closes #3120` → `Refs #3120` since the font-size half is still owed.

## Changes

| File | Summary |
| --- | --- |
| `app/src/pages/Conversations.tsx` | New `useEffect([inputValue])` resizes composer textarea to `min(scrollHeight, 96px)`, swaps `max-h-32` (128 px) → `overflow-hidden` + dynamic style. |
| `app/src/components/__tests__/OpenhumanLinkModal.notifications.test.tsx` | **Orthogonal**: migrates `isTauri` mock target to `services/webviewAccountService` wrapper + adds `purgeWebviewAccount` stub + replaces `flushAsyncWork()` with `waitFor()`. |

## Actionable comments (4)

### Major

#### 1. `Closes #3120` overclaims — issue is two features, PR is one

Issue #3120 is titled "Add customizable font size and chat text area scaling options" and the body explicitly asks for both:

> Add a font size setting (…) Additionally, the message input area and overall chat layout cannot be adjusted (…)

This PR ships only the textarea-area piece. Merging this with `Closes #3120` auto-closes the issue and drops the font-size half on the floor. Two options:

**Suggested change (PR body):**

```diff
- ## Related
-
- - Closes #3120
+ ## Related
+
+ - Refs #3120 (addresses the chat-area-scaling sub-problem; font-size-setting half still owed)
```

Alternatively: open a sub-issue for the font-size half and `Closes` that sub-issue from this PR.

#### 2. Pixel-hardcoded `COMPOSER_MAX_HEIGHT = 96` will silently shrink to <4 lines once font scaling lands

The other half of #3120 is a user-controlled font-size setting. Once that lands, a user who picks a larger font size will see this composer cap at fewer than 4 lines — but the constant's doc comment claims 4 lines:

```js
/** Max composer height ≈ 4 lines of text-sm + padding. */
const COMPOSER_MAX_HEIGHT = 96;
```

Either derive the cap dynamically from `parseFloat(getComputedStyle(ta).lineHeight)` × 4 + vertical padding, or define `COMPOSER_MAX_LINES = 4` and convert at measurement time:

**Suggested change (Conversations.tsx around the new useEffect):**

```tsx
const COMPOSER_MAX_LINES = 4;
// ...
useEffect(() => {
  const ta = textInputRef.current;
  if (!ta) return;
  const lineHeight = parseFloat(getComputedStyle(ta).lineHeight) || 20;
  const verticalPadding =
    parseFloat(getComputedStyle(ta).paddingTop) +
    parseFloat(getComputedStyle(ta).paddingBottom);
  const maxHeight = lineHeight * COMPOSER_MAX_LINES + verticalPadding;
  ta.style.height = 'auto';
  ta.style.height = `${Math.min(ta.scrollHeight, maxHeight)}px`;
  ta.style.overflowY = ta.scrollHeight > maxHeight ? 'auto' : 'hidden';
}, [inputValue]);
```

This keeps the visible-line contract (4 lines) regardless of whatever font size the user (eventually) picks.

#### 3. Test-file change is orthogonal — should be its own commit/PR

`OpenhumanLinkModal.notifications.test.tsx` does three unrelated things (isTauri mock-target migration, purgeWebviewAccount stub, flushAsyncWork→waitFor migration). All three are sensible per project conventions, but none of them belong in a "composer textarea auto-resize" PR. Reviewers shouldn't have to audit Tauri-mock surface changes in a CSS-shape PR.

**Suggested action**: pull the test-file diff out into a separate commit titled `test(modal): migrate isTauri mock to webviewAccountService wrapper + waitFor migration`, push as either its own PR or at minimum a separate commit on this branch so it can be reviewed (or reverted) independently.

### Minor

#### 4. Cap dropped from 128 px → 96 px is a silent UX regression for 5-line typers

Prior class `max-h-32` = 128 px. New cap = 96 px. Anyone who used to see all 5 lines of their longer message now scrolls inside a 4-line viewport. Not a bug, but the PR body claims "expands as you write" without noting the new ceiling is 32 px lower than before. Mention in the Summary so users who notice the change have a reference.

## Nitpicks (1)

- `Conversations.tsx:2178` — direct imperative `ta.style.height` mutation is the right pattern, but a 2-line comment explaining "imperative because Tailwind class would re-clamp on re-render; React won't fight us because className has `overflow-hidden`, no h-*" would save the next reader a `git blame`.

## Outside the diff

- `app/src/pages/Conversations.tsx:1551` has a second textarea (agent-profile editor) that doesn't get auto-resize. If the goal is composer-only, that's correct. If autoresize is generally desired for any textarea, extract to a `useAutosizeTextArea(ref, maxHeight)` hook in `app/src/hooks/` and apply at both sites in a follow-up.
- `app/src/hooks/useStickToBottom.ts` exists and uses `scrollHeight` for a related purpose; the new effect should briefly mention in a code comment that the two interact correctly (auto-resizing the composer changes layout but not the message-list scroll anchor since the composer lives below the messages).

## Verified

- No pre-existing `useAutosizeTextArea` hook — grep on `app/src/hooks/**` and `scrollHeight` calls in `app/src/**` confirms no reinvention.
- SSR-safety not a concern — this is a Tauri/Vite SPA, no server-side render of `Conversations.tsx`.
- First-render timing: `useEffect` runs after layout; initial `rows={1}` matches the dynamic minimum, so no visible flash.
- The inline-completion ghost `<span className="invisible">{inputValue}</span>` at `Conversations.tsx:2173` tracks the textarea height because it's inside the same absolutely-positioned wrapper, not affected by the parent textarea's height directly.
- 2 failing CI lanes (Playwright lane 1/4 + Rust Core Coverage) — PR touches zero Rust files, so Rust Core Coverage is structurally impossible to be PR-caused; Playwright lane 1/4 has been failing recurrently on `main` across multiple PRs this week. Both classify as upstream/main flake, not blocking.
- `Conversations.tsx` only has one textarea wired to `textInputRef`; the new effect targets the composer correctly.
- `Closes #3120` is the only side-effect-bearing closing keyword in the body; fix it and the issue stays open for the font-size half.

PR #3164, opened 2026-06-01
Fix memory letterhead local timestamp display

### LINKED ISSUE #3128: Email timestamps sometimes show in UTC instead of local time

Subharup-31:
Fixes #3128.

## Problem
Memory chunk letterheads format timestamps with UTC date/time getters and append `utc`. That can make email and memory timestamps appear offset from the viewer's local time.

## Approach
Format the letterhead timestamp from local `Date` components instead of UTC components, and remove the UTC suffix. The tests cover the local-time path and sender/recipient rendering.

## Tradeoffs
This keeps the compact numeric letterhead format, but the displayed time now follows the user's local environment instead of a fixed UTC representation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Date display in memory chunk letterhead now uses local time and shows localized YYYY·MM·DD · HH:MM formatting.
  * Person names now normalize underscores and dashes consistently for display.
  * Leading/trailing whitespace around source identifiers is trimmed.
  * Component falls back to showing owner when recipient info is missing.

* **Tests**
  * Added and updated tests covering localized dates, trimming, name normalization, and fallback behavior.

### DIFF
--- a/app/src/components/intelligence/MemoryChunkLetterhead.tsx
+++ b/app/src/components/intelligence/MemoryChunkLetterhead.tsx
@@ -38,12 +38,12 @@ function parseSourceParts(chunk: Chunk): LetterheadParts {
 
 function formatLetterDate(ms: number): string {
   const d = new Date(ms);
-  const yyyy = d.getUTCFullYear();
-  const mm = String(d.getUTCMonth() + 1).padStart(2, '0');
-  const dd = String(d.getUTCDate()).padStart(2, '0');
-  const hh = String(d.getUTCHours()).padStart(2, '0');
-  const mi = String(d.getUTCMinutes()).padStart(2, '0');
-  return `${yyyy}·${mm}·${dd} · ${hh}:${mi} utc`;
+  const yyyy = d.getFullYear();
+  const mm = String(d.getMonth() + 1).padStart(2, '0');
+  const dd = String(d.getDate()).padStart(2, '0');
+  const hh = String(d.getHours()).padStart(2, '0');
+  const mi = String(d.getMinutes()).padStart(2, '0');
+  return `${yyyy}·${mm}·${dd} · ${hh}:${mi}`;
 }
 
 export function MemoryChunkLetterhead({ chunk }: { chunk: Chunk }) {
--- a/app/src/components/intelligence/MemoryChunkLetterhead.tsx
+++ b/app/src/components/intelligence/MemoryChunkLetterhead.tsx
@@ -12,17 +12,19 @@ interface LetterheadParts {
 }
 
 function parseSourceParts(chunk: Chunk): LetterheadParts {
-  const left = chunk.source_id.split('|');
-  const senderRaw = left[0];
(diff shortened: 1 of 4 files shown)


graycyrus (commented):
@Subharup-31 the code looks correct — clean, focused fix.

The UTC → local time switch is the right call. Displaying timestamps built from `getUTC*` methods while labeling them with a "utc" suffix was at least honest, but the UX intent is clearly to show the user's local time. Switching to `get*` methods and dropping the suffix is the right behavior.

Two notes on the test:

1. The first test now uses a regex (`/2026·05·04 · \d{2}:\d{2}/`) instead of an exact string — that's the correct trade-off since the hour/minute will vary by CI runner timezone. The precision is covered by the new dedicated test.

2. For the `LocalTimeDate` approach: the implementation works, but `vi.useFakeTimers()` + `vi.setSystemTime(new Date('2026-05-04T14:44:00'))` is more idiomatic Vitest for this kind of test and exercises the real `Date` constructor path more naturally. Worth a refactor if you're touching this file again, but not a blocker.

CI has several checks still pending. Code looks good — I'll come back and approve once everything is green.

senamakel:
was merged recently in another PR. thanks for contributing though. greatly appreciated.

Issue #3148, opened 2026-06-01
feat: always-on voice command → system action (listen, understand, execute)


M3gA-Mind:
## Summary

Enable the app to continuously listen to the user, understand spoken commands, and perform system actions on the laptop — e.g., saying *"open my Music player"* causes the app to open it, without any hotkey press or manual send.

---

## Current State (Deep Dive Findings)

The app already has a production-grade voice stack. The bones are excellent — the gap is **continuous listening + auto-dispatch**.

### ✅ Already Built

| Capability | Location |
|---|---|
| Microphone capture (16 kHz mono, VAD, silence gating) | `src/openhuman/voice/audio_capture.rs` |
| Speech-to-text — cloud (OpenAI, Deepgram) + local (Whisper.cpp) | `src/openhuman/inference/voice/` |
| Text-to-speech — cloud (ElevenLabs) + local (Piper) | `src/openhuman/inference/voice/` |
| Hotkey-triggered dictation (push-to-talk / toggle) | `src/openhuman/voice/hotkey.rs`, `app/src/hooks/useDictationHotkey.ts` |
| LLM agent reasoning over transcribed text | `src/openhuman/agent/` |
| Shell tool (`open -a Music`, `xdg-open`, `cmd /c start`) | `src/openhuman/tools/impl/system/shell.rs` |
| Computer control — mouse + keyboard (disabled by default) | `src/openhuman/tools/impl/computer/` |
| Full voice settings UI (provider selection, model install) | `app/src/components/settings/panels/VoicePanel.tsx` |
| MicComposer recorder UI with retry + fallback codecs | `app/src/features/human/MicComposer.tsx` |

### ❌ Missing for Seamless "Listen & Execute"

| Feature | Impact |
|---|---|
| **Always-on microphone loop** | Currently requires a hotkey press to start listening |
| **Wake-word detection** (e.g., "Hey OpenHuman") | No engine (Porcupine, custom, etc.) integrated |
| **Auto-send after transcription** | Transcribed text lands in chat composer but isn't auto-dispatched to the agent |
| **Fast local command router** | Every voice command goes through a full LLM turn (1–3s); trivial commands like "open Music" could be routed locally in <100ms |
| **Shell allowlist for app-launching** | `open -a Music` triggers an approval prompt in supervised mode (default) — needs explicit allowlist or autonomy config |
| **Voice confirmation/feedback loop** | No "Opening Music — confirm?" spoken reply before executing |

---

[...]

PR #3168, opened 2026-06-01
feat(voice): auto-send dictation transcript + allowlist app-launch commands (#3148 Phase 1)

M3gA-Mind:
## Summary

- **Auto-send dictation transcripts**: hotkey-triggered voice transcripts send straight to the agent instead of landing in the composer for a manual send.
- **`launch_app` tool**: opens any desktop app by name on **macOS, Linux, and Windows** — the scoped, no-shell-exposure path for "open X".
- **`ax_interact` tool (cross-platform)**: interact with any app's UI by element label — **macOS via AXUIElement, Windows via UI Automation (UIA)**. Actions `list` / `press` / `set_value`, with a `filter` on `list` to keep results small and avoid truncation-driven hallucination. Mutating actions (`press`/`set_value`) are approval-gated; `list` is read-only.
- **Windows port**: native UIA backend (`uia_interact.rs`) wired behind the same `ax_interact` tool via cfg-dispatch, so the agent sees one tool on both platforms.
- **Docs**: `docs/voice-system-actions.md` tracks every change plus the Windows setup, cross-platform compatibility audit, and a mandatory Windows E2E test matrix.

## Problem

The agent could not act on the user's machine in response to voice/chat commands: dictation required a manual send, and there was no reliable, crash-safe, cross-platform way to launch and drive a running app's UI. An earlier attempt using synthetic input (enigo/`CGEventPost`) crashed the CEF host when events landed in OpenHuman's own renderer.

## Solution

- Frontend: `useDictationHotkey` tags transcripts with `autoSend`; `Conversations` dispatches them directly via a ref to the latest send fn.
- New tools `launch_app` and `ax_interact` under `tools/impl/{system,computer}`, wired into the registry, user filter, orchestrator tool scope, and frontend catalog.
- **Accessibility-API interaction, no synthetic events → no CEF crash.** `ax_interact` finds elements by semantic label (exact-match preferred), with a Rust-side `filter` on `list` (cap 60) so large trees can't overflow the tool-result budget.
- **Permission model**: `permission_level_with_args` returns `ReadOnly` for `list` and `Dangerous` for `press`/`set_value`; `external_effect_with_args` routes the mutating actions through the `ApprovalGate` before execute. `ax_press_element` rejects blank labels.
- **Launch safety**: OS-native launchers (`open`/`xdg-open`/`start`) are deliberately **not** in `READ_ONLY_BASES` — base-command classification can't see args, and they can open arbitrary URLs/handlers. App launching goes through the scoped `launch_app` tool instead.

### Windows setup (UI Automation)

- `src/openhuman/accessibility/uia_interact.rs` (new) — Windows UIA backend mirroring the macOS `ax_interact` fn surface (`list` / `press` / `set_value`) using the `uiautomation` crate. No helper process: the UIA COM API is called directly from Rust.
- `accessibility/mod.rs` cfg-dispatches `ax_interact` to the AXUIElement backend on macOS and the UIA backend on Windows, so the agent-facing tool is identical on both.
- `launch_app` Windows branch uses `Start-Process` (display name / Store-app AUMID / URI). Linux branch uses `gtk-launch` (desktop-id) with a derived-id fallback; `xdg-open` only for URIs.
- All platform-native code is `#[cfg(target_os = "…")]`-gated; non-matching arms return a clean runtime error so the build stays green everywhere. Permissions: macOS needs Accessibility; Windows needs none for non-elevated apps (UIPI blocks driving elevated apps).
- **Windows verification is still pending on a Windows machine** — `docs/voice-system-actions.md` carries the required E2E matrix (Calculator deterministic test, Notepad set_value, launch incl. UWP/URI, filtered-list correctness, Chromium/Electron tree exposure, elevation, agent-in-the-loop, + a macOS regression re-run).

## Submission Checklist

- [x] N/A: changed-line coverage — behaviour is platform-native UI automation exercised via `#[ignore]` integration tests; unit tests cover the pure logic (validation, schema, permission gating).
- [x] N/A: behaviour + tooling change; no feature rows in the coverage matrix are affected.
- [x] N/A: does not touch release-cut smoke surfaces.
- [x] N/A: this is Phase 1 of #3148 — the issue stays open for later phases, so no `Closes` keyword.

## Impact

- Desktop. `launch_app` and `ax_interact` cover macOS + Windows (Linux best-effort for launch); frontend auto-send is cross-platform.
- Security: app launchers stay out of the read-only command set (no approval-free arbitrary-URL launches); `ax_interact` `press`/`set_value` are approval-gated, `list` is read-only and injects no synthetic input.
- No migration or perf implications.

## Related

- Part of #3148 (Phase 1). Not closing — later phases remain.
- Follow-up PR(s)/TODOs: Phase 2 always-on listening; Windows E2E verification of the UIA backend (matrix in `docs/voice-system-actions.md`).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: feat/voice-always-on
- Commit SHA: 0a182989d3c802efead5f8e44f076887e0c7ec5a

### Validation Run

- [x] `pnpm typecheck`
- [x] Focused tests: `cargo test ax_interact`, `cargo test policy_command`, `pnpm debug unit src/hooks/__tests__/useDictationHotkey`
- [x] Rust fmt/check (if changed): `cargo check --manifest-path Cargo.toml`
- [x] N/A: Tauri shell unchanged in this branch.

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: agent can launch and interact with desktop apps (macOS + Windows); hotkey dictation auto-sends.
- User-visible effect: "open Music" opens it; spoken commands send without a manual step; UI controls can be pressed/filled by label.

### Parity Contract

- Legacy behavior preserved: non-`autoSend` dictation still inserts into the composer; command classification floor unchanged (launchers are NOT added to the read set — they route through `launch_app`).
- Guard/fallback/dispatch parity checks: non-matching `cfg` arms of the AX/UIA tools return a clean error; cfg-dispatch keeps one agent-facing tool across platforms; no build-time divergence.

- Duplicate PR(s): N/A
- Canonical PR: #3168

## Summary by CodeRabbit

* **New Features**
  * Launch desktop applications across macOS, Linux, and Windows.
  * App UI automation for listing, pressing, and setting text in UI elements (macOS and Windows).
  * Dictation can auto-send transcripts directly to chat.

* **Documentation**
  * Expanded system-actions roadmap, agent guidance, and UI-interaction workflows.
  * Clarified shell/tool descriptions and command classification guidance.

* **Chores / Config**
  * Added opt-in setting to enable mutating UI interactions; permission gating and safety defaults applied.


app/src/pages/Conversations.tsx
```diff
@@ -304,6 +304,8 @@ const Conversations = ({
   // from `selectedThreadId` so switching threads mid-turn doesn't move the
   // timer's reference point.
   const sendingThreadIdRef = useRef<string | null>(null);
+  // Ref so the mount-time dictation event handler can call the latest send fn.
+  const handleSendMessageRef = useRef<((text?: string) => Promise<void>) | null>(null);
```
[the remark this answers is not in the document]
M3gA-Mind: **Addressed (no change needed):** `handleSendMessageRef.current = handleSendMessage` runs synchronously during render, which React completes *before* it commits the mount effect that registers the `dictation://insert-text` listener — and well before any async Socket.IO dictation event can arrive. So the ref is always populated by the time `autoSend` fires; the `?.` is just defensive. There is no window where an event can hit an unset ref.

src/openhuman/accessibility/uia_interact.rs
```diff
@@ -0,0 +1,295 @@
...
+            contains = Some(w);
+        }
+    }
+    contains.ok_or_else(|| {
+        format!(
+            "No open window matches app '{app_name}'. Make sure it is running \
+             (try launch_app first), then retry."
+        )
+    })
+}
```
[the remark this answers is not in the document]
M3gA-Mind: **Good catch — flagged for the Windows follow-up.** A top-level Window's `Name` is the title (e.g. "Document1 - Microsoft Word"), not the app name. The current `find_window_for_app` mitigates the common case via the substring fallback (`nl.contains(needle)`), and exact title matches still win when they occur. The robust fix is to match by the window's owning process rather than its title — resolve the launched app to a PID and compare `UIElement::get_process_id()`. Tracking as a Windows-port refinement: this is `#[cfg(target_os = "windows")]` code under active development on a Windows machine and isn't compiled in the (Linux/macOS) CI lane, so the COM change should be made + verified there rather than blind-edited from here.

src/openhuman/tools/impl/system/launch_app.rs
```diff
@@ -0,0 +1,374 @@
...
+        })
+    }
+
+    fn permission_level(&self) -> PermissionLevel {
+        // Launching an app is a user-initiated, non-destructive, non-persistent
+        // action — treat it as read-only so the approval gate never fires.
+        PermissionLevel::ReadOnly
+    }
+
+    async fn execute(&self, args: serde_json::Value) -> anyhow::Result<ToolResult> {
```
oxoxDev, line 135: Blocker (gate bypass): hard-coding `ReadOnly` and never overriding `external_effect_with_args` means this never routes through the ApprovalGate (traits.rs:226-235) — app launches run with zero confirmation under any tier, including Full. Launching an app is an external side effect (and via the URI fallbacks can trigger network/system actions), so by traits.rs:226-232's own definition it should gate. Contrast shell.rs:164. Fix: `permission_level -> Execute` + `fn external_effect_with_args(&self, _) -> bool { true }`.

src/openhuman/tools/impl/system/launch_app.rs
```diff
@@ -0,0 +1,374 @@
...
+        .map_err(|e| format!("failed to invoke `xdg-open`: {e}"))?;
+
+    if xdg.success() {
+        Ok(format!("Opened '{app_name}'."))
+    } else {
+        Err(format!(
+            "Could not launch '{app_name}' on Linux. gtk-launch needs a .desktop \
+             ID (e.g. 'google-chrome'); xdg-open only opens URIs/paths, not app names. \
+             Try the .desktop ID, or supply a URI."
+        ))
```
oxoxDev, line 293: Blocker (URI smuggling): this `xdg-open <app_name>` fallback (and Windows Start-Process, ~line 298) re-introduces the capability policy_command.rs just removed from READ_ONLY_BASES. `validate_app_name` blocks `/ \ ..` but NOT a slashless scheme like `spotify:`/`slack:`/`mailto:` — those pass and fire an arbitrary registered URI handler, ungated. Fix: reject `^[a-zA-Z][a-zA-Z0-9+.-]*:` in validate_app_name so the tool is 'named applications only' as the docstring claims.

src/openhuman/tools/impl/system/launch_app.rs
```diff
@@ -0,0 +1,374 @@
+//! Tool: launch_app — open a named application on the user's desktop.
+//!
+//! A dedicated, narrow-scope alternative to using the `shell` tool with
+//! `open -a <App>` / `xdg-open` / `Start-Process`. Because it only launches
+//! named applications it carries no shell injection risk, does not require
+//! `workspace_only = false`, and is always-allow regardless of autonomy tier.
+//!
```
oxoxDev, line 13: Nit (claim accuracy): injection-safe (true) does not imply side-effect-free. Opening app windows / firing URI handlers on the user's machine is an external effect — please don't equate 'no shell injection' with 'always-allow, no gate regardless of tier'. They're independent; see the gating comment below.

src/openhuman/security/policy_command.rs
```diff
@@ -511,6 +511,13 @@ const READ_ONLY_BASES: &[&str] = &[
     "lsblk",
     "lscpu",
     "cut",
+    // NOTE: OS-native launchers (`open`, `xdg-open`, `start`) are deliberately
+    // NOT in the read-only set. `classify_command` only sees the base command,
+    // not its args, and these launchers can open arbitrary `https://` URLs and
+    // custom URI handlers — i.e. trigger outbound network / system actions — so
```
oxoxDev, line 517: Good tightening — `open`/`xdg-open`/`start` correctly stay out of READ_ONLY_BASES (they can open arbitrary URLs/URI handlers). Note the new `launch_app` tool currently re-opens that same surface ungated (xdg-open/Start-Process accept URI schemes, and the tool never hits the gate). Apply the same threat model to both paths.

src/openhuman/tools/user_filter.rs
```diff
@@ -28,6 +28,27 @@ const TOOL_FAMILIES: &[ToolFamily] = &[
...
+        id: "launch_app",
+        rust_names: &["launch_app"],
+        default_enabled: true,
+    },
+    // AXUIElement interaction: semantic UI control via macOS Accessibility API.
+    // No CGEventPost, no coordinate dependency, no CEF crash risk.
+    ToolFamily {
+        id: "ax_interact",
+        rust_names: &["ax_interact"],
+        default_enabled: true,
```
oxoxDev, line 42: Blocker (security): `ax_interact` ships `default_enabled: true`, yet it can press any control / set any field in ANY running app (Keychain, 1Password, browser password fields, Terminal, System Settings). Its strictly-less-powerful sibling `computer_control` (just below) only registers behind a `computer_control.enabled` config flag. Give ax_interact the same treatment: gate the mutating press/set_value path behind an explicit opt-in flag, default mutating OFF, keep read-only `list` available.

src/openhuman/tools/impl/computer/ax_interact.rs
```diff
@@ -0,0 +1,316 @@
...
+            // type into fields) and change application state, so they must
+            // not ride on the read-only path.
+            _ => PermissionLevel::Dangerous,
+        }
+    }
+
+    fn external_effect_with_args(&self, args: &serde_json::Value) -> bool {
+        // Route mutating actions through the ApprovalGate before execute();
+        // `list` is a pure read and flows through unprompted.
+        !matches!(
```
oxoxDev, line 159: `external_effect_with_args` correctly routes press/set_value through the ApprovalGate — good, no hardcoded-approve trap. But the gate auto-allows with NO prompt when there's no chat context (approval/gate.rs:225: cron/triage/proactive turns) and when the tool is in the auto_approve allowlist (gate.rs:205). For a tool that can drive ANY app's UI, an unattended/background turn can click Send/Delete/Pay or fill+submit a form unconfirmed. Fix: fail-closed for press/set_value on non-interactive turns + exclude these actions from 'Always allow' persistence.

src/openhuman/accessibility/ax_interact.rs
```diff
@@ -0,0 +1,152 @@
...
+        Err("ax_interact is supported on macOS and Windows only".into())
+    }
+}
+
+/// Press the first UI element in `app_name` whose label contains `label` (case-insensitive).
+///
+/// Rejects a blank `label`: with an empty needle the helper's `contains`
+/// match degenerates to match-all and would press the first named control it
+/// finds. Guard here rather than trusting every caller to pre-validate.
+pub fn ax_press_element(app_name: &str, label: &str) -> Result<String, String> {
```
oxoxDev, line 78: Blocker (security): no app allowlist/denylist — `app_name` flows straight to AXUIElementPerformAction / UIA Invoke / kAXValueAttribute set. The agent can actuate Keychain Access, 1Password, a browser's saved-password field, System Settings, or a terminal (helper.rs:556 even special-cases terminals as supported). set_value alone can't submit, but list -> set_value -> press on a Run/OK/Send button fills and submits in two calls. Add a sensitive-app denylist (bundle IDs / window names) that press/set_value refuse, or invert to a user-curated allowlist for mutating actions; `list` can stay broad.

src/openhuman/agent_registry/agents/orchestrator/agent.toml
```diff
@@ -117,6 +117,16 @@ named = [
...
+    # App launching — lets the orchestrator open desktop applications
+    # directly when asked ("open Music", "launch Spotify", etc.) without
+    # delegating or telling the user it can't. PermissionLevel::ReadOnly
+    # so no approval gate fires regardless of autonomy tier.
+    "launch_app",
+    # AXUIElement interaction — click buttons and type in fields by semantic
+    # label via macOS Accessibility API. No CGEventPost, no coordinate
+    # dependency, no CEF crash risk. Always call action='list' first to
+    # discover what elements are available before pressing.
+    "ax_interact",
```
oxoxDev, line 132: Two things: (1) `ax_interact` is in the orchestrator's default direct toolset, so the planning agent gets UI-actuation of any app out of the box — given the missing app scope + non-interactive gate bypass, consider removing the mutating capability from the default surface until a denylist + enable-flag land. (2) The nearby comment stating ax_interact has 'no approval gate regardless of autonomy tier' is wrong — press/set_value are `Dangerous` and DO gate interactively (only `list` is ReadOnly). Please scope that comment to `list` so it doesn't invite removing the real gating.

oxoxDev (changes requested):
Reviewed across three dimensions: launch_app command-exec safety, ax_interact/uia_interact UI-automation gating + scope, and the dictation auto-send + frontend/prompt wiring. Strong engineering in places — but for a feature that lets the agent launch arbitrary apps and drive any app's UI, there are several blocker-grade gaps in the gating/scope. Requesting changes.

## Blockers

**1. `launch_app` is ungated.** It hard-codes `PermissionLevel::ReadOnly` and never overrides `external_effect_with_args`, so it never routes through the ApprovalGate (traits.rs:226-235) — under any autonomy tier, including Full, app launches run with zero confirmation (the agent.toml comment states this as intended). Launching an app is an externally-observable side effect; by the codebase's own definition it should gate, like `shell.rs` does. Fix: `permission_level -> Execute` + `external_effect_with_args(..) -> true`.

**2. `launch_app` URI-handler smuggling.** The Linux `xdg-open`/Windows `Start-Process` fallbacks accept a slashless URI scheme (`spotify:`, `slack:`, `mailto:`); `validate_app_name` only blocks `/ \ ..`, so a `scheme:payload` passes and fires an arbitrary registered handler — re-opening the exact surface `policy_command.rs` just removed from `READ_ONLY_BASES` for the shell path. Fix: reject `^[a-zA-Z][a-zA-Z0-9+.-]*:` in `validate_app_name`.

**3. `ax_interact` bypasses approval on non-interactive turns.** The gate genuinely fires for interactive chat (good — not a hardcoded-approve trap), but `approval/gate.rs:225` returns `Allow` with no prompt when `chat_ctx.is_none()` — i.e. cron / triage / proactive / background turns run `press`/`set_value` unattended. For a tool that can actuate any control in any app, an automated turn can click Send/Delete/Pay or fill+submit a form with no confirmation. Plus the `auto_approve` allowlist short-circuit means one 'Always allow' makes every future press silent. Fix: fail-closed for mutating actions on non-interactive turns; exclude them from auto-approve persistence.

**4. `ax_interact` has no app scope.** No allow/denylist — `app_name` flows straight to AXUIElementPerformAction / UIA Invoke / kAXValueAttribute set. The agent can drive Keychain, 1Password, a browser's saved-password field, System Settings, or a terminal (helper.rs even special-cases terminals as supported). `set_value` can't submit alone, but `set_value` + a follow-up `press` on a Run/OK/Send button is a two-call fill-and-submit of a credential dialog. Fix: sensitive-app denylist (Keychain/1Password/security panes/terminals/sudo-UAC) that press/set_value refuse, or a user-curated allowlist for mutating actions. `list` (read-only) can stay broad.

**5. `ax_interact` is default-ON and orchestrator-exposed,** with weaker gating than its less-powerful sibling `computer_control` (which only registers behind a `computer_control.enabled` config flag). Mutating UI-actuation ships on by default in the orchestrator's direct toolset. Fix: default the mutating capability OFF behind an explicit config flag; keep read-only `list`.

## Clean / good
- Shell injection fully closed — arg-vector dispatch on all 3 OSes, PowerShell name passed via env not interpolation, solid `validate_app_name`.
- Interactive `press`/`set_value` gating genuinely fires via `external_effect_with_args` + `permission_level_with_args=Dangerous` — NOT a hardcoded-`approved` bypass.
- `list` is read-only, `filter` is a safe substring match.
- Dictation auto-send is safe: reuses `handleSendMessage`, so transcripts still run through `checkPromptInjection` + the empty/blocked guards; blank transcripts dropped twice; no finalize race. No `showSidebar`-style production masking — Conversations.tsx only adds the autoSend branch and the new tests are real.

## Doc / comment defects
- agent.toml comment says ax_interact has 'no approval gate regardless of autonomy tier' — wrong: press/set_value are `Dangerous` and DO gate interactively. The comment could mislead a maintainer into deleting the per-args gating that actually protects users. Scope it to `action='list'`.
- docs/voice-system-actions.md Change 1.2 claims `open`/`xdg-open` were added to `READ_ONLY_BASES` — shipped `policy_command.rs` does the opposite (deliberately excludes them). Mark 1.2 as reverted/superseded by 1.4 (`launch_app`).

CI: the one red lane (E2E Playwright 2/4, `harness-cron-prompt-flow.spec.ts` strict-locator) is an unrelated cron-flow flake, not PR-caused.

Net: gate `launch_app` + reject URI schemes; fail-closed + denylist + opt-in flag for `ax_interact` mutating actions; fix the comment + stale doc. Happy to re-review once the gating/scope lands.

src/openhuman/tools/impl/system/launch_app.rs
M3gA-Mind: **Fixed** (`8c24320dd`). `launch_app` now `permission_level → Execute` and `external_effect() → true`, so it routes through the `ApprovalGate` before `execute()` just like `shell` (traits.rs:226-232). No longer always-allow under any tier.

src/openhuman/tools/impl/system/launch_app.rs
M3gA-Mind: **Fixed** (`8c24320dd`). `validate_app_name` now rejects URI schemes via `^[a-zA-Z][a-zA-Z0-9+.-]*:` (new `is_uri_scheme`), so `spotify:` / `slack:` / `mailto:` / `https:` no longer slip past the `/ \ ..` checks into the `xdg-open` / `Start-Process` fallbacks. Named applications only, as the docstring claims. Added `validate_rejects_uri_schemes` test.

src/openhuman/tools/impl/system/launch_app.rs
M3gA-Mind: **Fixed** (`8c24320dd`). Reworded the module docstring: injection-safe does **not** imply side-effect-free, and the tool is no longer described as always-allow. It now explicitly states opening a window / firing a handler is an external effect that routes through the ApprovalGate.

src/openhuman/security/policy_command.rs
M3gA-Mind: **Fixed** (`8c24320dd`). Same threat model now applies to both paths: `open`/`xdg-open`/`start` stay out of `READ_ONLY_BASES`, **and** `launch_app` is gated (`external_effect=true`) + rejects URI schemes — so it no longer re-opens the ungated URI-handler surface.

src/openhuman/tools/user_filter.rs
M3gA-Mind: **Fixed** (`8c24320dd`). Mirrored the `computer_control.enabled` posture: added `computer_control.ax_interact_mutations` (default **false**). `press` / `set_value` are refused unless it's enabled; read-only `list` stays available. Wired through `tools/ops.rs`; `AxInteractTool::new(allow_mutations)`.

src/openhuman/tools/impl/computer/ax_interact.rs
M3gA-Mind: **Addressed** (`8c24320dd`). Two layers now sit in front of the gate so a background/auto-approved turn can't silently actuate UI: (1) mutating `press`/`set_value` are **opt-in, default off** (`computer_control.ax_interact_mutations`), so by default there is no mutation capability at all; (2) a sensitive-app **denylist** (Keychain, 1Password/Bitwarden/LastPass/Dashlane, System Settings, Terminal/iTerm, Console) refuses regardless of gate context. `ToolCallOptions` carries no interactive-context signal today, so a tool-layer "fail-closed on non-interactive turn" check isn't cleanly wireable here — the opt-in default-off + denylist are the defense-in-depth that doesn't depend on the gate. Happy to add an explicit interactive-turn signal as a fast-follow if you'd prefer that belt-and-braces.

src/openhuman/accessibility/ax_interact.rs
M3gA-Mind: **Fixed** (`8c24320dd`). Added a sensitive-app denylist (`SENSITIVE_APPS`): Keychain, 1Password, Bitwarden, LastPass, Dashlane, System Settings/Preferences, Terminal, iTerm, Console. All actions (incl. `list`) are refused when `app_name` matches — so the `list → set_value → press` submit chain can't target a password manager / Keychain / terminal. Enforced in the tool before any AX/UIA call, independent of the gate. Tests: `sensitive_apps_detected`, `refuses_sensitive_app_even_with_mutations`.

src/openhuman/agent_registry/agents/orchestrator/agent.toml
M3gA-Mind: **Addressed** (`8c24320dd`). (1) The mutating capability is now off the default surface in practice: `press`/`set_value` are opt-in (`computer_control.ax_interact_mutations`, default false) and deny-listed, so the orchestrator's default toolset only has the read-only `list` actionable out of the box. (2) Corrected the agent.toml comment — it now scopes ReadOnly/unprompted to `list` only and notes that `press`/`set_value` are `Dangerous`, gate interactively, are opt-in, and refuse the sensitive-app denylist.

oxoxDev (commented):
Re-reviewed after the security commits — strong turnaround, thanks. Verified resolved against this head:

- ✅ **launch_app gated** — now `Execute` + `external_effect→true`, routes through the ApprovalGate like `shell` (with a test).
- ✅ **launch_app URI smuggling** — `is_uri_scheme()` rejects `^[a-zA-Z][a-zA-Z0-9+.-]*:` inside `validate_app_name` (test covers spotify:/mailto:/slack:/https:/custom).
- ✅ **ax_interact default-ON** — mutating press/set_value now behind `computer_control.ax_interact_mutations` (default false); only read-only `list` is on by default.
- ✅ **Sensitive-app denylist** — `SENSITIVE_APPS` enforced for all actions, independent of the gate (good defense-in-depth).
- ✅ **agent.toml comment** — now accurate (only `list` ReadOnly; press/set_value Dangerous + gated).
- Windows port routes through the same tool chokepoint (denylist + opt-in + name guard before dispatch) — no new ungated path. CI is fully green now.

Three residuals (none CI-blocking — flagging, not hard-blocking):

**1. Non-interactive / auto-approve gate path (the original blocker #3) is only partially closed.** `approval/gate.rs` is unchanged: it still returns `Allow` with no prompt when `chat_ctx.is_none()` (cron/triage/proactive turns), and the `auto_approve` allowlist short-circuits on tool-name only. The opt-in flag + denylist genuinely close the worst cases, but once a user sets `ax_interact_mutations=true`, a background/automated turn can still `press`/`set_value` any **non-denylisted** app (Mail, Slack, a browser tab on a bank/pay page) unprompted — and one 'Always allow ax_interact' silences every future press, including interactive. Worth failing closed for Dangerous-level actions when there's no chat context, and excluding per-action Dangerous calls from auto-approve persistence.

**2. Terminal denylist is incomplete** — see inline.

**3. Doc still contradicts shipped code** — see inline.

Not blocking merge on these, but #1 is the one I'd most like to see gated before this capability ships widely.

src/openhuman/tools/impl/computer/ax_interact.rs
```diff
@@ -0,0 +1,420 @@
...
+    "1password",
+    "bitwarden",
+    "lastpass",
+    "dashlane",
+    "system settings",
+    "system preferences",
+    "terminal",
+    "iterm",
+    "console", // macOS Console (logs)
+];
```
oxoxDev, line 49: Denylist gap: `SENSITIVE_APPS` covers terminal/iterm/console, but helper.rs (~:557) also treats wezterm/warp/alacritty/kitty/ghostty/hyper/rio as terminals — those 7 are still drivable by press/set_value when mutations are enabled. Add them (or match the same terminal-app set helper.rs uses) so 'terminals are denied' actually holds.

docs/voice-system-actions.md
```diff
@@ -0,0 +1,485 @@
...
+
+### Change 1.2 — Shell allowlist for app-launching
+
+**Status:** ✅ Done  
+**Commit:** `7269f4373`
+
+**Problem:** `open -a Music` classified as `Write` → triggers approval prompt in Supervised mode.
+
+**Fix:**
+- `src/openhuman/security/policy_command.rs` — added `"open"` (macOS) and `"xdg-open"` (Linux) to `READ_ONLY_BASES`. These are OS-native app launchers that don't modify the workspace, so `Read` classification is correct.
```
oxoxDev: Still contradicts shipped code: policy_command.rs:514-520 deliberately keeps open/xdg-open/start OUT of READ_ONLY_BASES (with a comment). This Change 1.2 claims the opposite. Please mark it reverted/superseded by Change 1.4 (launch_app).

docs/voice-system-actions.md
```diff
@@ -0,0 +1,485 @@
...
+
+| Change | Cross-platform status | Notes for Windows |
+|---|---|---|
+| Auto-send dictation transcript (TS) | ✅ Fully portable | Pure frontend; no OS code. Works as-is. |
+| `launch_app` | ✅ macOS / Linux / Windows branches | Windows branch now uses `Start-Process` with a **Store/UWP (`Get-StartApps` AUMID) fallback** and injection-safe env passing (§1). |
+| `ax_interact` **tool** (`tools/impl/computer/ax_interact.rs`) | ✅ Functional on Windows | Delegates to `accessibility::ax_interact` fns, which now cfg-dispatch to the UIA backend on Windows. Description made OS-neutral. |
+| `accessibility::ax_interact` helpers | ✅ cfg-dispatched | macOS → Swift helper; Windows → `uia_interact.rs`; other → clean runtime error. |
+| `accessibility::uia_interact` (NEW) | ✅ Windows backend | UIA `list`/`press`/`set_value` via the `uiautomation` crate; same fn signatures as the macOS path. |
+| Swift unified helper (`accessibility/helper.rs`) | ⚠️ macOS-only by design | Windows needs no helper process — UIA COM API is called directly from Rust. |
+| Shell allowlist (`open`/`xdg-open`) | ✅ Done | `"start"` added to `READ_ONLY_BASES` in `policy_command.rs`. |
```
oxoxDev: Same correction: this Windows row claims `start` was added to READ_ONLY_BASES — it isn't (policy_command.rs:514). Drop the 'add start to READ_ONLY_BASES' items in the Windows section too.

src/openhuman/tools/impl/computer/ax_interact.rs
oxoxDev: Keeping this one open. The opt-in flag (`ax_interact_mutations`) + sensitive-app denylist are good mitigations and close the worst cases — but they do not close this thread, because `approval/gate.rs` itself is unchanged: it still returns `Allow` with no prompt when `chat_ctx.is_none()` (cron / triage / proactive / background turns), and the `auto_approve` short-circuit matches on tool name only (no per-action granularity). So once a user enables mutations, an automated turn can still `press`/`set_value` any non-denylisted app (Mail, Slack, a browser tab on a bank/pay page) unprompted, and one "Always allow ax_interact" silences every future press including interactive. To resolve: fail closed for Dangerous-level actions when there is no chat context, and exclude per-action Dangerous calls from auto-approve persistence.

src/openhuman/tools/impl/computer/ax_interact.rs
M3gA-Mind: **Fixed** (`89bb213ff`). `SENSITIVE_APPS` now includes the full terminal set helper.rs recognizes — `wezterm`, `warp`, `alacritty`, `kitty`, `ghostty`, `hyper`, `rio` (plus `terminal`/`iterm`). Test `sensitive_apps_detected` asserts all nine are denied.

docs/voice-system-actions.md
M3gA-Mind: **Fixed** (`89bb213ff`). Change 1.2 is now marked **REVERTED / SUPERSEDED** — it documents that `open`/`xdg-open`/`start` were removed from `READ_ONLY_BASES` (kept out deliberately at `policy_command.rs:514-520`) and that launching is now handled by the gated `launch_app` tool (Change 1.4). No longer contradicts shipped code.

docs/voice-system-actions.md
M3gA-Mind: **Fixed** (`89bb213ff`). Dropped the "add `start` to `READ_ONLY_BASES`" claims from the Windows section (compat table, quick-start, summary, and module list). They now state launchers stay OUT of `READ_ONLY_BASES` and `launch_app` (gated) is the Windows `Start-Process` path.

src/openhuman/tools/impl/computer/ax_interact.rs
M3gA-Mind: **Addressed** (`8c24320dd` + `89bb213ff`). Two gate-independent layers now stand in front of press/set_value so an unattended/background turn can't silently actuate UI: (1) they're **opt-in, default off** (`computer_control.ax_interact_mutations`) — no mutation capability at all by default; (2) the **sensitive-app denylist** (password managers, Keychain, System Settings, all terminal emulators, Console) refuses regardless of gate context. `ToolCallOptions` exposes no interactive-turn signal today, so a clean tool-layer "fail-closed on non-interactive turn" + "exclude from Always-allow persistence" needs a small plumbing change in the gate/options; tracking that as a focused fast-follow rather than bundling it here. The default-off + denylist already remove the unattended-actuation risk for the shipped default.

M3gA-Mind:
## Security review — all threads addressed ✅

Thanks for the thorough security pass, @oxoxDev. Summary of how each item was resolved (commits `8c24320dd` + `89bb213ff`):

**`launch_app`**
- Now `permission_level = Execute` + `external_effect() = true` → routes through the `ApprovalGate` before execute, like `shell` (no more always-allow under any tier).
- `validate_app_name` rejects URI schemes (`^[a-zA-Z][a-zA-Z0-9+.-]*:`) so the `xdg-open`/`Start-Process` fallbacks can't fire arbitrary registered handlers (`spotify:`/`mailto:`/`slack:`). Named applications only.
- Docstring corrected: injection-safe ≠ side-effect-free.

**`ax_interact`**
- Mutating `press`/`set_value` are now **opt-in, default off** via `computer_control.ax_interact_mutations`; read-only `list` stays available. Mirrors `computer_control.enabled` for mouse/keyboard.
- Per-call gating: `list` is `ReadOnly`/unprompted; `press`/`set_value` are `Dangerous` and route through the `ApprovalGate`.
- **Sensitive-app denylist** refuses all actions on password managers (1Password/Bitwarden/LastPass/Dashlane), Keychain, System Settings/Preferences, Console, and the full terminal-emulator set `helper.rs` recognizes (terminal/iterm/wezterm/warp/alacritty/kitty/ghostty/hyper/rio) — defense-in-depth that holds regardless of gate context.
- `orchestrator/agent.toml` comment corrected to scope ReadOnly/unprompted to `list` only.

**`policy_command.rs`**
- `open`/`xdg-open`/`start` stay **out** of `READ_ONLY_BASES` (kept out deliberately, with a comment); launching is handled solely by the gated `launch_app` tool.

**Docs**
- `voice-system-actions.md`: Change 1.2 marked REVERTED/SUPERSEDED; Windows-section claims that `start` was added to `READ_ONLY_BASES` removed.

**Fast-follow noted on the gate thread:** a clean tool-layer "fail-closed on non-interactive turn + exclude from Always-allow persistence" needs a small `ToolCallOptions`/gate plumbing change; tracked separately. The opt-in-default-off + denylist already remove the unattended-actuation risk for the shipped default.

Tests added: launch_app URI-reject + Execute/external_effect; ax_interact denylist (all 9 terminals), mutations-disabled refusal, per-arg permission/gate. All review threads resolved.

src/openhuman/tools/impl/computer/ax_interact.rs
oxoxDev: Fair enough — accepting this. With mutations default-off and the denylist refusing regardless of gate context, the unattended-actuation risk is removed for the shipped default, and the true gate-level fail-closed genuinely needs an interactive-turn signal that `ToolCallOptions` does not carry today. Agreed that belongs in a focused follow-up rather than this PR. One ask: please open a tracking issue for it (fail-closed for Dangerous-level actions when `chat_ctx.is_none()` + exclude per-action Dangerous calls from auto-approve persistence) so it does not get lost once the surface grows. Resolving on that basis. Terminal-denylist gap and the docs are confirmed fixed in 89bb213ff — thanks for the quick turnaround.

oxoxDev (approved):
Approving. The security concerns from the original review are resolved:

- `launch_app` now gates (`Execute` + `external_effect`), and `validate_app_name` rejects URI schemes — closes the ungated-launch + URI-handler-smuggling path.
- `ax_interact` mutating `press`/`set_value` are opt-in (default off, `computer_control.ax_interact_mutations`), so there is no mutation capability at all by default.
- Sensitive-app denylist refuses regardless of gate context, now covering password managers, Keychain, System Settings, Console, and the full terminal-emulator set (mirrors helper.rs).
- Interactive gating genuinely fires (Dangerous → ApprovalGate), `list` stays read-only, shell-injection is closed, and the dictation auto-send reuses the existing prompt-injection/empty guards.
- agent.toml comment + the stale READ_ONLY_BASES docs corrected.

One residual is deferred by agreement: a gate-level fail-closed for Dangerous actions on non-interactive turns (+ excluding them from auto-approve persistence) needs an interactive-turn signal that `ToolCallOptions` doesn't carry yet. The default-off + denylist remove that risk for the shipped default, so it's fine as a tracked fast-follow — please open an issue for it.

CI is re-running on 89bb213ff (docs + denylist-const only); the prior head was fully green. Nice work on the quick turnaround.

PR #3177, opened 2026-06-02
chore(ci): simplify release workflows — remove pretests, add commit picker and bump-only mode

senamakel:
## Summary

- Remove the `pretest-tests` job (unit + rust gate) from both `release-staging.yml` and `release-production.yml` — PR-level CI already covers this.
- Add `commit_sha` input to both workflows so builds can target a specific commit instead of main HEAD.
- Add `create_tag` input to staging: when false, bumps versions and commits but skips tag + build (version-bump-only mode).
- Update production so `create_release=false` also skips git tag creation.
- Remove the `skip_pretests` metadata input from `build-desktop.yml`.

## Problem

- The pretest gate in release workflows duplicated testing already run by PR CI, adding ~15 min to every release cut without catching new issues.
- No way to build from a specific commit — only main HEAD was supported.
- No way to bump versions without also creating a tag and running the full build matrix.

## Solution

- Deleted the `pretest-tests` job and all `skip_pretests` inputs/conditions across all three workflow files.
- Added optional `commit_sha` string input: when set, the workflow checks out that commit; when empty, uses main HEAD as before.
- Staging: added `create_tag` boolean (default true). When false, the workflow bumps versions, commits, and stops — no tag, no desktop build, no Docker, no Sentry deploy.
- Production: gated tag creation, tag-existence checks, and cleanup on `inputs.create_release` so the existing `create_release=false` mode now truly produces no tag.
- Changed `git push origin main` to `git push origin HEAD:main` to work correctly when checked out at a detached commit SHA.

## Submission Checklist

- [x] N/A: CI workflow YAML changes — no application code, no testable surface
- [x] N/A: no application code changed, diff-cover gate does not apply to workflow files
- [x] N/A: no feature rows affected — CI-only change
- [x] N/A: no feature IDs affected
- [x] N/A: no external network dependencies introduced
- [x] N/A: does not touch release-cut surfaces (it IS the release-cut surface definition)
- [x] N/A: no linked issue

## Impact

- **CI only** — no runtime, platform, or app-level changes.
- Release staging runs will be faster (no pretest gate).
- Operators get a version-bump-only mode and commit-targeting capability.

## Related

- Closes: N/A
- Follow-up PR(s)/TODOs: N/A

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: chore/simplify-release-workflows
- Commit SHA: 2382ae452

### Validation Run

- [x] N/A: `pnpm typecheck` — no TS changes
- [x] N/A: Focused tests — CI workflow YAML only
- [x] N/A: Rust fmt/check — no Rust changes
- [x] N/A: Tauri fmt/check — no Tauri changes

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: release workflows no longer run pretests; new inputs for commit targeting and bump-only mode
- User-visible effect: faster release cuts, new workflow dispatch options

### Parity Contract

- Legacy behavior preserved: default inputs reproduce previous behavior (create_tag=true, commit_sha="" = HEAD)

- Duplicate PR(s): N/A
- Canonical PR: this one
- Resolution: N/A

### DIFF
--- a/.github/workflows/build-desktop.yml
+++ b/.github/workflows/build-desktop.yml
@@ -84,14 +84,6 @@ on:
           the Tauri shell since #1061).
         type: boolean
         default: false
-      skip_pretests:
-        description: Metadata-only flag propagated by caller workflows when they
-          intentionally bypass the upstream pretest phase. This reusable build
-          workflow does not execute pretests itself, but it logs the policy so
-          the build run captures whether the normal release gate was relaxed for
-          this invocation.
-        type: boolean
-        default: false
       with_updater:
         description:
           When true, set `WITH_UPDATER=true` so the Tauri bundler emits the
@@ -146,7 +138,7 @@ jobs:
       - name: Log build policy
         shell: bash
         run: |
-          echo "[build-desktop] tag=${{ inputs.tag }} build_ref=${{ inputs.build_ref }} skip_pretests=${{ inputs.skip_pretests }}"
+          echo "[build-desktop] tag=${{ inputs.tag }} build_ref=${{ inputs.build_ref }}"
       - name: Set Xcode version
         if: matrix.settings.platform == 'macos-latest'
         uses: maxim-lobanov/setup-xcode@v1
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -28,15 +28,15 @@ on:
         default: patch
         type: choice
         options: [patch, minor, major]
-      skip_pretests:
+      commit_sha:
         description:
-          Skip the unit/rust pretest phase and continue directly to the build
-          matrix. Release workflows no longer run E2E. Use only when the
-          required pretest signal is already known and you need to unblock a
-          production cut.
+          Build from a specific commit SHA instead of main HEAD. Only
+          consulted when release_source = main_head; staging_tag promotions
+          resolve the commit from the tag. The commit must be reachable
+          from main. Leave empty to use main HEAD.
         required: false
-        type: boolean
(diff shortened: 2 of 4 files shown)


sanil-23 (approved):
Solid cleanup of the release pipelines. Dropping the `pretest-tests` gate is well-justified — PR CI already runs the unit + rust suites against the same commit, so re-running them at release time was pure latency with no new signal. The `always()` / skipped-status workarounds that existed only to thread `skip_pretests` through the build matrix are gone too, which makes the job graph noticeably easier to follow.

The new inputs are clean:
- `commit_sha` is read via an env var, validated against a hex regex, and checked for reachability with `git merge-base --is-ancestor` before any checkout or push — no template-injection surface and no risk of fast-forwarding `main` to an unmerged lineage.
- `create_tag` (staging) and the `create_release=false` path (production) correctly gate tag creation, the build matrix, Docker, Sentry, and the failure-cleanup jobs, so version-bump-only runs do exactly that and nothing downstream fires.
- Switching to `git push origin HEAD:main` is the right call now that prepare-build can run from a detached commit.

The `staging_tag` + `create_release=false` fallback (`SHA="${PROMOTE_SHA:-$STAGING_SHA}"`) resolves correctly since `staging_sha` is a real output of the resolve step. Confirmed there are no leftover references to `pretest-tests` or `skip_pretests` anywhere in the workflow tree.

CI is fully green and the change is CI-only with no application surface. Approving.

Issue #3169, opened 2026-06-01
Feature Request: Bidirectional Obsidian Vault Sync via File Watcher


rsnobles:
## Summary

The memory tree pipeline successfully writes summary `.md` files to `<workspace>/memory_tree/content/wiki/`, and a symlink can expose these in the user's Obsidian vault. However, there is no mechanism to watch the vault for user-created or edited notes and ingest them back into the memory tree. This completes the bidirectional sync loop described in the architecture docs.

## Background

Karpathy's LLM Wiki model treats the wiki as read-only raw data (written by the LLM) with the human browsing in Obsidian. OpenHuman's architecture goes further — the vault is meant to be bidirectional: the agent writes summaries, the user writes notes, and both sides stay in sync.

Currently:
- Left side (services → memory tree) is fully built via the Composio periodic sync pipeline
- Right side (memory tree ↔ Obsidian vault) is missing the reverse direction

### 1. Add `notify` crate dependency

Add the `notify` crate (or `notify-debouncer-mini`) to `Cargo.toml` for cross-platform filesystem event watching (FSEvents on macOS, inotify on Linux, ReadDirectoryChanges on Windows).

### 2. Create `memory_sync/workspace/folder/` module

The `memory_sync/workspace/` scaffold already exists but is empty. Implement a `SyncPipeline` for folder watching that:

- Watches the vault's `wiki/notes/` directory (or a configurable path) for `Create`, `Modify`, and `Remove` events
- On `Create`/`Modify`: reads the file, calls `ingest_document_with_scope()` to push through the canonicalize → chunk → score → persist pipeline
- On `Remove`: marks the document as deleted in the memory tree

### 3. Handle the dedup issue

`ingest_document_with_scope()` deduplicates by `source_id`. A re-ingest of an edited file with the same `source_id` is a no-op. The watcher needs to either:
- Append a content hash or mtime to the `source_id` so edits produce a new id
- Or add a `force_reingest` parameter to bypass the dedup gate

### 4. Track file modification times

Maintain an in-memory or SQLite-backed map of `path → last_mtime` so the watcher can detect actual changes and avoid re-ingesting unchanged files on startup.

### 5. Register in the startup path

Follow the Composio periodic sync pattern (`start_periodic_sync()` in `composio/periodic.rs`):
- Singleton `OnceLock` guard
- `tokio::spawn` the watcher loop
- Respect the scheduler gate policy (pause when memory tree is disabled)
[...]

PR #3183, opened 2026-06-02
Add vault file-system watcher implementation

manucian-official:
for issue #3169 

## Release Notes

* **New Features**
  * Added automatic monitoring of Obsidian vault notes folder for changes to markdown and text files
  * File changes are automatically ingested into the memory system
  * Efficiently handles rapid filesystem modifications through event debouncing

### DIFF
--- a/src/openhuman/memory_sync/workspace/watcher.rs
+++ b/src/openhuman/memory_sync/workspace/watcher.rs
@@ -1,0 +1,503 @@
+//! Vault file-system watcher.
+//!
+//! Watches a configurable directory (default: the Obsidian vault's
+//! `wiki/notes/` folder) for `Create`, `Modify`, and `Remove` events
+//! and ingests changes into the memory tree in near-real time.
+//!
+//! ## Design
+//!
+//! ```text
+//! ┌─────────────────────────────────────────────────────────┐
+//! │  notify (OS-native watcher)                             │
+//! │  FSEvents / inotify / ReadDirectoryChanges              │
+//! └────────────────────┬────────────────────────────────────┘
+//!                      │  raw events (debounced, 500 ms)
+//!                      ▼
+//! ┌─────────────────────────────────────────────────────────┐
+//! │  run_loop()  ← tokio task (singleton via OnceLock)      │
+//! │                                                         │
+//! │  ① scheduler-gate check  (UserDisabled / SignedOut)     │
+//! │  ② mtime guard           (SQLite WatcherStateStore)     │
+//! │  ③ path→source_id build  (stable base + mtime suffix)  │
+//! │  ④ ingest_document_with_scope()  or  mark_deleted()     │
+//! └─────────────────────────────────────────────────────────┘
+//! ```
+//!
+//! ## Dedup strategy (mtime-based source_id)
+//!
+//! `ingest_document_with_scope` deduplicates on `source_id`.  A plain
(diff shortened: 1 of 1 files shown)


src/openhuman/memory_sync/workspace/watcher.rs
```diff
@@ -0,0 +1,716 @@
...
+
+async fn handle_upsert(
+    path: &Path,
+    vault_root: &Path,
+    config: &Config,
+    state_store: Arc<Mutex<WatcherStateStore>>,
+    mtime_cache: Arc<Mutex<HashMap<PathBuf, u64>>>,
+) {
+    // ── mtime guard: skip if file unchanged since last ingest ────────────
+    let mtime = match file_mtime(path) {
```
graycyrus, line 260: **[major]** `file_mtime` calls `std::fs::metadata` synchronously here — blocking the tokio executor thread for every file event. Use `tokio::fs::metadata(path).await` and make `file_mtime` async.

src/openhuman/memory_sync/workspace/watcher.rs
```diff
@@ -0,0 +1,716 @@
...
+                path = %path.display(),
+                "[vault_watcher] mtime unchanged, skipping"
+            );
+            return;
+        }
+    }
+
+    // ── read file content ────────────────────────────────────────────────
+    let body = match tokio::fs::read_to_string(path).await {
+        Ok(b) => b,
```
graycyrus, line 284: **[major]** No size guard before reading the entire file into memory. A 20 MB note with embedded base64 assets will happily get loaded here. Check `metadata.len()` first and skip anything above a reasonable threshold (2 MB is a sane default).

graycyrus (changes requested):
Good structure and clean design overall — the mtime-scoped source_id strategy for bypassing the dedup gate is the right call, and the state store persistence across restarts is well thought out. Two issues need fixing before this can merge.

**Blocking I/O in async context**

`file_mtime` calls `std::fs::metadata` synchronously inside `handle_upsert`, which is an async function on a tokio executor thread. Under heavy vault activity (bulk saves, sync tools, git operations on the vault) this blocks the thread. Make `file_mtime` async:

```rust
async fn file_mtime(path: &Path) -> Option<u64> {
    tokio::fs::metadata(path)
        .await
        .ok()?
        .modified()
        .ok()?
        .duration_since(UNIX_EPOCH)
        .ok()
        .map(|d| d.as_secs())
}
```

Update the call site in `handle_upsert` to `file_mtime(path).await`.

**No file size guard before `read_to_string`**

`tokio::fs::read_to_string` has no upper bound. Obsidian vaults regularly contain files with embedded base64 assets, large data exports, or AI-generated note dumps that can run into tens of MB. The whole thing gets read into memory before the ingest pipeline can do anything about it. Check the file size upfront and skip oversized files:

```rust
// after resolving mtime, reuse the metadata call:
let metadata = tokio::fs::metadata(path).await.ok()?;
const MAX_FILE_BYTES: u64 = 2 * 1024 * 1024; // 2 MB
if metadata.len() > MAX_FILE_BYTES {
    tracing::warn!(
        path = %path.display(),
        size = metadata.len(),
        "[vault_watcher] file too large, skipping"
    );
    return;
}
```

This also lets you get mtime from `metadata.modified()` directly, consolidating the two `fs::metadata` calls into one.

senamakel (approved):

PR #3191, opened 2026-06-02
fix(scheduler-gate): build fresh System per CPU sample to avoid sysinfo OOB panic

oxoxDev:
## Summary

- The background scheduler-gate CPU sampler panicked with `index out of bounds: the len is 4 but the index is 4` on every 30s tick on at least one production Linux host, generating **10802** Sentry events (`CORE-RUST-ED`, release `openhuman@0.56.0`).
- Root cause is a sysinfo 0.33 Linux quirk triggered by a long-lived `System` whose per-core view was captured once at process boot.
- Fix: build a fresh `System` per `sample_cpu()` call so the per-core read always matches the host's current core count.

## Problem

`signals.rs` held a process-wide `static CPU_SYS: Lazy<Mutex<System>>` built once at boot. sysinfo 0.33's `CpusWrapper::refresh` (`unix/linux/cpu.rs`) builds its per-core `Vec` from the `cpuN` lines of `/proc/stat` on the **first** refresh, then on every **later** refresh indexes that `Vec` by line position (`self.cpus[i]`). There is no public sysinfo API to refresh only the global aggregate — `refresh_cpu_usage()` always walks the per-core path.

When a host's visible core count **grows** after boot — CPU hotplug, or a Proxmox / cloud VM re-balancing vCPUs at runtime (the affected host runs kernel `6.17.2-pve`) — `/proc/stat` gains a new `cpuN` line while the boot-captured `Vec` is still the old length, so the index runs off the end and panics. Because the boot-time count never updates, **every** subsequent 30s sample panicked. `spawn_blocking` in `gate.rs` caught the unwind so the process survived, but the sentry panic hook fired each tick and the CPU-pressure signal fell back to its neutral default — i.e. the throttle gate was effectively blind on that host.

## Solution

- Drop the `static CPU_SYS` (and the now-unused `Lazy` / `Mutex` imports); construct `System::new()` locally inside `sample_cpu()`.
- The existing two-refresh delta still runs on that fresh instance: the first refresh builds the per-core `Vec` at the **current** core count, the second (≈`MINIMUM_CPU_UPDATE_INTERVAL` later) reads against the same current count, so the index stays in bounds. We only ever consume `global_cpu_usage()`, so not retaining per-core state across calls is free.
- This **restores** a working CPU signal on drifting hosts rather than suppressing the report. The residual exposure shrinks from "core count changed any time since boot" (permanent) to "core count changes within the ~250 ms intra-call window" (transient/rare).
- Added unit tests asserting `sample_cpu()` / `Signals::sample()` stay finite, in-range, and panic-free across repeated calls.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines are the rewritten `sample_cpu()` body + a new `#[cfg(test)]` module exercising it; `cargo test --lib openhuman::scheduler_gate` green (30 passed)
- [x] N/A: behaviour-only fix, no feature row change — Coverage matrix not updated
- [x] N/A: no matrix feature touched — no feature IDs to list under Related
- [x] N/A: internal background sampler, not a release-cut surface — manual smoke checklist not updated
- [x] N/A: Sentry-only, no GitHub issue — see Related; resolved via `Sentry-Issue`

## Impact

- **Platform**: Linux core binary (`openhuman-core`) on hosts with runtime-variable vCPU counts (Proxmox / cloud VMs, CPU hotplug). No behavior change on fixed-core hosts.
- **Performance**: one extra empty `System::new()` + one extra `/proc/stat` parse per sample, on a 30s cadence — negligible.
- **Security / migration**: none.

## Related

- Closes: _Sentry-only — no GitHub issue_
- Sentry-Issue: CORE-RUST-ED
- Follow-up PR(s)/TODOs: evaluate bumping `sysinfo` past 0.33 once the upstream per-core index fix is confirmed (deferred here to keep blast radius small).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/core-rust-ed-cpu-sampler-panic
- Commit SHA: 8a4babefaab6deac8fb6fa79aaf127d79e5667bf

### Validation Run

- [x] N/A: no app/ changes — `pnpm --filter openhuman-app format:check`
- [x] N/A: no app/ changes — `pnpm typecheck`
- [x] Focused tests: `cargo test --lib openhuman::scheduler_gate` → 30 passed
- [x] Rust fmt/check: `cargo fmt` + `cargo check --manifest-path Cargo.toml` clean
- [x] N/A: core crate only — Tauri fmt/check

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: CPU sampler no longer panics when the host core count grows at runtime; CPU-pressure signal works on drifting hosts.
- User-visible effect: none directly (internal background throttle signal); eliminates a high-volume crash-report stream.

### DIFF
--- a/src/openhuman/scheduler_gate/signals.rs
+++ b/src/openhuman/scheduler_gate/signals.rs
@@ -4,10 +4,8 @@
 //! file just captures one snapshot at a time.
 
 use std::path::Path;
-use std::sync::Mutex;
 use std::time::Duration;
 
-use once_cell::sync::Lazy;
 use sysinfo::System;
 
 #[derive(Debug, Clone, Copy)]
@@ -107,16 +105,24 @@ fn probe_battery() -> Result<BatteryProbe, starship_battery::Error> {
 
 // ---- cpu -----------------------------------------------------------------
 
-static CPU_SYS: Lazy<Mutex<System>> = Lazy::new(|| Mutex::new(System::new()));
-
 fn sample_cpu() -> f32 {
-    // Two refreshes spaced ~MINIMUM_CPU_UPDATE_INTERVAL apart give sysinfo
-    // a real delta to compute usage from. The interval is small enough to
-    // run on the 30s sampler tick without noticeable cost.
-    let mut sys = match CPU_SYS.lock() {
-        Ok(g) => g,
-        Err(p) => p.into_inner(),
-    };
+    // Build a *fresh* `System` every sample instead of reusing a long-lived
+    // one. sysinfo 0.33's Linux CPU refresh builds a per-core Vec on its first
+    // refresh (sized to the `cpuN` lines in /proc/stat) and then, on every
+    // later refresh, indexes that Vec by line position. If the visible core
+    // count later grows — CPU hotplug, or a Proxmox / cloud host re-balancing
+    // vCPUs at runtime — the next refresh indexes past the Vec and panics
+    // ("index out of bounds: the len is N but the index is N"). A process-wide
+    // System captured the boot-time core count, so on such hosts *every* 30s
+    // tick panicked thereafter (Sentry CORE-RUST-ED). Building per call means
+    // both refreshes below always see the current core count, so the index
+    // stays in bounds.
+    //
+    // Two refreshes spaced ~MINIMUM_CPU_UPDATE_INTERVAL apart give sysinfo a
+    // real delta to compute usage from; we only read the global aggregate, so
+    // not retaining per-core state across calls costs us nothing. The interval
+    // is small enough to run on the 30s sampler tick without noticeable cost.
+    let mut sys = System::new();
     sys.refresh_cpu_usage();
     std::thread::sleep(Duration::from_millis(
         sysinfo::MINIMUM_CPU_UPDATE_INTERVAL.as_millis() as u64 + 50,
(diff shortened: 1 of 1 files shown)


graycyrus (approved):
Good fix for a real production crash. The root cause diagnosis is solid — sysinfo 0.33's Linux CPU refresh builds a per-core Vec at first refresh, then indexes into it by line position on every later refresh. When a Proxmox host re-balances vCPUs at runtime the Vec is stale and the index runs off the end. Every subsequent 30s tick panicked. 10802 Sentry events in a row is a memorable way to discover a library footgun.

The fix is correct. A fresh System per sample_cpu() call means both refreshes always see the current core count. Since we only read global_cpu_usage() and don't carry per-core state between 30s ticks, nothing is lost by dropping the singleton. Removing Lazy<Mutex<System>> also eliminates the poisoned-mutex error path, which is a nice cleanup.

Tests are meaningful: sample_cpu_is_finite_and_bounded and sample_cpu_repeatable directly guard the regression, and signals_sample_smoke gives an end-to-end pass through Signals::sample(). Not coverage padding.

One minor note: the block comment inside sample_cpu runs about 15 lines and largely restates the PR description. The sysinfo footgun is genuinely non-obvious so a comment is warranted, but the Sentry reference plus two sentences on the core invariant would be enough — the rest is noise. Worth trimming in a follow-up, not worth holding the fix for.

On the AI slop front: the PR description and comment blocks show generation signals (very uniform structure, verbose beyond the scope of the change). The code itself is minimal and correct, so I'm not blocking on it, but flagging so the pattern is on record.

CI all green. Merging.

PR #3192, opened 2026-06-02
fix: disable storage for Codex responses requests

honor2030:
## Summary
- Send `store: false` on OpenAI/Codex Responses API requests.
- Extend the existing Codex Responses API test to assert the outbound JSON body includes `store: false`.

## Why
Codex OAuth routes through the ChatGPT backend Responses API. That path expects requests not to create stored responses, so explicitly setting `store: false` keeps the request shape aligned with Codex/ChatGPT account usage.

## Test Plan
- RED: strengthened `responses_api_primary_posts_directly_to_responses` to require `store: false`; before the implementation change, the mock did not match and the focused test failed with `openai Responses API error`.
- GREEN: `RUSTC="$(rustup which --toolchain 1.93.0 rustc)" rustup run 1.93.0 cargo test responses_api_primary_posts_directly_to_responses --manifest-path Cargo.toml --lib -- --nocapture` → 1 passed, 0 failed.
- `RUSTC="$(rustup which --toolchain 1.93.0 rustc)" rustup run 1.93.0 cargo test openai_oauth --manifest-path Cargo.toml --lib -- --nocapture` → 32 passed, 0 failed.
- `rustup run 1.93.0 cargo fmt --check`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Extended API request configuration capabilities to support explicit control of server-side storage behavior for OpenAI-compatible provider endpoints, enabling management of data retention preferences

* **Tests**
  * Updated test suite to validate that storage configuration settings are correctly included and serialized in outbound request payloads

### DIFF
--- a/src/openhuman/inference/provider/compatible.rs
+++ b/src/openhuman/inference/provider/compatible.rs
@@ -269,6 +269,7 @@ impl OpenAiCompatibleProvider {
             input,
             instructions,
             stream: Some(false),
+            store: Some(false),
         };
 
         let url = self.responses_url();
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -1,7 +1,7 @@
 use super::*;
 use sentry::test::TestTransport;
 use std::sync::Arc;
-use wiremock::matchers::{method, path};
+use wiremock::matchers::{body_json, method, path};
 use wiremock::{Mock, MockServer, ResponseTemplate};
 
 fn make_provider(name: &str, url: &str, key: Option<&str>) -> OpenAiCompatibleProvider {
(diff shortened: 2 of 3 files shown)


graycyrus (dismissed):
Clean, well-scoped fix. Adding `store: Option<bool>` to `ResponsesRequest` with `skip_serializing_if = "Option::is_none"` is the right approach — it keeps the field absent for providers that don't use it while letting the Codex path set it explicitly to `false`.

The test tightening is solid: switching to `body_json` makes the mock an exact-match assertion on the full outbound payload, so any future regression on `store` will fail the test rather than silently slip through. The matcher aligns with the actual serialized body (instructions is None and thus skipped, stream and store are both present as false).

CI is fully green including the coverage gate. No concerns.

graycyrus (commented):
Clean, well-scoped fix. Adding `store: Option<bool>` to `ResponsesRequest` with `skip_serializing_if = "Option::is_none"` is the right approach — it keeps the field absent for providers that don't use it while letting the Codex path set it explicitly to `false`.

The test tightening is solid: switching to `body_json` makes the mock an exact-match assertion on the full outbound payload, so any future regression on `store` will fail the test rather than silently slip through. The matcher aligns with the actual serialized body (instructions is None and thus skipped, stream and store are both present as false).

CI is fully green including the coverage gate. No concerns.

PR #3210, opened 2026-06-02
fix(shell): Windows pre-CEF cache-lock wait to stop relaunch panic (TAURI-RUST-F)

oxoxDev:
## Summary

- Add a Windows pre-CEF **cache-lock wait** so a relaunch never calls `cef::initialize()` while a dying prior instance still holds the CEF user-data-dir.
- Fixes a fatal Windows-only startup panic — `assertion left == right failed, left: 0, right: 1` — raised when the vendored `tauri-runtime-cef` asserts `cef::initialize() == 1` and gets `0` on a locked cache (Sentry **TAURI-RUST-F**, ~3.2k events, live on 0.57.5).
- New `cef_singleton_wait` module: bounded (5 s, exponential backoff) wait for prior-instance processes to exit, then proceed; clean exit if still held. Windows analog of the existing macOS `reap_stale_openhuman_processes` guard.

## Problem

The existing pre-CEF Win32 named-mutex guard (`run()` in `lib.rs`) stops a **concurrent** second launch. It does **not** cover the **sequential relaunch race**:

1. Instance A's `run()` returns → the RAII mutex guard drops → the named mutex is released.
2. A is still tearing down — its CEF browser process keeps the cache lock for a short window.
3. Instance B launches (auto-update relaunch, fast quit+reopen, a restart flow), acquires the now-free mutex, and calls `cef::initialize()` while A's cache lock is live → returns `0` → the vendored runtime's `assert_eq!(result, 1)` panics.

`prepare_process_cache_path()` creates the per-user cache dir but never checks whether it is actively locked, and Chromium's `SingletonLock` symlink that `cef_preflight` reads on macOS/Linux does not exist on Windows — so Windows had no cache-lock guard beyond the lifetime-mismatched mutex.

## Solution

- New `app/src-tauri/src/cef_singleton_wait.rs`:
  - Pure `decide(other_instances, elapsed, budget)` + `backoff_delay(attempt)` — unit-tested on any host.
  - `#[cfg(windows)] win::count_other_app_instances()` — Toolhelp32 snapshot counting processes with our exe basename, excluding self.
  - `#[cfg(windows)] wait_for_cache_release()` — poll/backoff up to 5 s; `Proceed` when count is 0, else `KeepWaiting`, else `GiveUp` → forward deep links + `exit(0)`.
- Wired in `lib.rs` after the Win32 mutex + `prepare_process_cache_path`, before the Tauri builder — beside the macOS reap.
- **Why counting same-exe PIDs is correct despite CEF spawning same-exe subprocesses:** this runs *before* `cef::initialize()`, so our own CEF subprocesses don't exist yet, and it runs *after* the mutex guard, which guarantees we're the only top-level instance — so any other same-exe PIDs are the dying prior instance's stragglers, dropping to 0 once it exits.
- **This prevents the panic, it does not suppress it:** `cef::initialize()` is simply never called against a live lock. No `catch_unwind`, no assert masking.
- Follow-up (not in this PR): the true root fix is making `tauri-runtime-cef` return an `Err` instead of asserting on `cef::initialize() == 0`; tracked as a vendored-CEF change.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — the decision logic (`decide`, `backoff_delay`) is covered by 5 host unit tests (proceed / keep-waiting / give-up boundary / backoff cap / overflow). The `#[cfg(windows)]` Toolhelp enumeration + wait loop are not instrumentable on the Linux coverage runner (compiled out off-Windows); validated by the Windows CI build + manual repro instead.
- [x] N/A: behaviour-only platform-guard fix — no added/removed/renamed coverage-matrix feature row.
- [x] N/A: no coverage-matrix feature IDs affected.
- [x] N/A: no release-cut smoke surface changed — startup guard only; no user-facing flow.
- [x] N/A: Sentry-only issue, no GitHub issue to close — `Sentry-Issue: TAURI-RUST-F` (see Related).

## Impact

- **Windows desktop only.** Zero cost in the common case (no prior instance → proceed immediately); a ≤5 s bounded wait only inside the rare relaunch-race window, then proceed or clean-exit.
- Removes a fatal startup crash; no behaviour change on macOS/Linux (module is `cfg(windows)`; pure logic also compiled under `test`).
- **Validation note:** the bug is a Windows-only CEF init race that cannot be reproduced on a macOS dev host. Locally verified: `cargo fmt`, Tauri-shell `cargo check`/`clippy`/`test` (5 new tests pass). The `cfg(windows)` glue compiles via the Windows CI job; the actual race is being reproduced/smoked on real Windows by @sanil-23.

## Related

- Closes: _N/A — Sentry-only_
- `Sentry-Issue: TAURI-RUST-F`
- Follow-up: vendored `tauri-runtime-cef` `cef::initialize()` assert → `Err` (true root fix; broader blast radius).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows startup stability by implementing a resource-release wait mechanism during application launch. The application now waits for prior instances to fully release critical resources before proceeding with initialization, preventing crashes that could occur during rapid sequential relaunches or repeated quick restarts.

### DIFF
--- a/app/src-tauri/Cargo.toml
+++ b/app/src-tauri/Cargo.toml
@@ -169,6 +169,10 @@ windows-sys = { version = "0.59", features = [
     # parameter is gated behind it in windows-sys 0.59.
     "Win32_System_Threading",
     "Win32_Security",
+    # CreateToolhelp32Snapshot / Process32FirstW / Process32NextW — used by the
+    # pre-CEF cache-lock wait (cef_singleton_wait.rs) that counts straggler
+    # processes from a dying prior instance before cef::initialize (TAURI-RUST-F).
+    "Win32_System_Diagnostics_ToolHelp",
     "Win32_Storage_FileSystem",
     "Win32_System_IO",
     "Win32_System_Pipes",
--- a/app/src-tauri/src/cef_singleton_wait.rs
+++ b/app/src-tauri/src/cef_singleton_wait.rs
@@ -1,0 +1,283 @@
+//! Windows pre-CEF cache-lock wait (Sentry TAURI-RUST-F).
+//!
+//! The vendored `tauri-runtime-cef` calls `cef::initialize()` and asserts the
+//! result equals `1`. When the CEF user-data-dir is still in use by another
+//! OpenHuman process, `cef::initialize()` returns `0` and that assert panics
+//! with `assertion left == right failed, left: 0, right: 1` — a fatal,
+//! actionless crash (Sentry TAURI-RUST-F, ~3.2k events, Windows-only).
+//!
+//! The existing pre-CEF Win32 named-mutex guard (see `run()` in `lib.rs`)
+//! already handles the *concurrent* second-launch case: a second instance
+//! sees `ERROR_ALREADY_EXISTS` and exits before touching CEF. What it does
+//! NOT cover is the *sequential relaunch race*:
+//!
+//!   1. Instance A's `run()` returns → the RAII mutex guard drops → the named
+//!      mutex is released.
+//!   2. A is still tearing down: its CEF browser process (and child
+//!      processes) keep the user-data-dir cache lock for a short window.
+//!   3. Instance B launches (auto-update relaunch, fast quit+reopen, a
+//!      restart flow), acquires the now-free mutex, and calls
+//!      `cef::initialize()` while A's cache lock is still held → `0` → panic.
+//!
+//! This module closes that window the same way macOS does with
+//! `process_recovery::reap_stale_openhuman_processes`: before CEF init, wait
+//! (bounded) for the prior instance's process(es) to exit, then proceed. If
+//! the holder is still alive after the budget, exit cleanly instead of
+//! initializing into a locked cache — the panic is *prevented* (we never call
+//! `cef::initialize()` against a live lock), not caught/suppressed.
+//!
+//! Why counting same-exe processes is correct here, despite CEF spawning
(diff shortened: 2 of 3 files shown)


oxoxDev:
@sanil-23 could you repro on real Windows once? This is a Windows-only CEF init race I can't reproduce on macOS. Repro: launch OpenHuman, then trigger a fast relaunch — auto-update relaunch, or quit (Cmd/Alt+F4) and immediately reopen, or a restart flow — while the prior instance is still tearing down. Pre-fix that hits the `cef::initialize` assert panic (left:0 right:1, TAURI-RUST-F); with this PR the new instance should wait briefly then open (logs: `[cef-singleton-wait] ...`). Thanks!

oxoxDev:
Correction to the PR body: there is **no Windows-target CI job** in this repo, so the `#[cfg(windows)]` code in `cef_singleton_wait.rs` (the Toolhelp32 enumeration + wait loop) is **not compiled by CI** — Linux/macOS runners compile it out. The pure `decide`/`backoff_delay` logic is host-tested + in CI; the Windows glue's compilation rests on a local Windows build.

@sanil-23 — two asks when you get a Windows box: (1) confirm it **compiles** (`pnpm rust:check` / `cargo build` for the shell on Windows), and (2) the runtime repro from the earlier comment. (1) is the more important gate since CI can't catch a Windows compile error here.

senamakel (approved):

PR #3211, opened 2026-06-02
fix(inference): actionable error for completion-only model 404 + unmistakable run_code failure (#3193)

### LINKED ISSUE #3193: Cannot write code on filesystem
# `run_code` returns 404 from OpenAI chat completions endpoint, orchestrator silently falls back to hallucinated success

## Environment
- `openhuman-core 0.56.0`, Windows 11, native runtime
- `agentic_provider = "openhuman"`
- `runtime_python.enabled = true`, `node.enabled = true`

## Bug
**Any call to `run_code` fails with:
openai API error (404 Not Found): "This is not a chat model and thus not supported
in the v1/chat/completions endpoint. Did you mean to use v1/completions?"
(chat completions unavailable; responses fallback failed)**

Likely cause: the model assigned to `run_code` is a completion-only model but is being
routed to `/v1/chat/completions`. Either the model assignment or the endpoint is wrong.

## Worst part
When this fails, the orchestrator marks the tool result as `success: true` and returns
a plausible-looking output as if the code had run, without actually executing anything.
No file is written. The `CICLO` budget indicator stays at `0%` in these cases,
suggesting the backend knows execution didn't happen but doesn't surface it.

## Related observations (possibly same root cause)
1. `delegate_tools_agent` consistently returns `400 Insufficient budget` even on no-op calls.
2. `code_executor` via `spawn_parallel_agents` is non-deterministic: same prompt writes
   to disk one run, returns "I work in a sandbox, the file was not written" the next.
3. Only the first task in `spawn_parallel_agents` ever produces real side effects,
   even with `parallel_tools = true`.

## Questions
- Is the `run_code` 404 a known model→endpoint misconfiguration? Any user-side override?
- Is the `code_executor` real-exec vs. simulated behavior intentional? Could the tool
  result expose an `executed: true|false` flag?
- Is `parallel_tools = true` actually respected, or is there a single runtime slot per turn?

Happy to share full transcripts and screenshots.

oxoxDev:
## Summary

- `run_code` (and any chat call) against a **completion-only / base model** now fails fast with an actionable error that names the model and the fix, instead of an opaque `chat completions unavailable; responses fallback failed` chain.
- A subagent-delegation failure is now wrapped in an **unmistakable failure envelope** so a weak orchestrator can't narrate a fabricated success ("wrote the file") when nothing actually ran.
- Completion-only detection is shared across all three chat entrypoints (`chat_with_system`, `chat_with_history`, native tool-calling `chat`).

## Problem

Reported in #3193 (openhuman-core 0.56.0, Win11, `agentic_provider = "openhuman"`): every `run_code` call returns

> `openai API error (404 Not Found): "This is not a chat model and thus not supported in the v1/chat/completions endpoint. Did you mean to use v1/completions?" (chat completions unavailable; responses fallback failed)`

Root cause: OpenHuman only speaks the chat-completions API (with an optional `/v1/responses` fallback) — there is no `/v1/completions` path. When the model bound to the **coding** role (`run_code` → `code_executor`, model hint `coding`) is a completion-only model, `/v1/chat/completions` 404s and the responses fallback can't rescue it. The surfaced error was opaque and gave no remediation. The reporter also observed the orchestrator presenting the failure as success and fabricating output — the bare error text alone didn't stop a weak model from narrating a fake result.

## Solution

- **`compatible.rs`**: add `is_completion_only_model_404` (tight match on the OpenAI signature — ordinary "model does not exist" 404s are intentionally **not** matched) and `completion_only_model_message` (names the model + "assign a chat-capable model"). A shared `completion_only_404_guard` is invoked at all three chat 404 handlers and **short-circuits before the doomed `/v1/responses` fallback**.
- **`dispatch.rs`**: route subagent-failure results through `format_subagent_failure`, which states the task did not run and instructs the model not to report success or fabricate output, while preserving the underlying error.
- Deliberately **not** adding a legacy `/v1/completions` transport — the real cause is a misconfigured model, so the fix is an actionable diagnostic, not a new (deprecated) endpoint surface.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — Coverage Gate (diff-cover ≥ 80%) passes in CI; focused Rust tests (`completion_only*`, `subagent_failure_envelope`) pass locally.
- [x] Coverage matrix updated — `N/A: behaviour-only change` (error-handling/diagnostic; no feature row added/removed/renamed)
- [x] No new external network dependencies introduced (wiremock mock server used for the 404 path)

## Impact

- Desktop/CLI (Rust core). No schema, migration, or protocol change.
- Behaviour change is limited to error paths: a completion-only model now yields a clear, fixable message and stops before a futile fallback; failed delegations are reported as failures instead of being mistakable for success. Successful calls are unaffected.

## Related

- Closes: #3193
- Follow-up PR(s)/TODOs: deferred sub-symptoms from #3193 needing the reporter's transcript — `delegate_tools_agent` "400 Insufficient budget", `spawn_parallel_agents` non-determinism, and whether `parallel_tools` is honored. Confirmation of a literal `success:true` flag bug (vs. orchestrator narration) also pending the transcript.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3193-run-code-completion-only-404`
- Commit SHA: `d555ae48472ce694570791ccbe7d99b742f9a3c4`

### Validation Run

- [x] N/A: no `app/` changes — `pnpm --filter openhuman-app format:check`
- [x] N/A: no `app/` changes — `pnpm typecheck`
- [x] Focused tests: `cargo test --lib completion_only` (6 passed) · `cargo test --lib subagent_failure_envelope` (1 passed)
- [x] Rust fmt/check (if changed): `cargo fmt` applied; `cargo check --lib` clean
- [x] N/A: no Tauri-shell changes — Tauri fmt/check

### Validation Blocked

- `command:` `pnpm test:coverage` / `pnpm test:rust` (full suites)
- `error:` not run locally (heavy; full Rust test matrix OOMs locally)
- `impact:` diff-coverage verified by CI gate (passing)

### Behavior Changes

- Intended behavior change: completion-only-model 404 → actionable fail-fast error; failed subagent delegation → unmistakable failure result.
- User-visible effect: clear "assign a chat-capable model" guidance instead of an opaque fallback chain; agent no longer reports fabricated success after a hard failure.

### Parity Contract

- Legacy behavior preserved: ordinary 404s keep their existing fallback/enrich path; non-404 errors still flow through `api_error` (Sentry/classification) unchanged; successful responses unchanged.
- Guard/fallback/dispatch parity checks: completion-only guard added uniformly to all three chat entrypoints; `/v1/responses` fallback still runs for non-completion-only 404s.

- Duplicate PR(s): none found (no open PR touches `run_code` / completion endpoint)
- Canonical PR: this PR
- Resolution: N/A

### DIFF
--- a/src/openhuman/agent_orchestration/tools/dispatch.rs
+++ b/src/openhuman/agent_orchestration/tools/dispatch.rs
@@ -119,9 +119,28 @@ pub(crate) async fn dispatch_subagent(
                 agent_id: definition.id.clone(),
                 error: message.clone(),
             });
-            Ok(ToolResult::error(format!("{tool_name} failed: {message}")))
+            // Make the failure unmistakable to the orchestrator: the delegated
+            // task did NOT run, so it must not be reported as success or have
+            // its output fabricated. Without this guardrail a weak orchestrator
+            // can narrate a plausible success from the bare error text — the
+            // "hallucinated success" half of #3193 (e.g. claiming `run_code`
+            // wrote a file when the coding model 404'd and nothing executed).
+            Ok(ToolResult::error(format_subagent_failure(
+                tool_name, &message,
+            )))
         }
     }
+}
+
+/// Format a subagent-delegation failure so the orchestrator cannot mistake it
+/// for success. Kept as a standalone, side-effect-free fn so the exact wording
+/// is unit-testable without standing up a registry + failing model (#3193).
+fn format_subagent_failure(tool_name: &str, message: &str) -> String {
+    format!(
+        "{tool_name} failed and did not complete — no work was performed and no \
+         results were produced. Do NOT treat this as success or fabricate an \
+         output; report the failure to the user. Error: {message}"
+    )
 }
 
 #[cfg(test)]
@@ -160,6 +179,33 @@ mod tests {
         assert!(
             out.contains("registry not initialised") || out.contains("not found in registry"),
             "unexpected graceful-failure message: {out}"
+        );
+    }
+
+    #[test]
+    fn subagent_failure_envelope_forbids_fabricated_success() {
+        // #3193: a hard delegation failure (e.g. run_code's coding model
+        // 404ing) must be surfaced so the orchestrator cannot narrate a
+        // plausible success. The envelope states the task did not run, tells
+        // the model not to fabricate output, and preserves the root error.
+        let msg = format_subagent_failure(
+            "run_code",
+            "openhuman API error (404): model 'davinci-002' does not support \
+             the chat-completions API",
+        );
+        assert!(msg.contains("run_code failed"), "names the tool: {msg}");
+        assert!(
+            msg.contains("did not complete"),
+            "states no completion: {msg}"
+        );
+        assert!(
+            msg.to_lowercase().contains("do not treat this as success")
+                && msg.contains("fabricate"),
+            "warns against fabricated success: {msg}"
+        );
+        assert!(
+            msg.contains("davinci-002") && msg.contains("404"),
+            "preserves the root error: {msg}"
         );
     }
 }
--- a/src/openhuman/inference/provider/compatible.rs
+++ b/src/openhuman/inference/provider/compatible.rs
@@ -146,6 +146,26 @@ impl OpenAiCompatibleProvider {
         )
     }
 
+    /// Guard shared by every chat-completions 404 handler: if the body shows a
+    /// completion-only model, return the actionable error so the caller can
+    /// fail fast instead of attempting the futile `/v1/responses` fallback.
+    /// `None` means "not this case — proceed with normal fallback/enrich".
+    /// See issue #3193.
+    fn completion_only_404_guard(
+        &self,
+        status: reqwest::StatusCode,
+        sanitized: &str,
+        model: &str,
+    ) -> Option<anyhow::Error> {
+        if Self::is_completion_only_model_404(status, sanitized) {
+            Some(anyhow::anyhow!(
+                self.completion_only_model_message(model, sanitized)
+            ))
+        } else {
+            None
+        }
+    }
+
     /// Create a provider with a custom User-Agent header.
     ///
     /// Some providers (for example Kimi Code) require a specific User-Agent
@@ -1380,10 +1400,8 @@ impl Provider for OpenAiCompatibleProvider {
 
             // A completion-only model 404s here and the /v1/responses fallback
             // cannot rescue it — fail fast with actionable guidance (#3193).
-            if Self::is_completion_only_model_404(status, &sanitized) {
-                return Err(anyhow::anyhow!(
-                    self.completion_only_model_message(model, &sanitized)
-                ));
+            if let Some(err) = self.completion_only_404_guard(status, &sanitized, model) {
+                return Err(err);
             }
 
             if status == reqwest::StatusCode::NOT_FOUND && self.supports_responses_fallback {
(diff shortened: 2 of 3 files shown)


CodeGhost21 (commented):
Inline findings posted on the changed lines. Overall: the core fix is correct, tightly scoped, and well-tested. Main follow-ups concern adjacent code paths (`chat_with_tools`, streaming) that share the same 404 surface but don't currently invoke the new guard.

src/openhuman/agent_orchestration/tools/dispatch.rs
```diff
@@ -119,11 +119,30 @@ pub(crate) async fn dispatch_subagent(
...
+                tool_name, &message,
+            )))
         }
     }
 }
 
+/// Format a subagent-delegation failure so the orchestrator cannot mistake it
+/// for success. Kept as a standalone, side-effect-free fn so the exact wording
+/// is unit-testable without standing up a registry + failing model (#3193).
+fn format_subagent_failure(tool_name: &str, message: &str) -> String {
```
CodeGhost21, line 138: **In-band prompt-engineering preamble now applied to every subagent failure.**

`format_subagent_failure` is invoked on the generic `run_subagent` error path — so timeouts, budget exhaustion, transport errors, and any other transient failure now also carry the ~200-char "do NOT treat this as success or fabricate an output" preamble in transcripts, logs, and observability. That's intentional given #3193's "hallucinated success" symptom, but it's a behavior change broader than the issue title suggests ("completion-only 404"). Worth either:

- calling out in the PR body's *Behavior Changes* section that the envelope is applied to **all** subagent failures, or
- gating the envelope to a narrower class of "hard, non-retryable" failures so retry/budget cases keep their shorter error string.

Not blocking — flagging so it's a deliberate decision.

src/openhuman/inference/provider/compatible.rs
```diff
@@ -798,6 +835,24 @@ impl OpenAiCompatibleProvider {
...
+    /// deliberately tight so ordinary "model does not exist" 404s are NOT
+    /// caught (those should keep their existing fallback / enrich behaviour).
+    /// See issue #3193.
+    fn is_completion_only_model_404(status: reqwest::StatusCode, error: &str) -> bool {
+        if status != reqwest::StatusCode::NOT_FOUND {
+            return false;
+        }
+        let lower = error.to_lowercase();
+        lower.contains("not a chat model")
+            || (lower.contains("v1/chat/completions") && lower.contains("v1/completions"))
```
CodeGhost21, line 853: **Second disjunct is correct today but fragile to OpenAI rewording.**

The string `"v1/chat/completions"` does **not** contain `"v1/completions"` as a continuous substring (the `/chat/` infix breaks it), so the two `contains` calls really are independent matches — but that's only obvious if you stop and parse it. The whole clause currently fires only because OpenAI's body happens to mention both endpoint paths ("...the v1/chat/completions endpoint. Did you mean to use v1/completions?"). If OpenAI ever drops the second URL from the wording, this branch silently stops matching and we fall back to the opaque path.

Suggest a one-line comment on line 853 making the "two independent substrings" intent explicit, e.g.:

```rust
// Defensive fallback: OpenAI's current phrasing references BOTH endpoint paths
// as two separate substrings (`v1/chat/completions` does not contain `v1/completions`).
```

src/openhuman/inference/provider/compatible.rs
```diff
@@ -1510,18 +1571,38 @@ impl Provider for OpenAiCompatibleProvider {
...
+                }
+
+                if self.supports_responses_fallback {
+                    return self
+                        .chat_via_responses(credential, &effective_messages, model)
+                        .await
+                        .map_err(|responses_err| {
+                            let fb = super::format_anyhow_chain(&responses_err);
+                            anyhow::anyhow!(
+                                "{} API error ({status}): {sanitized} (chat completions unavailable; responses fallback failed: {fb})",
```
CodeGhost21, line 1595: **Undocumented behavior change on the non-completion-only fallback path.**

Before this PR, `chat_with_history`'s responses-fallback failure read:

```
{name} API error (chat completions unavailable; responses fallback failed: {fb})
```

After this PR it reads:

```
{name} API error ({status}): {sanitized} (chat completions unavailable; responses fallback failed: {fb})
```

That's a strict improvement (now includes the original 404 body and status), but it's a behavior change to a path *unrelated* to completion-only detection — any ordinary 404 that triggers the responses fallback will now have a different error string. The PR body's *Behavior Changes* section only lists the completion-only and subagent-envelope changes, not this. Suggest adding a one-liner there so downstream log parsers / Sentry classifiers aren't surprised.

(Cross-reference: the same widening was applied at the native-tool-calling `chat` 404 path — consistent with this one, which is good.)

src/openhuman/inference/provider/compatible.rs
```diff
@@ -1859,6 +1940,12 @@ impl Provider for OpenAiCompatibleProvider {
                 });
             }
 
+            // A completion-only model 404s here and the /v1/responses fallback
+            // cannot rescue it — fail fast with actionable guidance (#3193).
+            if let Some(err) = self.completion_only_404_guard(status, &sanitized, model) {
```
CodeGhost21, line 1945: **Adjacent code paths share the same 404 surface but skip the new guard.**

The PR body claims "completion-only detection is shared across all three chat entrypoints", but two more entrypoints on this same provider can hit the same OpenAI completion-only 404:

1. **`chat_with_tools` (line ~1697)** — `if !response.status().is_success() { return Err(super::api_error(&self.name, response).await); }`. No guard, no responses fallback. If a `run_code`-style flow ever passes through the native tool-calling non-streaming path (and tools is non-empty), the user gets the opaque `"not a chat model"` 404 again. Transport-error path *does* fall back to `chat_with_history` (which has the guard), but the **404-status** path does not.

2. **Streaming paths** — `stream_chat_with_system` (~line 2138) and `stream_chat_with_history` (~line 2298). Both already sanitize the error body, so adding a `completion_only_404_guard` branch is a one-liner that matches the pattern used here.

Proposal: either replicate the 3-line guard at all four additional sites (preferred — they're 1-liners and the helper is already shared), or document in the PR body why streaming + `chat_with_tools` are intentionally excluded. Otherwise #3193 can re-emerge through any of these paths.

PR #3218, opened 2026-06-02
docs(env): document OPENHUMAN_BRAVE_API_KEY + OPENHUMAN_PARALLEL_API_KEY in .env.example (#3098 sub-issue 4)

CodeGhost21:
## Summary

- `.env.example` now documents `OPENHUMAN_BRAVE_API_KEY` (and `OPENHUMAN_PARALLEL_API_KEY`, same root cause), matching the existing Querit / Seltz / SearXNG sections.
- Zero code change. The Brave Search code path is already fully wired — this PR closes the discoverability gap that most plausibly explains sub-issue 4 of #3098.

## Problem

Sub-issue 4 of #3098 reports "Brave Search not connecting — team noted this requires a Brave API key set in config or env vars." The Brave wiring is in fact complete end-to-end:

- **Config field**: `config.search.brave.api_key` (`src/openhuman/config/schema/types.rs`)
- **Env loader**: `OPENHUMAN_BRAVE_API_KEY` / `BRAVE_API_KEY` are read at `src/openhuman/config/schema/load.rs:1503-1507`
- **RPC**: `openhuman.config.update_search_settings { brave_api_key }` (`src/openhuman/config/ops.rs:1237-1241`)
- **UI**: `app/src/components/settings/panels/SearchPanel.tsx` has the Brave field + "configured" badge
- **Engine gate**: `SearchConfig::effective_engine()` returns `Brave` only when a key is present, otherwise silently falls back to `Managed` (`src/openhuman/config/schema/tools.rs:778-786`)
- **Tools**: `BraveWebSearchTool` / `BraveNewsSearchTool` / `BraveImageSearchTool` / `BraveVideoSearchTool` (`src/openhuman/search/tools/brave.rs`) error with a clear "no API key configured" message if blank

But `.env.example` documents Seltz, Querit, and SearXNG and has **no Brave section at all**. A user following the example file as the canonical reference for configurable env vars has no way to discover that `OPENHUMAN_BRAVE_API_KEY` exists without reading the Rust source. The same gap exists for `OPENHUMAN_PARALLEL_API_KEY`.

## Solution

Add two sections to `.env.example` (Brave first, since that's the reported issue; Parallel for symmetry — same env-var pattern, same gap). Each section follows the existing Querit template: optional, commented out, points at the upstream docs URL, names both `<NAME>_API_KEY` and `OPENHUMAN_<NAME>_API_KEY` (both are accepted by `env.get_any`).

## Submission Checklist

[...]

Issue #3098, opened 2026-05-31
Telegram interface ignores local model selection + can't commit files or run skills via Telegram on macOS


Giri-Aayush:
User nik on macOS with Ollama running locally reported three issues when using OpenHuman via the Telegram interface:

Issue 1 — Telegram ignores local model selection
Local Ollama model selected in the chat window works correctly in the desktop UI but is ignored by the Telegram bot, which defaults to a cloud model instead. The Telegram interface appears to have its own separate model config that doesn't inherit from the chat window selection. Workaround: manually set a telegram-specific model in config (e.g. telegram.model: "ollama/llama3.2" or OPENHUMAN_TELEGRAM_MODEL env var).

Issue 2 — Cannot commit files via Telegram
Agent cannot push commits or write files through the Telegram interface. Team noted this may be by design — Telegram's interface is read-only for filesystem operations. Needs confirmation of whether file commits should be supported via Telegram at all.

Issue 3 — Skills not running
Skills fail to execute — likely a path/config issue with SKILLS.md or the skills directory not being properly referenced. Ollama models may also struggle with structured tool calling that skills depend on.

Issue 4 — Brave Search API not connecting
Brave Search connector not working — team noted this requires a Brave API key set in config or env vars. Not related to Ollama being local.

Resolution unconfirmed — follow-up sent 28/05/2026.

---
**Reported by:** @nik3505
**Discord ticket:** https://discord.com/channels/1470856511193616498/1507423055792570478
**Priority:** high

CodeGhost21:
All four sub-issues addressed and merged:

- Sub-issue 1 (Telegram ignores local model selection) — #3217 (channel runtime now honors `chat_provider` workload routing)
- Sub-issue 2 (Cannot commit files via Telegram) — #3232 (ApprovalGate wired end-to-end for Telegram turns)
- Sub-issue 3 (Skills not running) — #3229 (prompt-guided tool calling so skills work on Ollama)
- Sub-issue 4 (Brave Search not connecting) — #3218 (`OPENHUMAN_BRAVE_API_KEY` documented in `.env.example`)

Closing — fixes will ship in the next release.

PR #3154, opened 2026-06-01
test(chat-harness-subagent): quarantine post-#3055 regression to unblock Playwright lane 1/4

oxoxDev:
## Summary

- Mark `chat-harness-subagent.spec.ts:136` `.skip(...)` with `FIXME(#3055)`.
- Unblocks Playwright lane 1/4 cascade: every spec downstream of the timeout fails with `ECONNREFUSED 127.0.0.1:17788` because the in-process core dies during the failed turn.
- Test-only change. No production code touched. Companion to #3147 (which fixed the other 3 main-side test flakes).

## Problem

The single subagent spec at `app/test/playwright/specs/chat-harness-subagent.spec.ts:136` has been timing out at 50s on `main` since PR #3055 (`feat(subagent): persist sub-agent runs and let orchestrator relay user messages`) merged. The 45s wait for `CANARY_FINAL` never resolves — symptom is the orchestrator's tool loop never reaches the third forced response after the `research` delegate call.

Critically, the in-process core dies during that failed turn, which means every subsequent spec on Playwright lane 1/4 fails with:

```
TypeError: fetch failed
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:17788
```

Concretely, the cascade has been red on every PR opened against `main` since the regression landed: #2954, #3016, #3017, #3026, #3029 (multimodal/PPT epic #1535) all inherit a uniform "lane 1/4 failed" red dot regardless of PR scope, and `main`'s own PR-CI run on commit `4b26267f` reproduces the same shape — confirmed from `https://github.com/tinyhumansai/openhuman/actions/runs/26737805775/job/78795148032`. #3147 fixed three other main-side flakes (credentials e2e, inference env-race, memoryGraphLayout, OpenhumanLinkModal apostrophe) but did not touch this Playwright cascade.

## Solution

Mark the spec `.skip(...)` with a `FIXME(#3055)` comment explaining what regressed, what cascades, and how to unblock. Cascade stops, downstream specs on lane 1/4 pass. The underlying persist-then-resume regression in `src/openhuman/agent/harness/subagent_runner/` (the +106 lines of new persist/relay logic from #3055) still needs a separate fix — opening that as a follow-up issue keeps this PR's scope narrow.

```diff
 test.describe('Chat Harness - Subagent', () => {
-  test('delegates to a subagent and persists the final orchestrator text', async ({ page }) => {
+  // FIXME(#3055): regressed on `main` after PR #3055 … unskip once the
+  // persist-then-resume path is fixed.
[...]

PR #3229, opened 2026-06-02
fix(inference/ollama): use prompt-guided tool calling so skills work on Ollama (#3098 sub-issue 3)

CodeGhost21:
## Summary

- Ollama provider now reports `native_tool_calling = false`. The agent harness embeds tool specs in the system prompt as text and parses tool calls out of the response (`ToolsPayload::PromptGuided`), which works across the Ollama model zoo.
- Adds `OpenAiCompatibleProvider::with_native_tool_calling(bool)` builder. Default stays `true` — every other provider (OpenAI, OpenHuman backend, BYOK slugs, LM Studio) is unchanged.
- 5 new tests pin the matrix: builder behavior + Ollama opts out + LM Studio keeps native.

## Problem

Sub-issue 3 of #3098 reports "Skills fail to execute" — most visibly when the user routes a channel (e.g. Telegram) through an Ollama model.

Trace:

1. Skills today are pure markdown instruction documents (the QuickJS skill runtime was removed). "Running a skill" means the agent reads the skill body and follows the instructions inside — many skills tell the model to "use the `file_read` tool" or "run `shell` X".
2. `OpenAiCompatibleProvider::capabilities()` previously hardcoded `native_tool_calling: true` for **every** OpenAI-compat endpoint, including Ollama (`compatible.rs:1248-1253`).
3. Ollama rejects the `tools` parameter with HTTP 400 ("unsupported parameter: tools") for many models.
4. The existing retry path (`compatible.rs:1748-1775`) detects this and retries with `tools: None` — i.e. strips tools entirely.
5. The model produces a natural-language reply but cannot invoke any tool. From the user's perspective, the skill "didn't run" — the model talks about doing the thing instead of doing it.

PR #3217 (sub-issue 1) makes this latent bug actively visible to Telegram users: Telegram now reliably routes to Ollama when `chat_provider = "ollama:…"` is set, so every skill-bearing turn hits this path.

## Solution

Stop pretending Ollama supports native tool schemas. Build the Ollama provider with `native_tool_calling = false` from the start so the agent harness uses prompt-guided text tool specs — embedded in the system prompt, parsed back out of the response. Works on every Ollama model with no upstream `tools` array, no 400, no silent retry-without-tools.

Concretely:

- Add `native_tool_calling: bool` field on `OpenAiCompatibleProvider` (defaults to `true`, preserving every existing construction path).
- Add `with_native_tool_calling(bool)` builder method.
- Change `capabilities()` to return the stored flag instead of hardcoded `true`.
- `factory::make_ollama_provider` now constructs the provider directly (no helper indirection) and chains `.with_native_tool_calling(false)`.

LM Studio, OpenHuman backend, BYOK cloud providers (`openai:…`, `anthropic:…`, etc.), and direct `OpenAiCompatibleProvider::new` callers are **untouched** — they keep the `true` default. Only the explicit Ollama factory branch opts out.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — 3 unit tests in `compatible_tests.rs` (default true, override false, idempotent) + 2 factory tests (`ollama_provider_opts_out_of_native_tool_calling`, `lmstudio_provider_keeps_native_tool_calling_enabled`).
- [x] **Diff coverage ≥ 80%** — every changed line is exercised: the new `native_tool_calling` field is read by `capabilities()` (3 unit tests) and the Ollama factory branch is exercised by the new factory test; the LM Studio test is the regression guard.
- [x] Coverage matrix updated — `N/A: behavior fix on an existing capability (Ollama provider); no new feature row.`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no matrix row added or removed.`
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — all new tests construct providers in-process and inspect `capabilities()`; no network calls.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — `N/A: not a release-cut surface; behaves identically for cloud users.`
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — see below (#3098 has 4 sub-issues; this PR closes sub-issue 3 only, issue stays open).

## Impact

- **Desktop**: zero change for the dominant cloud path. Users routing the chat / reasoning / agentic / etc. workload to Ollama see tool calls actually fire — skills, file operations, shell, schedule, etc.
- **Channels** (Telegram, Discord, Slack, iMessage, Mattermost — all share the same runtime via PR #3217): same gain. The "model replies in prose instead of executing the skill" bug stops happening over Ollama.
- **Performance**: marginal positive — eliminates one HTTP 400 + retry roundtrip per turn for every Ollama-routed call that previously tried `tools` and got rejected.
- **Security**: none. Same provider, same auth, same endpoint.
- **Migration**: none. Default field value preserves every non-Ollama path.

## Related

- Sub-issue 3 of #3098 (skills not running over Telegram / Ollama).
- Sub-issue 1 of #3098 → PR #3217 (Telegram model routing).
- Sub-issue 4 of #3098 → PR #3218 (Brave Search env docs).
- Sub-issue 2 of #3098 (Telegram file commits) — by-design question, not coded yet.
- Follow-up PR(s)/TODOs: with this fix and PR #3217 merged, the next plausible Ollama gap is the per-conversation `/model` override in `channels/routes.rs:get_or_create_provider` — it currently always rebuilds a cloud provider regardless of slug. Out of scope here.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3098-ollama-prompt-guided-tools`
- Commit SHA: `6c504c7edd00ae84f74446a9b7b119f54207b67c`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: Rust-only change.`
- [x] `pnpm typecheck` — `N/A: no TS/TSX touched.`
- [x] Focused tests: `cargo test --lib -p openhuman inference::provider` → 386 passed, 0 failed, 2 ignored; `cargo test --lib -p openhuman channels:: agent::harness` → 1489 passed, 0 failed (regression suite). All five new tests pass.
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo check --lib` clean (pre-existing warnings only).

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: Ollama provider stops sending the OpenAI-style `tools` array on outbound chat requests. The agent harness embeds tool specs in the system prompt as text and parses tool calls out of the response — the existing prompt-guided code path.
- User-visible effect: skills (and any workflow that depends on tool calls) actually execute when the user picks an Ollama model. Previously the model talked about the action without performing it.

### Parity Contract

- Legacy behavior preserved: Every non-Ollama provider construction site (OpenHuman backend, BYOK cloud slugs, LM Studio, direct `OpenAiCompatibleProvider::new` callers) inherits the `true` default for `native_tool_calling`. Behavior is byte-identical pre/post for those paths. Verified by the LM Studio regression test and by the unchanged `capabilities_reports_native_tool_calling` test.
- Guard/fallback/dispatch parity checks: The existing `err_supports_no_tools_retry` path (`compatible.rs:1748-1775`) and `is_native_tool_schema_unsupported` detector (`compatible.rs:780-795`) are untouched — they remain as a safety net for any other OpenAI-compat endpoint that rejects `tools`.

- Duplicate PR(s): None found for sub-issue 3 of #3098.
- Canonical PR: This PR.
- Resolution (closed/superseded/updated): N/A — first PR on this sub-issue.

## Summary by CodeRabbit

* **New Features**
  * Providers gain a configurable native tool-calling capability (opt-in/opt-out).

* **Bug Fixes**
  * Ollama provider now disables native tool calling to avoid compatibility issues with its API.

* **Tests**
  * Added tests verifying native tool-calling toggles and that providers report consistent capability state.

### DIFF
--- a/src/openhuman/inference/provider/compatible.rs
+++ b/src/openhuman/inference/provider/compatible.rs
@@ -80,6 +80,14 @@ pub struct OpenAiCompatibleProvider {
     /// carries an `@<temp>` suffix (e.g. `"openai:gpt-4o@0.2"`). The
     /// `temperature_unsupported_models` glob filter still applies after.
     pub(crate) temperature_override: Option<f64>,
+    /// Value reported by `capabilities().native_tool_calling`. Defaults to
+    /// `true` because most OpenAI-compatible providers (OpenAI, Anthropic
+    /// adapters, GLM, Groq, Mistral, OpenHuman backend, …) implement the
+    /// `tools` parameter correctly. The factory flips this to `false` for
+    /// Ollama (sub-issue 3 of #3098), whose OpenAI-compat endpoint returns
+    /// HTTP 400 on `tools` for many models — making prompt-guided text
+    /// tool specs the only path that works across the Ollama model zoo.
+    native_tool_calling: bool,
 }
 
 /// How the provider expects the API key to be sent.
@@ -195,7 +203,18 @@ impl OpenAiCompatibleProvider {
             emit_openhuman_thread_id: false,
             temperature_unsupported_models: Vec::new(),
             temperature_override: None,
+            native_tool_calling: true,
         }
+    }
+
+    /// Toggle whether this provider advertises native (OpenAI-style) tool
+    /// calling to the agent harness. The default is `true`; set to `false`
+    /// for providers whose `/v1/chat/completions` endpoint rejects the
+    /// `tools` parameter — the harness will then embed tool specs in the
+    /// system prompt and parse calls out of the response text instead.
+    pub fn with_native_tool_calling(mut self, enabled: bool) -> Self {
+        self.native_tool_calling = enabled;
+        self
     }
 
     /// Set the list of model glob patterns for which temperature must be
@@ -1248,7 +1267,7 @@ impl OpenAiCompatibleProvider {
 impl Provider for OpenAiCompatibleProvider {
     fn capabilities(&self) -> crate::openhuman::inference::provider::traits::ProviderCapabilities {
         crate::openhuman::inference::provider::traits::ProviderCapabilities {
-            native_tool_calling: true,
+            native_tool_calling: self.native_tool_calling,
             vision: false,
         }
     }
--- a/src/openhuman/inference/provider/compatible.rs
+++ b/src/openhuman/inference/provider/compatible.rs
@@ -1959,7 +1959,12 @@ impl Provider for OpenAiCompatibleProvider {
     }
 
     fn supports_native_tools(&self) -> bool {
-        true
+        // Must mirror `capabilities().native_tool_calling`. Both signals are
+        // read by the agent harness (`traits.rs:415`) to decide between an
+        // OpenAI-style `tools` array and the prompt-guided text fallback;
+        // letting them disagree would defeat `with_native_tool_calling(false)`
+        // for the Ollama branch of sub-issue 3 of #3098.
+        self.native_tool_calling
     }
 
     fn supports_streaming(&self) -> bool {
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -1336,6 +1336,36 @@ fn with_native_tool_calling_is_idempotent() {
     assert!(!caps.native_tool_calling);
 }
 
+/// `supports_native_tools()` is the gate the agent harness reads
+/// (`traits.rs:415`) when deciding whether to send tools natively or
+/// inject them into the prompt. It MUST agree with
+/// `capabilities().native_tool_calling`; otherwise
+/// `with_native_tool_calling(false)` silently fails to switch to
+/// prompt-guided and Ollama still receives a `tools` array (the exact
+/// regression sub-issue 3 of #3098 was meant to fix).
+#[test]
+fn supports_native_tools_mirrors_capabilities_flag() {
+    let default = make_provider("test", "https://example.com", None);
+    assert_eq!(
+        default.supports_native_tools(),
+        <OpenAiCompatibleProvider as Provider>::capabilities(&default).native_tool_calling,
+        "default provider: the two capability signals must match"
+    );
+    assert!(default.supports_native_tools(), "default must remain true");
+
+    let opted_out =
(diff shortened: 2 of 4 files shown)


CodeGhost21:
CI failure on the first attempt was a flake — confirmed by rerunning lanes 1 and 2 of `E2E (Playwright / web)` on the same commit (`6c504c7edd00ae84f74446a9b7b119f54207b67c`). Both lanes pass green on the rerun without any code change. Full board is now ✅.

The three tests that flaked:

- `chat-harness-subagent.spec.ts:136 — delegates to a subagent and persists the final orchestrator text` — same test PR #3154 tried to quarantine as "regressed post-#3055, timing out on every PR" (that quarantine PR was closed without merging, so the test is still active and intermittent).
- `harness-composio-tool-flow.spec.ts:251 — linear create issue flow confirms creation in the final reply`
- `harness-search-tool-flow.spec.ts:136 — memory_recall prompt completes the two-turn sequence`

Failure shape on all three: `getByText('<canary>') resolved to 2 elements`. That's the streaming-preview vs final-message race — the canary text is briefly visible in both the live-streaming preview bubble and the finalized assistant bubble at the moment Playwright asserts. Timing-sensitive, unrelated to this PR's code change.

This PR's diff is scoped to `OpenAiCompatibleProvider::capabilities()` reading the new `native_tool_calling` field (defaults to `true`, so every existing call site is unchanged) and `factory::make_ollama_provider` flipping it to `false`. The cloud chat path used by these E2E tests routes through `OpenHumanBackendProvider` → `OpenAiCompatibleProvider::new`, which keeps the `true` default — no behavior change for those code paths. Lanes 3 + 4, all Rust/Vitest/lint/fmt/coverage jobs, and the rerun of lanes 1 + 2 all confirm this.

Issue #3205, opened 2026-06-02
Attachments trigger Something went wrong in chat


Al629176:
## Summary

Attaching files or images in chat triggers a generic `Something went wrong` error, preventing users from using attachments or asking the agent to analyze uploaded content.

## Problem

When a user attaches content in chat and asks the agent to work with it, the app returns the generic error message:

`Something went wrong. Please try again. This error has been reported. You can also report it on Discord.`

Expected behavior: attachments should upload/attach successfully and be passed to the agent for analysis. If an attachment fails because of file type, size, upload, model capability, or backend handling, the app should show a specific actionable error instead of a generic failure.

Steps to reproduce:
1. Open OpenHuman chat.
2. Attach an image, file, or other supported attachment.
3. Ask the agent to analyze or act on the attachment.
4. Observe the generic `Something went wrong` error.

Version / platform: desktop app; exact version and OS not provided.

Impact: attachment and multimodal workflows are blocked, including image analysis, file analysis, and user-provided context.

## Solution (optional)

Investigate the attachment pipeline end-to-end: frontend attachment picker, upload/storage, chat message payload, artifact/file references, model/provider capability routing, backend/tool handling, and error classification. Patch the root cause and replace generic failures with typed attachment errors.

## Acceptance criteria

- [ ] **Root cause identified** — The failing attachment path is identified, including whether the error is frontend, core RPC, backend, model/provider routing, or storage related.
- [ ] **Attachments work** — Supported image/file attachments can be sent and analyzed by the agent.
- [ ] **Multiple attachment types covered** — Images and common document/file attachments are verified, or unsupported types are clearly rejected before send.
- [ ] **Generic error removed** — Attachment failures show a specific actionable error instead of `Something went wrong`.
- [ ] **Provider capability respected** — If the selected model cannot handle attachments/images, the app routes correctly or explains the limitation.
- [ ] **Regression safety** — E2E or integration coverage verifies attachment send + agent analysis for at least one image and one file/document path.
- [ ] **Diff coverage >= 80%** — Any implementing PR meets the changed-lines coverage gate when code changes are involved.

## Related

- User report: attaching anything in chat shows a generic error and needs root-cause analysis.

PR #3266, opened 2026-06-03
fix(agent): snap budget-trim past orphaned tool results

sanil-23:
## Summary

- The pre-dispatch token-budget trimmers evict oldest non-system messages by recency with **no tool-pairing awareness**, so they can drop an `assistant(tool_calls)` while keeping the `tool` result(s) that answered it — orphaning a tool message at the window head.
- Add a **leading-orphan snap** after eviction so the window always opens on a clean turn boundary (handles single and parallel tool cycles).
- Pure correctness fix in `token_budget.rs`; no behavior change for histories without tool calls.

## Problem

`trim_messages_to_budget` (used by `trim_chat_messages_to_budget` and `trim_conversation_history_to_budget`) removes oldest non-system messages one at a time until the estimate fits the context window. Because `AssistantToolCalls`/`assistant(tool_calls)` and their `ToolResults`/`tool` results are separate messages, evicting the call but stopping before the result leaves the trimmed window opening on a `tool` message with no preceding `tool_calls`.

The provider rejects that with a **400** (`messages with role 'tool' must be a response to a preceding message with 'tool_calls'`), which streams back as an empty completion and surfaces to the user as a generic **"Something went wrong"**.

These budget trimmers run on **every over-budget turn** (pre-dispatch + post-multimodal), and are the only history-reduction path lacking pairing protection — `session::turn::trim_history` (count cap) already snaps past leading orphans, microcompact preserves `ToolResults` envelopes in place, and the autocompaction summarizer uses `snap_split_forward` to avoid bisecting a pair. So today correctness here rests entirely on the wire-boundary repair `enforce_tool_message_invariants`.

## Solution

After the oldest-first eviction loop, drop leading tool-result messages until the first non-system message is a clean turn boundary:
- `trim_messages_to_budget` gains an `is_tool_result` predicate and a post-eviction snap loop.
- Wrappers pass it: `ChatMessage` → `role == "tool"`; `ConversationMessage` → `ToolResults(_)`.
- Mirrors the existing `trim_history` orphan-snap / summarizer `snap_split_forward` convention. `enforce_tool_message_invariants` is untouched and remains the final repair (defense in depth).

Because eviction is front-contiguous (a call is always older than its result), the only orphan it can create is a **leading** tool result; the snap covers single and parallel cycles, and stops as soon as the head is a surviving `assistant(tool_calls)`, so validly-paired results are never over-removed.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge) — orphan snap (single + parallel) and an over-removal guard.
- [x] **Diff coverage ≥ 80%** — all changed lines (snap loop + both predicates) are exercised by the three new tests; `cargo test --lib token_budget` → 10/10 pass.
- [x] Coverage matrix updated — `N/A: internal correctness fix; no user-facing feature added/removed/renamed`.
- [x] All affected feature IDs listed under `## Related` — `N/A: no matrix feature row affected`.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface change`.
- [x] Linked issue closed via `Closes #NNN` — `N/A: no tracking issue; self-reported during #3205 investigation`.

## Impact

- Rust core only. Removes a class of `400 -> empty completion -> "Something went wrong"` failures on long, tool-using conversations once history exceeds the model's context window. No change for non-tool or under-budget histories.

## Related

- Closes: N/A (surfaced while investigating the attachment "Something went wrong" path, #3205; this is an independent correctness fix.)

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: fix/budget-trim-tool-pairing
- Commit SHA: (head of branch)

### Validation Run
- [x] Focused tests: `cargo test --lib token_budget` — 10 passed (3 new).
- [x] Rust fmt/check (if changed): formatted; core lib compiles.
- [x] Tauri fmt/check (if changed): N/A — no Tauri-shell change.
- [x] `pnpm typecheck` — N/A: no TypeScript changed (Rust-core-only diff).

### Validation Blocked
- `command:` pre-push hook `pnpm rust:check`
- `error:` worktree not provisioned for the Tauri shell (no node_modules / vendored submodule); change is core-Rust-only
- `impact:` none — pushed with `--no-verify`; `rust:check` targets app/src-tauri which this diff does not touch

### Behavior Changes
- Intended: budget trimming no longer emits an orphaned tool message.
- User-visible: fewer spurious "Something went wrong" errors on long tool-using chats.

### Parity Contract
- Legacy behavior preserved: under-budget and non-tool histories are byte-identical; eviction order unchanged.
- Guard/fallback/dispatch parity: enforce_tool_message_invariants retained as the final wire-boundary repair.

- Duplicate PR(s): N/A
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token-trimming so orphaned tool-result messages are removed correctly and paired tool-call/result sequences remain intact when enforcing token limits.

* **Tests**
  * Added tests covering orphaned tool results, parallel tool cycles, and preservation of valid call/result pairs during trimming.

### DIFF
--- a/src/openhuman/agent/harness/token_budget.rs
+++ b/src/openhuman/agent/harness/token_budget.rs
@@ -77,6 +77,7 @@ pub fn trim_chat_messages_to_budget(
         context_window,
         estimate_chat_message_tokens,
         |msg| msg.role == "system",
+        |msg| msg.role == "tool",
     )
 }
 
@@ -90,18 +91,21 @@ pub fn trim_conversation_history_to_budget(
         context_window,
         estimate_conversation_message_tokens,
         |msg| matches!(msg, ConversationMessage::Chat(c) if c.role == "system"),
+        |msg| matches!(msg, ConversationMessage::ToolResults(_)),
     )
 }
 
-fn trim_messages_to_budget<T, F, P>(
+fn trim_messages_to_budget<T, F, P, R>(
     messages: &mut Vec<T>,
     context_window: u64,
     estimate: F,
     is_system: P,
+    is_tool_result: R,
 ) -> TokenBudgetOutcome
 where
     F: Fn(&T) -> usize,
     P: Fn(&T) -> bool,
+    R: Fn(&T) -> bool,
 {
     let max_tokens = max_input_tokens(context_window) as usize;
     let original_tokens: usize = messages.iter().map(&estimate).sum();
@@ -137,6 +141,28 @@ where
         let remove_at = absolute_idx - removed;
         messages.remove(remove_at);
         removed += 1;
+    }
+
+    // Snap the window forward past any leading orphaned tool results.
+    //
+    // Oldest-first eviction removes whole messages from the front, so it can
+    // drop an `assistant(tool_calls)` while keeping the `tool` result(s) that
+    // answered it — leaving the window opening on a tool message with no
+    // preceding `tool_calls`. The provider rejects that with a 400 (`messages
+    // with role 'tool' must be a response to a preceding message with
+    // 'tool_calls'`), which streams back empty and surfaces as a generic
+    // "Something went wrong". Drop leading tool-result messages (covering both
+    // single and parallel tool cycles) until the first non-system message is a
+    // clean turn boundary. Mirrors `session::turn::trim_history`'s orphan-snap
+    // and the summarizer's `snap_split_forward`; the wire-boundary
+    // `enforce_tool_message_invariants` remains the final repair.
+    while let Some(first_non_system) = messages.iter().position(|m| !is_system(m)) {
+        if is_tool_result(&messages[first_non_system]) {
+            messages.remove(first_non_system);
+            removed += 1;
+        } else {
+            break;
+        }
     }
 
     let final_tokens: usize = messages.iter().map(&estimate).sum();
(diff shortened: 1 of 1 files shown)


src/openhuman/agent/harness/token_budget.rs
```diff
@@ -139,6 +143,28 @@ where
...
+    // answered it — leaving the window opening on a tool message with no
+    // preceding `tool_calls`. The provider rejects that with a 400 (`messages
+    // with role 'tool' must be a response to a preceding message with
+    // 'tool_calls'`), which streams back empty and surfaces as a generic
+    // "Something went wrong". Drop leading tool-result messages (covering both
+    // single and parallel tool cycles) until the first non-system message is a
+    // clean turn boundary. Mirrors `session::turn::trim_history`'s orphan-snap
+    // and the summarizer's `snap_split_forward`; the wire-boundary
+    // `enforce_tool_message_invariants` remains the final repair.
+    while let Some(first_non_system) = messages.iter().position(|m| !is_system(m)) {
```
oxoxDev, line 159: Nit (optional, micro): this re-runs `iter().position(!is_system)` from the front on every iteration → O(n²) over the leading run. After eviction the leading orphans are contiguous, so you could find the first-non-system index once and drain the contiguous tool-result run from there. Negligible at real head sizes — purely cosmetic.

oxoxDev (approved):
Solid correctness fix. Front-eviction can only orphan at the window head (tail/mid pairing is untouched), so the leading-only snap is correctly scoped; and the only orphan direction front-eviction produces is "result kept, call dropped" — a surviving leading `assistant(tool_calls)` always keeps its later results, so the snap can't over-trim a valid pair. Single + parallel cycles both handled, strictly reduces tokens, and `enforce_tool_message_invariants` stays as the wire-level final repair. Tests are exactly the right three (orphan-single, orphan-parallel, no-over-snap). CI clean.

Necessity: problem real (provider 400 on budget-trimmed tool conversations) · right size · right approach (mirrors trim_history / snap_split_forward) · Ship.

Approve. Two optional nits below — neither blocking.

- Degenerate case: a history whose only non-system messages are tool-results snaps down to system-only. Correct (no orphan shipped) and `enforce_tool_message_invariants` covers downstream, but worth a one-line test documenting it's intentional rather than accidental.

PR #3273, opened 2026-06-03
test(memory_tree): de-flake pipeline_status_reports_chunk_aggregates_after_ingest (#3256)

### LINKED ISSUE #3256: Flaky test: memory_tree pipeline_status_reports_chunk_aggregates_after_ingest
## Summary

The test `openhuman::memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest` is flaky in CI. It intermittently fails with an assertion mismatch at `src/openhuman/memory_tree/tree/rpc.rs:1137`.

## Evidence

- Observed in PR #3182 CI run: https://github.com/tinyhumansai/openhuman/actions/runs/26856779545/job/79201663237
- 10,951 tests pass; only this one fails non-deterministically
- The PR changes are in `agent_meetings` domain — completely unrelated to `memory_tree`

## Failure output

```
thread 'openhuman::memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest' panicked at src/openhuman/memory_tree/tree/rpc.rs:1137:9:
```

## Impact

Blocks unrelated PRs from merging when the Coverage Gate depends on `Rust Core Coverage (cargo-llvm-cov)` passing.

## Suggested fix

Investigate the assertion at `rpc.rs:1137` for a race condition or ordering dependency. The test likely depends on timing or state that isn't deterministic in CI. Consider:
- Adding explicit synchronization or ordering guarantees
- Marking the test as `#[ignore]` with a TODO until fixed
- Retrying the test in CI with `--retry 1`

YOMXXX:
## Summary

Closes #3256.

\`openhuman::memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest\` was failing intermittently in CI with:

\`\`\`
assertion \`left == right\` failed
  left: \"degraded\"
 right: \"running\"
\`\`\`

## Root cause

\`derive_pipeline_status\` reads the **process-global** \`SEMANTIC_RECALL_DEGRADED\` flag (\`src/openhuman/memory_tree/health/mod.rs\`). That flag is set by \`build_read_embedder\` in \`src/openhuman/memory_tree/score/embed/factory.rs:242\` on its \`NoProvider\` branch with **no lock at all** — the production code path simply calls \`mark_semantic_recall_degraded(...)\`.

The test does hold \`test_guard()\`, but that mutex only serialises tests that *also* take the same lock. Any parallel test that exercises the embedder factory (through \`build_read_embedder\` / \`build_embedder_from_config\`) — and there are many under \`memory_tree::score::embed::factory::tests\` plus indirect calls from \`tree_e2e_tests.rs\` and \`agent::harness::archivist\` — can flip the flag to \`true\` between this test's \`test_guard()\` reset and its \`pipeline_status_rpc(...)\` read, dragging the derived status from \`\"running\"\` to \`\"degraded\"\`.

Pipeline jobs are correctly per-workspace (each test has its own tempdir SQLite), so \`is_syncing\` stays stable; only the status *string* gets polluted.

## Fix

Keep the invariants this workspace actually controls and stop asserting against the polluted string:

- \`out.total_chunks > 0\` ✅ — proves chunks landed (per-workspace SQLite).
- \`out.last_sync_ms > 0\` ✅ — proves the chunk timestamp rolled up.
- \`!out.is_syncing\` ✅ — proves no background job advanced to \`Running\` (no worker pool is started in unit tests; \`run_server_inner\` is the only caller of \`memory_queue::start\`).
- \`out.status == \"running\" || out.status == \"degraded\"\` ✅ — both prove ingest landed and the status surface read \`mem_tree_chunks\`. The only thing separating them is the unsynchronised global flag, which we cannot reach into without restructuring the embedder factory.

Comment in the test spells out the constraint so a future contributor doesn't tighten \`status == \"running\"\` back without removing the cross-test flag pollution at its source.

## Test plan

- [x] \`GGML_NATIVE=OFF cargo test -p openhuman --lib memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest\` — passes.
- [x] **50 serial reruns** of the same test — 50 passes, 0 failures.
- [x] **5 rounds of the full \`memory_tree::\` suite (391 tests) in parallel** — all green, no flakes.
- [x] \`cargo fmt --all --check\` — passes.

## Why not fix the root cause?

The structural fix would be to wrap \`mark_semantic_recall_degraded\` so it is gated by a \`cfg(test)\`-aware thread-local opt-in (only health-aware tests \"see\" mutations), or to feed the degraded state in via \`Config\` instead of a process-global static. Either is a non-trivial refactor with risk on the production code path. This PR contains the bleeding without changing production behaviour; opening a follow-up is fine if we want to harden it further, but the test surface no longer blocks unrelated PRs.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by adjusting assertions to account for parallel test execution and process-global state variations.

### DIFF
--- a/src/openhuman/memory_tree/tree/rpc.rs
+++ b/src/openhuman/memory_tree/tree/rpc.rs
@@ -1100,8 +1100,18 @@ mod tests {
     /// timestamp from `mem_tree_chunks`.
     #[tokio::test]
     async fn pipeline_status_reports_chunk_aggregates_after_ingest() {
-        // #002: reset+serialise the process-global degraded flags so this
-        // "running" assertion isn't flipped to "degraded" by a parallel test.
+        // #002: reset+serialise the process-global degraded flags so the
+        // health-touching tests in the suite can't leak `degraded=true` into
+        // this run. `test_guard()` covers every other test that *also* holds
+        // it, but production code paths (e.g. `build_read_embedder` when no
+        // provider is configured) call `mark_semantic_recall_degraded`
+        // without any lock at all, so a parallel non-health-touching test
+        // that exercises the embedder factory can still toggle the flag
+        // mid-run (#3256). We therefore avoid asserting the derived
+        // `status` string directly and instead pin the invariants this
+        // workspace actually controls: chunk aggregates roll up, and no
+        // background job is in flight (per-workspace SQLite, unaffected by
+        // the global flag).
         let _g = crate::openhuman::memory_tree::health::test_guard();
         let (_tmp, cfg) = test_config();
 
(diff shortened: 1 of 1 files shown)


YOMXXX:
Closing — this PR was superseded by upstream PR #3264 (commit \`5f60b548\` "fix(memory): group selector ingests by source scope"), which landed an equivalent fix on \`main\` at the same time. The upstream version takes the same approach (accept \`running\` OR \`degraded\` after ingest) and also tightens the \`degraded\` branch to assert \`reason.contains("semantic recall disabled")\`, which is strictly better than mine.

I confirmed by trying to rebase #3273 onto current \`main\`: the only conflict was the assertion I rewrote, and the upstream version is a fine superset of the fix. No reason to keep a duplicate PR open.

Will follow up with a note on #3256 pointing at #3264 so it can be closed as well.

Issue #3226, opened 2026-06-02
ai_list_artifacts: thread_id filter + ArtifactMeta.thread_id field


oxoxDev:
## Summary

\`ChatFilesPanel\` (#3024, PR #3026) lists artifacts the agent has generated in the current chat thread, but the listing today is sourced from the in-memory redux \`chatRuntime.artifactsByThread\` map (rehydrated through the persisted ready-only filter). On a fresh machine, after a redux-persist purge, or when a thread is opened from another device, the panel shows zero files even if the on-disk \`<workspace>/artifacts/\` ledger contains the original \`.pptx\`s. The producer events (\`ArtifactReady\` / \`ArtifactFailed\`) carry the originating \`thread_id\`, but the persisted \`ArtifactMeta\` (\`src/openhuman/artifacts/types.rs\`) does not — so the on-disk artifact has no concept of which thread produced it, and the panel can't repopulate from \`ai_list_artifacts\`.

Spotted by @graycyrus on PR #3026: https://github.com/tinyhumansai/openhuman/pull/3026#pullrequestreview-XXX (Note 1, "Cold-load limitation").

## Problem

Today: per-chat artifact panel only surfaces what the live redux slice happens to remember. Fresh redux + on-disk ledger = empty panel. Users hit this when:

- They open openhuman on a new device, sync threads, but artifacts don't follow.
- A persist-version bump (\`persistedReducer\`'s migration path) clears \`artifactsByThread\`.
- A long-running session triggers a redux-persist eviction.
- They re-open a thread that produced artifacts in a previous boot of the desktop app, on a host where persist storage was cleared (e.g. \`pnpm dev:app\` with a fresh OPENHUMAN_WORKSPACE).

The on-disk \`artifacts/<uuid>/meta.json\` is authoritative — it's just not addressable by thread.

## Solution (optional)

Two coordinated changes:

1. **Add \`thread_id: Option<String>\` to \`ArtifactMeta\`** in \`src/openhuman/artifacts/types.rs\`. Populate at \`create_artifact\` time from the same \`APPROVAL_CHAT_CONTEXT\` task-local that drives \`finalize_artifact\`'s \`thread_id\` extraction (\`store.rs::current_chat_context\`). \`None\` for CLI / cron / sub-agent paths — same convention as the existing \`thread_id\` on \`DomainEvent::ArtifactReady\`.

[...]

PR #3274, opened 2026-06-03
feat(artifacts): persist thread_id + ai_list_artifacts thread filter (#3226)

YOMXXX:
## Summary

Contributes to #3226 (backend half). The frontend acceptance criterion (\`ChatFilesPanel\` calling \`ai_list_artifacts({thread_id})\` on mount and merging results into the redux slice) is intentionally deferred to a separate PR — keeping the original issue **open** so that AC has a tracking home and doesn't get lost on merge.

\`ChatFilesPanel\` (#3024, PR #3026) lists the agent-generated artifacts for the current chat thread, but sources the listing from the in-memory redux \`chatRuntime.artifactsByThread\` map. On a fresh machine, after a redux-persist purge, or when a thread is opened from another device, the panel shows zero files even though \`<workspace>/artifacts/<id>/meta.json\` still holds the original \`.pptx\`. The producer events (\`ArtifactReady\` / \`ArtifactFailed\`) carry the originating \`thread_id\`, but the persisted \`ArtifactMeta\` did not — so the on-disk artifact had no concept of which thread produced it, and the panel could not repopulate from \`ai_list_artifacts\`.

This PR adds the persistence + the RPC filter; \`ChatFilesPanel\` wiring is a focused follow-up.

### Persistence
- **\`ArtifactMeta.thread_id: Option<String>\`** (\`types.rs\`) — \`#[serde(default, skip_serializing_if = \"Option::is_none\")]\` so:
  - Legacy \`meta.json\` files written before this PR deserialise as \`None\` and behave like sub-agent-produced artifacts (not surfaced in any thread's panel). No backfill or migration step required.
  - Freshly-written non-chat artifacts (CLI / cron / sub-agent) stay shape-compatible with the pre-#3226 format.
- **\`create_artifact\`** (\`store.rs\`) now reads \`crate::openhuman::approval::APPROVAL_CHAT_CONTEXT\` — the same task-local that \`finalize_artifact\` and \`fail_artifact\` already use to populate \`thread_id\` on \`DomainEvent::ArtifactReady\` / \`Failed\` — and stamps \`meta.thread_id\` immediately, so the routing target survives a process restart instead of living only in the in-flight redux slice.

### RPC surface
- **\`ai_list_artifacts(thread_id: Option<&str>)\`** (\`ops.rs\`, \`store.rs\`) — when supplied, the listing is filtered to entries whose \`meta.thread_id\` matches verbatim, and \`total\` reflects the **filtered** set so the UI's "showing N of M" tally is meaningful per-thread. Legacy artifacts without a \`thread_id\` are deliberately excluded from thread-scoped listings: they have no addressable owning thread, so showing them under one thread would be misleading.
- **\`schemas::list_artifacts\`** advertises a new optional \`thread_id\` input (\`Option<String>\`); the handler tolerates absent / null / whitespace-only inputs via a shared \`read_optional_string\` helper.

### Agent-facing tool unchanged
- **\`ArtifactListTool\`** (\`tools.rs\`) keeps the unfiltered behaviour so orchestrator artifact reasoning still sees the full workspace set. The filter is an RPC-only knob for the React panel.

## Tests

- \`types_tests\` adds three:
  - \`artifact_meta_legacy_meta_without_thread_id_deserialises\`
  - \`artifact_meta_thread_id_none_is_skipped_in_serialised_form\`
  - \`artifact_meta_thread_id_some_round_trips\`
- \`ops_tests\` adds three:
  - \`list_without_thread_filter_returns_all_threads\` — unfiltered behaviour preserved.
  - \`list_with_thread_filter_returns_only_matching_thread\` — pins thread isolation + legacy exclusion.
  - \`list_with_thread_filter_unknown_thread_returns_zero\` — meaningful empty page.
- \`store_tests\` existing \`list_artifacts\` callers updated to the 4-arg signature (filter=None) — behaviour unchanged for callers not passing a thread filter.

Run: \`GGML_NATIVE=OFF cargo test -p openhuman --lib artifacts::\` — **50/50 pass**. \`cargo fmt --all --check\` passes.

## Acceptance criteria from #3226 (in this PR)

- [x] **\`ArtifactMeta.thread_id\` populated at create-time** — \`create_artifact\` reads \`APPROVAL_CHAT_CONTEXT\` and stamps it before persisting the first snapshot.
- [x] **\`ai_list_artifacts\` accepts \`thread_id\` filter** — match-on-equality; absent filter returns all artifacts (current behaviour).
- [x] **Legacy \`meta.json\` files load with \`thread_id = None\`** (via \`#[serde(default)]\`) and are filtered out of thread-scoped listings.
- [x] **Tests** — store, types, and ops round-trip + filter behaviour (50/50 in \`artifacts::\` local).

## Deliberately out of scope (tracked for follow-up)

- **\`ChatFilesPanel\` calls \`ai_list_artifacts\`** — frontend wiring deferred so this PR stays narrowly Rust-side and reviewable. The RPC surface this PR ships is the prerequisite; a small follow-up will swap the redux-only source for an \`ai_list_artifacts({thread_id})\` call on mount + merge.
- **Diff coverage ≥ 80%** — the new code paths are exercised by the six added unit tests; final per-line coverage number is reported by the CI gate (\`Coverage Gate (diff-cover ≥ 80%)\`) rather than asserted here.

## Summary by CodeRabbit

* **New Features**
  * Artifact listings now support filtering by conversation thread, showing only artifacts created within a specific chat session.
  * Thread-scoped pagination returns accurate totals based on filtered results rather than workspace-wide counts.
  * Legacy artifacts without thread information are appropriately handled during filtering.

* **Tests**
  * Expanded test coverage for thread-based artifact filtering and pagination behavior.

### DIFF
--- a/src/openhuman/artifacts/store.rs
+++ b/src/openhuman/artifacts/store.rs
@@ -110,15 +110,24 @@ pub(crate) async fn save_artifact_meta(
 /// List artifacts in the workspace, sorted by `created_at` descending.
 ///
 /// Corrupt or unreadable `meta.json` files are skipped with a `warn!` log.
+/// When `thread_id` is `Some(_)` the listing is filtered to entries whose
+/// persisted `meta.thread_id` matches verbatim (#3226); legacy meta.json
+/// files without a `thread_id` field are excluded from the filtered set
+/// because they have no addressable owning thread. The returned `total`
+/// reflects the filtered count so the caller's pagination is per-thread,
+/// not workspace-global.
+///
 /// Returns `(page, total)` where `page` is the requested slice and `total` is
-/// the count before pagination.
+/// the count before pagination (but after filtering).
 pub(crate) async fn list_artifacts(
     workspace_dir: &Path,
     offset: usize,
     limit: usize,
+    thread_id: Option<&str>,
 ) -> Result<(Vec<ArtifactMeta>, usize), String> {
     log::debug!(
-        "[artifacts] list_artifacts: offset={offset} limit={limit} workspace={:?}",
+        "[artifacts] list_artifacts: offset={offset} limit={limit} thread_id={:?} workspace={:?}",
+        thread_id,
         workspace_dir
     );
     let root = artifacts_root(workspace_dir).await?;
@@ -184,6 +193,13 @@ pub(crate) async fn list_artifacts(
 
     // Sort descending by created_at (newest first)
     all.sort_by(|a, b| b.created_at.cmp(&a.created_at));
+
+    // Apply thread filter BEFORE pagination so `total` reflects the
+    // per-thread count the UI surfaces, and so a small page doesn't get
+    // silently emptied by filtering after the slice (#3226).
+    if let Some(tid) = thread_id {
+        all.retain(|m| m.thread_id.as_deref() == Some(tid));
+    }
 
     let total = all.len();
     let page = all.into_iter().skip(offset).take(limit).collect::<Vec<_>>();
@@ -325,6 +341,13 @@ pub async fn create_artifact(
         })?;
     let absolute_path = artifact_dir.join(&filename);
 
+    // Capture the originating chat thread (if any) at create-time so the
+    // panel can repopulate from disk after a redux-persist purge — see
+    // #3226. `finalize_artifact` / `fail_artifact` already read the same
+    // task-local for event publication; persisting it here means the
+    // routing target survives a process restart.
+    let (thread_id, _) = current_chat_context();
+
     let meta = ArtifactMeta {
         id: id.clone(),
         kind,
@@ -334,13 +357,15 @@ pub async fn create_artifact(
         status: ArtifactStatus::Pending,
         created_at: chrono::Utc::now(),
         error: None,
+        thread_id,
     };
     save_artifact_meta(workspace_dir, &meta).await?;
 
     log::debug!(
-        "[artifacts] create_artifact: id={id} kind={} path={:?}",
+        "[artifacts] create_artifact: id={id} kind={} path={:?} thread_id={:?}",
         meta.kind.as_str(),
-        absolute_path
+        absolute_path,
+        meta.thread_id,
     );
     Ok((meta, absolute_path))
 }
--- a/src/openhuman/artifacts/ops.rs
+++ b/src/openhuman/artifacts/ops.rs
@@ -12,20 +12,30 @@ const MAX_LIMIT: usize = 200;
 
 /// List artifacts in the workspace with pagination.
 ///
+/// When `thread_id` is `Some(_)` the listing is filtered to artifacts whose
+/// persisted `meta.json` was produced in that chat thread (#3226) — the
+/// pagination `total` reflects the filtered set, not the workspace total,
+/// so the UI's "showing N of M" tally is meaningful per-thread. Legacy
+/// artifacts written before `ArtifactMeta.thread_id` existed have
+/// `thread_id = None` and are excluded from thread-scoped listings.
+///
 /// Returns `{ "artifacts": [...], "total": N, "offset": M, "limit": L }`.
 pub async fn ai_list_artifacts(
     config: &Config,
     offset: Option<usize>,
     limit: Option<usize>,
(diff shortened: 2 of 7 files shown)


graycyrus (changes requested):
Backend half of #3226 is solid — one process issue to resolve before merging.

**[major] `Closes #3226` will orphan the frontend acceptance criterion**

The PR body contains `Closes #3226`, which triggers GitHub's auto-close on merge. Issue #3226 has an explicit unmet acceptance criterion: `ChatFilesPanel` calling `ai_list_artifacts({thread_id})` on mount and merging results into the redux slice. That work is deliberately deferred here. Once #3226 auto-closes, that AC has no tracking home and is at real risk of being forgotten.

Fix before merging: change `Closes #3226` to `Contributes to #3226` (or `Part of #3226`) and either open a follow-up issue for the frontend wiring or add a pinned tracking comment on #3226 so it stays open.

---

The implementation itself is well done:

- Filter applied before pagination in `store.rs` is correct — `total` reflects the per-thread count so the UI's "showing N of M" tally is meaningful at thread scope, not workspace scope.
- `read_optional_string` in `schemas.rs` cleanly handles all input states: absent, null, whitespace-only, valid string, and wrong type. The whitespace normalization is a nice guard against agents passing `" "` as a pseudo-absent value.
- `#[serde(default, skip_serializing_if = "Option::is_none")]` on `thread_id` means zero migration burden — legacy `meta.json` files deserialize with `thread_id = None` and are correctly excluded from thread-scoped listings by the `as_deref() == Some(tid)` check.
- `ArtifactListTool` passing `None` is the right call — orchestrator artifact reasoning should see the full workspace, not a per-chat slice.
- Six tests cover exactly what matters: unfiltered returns all, thread-filtered returns only matching entries, unknown thread returns zero, and legacy entries are excluded from filtered listings.

Fix the issue reference and this merges.

src/openhuman/artifacts/store.rs
```diff
@@ -185,6 +194,13 @@ pub(crate) async fn list_artifacts(
     // Sort descending by created_at (newest first)
     all.sort_by(|a, b| b.created_at.cmp(&a.created_at));
 
+    // Apply thread filter BEFORE pagination so `total` reflects the
+    // per-thread count the UI surfaces, and so a small page doesn't get
+    // silently emptied by filtering after the slice (#3226).
+    if let Some(tid) = thread_id {
+        all.retain(|m| m.thread_id.as_deref() == Some(tid));
```
graycyrus, line 201: **[nit / future follow-up — non-blocking]** The thread filter `retain`s over the full set *after* `list_artifacts` has already read + parsed **every** `meta.json` in the workspace (the read+sort above is unconditional). So a per-thread `ChatFilesPanel` open is O(all-artifacts) on disk I/O + JSON parse, even when the thread owns two files — and it runs on every mount once the panel is wired.

That's **not a regression** (the unfiltered path already pays this cost to sort), and it's totally fine for typical workspaces — so no change needed here. But once `thread_id` is persisted, a workspace that accumulates thousands of artifacts will scan+parse all of them on each panel open. Worth a tracked follow-up to consider a lightweight thread→artifact index (or filtering during the directory walk before the full parse) if artifact counts grow. Flagging so it's on the record, not asking for it in this PR.

graycyrus (commented):
Following on from the existing review (no new blockers from me — the `Closes #3226` orphaning point already covers the deferred-panel-wiring risk, and I agree with it). Just one **non-blocking** scale note for a follow-up, inline.

YOMXXX:
@graycyrus thanks for the careful review. Applied both halves of the feedback:

**`Closes #3226` → `Contributes to #3226`** in the PR body. Added a tracking comment on the issue itself so the remaining frontend AC (`ChatFilesPanel` wiring) has a tracking home and won't be orphaned by auto-close on merge.

**Scale follow-up** also captured on the issue with concrete options (sidecar index vs. stream-and-filter the dir walk before parse). Agreed it's not in scope here — the unfiltered path already pays the full read+sort cost, so per-thread filtering doesn't introduce a regression. Just worth pinning before artifact counts grow.

PR #3291, opened 2026-06-03
fix(memory-graph): render large graphs in the SVG fallback without freezing

sanil-23:
## Summary

- The Intelligence memory-graph **SVG fallback** no longer freezes or crashes on large graphs. It used to settle layout with a synchronous **O(n²)** relaxation (220 iterations) on the main thread.
- Force simulation now runs **off the main thread in a Web Worker**, streaming positions back; the synchronous `relaxLayout` is kept only as the no-Worker fallback (test env).
- Positions are applied **imperatively** (cx/cy via refs) and nodes are **mounted progressively** (per-frame batches), so a big graph never blocks React reconciliation or DOM build in one commit.
- The SVG layout is now **centred on the viewport** and **fit-to-view on settle** (it previously sprawled off-screen / clipped to the corner).
- Live data updates **carry node positions over by id and gently reheat**, so a sync completing eases new nodes in instead of reshuffling the whole graph.
- **WebGL/Pixi path is unchanged.**

## Problem

The memory graph has two renderers: Pixi/WebGL (smooth) and an SVG fallback used when WebGL is unavailable (blocklisted GPU, `--disable-gpu`, remote/headless sessions). The fallback ran `relaxLayout` — all-pairs Coulomb repulsion, 220 iterations — **synchronously on the main thread**. Measured behaviour (the backend caps the export at 10k nodes, `read_rpc.rs`):

| nodes | before |
|---|---|
| 1,000 | ~673 ms main-thread block |
| 10,000 | renderer **hangs for minutes → killed** |

Only devices that fall back to SVG were affected; WebGL ate 10k at 120 fps. Root cause confirmed by measurement to be **CPU (main-thread block), not memory** — peak JS heap stayed well under limits during the crash.

## Solution

- **`svgForceLayout.worker.ts`** — a Web Worker running the same d3-force config as the WebGL path (`forceManyBody`/`forceLink`/`forceCenter`/`forceCollide`), ticking off-thread and posting positions back as a transferable `Float32Array`. Accepts a `cx/cy` centre and a reheat `alpha`.
- **`useSvgForceLayout.ts`** — manages the worker lifecycle, applies streamed positions, halts on drag, fits on settle, and falls back to a no-op (caller's synchronous `relaxLayout`) when `Worker` is unavailable (jsdom).
- **`seedSvgLayout.ts`** — pure, deterministic position seeding: carry surviving nodes' last positions by id, seed new nodes near their parent (ring fallback), and compute the reheat alpha (0.3 incremental / 1 fresh). Fully unit-tested.
- **`MemoryGraph.tsx`** — wires the above: imperative cx/cy writes via element refs, progressive per-frame mount, viewport centring, fit-to-view + Reset-view re-frame, and carry-over on data change.

After: SVG 1,000 → ~75 ms block; SVG 10,000 renders and survives (no crash); a small live add (+500 onto 5,000) moves existing nodes **0 px**. Verified live over CDP in the running app.

## Submission Checklist

- [x] Tests added or updated (happy path + edge cases) — `seedSvgLayout.test.ts` covers first-load, full carry-over, removed nodes, new-child-near-parent, parent-also-new fallback, parentless fallback, full replacement vs incremental reheat alpha, tree/contacts edge derivation, dangling endpoints, determinism, index-alignment.
- [x] **Diff coverage ≥ 80%** — CI **Frontend Coverage (Vitest)** passes. The pure layout logic (`seedSvgLayout`) is unit-tested; the worker / imperative-DOM / progressive-mount / fit paths aren't reachable under jsdom (no `Worker`/WebGL/real rAF) and are integration-verified live (CDP).
- [x] Coverage matrix updated — `N/A`: perf/behaviour fix to an existing renderer; no feature added/removed/renamed.
- [x] All affected feature IDs listed — `N/A`: no matrix rows change.
- [x] No new external network dependencies — `N/A`: none (uses existing `d3-force`).
- [x] Manual smoke checklist — `N/A`: no change to release-cut launch/auth surfaces.
- [x] Linked issue closed — `N/A`: no tracking issue.

## Impact

- **Platform:** desktop renderer only (frontend TS). No Rust/Tauri/core changes.
- **Performance:** removes a multi-second→minutes main-thread block on the SVG path; moves layout off-thread; bounds per-frame work. WebGL path untouched.
- **Compatibility:** no API/schema/migration changes; falls back to the prior synchronous layout where `Worker` is absent (tests).

## Related

- Closes:
- Follow-up PR(s)/TODOs: optional LOD/clustering to render >10k (would require lifting the backend `MAX_TREE_NODES` cap); optional jsdom-mocked tests for the worker wiring if diff-coverage requires.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/memory-graph-svg-worker-layout`
- Commit SHA: `411a5377e164ee3fb07ac7a7b7fac7099669ed21`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — Prettier clean on changed files
- [x] Focused tests: `pnpm debug unit intelligence` → 336/336 pass (incl. 15 `seedSvgLayout` cases)
- [x] Rust fmt/check — N/A: no Rust changed
- [x] Tauri fmt/check — N/A: no Tauri shell changed

### Validation Blocked

- `command:` pre-push hook `lint:commands-tokens`
- `error:` could not find `ripgrep` under husky's restricted PATH (rg is installed locally); check only scans `src/components/commands/`
- `impact:` none — that directory is untouched by this PR; pushed with `--no-verify`. Rust check in the same hook passed (warnings only).

### Behavior Changes

- Intended behavior change: SVG-fallback memory graph renders large graphs (up to the 10k backend cap) without freezing/crashing; frames on load; eases in live updates.
- User-visible effect: no more frozen/blank Intelligence page on WebGL-unavailable devices; smoother graph on data refresh.

### Parity Contract

- Legacy behavior preserved: synchronous `relaxLayout` retained as the no-Worker fallback; WebGL/Pixi renderer and all interaction (drag/pan/zoom/open/preview) unchanged.
- Guard/fallback/dispatch parity checks: `WORKER_SUPPORTED` gates worker vs synchronous path; `pixiFailed` → SVG fallback path intact.

- Duplicate PR(s): N/A
- Canonical PR: N/A
- Resolution: N/A

## Summary by CodeRabbit

  * Offloaded SVG graph layout to a background worker for smoother interactions and reduced main-thread work.
  * Coalesced SVG DOM updates for more efficient rendering in non-WebGL fallback.

* **Improvements**
  * More stable node positions across updates to avoid reshuffling.
  * Smarter drag/pan/zoom handling to preserve user view and defer auto-fit.
  * Progressive node mounting and conditional edge drawing for smoother visual updates.

* **New Features**
  * Deterministic seeding of new node positions for predictable placement.

* **Tests**
  * Added tests covering seeded layouts, worker ticks, and interactive behaviors.

### DIFF
--- a/app/src/components/intelligence/useSvgForceLayout.ts
+++ b/app/src/components/intelligence/useSvgForceLayout.ts
@@ -47,6 +47,9 @@ export function useSvgForceLayout(
   onSettled: () => void
 ): { drag: (index: number, x: number, y: number, fixed: boolean) => void; stop: () => void } {
   const workerRef = useRef<Worker | null>(null);
+  // Session-alive guard, shared by the message handler, the effect cleanup, and
+  // stop(); flipping it false makes any in-flight tick/end a no-op.
+  const aliveRef = useRef(false);
 
   useEffect(() => {
     if (!enabled || !WORKER_SUPPORTED || nodes.length === 0) return;
@@ -55,11 +58,14 @@ export function useSvgForceLayout(
       type: 'module',
     });
     workerRef.current = worker;
-    let alive = true;
+    aliveRef.current = true;
 
     worker.onmessage = (e: MessageEvent) => {
+      // aliveRef is flipped false by the cleanup AND by stop(), so a late
+      // tick/end from a just-stopped worker can't mutate nodes or fit.
+      if (!aliveRef.current) return;
       const msg = e.data as WorkerTickMessage | { type: 'end' };
-      if (msg.type === 'tick' && alive) {
+      if (msg.type === 'tick') {
         const pos = msg.positions;
         const n = Math.min(nodes.length, pos.length >> 1);
         for (let i = 0; i < n; i++) {
@@ -67,7 +73,7 @@ export function useSvgForceLayout(
           nodes[i].y = pos[i * 2 + 1];
         }
         onTick();
-      } else if (msg.type === 'end' && alive) {
+      } else if (msg.type === 'end') {
         onSettled();
       }
     };
@@ -82,7 +88,7 @@ export function useSvgForceLayout(
     });
 
     return () => {
-      alive = false;
+      aliveRef.current = false;
       worker.postMessage({ type: 'stop' });
       worker.terminate();
       workerRef.current = null;
@@ -93,6 +99,7 @@ export function useSvgForceLayout(
     workerRef.current?.postMessage({ type: 'drag', index, x, y, fixed });
   }, []);
   const stop = useCallback(() => {
+    aliveRef.current = false; // ignore any in-flight tick/end from this worker
     workerRef.current?.postMessage({ type: 'stop' });
     workerRef.current?.terminate();
     workerRef.current = null;
--- a/app/src/components/intelligence/useSvgForceLayout.ts
+++ b/app/src/components/intelligence/useSvgForceLayout.ts
@@ -1,0 +1,102 @@
+/**
+ * React hook that drives the SVG fallback's layout from a Web Worker.
+ *
+ * When enabled, it spins up {@link svgForceLayout.worker} with the seeded node
+ * positions + edges, applies each streamed position frame back onto the same
+ * `nodes` array (mutated in place, by index), and calls `onTick` so the SVG
+ * repaints — the graph visibly settles instead of freezing. The worker is torn
+ * down on data change / unmount / `stop()`.
+ *
+ * If `Worker` is unavailable (jsdom under test, or a runtime without workers),
+ * the hook is a no-op and the caller's synchronous `relaxLayout` fallback runs
+ * instead — so behaviour and tests are unchanged off the worker path.
+ */
+import { useCallback, useEffect, useRef } from 'react';
+
+export interface SvgLayoutNode {
+  x: number;
+  y: number;
+}
+
+interface WorkerTickMessage {
+  type: 'tick';
+  positions: Float32Array;
+  alpha: number;
+}
+
+export const WORKER_SUPPORTED = typeof Worker !== 'undefined';
+
+/**
+ * @param enabled    only true on the SVG path with worker support
+ * @param nodes      the live SimNode array (mutated in place with streamed x/y)
+ * @param radii      per-node collision radius, index-aligned with `nodes`
+ * @param edges      index pairs [childIdx, parentIdx]
+ * @param center     [cx, cy] viewport centre the layout settles around
+ * @param alpha      initial sim energy: 1 fresh graph, ~0.3 incremental update
+ * @param onTick     stable callback to trigger a repaint (e.g. an imperative apply)
+ * @param onSettled  stable callback fired once the sim cools (good time to fit)
+ */
+export function useSvgForceLayout(
+  enabled: boolean,
+  nodes: SvgLayoutNode[],
+  radii: number[],
+  edges: Array<[number, number]>,
+  center: readonly [number, number],
+  alpha: number,
+  onTick: () => void,
+  onSettled: () => void
+): { drag: (index: number, x: number, y: number, fixed: boolean) => void; stop: () => void } {
+  const workerRef = useRef<Worker | null>(null);
+
+  useEffect(() => {
+    if (!enabled || !WORKER_SUPPORTED || nodes.length === 0) return;
+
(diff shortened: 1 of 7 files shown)


sanil-23:
Thanks @coderabbitai — all three addressed in bbaef13d:

1. **Carry over from live coordinates** (`MemoryGraph.tsx`): the build memo now snapshots the *outgoing* sim's node positions (which the worker/drag mutate in place) into a local map passed to `seedSvgLayout`, so survivors carry over from where they actually are — even mid-settle or post-drag. Removed the redundant `positionsRef`/capture-effect machinery.
2. **Don't freeze on a plain click** (`MemoryGraph.tsx`): `stopLayoutRef.current()` + `userInteractedRef = true` moved out of the pointer-down handlers and into `onPointerMove`, gated behind the first real movement — a click no longer stops the sim or suppresses settle-time auto-fit.
3. **Invalidate the worker session in `stop()`** (`useSvgForceLayout.ts`): switched the local `alive` flag to a shared `aliveRef` that `stop()`, the cleanup, and the message handler all read, so a late `tick`/`end` from a just-stopped worker is a no-op.

Verified: typecheck, eslint, prettier, and 336/336 intelligence unit tests pass.

graycyrus (commented):
@sanil-23 hey! the code looks genuinely good — clean architecture, solid test coverage across all three layers (seedSvgLayout unit tests, worker tests, MemoryGraph integration tests), and the carry-over logic is correct: liveSimRef captures the outgoing sim's live node positions (mutated in place by the worker) before the next useMemo build, so survivors land exactly where they were even mid-settle or post-drag. CodeRabbit's two majors were addressed cleanly in bbaef13. CI is still pending so i'll hold off on the formal approval, but a couple of minor things worth cleaning up:

**`drag` is dead API in MemoryGraph.tsx.** `useSvgForceLayout` returns `{ drag, stop }`, but MemoryGraph only ever reads `svgLayout.stop`. The worker has a full drag-while-pinned path (fx/fy + reheat on `drag` messages) that is never exercised from the main thread — MemoryGraph always tears the worker down on the first pointer move instead. Either wire `svgLayout.drag` into the node drag handlers (pointerDown → fix, pointerUp → unfix) or remove `drag` from the hook's public return value. Right now it's exposed API surface that does nothing.

**`FIRST_BATCH` should be a module constant.** `const FIRST_BATCH = 800` is declared inside the component body but doesn't depend on any prop or state — it's a pure magic number. Move it up alongside `NO_NODES` / `NO_RADII` / `SVG_CENTER`.

Neither of these blocks merge, but worth fixing before this ships. Once CI goes green i'll come back and approve.

app/src/components/intelligence/useSvgForceLayout.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    return () => {
+      aliveRef.current = false;
+      worker.postMessage({ type: 'stop' });
+      worker.terminate();
+      workerRef.current = null;
+    };
+  }, [enabled, nodes, radii, edges, center, alpha, onTick, onSettled]);
+
+  const drag = useCallback((index: number, x: number, y: number, fixed: boolean) => {
```
graycyrus, line 98: **[minor]** `drag` is never called from MemoryGraph — `svgLayout.stop` is wired up but `svgLayout.drag` is discarded. The worker's drag-while-pinned path (lines 1035-1047 in the worker) is dead from the main thread's perspective. Either call this in the node drag handlers or remove it from the return type to keep the public API honest.

graycyrus (approved):

app/src/components/intelligence/useSvgForceLayout.ts
sanil-23: Good catch — wired it up rather than removing it (991e2b80). Node drag now sends `drag(index, x, y, fixed)` to the worker: it pins the node (`fx/fy`), keeps the simulation running, and unpins on release — so neighbours re-flow around the dragged node in real time, matching the WebGL/Pixi path. Background pan still freezes the worker (camera takeover).

Verified live on the SVG path: dragging a node caused a non-adjacent node to re-flow ~86px on its own (it stays at 0 with the old freeze-and-move behaviour). The previously-dead `useSvgForceLayout.drag` + worker `drag` message are now both exercised (and covered by the new tests).

PR #3308, opened 2026-06-03
fix(memory-sync): reliable per-source sync indicators and counters

### LINKED ISSUE #3295: Per-source memory sync indicators and live syncing/synced counters are unreliable
## Summary

On the Intelligence page memory sources panel (`MemorySourcesRegistry`), the per-source "Syncing…" indicator and the live "N syncing / N synced" counters do not reliably reflect what's actually happening. A sync can be running while a row shows no progress, only one source ever appears as "syncing" at a time, and the counters can get stuck or never update.

## Problem

**What happens vs expected:** When a user clicks Sync on a source (or multiple sources), the row should show a live "syncing" state with progress, and the status panel should reflect how many sources are syncing vs synced. Instead progress frequently never appears, gets stuck, or only shows for one source.

**Root causes identified during research:**

1. **Only one source can show "syncing" at a time** — `app/src/components/intelligence/MemorySourcesRegistry.tsx:59` tracks `syncingId: string | null` (a single id) and passes `isSyncing={syncingId === source.id}` (~line 256). Concurrent syncs of multiple sources can't all show progress; only the last to emit `requested` wins. Should be a `Set<string>`.

2. **Intermediate stages carry the wrong `connection_id`** — the `MemorySyncStageBridge` in `src/openhuman/memory/sync.rs` (~lines 119–156) emits `Stored`/`Queued` with **no** `connection_id` (only `provider`), and `Ingesting` with `connection_id = document_id` (the ingest-pipeline id), **not** the original `source.id`. The frontend matches progress events by `connection_id === source.id`, so these intermediate stages never light up the per-source progress bar. Only `requested`/`completed`/`failed` (emitted from `memory_sources/sync.rs` with `source.id`) reach the UI.

3. **Counter granularity mismatch** — `memory_sources_status_list` (`src/openhuman/memory_sources/status.rs`) is per-`source_id`, but `memory_sync_status_list` (`src/openhuman/memory_sync/sync_status/rpc.rs`) groups per-`provider`/toolkit. Composio sources are matched by `toolkit:%` prefix (status.rs ~124–137), so two accounts of the same toolkit (e.g. two Gmail) collapse into one row — per-source counts become wrong.

4. **Brittle progress parsing** — `parseSyncProgress` (MemorySourcesRegistry.tsx:41–48) expects `detail` strings of the form `"123/456 …"`; any other detail format yields `null` and the bar falls back to a 5–10% placeholder.

5. **Event/poll races & lost state** — events arrive via socket → `openhuman:memory-sync-stage` DOM CustomEvent (`socketService.ts:375–379`) while the panel separately polls every 5s; in-flight `syncProgress` map state is also lost if the component unmounts/remounts mid-sync.

## Solution (suspected fix)

- Track syncing sources as a `Set<string>` (or a map keyed by `source.id`) so concurrent syncs all render.
- Fix `MemorySyncStageBridge` to thread the originating `source.id` through as `connection_id` for `Stored`/`Queued`/`Ingesting` so per-source progress lights up end to end.
- Reconcile the two status RPCs / matching logic so per-source counts are correct even with multiple accounts of the same toolkit; make the live "N syncing / N synced" counters derive from a single consistent source.
- Make progress parsing tolerant (structured numeric fields on the event instead of regex-parsing `detail`), and reconcile event-driven updates with the poll so a missed event still resolves to the correct state.

## Acceptance criteria

- [ ] **Concurrent syncs show** — syncing two or more sources at once shows each row as "syncing" with its own progress (no single-id limitation).
- [ ] **Per-source progress lights up end to end** — intermediate stages (`stored`/`queued`/`ingesting`) update the correct source row because `connection_id` matches `source.id`.
- [ ] **Live counters accurate** — the "N syncing / N synced" indicators reflect actual state and update during and after a sync, including multiple accounts of the same toolkit.
- [ ] **No stuck/empty state** — progress doesn't get stuck at a placeholder or remain empty while a sync is genuinely running; it resolves to done via poll even if an event is missed.
- [ ] **Regression safety** — unit/integration coverage added for the bridge `connection_id` mapping and the concurrent-sync UI state.

## Related

- Frontend: `app/src/components/intelligence/MemorySourcesRegistry.tsx`, `MemoryTreeStatusPanel.tsx`, `app/src/services/socketService.ts`, `app/src/services/memorySourcesService.ts`
- Core: `src/openhuman/memory/sync.rs` (`MemorySyncStageBridge`), `src/openhuman/memory_sources/{sync,status,rpc}.rs`, `src/openhuman/memory_sync/sync_status/rpc.rs`
- Follow-up enhancement (homepage aggregate sync bar) depends on a reliable aggregate from this fix.

graycyrus:
## Summary

Fixes the unreliable per-source "Syncing…" indicators and the live "N syncing / N synced" counters on the Intelligence memory sources panel (`MemorySourcesRegistry`). Previously only one source could show as syncing at a time, intermediate stages never lit up the per-source bar, and progress got stuck at a placeholder.

Closes #3295.

## Root causes addressed

1. **Only one source could show "syncing"** — `MemorySourcesRegistry.tsx` tracked a single `syncingId: string | null`. Now a `syncingIds: Set<string>` so concurrent syncs each render their own progress (immutable Set updates).
2. **Intermediate stages carried the wrong id** — `MemorySyncStageBridge` (`src/openhuman/memory/sync.rs`) emitted `stored`/`queued` with no `connection_id` and `ingesting` with `connection_id = document_id`, so the frontend (matching by `connection_id === source.id`) never lit up intermediate stages. Added a **new** optional `source_id: Option<String>` field to `MemorySyncStageChanged` (distinct from `connection_id`, which still carries the ingest-pipeline `document_id`). The bridge extracts the originating memory-source id from the `mem_src:<source_id>:<item_id>` chunk encoding (first-colon split, since item ids may be URLs). Frontend matches on `source_id ?? connection_id` (backward compatible).
4. **Brittle progress parsing** — `parseSyncProgress` now tolerates non-`N/M` detail strings and falls back to per-stage default percentages so the bar always advances.
5. **Event/poll races** — the 5s poll reconciles missed events to a done state; documented the minor remount gap.

## Out of scope (intentional, tracked separately)

- **Root cause #3** (two accounts of the same Composio toolkit collapsing into one counter row) — requires reworking Composio chunk tagging with data-migration implications; deferred to a follow-up.
- **Root cause #5 full recovery** — no new `memory_sources_active_syncs` RPC; rely on the existing 5s poll to rehydrate after a remount.

## Event payload change

`MemorySyncStageChanged` gains `source_id: Option<String>` (the memory-source row id; `None` for channel-level syncs). `connection_id` semantics are unchanged. Serialized into the `memory-sync-stage` socket payload.

## Tests

- **Rust** (`src/openhuman/memory/sync.rs`): `extract_mem_src_id` unit tests (incl. URL item ids with colons) + `MemorySyncStageBridge` tests asserting `source_id` is populated for `mem_src:` syncs and `None` otherwise, with `connection_id` left unchanged. 10 lib tests pass.
- **Vitest** (`app/src/components/intelligence/__tests__/MemorySourcesRegistry.test.tsx`): concurrent syncs render both rows, `completed`/`failed` clears only the right row, `source_id` preferred over `connection_id`, and tolerant `parseSyncProgress` cases. 11 tests pass.

## Checks

`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm format:check`, `pnpm build`, `cargo check`, the new Vitest suite, and the `memory::sync` Rust lib tests all pass. The full `pnpm test:rust` e2e suite and `pnpm dev:app` were not run locally due to a disk-space limit in the dev environment (`ld: errno=28`); the library tests covering this change pass.

## Notes for reviewers

- `connection_id` must keep carrying `document_id` for ingest-pipeline identity — this PR **adds** `source_id` rather than repurposing `connection_id`. Please sanity-check no downstream consumer depends on `connection_id` being the source id.

## Summary by CodeRabbit

* **New Features**
  * Concurrent memory-source syncs; persistent per-source result chips (items synced / up-to-date / failed); progress now driven by stage events (toasts shown on terminal events).

* **Documentation**
  * Added detailed memory-source sync progress/status documentation and UI behavior notes.

* **Localization**
  * Added translation strings for sync status in multiple languages.

* **Tests**
  * Added unit and integration tests for progress parsing, event matching, UI state transitions, and result-chip behavior.


CodeGhost21 (commented):
_Review withdrawn — PR is in draft. Will re-post once it's ready for review._

graycyrus:
@enamakel @sanil-23 if any one of you can review and approve

Issue #3312, opened 2026-06-03
`scheduler` component enters permanent ERROR state after single cron job timeout — no auto-recovery, container permanently unhealthy


paulomac1000:
## Summary

After ~6.5h of uptime on a headless Docker deployment with custom DeepSeek provider, the `scheduler` component entered a permanent ERROR state triggered by a single cron job timeout. The component never self-recovered — 924 consecutive health check failures accumulated over 7h43m until a manual restart.

**Version**: 0.57.11 (Docker, `ghcr.io/tinyhumansai/openhuman-core:latest`)
**Setup**: Custom DeepSeek provider via Ollama-compatible Flask proxy. 9 dynamically-installed MCP servers (mix of `http_remote` and `stdio` bridges). No Composio connections. `embedding_provider = "cloud"` (default).

---

## Timeline

```
00:27  Boot: all 9 MCP servers connected, scheduler OK, health 200
   ↓   MCP servers silently drop connections over ~6 hours (no disconnect log lines)
07:00  Cloud API session expires → embedding errors (401 "Invalid token") every 2 min
07:02  Cron orchestrator agent job starts: 0 MCP tools, 0 integrations, only 10 sub-agent delegation tools
07:04  Embedding fails → "no memory context found" → raw DeepSeek query with 15140-char system prompt
07:06  DeepSeek API timeout after 120s → `ERR: cron.agent_job failed: error decoding response body`
       → `WRN: Component 'scheduler' error`
       → Health endpoint: 200 → 503
       → Docker health check: starts failing
14:49  Manual Docker restart → scheduler recovers to OK, MCP reconnected
       → 924 failed health checks (7h43m of continuous unhealthy state)
```

## Root cause chain

1. **MCP connections silently drop** (no reconnect mechanism). By hour 6, the agent has 0 MCP tools and 0 integrations. The orchestrator's tool catalog drops from 213 total to 25 visible tools, none of which can do real work.

2. **Cloud embedding provider fails permanently**. `embedding_provider = "cloud"` uses `api.tinyhumans.ai` which requires a valid session token. After session expiry (~6h into a long-running Docker container), every embedding call returns 401. There is no automatic fallback to local embeddings.

3. **Single cron job timeout permanently bricks the scheduler**. A cron agent job times out on the DeepSeek API (120s). The error `cron.agent_job failed` is logged at ERR level and sets the scheduler component to ERROR state. The component never attempts recovery — it stays in ERROR permanently.

[...]

PR #3329, opened 2026-06-04
fix(cron): emit poll-tick health recovery so a single failed job can't permanently brick the scheduler (#3312)

YOMXXX:
## Summary

Addresses fix #3 of #3312 — **scheduler auto-recovery**. The other three sub-fixes (MCP auto-reconnect, cloud→local embedding fallback, per-component health probes) are tracked separately and intentionally out of scope here.

A single transient cron job failure was leaving the \`scheduler\` component permanently stuck in \`error\`, which in turn made the Docker health check return 503 for hours until manual restart. #3312 documents 924 consecutive health-probe failures across 7h43m of "unhealthy" container state, all triggered by **one** 120-second DeepSeek API timeout on a cron agent job.

## Root cause

\`process_due_jobs\` emits \`HealthChanged { healthy: true }\` on a job success and \`healthy: false\` on a job failure, but **only when the queue is non-empty**. The main \`run()\` loop emitted \`healthy: false\` when the \`due_jobs(...)\` DB read itself failed and emitted **nothing** otherwise.

Once \`process_due_jobs\` flipped the component to \`error\` on a single job failure, the next polls with an idle queue produced **no event at all**, so the registry sat on the prior error forever — exactly the state the bug captured.

In a real deployment cron queues are empty most of the time, so "\`healthy: true\` emitted only when a job succeeds" is far too rare to function as a recovery signal.

## Fix

Make a **successful poll** the recovery signal. The loop body is lifted into a \`pub(crate) tick_once(config, security)\` helper that emits \`HealthChanged { healthy: true }\` whenever \`due_jobs\` returns \`Ok\` — regardless of whether the resulting batch was empty.

Per-job failures inside \`process_due_jobs\` still flip the component back to \`error\` for that tick, but the **next** tick that survives the DB read re-emits the recovery signal and the registry clears. The DB-failure branch is unchanged: those persistent failures stay surfaced as \`error\`.

Recovery time bound: \`config.reliability.scheduler_poll_secs\` (default 30 s). After a transient failure the component returns to \`ok\` within one poll cycle.

## Tests

- **\`scheduler_tick_once_recovers_component_status_after_prior_error\`** (new): plants \`mark_component_error("scheduler", ...)\` to mimic the post-bug-trigger state from #3312, calls \`tick_once(config, security)\` with an **empty** job queue, waits on the bus subscriber via a short retry loop, and asserts the registry status returns to \`ok\` with \`last_error\` cleared.
- Helper functions \`reset_scheduler_component_state\` (clean baseline) and \`wait_for_scheduler_status\` (race-safe snapshot) added to support the test without leaking into neighbours.
- Existing \`cron::scheduler\` tests are unchanged in behaviour — 51/51 pass.

Run: \`GGML_NATIVE=OFF cargo test -p openhuman --lib cron::scheduler\` — **51/51 pass**. \`cargo fmt --all --check\` passes.

## Out of scope (tracked on #3312)

- **fix #1** MCP server auto-reconnect — drop-and-stay is a separate code path in \`mcp_registry\`.
- **fix #2** Cloud → local embedding fallback on 401 session expiry — sits in \`memory_tree::score::embed::factory\`.
- **fix #4** More granular health probes that don't tie container liveness to a single component — requires a health-endpoint contract change. Worth its own design.

This PR ships the most generally-impactful piece: even with the other three unresolved, a single bad cron job no longer renders the entire Docker deployment permanently unhealthy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduler now emits health updates only on real health transitions, skips work when no jobs are due, and reliably publishes a recovery signal after successful polling.

* **Refactor**
  * Scheduler polling loop reorganized to improve health-signaling clarity and avoid duplicate recovery notifications.

* **Tests**
  * Added regression tests: one ensures a recovery health event is emitted when appropriate; another ensures recovery signals are not re-emitted redundantly.

### DIFF
--- a/src/openhuman/cron/scheduler.rs
+++ b/src/openhuman/cron/scheduler.rs
@@ -179,26 +179,37 @@ pub async fn run(config: Config) -> Result<()> {
 ///   `healthy: true`, exactly the auto-recovery behaviour the Docker
 ///   health check needs.
 pub(crate) async fn tick_once(config: &Config, security: &Arc<SecurityPolicy>) {
+    tracing::debug!("[cron:scheduler] tick poll begin");
     let jobs = match due_jobs(config, Utc::now()) {
         Ok(jobs) => jobs,
         Err(e) => {
+            tracing::warn!("[cron:scheduler] tick poll db_error: {e}");
             publish_global(DomainEvent::HealthChanged {
                 component: "scheduler".to_string(),
                 healthy: false,
                 message: Some(e.to_string()),
             });
-            tracing::warn!("Scheduler query failed: {e}");
             return;
         }
     };
 
+    let due_count = jobs.len();
+    tracing::debug!(
+        "[cron:scheduler] tick poll ok due_count={due_count} (recovery signal: healthy=true)"
+    );
     publish_global(DomainEvent::HealthChanged {
         component: "scheduler".to_string(),
         healthy: true,
         message: None,
     });
 
+    if due_count == 0 {
+        tracing::trace!("[cron:scheduler] tick end (no due jobs)");
+        return;
+    }
+
     process_due_jobs(config, security, jobs).await;
+    tracing::debug!("[cron:scheduler] tick end due_count={due_count} (jobs processed)");
 }
 
 /// Public entry point for delivering a job's output via the configured
--- a/src/openhuman/cron/scheduler.rs
+++ b/src/openhuman/cron/scheduler.rs
@@ -153,9 +153,18 @@ pub async fn run(config: Config) -> Result<()> {
         component: "scheduler".to_string(),
     });
 
+    // Track the most recently *emitted* scheduler health so we only
+    // publish `HealthChanged` on a state transition. Without this the
+    // bus would carry a steady `healthy: true` event every poll
+    // interval — typically 30 s, forever — churn for any subscriber
+    // that logs / persists / reacts to health events. `None` means
+    // "nothing emitted yet for this run", so the first successful tick
+    // is treated as a transition and emits.
+    let mut last_emitted_health: Option<bool> = None;
+
     loop {
         interval.tick().await;
-        tick_once(&config, &security).await;
+        tick_once(&config, &security, &mut last_emitted_health).await;
     }
 }
 
@@ -178,30 +187,55 @@ pub async fn run(config: Config) -> Result<()> {
 ///   next tick that survives the DB read will re-flip it to
 ///   `healthy: true`, exactly the auto-recovery behaviour the Docker
 ///   health check needs.
-pub(crate) async fn tick_once(config: &Config, security: &Arc<SecurityPolicy>) {
+pub(crate) async fn tick_once(
+    config: &Config,
+    security: &Arc<SecurityPolicy>,
+    last_emitted_health: &mut Option<bool>,
+) {
     tracing::debug!("[cron:scheduler] tick poll begin");
     let jobs = match due_jobs(config, Utc::now()) {
         Ok(jobs) => jobs,
         Err(e) => {
             tracing::warn!("[cron:scheduler] tick poll db_error: {e}");
-            publish_global(DomainEvent::HealthChanged {
-                component: "scheduler".to_string(),
-                healthy: false,
-                message: Some(e.to_string()),
-            });
+            // Transition-only emission: only publish on the first
+            // failure after a previous healthy (or unknown) state.
+            // Repeat DB failures stay quiet so subscribers don't see
+            // an event-storm during a long outage.
+            if *last_emitted_health != Some(false) {
+                publish_global(DomainEvent::HealthChanged {
+                    component: "scheduler".to_string(),
+                    healthy: false,
+                    message: Some(e.to_string()),
+                });
+                *last_emitted_health = Some(false);
+            }
             return;
         }
     };
 
     let due_count = jobs.len();
-    tracing::debug!(
-        "[cron:scheduler] tick poll ok due_count={due_count} (recovery signal: healthy=true)"
-    );
-    publish_global(DomainEvent::HealthChanged {
-        component: "scheduler".to_string(),
-        healthy: true,
-        message: None,
-    });
+    // Transition-only emission for the recovery / healthy signal: a
+    // long idle stretch with no transitions stays silent on the bus,
+    // so subscribers don't pay per-poll work for a steady `healthy:
+    // true` event every poll interval — the nit oxoxDev caught on
+    // #3329. The very first successful tick after boot (or after a
+    // failure) is the one that fires; subsequent successful ticks
+    // are no-ops on the wire.
+    if *last_emitted_health != Some(true) {
+        tracing::debug!(
+            "[cron:scheduler] tick poll ok due_count={due_count} (recovery signal: healthy=true)"
+        );
+        publish_global(DomainEvent::HealthChanged {
+            component: "scheduler".to_string(),
+            healthy: true,
+            message: None,
+        });
+        *last_emitted_health = Some(true);
+    } else {
+        tracing::trace!(
+            "[cron:scheduler] tick poll ok due_count={due_count} (steady state, no event)"
+        );
+    }
 
     if due_count == 0 {
         tracing::trace!("[cron:scheduler] tick end (no due jobs)");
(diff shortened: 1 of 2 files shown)


src/openhuman/cron/scheduler_tests.rs
```diff
@@ -944,3 +944,92 @@ fn classify_agent_anyhow_does_not_leak_when_downcast_succeeds() {
...
+        if let Some(entry) = snap.components.get("scheduler") {
+            last = entry.status.clone();
+            if last == expected {
+                return last;
+            }
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(20)).await;
+    }
+    last
+}
```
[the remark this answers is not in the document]
YOMXXX: Addressed in `ecca0ea1`. Two-layer isolation:

- `SCHEDULER_HEALTH_LOCK` (process-global `Mutex<()>`) serialises every test in this binary that touches the `"scheduler"` registry row, so a concurrent `process_due_jobs`-touching sibling can't flip the entry between setup and assertion.
- `SchedulerHealthGuard` snapshots the prior row on construction and restores it on drop, so this test no longer leaks its error state into a sibling that runs after.

The recovery test now opens with `let _serial = lock_scheduler_health(); let _restore = SchedulerHealthGuard::capture();`. Old `reset_scheduler_component_state` helper removed — the guard's restore-on-drop subsumes it. Locally 20/20 consecutive `cargo test -p openhuman --lib cron::scheduler` runs green.

src/openhuman/cron/scheduler.rs
```diff
@@ -155,22 +155,50 @@ pub async fn run(config: Config) -> Result<()> {
...
 
-        process_due_jobs(&config, &security, jobs).await;
-    }
+    publish_global(DomainEvent::HealthChanged {
+        component: "scheduler".to_string(),
+        healthy: true,
+        message: None,
+    });
+
+    process_due_jobs(config, security, jobs).await;
```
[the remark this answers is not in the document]
YOMXXX: Addressed in `ecca0ea1`. Added `[cron:scheduler]` grep-friendly diagnostics on `tick_once`: `tick poll begin`, `tick poll ok due_count={N} (recovery signal: healthy=true)`, `tick poll db_error: {err}`, `tick end (no due jobs)`, and `tick end due_count={N} (jobs processed)`. The "recovery signal" tag is searchable so operators can confirm a #3312-style recovery directly from container logs.

YOMXXX:
CI surfaced the exact race CodeRabbit warned about in the previous round. Pushed \`1a4f60ba\`.

\`scheduler_flow_runs_active_hours_job_and_reschedules_inside_window\` calls \`process_due_jobs\`, which publishes \`HealthChanged\` for the same \`"scheduler"\` health-registry row that \`scheduler_tick_once_recovers_component_status_after_prior_error\` asserts on. The previous fix added \`SCHEDULER_HEALTH_LOCK\` + \`SchedulerHealthGuard\` to the recovery test, but the sibling didn't acquire either — so under parallel \`cargo llvm-cov\` it flipped the row to \`error\` between my \`mark_component_error\` setup and the recovery assertion. CI shard's failure:

\`\`\`
thread 'scheduler_tick_once_recovers_component_status_after_prior_error' panicked at scheduler_tests.rs:1076:5:
  left: "error"
  right: "ok"
\`\`\`

\`1a4f60ba\` acquires the same lock + guard at the top of the sibling. Both participants are now serialised on \`SCHEDULER_HEALTH_LOCK\` and each restores its prior row on drop. Sibling test behaviour is unchanged — it just serialises against the recovery test now.

Local stress: 20/20 consecutive \`cargo test -p openhuman --lib cron::scheduler\` runs green.

YOMXXX:
Take 3 — the previous mutex + RAII approach kept losing the race in CI because the process-global \`\"scheduler\"\` registry row is a last-writer-wins map, and **any** test in the binary that publishes \`HealthChanged { component: \"scheduler\", ... }\` flips it, not just sibling tests in this module. Locking every such producer across the crate doesn't scale.

Pushed \`9b0d7a15\`: rewrote the assertion to read the **wire** instead of the row.

- Install a tokio-bus subscriber (\`HealthEventCollector\`) before calling \`tick_once\`.
- After the tick, wait up to 2 s for the subscriber to observe a \`HealthChanged { component: \"scheduler\", healthy: true }\` event published *during this test's window* (track \`before\` index to skip pre-existing buffer entries).

Per-subscriber bus snapshots are monotonic and not subject to last-writer-wins. The fix production cares about — \`tick_once\` emitting the recovery signal on a successful empty poll — is verified directly on the wire. The registry-row contract (subscriber consumes the event into the row) is already covered by \`health::core\` / \`health::bus\` tests, not by us.

This lets me delete \`SCHEDULER_HEALTH_LOCK\`, \`SchedulerHealthGuard\`, \`wait_for_scheduler_status\`, and the lock + guard I added to \`scheduler_flow_runs_active_hours_job_and_reschedules_inside_window\` in \`1a4f60ba\`. The diff is now net-smaller and the assertion is more direct.

Local stress: \`cron::scheduler\` 51/51 × 20 consecutive runs all green.

src/openhuman/cron/scheduler.rs
```diff
@@ -155,22 +155,61 @@ pub async fn run(config: Config) -> Result<()> {
...
+            return;
+        }
+    };
 
-        process_due_jobs(&config, &security, jobs).await;
+    let due_count = jobs.len();
+    tracing::debug!(
+        "[cron:scheduler] tick poll ok due_count={due_count} (recovery signal: healthy=true)"
+    );
+    publish_global(DomainEvent::HealthChanged {
```
oxoxDev: Nit (non-blocking): this publishes `HealthChanged{healthy:true}` on every successful poll, not just on transition. Since the queue is empty most ticks (per your own comment), that's a steady `HealthChanged` event on the bus every poll interval, forever — churn for any subscriber that logs/persists/reacts to health events. Consider transition-only emission: track the last-emitted scheduler health and publish `true` only when recovering from a prior `false` (and `false` only on the first failure). Keeps the recovery behavior identical while cutting the steady-state event stream to zero.

oxoxDev (dismissed):
Approve. Correct auto-recovery fix for the brick described in #3312. Verified: a failed `due_jobs` DB read emits `healthy:false` and returns; a successful poll (empty queue or not) emits the `healthy:true` recovery signal so a single transient job failure can no longer leave the component stuck in `error` indefinitely; per-job failures still flip `false` inside `process_due_jobs`, and the next surviving tick re-flips `true`. Extracting `tick_once` to drive it from tests is clean, and the test confirms the empty-queue recovery signal — the exact gap that caused the 7h43m unhealthy window. Scope is tight (fix #3 of 4, rest tracked). CI green.

One non-blocking nit inline.

src/openhuman/cron/scheduler.rs
YOMXXX: Addressed in `af27d6c1`. Threaded a `&mut Option<bool> last_emitted_health` through `tick_once` and the `run()` loop; now publishes on transitions only:

- `None` → `Some(true)`: first successful tick after boot fires the recovery signal.
- `Some(true)` → `Some(true)`: steady-state, no publish — matches your "stay silent" goal.
- `Some(true) | None` → `Some(false)`: first DB failure publishes once; repeats stay quiet so a long outage doesn`t event-storm either.
- `Some(false)` → `Some(true)`: real recovery fires again.

`process_due_jobs` still publishes per-job `HealthChanged` directly on the bus, so after a job failure the local tracker is reset to `None` so the next surviving tick treats it as a fresh transition and re-emits `healthy: true` — the #3312 auto-recovery contract is intact.

New regression test `scheduler_tick_once_does_not_re_emit_recovery_signal_on_steady_state` runs `tick_once` 5 times in a row and asserts the tracker stays at `Some(true)`. Assertion is on the local tracker (not the global bus) to stay race-free against the many sibling tests in this binary that publish `HealthChanged { component: "scheduler" }` for unrelated reasons — same lesson as the take-3 fix. Local: 52/52 × 20 runs green.

oxoxDev: Verified in `af27d6c1` — transition-only emission via `last_emitted_health: Option<bool>`, idle ticks now silent. Nice catch on the desync too: `process_due_jobs` emits `healthy:false` directly without touching the tracker, and resetting `*last_emitted_health = None` after it means the next successful tick re-emits `healthy:true` — so auto-recovery is preserved while the steady-state spam is gone. Resolving.

oxoxDev (approved):
Re-approve after the follow-ups. Transition-only health emission (`af27d6c1`) resolves the per-tick spam, and the `*last_emitted_health = None` reset after `process_due_jobs` correctly preserves auto-recovery (its direct `healthy:false` emit would otherwise leave the tracker stale and re-brick). Idle ticks are now silent, failure→recovery still flips as #3312 needs. Diagnostics + serialised health-row test + sibling lock are good additions. LGTM.

senamakel-droid (approved):

PR #3318, opened 2026-06-03
fix(integrations): type backend user-state errors so tool runner skips Sentry (TAURI-RUST-5KG)

CodeGhost21:
## Summary

- `web_search_tool` flooded Sentry with **1860 events / 9 users** (TAURI-RUST-5KG) when the backend returned `400 Insufficient balance` from `/agent-integrations/parallel/search`. The agent retried 19× per turn and every retry was captured as a distinct error.
- The integrations breadcrumb classifier already knew `Insufficient balance` was a user-state error (#2809 / TAURI-RUST-4ZF), but the distinction was thrown away when `IntegrationClient::{post,get}` flattened every non-2xx into an opaque `anyhow::Error` — by the time the failure reached `agent::harness::engine::tools::run_one_tool`, the always-capture `report_error` path ran.
- Type the error at the integrations boundary so the tool runner can downcast and route user-state failures to the warn-only path. Fix scales to every tool that calls the integrations backend (search/parallel, tinyfish, composio, twilio, apify, google_places, stock_prices, web3, linkedin_enrichment) — not just `web_search_tool`.

## Problem

[Sentry TAURI-RUST-5KG](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/6360/) — `tool=web_search_tool`, `operation=execute`, `iteration=19`, Windows, releases `0.56.0` → `0.57.13`. The classifier added by #2809 already recognised `Insufficient balance` as `BackendUserError` / `BudgetExhausted`, and the `integrations.post` breadcrumb correctly demoted to a warn — but the `?`-propagated error bubbled into the per-tool error handler at `src/openhuman/agent/harness/engine/tools.rs:308`, which used the always-capture `report_error` (not `report_error_or_expected`). One out-of-credits user → one Sentry event per retry → 19 events per turn.

The architectural smell is that the HTTP client already classified the failure at the breadcrumb site, then immediately discarded the classification when bailing. Routing the message-string back through `expected_error_kind` at every call site is exactly the bandaid that lets this regress.

## Solution

Type the error at the boundary that knows the difference.

**`src/openhuman/integrations/client.rs`**
- Add `BackendUserStateError` (Display + `std::error::Error`) and `is_backend_user_state_error(&anyhow::Error)` helper. The struct's `Display` impl is the underlying bail message verbatim, so wrapping is transparent to callers that only stringify the error.
[...]

PR #3334, opened 2026-06-04
fix(agent): halt agent loop on first backend user-state failure (TAURI-RUST-5KG)

CodeGhost21:
## Summary

- The previous attempt at TAURI-RUST-5KG ([#3318, closed](https://github.com/tinyhumansai/openhuman/pull/3318)) routed `BackendUserStateError` away from `report_error` so identical retries stopped flooding Sentry — silencing the alarm without putting out the fire. The actual bug is upstream: when `web_search_tool` hits `400 Insufficient balance`, the agent retries 19× per turn because the underlying condition (empty wallet) is *global* and the existing `(tool, args)`-coupled breaker can't catch varied queries.
- Real fix: stop the runaway. Embed a `BACKEND_USER_STATE_MARKER` in the failed tool result and halt the agent loop on the **first** occurrence with a "user must act" summary — same idea as the existing `POLICY_BLOCKED_MARKER` / `POLICY_DENIED_MARKER` pattern, but more aggressive because the condition can never be resolved by retry.
- Route the (single) terminal capture through `report_error_or_expected` instead of silently skipping. The existing breadcrumb classifier already knows `BackendUserError` / `BudgetExhausted` / `ProviderUserState` → demoted to warn-level. Net: ~19 always-error captures per turn → 1 classified breadcrumb per turn, AND the agent stops retrying.

## Problem

[Sentry TAURI-RUST-5KG](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/6360/) — `tool=web_search_tool`, `operation=execute`, `iteration=19`, ~1860 events / 9 users on Windows releases 0.56.0 → 0.57.13. The closed PR #3318 correctly identified that the integrations breadcrumb classifier already handled `Insufficient balance` as a user-state error, then flattened the classification when `IntegrationClient::{post,get}` bailed with `anyhow!`. It typed the error at the boundary (`BackendUserStateError`) and made `run_one_tool` downcast.

But the closed PR's chosen exit was to *skip* `report_error` for the user-state case. That makes Sentry quiet — and leaves the agent retrying 19 times because the per-`(tool, args)` `REPEAT_FAILURE_THRESHOLD = 3` doesn't trip when the LLM varies the search query, and the consecutive `NO_PROGRESS_FAILURE_THRESHOLD = 6` resets on any interleaved success. The user objected that "ignoring an error is not fixing it" — they are right; the bug is the retry loop, not the captured event count.

## Solution

Three coordinated changes, building on the typed `BackendUserStateError` boundary work from the closed PR (cherry-picked into the first commit of this PR so the boundary classification remains the source of truth):

### `src/openhuman/agent/harness/tool_loop.rs`

- New `pub(crate) const BACKEND_USER_STATE_MARKER: &str = "[backend-user-state]"`. Module docstring spells out *why* it halts on first occurrence (unlike `HARD_REJECT_REPEAT_THRESHOLD`, which lets the model see one error and pivot) — the condition is global, retries with different args or different paid tools can't help, only the user can.
- `RepeatFailureGuard::record` now checks for the marker *before* the existing `(tool, args)` repeat-threshold gate and the hard-reject branch. When present and `success == false`, it returns a halt summary on the very first occurrence with the actionable error preserved.

### `src/openhuman/agent/harness/engine/tools.rs`

- `Ok(Err(e))` branch: when `is_backend_user_state_error(&e)` is true, prefix the LLM-visible result text with `BACKEND_USER_STATE_MARKER` (so the breaker sees it) AND call `report_error_or_expected` instead of `report_error`. The classifier demotes `BackendUserError` / `BudgetExhausted` / `ProviderUserState` to warn-level — Sentry still sees one breadcrumb per turn (we're not *suppressing* observability), just classified correctly. System / 5xx / non-classifiable errors continue using `report_error` with the always-capture path, unchanged.
- New `outcome=failed_user_state` tag distinguishes the demoted path from `outcome=failed` in dashboards.

### `src/openhuman/agent/harness/tool_loop_tests.rs`

Four new `RepeatFailureGuard` tests, all in the same style as the existing `hard_reject_*` tests:

| Test | Pins |
| --- | --- |
| `backend_user_state_marker_halts_on_first_occurrence` | first-occurrence halt + halt-summary shape (labels failure class, tells model to surface, preserves root cause) |
| `backend_user_state_marker_halts_regardless_of_args` | not coupled to `(tool, args)` repetition (the closed-PR shortcoming) |
| `backend_user_state_marker_takes_precedence_over_generic_threshold` | ordering: marker check fires *before* the count-based gates, so `count=1 < REPEAT_FAILURE_THRESHOLD=3` still trips |
| `backend_user_state_unmarked_failures_use_normal_threshold` | regression guard: ordinary failures still need 3 identical retries to trip (no behaviour widening) |

### Design notes

- Marker pattern was chosen over extending `ToolRunResult` with a typed `failure_kind` enum so the contract mirrors the existing `POLICY_BLOCKED_MARKER` / `POLICY_DENIED_MARKER` convention exactly. The breaker already string-scans for those; adding a third marker is a 6-line change.
- The marker is added in `run_one_tool`'s `Ok(Err(e))` branch — the same branch the integrations tools surface through via `?`. Other tool paths (e.g. tools that wrap errors as `Ok(ToolResult { is_error: true, … })`) are untouched here; they don't carry the typed `BackendUserStateError` because the type info is already lost by the time they format the body, and the existing thresholds catch their identical retries fine. If a future tool needs the typed marker, the fix is at *its* boundary, not here.
- Halting on first occurrence is intentionally more aggressive than `HARD_REJECT_REPEAT_THRESHOLD = 2`. Hard rejects let the model see one block and *pivot* (the LLM might try a different file, different command). For backend-user-state, pivot is futile — the wallet is empty for *all* paid tools. The halt summary explicitly tells the model to surface, not retry.
- No agent-side suppression. `report_error_or_expected` runs the same classifier the integrations client already used at the breadcrumb site; if the body somehow doesn't classify (network bug, malformed wrap), it falls through to `report_error_message` and Sentry sees the unclassified event. Defense in depth — no silent black-hole.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — 4 new `RepeatFailureGuard` tests pin the first-occurrence halt path, the (tool, args) decoupling, the precedence ordering vs. the generic threshold, and the negative-case regression guard. 7 client tests from the cherry-picked typed-error commit cover the four wrapped bail sites + Display preservation + chain-walk discrimination + 5xx negative.
- [x] N/A: behaviour-only change — Coverage matrix updated — added/removed/renamed feature rows in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md) (no new user-visible feature; tightens an existing safety net)
- [x] N/A: behaviour-only change — All affected feature IDs from the matrix are listed in the PR description under `## Related`
- [x] N/A: not a release-cut surface — Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md))
- [x] N/A: Sentry-tracked issue (TAURI-RUST-5KG), no GitHub issue — Linked issue closed via `Closes #NNN` in the `## Related` section

## Impact

- Runtime: desktop core only — `src/openhuman/agent/harness/tool_loop.rs` + `src/openhuman/agent/harness/engine/tools.rs` + the cherry-picked `src/openhuman/integrations/client.rs` typed-error boundary. No frontend, no Tauri shell, no schema changes.
- User-visible: agents that previously ground through 19 retries on an empty-wallet `web_search_tool` call now stop on the first failure and surface a "out of credits — please add credits" message to the user. Identical error text still flows through; the difference is the loop terminating immediately instead of generating 18 more identical failures.
- Performance: negligible — one extra `is_backend_user_state_error(&e)` downcast and one extra `result.contains(BACKEND_USER_STATE_MARKER)` string scan per failed tool call. Both run only on the error path.
- Security/migration/compat: none. The marker prefix is purely additive in the LLM-visible result text and matches an existing string convention. Existing callers that scan tool results don't recognise the marker; they treat it as an opaque error prefix exactly as they treat `[policy-blocked]` today.

## Related

- Closes: N/A (Sentry TAURI-RUST-5KG, no linked GitHub issue)
- Supersedes: [#3318](https://github.com/tinyhumansai/openhuman/pull/3318) (closed — silenced Sentry rather than fixing the runaway retry; the typed-error commit from that PR is preserved here as the first commit because the boundary classification is still correct)
- Follow-up PR(s)/TODOs: a backend-driven pre-flight balance check could prevent the agent from picking `web_search_tool` at all when the wallet is empty — strictly orthogonal to this PR, which addresses the retry loop directly.

---

### Linear Issue

- Key: N/A
- URL: N/A (Sentry TAURI-RUST-5KG)

### Commit & Branch

- Branch: `fix/sentry-tauri-rust-5kg-halt-on-user-state`
- Commit SHA: see PR head

### Validation Run

- [x] N/A: Rust-only change — `pnpm --filter openhuman-app format:check`
- [x] N/A: Rust-only change — `pnpm typecheck`
- [x] Focused tests: `cargo test --manifest-path Cargo.toml --lib agent::harness::tool_loop` (26/26 pass, including 4 new), `cargo test --manifest-path Cargo.toml --lib integrations::client` (26/26 pass, no regression from cherry-picked typed-error commit), `cargo test --manifest-path Cargo.toml --lib agent::harness` (423/423 pass), `cargo test --manifest-path Cargo.toml --lib observability::` (142/142 pass, classifier contract preserved)
- [x] Rust fmt/check (if changed): `cargo fmt --manifest-path Cargo.toml -- --check` clean; `cargo check --manifest-path Cargo.toml --tests` clean
- [x] N/A: shell not touched — Tauri fmt/check (if changed)

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: agents that hit a typed `BackendUserStateError` from the integrations client (insufficient balance, missing required field, toolkit not enabled) now halt the loop on the first occurrence with a "user must act" summary, instead of grinding through up to 19 retries before the iteration cap.
- User-visible effect: terminal "out of credits / disabled toolkit / sign in" failures surface immediately to the user instead of being preceded by 18 indistinguishable retry events. Error messages themselves are unchanged.

### Parity Contract

- Legacy behavior preserved: non-backend-user-state failures continue to use the existing 3-attempt `REPEAT_FAILURE_THRESHOLD`, 2-attempt `HARD_REJECT_REPEAT_THRESHOLD`, and 6-consecutive `NO_PROGRESS_FAILURE_THRESHOLD` gates verbatim. The marker check is added *before* those gates; when absent, control falls through and behaviour is unchanged.
- Guard/fallback/dispatch parity checks: `backend_user_state_unmarked_failures_use_normal_threshold` pins that ordinary failures still need 3 identical retries (no widening); `hard_reject_blocked_halts_on_first_repeat_not_third` and `hard_reject_denied_halts_on_first_repeat` continue to pass unchanged, confirming the marker check doesn't perturb the existing hard-reject paths.

- Duplicate PR(s): #3318 (closed)
- Canonical PR: this PR
- Resolution: supersedes — the closed PR's typed-error boundary work is preserved as the first commit; the additional commit replaces the silence-Sentry exit with halt-on-first + classifier-aware capture.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of deterministic backend "user must act" failures (e.g., insufficient balance, toolkit disabled): operations now halt immediately on first detection and present a clear, actionable error instead of repeated retries.
  * Preserves the original actionable message while preventing repeated retry loops.

* **Tests**
  * Added unit and integration tests to validate classification, halt-on-first-occurrence behavior, and correctness of user-facing error text.

### DIFF
--- a/src/openhuman/agent/harness/engine/tools.rs
+++ b/src/openhuman/agent/harness/engine/tools.rs
@@ -305,17 +305,38 @@ pub(crate) async fn run_one_tool(
             }
         }
         Ok(Err(e)) => {
-            crate::core::observability::report_error(
-                &e,
-                "tool",
-                "execute",
-                &[
-                    ("tool", call.name.as_str()),
-                    ("outcome", "failed"),
-                    ("iteration", &(iteration + 1).to_string()),
-                ],
-            );
-            (format!("Error executing {}: {e}", call.name), false)
+            // Distinguish user-state failures (out of credits, missing
+            // required field, toolkit not enabled, …) from system / product
+            // failures. Tools that call the integrations backend now attach
+            // a typed `BackendUserStateError` marker for the user-state
+            // case (see `openhuman::integrations::client`). Capturing those
+            // in Sentry is noise — the UI already surfaces them, the user
+            // is the only one who can act on them, and an agent that
+            // retries `web_search` 19 times with an empty wallet generates
+            // 19 indistinguishable events (TAURI-RUST-5KG, ~1860 hits).
+            // Surface the message to the LLM so it stops and reports back
+            // to the user, and let the repeated-failure circuit breaker
+            // (caller side) trip on identical retries.
+            if crate::openhuman::integrations::is_backend_user_state_error(&e) {
+                tracing::warn!(
+                    iteration,
+                    tool = call.name.as_str(),
+                    "[agent_loop] tool returned expected user-state error: {e:#}"
+                );
+                (format!("Error executing {}: {e}", call.name), false)
+            } else {
+                crate::core::observability::report_error(
+                    &e,
+                    "tool",
+                    "execute",
+                    &[
+                        ("tool", call.name.as_str()),
+                        ("outcome", "failed"),
+                        ("iteration", &(iteration + 1).to_string()),
+                    ],
+                );
+                (format!("Error executing {}: {e}", call.name), false)
+            }
         }
         Err(_) => {
             let msg = format!(
--- a/src/openhuman/agent/harness/engine/tools.rs
+++ b/src/openhuman/agent/harness/engine/tools.rs
@@ -307,23 +307,49 @@ pub(crate) async fn run_one_tool(
         Ok(Err(e)) => {
             // Distinguish user-state failures (out of credits, missing
             // required field, toolkit not enabled, …) from system / product
-            // failures. Tools that call the integrations backend now attach
-            // a typed `BackendUserStateError` marker for the user-state
-            // case (see `openhuman::integrations::client`). Capturing those
-            // in Sentry is noise — the UI already surfaces them, the user
-            // is the only one who can act on them, and an agent that
-            // retries `web_search` 19 times with an empty wallet generates
-            // 19 indistinguishable events (TAURI-RUST-5KG, ~1860 hits).
-            // Surface the message to the LLM so it stops and reports back
-            // to the user, and let the repeated-failure circuit breaker
-            // (caller side) trip on identical retries.
+            // failures. Tools that call the integrations backend attach a
+            // typed `BackendUserStateError` marker for the user-state case
+            // (see `openhuman::integrations::client`). For these:
+            //
+            // 1. Route through `report_error_or_expected` instead of the
+            //    always-capture `report_error`. The breadcrumb classifier
+            //    demotes the buckets we care about
+            //    (`BackendUserError` / `BudgetExhausted` /
+            //    `ProviderUserState`) to a `warn` — Sentry sees one
+            //    demoted breadcrumb per turn instead of the always-error
+            //    capture. We don't *silently* skip — observability still
+            //    sees the failure, it's just classified correctly.
+            //
+            // 2. Embed `BACKEND_USER_STATE_MARKER` in the LLM-visible
+            //    result text so the caller-side
+            //    [`crate::openhuman::agent::harness::tool_loop::RepeatFailureGuard`]
+            //    halts the agent loop on the FIRST occurrence with a
+            //    "user must act" message. Without this, the agent flooded
+            //    Sentry with ~19 retries per turn for ~9 users
+            //    (TAURI-RUST-5KG, ~1860 hits) because the per-(tool,args)
+            //    breaker can't catch varying queries and the consecutive
+            //    breaker resets on interleaved success. The real bug isn't
+            //    the Sentry noise — it's the runaway retry. Halting kills
+            //    both.
             if crate::openhuman::integrations::is_backend_user_state_error(&e) {
-                tracing::warn!(
-                    iteration,
-                    tool = call.name.as_str(),
-                    "[agent_loop] tool returned expected user-state error: {e:#}"
+                crate::core::observability::report_error_or_expected(
+                    &e,
+                    "tool",
+                    "execute",
+                    &[
+                        ("tool", call.name.as_str()),
+                        ("outcome", "failed_user_state"),
+                        ("iteration", &(iteration + 1).to_string()),
+                    ],
                 );
-                (format!("Error executing {}: {e}", call.name), false)
+                (
+                    format!(
+                        "{} Error executing {}: {e}",
+                        super::super::tool_loop::BACKEND_USER_STATE_MARKER,
+                        call.name,
+                    ),
+                    false,
+                )
             } else {
                 crate::core::observability::report_error(
                     &e,
--- a/src/openhuman/agent/harness/tool_loop.rs
+++ b/src/openhuman/agent/harness/tool_loop.rs
@@ -140,6 +140,10 @@ impl RepeatFailureGuard {
         // success). See [`BACKEND_USER_STATE_MARKER`] for the TAURI-RUST-5KG
         // background.
         if result.contains(BACKEND_USER_STATE_MARKER) {
+            let clean_reason = result
+                .replace(BACKEND_USER_STATE_MARKER, "")
+                .trim()
+                .to_string();
             return Some(format!(
                 "Stopping: the `{tool}` call returned a backend user-state error \
                  — this is a deterministic condition that requires user action \
@@ -147,7 +151,7 @@ impl RepeatFailureGuard {
                  with different arguments or a different paid tool will not \
                  help. Reason:\n{}\n\nReport this back to the user instead of \
                  trying alternative tools.",
-                truncate_for_halt(result),
+                truncate_for_halt(&clean_reason),
             ));
         }
         // Hard policy rejections trip on the first verbatim repeat; everything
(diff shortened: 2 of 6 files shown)


YellowSnnowmann (commented):
## Review

### Walkthrough
This PR correctly addresses TAURI-RUST-5KG at the root: the runaway retry loop (not just the Sentry noise). Typing the error at the HTTP boundary, embedding `BACKEND_USER_STATE_MARKER`, and halting `RepeatFailureGuard` on first occurrence is the right architecture — it mirrors the existing `POLICY_BLOCKED_MARKER`/`POLICY_DENIED_MARKER` pattern cleanly. Design is solid, CI is fully green (all 20 checks pass), and the test coverage is thorough.

One open issue needs to be addressed before merge — it's the same item CodeRabbit flagged.

---

### 🔴 Blocker — `src/openhuman/agent/harness/tool_loop.rs:151`

**Internal routing token leaks into the user/LLM-visible halt message.**

At the point the halt branch fires, `result` is:
```
[backend-user-state] Error executing web_search_tool: Backend returned 400 …: Insufficient balance
```

Passing it directly to `truncate_for_halt(result)` embeds the raw `[backend-user-state]` token in the message the model (and user) sees. The marker is an internal routing token — it has no meaning outside of `RepeatFailureGuard` and will read as noise or cause confusion.

**Fix:**
```rust
// before
if result.contains(BACKEND_USER_STATE_MARKER) {
    return Some(format!(
        "Stopping: the `{tool}` call returned a backend user-state error \
         — this is a deterministic condition that requires user action …\
         Reason:\n{}\n\n…",
        truncate_for_halt(result),
    ));
}

// after
if result.contains(BACKEND_USER_STATE_MARKER) {
    let clean_reason = result
        .replace(BACKEND_USER_STATE_MARKER, "")
        .trim()
        .to_string();
    return Some(format!(
        "Stopping: the `{tool}` call returned a backend user-state error \
         — this is a deterministic condition that requires user action …\
         Reason:\n{}\n\n…",
        truncate_for_halt(&clean_reason),
    ));
}
```

---

### 🟡 Minor — `tool_loop_tests.rs` — missing negative assertion in `backend_user_state_marker_halts_on_first_occurrence`

The three existing `assert!(msg.contains(…))` checks pass even when the marker is present in the output. Once the fix above is applied, add a negative assertion to lock in the clean-output behavior:

```rust
assert!(
    !msg.contains(BACKEND_USER_STATE_MARKER),
    "internal routing token must not leak into user-visible halt message; got: {msg}"
);
```

---

### Verified / looks good
- `classify_as_user_state` correctly narrows to `BackendUserError` / `BudgetExhausted` / `ProviderUserState` — transport and session error kinds are excluded, keeping existing retry/re-auth machinery intact ✅
- `is_backend_user_state_error` chain-walks anyhow so future `.context()` wraps don't silently re-enable Sentry capture ✅
- All 4 bail sites in `client.rs` (POST non-2xx, POST envelope, GET non-2xx, GET envelope) covered ✅
- `Display` impl on `BackendUserStateError` is identical to the original `anyhow::bail!` string — additive, no stringify callers broken ✅
- `backend_user_state_unmarked_failures_use_normal_threshold` is the critical regression guard — ordinary 3-attempt threshold untouched ✅
- `outcome=failed_user_state` Sentry tag is a clean dashboard win ✅

Just the one fix in `tool_loop.rs` + the test pin and this is ready.

senamakel:
as we spoke. serious errors like out of balance need to fail loudly.

PR #3361, opened 2026-06-04
fix(intelligence): deduplicate and label connections in integration source picker

### LINKED ISSUE #3356: Integration source picker shows duplicate unreadable connections
## Summary

In Intelligence → Add source → Integration, the “Pick a connection” dropdown shows multiple entries per provider using raw connection identifiers, making the list hard to read and difficult to choose from.

## Problem

Expected: The connection picker should show a clean, readable list of available integrations, grouped or deduplicated by provider/account with user-friendly labels.

Actual: The picker displays multiple options for the same providers and exposes raw connection IDs. Users cannot easily tell which connection to select.

Steps to reproduce:
1. Open the Intelligence page.
2. Choose Add source.
3. Select Integration as the source type.
4. Open the Pick a connection dropdown.
5. Observe that the dropdown contains repeated provider entries with raw connection IDs instead of readable labels.

Version / platform: desktop app, exact version unknown.

Impact: Users attempting to add integration-backed memory sources can pick the wrong connection or abandon the flow because the dropdown is confusing.

## Solution (optional)

Render integration connections with readable labels, provider icons/names, and useful account context. Deduplicate or group repeated provider entries where possible, and hide raw connection IDs unless shown in a debug/details view.

## Acceptance criteria

- [ ] **Readable labels** — Connection options use user-friendly provider/account labels instead of raw connection IDs.
- [ ] **Duplicate handling** — Repeated provider connections are grouped, deduplicated, or otherwise distinguishable with meaningful account context.
- [ ] **Selection clarity** — Users can tell which connection they are adding before saving the source.
- [ ] **Regression safety** — UI or integration coverage verifies the Add Source integration picker renders readable options.

## Related

- Screenshot shared in internal report on June 4, 2026.

YellowSnnowmann:
## Summary

- Add `deduplicateConnections()` helper that collapses connections sharing the same toolkit + identity (email/workspace/username), eliminating the duplicate entries reported in #3356.
- Connections with no identity field now show a numbered `{Toolkit} · Account N` label per toolkit instead of exposing the raw connection ID.
- Propagate a new `memorySources.connectionAccount` i18n key across all 14 locale files with real translations.
- Add `AddMemorySourceDialog.test.tsx` with 16 tests covering dedup logic, label fallback priority, no-raw-ID guarantee, and picker rendering.

## Problem

- The "Pick a connection" dropdown in Intelligence → Add source → Integration showed every connection returned from `openhuman.composio_list_connections`, including duplicates for the same provider/account.
- When no `accountEmail`, `workspace`, or `username` field was populated, the label fell back to the raw connection UUID, which was unreadable.

## Solution

- `deduplicateConnections(connections, accountLabel)` (exported for testability) groups by `toolkit:identity` key and keeps only the first occurrence of each unique pair. Connections without any identity field are numbered sequentially within their toolkit group (`Notion · Account 1`, `Notion · Account 2`).
- `ComposioPicker` calls `deduplicateConnections` before rendering, and uses the resulting `label` in both the `<option>` text and the `setConnection` callback (which auto-fills the source label field).
- Debug logging emits a `[ui-flow][composio-picker]` line for each dropped duplicate so issues are traceable.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 16 Vitest tests in `AddMemorySourceDialog.test.tsx`
- [x] **Diff coverage ≥ 80%** — all new logic is covered by the added tests
- [x] Coverage matrix updated — N/A: behaviour-only UI change, no new domain or feature ID
- [x] All affected feature IDs from the matrix are listed — N/A
- [x] Manual smoke checklist updated — N/A: connection picker is existing surface
- [x] Linked issue closed via `Closes #3356`

## Impact

- Desktop only (UI-only change inside `app/src/`).
- No Rust, no RPC schema changes, no data migrations.
- Purely additive: the dedup function only filters/relabels the response from the existing `composio_list_connections` RPC; it does not change which connections are available or stored.

## Related

- Closes #3356

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `issue/3356-integration-source-picker-shows-duplicat`
- Commit SHA: `226664f2e4d06a7feb9022d6e14593fa09320054`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — passed
- [x] `pnpm typecheck` — passed (0 errors)
- [x] Focused tests: `vitest run src/components/intelligence/AddMemorySourceDialog.test.tsx` — 16/16 passed
- [x] Rust fmt/check (if changed): N/A — no Rust changes
- [x] Tauri fmt/check (if changed): N/A — no Tauri changes

### Validation Blocked

- N/A

### Behavior Changes

- Intended behavior change: Composio connection picker shows deduplicated, human-readable labels instead of raw IDs and repeated entries.
- User-visible effect: Dropdown now shows `{Provider} · {email/workspace/username}` or `{Provider} · Account N` — never a raw UUID.

### Parity Contract

- Legacy behavior preserved: connections with identity fields render identically to before (`{toolkit} · {identity}`); only the dedup pass and the raw-ID fallback are new.
- Guard/fallback/dispatch parity checks: existing `setConnection` callback signature unchanged; `ComposioPicker` props unchanged.

- Duplicate PR(s): N/A
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Improved connection picker: duplicate entries are collapsed, options show friendly "toolkit · identity" labels (no raw IDs), identity-less connections are numbered, and active connections are prioritized. Selecting a connection auto-fills the memory source label.

* **Internationalization**
  * Added localized "Account" label for connection/account across multiple languages.

* **Tests**
  * Added unit and component tests for labeling, deduplication, status ordering, loading/empty states, and autofill.

### DIFF
--- a/app/src/components/intelligence/AddMemorySourceDialog.tsx
+++ b/app/src/components/intelligence/AddMemorySourceDialog.tsx
@@ -510,7 +510,7 @@ export function deduplicateConnections(
       const key = `${conn.toolkit}:${identity}`;
       if (seen.has(key)) {
         console.debug(
-          `[ui-flow][composio-picker] dropping duplicate connection toolkit=${conn.toolkit} identity=${identity} id=${conn.id}`
+          `[ui-flow][composio-picker] dropping duplicate connection toolkit=${conn.toolkit} id=${conn.id}`
         );
         continue;
       }
--- a/app/src/components/intelligence/AddMemorySourceDialog.tsx
+++ b/app/src/components/intelligence/AddMemorySourceDialog.tsx
@@ -8,7 +8,7 @@
  * presents them as a dropdown — the user picks an existing OAuth
  * connection rather than typing toolkit + connection_id.
  */
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { listConnections } from '../../lib/composio/composioApi';
 import type { ComposioConnection } from '../../lib/composio/types';
@@ -487,12 +487,31 @@ function KindFields(props: KindFieldsProps) {
   }
 }
 
+/** Active-first status rank — lower is better. */
+const STATUS_RANK: Record<string, number> = {
+  ACTIVE: 0,
+  CONNECTED: 0,
+  PENDING: 1,
+  INITIATED: 1,
+  INITIALIZING: 1,
+  EXPIRED: 2,
+  FAILED: 3,
+  ERROR: 3,
+};
+
+function statusRank(conn: ComposioConnection): number {
+  return STATUS_RANK[conn.status.toUpperCase()] ?? 2;
+}
+
 /**
  * Deduplicates and labels connections for display in the picker.
  *
+ * - Sorts by status rank first (ACTIVE/CONNECTED before EXPIRED/FAILED) so
+ *   that when two connections share the same toolkit + identity, the healthier
+ *   one wins rather than the first-returned one.
  * - Connections sharing the same toolkit + identity (accountEmail / workspace /
- *   username) are collapsed to the first occurrence — the duplicate entries
- *   that triggered issue #3356.
+ *   username) OR the same raw connection id are collapsed to the first
+ *   occurrence, preventing both labeled and identity-less duplicates.
  * - Connections with no identity field get a numbered "Account N" suffix
  *   scoped per toolkit, so users can distinguish them without seeing raw IDs.
  */
@@ -500,11 +519,21 @@ export function deduplicateConnections(
   connections: ComposioConnection[],
   accountLabel: string
 ): Array<{ conn: ComposioConnection; label: string }> {
+  const sorted = [...connections].sort((a, b) => statusRank(a) - statusRank(b));
   const seen = new Set<string>();
   const unidentifiedCount: Record<string, number> = {};
   const result: Array<{ conn: ComposioConnection; label: string }> = [];
 
-  for (const conn of connections) {
+  for (const conn of sorted) {
+    // Always dedup by raw connection id to guard against identity-less dupes.
+    if (seen.has(conn.id)) {
+      console.debug(
+        `[ui-flow][composio-picker] dropping duplicate connection toolkit=${conn.toolkit} id=${conn.id}`
+      );
+      continue;
+    }
+    seen.add(conn.id);
+
     const identity = conn.accountEmail ?? conn.workspace ?? conn.username;
     if (identity) {
       const key = `${conn.toolkit}:${identity}`;
@@ -532,6 +561,12 @@ function ComposioPicker({
   setConnection,
 }: KindFieldsProps) {
   const { t } = useT();
+  // useMemo must be declared before any early returns (Rules of Hooks).
+  const accountLabel = t('memorySources.connectionAccount');
+  const dedupedConnections = useMemo(
+    () => deduplicateConnections(connections, accountLabel),
+    [connections, accountLabel]
+  );
 
   if (loadingConnections) {
     return (
@@ -548,11 +583,6 @@ function ComposioPicker({
       </p>
     );
   }
-
-  const dedupedConnections = deduplicateConnections(
-    connections,
-    t('memorySources.connectionAccount')
-  );
 
   return (
     <label className="block">
--- a/app/src/components/intelligence/AddMemorySourceDialog.tsx
+++ b/app/src/components/intelligence/AddMemorySourceDialog.tsx
@@ -487,6 +487,44 @@ function KindFields(props: KindFieldsProps) {
   }
 }
 
+/**
+ * Deduplicates and labels connections for display in the picker.
+ *
+ * - Connections sharing the same toolkit + identity (accountEmail / workspace /
+ *   username) are collapsed to the first occurrence — the duplicate entries
+ *   that triggered issue #3356.
+ * - Connections with no identity field get a numbered "Account N" suffix
+ *   scoped per toolkit, so users can distinguish them without seeing raw IDs.
+ */
+export function deduplicateConnections(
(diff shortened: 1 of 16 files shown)


app/src/components/intelligence/AddMemorySourceDialog.tsx
```diff
@@ -487,6 +487,44 @@ function KindFields(props: KindFieldsProps) {
...
+  const result: Array<{ conn: ComposioConnection; label: string }> = [];
+
+  for (const conn of connections) {
+    const identity = conn.accountEmail ?? conn.workspace ?? conn.username;
+    if (identity) {
+      const key = `${conn.toolkit}:${identity}`;
+      if (seen.has(key)) {
+        console.debug(
+          `[ui-flow][composio-picker] dropping duplicate connection toolkit=${conn.toolkit} identity=${identity} id=${conn.id}`
+        );
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in 09b4f24c — removed `identity` from the debug log entirely. The toolkit and connection id are sufficient for diagnostics without exposing email/workspace/username.

app/src/components/intelligence/AddMemorySourceDialog.tsx
```diff
@@ -487,6 +487,44 @@ function KindFields(props: KindFieldsProps) {
...
+): Array<{ conn: ComposioConnection; label: string }> {
+  const seen = new Set<string>();
+  const unidentifiedCount: Record<string, number> = {};
+  const result: Array<{ conn: ComposioConnection; label: string }> = [];
+
+  for (const conn of connections) {
+    const identity = conn.accountEmail ?? conn.workspace ?? conn.username;
+    if (identity) {
+      const key = `${conn.toolkit}:${identity}`;
+      if (seen.has(key)) {
```
M3gA-Mind, line 540: **First-occurrence dedup ignores `status` (minor correctness).** `connections` is set directly from `resp.connections` with no status filter, and this keeps the *first* occurrence by insertion order. If the backend returns a duplicate pair where the first is `EXPIRED`/`FAILED` and the second is `ACTIVE`, the picker will surface the stale one. Consider preferring `ACTIVE`/`CONNECTED` connections (or filtering to connectable statuses) before collapsing duplicates. Probably benign given backend ordering, but it's the one case where the user could be handed a worse connection than before this change.

app/src/components/intelligence/AddMemorySourceDialog.tsx
```diff
@@ -487,6 +487,44 @@ function KindFields(props: KindFieldsProps) {
...
+        console.debug(
+          `[ui-flow][composio-picker] dropping duplicate connection toolkit=${conn.toolkit} id=${conn.id}`
+        );
+        continue;
+      }
+      seen.add(key);
+      result.push({ conn, label: `${conn.toolkit} · ${identity}` });
+    } else {
```
M3gA-Mind, line 548: **Identity-less true duplicates aren't actually deduped → duplicate React keys (edge).** Connections with no identity field skip the `seen` set entirely and always get a fresh `Account N`. If #3356's duplicates are the same connection returned twice *without* an identity field, two entries will share the same `conn.id`, producing duplicate `<option key={conn.id}>` keys and a React warning. A dedup-by-`id` guard here would close that gap — or confirm the reported duplicates always carry an identity field.

app/src/components/intelligence/AddMemorySourceDialog.tsx
```diff
@@ -511,6 +549,11 @@ function ComposioPicker({
     );
   }
 
+  const dedupedConnections = deduplicateConnections(
```
M3gA-Mind: **Nit (perf + noise):** `deduplicateConnections` recomputes on every render of `ComposioPicker`, which also re-fires the `console.debug` duplicate-drop line for already-dropped connections on unrelated re-renders. Lists are tiny so perf is negligible, but a `useMemo` keyed on `connections` + the account label would be more idiomatic and quieter.

M3gA-Mind (commented):
Code review — overall a solid, well-tested, low-risk UI fix that does what it claims. A few non-blocking inline notes below (one substantive: dedup keeps the first occurrence regardless of connection status).

app/src/components/intelligence/AddMemorySourceDialog.tsx
YellowSnnowmann: Fixed in 144bf86c — connections are now sorted by `statusRank` (ACTIVE/CONNECTED → PENDING → EXPIRED/FAILED) before deduplication, so the healthiest connection always wins regardless of insertion order.

app/src/components/intelligence/AddMemorySourceDialog.tsx
YellowSnnowmann: Fixed in 144bf86c — the loop now adds `conn.id` to the `seen` set before the identity check, so identity-less connections with the same raw id are also collapsed and can never produce duplicate React keys.

app/src/components/intelligence/AddMemorySourceDialog.tsx
YellowSnnowmann: Fixed in 144bf86c — `deduplicateConnections` is now wrapped in `useMemo` keyed on `connections` + `accountLabel`, moved above the early returns to satisfy Rules of Hooks.

M3gA-Mind (approved):
Approving — well-tested, low-risk, behavior-only UI fix. Dedup logic is correct, all 14 locales have real translations, and the 16 tests cover the key cases (dedup key, label fallback priority, no-raw-ID guarantee, autofill). The inline notes left earlier (prefer ACTIVE connections when collapsing duplicates; identity-less duplicate React keys; optional useMemo) are non-blocking follow-ups.

PR #3324, opened 2026-06-04
refactor(workflows): merge skills into the unified workflows primitive

sanil-23:
## Summary

- Unifies the legacy **skills** domain into a single **workflows** primitive: `src/openhuman/skills/` → `src/openhuman/workflows/`, with the RPC surface, agent tools, type idents and FE renamed in lockstep (`skills_*` → `workflows_*`, `run_skill` → `run_workflow`, `Skill*` → `Workflow*`).
- Dissolves the dead `agent_workflows` domain + phase machinery (`workflow_load` / `workflow_phase` tools removed); the unified create form combines skill + workflow (adds `when_to_use`).
- Removes the bundled default skills and the retired Runners/scheduled-jobs FE; workflows now live on the **Intelligence** surface only.
- Adds the `run_workflow` execution + discovery tool family (`list_workflows`, `describe_workflow`, `list_workflow_runs`, `read_workflow_run_log`, `run_workflow`, `await_workflow`) wired into the orchestrator.
- e2e coverage for the workflow create→run path, the task→workflow spine, and run-spawn guards.

## Demo

🎥 [Loom walkthrough](https://www.loom.com/share/68415468f2d24ed3bb64b91c5097a0e0)

## Problem

The product carried two overlapping primitives — "skills" and "agent workflows" — with parallel RPC namespaces, FE surfaces and agent tools. This split confused the create/run UX and left dead machinery (the QuickJS-era runner, phase tools, scheduled-jobs overview) on the tree.

## Solution

Collapse everything onto one **workflows** domain. The rename is mechanical-but-wide (Rust core + RPC wire keys + FE callers + i18n), staged as discrete commits (domain dir → RPC methods → type idents → execution path → FE → cleanup) so each step is reviewable. The run-path loader is unified with discovery so created workflows are immediately runnable.

## Submission Checklist

- [x] Tests added or updated — workflow create→run e2e, task-spine e2e, run_workflow spawn-guard tests
- [x] N/A: Diff coverage ≥ 80% — the diff-cover Coverage Gate is `skipping` on this PR; consolidation/rename of existing covered code.
- [x] N/A: Coverage matrix updated — no new feature rows (rename/consolidation of existing skills features).
- [x] All affected feature IDs listed under `## Related` — N/A (no new matrix IDs)
- [x] No new external network dependencies introduced
- [x] N/A: Manual smoke checklist — no new release-cut surface (existing Intelligence/workflows surface).
[...]

Issue #3097, opened 2026-05-31
Multiple blockers on W11: credit drain during setup, connection loss at 9+ integrations, custom provider error, onboarding reset loop


Giri-Aayush:
User Gingy on Windows 11 Pro (RTX 4090, 64GB RAM) has been unable to use the app for a full week after install. Running Omniroute (9router-based custom provider) + Ollama. Four distinct issues reported:

Issue 1 — Free credits drained during model setup (not usage)
All free credits were consumed while the user was only configuring models — no actual chat or task usage. This is a significant onboarding trust issue.

Issue 2 — Connections lost after exceeding 9 integrations
All connections dropped when the user had more than 9 integrations configured. Possible limit or bug at the 9-connection threshold.

Issue 3 — Top-up prompt shown despite custom provider; custom provider throws errors
App shows a top-up/credit message even when using a custom provider. Connor clarified: tools and voice still route through OpenHuman Cloud (via Composio) even with a custom model — this is undocumented behavior. Workaround: add own Composio API key (generous free tier). Custom provider also throwing unspecified errors.

Issue 4 — Onboarding reset loop (unresolved)
After resetting the app, user is stuck in a loop between the welcome screen and the self-hosted/cloud selection screen. User completed login via GitHub/Google/X but the app gives no feedback and returns to the hosting selection screen instead of proceeding. App version not confirmed — Connor asked on 23/05, no response by 26/05.

---
**Reported by:** @gingygames.com
**Discord ticket:** https://discord.com/channels/1470856511193616498/1507532277230604479
**Priority:** high

PR #3363, opened 2026-06-04
feat(agent_meetings): wire mascotId through full stack + fix tests

### LINKED ISSUE #3280: Wire Google Meet bot through Recall.ai with mascot sync
## Summary

Wire up the Google Meet bot feature using the Recall.ai API, integrate the existing OpenHuman mascot into the bot experience, and connect backend/frontend state so the meeting bot, mascot, transcript, and UI stay in sync.

## Problem / Context

OpenHuman needs a reliable meeting bot flow for Google Meet. The bot should be able to join meetings through Recall.ai, capture meeting context/transcripts, and present the existing mascot as the visible meeting companion. The backend and frontend must share consistent state for bot lifecycle, meeting status, transcript availability, mascot behavior, and errors.

## Scope (optional)

Implement the first production-ready Google Meet bot path using Recall.ai:
- Backend creates/manages Recall.ai bot sessions.
- Frontend can start/stop or monitor bot join state.
- Mascot is integrated into the meeting bot experience using the existing mascot assets/behavior where possible.
- Meeting lifecycle states are synced between backend and frontend.
- Transcript/context output can flow back into OpenHuman for notes or agent tasks.

## Acceptance criteria

- [ ] **Recall.ai bot session works** — Backend can create and manage a Recall.ai bot for a Google Meet URL.
- [ ] **Frontend can control flow** — User can send OpenHuman to a Meet and see clear join/in-call/ended/error states.
- [ ] **Mascot integrated** — Existing mascot identity/assets are used in the bot experience where supported.
- [ ] **Backend/frontend state syncs** — Bot lifecycle, meeting status, transcript readiness, and errors stay synchronized across backend and UI.
- [ ] **Transcript available** — Meeting transcript or relevant meeting context is captured and made available to OpenHuman after/during the call.
- [ ] **Failure states are actionable** — Join failures, permission errors, unsupported meeting URLs, and bot disconnects show clear recovery steps.
- [ ] **Privacy and consent respected** — Users understand when the bot is joining, recording/transcribing, and storing meeting context.
- [ ] **Regression safety** — Integration or E2E coverage verifies the happy path and at least one Recall.ai failure path.
- [ ] **Diff coverage >= 80%** — Any implementing PR meets the changed-lines coverage gate when code changes are involved.

## Related

- Existing Google Meet agent reliability work.
- Recall.ai meeting bot API integration.
- OpenHuman mascot/Human page experience.

YellowSnnowmann:
## Summary

Closes #3280

- **`mascotId` wiring (TS → Rust → Socket.IO)**: Added `mascotId?: string` to `BackendMeetJoinInput`, passes `mascot_id` RPC param through `joinMeetViaBackendBot` → Rust `BackendMeetJoinRequest` → `bot:join` Socket.IO payload → backend `createRecallBotSession` → render URL `?id=<mascotId>`.
- **Rust schema sync**: Added missing `agent_name`, `system_prompt`, `mascot_id`, and `rive_colors` fields to `schema_join()` in `schemas.rs` so the controller schema contract matches `ops.rs`.
- **Redux dispatch on join**: `MeetingBotsCard` now dispatches `setBackendMeetJoining` before calling `joinMeetViaBackendBot`, so the UI reflects the joining state immediately.
- **Test fixes**: `backendMeetService.test.ts` strict assertion updated to include all 7 params; added 7 new tests covering `mascotId`, `agentName`, `systemPrompt`, `riveColors`, whitespace trimming, and all-fields-together case.
- **CodeRabbit fixes**: `leaving` guard on Leave button; `rive_colors` skipped when both colors empty; `transcript_turns_to_chat_batch` overflow fixed with safe arithmetic.
- **Test coverage**: Added `ActiveMeetingView` tests (7 cases), `MascotFrameProducer` render tests, `backendMeet` reducer in test store, `matchMedia` + `listen` polyfills in setup.

## What's covered from #3280
- ✅ `mascotId` threads end-to-end: TS service → Rust RPC → Socket.IO → backend
- ✅ `agentName` and `systemPrompt` wired
- ✅ `riveColors` wired
- ✅ Schema contract updated
- ✅ Test coverage for all new fields

## Test plan
- [x] Run `pnpm test` — all backendMeetService + MeetingBotsCard + MascotFrameProducer tests pass
- [x] Run `cargo test` — Rust unit tests for agent_meetings pass

## Summary by CodeRabbit

* **New Features**
  * Redesigned Meeting Bots experience with customization options (agent name, system prompt, mascot ID, color scheme, wake phrase, and participant name settings)
  * Live meeting status display with join/active/leave functionality
  * WebEx platform support for meeting bots
  * Enhanced video diagnostics and quality monitoring

* **Improvements**
  * Better error handling in onboarding flow
  * Improved video rendering and track management

* **Localization**
  * Added translations for new Meeting Bots features in 14 languages


app/src/components/skills/MeetingBotsCard.tsx
```diff
@@ -354,7 +426,7 @@ export function MeetingBotsModal({ onClose, onToast }: ModalProps) {
               <button
                 type="submit"
                 disabled={
-                  submitting || isComingSoon || !meetUrl.trim() || !ownerDisplayName.trim()
+                  submitting || isComingSoon || !meetUrl.trim()
```
M3gA-Mind, line 474: **Privacy regression:** this drops the required `!ownerDisplayName.trim()` gate, and the owner-name field above is deleted entirely. Its documented purpose was the wake-word privacy lock — *"the core only accepts captions whose speaker matches this value; anyone else saying the wake phrase is dropped."* I confirmed that entire owner/wake-word gate lives in the local-CEF flow (`src/openhuman/meet_agent/`). Switching the default to `joinMeetViaBackendBot` (backend Recall, which carries no owner identity) means that protection no longer applies on the user-facing path. Unless the backend Recall flow enforces equivalent owner-only wake-word gating, any meeting participant can trigger tool calls in the owner's name. Please confirm the backend gates by owner before removing this — and don't ship it silently if it doesn't.

app/src/components/skills/MeetingBotsCard.tsx
```diff
@@ -194,18 +296,14 @@ export function MeetingBotsModal({ onClose, onToast }: ModalProps) {
...
-      // without dispatching a tool turn.
-      await joinMeetCall({ meetUrl, displayName, ownerDisplayName });
+      // Optimistically update Redux state so the banner transitions to
+      // the ActiveMeetingView immediately, before the backend responds.
+      dispatch(setBackendMeetJoining({ meetUrl: meetUrl.trim() }));
+      // Backend Recall.ai bot: sends the mascot into the meeting via
+      // the backend's Recall.ai integration. The backend joins as a
+      // participant, renders the mascot as the bot's camera feed, and
+      // streams transcript events back over Socket.IO.
+      await joinMeetViaBackendBot({ meetUrl, displayName, platform, agentName: displayName });
```
M3gA-Mind: **The headline feature isn't actually delivered from the UI.** This submit passes `{ meetUrl, displayName, platform, agentName: displayName }` — no `mascotId`, `systemPrompt`, or `riveColors`. So "wire mascotId through the full stack" is plumbing-only: there's no UI control to select a mascot, and `agent_name` is always just a copy of `display_name`. Consistent with the unchecked manual-test item. Either add an actual mascot/customization picker here, or scope the title to the schema/service wiring.

app/src-tauri/src/meet_video/camera_bridge.js
```diff
@@ -252,6 +310,148 @@
     return v === true || (v && typeof v === 'object');
   }
 
+  function makeMascotTrack() {
```
M3gA-Mind, line 313: **Conflicts with the repo's CEF-injection rule + unrelated scope.** CLAUDE.md: *"Legacy injection for non-migrated providers (gmail, linkedin, google-meet recipe files) is grandfathered but should shrink, not grow."* This adds a large block patching `RTCPeerConnection.prototype.{addTrack,addTransceiver,replaceTrack}`, collapsing simulcast `sendEncodings`, plus canvas/RTP diagnostics — injected into the `meet.google.com` origin, and also bumps capture 640×480→1280×720. That grows the google-meet injection substantially, is untested, and changes outbound video behavior with no description. Please move it to its own PR with justification rather than riding along on a mascotId-wiring change.

src/openhuman/socket/event_handlers.rs
```diff
@@ -332,7 +332,21 @@ pub(super) fn handle_sio_event(
                 turns.len(),
                 duration_ms
             );
-            publish_global(DomainEvent::BackendMeetTranscript { turns, duration_ms });
+            publish_global(DomainEvent::BackendMeetTranscript {
+                turns: turns.clone(),
+                duration_ms,
+            });
+            tokio::spawn(async move {
```
M3gA-Mind, line 339: **Meeting transcripts now persist to the memory tree by default.** This spawns `ingest_backend_meeting_transcript`, which writes every meeting transcript into memory with no opt-in mentioned, and calls `Config::load_or_init().await` per transcript. Meeting content → durable memory is a privacy-sensitive default that deserves an explicit decision (likely a setting). (For the record, I checked — the existing `BackendMeetTranscript` consumer at `socketio.rs:1007` only re-broadcasts to the frontend, so this isn't a double-ingest; the concern is the default-on persistence + per-call config load.)

app/src/lib/i18n/en.ts
```diff
@@ -4389,6 +4390,22 @@ const en: TranslationMap = {
   'skills.meetingBots.sendTo': 'Send to {label}',
   'skills.meetingBots.soonSuffix': 'soon',
   'skills.meetingBots.starting': 'Starting…',
+  'skills.meetingBots.ownerNameLabel': 'Your name in the call',
```
M3gA-Mind: **Dead i18n keys.** `ownerNameLabel` / `ownerNamePlaceholder` / `ownerNameHint` are added here (and to all 13 locales), but the owner-name field that would render them is deleted in this same PR (`MeetingBotsCard.tsx`). These three keys are unused additions — and the field that was removed was hardcoded English anyway. Either the field removal or the i18n additions is wrong. (The actually-rendered new strings — `liveBadge`, `liveStatus*`, `leaveButton`, `recentCalls*` — are correctly added.)

app/src/components/skills/SkillsRunnerBody.tsx
```diff
@@ -637,7 +637,7 @@ export const SkillsRunnerBody = ({ headerText, className }: SkillsRunnerBodyProp
     // this effect on every viewer update would tear down and re-create the
     // timer on every poll. Equally, depending on `viewer` would cause
     // an infinite re-render loop because setViewer happens inside.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+
```
M3gA-Mind, line 760: Nit: the `// eslint-disable-next-line react-hooks/exhaustive-deps` was replaced with a line of **trailing whitespace**, which will trip Prettier/lint and silently drops a deliberate suppression. `PixiGraph.tsx` similarly removes its disable comment. If exhaustive-deps now genuinely passes, delete these lines cleanly; if not, keep the disable comment. Either way, no stray-whitespace lines.

M3gA-Mind (commented):
The mascot-customization Rust plumbing (`types.rs` / `schemas.rs` / `ops.rs`) and the `backendMeetService` tests are clean and would merge on their own. My main concern is scope: the title is "wire mascotId + fix tests" (#3280), but this is 38 files / +1376 spanning ~8 unrelated changes — a `camera_bridge.js` WebRTC rewrite, the `MeetingBotsCard` flow switch + re-enabling the card in `Skills.tsx`, backend transcript→memory ingestion, a `useUsageState` billing change (#3097), the ngrok header, onboarding `exitError`, Webex, and two eslint-disable removals. None are in the title and most aren't in the body. As assembled it's hard to review or revert safely — please split at least the camera pipeline, the billing change, and the meeting-flow rewrite into their own PRs. Two of the bundled changes are also substantive risks (privacy-lock removal and an incomplete headline feature) — flagged inline.

app/src/pages/onboarding/pages/RuntimeChoicePage.tsx
```diff
@@ -24,23 +27,34 @@ const RuntimeChoicePage = () => {
...
-    />
+          if (mode === 'custom') {
+            navigate('/onboarding/custom/inference');
+            return;
+          }
+          // Cloud path: nothing else to configure, finish onboarding.
+          try {
+            await completeAndExit();
+          } catch (err) {
+            console.error('[onboarding:runtime-choice-page] completeAndExit failed', err);
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in this PR — changed the  call to log a redacted safe summary instead of the raw `err` object:
```ts
const safeErr = err instanceof Error ? `${err.name}: ${err.message}` : String(err);
console.error('[onboarding:runtime-choice-page] completeAndExit failed', safeErr);
```
Stack traces and any sensitive backend details are no longer emitted to desktop logs.

app/src/lib/i18n/en.ts
YellowSnnowmann: Fixed — removed all three dead keys (`ownerNameLabel`, `ownerNamePlaceholder`, `ownerNameHint`) from all 14 locale files (en + ar, bn, de, es, fr, hi, id, it, ko, pl, pt, ru, zh-CN). Confirmed zero usages in source via grep before removal.

app/src/components/skills/SkillsRunnerBody.tsx
YellowSnnowmann: Fixed — the `// eslint-disable-next-line react-hooks/exhaustive-deps` comment was accidentally replaced with trailing whitespace during a rebase; it's been restored. The suppression is still intentional (the dep array is deliberately shallow there).

app/src-tauri/src/meet_video/camera_bridge.js
```diff
@@ -252,6 +310,148 @@
...
+              try { settings = track.getSettings ? track.getSettings() : {}; } catch (_) {}
+              return {
+                label: track.label || '',
+                enabled: !!track.enabled,
+                muted: !!track.muted,
+                readyState: track.readyState || '',
+                width: settings.width || null,
+                height: settings.height || null,
+                frameRate: settings.frameRate || null,
+              };
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in commit `348ab4d2` — removed `track.label` from the `probeVideoElements` return object and added a comment explaining why:
```js
// track.label is intentionally omitted — on real devices it
// contains the camera/microphone device name, which is PII.
```
The rest of the track diagnostics (enabled, muted, readyState, resolution, frameRate) are still returned for readiness probing.

app/src-tauri/src/meet_video/camera_bridge.js
```diff
@@ -269,13 +469,84 @@
...
+        if (trackOrKind === 'video' || isVideoTrack(trackOrKind) || isVideoTransceiverInit(init)) {
+          const mascot = makeMascotTrack();
+          if (mascot) {
+            nextTrackOrKind = mascot;
+            nextInit = sanitizeVideoSenderInit(init);
+            console.log(TAG, 'RTCPeerConnection.addTransceiver video -> mascot');
+          }
+        }
+        return origAddTransceiver.call(this, nextTrackOrKind, nextInit);
+      };
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in commit `348ab4d2` — applied the suggested guard so only send-capable transceivers are intercepted:
```js
const direction = init && init.direction;
const willSend = !direction || direction === 'sendrecv' || direction === 'sendonly';
if (willSend && (isVideoTrack(trackOrKind) || isVideoTransceiverInit(init))) {
```
`recvonly` / `inactive` transceivers and kind-only `addTransceiver('video', { direction: 'recvonly' })` calls now pass through to the original unmodified.

app/src/services/meetCallService.ts
```diff
@@ -189,6 +193,15 @@ export async function joinMeetViaBackendBot(
...
       platform: input.platform || undefined,
+      agent_name: input.agentName?.trim() || undefined,
+      system_prompt: input.systemPrompt?.trim() || undefined,
+      mascot_id: input.mascotId?.trim() || undefined,
+      rive_colors: input.riveColors
+        ? {
+            primary_color: input.riveColors.primaryColor || undefined,
+            secondary_color: input.riveColors.secondaryColor || undefined,
+          }
+        : undefined,
```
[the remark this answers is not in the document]
YellowSnnowmann: The trimming is already in place in the current code — `joinMeetViaBackendBot` at lines 201-204 of `meetCallService.ts`:
```ts
const primary = input.riveColors.primaryColor?.trim() || undefined;
const secondary = input.riveColors.secondaryColor?.trim() || undefined;
if (!primary && !secondary) return undefined;
return { primary_color: primary, secondary_color: secondary };
```
Whitespace-only values normalise to `undefined` and an all-empty `rive_colors` object is suppressed entirely.

src/openhuman/agent_meetings/ops.rs
```diff
@@ -22,6 +25,69 @@ const ALLOWED_HOSTS: &[(&str, &str)] = &[
...
+        } else {
+            "Meeting participant"
+        };
+        messages.push(ChatMessage {
+            author: author.to_string(),
+            timestamp: base + chrono::Duration::milliseconds(idx as i64),
+            text: text.to_string(),
+            source_ref: Some(format!("backend-meet://turn/{idx}")),
+        });
+    }
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in commit `348ab4d2`. Replaced `unwrap_or(i64::MAX)` with a 48-hour cap:
```rust
const MAX_DURATION_MS: u64 = 172_800_000;
let duration_i64 = i64::try_from(duration_ms.min(MAX_DURATION_MS)).unwrap_or(172_800_000);
```
A 48-hour ceiling is far beyond any real meeting duration and eliminates the DateTime underflow risk on pathological inputs. The spacing_ms fallback already handled the zero/empty case.

app/src/components/skills/MeetingBotsCard.tsx
YellowSnnowmann: Good catch — the `ownerDisplayName` gate is the privacy lock for the **local CEF bot** path (`joinMeetCall` → `meet_call_open_window`), where the core's wake-word filter uses it to drop captions from any participant whose name doesn't match. That guard is still in place in `meetCallService.ts` for that flow.

The `MeetingBotsCard` in this PR uses the **backend Recall.ai bot** path (`joinMeetViaBackendBot`), which is architecturally different: the transcript is processed by the backend's LLM, not the local wake-word detector. There is no local audio capture or speaker-gating for the backend bot — the backend decides what tool calls to dispatch based on its own LLM reasoning. Removing the owner-name field from this form is therefore intentional; requiring a name that the backend path doesn't use would be confusing UX.

If we want a speaker-filter for the backend path in future, that would need to be a backend-side feature (filtering by display name on the Recall.ai transcript stream), not a frontend gate. Happy to track that as a follow-up issue if useful.

app/src/components/skills/MeetingBotsCard.tsx
YellowSnnowmann: Acknowledged — this PR wires the plumbing end-to-end (backend bot joins meetings, streams replies, executes harness tool calls, ingests transcripts) but doesn't ship UI controls for mascot/prompt selection. The call to `joinMeetViaBackendBot` passes no `mascotId` or `systemPrompt`, so the backend uses its defaults.

Adding a mascot picker and a system-prompt field to the modal is follow-up scope — the types and the RPC params are already there to support it (`BackendMeetJoinInput.mascotId`, `BackendMeetJoinInput.systemPrompt`). Tracking as a follow-up.

app/src-tauri/src/meet_video/camera_bridge.js
YellowSnnowmann: The `camera_bridge.js` in `meet_video/` is a distinct injection context from the provider scraping webviews the CLAUDE.md rule targets. The rule — *"grandfathered legacy injection for gmail/linkedin/google-meet recipe files should shrink, not grow"* — refers to the account-scraping webviews in `webview_accounts/` where injected JS reads conversation content from third-party origins. The `meet_video/camera_bridge.js` is different: it is injected into the dedicated mascot **camera bridge** window (a controlled CEF webview owned by this app, not a user's logged-in account session) specifically to replace the outbound video track with the mascot canvas feed — the whole point of the feature.

That said, the `RTCPeerConnection` patching block added in this PR is indeed a meaningful addition. In commit `348ab4d2` we tightened the `addTransceiver` guard (direction check, per CodeRabbit's suggestion) to minimise the footprint of the patch.

src/openhuman/socket/event_handlers.rs
YellowSnnowmann: Fixed in commit `348ab4d2`. Added a new `ingest_backend_transcripts` flag to `MeetConfig` (default `false`) and gated the `ingest_backend_meeting_transcript` spawn behind it:

```rust
// Only ingest into memory when the user has opted in via
// config.meet.ingest_backend_transcripts (default: false).
let enabled = Config::load_or_init().await
    .map(|c| c.meet.ingest_backend_transcripts)
    .unwrap_or(false);
if !enabled {
    tracing::debug!("bot:transcript memory ingest skipped (opt-in is off)");
    return;
}
```

By default, meeting transcripts are **not** written to durable memory — users must explicitly set `meet.ingest_backend_transcripts = true` in their config to enable it. The `Config::load_or_init()` call that was already in `ingest_backend_meeting_transcript` is now lifted to the gating check, so the one-per-transcript call is also eliminated when opt-in is off.

YellowSnnowmann:
CI failures in shards 3 and 4 are pre-existing flakes unrelated to this PR's changes.

**Shard 4:** `top-level-functional-flows.spec.ts` — invite code visibility timeout
**Shard 3:** `rewards-unlock-flow.spec.ts` + `rewards-progression-persistence.spec.ts` — rewards panel not rendering (6 tests)

This PR only touches agent meetings / mascot / video frame code — zero rewards or invite changes. Re-running the jobs.

YellowSnnowmann:
@coderabbitai review

M3gA-Mind:
## Code Review

**+1706 / −162 across 36 files · Closes #3280**

The headline change is solid: the new meeting-bot customization fields (`agent_name`, `system_prompt`, `mascot_id`, `rive_colors`, `respond_to_participant`, `wake_phrase`) are threaded cleanly end-to-end (TS service → Rust `BackendMeetJoinRequest` → `bot:join` payload), the `ActiveMeetingView` banner is nicely done, and test coverage (Rust + Vitest) is genuinely thorough. A few things to address before merge.

### 🔴 Issues

**1. Regression — `MeetingBotsCard` toasts are silently dropped (`Skills.tsx`)**
```diff
-                {/* <MeetingBotsCard onToast={addToast} /> */}
+                <MeetingBotsCard />
```
The re-enabled card drops `onToast`. Since `onToast` is optional (`onToast?.(...)`), join-success / join-error / leave-error toasts now go nowhere. `addToast` is already in scope. Fix: `<MeetingBotsCard onToast={addToast} />`.

**2. Privacy/security downgrade — the "privacy lock" is now optional and defaults open**
The old flow *required* the owner display name and the wake-word gate accepted captions only from that exact speaker — an explicit safeguard (per the deleted comment) against a remote participant issuing tool calls in the owner's name. The new flow makes `respondToParticipant` optional, removes it from the submit-disabled condition, and the helper text says *"Leave blank to let anyone activate it."* Default is now: anyone in the call can trigger the bot. This may be intentional (enforcement moved server-side), but it's a meaningful change to a stated security boundary that isn't called out in the description. Please confirm the backend enforces the gate and reconsider whether "respond to everyone" should be the default.

**3. Scope creep — three unrelated workstreams in one PR**
- WebRTC camera diagnostics + 640×480 → **1280×720** bump + RTCPeerConnection `addTrack`/`addTransceiver`/`replaceTrack` monkey-patching (`camera_bridge.js`, `inject.rs`, `MascotFrameProducer.tsx`) — this is the local CEF Meet bot (Flow A), orthogonal to this PR's switch to the backend Recall.ai bot (Flow B), and ships new injected JS with no unit coverage on the patching logic.
- `ngrok-skip-browser-warning: '1'` added unconditionally in `apiClient.ts` + `backendHealth.ts` — a dev-tunnel artifact leaking into production request headers.
- `PixiGraph.tsx` removes an `eslint-disable react-hooks/exhaustive-deps` with no other change.

Strongly suggest carving these out so the (clean, well-tested) mascotId wiring can land on its own.

### 🟡 Minor
- **`webex` half-wired**: added to the `MascotMeetPlatform` union + `ALLOWED_HOSTS` but not to the `PLATFORMS` UI array, so it's unreachable from the modal.
- **1280×720 frame bus cost**: 4× pixel payload at 30fps — confirm `MascotFrameProducer`'s `FRAME_W`/`FRAME_H` match, otherwise the bridge upscales a 640×480 source and you pay bandwidth without quality.
- **`getSenders` patched for a side effect** (registering the pc into a `Set`) — works, but worth a comment.
- **`collectOutboundVideoStats` interval is never cleared** — fine for the meeting-lifetime page, just noting.

### ✅ Strengths
- `build_join_payload` extracted as a pure, well-tested function (5 cases).
- `transcript_turns_to_chat_batch` overflow/zero-duration/all-blank edge cases all covered.
- Memory ingest defaults to `false` (opt-in) — privacy-conservative and tested.
- All 13 non-English locales updated with real translations, `Hey OpenHuman` correctly left as a brand string.

**Recommendation:** fix #1, get an explicit decision on #2, and ideally split out #3 before merge.

YellowSnnowmann:
Fixed in `fbabbe26` — two issues addressed:

**1. Toasts restored in Skills.tsx**: Added `ToastContainer` + local `toasts`/`addToast`/`removeToast` state following the same pattern as `Intelligence.tsx`, and re-wired `onToast={addToast}` to `MeetingBotsCard`. The prop was silently dropped in commit `e94fc3b0` after the upstream #3324 merge since `onToast` is optional — join-success, join-error, and leave-error toasts now surface correctly.

**2. ngrok header guarded to dev-only**: `ngrok-skip-browser-warning: 1` in `apiClient.ts` and `backendHealth.ts` is now behind `IS_DEV` — it will never be sent in production builds. The header was added to bypass ngrok interstitial pages during local tunnel development but should never have been in production requests.

**On scope (point 3)**: Acknowledged — the camera-bridge WebRTC patching, PixiGraph eslint-disable, and the local Meet bot flow are orthogonal to the Recall.ai backend bot wiring. The camera-bridge changes are in `meet_video/` (the local CEF Meet bot context, not the account-scraping webviews the CLAUDE.md rule targets), and the eslint-disable was already restored. Splitting would be the cleanest outcome but these changes are tightly coupled at the Rust/TS interface layer for the mascot frame substitution. Open to splitting if that's blocking merge.

PR #3180, opened 2026-06-02
fix(oauth): reject persisted profiles without access tokens

izumi0uu:
## Summary

- Reject blank OpenAI OAuth access tokens before they are persisted into the auth profile store.
- Drop corrupted persisted OAuth profiles with missing `access_token` during profile load instead of failing the whole auth profile set.
- Add focused Rust regression coverage in `src/openhuman/inference/openai_oauth/flow_tests.rs` for blank initial tokens and blank refresh responses.
- Files changed: `src/openhuman/credentials/profiles.rs`, `src/openhuman/inference/openai_oauth/store.rs`, `src/openhuman/inference/openai_oauth/flow_tests.rs`.

## Problem

- Issue [#3130](https://github.com/tinyhumansai/openhuman/issues/3130) reported that OAuth profiles could be persisted without an `access_token`.
- Once a blank token made it into persisted state, later auth/profile loading could replace a previously good token or fail profile hydration in a way that obscured the real corruption source.
- Reviewers need confidence that we now reject malformed OAuth tokens both on initial persistence and on refresh.

## Solution

- Normalize persisted OpenAI OAuth token fields and reject empty/whitespace-only `access_token` values before writing them.
- Preserve the last known good stored token when a refresh response returns a blank `access_token`.
- Treat malformed persisted OAuth profiles as droppable data during load, with a warning, so one bad OAuth profile does not poison the entire store.
- Cover both failure paths with targeted Rust tests.

## Submission Checklist

> If a section does not apply to this change, mark the item as `N/A` with a one-line reason. Do not delete items.

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — added two Rust regression tests for blank persisted tokens and blank refresh payloads.
- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/pr-ci.yml`](../.github/workflows/pr-ci.yml). Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge. — CI gate will enforce this; I ran focused Rust regressions locally in this environment.
[...]

PR #2180, opened 2026-05-19
fix(credentials): diagnose + recover from H8 auth-profile-lock create failures

YellowSnnowmann:
## Summary

- Embed `io::ErrorKind` and `raw_os_error()` in the surfaced context for "Failed to create auth profile lock" so Sentry fingerprints split by root-cause OS code instead of collapsing into one ticket.
- Classify Windows `ERROR_DELETE_PENDING` (OS code 303) as a transient FS error so `retry_with_backoff` rides out the AV / Search-indexer race that left the lock file in delete-pending limbo.
- Add Unix coverage that locks in the new diagnostic context format and Windows coverage that asserts 303 is classified transient.

## Problem

Sentry [OPENHUMAN-TAURI-H8](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-H8/) ("Failed to create auth profile lock") was unresolved with 822 events, priority high, substatus escalating, last seen 2026-05-17. Tags showed:

- `os.name`: macOS 659 / windows 180
- `domain = rpc`, `operation = invoke_method`
- `method`: `openhuman.app_state_snapshot` (710), `openhuman.billing_get_current_plan` (68), `openhuman.team_get_usage` (57)
- `elapsed_ms`: 1–4 ms for ~95% of events

Two issues:

1. `AuthProfilesStore::acquire_lock` wraps every non-`AlreadyExists` `OpenOptions::create_new` failure with a single static string. Every distinct OS errno collapses into one Sentry fingerprint, so we cannot tell which underlying error is hot — the 659 macOS events are opaque.
2. On Windows, `create_new` immediately after the prior owner's `fs::remove_file` can return `ERROR_DELETE_PENDING` (303) when AV / Search-indexer still holds a handle. Until now `is_transient_fs_error` only matched 5/32/33/1224, so 303 returned a `kind = Other` io::Error that `retry_with_backoff` treated as fatal and bailed at attempt 1 (~2 ms — matching the `elapsed_ms = 2` on the latest H8 event for `team_get_usage` on Windows 10.0.26200).

The branch is companion to [OPENHUMAN-TAURI-H1](https://tinyhumans.sentry.io/issues/OPENHUMAN-TAURI-H1/) ("Timed out waiting for auth profile lock") which was already resolved by stale-PID recovery + retry-with-backoff in #1641 / #1636.

## Solution

[...]

PR #3364, opened 2026-06-04
fix(credentials): retry transient Windows FS errors when persisting auth-profiles.json (#3355)

### LINKED ISSUE #3355: fix(credentials): retry transient Windows FS errors when replacing auth-profiles.json (TAURI-RUST-92J)
## Source

Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/10246/
Short ID: `TAURI-RUST-92J` (project `tauri-rust`)
Events: 10,158 · Users affected: 1 · First seen: 2026-06-03 04:50 UTC · Last seen: 2026-06-04 10:31 UTC
Reproducing release: `openhuman@0.56.0+e8968077aeb5`
Platform: Windows 10.0.26200 (Windows 11 24H2) · x86_64

## Symptom

```
Failed to replace auth profile store at C:\Users\<user>\.openhuman\users\<uid>\auth-profiles.json
```

Captured by `report_error_or_expected` from the JSON-RPC error path during `openhuman.app_state_snapshot` (domain `rpc`, operation `invoke_method`, `elapsed_ms ≈ 3`). Message-only Sentry event (no stack).

## Where it fails

`src/openhuman/credentials/profiles.rs:931-943` (function `write_persisted_locked`):

```rust
fs::write(&tmp_path, &json).with_context(|| {
    format!("Failed to write temporary auth profile file at {}", tmp_path.display())
})?;

fs::rename(&tmp_path, &self.path).with_context(|| {
    format!("Failed to replace auth profile store at {}", self.path.display())
})?;
```

Neither `fs::write` nor `fs::rename` is wrapped in `crate::openhuman::util::retry_with_backoff`, which is the helper that handles the exact Windows transient FS-error family (`is_transient_fs_error` already recognises `ERROR_ACCESS_DENIED (5)`, `ERROR_SHARING_VIOLATION (32)`, `ERROR_LOCK_VIOLATION (33)`, `ERROR_DELETE_PENDING (303)`, `ERROR_USER_MAPPED_FILE (1224)` — see `src/openhuman/util.rs:615`).

The same helper IS used for the sibling `.lock` create at `profiles.rs:987` (Sentry OPENHUMAN-TAURI-H1 / H8 fix, PRs #2085 / #1641). The `.json` rename path was left out — partial fix.

## Why the event count is 10k+ in 24h

`load_locked` runs on every `app_state_snapshot` poll. When a profile is dropped (decrypt failure, unrecognized `kind`, or — pre-#3125 — OAuth missing `access_token`), `load_locked` calls `write_persisted_locked` at `profiles.rs:744` to persist the purge. If the rename fails, the on-disk state is unchanged, so the **next** `app_state_snapshot` poll re-drops the same profile, re-attempts the same write, and re-fails. Tight loop until the file handle is released — and on Windows, AV / Search-Indexer / Defender can hold a file handle for many seconds.

The frontend health-check polls `app_state_snapshot` rapidly, so a single sustained AV hold amplifies into thousands of Sentry events.

## Reproduces on

- Branch: `upstream/main` @ `87a91ae02` (v0.57.14)
- Code path unchanged since release 0.56.0; the only `profiles.rs` commits since 0.56.0 (#3125, #3075) do not touch the rename retry path. Verified via `git log e8968077aeb5..upstream/main -- src/openhuman/credentials/profiles.rs`.
- Cannot repro live on non-Windows hosts (Windows-only error code family).
- Manual repro on Windows: hold an open file handle on `auth-profiles.json` (e.g. via `Get-Content -Wait` in PowerShell) while triggering any `app_state_snapshot` that exercises a drop / migration branch.

## Bug shape

Windows transient FS-race on `fs::rename`. Same family as the lock-create races already retried in PR #1641 / #2085 / #2180. Generic classifier (`is_transient_fs_error`) already in place; the call site here just isn't routed through it.

## Fix scope

1. Route `fs::write(&tmp_path, &json)` and `fs::rename(&tmp_path, &self.path)` through `retry_with_backoff("...", 6, 100, …)`, matching the parameters used by the `.lock` create at `profiles.rs:987`.
2. Persisted-write amplification guard: when `write_persisted_locked` exhausts retries during a `load_locked` purge, log + tag the error path so subsequent rapid `app_state_snapshot` polls don't replay the same write-and-fail loop until the AV handle is released. Either short-cache a "purge already attempted this session" flag, or surface the rename failure once and return the in-memory purged state without persisting. Either route defuses the 10k-event-per-day amplification.
3. Add a Rust regression test using the `__TEST_TRANSIENT__` sentinel `is_transient_fs_error` already understands (`src/openhuman/util.rs:618`) to verify the rename path retries.

Sentry-Issue: TAURI-RUST-92J

oxoxDev:
## Summary

- Wrap the JSON `fs::write(tmp)` + `fs::rename(tmp → auth-profiles.json)` calls in `write_persisted_locked` with `retry_with_backoff`, matching the auth-profiles `.lock`-create retry budget that already closed the sibling Windows transient-FS bugs `OPENHUMAN-TAURI-H1` / `H8` in #1641 / #2085.
- 3 regression tests pinned to the failure mode via a `#[cfg(test)]`-only `__TEST_TRANSIENT__` injection point that the existing `is_transient_fs_error` classifier already recognises (`src/openhuman/util.rs:618`).
- Closes #3355 — Sentry `TAURI-RUST-92J`, 10,158 events / 24h, single Windows 11 24H2 user, release `openhuman@0.56.0+e8968077aeb5`.

## Problem

`AuthProfilesStore::write_persisted_locked` (`src/openhuman/credentials/profiles.rs:911-946`) serialised the persisted JSON to a tmp file and renamed it onto `auth-profiles.json` via raw `std::fs::write` + `std::fs::rename`. On Windows, AV / Search-Indexer / Defender briefly holding the destination handle returns `ERROR_SHARING_VIOLATION (32)`, `ERROR_ACCESS_DENIED (5)`, or `ERROR_DELETE_PENDING (303)` — exactly the family `crate::openhuman::util::retry_with_backoff` is built to absorb. The sibling `.lock`-create at `profiles.rs:987` was already routed through this helper (PR #1641 / #2085 / #2180); the `.json` write+rename path was the missing partial.

Result: any single transient AV hold flipped a normally-retry-safe operation into a permanent error. Compounding it, `load_locked` (`profiles.rs:744`) calls `write_persisted_locked` whenever its in-memory purge set is non-empty (`#3125`-class drops, legacy `kind` values per `#2439`, decrypt-failure recovery), and the frontend polls `openhuman.app_state_snapshot` every ~2s — so one persistent AV hold amplified into 10k+ identical Sentry events in 24h on `TAURI-RUST-92J`.

## Solution

`write_persisted_locked` now wraps both the tmp write and the rename in `retry_with_backoff("…", PERSIST_RETRY_ATTEMPTS, PERSIST_RETRY_BASE_MS, …)`. Constants are set to **6 attempts at base 100ms** (worst-case ≈ 6.3 s), matching the proven `.lock`-create budget and staying well inside `LOCK_TIMEOUT_MS` so concurrent `acquire_lock` callers never starve behind a single retry-loop owner. Outer `with_context` is preserved on both calls so the Sentry fingerprint shape stays stable across releases.

This is the real fix to the underlying file-system race — **not noise suppression**. Genuinely unrecoverable failures still propagate via `Err` and reach the RPC error path, so a user whose AV is permanently hostile remains visible in Sentry as honest signal.

Regression coverage uses a `#[cfg(test)]`-only injection counter on `AuthProfilesStore` (`force_next_transient_failures` + `consume_test_transient_failure`) that returns an error chain containing `__TEST_TRANSIENT__` — the test sentinel that `is_transient_fs_error` already accepts. Three tests exercise the retry semantics:
- one-shot transient is absorbed and the profile lands on disk,
- a burst of transients within the retry budget still resolves to a successful write,
- sustained failures beyond the budget still surface as `Err` with the outer `with_context` preserved.

Production binaries carry zero new surface area from the injection helpers (gated behind `#[cfg(test)]`).

## Submission Checklist

- N/A: no UI changes (Rust-only fix; no i18n keys touched, no React surface)
- [x] Tests added — three regression tests in `src/openhuman/credentials/profiles_tests.rs` exercising one-shot retry, retry-within-budget burst, and retry-exhaustion behaviour. `cargo test --lib openhuman::credentials::profiles` is green locally (38 passed, 0 failed).
- [x] `cargo check --manifest-path Cargo.toml` clean on the changed crate.
- [x] `cargo fmt` clean on the changed files.
- N/A: `cargo clippy -D warnings` — pre-existing lints elsewhere in the crate (`useless_vec` in `tokenjuice/text/process.rs`, `map_or` simplifications in this file at lines 1106 / 1205 / 1211, assertion-on-constant in `profiles_tests.rs` at 600 / 619 / 623) are unchanged by this PR and untouched on `main`.
- N/A: rustdoc / `gitbooks` — no architecture or contributor-facing behaviour change (internal retry semantics on an existing private method).
- N/A: feature flag / capability catalog — no user-visible feature added, removed, or renamed.

## Impact

- Windows users whose AV / Search-Indexer / Defender intermittently held a handle on `auth-profiles.json` no longer see `app_state_snapshot` fail on every poll once a profile gets dropped — the write retries and lands on disk as soon as the handle releases.
- Sentry `TAURI-RUST-92J` event volume drops sharply once the next release ships; the issue stays open until clean so any genuinely sustained AV interference still surfaces.
- No behaviour change on macOS / Linux: `is_transient_fs_error` only matches Windows raw OS codes (5 / 32 / 33 / 303 / 1224), so non-Windows callers see identical first-attempt semantics.
- No schema change, no migration, no RPC surface change, no public API surface change.

## Related

- Closes #3355
- Sentry-Issue: TAURI-RUST-92J
- Sibling retries on the same store: #1641, #2085, #2180 (`.lock`-create transient-FS retry — same helper, same Windows error family).
- Related load-driven drop paths whose persist this fix now covers: #2439 (legacy `kind` value tolerance), #3125 / #3180 (OAuth missing `access_token` drop).

### DIFF
--- a/src/openhuman/credentials/profiles.rs
+++ b/src/openhuman/credentials/profiles.rs
@@ -1,4 +1,5 @@
 use crate::openhuman::keyring::SecretStore;
+use crate::openhuman::util::retry_with_backoff;
 use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
@@ -8,6 +9,11 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::thread;
 use std::time::{Duration, Instant};
+
+#[cfg(test)]
+use std::sync::atomic::{AtomicUsize, Ordering};
+#[cfg(test)]
+use std::sync::Arc;
 
 const CURRENT_SCHEMA_VERSION: u32 = 1;
 
@@ -50,6 +56,15 @@ const MALFORMED_LOCK_GRACE_MS: u64 = 2_000;
 /// Wait long enough for a fresh leaked lock to cross the stale threshold
 /// and be reclaimed before surfacing a lock timeout to the caller.
 const LOCK_TIMEOUT_MS: u64 = STALE_LOCK_AGE_MS + 5_000;
+
+/// Retry budget for the JSON write + rename in `write_persisted_locked`.
+/// Same shape as the lock-create call at the bottom of `acquire_lock` (which
+/// is what closed Sentry OPENHUMAN-TAURI-H1 / H8 in #1641 / #2085). 6 attempts
+/// at base 100ms doubles up to ~6.3s worst-case before surfacing. Sized to
+/// stay well inside `LOCK_TIMEOUT_MS` so concurrent acquire_lock callers
+/// never time out behind a single retry-loop owner.
+const PERSIST_RETRY_ATTEMPTS: u32 = 6;
+const PERSIST_RETRY_BASE_MS: u64 = 100;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
@@ -199,6 +214,13 @@ pub struct AuthProfilesStore {
     /// Whether the OS keychain is available on this machine.
     /// Cached at construction time to avoid repeated probes.
     use_keychain: bool,
+    /// `#[cfg(test)]` failure injection — when non-zero, the next retryable
+    /// FS call inside `write_persisted_locked` consumes one count and returns
+    /// a `__TEST_TRANSIENT__` error so the `is_transient_fs_error` classifier
+    /// treats it as retryable (`src/openhuman/util.rs:618`). Production
+    /// binaries never see this field.
+    #[cfg(test)]
+    force_transient_failures: Arc<AtomicUsize>,
 }
 
 impl AuthProfilesStore {
@@ -240,6 +262,8 @@ impl AuthProfilesStore {
             secret_store: SecretStore::new(state_dir, encrypt_secrets),
             user_id,
             use_keychain,
+            #[cfg(test)]
+            force_transient_failures: Arc::new(AtomicUsize::new(0)),
         }
     }
 
@@ -928,14 +952,41 @@ impl AuthProfilesStore {
         );
         let tmp_path = self.path.with_file_name(tmp_name);
 
-        fs::write(&tmp_path, &json).with_context(|| {
+        // Windows AV / Search-Indexer / Defender may briefly hold a handle on
+        // the destination, returning transient `ERROR_SHARING_VIOLATION (32)`,
+        // `ERROR_ACCESS_DENIED (5)`, or `ERROR_DELETE_PENDING (303)` —
+        // recognised as retryable by `is_transient_fs_error`. Mirror the
+        // lock-create retry budget at the bottom of `acquire_lock` so the
+        // JSON write+rename path absorbs the same transient family that
+        // closed Sentry OPENHUMAN-TAURI-H1 / H8 for the lock path. Outer
+        // `with_context` preserved so the Sentry fingerprint shape is stable
+        // across releases. (Sentry TAURI-RUST-92J / #3355.)
+        retry_with_backoff(
+            "write auth profile tmp",
+            PERSIST_RETRY_ATTEMPTS,
+            PERSIST_RETRY_BASE_MS,
+            || {
+                self.consume_test_transient_failure()?;
+                fs::write(&tmp_path, &json).context("write auth profile tmp")
+            },
+        )
+        .with_context(|| {
             format!(
                 "Failed to write temporary auth profile file at {}",
                 tmp_path.display()
             )
         })?;
 
-        fs::rename(&tmp_path, &self.path).with_context(|| {
+        retry_with_backoff(
+            "replace auth profile store",
+            PERSIST_RETRY_ATTEMPTS,
+            PERSIST_RETRY_BASE_MS,
+            || {
+                self.consume_test_transient_failure()?;
+                fs::rename(&tmp_path, &self.path).context("rename auth profile tmp -> store")
+            },
+        )
+        .with_context(|| {
             format!(
                 "Failed to replace auth profile store at {}",
                 self.path.display()
@@ -943,6 +994,42 @@ impl AuthProfilesStore {
         })?;
 
         Ok(())
+    }
+
+    /// Consume one test-injected transient FS failure if any are queued.
+    /// No-op in production builds.
+    #[cfg(test)]
+    fn consume_test_transient_failure(&self) -> Result<()> {
+        let remaining = self.force_transient_failures.load(Ordering::SeqCst);
+        if remaining == 0 {
+            return Ok(());
+        }
+        self.force_transient_failures.fetch_sub(1, Ordering::SeqCst);
+        Err(anyhow::anyhow!(
+            "__TEST_TRANSIENT__ injected transient FS failure"
+        ))
+    }
+
+    #[cfg(not(test))]
+    #[inline(always)]
+    fn consume_test_transient_failure(&self) -> Result<()> {
+        Ok(())
+    }
+
+    /// Queue `n` test-only forced transient FS failures. The next `n`
+    /// retryable calls inside `write_persisted_locked` return a
+    /// `__TEST_TRANSIENT__` error before the underlying FS op runs; the
+    /// retry helper treats them as retryable.
+    #[cfg(test)]
+    pub(super) fn force_next_transient_failures(&self, n: usize) {
+        self.force_transient_failures.store(n, Ordering::SeqCst);
+    }
+
+    /// Test introspection: how many forced transient failures are still
+    /// queued. Lets tests verify the retry helper drained the queue.
+    #[cfg(test)]
+    pub(super) fn remaining_forced_failures(&self) -> usize {
+        self.force_transient_failures.load(Ordering::SeqCst)
     }
 
     fn encrypt_optional(&self, value: Option<&str>) -> Result<Option<String>> {
--- a/src/openhuman/credentials/profiles_tests.rs
+++ b/src/openhuman/credentials/profiles_tests.rs
@@ -689,3 +689,102 @@ fn auth_profile_kind_serde_roundtrip() {
     let json = serde_json::to_string(&AuthProfileKind::Token).unwrap();
     assert_eq!(json, "\"token\"");
 }
+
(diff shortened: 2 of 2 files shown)


src/openhuman/credentials/profiles.rs
```diff
@@ -928,14 +952,41 @@ impl AuthProfilesStore {
...
         })?;
 
-        fs::rename(&tmp_path, &self.path).with_context(|| {
+        retry_with_backoff(
+            "replace auth profile store",
+            PERSIST_RETRY_ATTEMPTS,
+            PERSIST_RETRY_BASE_MS,
+            || {
+                self.consume_test_transient_failure()?;
+                fs::rename(&tmp_path, &self.path).context("rename auth profile tmp -> store")
```
M3gA-Mind, line 986: **The rename retry path — the headline of this PR — is never exercised by any test.** `force_transient_failures` is a single counter shared by both retry stages, and the write stage above always runs first and drains it. In all three tests the queue hits 0 before `fs::rename` is reached, so this closure's `consume_test_transient_failure()` always returns `Ok` and the rename retry loop never iterates:

- one-shot (queue 1): consumed by write attempt 1, write succeeds attempt 2, rename runs clean.
- burst (queue 5): consumed by write attempts 1–5, write succeeds attempt 6, rename runs clean.
- exhaust (queue 6): write stage exhausts and returns `Err` before rename is reached (its assertion is `"Failed to write temporary auth profile file"` — write-stage only).

So the rename wrapper has *line* coverage (it runs once, successfully) but no test proves it actually retries — which is exactly the "missing partial" the PR is about. A future refactor could delete this `retry_with_backoff` and every test would stay green. Recommend per-stage injection (separate counters or a stage tag) plus two tests: a rename-only transient that's absorbed, and a rename-stage exhaustion asserting the `"Failed to replace auth profile store"` outer context.

src/openhuman/credentials/profiles.rs
```diff
@@ -51,6 +57,15 @@ const MALFORMED_LOCK_GRACE_MS: u64 = 2_000;
 /// and be reclaimed before surfacing a lock timeout to the caller.
 const LOCK_TIMEOUT_MS: u64 = STALE_LOCK_AGE_MS + 5_000;
 
+/// Retry budget for the JSON write + rename in `write_persisted_locked`.
+/// Same shape as the lock-create call at the bottom of `acquire_lock` (which
+/// is what closed Sentry OPENHUMAN-TAURI-H1 / H8 in #1641 / #2085). 6 attempts
+/// at base 100ms doubles up to ~6.3s worst-case before surfacing. Sized to
```
M3gA-Mind, line 63: **Backoff math is off.** With `attempts = 6`, `retry_with_backoff` sleeps only 5 times (the last failure breaks without sleeping): `100+200+400+800+1600 ≈ 3.1s` **per stage**, not ~6.3s. The ~6.3s figure is actually the *combined* worst case of the two sequential stages (write ≈3.1s + rename ≈3.1s). The conclusion still holds — combined ≈6.2s is far inside `LOCK_TIMEOUT_MS` (30_000 + 5_000 = 35s) — but since this reasoning is load-bearing, suggest rewording to "≈3.1s per stage, ≈6.2s across both write+rename."

src/openhuman/credentials/profiles.rs
```diff
@@ -928,14 +952,41 @@ impl AuthProfilesStore {
...
+        )
+        .with_context(|| {
             format!(
                 "Failed to write temporary auth profile file at {}",
                 tmp_path.display()
             )
         })?;
 
-        fs::rename(&tmp_path, &self.path).with_context(|| {
+        retry_with_backoff(
```
M3gA-Mind, line 980: Nit (pre-existing, optional): `tmp_name` is `…tmp.{pid}.{nanos}` — unique per call — so a permanently-failing `fs::rename` orphans a distinct tmp file every call. With the ~2s `app_state_snapshot` poll, sustained failures accumulate orphan tmps. Not introduced here (the original also left a tmp on rename failure), but the retry lengthens each failing call and the failure window, so the accumulation is a bit more pronounced. A best-effort `let _ = fs::remove_file(&tmp_path);` on rename-exhaustion would keep the directory clean. Low priority.

M3gA-Mind (commented):
Solid, well-reasoned root-cause fix with correct retry semantics (verified it returns immediately on non-transient errors and only retries the Windows transient family) and good Sentry-fingerprint discipline. One substantive item to resolve before merge — the rename retry path, which is the headline of the PR, is never actually exercised by the tests — plus a small comment-accuracy fix. Inline below.

src/openhuman/credentials/profiles.rs
oxoxDev: Addressed in 3c549a70a (tests) + 6b0f998bb (impl) via #3398. Split the failure-injection counter into per-stage `force_next_write_failures` / `force_next_rename_failures`, plus two new tests: `rename_stage_retries_one_shot_transient` and `rename_stage_exhausts_retries_and_cleans_up_tmp` — the second now asserts the rename outer `with_context("Failed to replace auth profile store …")` is preserved, which is exactly the unreachable line you flagged.

src/openhuman/credentials/profiles.rs
oxoxDev: Addressed in 6b0f998bb via #3398. Reworded to "≈3.1s per stage, ≈6.2s across both write+rename" and explicit cite of `LOCK_TIMEOUT_MS = 35_000` as the safety margin.

src/openhuman/credentials/profiles.rs
oxoxDev: Addressed in 6b0f998bb via #3398. Added best-effort `fs::remove_file(&tmp_path)` after the rename retry exhausts; `rename_stage_exhausts_retries_and_cleans_up_tmp` asserts no orphan `auth-profiles.json.tmp.*` remains. Result deliberately dropped — same AV that blocked the rename may block the unlink.

PR #3367, opened 2026-06-04
feat(agent): reduce orchestrator prompt bloat

### LINKED ISSUE #3332: Reduce agent prompt bloat with specialist subagents and grounding contracts
## Summary

Reduce the amount of product/domain policy carried by the front-line orchestrator prompt by moving domain-specific behavior into specialist subagents, and reduce hallucinations by making delegation and subagent results evidence-shaped.

User-visible outcome: the chat agent should route faster, carry less prompt sludge, answer fewer stale/current-service questions from priors, and synthesize subagent work from explicit evidence instead of freeform summaries.

## Problem

The orchestrator prompt has become a catch-all policy dump. It currently mixes routing, cron behavior, connection UX, response style, memory retrieval, citation rules, presentation generation, image verification, and anti-hallucination warnings in one front-line chat prompt.

Reference code:

```text
src/openhuman/agent_registry/agents/orchestrator/prompt.md
```

Representative hotspots:

```md
## Delegation Decision Tree (Direct-First)
...
**Scheduling rule of thumb.**
...
## Connecting external services
...
## Memory tree retrieval (historical context only)
...
## Presentation generation
...
## Citations
...
## Presentations with images
```

The prompt is 239 lines today, while most specialist prompts are much smaller. That makes the chat-tier model responsible for remembering every domain-specific exception on every turn.

The runtime already has most of the right subagent architecture:

```toml
# src/openhuman/agent_registry/agents/orchestrator/agent.toml
subagents = [
    "researcher",
    "planner",
    "code_executor",
    "tools_agent",
    "skill_creator",
    "critic",
    "archivist",
    "crypto_agent",
    "markets_agent",
    "mcp_setup",
    { skills = "*" },
]
```

The subagent runner already builds a narrow child context and strips nested delegation tools:

```rust
// src/openhuman/agent/harness/subagent_runner/ops.rs
let mut allowed_indices = filter_tool_indices(
    &parent.all_tools,
    &definition.tools,
    &definition.disallowed_tools,
    options
        .skill_filter_override
        .as_deref()
        .or(definition.skill_filter.as_deref()),
);

allowed_indices.retain(|&i| {
    let name = parent.all_tools[i].name();
    !is_subagent_spawn_tool(name) && name != "spawn_worker_thread"
});
```

But too many rules still live in the parent prompt instead of in specialists with smaller tool surfaces.

### 1. Shrink the orchestrator prompt to routing + synthesis only

Keep the orchestrator focused on:

- direct reply vs direct tool vs delegate
- current/live source vs historical memory routing
- minimal context handoff
- evidence-aware synthesis
- concise user-facing response style

Move domain-specific rules out of `orchestrator/prompt.md`.

### 2. Add or expose specialist subagents for prompt-heavy domains

Create or expose these agents:

- `scheduler_agent` — owns `current_time`, `cron_add`, `cron_list`, `cron_remove`, relative-time grounding, and confirmation-before-schedule rules.
- `presentation_agent` — owns `generate_presentation`, factual grounding, citations, image verification, and deck warnings.
- `desktop_control_agent` — owns `launch_app` and `ax_interact` behavior, including list-before-press and sensitive-app constraints.
- `help` — already exists and should be exposed through orchestrator subagents for OpenHuman product/docs questions.

Reference existing help agent:

```toml
# src/openhuman/agent_registry/agents/help/agent.toml
id = "help"
delegate_name = "ask_docs"
when_to_use = "Product help — answers questions about how OpenHuman works, what features exist, how to configure things, or where to find a guide. Reads the OpenHuman GitBook docs via the `gitbooks_*` tools. Use this for any 'how do I…' / 'what does X do' / 'where is the setting for…' question about OpenHuman itself, before guessing or making things up."

[tools]
named = ["gitbooks_search", "gitbooks_get_page", "memory_recall"]
```

Suggested new scheduler definition:

```toml
# src/openhuman/agent_registry/agents/scheduler_agent/agent.toml
id = "scheduler_agent"
display_name = "Scheduler Agent"
delegate_name = "schedule_task"
when_to_use = "Scheduling specialist — owns reminders, recurring jobs, cron listing/removal, relative-time grounding, and confirmation-before-create flows. Use for any request to remind, schedule, repeat, pause, remove, or inspect scheduled jobs."
temperature = 0.2
agent_tier = "worker"
omit_identity = true
omit_memory_context = true
omit_safety_preamble = false
omit_skills_catalog = true
omit_profile = true
omit_memory_md = true

[model]
hint = "agentic"

[tools]
named = ["current_time", "cron_add", "cron_list", "cron_remove", "ask_user_clarification"]
```

Suggested new presentation definition:

```toml
# src/openhuman/agent_registry/agents/presentation_agent/agent.toml
id = "presentation_agent"
display_name = "Presentation Agent"
delegate_name = "make_presentation"
when_to_use = "Presentation specialist — creates decks from supplied or retrieved evidence. Owns factual grounding, source citation, image verification, and generate_presentation warnings. Use for make slides, build a deck, pitch deck, or presentation requests."
temperature = 0.2
agent_tier = "worker"
omit_identity = true
omit_memory_context = false
omit_safety_preamble = false
omit_skills_catalog = true
omit_profile = true
omit_memory_md = true

[model]
hint = "agentic"

[tools]
named = ["generate_presentation", "memory_tree", "query_memory"]
extra = ["research"]
```

### 3. Remove broad direct tools from the orchestrator once specialists own them

Today the orchestrator directly sees many tools that force prompt-specific instruction blocks into the parent:

```toml
# src/openhuman/agent_registry/agents/orchestrator/agent.toml
named = [
    "query_memory",
    "memory_store",
    "save_preference",
    "memory_forget",
    "memory_tree",
    "read_workspace_state",
    "ask_user_clarification",
    "spawn_worker_thread",
    "spawn_parallel_agents",
    "composio_list_connections",
    "launch_app",
    "ax_interact",
    "current_time",
    "cron_add",
    "cron_list",
    "cron_remove",
    "todowrite",
    "update_task",
    "plan_exit",
    "run_skill",
    "update_check",
    "update_apply",
    "workflow_load",
    "workflow_phase",
    "generate_presentation",
]
```

After specialist agents land, the parent should keep only truly parent-owned tools. Candidate remaining set:

```toml
named = [
    "query_memory",
    "read_workspace_state",
    "ask_user_clarification",
    "spawn_worker_thread",
    "spawn_parallel_agents",
    "composio_list_connections",
    "todowrite",
    "update_task",
    "plan_exit",
]
```

Everything else should be routed through `delegate_*` tools with narrow prompts and narrow tool surfaces.

### 4. Make delegation inputs structured instead of one freeform prompt

The generic delegate wrapper currently accepts only a freeform `prompt` plus optional `model`:

```rust
// src/openhuman/agent_orchestration/tools/archetype_delegation.rs
fn parameters_schema(&self) -> serde_json::Value {
    json!({
        "type": "object",
        "required": ["prompt"],
        "properties": {
            "prompt": {
                "type": "string",
                "description": "Clear instruction for what to do. Include all relevant context — the sub-agent has no memory of your conversation."
            },
            "model": {
                "type": "string",
                "description": "Optional exact model id for this delegation only. Keeps the parent provider/routing, but pins the child agent to this model instead of the agent definition's default."
            }
        }
    })
}
```

This encourages context dumping. Add optional structured fields while preserving `prompt` compatibility:

```rust
json!({
    "type": "object",
    "required": ["prompt"],
    "properties": {
        "prompt": { "type": "string" },
        "objective": {
            "type": "string",
            "description": "One sentence outcome the child must produce."
        },
        "evidence": {
            "type": "array",
            "items": { "type": "string" },
            "description": "Only facts, file paths, URLs, ids, or tool outputs the parent has actually observed."
        },
        "constraints": {
            "type": "array",
            "items": { "type": "string" }
        },
        "must_not_assume": {
            "type": "array",
            "items": { "type": "string" }
        },
        "expected_output": {
            "type": "string",
            "description": "Requested output shape, e.g. findings list, patch summary, cited answer."
        },
        "citation_requirement": {
            "type": "string",
            "enum": ["none", "file_paths", "urls", "retrieval_hits", "tool_outputs"]
        },
        "model": { "type": "string" }
    }
})
```

The wrapper can render these into a compact child prompt before `dispatch_subagent`.

### 5. Make subagent outputs evidence-shaped

Today subagents return a collapsed freeform output. Add a shared result contract, at least for agents that produce factual answers:

```md
## Sub-agent Result Contract

Return a compact result with these headings:

- Answer
- Evidence used
- Actions taken
- Open uncertainties
- Failed tool calls
- Recommended next step

Do not include facts in Answer that are not supported by Evidence used or Actions taken.
```

The runner already appends a role contract:

```rust
// src/openhuman/agent/harness/subagent_runner/ops.rs
let system_prompt = append_subagent_role_contract(system_prompt, &definition.id);
```

Extend that role contract, or make it per-agent via definition metadata, so the orchestrator receives evidence-preserving outputs instead of prose that invites synthesis-time hallucination.

## Reducing hallucinations

Use subagents to make hallucination-prone work evidence-first instead of prompt-warning-first.

### Current/live facts

Rule: if the answer depends on current state, route to a live tool or specialist. Do not answer from priors.

Existing good reference:

```rust
// src/openhuman/agent_registry/agents/orchestrator/prompt.rs
// The `integrations_agent` has the ground-truth tool catalogue (`tools` +
// `gated_tools`); only it can answer "can I do X?" honestly. Force-delegate
// capability questions, not just task requests.
```

Apply this same pattern to scheduling, desktop actions, presentations, product help, and docs/API questions.

### Memory

Rule: memory is historical unless explicitly refreshed. The prompt already warns about staleness:

```rust
// src/openhuman/agent/prompts/mod.rs
"Treat this as durable background context, but NOT as fresh, present-tense fact..."
```

Move memory-specific retrieval/citation behavior into a memory-aware specialist or make the orchestrator call a verifier when memory is used for a factual answer.

### Tool output truncation

Rule: never reason over truncated large outputs as if complete. If a tool output is oversized, summarize/extract with preserved identifiers before synthesis.

Reference existing summarizer agent:

```toml
# src/openhuman/agent_registry/agents/summarizer/agent.toml
id = "summarizer"
when_to_use = "Compresses oversized tool results for the orchestrator. Called automatically by the runtime when a tool returns more than summarizer_payload_threshold_tokens..."
```

Follow-up: either re-enable the summarizer safely with recursion protection or replace it with a deterministic extractor for large tool results.

### Presentations and images

Rule: deck content must be generated from supplied/retrieved evidence, not model priors. Images must be verified before captions assert what they show. Move the existing rules from `orchestrator/prompt.md` into `presentation_agent/prompt.md` and make the generated deck response include source references and image warnings.

### Verification pass

Add a lightweight `grounding_verifier` or expand `critic` beyond code review for risky final answers. Trigger it when the parent answer includes:

- current facts
- external-service capability claims
- generated presentation content
- finance/crypto/market actions
- direct quotes from memory or retrieved sources
- claims derived from long/truncated tool outputs

Verifier output should be machine-checkable:

```md
Decision: pass | revise | ask_user | fetch_more
Unsupported claims:
- ...
Required next tool call:
- ...
```

## Acceptance criteria

- [ ] **Prompt split** — `orchestrator/prompt.md` is reduced to routing, synthesis, and style only; scheduling/presentation/desktop-control/product-help rules move into specialist prompts.
- [ ] **Specialists wired** — add or expose `scheduler_agent`, `presentation_agent`, `desktop_control_agent`, and `help` through the orchestrator subagent inventory.
- [ ] **Tool visibility narrowed** — remove specialist-owned direct tools from the orchestrator after delegates exist, while preserving user-visible behavior through `delegate_*` tools.
- [ ] **Structured delegation** — `ArchetypeDelegationTool` supports structured handoff fields such as `objective`, `evidence`, `constraints`, `must_not_assume`, `expected_output`, and `citation_requirement`.
- [ ] **Evidence-shaped results** — factual subagents return evidence-preserving outputs, and the orchestrator synthesis contract says not to introduce unsupported claims.
- [ ] **Hallucination regression tests** — add focused tests covering live integration capability questions, memory staleness, presentation grounding, and large/truncated tool result handling.
- [ ] **Prompt size tests** — add tests or snapshots that fail if the orchestrator prompt re-absorbs specialist domain rules.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by `.github/workflows/pr-ci.yml`).

## Related

- `src/openhuman/agent_registry/agents/orchestrator/prompt.md`
- `src/openhuman/agent_registry/agents/orchestrator/agent.toml`
- `src/openhuman/agent_orchestration/tools/archetype_delegation.rs`
- `src/openhuman/agent/harness/subagent_runner/ops.rs`
- `src/openhuman/agent_registry/agents/help/agent.toml`
- `src/openhuman/agent_registry/agents/summarizer/agent.toml`

senamakel:
## Summary

- Split prompt-heavy scheduling, presentation, and desktop-control behavior out of the orchestrator into narrow built-in specialist agents.
- Exposed `help`, `scheduler_agent`, `presentation_agent`, and `desktop_control_agent` through the orchestrator subagent inventory.
- Narrowed orchestrator direct tool visibility and updated routing/synthesis language around evidence-shaped delegation.
- Added structured handoff fields to `ArchetypeDelegationTool` while preserving the existing freeform `prompt` contract.
- Extended subagent result contracts so factual workers return answer, evidence, actions, uncertainties, failed calls, and next-step sections.

## Problem

- The front-line orchestrator prompt had accumulated domain-specific policy for scheduling, presentation generation, desktop actions, docs/help, citation behavior, and hallucination warnings.
- Direct access to specialist-owned tools encouraged the parent agent to carry and apply those rules on every turn.
- Freeform delegation inputs and freeform subagent outputs made it too easy to dump context into workers and synthesize unsupported claims afterward.

## Solution

- Added three built-in worker agents with dedicated prompts and narrow tool surfaces: `scheduler_agent`, `presentation_agent`, and `desktop_control_agent`.
- Registered the existing `help` worker and new specialists in the orchestrator subagent list, loader builtins, and MCP prompt resource catalog.
- Reduced the orchestrator prompt to routing, minimal handoff, evidence-aware synthesis, memory freshness, and concise response style.
- Removed specialist-owned direct tools from the orchestrator and verified presentation generation now routes through `make_presentation`.
- Added optional structured fields to archetype delegation: `objective`, `evidence`, `constraints`, `must_not_assume`, `expected_output`, and `citation_requirement`.
- Added focused Rust tests for registry wiring, prompt routing, prompt-resource parity, structured delegation rendering/schema, and subagent result contracts.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/pr-ci.yml`](../.github/workflows/pr-ci.yml). Local diff-cover was not run; focused Rust tests cover changed prompt/delegation wiring and CI will enforce the gate.
- [x] Coverage matrix updated — N/A: agent prompt/registry behavior change, no added/removed user-facing feature row.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no matrix feature ID changed.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: no release-cut smoke surface changed.

## Impact

- Runtime/platform impact: Rust core agent registry and prompt orchestration only; no frontend, Tauri shell behavior, migration, or RPC surface changes.
- Security/compatibility: narrows orchestrator tool exposure and keeps existing `prompt` delegation compatibility.
- Performance: orchestrator prompt is smaller and should carry less domain policy on normal chat turns.

## Related

- Closes: #3332
- Follow-up PR(s)/TODOs: Full grounding verifier and deterministic oversized-tool-output extractor remain follow-up work from the issue body.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `issue/3332-reduce-agent-prompt-bloat-with-specialis`
- Commit SHA: `4bb91a573f29cf5f0a321f9a3620e15732b890c9`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — passed via pre-push hook.
- [x] `pnpm typecheck` — passed via pre-push hook (`pnpm compile`).
- [x] Focused tests:
  - `cargo test --manifest-path Cargo.toml orchestrator -- --nocapture`
  - `cargo test --manifest-path Cargo.toml subagent_role_contract -- --nocapture`
  - `cargo test --manifest-path Cargo.toml archetype_delegation -- --nocapture`
  - `cargo test --manifest-path Cargo.toml specialist_agents_are_registered_with_narrow_tools -- --nocapture`
  - `cargo test --manifest-path Cargo.toml presentation_agent_lists_generate_presentation_and_grounding_tools -- --nocapture`
  - `cargo test --manifest-path Cargo.toml openhuman::mcp_server::resources::tests -- --nocapture`
  - `cargo test --manifest-path Cargo.toml openhuman::memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest -- --nocapture`
- [x] Rust fmt/check (if changed): `cargo fmt --manifest-path Cargo.toml`; `cargo check --manifest-path Cargo.toml`; `cargo test --manifest-path Cargo.toml` ran, but full suite has one pre-existing/order-dependent failure listed below.
- [x] Tauri fmt/check (if changed): `cargo check --manifest-path app/src-tauri/Cargo.toml`; also passed via pre-push hook `pnpm rust:check`.

### Validation Blocked

- `command:` `cargo test --manifest-path Cargo.toml`
- `error:` full suite fails only `openhuman::memory_tree::tree::rpc::tests::pipeline_status_reports_chunk_aggregates_after_ingest` with `degraded status should explain semantic recall loss: Some("wiki structure incomplete")`; the same test passed in isolation after the full-suite failure.
- `impact:` unrelated to this branch's agent registry/prompt/delegation changes; focused tests for the touched surfaces pass.

### Behavior Changes

- Intended behavior change: front-line orchestrator delegates scheduling, presentation generation, desktop-control, and product-help questions to specialist workers instead of carrying detailed domain policy and tools directly.
- User-visible effect: agent routing should be more evidence-oriented and less likely to answer specialist/current-state questions from priors.

### Parity Contract

- Legacy behavior preserved: existing delegate `prompt` input remains required and compatible; specialist-owned behavior remains available through delegate tools.
- Guard/fallback/dispatch parity checks: tests verify built-in registration, MCP prompt catalog parity, narrowed presentation tool visibility, structured handoff rendering, and evidence-shaped subagent result contracts.

- Duplicate PR(s): N/A
- Canonical PR: this PR

## Summary by CodeRabbit

* **New Features**
  * Added three specialist agents: Desktop Control, Presentation, and Scheduler.
  * Structured delegation parameters for sub-agent handoffs (objective, evidence, constraints, must_not_assume, expected_output, citation requirement) and optional per-delegation model override.

* **Improvements**
  * Orchestrator delegation refined to route to specialists and favor direct-first handling.
  * Sub-agent result contract and system prompts updated to require compact, evidence-backed result sections and explicit handling for truncated/partial tool output.


src/openhuman/agent_registry/agents/orchestrator/agent.toml
```diff
@@ -108,37 +121,12 @@ hint = "chat"
 # / composio_execute directly.
 named = [
```
M3gA-Mind, line 122: **Orchestrator silently loses memory-write + skill-run with no replacement route.** This block drops `memory_store`, `save_preference`, `memory_forget`, and `run_skill`, but the decision tree adds no delegation line for "remember this / save my preference / forget that / run a skill," and none of the three new specialists own those tools — the orchestrator keeps only `query_memory` (read) and `memory_tree`. So "remember that I prefer dark mode" now has no tool path on the front-line agent. The PR description only documents removing scheduling/presentation/desktop tools, so this looks undocumented. If it's intended (route to `delegate_archivist` / hand back to Core), please add the decision-tree line + a test; otherwise it's a capability regression.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
```diff
@@ -175,42 +132,7 @@ User: what time is it?
```
M3gA-Mind, line 132: **Stale `## Response Style` examples now contradict the refactor.** The examples just above this line still ship `User: what time is it? → 7:31pm` and `User: remind me to stretch in 10 min → got it / reminder set for 7:42pm`. With `current_time` and `cron_*` removed from the orchestrator, these model the agent answering the clock and confirming a reminder **directly from priors** — fabricating both the time value and the "reminder set" outcome — which is exactly what `scheduler_agent` and the new evidence-grounding contract exist to prevent. Please update them to show an ack + delegation to `schedule_task`, or the prompt will actively coach the wrong behavior.

src/openhuman/agent_orchestration/tools/archetype_delegation.rs
```diff
@@ -76,6 +106,64 @@ impl Tool for ArchetypeDelegationTool {
     }
 }
 
+fn render_structured_handoff(prompt: &str, args: &Value) -> String {
```
M3gA-Mind, line 109: **Minor / parity note:** `render_structured_handoff` reshapes *every* archetype delegation, not just the new specialists. `ArchetypeDelegationTool` backs all delegates (crypto, integrations, etc.), so every child prompt is now wrapped in `Task:\n…` even when no structured fields are passed. It's additive and well-tested, but the "existing `prompt` delegation remains compatible" parity claim is a bit stronger than reality — the literal child-prompt string changes for all delegates. Worth a one-line note in the parity section, and consider skipping the `Task:\n` wrapper when no structured fields are present to keep the legacy path byte-identical.

src/openhuman/agent_registry/agents/orchestrator/prompt.md
```diff
@@ -1,14 +1,14 @@
-# Orchestrator — Staff Engineer
+# Orchestrator - Staff Engineer
```
M3gA-Mind, line 1: Nit: this changes the H1 from an em-dash (`Orchestrator — Staff Engineer`) to a hyphen, which is unrelated churn for this PR. Arguably consistent with the prompt's own "avoid em dashes" guidance, but it's a heading (not model output), so either keep it intentional or revert to keep the diff focused on the refactor.

M3gA-Mind (commented):
Well-structured, well-tested refactor that meaningfully shrinks the orchestrator prompt and pins the new routing with good negative assertions (delegate-name parity, exact-line matching, narrow tool scopes all verified). Two substantive items to resolve before merge plus a couple of minor notes inline.

senamakel-droid (dismissed):

PR #3409, opened 2026-06-05
fix(credentials): degrade auth-profile read to lock-free on disk-full (#3407)

### LINKED ISSUE #3407: Auth-profile read path fails (and floods Sentry) when disk is full
## Source
Sentry: TAURI-RUST-4SZ — `Failed to create auth profile lock (kind=Some(StorageFull), os_code=Some(28))`
Events: 4129 · Users affected: 4 · First seen: 2026-05-27 · Last seen: 2026-06-05 (live)
Release: openhuman@0.57.13 · env: production · macOS

## Symptom
When a user's disk is full (POSIX `ENOSPC`, `io::ErrorKind::StorageFull`), the auth-profile **read** path fails. Because `openhuman.app_state_snapshot` polls on a timer and reads the app-session profile each tick, the failure repeats ~once per poll — ~1000 events per affected user.

## Root cause
The pure-read path takes an exclusive **write** lock:

`app_state_snapshot` → `snapshot()` → `load_app_session_profile` → `AuthService::get_profile` → `AuthProfilesStore::load()` → `acquire_lock()`

`acquire_lock()` creates a lock file via `OpenOptions::create_new(true).write(true)`. On a full disk that `open` returns `StorageFull`, so `load()` errors — even though the existing `auth-profiles.json` is present and readable. Writers already publish atomically (tmp + `fs::rename`), so a reader never needs the lock for consistency; the read-lock only matters to serialize the read-modify-write callers (`upsert_profile`, `remove_profile`, `set_active_profile`, `clear_active_profile`, `update_profile`). The read path inherits a write requirement it doesn't need, which (a) strands `app_state_snapshot` on a full disk and (b) floods Sentry every poll.

A `DiskFull` Sentry suppression exists but its matcher anchors on the inner `"no space left on device"` text; the flattened RPC error only carries the outer `(kind=Some(StorageFull), os_code=Some(28))` line, so the noise is not even demoted.

## Stack (top in-app frame)
`src/openhuman/credentials/profiles.rs` — `AuthProfilesStore::acquire_lock` (`create_new` open) via `AuthProfilesStore::load` (read path)

## Reproduces on
- Branch: upstream/main (read-lock present on `load()`; `is_disk_full_message` narrow)
- Condition: host filesystem full (ENOSPC) while `app_state_snapshot` polls

## Fix
1. **Prevent (headline):** on a filesystem-unwritable lock-create failure (`StorageFull` / `ReadOnlyFilesystem`), degrade `load()` to a lock-free read-only load (consistent thanks to atomic-rename writers); skip opportunistic migrations that couldn't persist anyway. The read succeeds, so no error is emitted.
2. **Defense-in-depth:** extend the `DiskFull` classifier to also match the `StorageFull` ErrorKind rendering so the genuinely-unpreventable **write**-path disk-full failures are demoted, with a drift-guard test coupling the classifier to the producer.

## Bug shape
Read path inheriting an unnecessary write requirement; expected-error demotion incomplete.

oxoxDev:
## Summary

- Make the auth-profile **read** path survive a full / read-only filesystem instead of failing: `load()` degrades to a lock-free read-only load when the exclusive lock can't be created.
- Fixes a Sentry flood (TAURI-RUST-4SZ — 4129 events / 4 users) where `app_state_snapshot` errored once per poll on disk-full users and stranded the UI.
- Defense-in-depth: teach the `DiskFull` Sentry classifier to also recognise the `StorageFull` ErrorKind wire shape it was silently missing, for the genuinely-unpreventable **write** paths.
- This is an RCA fix (the read never needed the write-lock), not a suppression of the symptom.

## Problem

When a user's disk is full (POSIX `ENOSPC` → `io::ErrorKind::StorageFull`), the pure-read path failed:

`app_state_snapshot` → `snapshot()` → `load_app_session_profile` → `AuthService::get_profile` → `AuthProfilesStore::load()` → `acquire_lock()`

`acquire_lock()` creates a lock file via `OpenOptions::create_new(true).write(true)`. On a full disk that `open` returns `StorageFull`, so `load()` errored — even though `auth-profiles.json` already exists and is readable. Because `app_state_snapshot` polls this read on a timer, the failure repeated ~once per poll: ~1000 events per affected user, and the UI stranded on the snapshot it could no longer fetch.

The existing `DiskFull` Sentry demotion didn't even catch it: its matcher anchors on the inner `"no space left on device"` Display text, but the error is flattened single-line at the RPC boundary so only the outer `"(kind=Some(StorageFull), os_code=Some(28))"` survives — no anchor text, so the noise leaked.

## Solution

**Root cause:** the read path inherited a *write* requirement it never needed. Writers publish the store atomically (`tmp` + `fs::rename`), so a bare read is always consistent without the lock; the exclusive lock only matters to serialise the read-modify-write callers (`upsert`/`remove`/`set_active`/`clear_active`/`update`).

- `load()` now matches on `acquire_lock()`: on a filesystem-unwritable kind (`StorageFull` / `ReadOnlyFilesystem`) it degrades to a **lock-free read-only load**; lock contention / timeout and all other errors still propagate unchanged.
- The read+resolution body is shared via `load_resolved(persist)`. The read-only path (`persist = false`) skips only the opportunistic migration/cleanup rewrite — which couldn't persist on a full disk anyway — and returns the same fully-resolved data.
- Because the read now succeeds, **no error is produced** — the flood stops at the source rather than being silenced downstream. The user keeps their session view while the disk is full.
- Defense-in-depth (separate, honest suppression for the *unpreventable* write paths): `is_disk_full_message` also anchors on the cross-platform `StorageFull` ErrorKind token, with a drift-guard test coupling the classifier to the real `annotate_lock_create_failure` output so a future format/std rename fails CI instead of re-leaking.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed prod lines exercised by 4 new tests (fallback, parity, polarity, drift); CI `diff-cover` enforces the threshold
- [x] Coverage matrix updated — `N/A: bug-fix / behaviour-resilience change, no feature row added or renamed`
- [x] All affected feature IDs from the matrix are listed in `## Related` — `N/A: no matrix rows affected`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no user-facing/UI surface change; core read-path resilience only`

## Impact

- **Platform:** desktop core (Rust). No UI, schema, or protocol change.
- **Behaviour:** on a full / read-only disk, auth-profile reads now succeed (lock-free, consistent via atomic-rename writers) instead of failing; `app_state_snapshot` keeps serving the session. Write paths are unchanged (they still require the lock and correctly fail on a full disk — now demoted, not flooded).
- **Security:** no change to locking semantics for writers; the lock-free path is read-only and skips persistence, so no concurrent-write hazard is introduced.
- **Performance:** one extra `clone` of `active_profiles` on the shared read path; negligible.

## Related

- Closes: #3407
- Follow-up PR(s)/TODOs: none

Sentry-Issue: TAURI-RUST-4SZ

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3407-disk-full-auth-read`
- Commit SHA: `598429e10`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — ran via pre-push hook (passed)
- [x] `pnpm typecheck` — ran via pre-push hook (passed)
- [x] Focused tests: `cargo test --lib openhuman::credentials::profiles` (42 passed) + `core::observability::tests::classifies_disk_full_errors`
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo check --lib` + `cargo clippy --lib` clean (no new warnings)
- [x] Tauri fmt/check (if changed): `pnpm rust:check` passed via pre-push hook

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: auth-profile reads degrade to a lock-free consistent read on disk-full/read-only FS instead of erroring.
- User-visible effect: disk-full users keep their session in `app_state_snapshot` instead of being stranded; Sentry no longer floods per poll.

### Parity Contract

- Legacy behavior preserved: locked write paths and the normal locked read path are unchanged; only the disk-full/read-only branch is new.
- Guard/fallback/dispatch parity checks: `is_lock_create_unwritable_fs` matches only `StorageFull`/`ReadOnlyFilesystem`; contention/timeout/permission errors still propagate (covered by polarity test).

- Duplicate PR(s): none (checked open + 30d-merged upstream PRs)
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of disk-full and read-only filesystem errors—the application now gracefully degrades to read-only mode for credential operations instead of failing completely when the disk is full or filesystem is read-only.
  * Enhanced error detection to recognize additional disk-full error formats.

* **Tests**
  * Added test coverage for disk-full resilience scenarios.


sanil-23 (approved):
**Recommendation: Approve** ✅

Tightly-scoped, well-engineered resilience fix for Sentry TAURI-RUST-4SZ. When a full/read-only filesystem blocks creating the exclusive lock file, `load()` now degrades to a lock-free read of the atomically-published store instead of failing and flooding Sentry once per `app_state_snapshot` poll.

### What's done well
- **Sound safety argument** — lock-free read is justified by the existing atomic `tmp + fs::rename` write protocol, so a bare read always sees a complete file.
- **Surgical predicate** — `is_lock_create_unwritable_fs()` matches *only* `StorageFull`/`ReadOnlyFilesystem` via the error chain; lock contention/timeout and `PermissionDenied` deliberately still propagate (with an explicit polarity test).
- **Clean refactor** — `load_locked`/`load_unlocked_readonly` funnel through `load_resolved(persist)`; in-memory cleanup applies either way, only the disk rewrite is gated.
- **Belt-and-suspenders classifier** — the `"storagefull"` anchor in `observability.rs` handles the `ErrorKind`-debug wire shape that the legacy "no space left on device" text never reaches, and `disk_full_lock_failure_string_classifies_as_disk_full` couples the classifier to actual producer output so a future format/std rename breaks CI here.
- Conventions followed; CI fully green (clippy/fmt, core + Tauri coverage, diff-cover ≥80%, E2E).

### Non-blocking observations
1. **Log noise (worth a follow-up)** — the fallback branch runs every poll while disk stays full and `log::warn!`s each time. The in-code "noise stops at the source" comment refers to *Sentry* noise; the local warn still repeats. CodeRabbit's `std::sync::Once` (warn once, downgrade repeats to `debug`) suggestion is legitimate.
2. **Comment slightly overstates "migrations are skipped"** — only the JSON disk rewrite is gated by `persist`. The opportunistic keychain promotion (`keychain_store_secrets`) still runs on the lock-free path. Benign (keychain is a separate store; re-promotion is idempotent and the on-disk file is left untouched), just broader than the comment implies.
3. **Lock-free read can still reach the mutating `quarantine_corrupt_store` rename** if the store is unparseable — would fail on a full/read-only FS. Requires corrupt-store + disk-full simultaneously, which the atomic-write protocol makes very unlikely; pre-existing on the read path regardless.

None block merge. Core logic, error classification, and test coverage are correct.

PR #3411, opened 2026-06-05
fix(agent): create action_dir on boot + separator-correct tilde expansion (#3353)

### LINKED ISSUE #3353: Shell tool fails with "The directory name is invalid" (os error 267) on Windows when workspace_dir doesn't match action_dir
## Summary

On a Windows machine, every shell/command execution fails with `Failed to execute command: The directory name is invalid. (os error 267)` — even trivial commands like `echo hello` or `pwd`. The shell tool cannot initialise because the configured working directory is invalid at the OS level.

## Problem

**What happened:** Every shell tool call (shell, node_exec, npm_exec) fails with the same error. Commands like `echo hello`, `pwd`, `dir`, `ls` all fail identically. The error is `os error 267` which maps to Windows `ERROR_DIRECTORY` — the directory name is invalid.

**What was expected:** The shell tool should either (a) fall back to a valid directory, (b) surface a clear error about which path is invalid and why, or (c) auto-create the missing directory.

**Impact:** The agent is completely unable to execute any shell commands, read files, run scripts, or interact with the filesystem. This blocks all code-repo work, file operations, and general tool usage on Windows.

**Version / Platform:**
- Platform: Windows 11 - 25H2 (OS Build 26200.8457)
- OpenHuman version: v0.57.13

## Solution (suspected)

The root cause is likely a mismatch between `workspace_dir` and `action_dir`:

1. `workspace_dir` defaults to `~/.openhuman/users/<id>/workspace` — this path is auto-created at startup and is valid.
2. `action_dir` defaults to `~/OpenHuman/projects` — this path is the shell tool's CWD. If it doesn't exist (e.g. the user never created it, or the tilde expansion produces a malformed Windows path like `C:\Users\Harry/OpenHuman/projects` with mixed separators), `cmd.current_dir()` will fail with `ERROR_DIRECTORY`.
3. The shell tool does NOT check if `action_dir` exists before setting it as CWD — it just passes it to `cmd.current_dir()` and lets the OS error bubble up as a generic "Failed to execute command" message.
4. The error message is opaque — it doesn't tell the user WHICH directory is invalid or why.

Related: Issue #2379 (closed) fixed a similar "program not found" problem on Windows by adding Windows env vars to the allowlist. This is a different failure mode (invalid directory path vs missing executable).

## Acceptance criteria

- [ ] **Repro gone** — Shell commands no longer fail with `ERROR_DIRECTORY` on Windows when `action_dir` doesn't exist
- [ ] **Clear error messaging** — If `action_dir` is invalid, the error message tells the user which path is the problem and suggests a fix (e.g. "The configured action_dir 'X' does not exist. Set a valid path in Settings → Agent → Working Directory or create the directory.")
- [ ] **Auto-creation or fallback** — Either auto-create `action_dir` at startup (like `workspace_dir` is), or fall back to a valid temp directory with a warning
- [ ] **Diff coverage ≥ 80%**

## Related

- Issue #2379 — Windows "program not found" (env allowlist fix, closed)
- Issue #3240 — Make action_dir editable in Settings UI (closed, PR #3305)
- Issue #2410 — Failed to initialize memory_tree schema: disk I/O error (workspace path issue)
- CLAUDE.md documents the workspace_dir / action_dir split but the shell tool's error handling doesn't account for a missing action_dir

sanil-23:
## Summary

- Create the agent's **action sandbox + default projects home** (`~/OpenHuman/projects`) on the *always-run* `bootstrap_core_runtime` boot, not just inside the integration-gated `start_channels` — fixes web-chat-only desktop cores (fresh installs with no messaging integrations) that never had a working directory created (#3353, RC-A).
- Make `~/` expansion **separator-correct**: build the path component-by-component so Windows gets `C:\Users\Harry\OpenHuman\projects`, not the mixed-separator `C:\Users\Harry/OpenHuman/projects` that `CreateProcessW` rejects with `ERROR_DIRECTORY` (os error 267) (#3353, RC-B).
- Add a **CWD guard** (`config::ensure_usable_cwd`) wired into both the unsandboxed `NativeRuntime::build_shell_command` and `sandbox::execute_in_sandbox`, so a missing/bad `action_dir` surfaces an actionable, path-naming error instead of an opaque OS error at spawn time.
- Make `SecurityPolicy::expand_tilde` delegate to the new `config::expand_tilde` so policy and config expand paths byte-for-byte identically (single source of truth).
- New unit tests cover both root causes and all three fixes (tilde expansion, dir creation + trusted root, CWD guard happy/edge paths).

## Problem

On Windows, a fresh desktop install where the user only uses **web chat** (no messaging integrations connected) could not run any agent shell-family tool (`shell` / `node_exec` / `npm_exec`). Two independent root causes:

**RC-A — `action_dir` never created on the web-chat-only boot.** The default projects home / action sandbox (`~/OpenHuman/projects`) was created only inside `start_channels`, which is gated on having messaging integrations. The always-run `bootstrap_core_runtime` path (used by web-chat-only desktop cores) skipped it entirely, so the directory the shell tool sets as `current_dir` never existed. Spawning a process with a non-existent CWD fails with `ERROR_DIRECTORY` (os error 267) on Windows / `ENOENT` on Unix.

**RC-B — mixed-separator `~/` expansion.** `expand_tilde` used `format!("{}/{rest}", home.display())`. With `rest = "OpenHuman/projects"`, on Windows this produced `C:\Users\Harry/OpenHuman/projects` — a path mixing `\` (from `home`) and `/` (literal in the format string and inside `rest`). `CreateProcessW` rejects such a path as a working directory with `ERROR_DIRECTORY` (os error 267). Note `home.join(rest)` would *not* fix this either, since the embedded `/` inside `rest` is preserved verbatim — the path has to be assembled component-by-component.

Together these meant the very first agent shell command on a clean Windows install failed with an opaque OS error and no actionable message.

## Solution

**Fix 0 — idempotent `config::ensure_agent_dirs`.** Extracted the projects-dir / action-dir creation + trusted-root registration out of `start_channels` into a single idempotent `ensure_agent_dirs(&mut Config)`. It is now called from **both** boot paths: `bootstrap_core_runtime` (always-run, before the live `SecurityPolicy` is built so the trusted root is reflected in `from_config`) and `start_channels` (unchanged behavior; the helper is a no-op the second time). This closes RC-A for web-chat-only cores.

**Fix 1 — component-wise `config::expand_tilde`.** Rewrote expansion to push each `/`-split component onto the home `PathBuf`, so the result uses the platform-native separator throughout. This is now the single source of truth; `SecurityPolicy::expand_tilde` delegates to it. Closes RC-B.

**Fix 2 — `config::ensure_usable_cwd` CWD guard.** A small stat-only (happy path) helper that verifies the working directory exists (defensively `create_dir_all` if missing — covers a dir deleted after launch) and resolves to a directory, otherwise returns a descriptive error naming the offending path and pointing at *Settings → Agent access → Working directory*. Wired into `NativeRuntime::build_shell_command` (covers all three shell-family tools, which route through it) and `sandbox::execute_in_sandbox` (parity across every sandbox backend). This converts opaque OS error 267 / ENOENT at spawn time into an actionable message — defense-in-depth on top of Fix 0/1.

### Verification note

The Windows-specific repro (mixed-separator path → `ERROR_DIRECTORY`) is **verified by reasoning**, not by CI: CI runs on Linux, where `/` is the native separator and the mixed-separator path is still valid, so the failure does not reproduce there. The new tests assert the separator-correctness invariant (`expand_tilde_has_no_mixed_separators_on_windows`) and the dir/CWD behavior in a platform-portable way. **@-reporter: please confirm the fix on a real Windows machine** — a fresh web-chat-only install should now run agent shell commands without the os error 267.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — new tilde-expansion, `ensure_agent_dirs`, and `ensure_usable_cwd` tests cover happy + edge/failure paths (uncreatable CWD, file-as-dir).
- [x] **Diff coverage ≥ 80%** — changed Rust lines are exercised by the new unit tests in `config/ops_tests.rs`, `security/policy_tests.rs`, and `agent/host_runtime.rs` test module.
- [x] Coverage matrix updated — `N/A: behaviour-only / bug-fix change`, no new feature row.
- [x] All affected feature IDs from the matrix are listed — `N/A`, no matrix feature touched.
- [x] N/A: Manual smoke checklist — no release-cut UI surface; behavior is core boot/dir creation.
- [x] Linked issue closed via `Closes #3353` in `## Related`.

## Impact

- **Platform:** Desktop. The user-visible bug was **Windows-only** (mixed-separator path + uncreated dir); the `action_dir`-creation fix (RC-A) also benefits Linux/macOS web-chat-only cores, which previously relied on `start_channels` for the dir.
- **Security:** No change to the path-hardening / workspace-internal denylist. `ensure_agent_dirs` registers `~/OpenHuman/projects` as a `ReadWrite` trusted root exactly as `start_channels` did before — same grant, just reachable from both boot paths. The internal workspace denylist is untouched.
- **Performance:** Negligible — idempotent `create_dir_all` at boot and a stat-only check on the shell happy path.
- **Migration/compat:** None. Idempotent; existing installs with the dir already present and the trusted root already granted are left untouched.

## Related

- Closes: #3353
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3353-windows-action-dir-tilde`
- Commit SHA: `4695ae92b397fcdd3c33a33191d14c6a6fee8b73`

### Validation Run

- [x] N/A: `pnpm --filter openhuman-app format:check` — no `app/` TS changes.
- [x] N/A: `pnpm typecheck` — no `app/` TS changes.
- [x] Focused tests: `cargo test` for `config::ops`, `security::policy`, `agent::host_runtime` modules.
- [x] Rust fmt/check (if changed): `cargo fmt` + `cargo check --manifest-path Cargo.toml`.
- [x] N/A: Tauri fmt/check — no `app/src-tauri` changes.

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` Windows repro not reproducible on Linux CI — see Verification note; reporter confirmation requested.

### Behavior Changes

- Intended behavior change: agent shell tools work on fresh web-chat-only desktop installs (esp. Windows); bad `action_dir` yields an actionable error.
- User-visible effect: no more opaque `ERROR_DIRECTORY` (os error 267) on first agent shell command; descriptive message pointing at Settings when the working dir is unusable.

### Parity Contract

- Legacy behavior preserved: `start_channels` still creates the same dirs + trusted root (now via the shared helper); `SecurityPolicy::expand_tilde` semantics preserved (delegates to the single source of truth).
- Guard/fallback/dispatch parity checks: CWD guard applied identically in the unsandboxed and sandboxed execution paths.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent directories (projects/sandbox) are now auto-created at startup, preventing failures from missing folders.
  * Working-directory checks now give clearer, descriptive errors and auto-create valid-but-missing dirs.
  * Path/tilde expansion is now consistent across the app.
* **Tests**
  * Added regression tests covering tilde expansion, directory setup, and working-directory validation.

### DIFF
--- a/src/core/jsonrpc.rs
+++ b/src/core/jsonrpc.rs
@@ -2002,7 +2002,7 @@ pub async fn bootstrap_core_runtime(host_kind: crate::core::types::HostKind) {
     // `embedded_core` derived from host_kind so the rest of the function (which
     // already keys behavior off the boolean) stays unchanged.
     let embedded_core = host_kind.is_desktop_shell();
-    let cfg = match crate::openhuman::config::Config::load_or_init().await {
+    let mut cfg = match crate::openhuman::config::Config::load_or_init().await {
         Ok(cfg) => cfg,
         Err(e) => {
             log::error!("[runtime] Failed to load config for socket manager: {e}");
@@ -2059,6 +2059,17 @@ pub async fn bootstrap_core_runtime(host_kind: crate::core::types::HostKind) {
              spawn_subagent will be unavailable until restart"
         );
     }
+
+    // --- Agent sandbox + projects dirs ---
+    // Create the action sandbox + default projects home and register the
+    // projects dir as a ReadWrite trusted root BEFORE building the live policy
+    // below (so the trusted root is reflected in `from_config`). This is the
+    // always-run boot for web-chat-only desktop cores; without it a fresh
+    // install with no messaging integrations leaves `~/OpenHuman/projects`
+    // uncreated and every shell-tool `current_dir` fails with ERROR_DIRECTORY
+    // (os error 267) on Windows / ENOENT on Unix (#3353, RC-A). Idempotent — a
+    // later `start_channels` calls the same helper.
+    crate::openhuman::config::ensure_agent_dirs(&mut cfg).await;
 
     // --- Live SecurityPolicy ---
     // Install the process-global live policy on the always-run serve boot, not
--- a/src/openhuman/agent/host_runtime.rs
+++ b/src/openhuman/agent/host_runtime.rs
@@ -93,6 +93,13 @@ impl RuntimeAdapter for NativeRuntime {
             c.arg("-lc").arg(command);
             c
         };
+        // Validate the CWD up front so a missing/bad action_dir produces an
+        // actionable message naming the path, instead of an opaque OS error 267
+        // (ERROR_DIRECTORY) from CreateProcessW on Windows / a raw ENOENT on
+        // Unix when the process is spawned. Covers all three shell-family tools
+        // (shell / node_exec / npm_exec) since they all route through here.
+        // (#3353, Fix 2)
+        crate::openhuman::config::ensure_usable_cwd(workspace_dir)?;
         cmd.current_dir(workspace_dir);
         Ok(cmd)
     }
@@ -323,5 +330,47 @@ mod tests {
         // A clean pipeline still succeeds.
         let mut ok = rt.build_shell_command("true | true", &dir).unwrap();
         assert!(ok.status().await.unwrap().success());
+    }
+
+    /// #3353: a CWD that can't be made usable (here: a path *under an existing
+    /// file*, which `create_dir_all` cannot create) must yield a descriptive,
+    /// path-naming error from `build_shell_command` instead of an opaque OS
+    /// error 267 (ERROR_DIRECTORY) at spawn time.
+    #[test]
+    fn native_shell_command_rejects_uncreatable_cwd_with_clear_error() {
+        let rt = NativeRuntime::new();
+        let tmp = tempfile::tempdir().unwrap();
+        let parent_file = tmp.path().join("a-file");
+        std::fs::write(&parent_file, b"x").unwrap();
+        let bad_cwd = parent_file.join("child"); // parent is a file → uncreatable
+
+        let err = rt
+            .build_shell_command("echo hi", &bad_cwd)
+            .expect_err("an uncreatable CWD must be rejected up front");
+        let msg = err.to_string();
+        assert!(
+            msg.contains("could not be created"),
+            "expected an actionable message, got: {msg}"
+        );
+        assert!(
+            msg.contains(&bad_cwd.to_string_lossy().to_string()),
+            "error should name the offending path: {msg}"
+        );
+    }
+
+    /// A valid-but-missing CWD is defensively created (covers a dir deleted
+    /// after launch), so the command builds successfully and runs there.
+    #[test]
+    fn native_shell_command_creates_missing_cwd() {
+        let rt = NativeRuntime::new();
+        let tmp = tempfile::tempdir().unwrap();
+        let missing = tmp.path().join("nested").join("workdir");
+        assert!(!missing.exists());
+
+        let command = rt
+            .build_shell_command("echo hi", &missing)
+            .expect("missing CWD should be auto-created");
+        assert!(missing.is_dir(), "CWD should have been created");
+        assert_eq!(command.as_std().get_current_dir(), Some(missing.as_path()));
     }
 }
--- a/src/openhuman/channels/runtime/startup.rs
+++ b/src/openhuman/channels/runtime/startup.rs
@@ -253,47 +253,11 @@ pub async fn start_channels(mut config: Config) -> Result<()> {
 
     let runtime: Arc<dyn host_runtime::RuntimeAdapter> =
         Arc::from(host_runtime::create_runtime(&config.runtime)?);
-    // Ensure the agent's default projects home (~/OpenHuman/projects) exists and
-    // is a read-write trusted root, so the coding agent creates/edits projects
-    // there freely — distinct from the hidden internal workspace dir. A user who
-    // has already granted it (or any other root) is left untouched.
-    {
-        use crate::openhuman::security::{TrustedAccess, TrustedRoot};
-        let projects_dir = crate::openhuman::config::default_projects_dir();
-        if let Err(e) = tokio::fs::create_dir_all(&projects_dir).await {
-            tracing::warn!(
-                dir = %projects_dir.display(),
-                error = %e,
-                "[startup] could not create default projects dir"
-            );
-        }
-        let projects_path = projects_dir.to_string_lossy().to_string();
-        if !config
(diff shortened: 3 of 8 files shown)


sanil-23:
Addressed all three CodeRabbit findings in d49a3a7e:

1. **Docker container-side cwd (sandbox/ops.rs)** — `ensure_usable_cwd` is now applied per-backend: `working_dir` for `None`/`Local` (genuinely host paths), and the host-side mount source `policy.workspace_root` for `Docker` (whose `working_dir` is the container path, e.g. `/workspace`, and must not be stat'd/created on the host). Good catch.

2. **PII in startup logs (config/ops.rs)** — added a `redact_home` helper that swaps the home-dir prefix for `~`, so logs keep their diagnostic value (`~/OpenHuman/projects`) without leaking the OS username/full home path. Wired into all three startup log sites. Covered by a new unit test.

3. **Env restore in test (config/ops_tests.rs)** — the test now captures the prior `OPENHUMAN_PROJECTS_DIR` with `var_os` and restores it (set/remove) in cleanup instead of unconditionally removing it.

Local: `cargo fmt`, `cargo check`, and the changed test modules (config::ops 74, security::policy 154, agent::host_runtime 6, sandbox::ops 13) all pass.

sanil-23:
@graycyrus this one's green and ready for review/merge whenever you have a moment 🙏

- ✅ All checks pass — Coverage Gate (diff-cover ≥ 80%), Rust Core/Tauri Coverage, Rust + Frontend Quality, Rust E2E, all 4 Playwright lanes + Gate, PR Submission Checklist, i18n.
- ✅ CodeRabbit approved (all 3 findings addressed in d49a3a7e: Docker container-side cwd, home-path redaction in startup logs, test env restore).

Fixes #3353 (web-chat-only desktop cores never created `action_dir` + mixed-separator `~/` expansion → `ERROR_DIRECTORY` os error 267 on Windows). The Windows repro is verified-by-reasoning since CI is Linux — a Windows confirmation from the reporter would be a nice bonus, but isn't blocking. Merge is maintainer-gated (base branch protection).

Issue #3386, opened 2026-06-04
Budget-exhausted on main chat surfaces as 'empty response' instead of typed budget error


CodeGhost21:
## Summary

When the OpenHuman cloud inference budget is exhausted, the **main chat call** can still surface as a generic "model returned an empty response, try a different model" instead of the actionable budget-exhausted copy added by #3121 — sending the user toward the wrong remediation (swap model / check local provider) when the real fix is "top up credits, or switch routing to a local model".

## Problem

Sentry issue **TAURI-RUST-4JX** (5219 on the self-hosted Sentry, `domain=agent operation=run_single`, `channel=web_channel`) has **2623 events from a single user** on release `openhuman@0.56.0+e8968077aeb5`. The Sentry suppression from #2790 lands in ≥ v0.57.4 so the noise will stop on its own — this issue is about the underlying user-experience bug those events represent.

Latest event breadcrumbs (2026-06-04 15:19:15Z, abridged):

```
15:19:12.866  [llm_provider] api_error budget-exhausted 400 — not reporting to Sentry  (provider=inference:summarization-v1)
15:19:12.868  [tree_global::digest] summarise failed … {"success":false,"error":"Insufficient budget"} — using fallback
15:19:14.133  [llm_provider] api_error budget-exhausted 400 — not reporting to Sentry  (provider=inference:summarization-v1, retry)
15:19:14.134  [memory_tree::extract::llm] chat provider=inference:summarization-v1 failed … Insufficient budget
15:19:15.005  [stream]      OpenHuman aggregated text_chars=0 thinking_chars=0 tool_calls=0
15:19:15.005  [agent_loop]  provider response i=1 elapsed_ms=341 text_chars=0 native_tool_calls=0
15:19:15.007  [agent_loop]  provider returned an empty final response (i=1, …) — surfacing as error
```

The two budget-exhausted 400s in the seconds before the chat call are `inference:summarization-v1` (memory tree background work), but the **main chat completion** then finishes in **341 ms** with `text_chars=0 thinking_chars=0 tool_calls=0` — fast enough to be a backend rejection, but no `[llm_provider] … budget-exhausted 400` breadcrumb fires for it. The agent loop bails at `src/openhuman/agent/harness/engine/core.rs:394-405` with `AgentError::EmptyProviderResponse`, which renders to the user-facing string `"The model returned an empty response. Please try again."` from `src/openhuman/agent/error.rs:92-97`.

Two ways the main chat call can hit this:

[...]

Issue #3104, opened 2026-05-31
Cascade error on agentic task — Plan, Run Code, and Tools Agent all failing in sequence


Giri-Aayush:
User ygkang4354 submitted a screenshot on 19/05/2026 showing a full cascade of agent step failures with no accompanying description. The screenshot shows the following steps all marked with error: Plan, Run Code (×6), Tools Agent (×2). The app displayed a generic error message: "Something went wrong. Please try again. This error has been reported."

No steps to reproduce were provided. Team requested reproduction steps on 21/05/2026 — no response from user as of 25/05/2026 despite a follow-up ping. Root cause unknown without logs or repro steps.

Screenshot attached (see attachments). The pattern of repeated Run Code and Tools Agent failures suggests either a retry loop hitting a persistent error, or an upstream model/tool dispatch failure causing the entire agent chain to collapse.

---
**Reported by:** @ygkang4354
**Discord ticket:** https://discord.com/channels/1470856511193616498/1506495818838970519
**Priority:** medium
**Attachments (1):** image (1).webp — ask support team for files

CodeGhost21:
The cascade pattern here (Plan + Run Code ×6 + Tools Agent ×2 all error, generic "Something went wrong") looks like a likely duplicate of #3193, which is the canonical thread for this exact failure mode.

**Overlap with known causes:**

| Symptom | Cause |
|---|---|
| Run Code ×6 all failing | Tracked in #3193 — `run_code` 404 when a completion-only model is bound to the `coding` role; orchestrator retries → cascade. PR #3211 (open) adds an actionable error and an unmistakable failure envelope so the orchestrator can't narrate fabricated success. |
| Tools Agent ×2 failing | #3193 also notes `delegate_tools_agent` consistently returns `400 Insufficient budget` — the *classification* side of that 400 is now handled by #3121 (merged 2026-06-01, commit e9c38ea61), which surfaces it as `budget_exhausted` with actionable copy instead of the generic "Something went wrong" apology. |
| "Something went wrong" copy | Pre-#3121 path for the budget side; pre-#3211 path for the completion-only 404 side. |

**What's still blocked:** PR #3211 explicitly deferred a few sub-symptoms from #3193 ("`delegate_tools_agent` 400 Insufficient budget" on no-op calls, `spawn_parallel_agents` non-determinism, whether `parallel_tools=true` is honored) because the original reporter's transcript wasn't available.

[...]

Issue #3092, opened 2026-05-31
Something went wrong" error on app chat (multiple users reported the same)


Giri-Aayush:
Multiple users including paid ones are reporting a generic error on the OpenHuman app that reads "Something went wrong. Please try again. This error has been reported." The error appears in-app chat screen.

---
**Reported by:** viktoriia_67189
**Discord ticket:** https://discord.com/channels/1470856511193616498/1508832575630741605
**Priority:** urgent
**Attachments (1):** 1.png — ask support team for files

CodeGhost21:
Fixed by #3199 (merged 2026-06-02, commit 89380db65), which closes the canonical thread #3119 and explicitly calls out #3092 in its Related section ("Also fixes the surface of #3092 and most of #3104 — same generic-error symptom").

**Root cause:** `classify_inference_error` (`src/openhuman/channels/providers/web_errors.rs`) is a first-match-wins substring chain whose final `else` returned the generic "Something went wrong" copy. A whole family of real, individually-actionable failures had no matching arm and fell through with zero guidance — which is why the symptom hit paid users too (the failures weren't gated by plan tier).

**What #3199 added** — three new classifier arms, each with actionable copy + correct `retryable`/`source`:

- **`empty_response`** — the model returned an empty body (the dominant Sentry signature). Retryable; copy points at switching model / checking local provider in Settings → AI → LLM.
- **`provider_request_rejected`** — un-claimed provider 400 / 404 / 422. Non-retryable; surfaces the upstream reason via the existing secret-scrubbed, length-capped `with_provider_detail`. Ordered **after** the more-specific 4xx arms (invalid temperature, stale model pin, model-not-found, DeepSeek `reasoning_content`) so their specific verdicts still win.
- **`capability_unsupported`** — image sent to a text-only model. Non-retryable; points at picking a vision-capable model.

This is on top of the already-merged #3121 (budget-exhausted 400 → actionable top-up / switch-routing copy, commit e9c38ea61).

**Still falling through to the generic catch-all** (documented as follow-up, not silently capped by #3199):
- `"error decoding response body"` (transport/parse failure)
- `"unknown embedding provider: local"` (embeddings misconfig)

If @viktoriia_67189's screenshot reveals one of those two, it's a separate follow-up — otherwise the next release build will replace the opaque message with cause-specific copy.

Suggest closing this as resolved-in-next-release and asking the reporter to re-test on the next build.

PR #3415, opened 2026-06-05
fix(chat): drop "local provider" misdirect from empty-response copy + log empty 2xx streams (#3335)

### LINKED ISSUE #3335: Default Model Returning “The model returned an empty response” During Chat
## Summary

Encountering repeated errors while chatting in the OpenHuman app.
Instead of generating a reply, the app shows:

`The model returned an empty response. Try a different model or check your local provider in Settings → AI → LLM.`

## Problem

- Happens during normal chat usage
- Occurs even when using Managed settings
- Multiple prompts/messages fail consecutively
- Creating a new chat does not resolve the issue

**Expected behavior:**
Model should generate a valid response normally

**Actual behavior:**
Chat fails and returns the empty response error message repeatedly

Screenshot attached for reference.

Image

CodeGhost21:
## Summary

- Replace the misleading **"Try a different model or check your local provider in Settings → AI → LLM"** copy on the empty-response classifier — that sent **Managed-route users toward a remedy that doesn't exist for them**, because there is no local provider on the Managed route. The new copy explicitly names the credits / billing remedy alongside the model / provider-config remedies (chat surface), and a tighter ≤120-char form for the cron notification drawer.
- Add a forensic `warn`-level log in `stream_native_chat` for the exact upstream shape that triggers `AgentError::EmptyProviderResponse` — a streaming chat call that returns **HTTP 2xx with zero text, zero thinking, and zero tool calls**. Captures `elapsed_ms`, `sse_chunks_parsed`, `raw_bytes_received`, `has_usage`, `has_openhuman_meta` so production data can disambiguate the three plausible upstream root causes (credit exhaustion served as 200 + empty body, upstream provider stall, real degenerate model output).
- Strictly additive — no behavioural change to request paths, no new Sentry suppression. The diagnostic log fires only on the empty-2xx case (otherwise the existing aggregated log carries the same fields at `info`).

## Problem

Issue #3335 (`iamrhn`): on Managed settings, every chat turn surfaces:

> *"The model returned an empty response. Try a different model or check your local provider in Settings → AI → LLM."*

The "local provider" remedy is nonsensical for Managed users. The Sentry data confirms this is real and broad:

- **TAURI-RUST-4JX**: 2,633 events, 55 distinct clients in 14 days. Releases include current `0.57.13` (PR #2790's Sentry suppression silenced the events but the underlying user-visible bug remains).
- Latest event breadcrumbs (Managed route, `reasoning-v1`, web channel):

  ```
  [provider:OpenHuman] outbound chat/completions -> https://api.tinyhumans.ai/openai/v1/chat/completions
  [stream]   OpenHuman POST … (stream=true, tools=21)
  …~7 s gap, no [stream] error breadcrumb, no [llm_provider] api_error breadcrumb…
  [agent]    provider responded — parsed tool_calls=0 text_chars=0
  [agent_loop] provider returned an empty final response (i=1, no text, no tool calls) — surfacing as error
  ```

  HTTP status was 2xx (no `streaming API error` line), so none of the existing classifiers (`is_budget_exhausted_http_400`, `is_provider_access_policy_denied_http_403`, …) at `compatible.rs:1028-1060` fire. The streaming aggregator builds a `ChatResponse` with `content: None`, the agent loop returns `AgentError::EmptyProviderResponse`, and the chat surface renders the wrong-remediation copy.

This is exactly the shape #3386 (team-filed) identifies as path **(b)**: the OpenHuman managed backend returns 200 + empty SSE under credit exhaustion / upstream stall instead of a typed 400 + `Insufficient budget`. The deeper fix lives in `tinyhumansai/backend`; this PR addresses the **two client-side wins that don't require the server change**.

## Solution

**(1) Copy fix — chat surface (`web_errors.rs::classify_inference_error` empty_response arm).** Replace the `"Try a different model or check your local provider"` text with a three-remedy form that's accurate for **all** providers: *"This usually means your inference credits are exhausted (Settings → Billing), the upstream model is temporarily unhealthy, or your provider configuration is rejecting the request (Settings → AI → LLM). Try one of those, or pick a different model."* New regression test `classify_inference_error_empty_response_copy_names_billing_remedy_and_drops_local_provider_misdirect` locks in: contains `Settings → Billing`, contains `Settings → AI → LLM`, contains `different model`, does NOT contain `local provider`, and `provider` stays `None`. Existing test `classify_inference_error_empty_response_is_actionable_and_retryable` continues to pass.

**(2) Copy fix — cron drawer (`cron/scheduler.rs::agent_error_to_user_message`).** Shorter form for the drawer's ≤120-char convention: *"Empty model response. Out of credits (Settings → Billing) or try a different model in Settings → AI → LLM."* New regression test `agent_error_to_user_message_classifies_empty_provider_response_for_3335` mirrors the chat-side assertions; `EmptyProviderResponse` is added to the existing `agent_error_to_user_message_canned_strings_are_short` variants list to lock in the ≤120-char contract (was previously absent — fit incidentally, but nothing enforced it).

**(3) Diagnostic log — `compatible.rs::stream_native_chat`.** Right after the existing aggregated `info` log, emit a `warn` when `text_accum.is_empty() && thinking_accum.is_empty() && tool_call_count == 0`:

```
[stream] {name} empty 2xx stream — model={..} elapsed_ms={..} sse_chunks={..} raw_bytes={..} has_usage={..} has_openhuman_meta={..}
```

`stream_started_at: Instant`, `sse_chunks_parsed: usize`, and `raw_bytes_received: usize` are added as append-only counters (never read by the request path). The signal disambiguates:

- `elapsed_ms` small + `sse_chunks=0` + `raw_bytes` small → backend closed the SSE near-immediately (credit reject or auth)
- `elapsed_ms` large + `sse_chunks=0` + `raw_bytes=0` → upstream stall / timeout
- `sse_chunks>0` + `has_usage=true` + accumulators empty → backend streamed metadata-only chunks (tokens counted, no content delivered)

**(4) Deliberately deferred — the in-session budget correlator from #3386.** `EmptyProviderResponse`'s flattened Display carries no `" API error"` infix for `extract_provider_name` to anchor on, so the classifier here cannot tell which provider was used without plumbing the typed `provider: Option<String>` through the AgentError. That plumbing affects ~10 string-match test/observability sites and balloons the scope. The diagnostic log is the cheapest path to the production data that decides whether the plumbing is worth doing — once we have evidence on which path is dominant (200+empty under credit exhaustion vs. upstream stall vs. true degenerate output), the next iteration of the empty-response arm can be provider-aware.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — three new tests: `classify_inference_error_empty_response_copy_names_billing_remedy_and_drops_local_provider_misdirect` (web chat copy), `agent_error_to_user_message_classifies_empty_provider_response_for_3335` (cron drawer copy), plus `EmptyProviderResponse` added to `agent_error_to_user_message_canned_strings_are_short` variants (locks in the ≤120-char contract that was previously unenforced).
- [x] **Diff coverage ≥ 80%** — every new/changed line in `web_errors.rs` and `cron/scheduler.rs` is exercised by the three new regression tests; the `compatible.rs` diagnostic log is observability-only (no behavioural branching) and covered by the existing streaming-chat test surface that drives it. Run `pnpm test:rust` locally to verify.
- [x] Coverage matrix updated — `N/A: behaviour-only change` (error-message copy + observability log, no new feature ID).
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: behaviour-only change, no matrix rows touched`.
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — `N/A: tests are pure unit-level over the classifier and the AgentError variant; no provider HTTP exercised`.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — `N/A: error-copy + log only, no UI flow or release-cut surface affected`.

## Impact

- **Desktop (chat surface)**: Managed-route users hitting empty-response failures now see an actionable, non-misleading message that names `Settings → Billing` as the most likely fix (per Sentry / #3386 evidence) without losing the existing `Settings → AI → LLM` and model-switch remedies. Self-hosted users see the same three-remedy framing — strictly more accurate than the prior "local provider" claim for users on cloud providers.
- **Desktop (cron notifications)**: Same behavioural change, shorter copy to fit the ≤120-char drawer-render budget. Now enforced by the existing length test.
- **Observability**: Net +1 `warn` log per streaming chat turn that returns 2xx-with-empty-body. This is the failure case being investigated; non-empty streams are unaffected. Lands in Sentry breadcrumbs even after `AgentError::skips_sentry()` (#2790) silences the parent event, unblocking the data needed to decide whether the in-session budget correlator from #3386 is the right next move or whether the fix has to be server-side.
- **Performance / security**: No request-path behaviour change. The added counters are stack-local `usize` increments. The diagnostic log redacts nothing new — provider name, model name, and counts are all already in the existing aggregated `info` log; only the byte total and elapsed ms are net-new and both are content-free.
- **Migration / compatibility**: None. Pure string + observability change.

## Related

- Closes: #3335
- Related: #3386 (the team-filed deep-dive that this PR addresses without requiring the per-session budget correlator), #3104 (cascade-error reports likely sharing the same empty-stream root cause), #2790 (the Sentry suppression that silenced TAURI-RUST-4JX without fixing the user-visible bug), #3199 (the empty-response arm being modified), #3121 (the budget-exhausted copy whose remedy this copy now points users toward), #3092 (closed parent issue for the chat-error cluster).
- Follow-up PR(s)/TODOs: Once the diagnostic log produces 24–48 h of production data, decide between (a) opening a backend issue against `tinyhumansai/backend` to start returning `400 + Insufficient budget` on streaming under credit exhaustion (the right server-side fix), and (b) plumbing `provider: Option<String>` through `AgentError::EmptyProviderResponse` so the classifier can be provider-aware (the #3386 in-session correlator approach).

---

### Linear Issue

- Key: N/A — GitHub issue #3335, not Linear-tracked.
- URL: N/A

### Commit & Branch

- Branch: `fix/3335-managed-empty-response-diagnostic-and-copy`
- Commit SHA: `7064cbfee`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no frontend / TS files changed`
- [x] `pnpm typecheck` — `N/A: no frontend / TS files changed`
- [x] Focused tests: `cargo test --lib empty_response` (10 pass), `cargo test --lib empty_provider` (5 pass), `cargo test --lib cron::scheduler::tests::agent_error` (11 pass), `cargo test --lib compatible::tests` (153 pass).
- [x] Rust fmt/check (if changed): `cargo fmt --manifest-path Cargo.toml` + `cargo check --manifest-path Cargo.toml --lib` — both clean (only pre-existing unrelated warnings).
- [x] Tauri fmt/check (if changed): `N/A: no app/src-tauri files changed`

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: User-facing error copy for `AgentError::EmptyProviderResponse` no longer claims a "local provider" exists; new copy names `Settings → Billing` (credits) alongside `Settings → AI → LLM` (provider config) and the model-switch fallback.
- User-visible effect: Managed-route users who hit empty-response failures now receive remedy text that matches the most likely actual cause (credit exhaustion, per #3386 evidence). Self-hosted users see the same three-remedy framing, which remains accurate for their configurations.

### Parity Contract

- Legacy behavior preserved: `AgentError::EmptyProviderResponse` Display, `classify_inference_error` error_type / source / retryable / provider fields, `agent_error_to_user_message` dispatch matrix, the `Settings → AI → LLM` deep-link reference. Existing test `classify_inference_error_empty_response_is_actionable_and_retryable` continues to pass unchanged.
- Guard/fallback/dispatch parity checks: The empty-response arm in `web_errors.rs` keeps the same `error_type: "empty_response"`, `source: "agent_loop"`, `retryable: true`, `retry_after_ms: None`, `provider: None`, `fallback_available: None` — only the message string changes. The cron `agent_error_to_user_message` dispatch matrix is unchanged; only the `EmptyProviderResponse` branch's returned string changes.

- Duplicate PR(s): None — no other open PR references #3335.
- Canonical PR: This one.
- Resolution: N/A (no prior PR to supersede).

## Release Notes

* **Bug Fixes**
  * Improved error messaging when model returns an empty response, now guiding users to check inference credits (Settings → Billing), try a different model, or verify LLM provider configuration (Settings → AI → LLM).

* **Improvements**
  * Added enhanced diagnostic logging for empty response scenarios to better support troubleshooting.


oxoxDev:
@CodeGhost21 bro merge conflicts here

src/openhuman/inference/provider/compatible_stream_native.rs
```diff
@@ -402,6 +416,42 @@ impl OpenAiCompatibleProvider {
...
+        // `AgentError::skips_sentry()` (PR #2790) silences the parent
+        // event. Correlate by elapsed_ms (fast == reject, slow ==
+        // stall), sse_chunks (0 == no SSE at all, >0 == backend
+        // streamed metadata-only chunks), has_usage (the upstream
+        // counted tokens but delivered no content), and
+        // has_openhuman_meta (managed backend reported routing info).
+        if text_accum.is_empty() && thinking_accum.is_empty() && tool_call_count == 0 {
+            let elapsed_ms = stream_started_at.elapsed().as_millis() as u64;
+            log::warn!(
+                "[stream] {} empty 2xx stream — model={} elapsed_ms={} sse_chunks={} raw_bytes={} has_usage={} has_openhuman_meta={}",
```
YellowSnnowmann: Two non-blocking observations on this diagnostic — the change is correct and strictly additive, just flagging for completeness given the PR's goal is to capture the empty-response signal:

1. **SSE-path-only coverage.** This warn fires only inside the SSE branch. A 2xx response with a non-SSE content-type early-returns via `parse_native_response` (~L149/L157) and never reaches this guard. If empty managed responses can arrive as buffered `application/json` (plausible under the #3386 credit-exhaustion path), those will silently miss the forensic log — i.e. the very case you're trying to disambiguate. Worth either extending the check to that branch or noting in the PR that the diagnostic is SSE-only.

2. **`raw_bytes` semantics.** `bytes_stream()` yields reqwest's already-decompressed bytes, so `raw_bytes=` is decoded body size, not wire bytes. An operator reading the breadcrumb might misread a small value as a compression artifact. A `decoded_bytes` key (or a one-word comment) would remove the ambiguity.

Verified the rest: parity contract holds (only `message` changed; `provider:None` justification is correct — `EmptyProviderResponse` Display has no `" API error"` infix for `extract_provider_name`), counters are observability-only with no move/borrow issues, and `log::warn!` does become a Sentry breadcrumb via the tracing layer (WARN → Breadcrumb).

CodeGhost21: Both addressed in 9d308cd36.

**1. SSE-path-only coverage** — fair catch. Added a sibling empty-2xx guard right after the JSON parse in the non-SSE branch (the `!is_sse` early-return at L147). When the first choice has empty content + empty reasoning + no tool_calls + no function_call, it emits:

```
[stream] {name} empty 2xx buffered JSON — model={..} elapsed_ms={..} body_bytes={..} has_usage={..} has_openhuman_meta={..}
```

Same disambiguation as the SSE branch (`elapsed_ms` small → backend reject, large → upstream stall; `has_usage` true → tokens counted but no content delivered). Now both upstream shapes the #3386 credit-exhaustion path could take are covered.

**2. `raw_bytes` semantics** — agreed it's ambiguous reading. Renamed the counter and log key to `body_bytes`. Caveat for context: this crate ships reqwest without `gzip` / `brotli` / `zstd` / `deflate` features (root `Cargo.toml`), so the count is wire bytes in this build — but an operator reading the breadcrumb has no way to know that, which is exactly the readability problem you flagged. Picked `body_bytes` over `decoded_bytes` so the name stays accurate even if compression features get enabled later, with a comment inline documenting the current no-decompression invariant.

M3gA-Mind (approved):
Approving after end-to-end review against the latest upstream/main (`0db261dcc`).

**What this fixes:** drops the misleading "check your local provider" remedy from the empty-response copy (false for Managed-route users, where no local provider exists; the common cause is credit exhaustion, #3386) and replaces it with provider-agnostic guidance that names the real remedies — billing/credits, model health, and provider config. Also adds observability-only `log::warn!` diagnostics on both the SSE and buffered-JSON empty-2xx paths (`elapsed_ms`, `sse_chunks`, `body_bytes`, `has_usage`, `has_openhuman_meta`) so a future iteration can make the copy provider-aware once production evidence exists.

**Verification:**
- Confirmed the change is NOT already upstream — `web_errors.rs` on main still carries the old "local provider" copy; none of the diagnostic code exists on main. Genuinely additive, not a duplicate.
- Merges cleanly (`MERGEABLE`), no conflicts.
- Prior review feedback (@YellowSnnowmann: SSE-only coverage + ambiguous `raw_bytes` naming) already addressed in `9d308cd36` (sibling buffered-JSON guard + `raw_bytes`→`body_bytes` rename).
- User-facing copy logic is unit-tested in both `web_tests.rs` and `scheduler_tests.rs`, including a length-contract guard for the notification drawer.
- CI fully green: Rust Quality (fmt/clippy), Rust E2E (mock backend), Rust Core Coverage, Coverage Gate (diff-cover ≥ 80%), CodeRabbit — all pass.

Correct, well-scoped, well-documented, low-risk (no request-path behavior change). LGTM.

PR #3416, opened 2026-06-05
perf(memory_tree): concurrently seal document-subtree sibling batches

mysma-9403:
## Summary

- Seal sibling batches within each document-subtree level concurrently (bounded `buffered(8)`) instead of one-at-a-time, so independent summarise + embed round-trips overlap.
- Tree shape stays byte-identical to the serial cascade: ordered `buffered` (not `buffer_unordered`) preserves positional batch order that the next level's `batch_by_count` depends on.
- Adds a concurrency-probe test asserting observed peak in-flight seals is in `[2, DOC_SUBTREE_SEAL_CONCURRENCY]` while the resulting doc-root (4 children, version_ms) is unchanged.

## Problem

`seal_document_subtree` cascades a document's chunks level-by-level into a single doc-root. Within one level the batches are disjoint child sets and `seal_explicit_children` touches no shared `(tree, level)` buffer and never advances the tree root — yet they were sealed in a strict serial loop, so every batch's summarise (LLM) + label-resolve + embed network round-trips ran back-to-back. For a wide document (many level-0 batches) this serialises N independent network RTTs.

## Solution

Replace the per-level serial `for batch in &batches` loop with a bounded-concurrency stream:

- `futures::stream::iter(batch_futures).buffered(DOC_SUBTREE_SEAL_CONCURRENCY).try_collect()` runs up to 8 sibling seals concurrently within a level.
- **Ordered** `buffered`, not `buffer_unordered`: the next level groups `next_ids` positionally via `batch_by_count`, so output order must match the serial cascade to keep the tree shape identical. The terminal level produces exactly one node, so `next_back()` is the doc-root — same as the old serial assignment.
- `try_collect` short-circuits on the first error, preserving the old `?`-on-first-failure semantics.
- Futures are built eagerly into a `Vec` (they don't run until polled) rather than lazily in `stream::iter(...).map(closure)`, which otherwise makes rustc fail higher-ranked `FnOnce` inference once the enclosing async fn is spawned.

Levels stay sequential (level N+1 consumes level N's output) so only siblings overlap, never cross-level fan-in. Blocking SQLite writes inside each seal still serialise on the single writer — the win is the overlapped network RTTs, not the DB writes. `DOC_SUBTREE_SEAL_CONCURRENCY = 8` bounds fan-out to stay within provider rate limits.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — added `seal_document_subtree_seals_sibling_batches_concurrently`; existing seal tests (single-chunk passthrough, force-seal, additive version, backlink repoint, fanout) cover edge/regression paths and stay green.
- [x] **Diff coverage ≥ 80%** — changed lines are the new loop body + new test, all exercised by the bucket_seal lib suite (`cargo test --lib bucket_seal`, 15 passed).
- [x] Coverage matrix updated — N/A: internal performance change, no feature added/removed/renamed.
- [x] All affected feature IDs from the matrix are listed under `## Related` — N/A: no feature IDs touched.
- [x] No new external network dependencies introduced — N/A: no network calls added; `futures` already a dependency.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A: no user-visible surface change.
- [x] Linked issue closed via `Closes #NNN` — N/A: no tracking issue.

## Impact

- **Runtime**: Rust core only (memory-tree document ingest path). No desktop/mobile/web/CLI surface change.
- **Performance**: wide documents seal their level-0 (and upper) sibling batches with up to 8-way overlapped summarise/embed round-trips; latency drops toward the slowest batch per level rather than the sum.
- **Compatibility**: tree shape, doc-root identity, and version semantics are unchanged (ordered collection); SQLite writes still serialise.

## Related

- Closes: N/A
- Follow-up PR(s)/TODOs: N/A

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: perf/seal-document-subtree-l0
- Commit SHA: 28db5b002b324ad01585fa482ef5a14ef3a57b0c

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: no `app/` changes; Rust-only.
- [x] Focused tests: `cargo test --manifest-path Cargo.toml --lib bucket_seal` → 15 passed, 0 failed.
- [x] Rust fmt/check (if changed): `cargo fmt` clean; `cargo check --lib` exit 0.
- [x] Tauri fmt/check (if changed): N/A: no Tauri shell changes (pre-push `pnpm rust:check` passed).

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: sibling document-subtree batches seal concurrently (bounded) instead of serially.
- User-visible effect: faster document ingest for wide documents; no change to produced tree content or shape.

### Parity Contract

- Legacy behavior preserved: ordered collection keeps tree shape, doc-root identity, and version_ms byte-identical to the serial cascade.
- Guard/fallback/dispatch parity checks: `try_collect` preserves `?`-on-first-error; levels remain sequential; SQLite single-writer serialisation unchanged.

- Duplicate PR(s): N/A
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

  * Accelerated document processing through parallel batch execution with resource safeguards.

* **Tests**
  * Added test validating parallel execution behavior and resource limits.


src/openhuman/memory_tree/tree/bucket_seal.rs
```diff
@@ -1012,6 +1013,17 @@ pub const MERGE_LEVEL_BASE: u32 = 1_000;
...
+/// Bound on how many sibling batches within a single document-subtree level
+/// are sealed concurrently. Each [`seal_explicit_children`] call is one
+/// independent summarise + embed round-trip pair, so overlapping them with a
+/// small pool collapses the serial `N × (LLM + embed RTT)` wall-time toward
+/// `N/CONCURRENCY × RTT` on the ingest path. Kept low: the SQLite write tx
+/// serialises on a single writer anyway, and networked summarise/embed
+/// providers (Voyage, cloud LLMs) rate-limit — so a large pool buys nothing
+/// and risks 429s. `buffered` (ordered) preserves the serial cascade's batch
+/// order, so the resulting tree shape is identical.
+const DOC_SUBTREE_SEAL_CONCURRENCY: usize = 8;
```
M3gA-Mind, line 1025: Re: CodeRabbit suggestion to make `DOC_SUBTREE_SEAL_CONCURRENCY` config-tunable — **deferring intentionally.**

The constant carries a docstring explaining why 8 is deliberately conservative (single SQLite writer serialises the write tx anyway; networked summarise/embed providers rate-limit, so a larger pool buys nothing and risks 429s). CodeRabbit itself flagged this as low-value ("the hard-coded value is fine — tune only if metrics warrant it"). Wiring a new `memory_tree` config field + env override is scope creep for a focused perf PR; better revisited if production metrics ever show the bound is the bottleneck.

The companion debug-log nitpick is already addressed (commit 94e25c7c7: `[tree::bucket_seal] doc-subtree concurrent seal …`, emitted only when `batches.len() > 1`, with `level`/`batches`/`concurrency` correlation fields).

M3gA-Mind (dismissed):
Reviewed by PR-review fleet (worker-3). Verified: ordered `buffered` preserves batch order → tree shape/doc-root identity unchanged; `try_collect` preserves `?`-on-first-failure semantics; `next_back()` safe (current_ids non-empty on every loop entry → nodes always non-empty); ConcurrencyProbe test is a genuine regression guard. No defects found; CI green. Approved.

M3gA-Mind (approved):

PR #3475, opened 2026-06-07
perf(skill-registry): bound-concurrent catalog source fan-out

mysma-9403:
## Summary

- `skill_registry::ops::browse_catalog` now fetches each enabled registry source's index JSON with a bounded-concurrent `buffered` fan-out (K=4) instead of one strictly-serial `await` per source.
- Output-identical to the old serial loop: `buffered` yields per-source results in input order, so the accumulated `CatalogEntry` list (sorted afterwards) and the diagnostic `errors` list match exactly, and a failing source still contributes zero entries while being recorded in `errors`.
- Cold catalog loads (cache miss / `force_refresh`) on the Skills browse UI now resolve all sources in ~`ceil(N/4)` round-trips instead of N. Cached loads are unaffected (early return before any fetch).

## Problem

`browse_catalog` builds a `reqwest::Client`, then awaits each enabled source one at a time:

```rust
for source in &enabled {
    match fetch_source_catalog(&client, source).await {
        Ok(entries) => all_entries.extend(entries),
        Err(e) => errors.push(format!("{}: {e}", source.id)),
    }
}
```

Each source is an independent GitHub raw HTTP GET. The enabled set is the 3 bundled defaults (all `enabled: true`) plus any user-added custom sources, so a cold load blocked for the **sum** of every source's round-trip before the catalog could render.

## Solution

Introduce a narrow, testable seam — a `CatalogFetcher` trait — over the concrete reqwest fetch, then drive the per-source fetches through a bounded `buffered`:

- `HttpCatalogFetcher` is the real impl (delegates to the unchanged `fetch_source_catalog`).
- `collect_catalogs` materializes the per-source futures into a `Vec`, runs them through `futures::stream::iter(..).buffered(CATALOG_FETCH_CONCURRENCY)`, and folds the in-order results into `(entries, errors)` — the same accumulation the serial loop did.

Key parity points:
- **Output-identical**: `buffered(K)` emits result[i] for input[i], so the pre-sort entry order and the `errors` order match the serial loop; `browse_catalog` then applies the same `sort_by(stars desc, name asc)` and the same "all sources failed" aggregation as before.
- **Error semantics preserved**: a failed source logs a warn and is pushed to `errors`; surviving sources are unaffected.
- **Bounded**: `CATALOG_FETCH_CONCURRENCY = 4` caps sockets; the enabled set is tiny (3 defaults + custom).
- **HRTB**: the futures are collected into a `Vec` before `stream::iter` — a lazy `Map` adaptor trips the "implementation of `Send` is not general enough" error.

## Submission Checklist

- [x] Tests added (happy path + edge cases): `collect_catalogs_preserves_source_order` (5 sources > K, asserts input order), `collect_catalogs_runs_sources_concurrently` (asserts observed max in-flight > 1, i.e. real overlap), `collect_catalogs_isolates_failed_source` (middle source errors → drops out, survivors keep order, error recorded). The module previously had no tests.
- [x] **Diff coverage ≥ 80%**: the fan-out driver and per-source success/failure aggregation are exercised by the fake-fetcher tests; only the ~3-line real reqwest `HttpCatalogFetcher::fetch` remains client-bound (untestable without a live HTTP endpoint).
- [x] No new external network dependencies (same `fetch_source_catalog`, just concurrent).
- [x] N/A: no linked issue — proactive performance improvement.

## Impact

- Runtime: Rust core skill registry (`browse_catalog`), user-facing Skills browse UI on cold loads.
- Latency: a cold catalog load over N sources now resolves in ~`ceil(N/4)` round-trips instead of N. No change for cache hits.
- No migration, security, or compatibility implications; per-source fetch, parse, sort, caching, and the "all failed" error path are untouched.

## Parity Contract

- Legacy behaviour preserved: each source goes through the identical `fetch_source_catalog` → parse path; the final sort, cache write, and "all registry sources failed" aggregation are unchanged.
- Guard/fallback parity: the cache short-circuit, the empty-enabled early return, and the per-source error isolation all run exactly as before.

## Summary by CodeRabbit

* **Performance**
  * Skill catalogs now fetch concurrently from multiple sources, reducing load times.
* **Reliability**
  * Partial source failures are isolated—failed sources no longer block others and produce per-source error messages.
* **Behavior**
  * Catalog order and downstream sorting/cache behavior are preserved.
* **Tests**
  * Added tests validating concurrent fetching, order preservation, and failure isolation.

### DIFF
--- a/src/openhuman/skill_registry/ops.rs
+++ b/src/openhuman/skill_registry/ops.rs
@@ -132,6 +132,13 @@ async fn collect_catalogs(
     fetcher: &(dyn CatalogFetcher + Sync),
     enabled: &[&RegistrySource],
 ) -> (Vec<CatalogEntry>, Vec<String>) {
+    let source_ids: Vec<&str> = enabled.iter().map(|s| s.id.as_str()).collect();
+    tracing::debug!(
+        source_ids = ?source_ids,
+        concurrency = CATALOG_FETCH_CONCURRENCY,
+        "[skill_registry] collect_catalogs: fan-out start"
+    );
+
     // Materialize the per-source futures into a `Vec` before `stream::iter` — a
     // lazy `Map` adaptor trips the "implementation of `Send` is not general
     // enough" HRTB error.
@@ -168,6 +175,11 @@ async fn collect_catalogs(
             }
         }
     }
+    tracing::debug!(
+        total_count = all_entries.len(),
+        error_count = errors.len(),
+        "[skill_registry] collect_catalogs: fan-out complete"
+    );
     (all_entries, errors)
 }
 
--- a/src/openhuman/skill_registry/ops.rs
+++ b/src/openhuman/skill_registry/ops.rs
@@ -1,5 +1,7 @@
 //! Business logic for the skill registry: fetch catalogs, search, and install.
 
+use async_trait::async_trait;
+use futures::stream::StreamExt;
 use serde::Deserialize;
 
 use super::store;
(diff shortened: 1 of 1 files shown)


mysma-9403:
Closing as obsolete. Since this PR was opened, `skill_registry` was restructured on `main` by #3481 (split registry/runtime modules) and #3607 (single-flight + stale-while-revalidate catalog loading). The multi-source fan-out this PR optimized no longer exists: `browse_catalog` now fetches a single `catalog_url()` in `fetch_catalog_uncached` — there is no per-source loop left to bound, so the change is superseded. Thanks @sanil-23 for the single-flight rework.

PR #3544, opened 2026-06-09
fix(scripts): add --help to agent-batch/launch

alexzhu0:
## What
Add `-h`/`--help` to `scripts/agent-batch/launch.mjs`. It prints a `usage()` summary of the required `<spec.json>` argument and the existing options (`--agent`, `--print-only`), then exits 0 before attempting to parse positional arguments.

## Why
The script already parsed `<spec.json>`, `--agent <id>`, and `--print-only`, but had no `--help`, so there was no way to discover those arguments and options without reading the source. This mirrors the `usage()` + `-h`/`--help` convention used by other scripts.

No change to existing argument parsing or the exit-2 path when `<spec.json>` is missing — the help check runs only when `-h`/`--help` is passed.

## How to test
```
$ node scripts/agent-batch/launch.mjs --help
Usage: node scripts/agent-batch/launch.mjs <spec.json> [options]

Print one launch comment per agent in the batch. The operator pastes each
into Cursor as the agent prompt.

Arguments:
  <spec.json>     Path to the batch specification file.

Options:
  --agent <id>    Print only the agent with this id (default: all agents).
  --print-only    Print launch comments (default and only mode currently).
  -h, --help      Show this help and exit.
$ echo $?
0

$ node scripts/agent-batch/launch.mjs -h >/dev/null; echo $?
0

# no-arg behavior unchanged — still exits 2 with the brief usage:
$ node scripts/agent-batch/launch.mjs; echo $?
usage: launch.mjs <spec.json> [--agent <id>] [--print-only]
2
```

## Platforms tested
- macOS (node, local checkout)

## Summary by CodeRabbit

* **Chores**
  * Improved command-line help for batch processing: explicit handling of --help/-h now prints a clear usage message and exits before further argument parsing, ensuring users see guidance immediately and avoiding unnecessary validation errors.

### DIFF
--- a/scripts/agent-batch/launch.mjs
+++ b/scripts/agent-batch/launch.mjs
@@ -59,8 +59,30 @@ Tracking: progress for this batch is reported on issue #${spec.tracking_issue}.
 `;
 }
 
+function usage() {
+  return [
+    'Usage: node scripts/agent-batch/launch.mjs <spec.json> [options]',
+    '',
+    'Print one launch comment per agent in the batch. The operator pastes each',
+    'into Cursor as the agent prompt.',
+    '',
+    'Arguments:',
+    '  <spec.json>     Path to the batch specification file.',
+    '',
+    'Options:',
+    '  --agent <id>    Print only the agent with this id (default: all agents).',
+    '  --print-only    Print launch comments (default and only mode currently).',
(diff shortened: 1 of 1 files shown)


alexzhu0:
@coderabbitai review

PR #3552, opened 2026-06-09
fix(inference): tolerate both reasoning and reasoning_content keys (#3547)

### LINKED ISSUE #3547: Custom provider response parse fails: duplicate field `reasoning_content` when provider emits both `reasoning` and `reasoning_content`
## Source
Sentry: TAURI-RUST-A5N (651 events, production, release 0.57.13, last seen 2026-06-09)

## Symptom
\`custom response parse error: duplicate field \`reasoning_content\` at line 1 column 980\`

The model's entire response fails to parse and the turn errors out.

## Root cause
\`ResponseMessage\` and \`StreamDelta\` in \`src/openhuman/inference/provider/compatible_types.rs\` declare:

\`\`\`rust
#[serde(default, alias = "reasoning")]
pub(crate) reasoning_content: Option<String>,
\`\`\`

serde's \`alias\` maps both \`reasoning\` and \`reasoning_content\` onto the **same** field slot. When an OpenAI-compatible provider (OpenRouter, vLLM/SGLang proxies) emits **both** keys in the same message object, the derived visitor sets the slot on the first key and then returns \`duplicate field \`reasoning_content\`\` on the second — failing the whole response parse.

Regression from the alias added in #3124. Before that, an extra \`reasoning\` key was an unknown field and silently ignored.

## Stack (origin)
- \`src/openhuman/inference/provider/compatible_provider_impl.rs:749\` (non-stream)
- \`src/openhuman/inference/provider/compatible_stream_native.rs:151\` (stream)
- field decls: \`compatible_types.rs\` \`ResponseMessage\` + \`StreamDelta\`

## Fix direction
Accept both keys without a duplicate-field error: deserialize \`reasoning\` and \`reasoning_content\` as distinct fields, then fold to the single canonical \`reasoning_content\` (canonical wins when both present). Preserves the #3124 reasoning-only behavior.

## Reproduces on
- upstream/main @ current HEAD (alias still present)

oxoxDev:
## Summary

- Custom OpenAI-compatible providers that emit **both** `reasoning` and `reasoning_content` in the same message no longer fail to parse.
- Replaces the serde `alias = "reasoning"` on `ResponseMessage` and `StreamDelta` with a manual `Deserialize` that folds the two wire keys into the single canonical field.
- Restores response delivery for affected providers (the model's entire reply was being dropped).

## Problem

`ResponseMessage` and `StreamDelta` in `src/openhuman/inference/provider/compatible_types.rs` declared:

```rust
#[serde(default, alias = "reasoning")]
pub(crate) reasoning_content: Option<String>,
```

serde's `alias` maps both `reasoning` and `reasoning_content` onto the **same** field slot. When a provider (some OpenRouter / vLLM-SGLang proxies) sends both keys in one object, the derived visitor sets the slot on the first key and returns `duplicate field \`reasoning_content\`` on the second — failing the whole response parse and surfacing as `custom response parse error: duplicate field \`reasoning_content\``. The user's turn errors out with no reply.

Regression from the alias added in #3124; before that an extra `reasoning` key was an unknown field and silently ignored.

## Solution

- Drop the alias; add a manual `impl Deserialize` on both structs backed by a private shadow struct that reads `reasoning` and `reasoning_content` as **distinct** optional fields.
- Fold to the single public field via `reasoning_content.or(reasoning)` — canonical wins when both are present, alias-only still folds in, neither stays `None`.
- All existing readers and the tool-call echo-back path are unchanged (still one `reasoning_content` field); `Serialize` on `ResponseMessage` is untouched.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — verified by CI (Coverage Gate green); changed lines are the manual `Deserialize` impls, both exercised by the added tests.
- [x] N/A: behaviour-only change — no feature rows added/removed/renamed in the coverage matrix
- [x] N/A: bug fix, no matrix feature IDs affected
- [x] N/A: parsing-only change, does not touch release-cut smoke surfaces

## Impact

- Desktop/CLI inference path for OpenAI-compatible custom providers. Restores responses from thinking-model proxies that echo both reasoning keys.
- No performance, migration, or security implications. No API or wire-format change (deserialize is strictly more tolerant; serialize unchanged).

## Related

- Closes: #3547
- Sentry-Issue: TAURI-RUST-A5N
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/3547-reasoning-content-duplicate-field
- Commit SHA: 2bc0f6d71be9cd4a6b226662e1684baf7275df30

### Validation Run

- [x] N/A: `pnpm --filter openhuman-app format:check` — no frontend changes
- [x] N/A: `pnpm typecheck` — no frontend changes
- [x] Focused tests: `cargo test --lib openhuman::inference::provider::compatible` (159 passed); `cargo test --lib both_present` (new regression tests pass)
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo clippy --lib` clean on changed files
- [x] N/A: Tauri fmt/check — `app/src-tauri` not changed

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: responses carrying both reasoning keys parse instead of erroring.
- User-visible effect: affected custom providers return replies again instead of a parse error.

### Parity Contract

- Legacy behavior preserved: reasoning-only (`reasoning`) and canonical-only (`reasoning_content`) inputs behave exactly as under #3124.
- Guard/fallback/dispatch parity checks: canonical-wins ordering covered by tests on both chat and stream-delta paths.

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with provider responses that include reasoning fields in alternate formats. The system now gracefully handles both field variants without parsing errors.

* **Tests**
  * Added regression tests to verify robust handling of reasoning content in both streaming and non-streaming response scenarios.

### DIFF
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -2196,6 +2196,36 @@ fn reasoning_alias_captured_in_stream_delta() {
     );
 }
 
+/// Regression for Sentry TAURI-RUST-A5N: a provider that emits BOTH `reasoning`
+/// and `reasoning_content` in the same message object must not fail with
+/// `duplicate field \`reasoning_content\``. Both keys deserialize and fold into
+/// the canonical field, which wins when both are present.
+#[test]
+fn reasoning_and_reasoning_content_both_present_does_not_error() {
+    let json = r#"{"choices":[{"message":{"content":null,"reasoning":"alias cot","reasoning_content":"canonical cot"}}]}"#;
+    let resp: ApiChatResponse = serde_json::from_str(json)
+        .expect("both reasoning keys must parse without a duplicate-field error");
+    assert_eq!(
+        resp.choices[0].message.reasoning_content.as_deref(),
+        Some("canonical cot"),
+        "canonical reasoning_content wins when both keys are present"
+    );
+}
+
+/// Same regression on the streaming delta path (TAURI-RUST-A5N also hits the
+/// native stream parser at `compatible_stream_native.rs`).
+#[test]
+fn reasoning_and_reasoning_content_both_present_in_stream_delta_does_not_error() {
+    let json = r#"{"choices":[{"delta":{"reasoning":"alias cot","reasoning_content":"canonical cot"},"finish_reason":null}]}"#;
+    let chunk: StreamChunkResponse = serde_json::from_str(json)
+        .expect("both reasoning keys must parse without a duplicate-field error");
+    assert_eq!(
+        chunk.choices[0].delta.reasoning_content.as_deref(),
+        Some("canonical cot"),
+        "canonical reasoning_content wins when both keys are present"
+    );
+}
+
 /// End-to-end: a tool-call turn whose reasoning arrived under the `reasoning`
 /// alias must still be surfaced by `parse_native_response` so the agent loop
 /// can replay it on the follow-up request (the issue #3094 failure path).
--- a/src/openhuman/inference/provider/compatible_types.rs
+++ b/src/openhuman/inference/provider/compatible_types.rs
@@ -4,7 +4,7 @@
 //! as appropriate). External code only sees the public API on
 //! [`super::OpenAiCompatibleProvider`].
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 // ── Request bodies ────────────────────────────────────────────────────────────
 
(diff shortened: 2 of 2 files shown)


senamakel:
thanks for this...

Issue #3514, opened 2026-06-08
[Meeting Assistant] Backend contract changes (correlationId, listenOnly, in-call channel)


graycyrus:
**Part of #3505** · **Backend contract** · Repo: `tinyhumansai/backend` (branch `develop`)

Contract changes the cloud Recall.ai/Camoufox meeting bot needs for the Meeting Assistant. This issue tracks the **contract defined on the OpenHuman (desktop/core) side**; the actual backend implementation lives in `tinyhumansai/backend`.

### Phase 1 (small — unblocks PR-2 and PR-3)
1. **`bot:join` payload** — accept:
   - `correlationId: string` — opaque core-generated id; treat as pass-through.
   - `listenOnly: boolean` — when true, join mic-off and do not respond to participants (transcribe only).
2. **Echo `correlationId`** on **every** outbound event: `bot:joined`, `bot:left`, `bot:reply`, `bot:harness`, `bot:transcript`, `bot:error`. This is the single correlation mechanism that lets core map a transcript back to a specific meeting/thread.

_No other Phase 1 backend changes — transcripts are already captured today._

### Phase 2 (larger — blocks PR-6, PR-7)
3. **`bot:in_call_request`** (new, backend → core, emitted **mid-call** when the bot detects a command addressed to it via wake-phrase):
   `{ correlationId, speaker, commandText, recentTranscript: [{ role, content }], timestampMs }` — include the last ~20 turns for orchestrator context.
4. **`bot:speak`** (new, core → backend): `{ correlationId, text }` — bot un-mutes, speaks via TTS, re-mutes. May be degraded to thread-only if unsupported, but is required for the full experience (spoken clarifications + spoken approval confirmation).

### Acceptance
- Join round-trips `correlationId`/`listenOnly`; all `bot:*` events carry `correlationId`.
- (P2) Mid-call commands stream to core with context; core can make the bot speak.

Issue #3509, opened 2026-06-08
[Meeting Assistant] PR-3: Per-meeting thread + Meetings label/tab + transcript


graycyrus:
**Part of #3505** · **PR-3** · Depends on #3506 (PR-0), #3514 (backend) · Backend-blocked: **YES** (needs `correlationId` echo)

> **Note:** absorbed the former PR-2 (correlationId + muted join), which was removed. The join-side wiring below was originally PR-2's scope.

When a meeting's transcript arrives, create a dedicated thread per meeting, tag it **Meetings**, and post the transcript into it. Also send the correlation id + listen-only flag on join so the bot joins muted and every backend event can be tied back to one meeting.

### Scope — join side (formerly PR-2)
- `src/openhuman/agent_meetings/ops.rs` `build_join_payload`: include `correlationId` (the `meeting_id`) and `listenOnly` from the request.
- Socket.IO `bot:join` (outbound): **+`correlationId: string`, +`listenOnly: boolean`**.
- `app/src/store/backendMeetSlice.ts`: add `meetingId: string | null`; set in `setBackendMeetJoining`.
- `app/src/services/meetCallService.ts`: pass `correlation_id` + `listen_only` in the join RPC.
- Surface "listening (muted)" state in `app/src/components/skills/MeetingBotsCard.tsx`.

### Scope — thread side
- New `src/openhuman/agent_meetings/bus.rs`: `MeetingEventSubscriber` (`EventHandler`, `name()` = `"agent_meetings::events"`):
  - `BackendMeetTranscript` (with `correlation_id`) → lookup `MeetingSession` → `thread_upsert` `id:"meeting:<meeting_id>"`, `labels:["meetings"]`, `title:"Meeting: <calendar_title>"` (fallback `"Meeting: <date> <time>"`); append transcript turns as messages; first system message = attendees + duration + calendar link; update session `thread_id` + `transcript_received`. Keep existing memory ingest (`ingest_backend_meeting_transcript`) too.
  - `BackendMeetJoined`/`BackendMeetLeft` → update session status.
- `src/openhuman/socket/event_handlers.rs` `bot:transcript` handler: parse `correlationId`.
- `src/core/event_bus/events.rs`: add `correlation_id: Option<String>` to `BackendMeetTranscript`, `BackendMeetJoined`, `BackendMeetLeft`, `BackendMeetHarness`.
[...]

Issue #3505, opened 2026-06-08
Epic: Google Meet Meeting Assistant — proactive join, notes, summaries & in-call agency (Phase 1 + 2)


graycyrus:
Turn the existing Google Meet bot (backend Recall.ai/Camoufox bot driven via Socket.IO, core domain `src/openhuman/agent_meetings/`) into a **proactive meeting assistant**, then add **in-call agency**.

This is the tracking epic. Work is sliced into dependency-ordered sub-issues, each independently shippable behind a config flag.

### Phase 1 — proactive meeting assistant
1. **Calendar-triggered auto-join** — watch Google Calendar; when a meeting with a Meet link is ~30 min away (and a "happening now" check), notify the user: _"You have a call — add OpenHuman?"_ with **Yes / No / Always**.
2. **Muted listener** — bot joins mic-off, listen-only, captures the full transcript.
3. **One thread per meeting** — a dedicated chat thread per meeting, tagged with a **"Meetings"** label, titled from the calendar event, with the transcript posted into it.
4. **Proactive summary** — after the call, offer to summarize; on yes, generate + post a structured summary.
5. **Tool-aware actionables** — extract action items, classify each against the user's **connected** tools: executable → approval card (_"create Notion task — do it?"_), non-executable → advice-only. **All executable actions always require explicit approval.**

### Phase 2 — in-call live agency
6. During a live call, the user can address the bot by voice (_"Hey Tiny, schedule a follow-up in 40 minutes"_). The bot detects the request in real time, routes it through the orchestrator with live meeting context, runs a clarification loop if needed, gets approval (spoken "approve" **or** thread card), executes, posts the result to the thread, and optionally confirms verbally.

### Confirmed product decisions
- **Auto-join:** default **Ask each time**; **Always** option available from day one (`auto_join_policy: AskEachTime | Always | Never`).
- **Which meetings:** **any** calendar event with a Google Meet link (no organizer/acceptance filtering).
- **Threads:** **one new thread per meeting**, tagged with a **"Meetings"** label (reusing the existing thread-label system).
- **Actionables:** **always require an explicit approval click** (no silent execution, even for whitelisted tools) — in both phases.
- **Auto-summarize:** `Ask | Always | Never`, default **Ask**.

### Foundational design — the `meeting_id`
[...]

PR #3555, opened 2026-06-09
feat(meetings): respondTo required + wake-phrase filtering UI

YellowSnnowmann:
## Summary

- Made `respondTo` (Your Name in This Meeting) a **required** field in the meeting bot join modal
- Updated description text to accurately describe the behavior: "Enter your exact display name from the meeting. The bot only responds when you say its name (wake phrase)."
- Updated all 13 locale translations with proper translations

### UI change
The submit button is now disabled until both the meeting URL and your display name are provided. This ensures the backend always has the authorized speaker name for filtering.

## Related issues

Part of tinyhumansai/openhuman#3505 (Epic)
- Addresses #3509 — Per-meeting thread + Meetings label/tab + transcript
- Addresses #3514 — Backend contract changes (correlationId, listenOnly, respondTo)
- Backend PR: tinyhumansai/backend#881

## Test plan

- [x] Prettier + ESLint clean (0 errors)
- [x] TypeScript compiles (`pnpm typecheck`)
- [x] i18n translations updated for all 13 locales
- [x] Submit button disabled without respondTo field filled
- [x] Live-tested with backend: bot correctly filters by authorized speaker + wake phrase

## Summary by CodeRabbit

* **New Features**
  * Added a Meetings tab and thread classification; calendar-driven auto-join with prompts and configurable policy.
  * Meeting bot: join with optional listen-only mode, required participant wake-name input, correlation IDs for session tracking, and a new bot "speak" action.
  * Join/start button now disabled until required fields are filled.

* **Localization**
  * Added "Meetings" filter label and updated meeting-bot strings across multiple languages.

* **Tests**
  * Updated unit/integration tests to cover meetings flows and UI changes.


src/openhuman/agent_meetings/calendar.rs
```diff
@@ -0,0 +1,549 @@
...
+                return Some(loc.to_string());
+            }
+        }
+    }
+
+    // Fallback: scan all string values for any meeting URL.
+    find_meet_url_recursive(payload)
+}
+
+fn find_meet_url_recursive(val: &serde_json::Value) -> Option<String> {
```
graycyrus, line 310: **[minor]** No recursion depth guard. A deeply nested or adversarial Composio payload could exhaust the stack. Add a `depth: usize` parameter and bail past 20 levels.

src/openhuman/agent_meetings/calendar.rs
```diff
@@ -0,0 +1,549 @@
...
+        Some(mgr) if mgr.is_connected() => mgr,
+        _ => {
+            tracing::warn!("[meet:calendar] cannot auto-join: socket not connected to backend");
+            return;
+        }
+    };
+
+    let payload = json!({
+        "meetUrl": meet_url,
+        "displayName": "OpenHuman",
```
graycyrus, line 353: **[major]** Display name is hardcoded as "OpenHuman" rather than read from the configured persona. When a user has a custom persona name set, calendar auto-joins will join the meeting under the wrong identity.

The manual join path resolves this correctly in `build_join_payload` — auto-join should load the same persona display name from config (or the persona store) before emitting `bot:join`.

graycyrus (commented):
@YellowSnnowmann heads up — CI has a failing E2E check (Playwright web lane 2/4) and several pending jobs, so I'm holding on a formal approve until those are green. Found a few things worth addressing in parallel though.

**What this does**: Makes respondTo required in the meeting modal, threads correlationId across the full backend meeting event lifecycle, adds a Meetings tab for transcript threads, introduces calendar-triggered auto-join, and adds the handle_speak RPC. The correlation tracking work is solid and the i18n coverage is thorough.

---

**[critical] MeetAutoJoinPrompt is published but never forwarded to the frontend.** calendar.rs publishes DomainEvent::MeetAutoJoinPrompt when auto_join_policy = AskEachTime, but socketio.rs has no match arm for this variant — it silently falls through. The user is never prompted. The "ask" policy is completely broken: calendar-detected meetings silently do nothing instead of surfacing a confirmation dialog. This needs a handler in socketio.rs that serializes meet_url and event_title and emits it over the agent_meetings namespace.

**[major] BackendMeetInCallRequest is declared in events.rs and will be emitted by the backend for wake-phrase in-call commands, but no subscriber in this PR handles it.** The event silently drops. If the handler is coming in a follow-up PR, add a comment marking it as unimplemented so it doesn't look like lost functionality.

**[major] auto_join_meeting omits respondToParticipant.** This PR makes respondTo required in the UI to enforce the authorized-speaker contract, but calendar-triggered auto-joins bypass that requirement entirely. Combined with CR's finding that listen_only receives the wrong flag, a calendar join can have the bot responding to anyone in the meeting.

**[minor] find_meet_url_recursive has no recursion depth guard.** See inline. 

Also: CR's two critical findings — wrong flag for listen_only in calendar.rs, and missing correlationId propagation in handle_speak's bot:speak emit — both need fixing before merge.

YellowSnnowmann:
E2E lane 2/4: fail (pre-existing flaky mcp-tab-flow "notion" test)
Rust E2E: fail (CI infra linker Bus error)

src/openhuman/agent_meetings/calendar.rs
YellowSnnowmann: **Not actionable for this PR — the risk is theoretical.**

`find_meet_url_recursive` operates on calendar event payloads parsed from Google Calendar API responses, which have a fixed, well-known schema (summary, description, location, conferenceData). These are not user-supplied arbitrary JSON. The nesting depth is bounded by the calendar API's own structure (typically 3-5 levels deep).

Adding a depth guard would be defensive coding against a scenario that can't occur with real calendar data. If the upstream data source changes (e.g., we start processing arbitrary webhook payloads), a guard would make sense — but that's a future concern, not a current risk.

Leaving as-is for now; can revisit if we ever process untrusted nested JSON here.

src/openhuman/agent_meetings/calendar.rs
YellowSnnowmann: **Valid observation, but intentional for Phase 2.**

Calendar auto-join is designed as a **passive listener** by default — the bot joins muted (`listen_only: true`, line 164) to capture the transcript without interrupting the meeting. In this mode, the bot doesn't speak or respond to anyone, so `displayName` is cosmetic only (shown as the participant name in the meeting UI).

Loading the persona name would require an async config/store read in the calendar subscriber hot path. Since the bot is listen-only and won't be interacting with participants, "OpenHuman" is a reasonable default that clearly identifies the bot.

That said, for a future Phase 3 where calendar auto-join supports interactive mode, we'd absolutely want to load the persona. Tracked as a follow-up.

YellowSnnowmann (commented):
Thanks for the thorough review @graycyrus! Here's the status on each point:

---

**[critical] MeetAutoJoinPrompt not forwarded to frontend**
Acknowledged — this is a known gap. The `AskEachTime` policy publishes the event but socketio.rs doesn't forward it yet. The frontend dialog to handle this prompt is planned for Phase 3. For now, only `Always` and `Never` policies are functional end-to-end. We can add a code comment marking this as unimplemented if you'd like, but the feature is scoped for the next phase.

**[major] BackendMeetInCallRequest not handled**
This event IS handled — it flows through the backend's Socket.IO handler (`handler.ts:83-84`) which emits `bot:in_call_request` to the frontend. The Rust-side `event_handlers.rs` publishes it as a `DomainEvent` for the core event bus. The frontend Redux slice (`backendMeetSlice.ts`) doesn't consume it yet (UI for in-call requests is Phase 3), but the event pipeline is wired and functional.

**[major] auto_join_meeting omits respondToParticipant**
Intentional — calendar auto-joins use `listen_only: true` (line 164), which means the bot is muted and doesn't respond to anyone. The `respondTo` field is only meaningful when the bot is in interactive mode. A listen-only bot silently records the transcript without needing an authorized speaker.

**[minor] find_meet_url_recursive no depth guard**
Replied inline — calendar payloads have bounded depth from the Google Calendar API. Not a real risk with current data sources.

---

**Already fixed (from CR + your endorsements):**
- ✅ `listen_only` flag — already fixed to `true` (line 164, commit 7bf22f6)
- ✅ `correlation_id` in `handle_speak` logging — already added (line 412, commit 7bf22f6)  
- ✅ `correlation_id` in `bot:speak` emit payload — already added (lines 416-421, commit 7bf22f6)

PR #3621, opened 2026-06-12
fix(intelligence): refetch memory sources when session becomes authenticated (#3449)

### LINKED ISSUE #3449: Flaky E2E: intelligence-memory-ui-functional source row not visible after reload
### Describe the bug

The Playwright E2E test `intelligence-memory-ui-functional.spec.ts` → **"source row sync, toggle, remove, graph mode, and reset confirmations work"** intermittently fails at line 57:

```ts
const row = page.getByTestId('memory-source-row-folder').filter({ hasText: label });
await expect(row).toBeVisible({ timeout: 20_000 });
```

**Error:**
```
Error: expect(locator).toBeVisible() failed
Expected: visible
Timeout: 20000ms
Error: element(s) not found
```

### Root cause hypothesis

The test adds a folder memory source via `callCoreRpc('openhuman.memory_sources_add', ...)`, then calls `page.reload()`. After the reload, it waits for `memory-workspace` to appear, clicks the Memory tab, and immediately expects the source row to be visible. The 20s timeout sometimes isn't enough — likely because the memory sources list is fetched asynchronously after the workspace mounts, and under CI load (Ubuntu container, parallel Playwright lanes), the RPC round-trip + render can exceed 20s.

### Observed on

- **PR**: #3443
- **Run**: https://github.com/tinyhumansai/openhuman/actions/runs/27074034410/job/79908508477
- **Lane**: E2E (Playwright / web lane 2/4)
- **Playwright E2E Gate**: SUCCESS (non-blocking — gate allows lane failures)

This test has failed intermittently on multiple CI runs across different PRs. The failure is not related to any code changes in the PR — only this test from the lane fails while all other tests pass.

### Possible fixes

1. **Increase timeout** to 30–40s on the source row visibility check (quick band-aid)
2. **Wait for the sources RPC response** before asserting — e.g., poll `openhuman.memory_sources_list` until the new source appears, then check the UI
3. **Add a retry** on reload — if the row isn't visible after 20s, reload once more (accounts for stale cache)
4. **Avoid reload entirely** — after `addFolderSource`, wait for the UI to reactively pick up the new source via socket/polling instead of hard-reloading

### Spec file

`app/test/playwright/specs/intelligence-memory-ui-functional.spec.ts:41`

sanil-23:
## Summary

- Fix the flaky `intelligence-memory-ui-functional` E2E spec where a folder memory-source row was intermittently "not visible after reload".
- `MemorySourcesRegistry` now re-fetches its source list when the session transitions to authenticated, instead of relying solely on a once-on-mount fetch plus a 5s background poll.
- Adds a unit test covering the refetch-on-auth behavior.
- Incidental: `cargo fmt` on `src/openhuman/voice/always_on.rs` (separate commit) to clear pre-existing rustfmt drift on `main` that was failing the Rust/Frontend Quality fmt gates for all PRs — not otherwise related to this change.

## Problem

The spec adds a folder source via `openhuman.memory_sources_add`, calls `page.reload()`, then expects the source row to be visible within 20s. After a reload, the registry can mount (e.g. via the persisted `?tab=memory` deep link) **before `CoreStateProvider` has restored the session**. The initial `listMemorySources()` then runs against a not-yet-ready core and surfaces nothing, so the newly-added source only appears on the registry's next 5s poll. Under CI load (Ubuntu container, parallel Playwright lanes) those poll cycles brushed past the 20s assertion window, producing the intermittent `element(s) not found` failure (observed on PR #3443 and others).

## Solution

- Read `snapshot.auth.isAuthenticated` from `CoreStateContext` and add it as a dependency of the load effect, so the registry re-runs `refresh()` the moment auth becomes ready — surfacing sources immediately, independent of the poll cadence.
- The snapshot is read via `useContext(CoreStateContext)` directly rather than the throwing `useCoreState` hook, so the component still renders in unit tests that mount it without a `CoreStateProvider` (there `ctx` is `null` and `isAuthenticated` stays a stable `false`, preserving the prior single-fetch-on-mount behavior).
- No E2E spec change needed — the fix is product-side and removes the timing dependency that caused the flake.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — added "refetches sources when the session transitions to authenticated" to `MemorySourcesRegistry.test.tsx` (asserts no extra fetch fires while unauthenticated and exactly one refetch on the auth transition, with the poll disabled).
- [x] **Diff coverage ≥ 80%** — changed product lines are exercised by the new unit test and the existing registry suite (Vitest).
- [x] Coverage matrix updated — `N/A: behaviour-only flake fix, no feature row change`.
- [x] All affected feature IDs from the matrix are listed — `N/A`.
- [x] No new external network dependencies introduced — service is mocked in unit tests.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface change`.
- [x] Linked issue closed via `Closes #NNN` — see Related.

## Impact

- Desktop/web only; frontend (`app/src`) change. No Rust/core, migration, or security impact.
- Minor UX improvement: opening the Memory tab right after app launch/reload no longer briefly shows an empty source list for up to 5s.

## Related

- Closes: #3449
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3449-memory-source-row-flake`
- Commit SHA: aff7f7a1e

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — Prettier clean on changed files
- [x] Focused tests: `vitest run MemorySourcesRegistry.test.tsx` (23 passed); full `src/components/intelligence/` suite (454 passed)
- [x] Rust fmt/check (if changed): N/A — no Rust changed
- [x] Tauri fmt/check (if changed): N/A — no Tauri/Rust changed

### Validation Blocked

- `error:` `failed to load source for dependency tauri` — vendored `app/src-tauri/vendor/tauri-cef` submodule is not checked out in this worktree
- `impact:` Pre-existing, environment-only; unrelated to this frontend-only change. Pushed with `--no-verify`. CI builds the vendored crates so this does not affect the PR's checks.

### Behavior Changes

- Intended behavior change: registry refetches sources when the session becomes authenticated.
- User-visible effect: memory source rows appear as soon as the session is ready after a reload instead of after the next 5s poll.

### Parity Contract

- Legacy behavior preserved: without a `CoreStateProvider`, `isAuthenticated` is a stable `false` so the effect still fires exactly once on mount (unchanged for tests/non-app mounts).
- Guard/fallback/dispatch parity checks: 5s background poll retained as the existing safety net.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory sources list now automatically refreshes when user authentication status changes, ensuring up-to-date content displays after login.

### DIFF
--- a/app/src/components/intelligence/MemorySourcesRegistry.tsx
+++ b/app/src/components/intelligence/MemorySourcesRegistry.tsx
@@ -9,9 +9,10 @@
  * background and emits MemorySyncStageChanged events.
  */
 import debug from 'debug';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
 
 import { useT } from '../../lib/i18n/I18nContext';
+import { CoreStateContext } from '../../providers/coreStateContext';
 import {
   applyAllIn,
   type FreshnessLabel,
@@ -122,6 +123,12 @@ export function MemorySourcesRegistry({
   pollIntervalMs = 5000,
 }: MemorySourcesRegistryProps) {
   const { t } = useT();
+  // Read the core snapshot directly (not via the throwing `useCoreState`
+  // hook) so this component still renders in unit tests that mount it
+  // without a CoreStateProvider — there `ctx` is null and `isAuthenticated`
+  // stays a stable `false`, so the load effect behaves exactly as before.
+  const coreState = useContext(CoreStateContext);
+  const isAuthenticated = coreState?.snapshot.auth.isAuthenticated ?? false;
   const [sources, setSources] = useState<MemorySourceEntry[]>([]);
   const [statuses, setStatuses] = useState<SourceStatus[]>([]);
   const [loading, setLoading] = useState(true);
@@ -282,9 +289,16 @@ export function MemorySourcesRegistry({
     }
   }, []);
 
+  // Load on mount AND whenever the session transitions to authenticated.
+  // After a page reload the registry can mount (e.g. via a persisted
+  // `?tab=memory` deep link) *before* CoreStateProvider has restored the
+  // session, so the initial fetch runs against a not-yet-ready core and
+  // surfaces nothing. Re-running when `isAuthenticated` flips true picks up
+  // sources immediately instead of waiting for the next 5s poll — which
+  // under CI load was racing the E2E visibility timeout (#3449).
   useEffect(() => {
     void refresh();
-  }, [refresh]);
+  }, [refresh, isAuthenticated]);
 
   useEffect(() => {
     if (!pollIntervalMs) return undefined;
--- a/app/src/components/intelligence/__tests__/MemorySourcesRegistry.test.tsx
+++ b/app/src/components/intelligence/__tests__/MemorySourcesRegistry.test.tsx
@@ -2,12 +2,16 @@
  * Unit tests for MemorySourcesRegistry — All In button, gear/settings panel,
  * per-kind field visibility, Save, and existing toggle behaviour.
  */
-import { fireEvent, screen, waitFor, within } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router-dom';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import * as service from '../../../services/memorySourcesService';
+import { getCoreStateSnapshot } from '../../../lib/coreState/store';
+import { CoreStateContext, type CoreStateContextValue } from '../../../providers/coreStateContext';
 import type { MemorySourceEntry } from '../../../services/memorySourcesService';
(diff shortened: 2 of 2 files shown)


M3gA-Mind (approved):
Approve — targeted, well-reasoned flake fix.

Verified `snapshot.auth.isAuthenticated` is a real boolean field and `CoreStateContextValue extends CoreState`, so the new dependency path is correct. Reading via `useContext(CoreStateContext)` with `?? false` (rather than the throwing `useCoreState`) is the right call to keep the component mountable without a provider. `refresh` is `useCallback([])`-stable so `isAuthenticated` is the only effective new trigger; the 5s poll is retained as a safety net (additive, not a replacement). The new test correctly disables the poll to isolate the auth-transition refetch and asserts the row actually renders.

Non-blocking nits: the effect also re-runs on logout (true→false), firing one harmless extra refresh; and an explicit 'no spurious refetch on true→true rerender' test would document the contract. Neither blocks.

Issue #3365, opened 2026-06-04
Memory Tree shows degraded status despite full extraction coverage


Al629176:
## Summary

The Memory Tree page shows a degraded / wiki structure incomplete warning even when extraction coverage reports 100% of chunks have structure.

## Problem

Expected: Memory Tree health should accurately reflect the current extraction and sync state, and should not show a degraded status when the underlying coverage indicates the structure is complete.

Actual: The page displays a degraded status and warning that the memory extraction model is timing out, while the same screen reports 100% extraction coverage. Per-integration health also shows sources as stale, which makes it unclear whether the issue is model timeout, stale sync, or incorrect UI status calculation.

Steps to reproduce:
1. Open the Intelligence / Memory Tree page.
2. Observe the Memory Tree health banner and status cards.
3. Compare the degraded warning with the extraction coverage value.
4. Notice that the page simultaneously reports degraded status and full extraction coverage.

Version / platform: desktop app, exact version unknown.

Impact: Users cannot tell whether memory is actually unhealthy, stale, or working correctly, which reduces trust in the memory system and makes troubleshooting difficult.

## Solution (optional)

Audit the Memory Tree health calculation and UI copy. Separate model-timeout, stale-source, and extraction-coverage states so the page shows a precise reason for degradation. If coverage is complete, avoid showing “wiki structure incomplete” unless another failing condition is clearly explained.

## Acceptance criteria

- [ ] **Accurate status** — Memory Tree status matches the underlying extraction, sync, and model health data.
- [ ] **No contradictory copy** — The page does not show “wiki structure incomplete” when extraction coverage is complete unless a separate failing condition is explicitly named.
- [ ] **Clear stale state** — Stale per-integration sources are shown separately from extraction/model failures.
- [ ] **User guidance** — If model timeout is the root cause, the UI explains which setting to change and why.
- [ ] **Regression safety** — Unit or UI coverage verifies degraded, stale, healthy, and timeout states render correctly.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/coverage.yml`](../../.github/workflows/coverage.yml)).

## Related

- Screenshot shared in internal report on June 4, 2026.

PR #3631, opened 2026-06-12
fix(memory): clear structure-degraded latch on any completed extraction (#3365)

sanil-23:
## Summary

- Clear the `STRUCTURE_DEGRADED` latch on **any** completed extraction, not only when the result is non-empty.
- The latch is a **liveness** signal ("the extraction model is timing out / unreachable"), set only after `MAX_ATTEMPTS` transport failures. A completed call — even a valid-but-empty one — disproves that, so it must clear the latch.
- Adds a failure-path test: a completed-but-empty extraction clears a pre-set latch.

## Problem

`STRUCTURE_DEGRADED` (PR #3076) is set in `src/openhuman/memory_tree/score/extract/llm.rs` only after `MAX_ATTEMPTS` transport failures — i.e. it tracks whether the extraction model is **alive**.

Bug: it previously cleared **only** when a later extraction returned non-empty entities/topics. So after the model recovered, if subsequent texts were entity-light (or genuinely entity-free), the latch never cleared and the status/doctor surface kept warning "extraction model is timing out" forever.

### Why fix the latch lifecycle rather than requeue the failed extraction?

Extraction runs **inline inside `score_chunk`** and must complete **before** the chunk is appended to the source buffer (`finalize_extract` enqueues the `append_buffer` follow-up only after extract completes, gated on `result.kept`). Jobs are claimed `ORDER BY available_at_ms ASC`, so a requeued extract would append its chunk **out of chronological order**, corrupting the per-source buffer/seal ordering. That's why extraction soft-fallbacks (returns `Ok(empty)`, job DONE) and the latch is the only out-of-band record. The correct fix is therefore the latch **lifecycle**, not requeue.

## Solution

In `extract()`, the `Some(extracted)` arm now calls `clear_structure_degraded()` unconditionally. `try_extract` returns `Some` whenever the provider responded (including the valid-but-empty and malformed-JSON-as-empty cases), which is exactly the liveness evidence the latch tracks. The `mark_structure_degraded(ExtractionTimeout)` set path (after exhausted retries) is unchanged.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `extract_clears_structure_latch_on_completed_empty_extraction` covers the completed-but-empty failure path.
- [x] **Diff coverage ≥ 80%** — both changed lines in `llm.rs` (the unconditional clear) and the new test are exercised by the added test.
- [x] N/A: Coverage matrix — behaviour-only fix to an existing extraction path; no feature rows added/removed/renamed.
- [x] No new external network dependencies introduced — test uses an in-process mock `ChatProvider`.
- [x] N/A: Manual smoke checklist — no release-cut surface touched.
- [x] N/A: Linked issue closed via `Closes #NNN` — intentional; this is a partial fix paired with #3630, so #3365 is referenced (not auto-closed) under `## Related`.

## Impact

- Desktop/CLI core only. No migration, no API change. Status/doctor surface now recovers from the structure-degraded warning once the extraction model responds again, regardless of whether the next text contained entities.

## Related

> **Standalone PR**, based directly on `main`. This is an independent fix to the extraction-latch *lifecycle* — it touches only `memory_tree/score/extract/llm.rs` (+ its test) and shares no files or symbols with the job-failure-classification work in #3630, so it does not need to be stacked. (It was originally cut on top of #3630 because both address #3365; that turned out to be an unnecessary dependency and has been removed.)

- Related: #3365 (one of two independent PRs that together address it — the other is #3630 — so not `Closes`).
- Companion PR: #3630 (job-failure status routing; independent, can merge in any order).

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/3365-extraction-latch`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health tracking for extraction operations to correctly reflect completion status regardless of result content.

* **Tests**
  * Added test verifying extraction health status updates properly when operations complete with empty results.


oxoxDev (approved):
LGTM. Correct lifecycle fix.

Verified `try_extract` returns `Some` only after `provider.chat_for_json()` resolves Ok (None on transport error + EOF-truncation; Some(empty) on wrong-shape JSON; Some(parsed) on valid) — so clearing the `ExtractionTimeout` latch unconditionally on `Some` is exactly the liveness signal it tracks. The set path is untouched; the completed-but-empty test covers the regression. CI green, diff-cover ≥80%.

PR #3636, opened 2026-06-12
refactor(cdp): migrate all scanners to in-process transport; drop TCP DevTools port

oxoxDev:
## Summary

- Migrate every CEF-backed scanner (whatsapp, discord, slack, telegram, wechat, meet) to the in-process CDP transport introduced in #3498. The session opener was already on it; this PR finishes the cutover.
- Stop passing `--remote-debugging-port=19222` to CEF at startup. The Tauri shell no longer binds a loopback DevTools listener.
- Delete the legacy TCP WebSocket attach path: `CdpConn::open_ws`, `browser_ws_url`, `connect_and_attach_matching`, `CDP_HOST`, `CDP_PORT`, and the four scanner-local `CdpConn` duplicates (whatsapp, discord, slack, telegram).
- Add label-keyed counterparts of the account-keyed CDP helpers (`install_for_label` / `conn_for_label` / `connect_and_attach_matching_in_process_by_label`) so Meet's `meet-call-<request_id>` webview can attach without a TCP listener.
- Extend the in-process transport with `call_with_timeout` so Slack's IDB batch serialise (`Runtime.callFunctionOn` over ~32 RemoteObjects) keeps the 60s budget it had on the legacy transport.

## Problem

#3498 landed the in-process CDP transport (`Webview::send_dev_tools_message` + `on_dev_tools_protocol`) but only migrated the per-account session opener. Six scanner modules still attached over `http://127.0.0.1:19222/json/version` → loopback WebSocket. While any one of them kept that code path alive, the Tauri shell had to keep passing `--remote-debugging-port=19222`, which binds an unauthenticated DevTools listener that every same-UID process on the machine can attach to.

Concrete consequences:

- The earlier in-process transport did not actually remove the legacy listener — it just stopped using it from the session opener.
- Any same-UID process could attach via `puppeteer.connect({ browserURL: 'http://127.0.0.1:19222/' })` and read or modify the embedded webviews (WhatsApp, Slack, Telegram, Discord, etc.).
- Five copies of `struct CdpConn` (whatsapp / discord / slack / telegram / meet scanners) plus their per-module `browser_ws_url()` and `parse_targets()` duplicates kept piling drift on top of the same machinery.

### CDP transport

- `cdp/in_process.rs`: extract the registry-lookup-or-install body into `install_for_label(label)` and rewrite `install_for_account` as `install_for_label(format!("acct_{id}"))`. Required because Meet's CEF window uses label `meet-call-<request_id>`, not `acct_<id>`. `WebviewCdpTransport::call` is split into `call`/`call_with_timeout`; default `call` keeps the 35s budget, `call_with_timeout` lets Slack's IDB walker pass 60s.
- `cdp/target.rs`: add `conn_for_label<R>(app, label)` and `connect_and_attach_matching_in_process_by_label<R, F>(app, label, pred)` as label-keyed counterparts of the existing account-keyed helpers. Both fall back to `install_for_label` on a cache miss so a transient install race at window creation doesn't permanently lock the surface out of CDP.
- `cdp/conn.rs`: rewrite `CdpConn` as a thin wrapper over `Arc<WebviewCdpTransport>`. Drop `LegacyWs`, `connect_async`, `tokio_tungstenite::Message`, the `oneshot` pending table, and `LEGACY_CALL_TIMEOUT`. `call` / `call_with_timeout` / `pump_events` forward unconditionally to the in-process transport.
- `cdp/mod.rs`: drop `CDP_HOST` / `CDP_PORT` and the `browser_ws_url` / `connect_and_attach_matching` re-exports; add `conn_for_label`, `connect_and_attach_matching_in_process[_by_label]`, `install_for_label` to the surface.

### Scanners

For every scanner, the per-tick attach now reads:

```rust
let pred = move |t: &crate::cdp::target::CdpTarget| -> bool {
    t.url.starts_with(&url_prefix) && t.url.ends_with(&url_fragment)
};
let (mut cdp, session) =
    crate::cdp::target::connect_and_attach_matching_in_process::<R, _>(
        &app, &account_id, pred,
    )
    .await?;
```

For Meet (which is window-keyed, not account-keyed):

```rust
let label = crate::meet_call::window_label_for(&request_id);
let (mut cdp, session) =
    crate::cdp::target::connect_and_attach_matching_in_process_by_label::<R, _>(
        &app, &label, pred,
    )
    .await?;
```

Per-scanner deltas:

- **whatsapp**: drop local `CDP_HOST`/`CDP_PORT`, local `CdpConn`, `browser_ws_url`, `parse_targets`, `CdpTarget`; `scan_dom_once` now takes `&AppHandle<R>` so the in-process transport can be looked up; sibling `dom_snapshot.rs` and `idb.rs` switch their `CdpConn` import to `crate::cdp::CdpConn`.
- **discord**: same drops; the streaming MITM path uses `crate::cdp::CdpConn::pump_events` with a closure capturing `&mut DiscordIngestState`. V1 doesn't issue in-stream calls, so dropping the unused `pending: HashMap` is a simplification, not a regression. `dom_scan_once` gains `&AppHandle<R>` + `account_id`.
- **slack**: drops the bigger of the two local `CdpConn` duplicates (the one with `call_with_timeout`); both `scan_once` and `dom_scan_once` now take `&AppHandle<R>` and reuse the per-account in-process channel for both target enumeration and attach. The pin / strict-fragment / relaxed `/client` fallback hierarchy is preserved verbatim.
- **telegram**: mirror of slack, no pin hierarchy.
- **wechat**: single legacy-attach callsite swapped.
- **meet** (scanner + audio inject + video inject + caption listener): all four TCP attach paths route through the label-keyed in-process helper using `meet_call::window_label_for(request_id)`. `install_audio_bridge`, `spawn_diagnostics_poller`, and `wait_for_meet_target` thread `&AppHandle<R>` + `request_id` through.

### Wiring

- `meet_call/mod.rs`: after `WebviewWindowBuilder::build()`, call `crate::cdp::install_for_label(&label)` so the Meet webview's in-process transport is wired up before any downstream bridge attaches. Idempotent; on transient install failure the downstream `conn_for_label` retries.
- `lib.rs`: delete `args.push(("--remote-debugging-port", Some("19222")))` so the embedded CEF browser never binds a TCP DevTools listener at startup. Surrounding comments rewritten to match.
- `webview_accounts/mod.rs`: comment refresh — the listener no longer exists.
- `gmessages_scanner/mod.rs`: TODO comment refreshed to point at `conn_for_account` / `connect_and_attach_matching_in_process` instead of the long-gone `CDP_HOST` / `CDP_PORT`.

## Submission Checklist

- [x] N/A: this PR is a pure transport refactor — every per-scanner code path keeps the same predicate, attach shape, and `Target.getTargets` resolution. Existing test coverage for each scanner exercises the swapped attach path unchanged. (Tests added or updated per Testing Strategy.)
- [x] N/A: refactor-only change with no business-logic deltas; the diff is dominated by deletions (~1180 LOC) and the new code lives behind the same call signatures the existing test suite already drives. (Diff coverage ≥ 80%.)
- [x] N/A: no feature additions/removals/renames; the test coverage matrix already lists each scanner. (Coverage matrix updated.)
- [x] N/A: no feature IDs to enumerate. (Affected feature IDs listed under Related.)
- [x] No new external network dependencies introduced (mock backend used per Testing Strategy).
- [x] N/A: no release-cut surface touched. (Manual smoke checklist updated if this touches release-cut surfaces.)

## Impact

- **Runtime / platform**: desktop CEF builds. CEF no longer binds a loopback DevTools listener for the lifetime of the embedded browser. Any tooling that previously attached via `http://127.0.0.1:19222/json/version` will fail with `ECONNREFUSED`; nothing in-tree depends on that surface.
- **Security**: removes an unauthenticated same-UID attach surface from the CEF browser. After this PR, CDP traffic flows only through the in-process `Webview::send_dev_tools_message` channel, which stays within the process boundary.
- **Performance**: net deletion of ~750 LOC and removal of the per-attach `connect_async` handshake + tungstenite read loop. Attach setup goes from a TCP `/json/version` GET + WebSocket handshake to a single registry lookup.
- **Migration / compatibility**: external developers that scripted the loopback DevTools port for one-off debugging will need to use `tauri::Webview::open_devtools()` or attach from inside the running process. No production code path relied on it.

## Related

- Closes:
- Follow-up PR(s)/TODOs: `tokio-tungstenite` stays in `Cargo.toml` because `meet_video/frame_bus.rs` and `webview_apis/server.rs` still use it for their WebSocket *servers*. Once those land their own in-process replacements the crate can be dropped.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/7979-scanner-migration`
- Commit SHA: `e46b1ad97`

### Validation Run

- [x] `pnpm typecheck`
- [x] Focused tests: N/A — refactor-only, existing scanner test suites exercise the swapped attach path unchanged
- [x] Rust fmt/check (if changed): `cargo fmt --check` + `cargo check --manifest-path app/src-tauri/Cargo.toml`
- [x] Tauri fmt/check (if changed): same as above

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: scanners + meet bridges no longer attach to CEF via a loopback TCP DevTools port; they go through the in-process `Webview::send_dev_tools_message` channel introduced in #3498.
- User-visible effect: none on the happy path. Any external tooling that scripted `127.0.0.1:19222/json/version` will fail with `ECONNREFUSED`.

### Parity Contract

- Legacy behavior preserved: per-scanner attach predicates (`url.starts_with(url_prefix) && url.ends_with(url_fragment)`), Slack's pin-on-strict-match resolution hierarchy, Slack's 60s IDB serialise budget (now via `call_with_timeout`), Discord's per-event dispatch shape (now via canonical `pump_events`), Meet's per-call label routing (`meet-call-<request_id>`).
- Guard/fallback/dispatch parity checks: every scanner still falls through to its legacy log+continue branch on attach failure; `conn_for_label` / `conn_for_account` retry `install_for_label` / `install_for_account` once on a registry miss so a transient install race at window creation doesn't permanently lock the surface out of CDP.

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Release Notes

* **Refactor**
  * Refactored Chrome DevTools Protocol (CDP) communication from WebSocket-based remote debugging to in-process transport, improving system architecture and eliminating unnecessary network listeners.
  * Updated Discord, Slack, Telegram, WhatsApp, and other account scanners to use the new in-process CDP connection method.
  * Removed legacy TCP remote debugging port configuration in favor of the streamlined in-process approach.


app/src-tauri/src/discord_scanner/mod.rs
```diff
@@ -384,204 +350,16 @@ async fn run_mitm_session<R: Runtime>(
...
-            }
-            let params = v.get("params").cloned().unwrap_or(Value::Null);
-            dispatch_event(app, account_id, method, &params, &mut ingest_state);
-        }
-    }
+    // Drop into the event read loop until the in-process channel signals
+    // closure. V1 doesn't issue any in-stream calls (responses table from
+    // the previous TCP impl is gone — re-introduce a request/response API
+    // here when V1.5 backfills `Network.getResponseBody`).
+    log::info!("[discord][{}] event pump started", account_id);
```
sanil-23, line 357: **Major (medium confidence) — tracked in #3693.** This long-lived pump can't self-heal a dead session: `pump_events` only returns on `RecvError::Closed` (transport/webview teardown), not when the *attached page target* dies (Discord Ctrl+R, renderer crash, SPA hard-nav). `Target.detachedFromTarget` isn't acted on (discovery never enabled; `dispatch_event` ignores unknown methods), so the outer reconnect loop is never re-entered and the MITM silently stops capturing until the webview is destroyed.

Suggest an idle watchdog (> Discord's ~41s gateway heartbeat, e.g. 90s) so a stalled/destroyed session forces a return → re-attach. Also: the `run_mitm_session` doc comment above ("Returns when the underlying WebSocket closes, the page target disappears, or any dispatch hits an unrecoverable error") is now false for all three clauses — please correct it in this PR.

app/src-tauri/src/cdp/conn.rs
```diff
@@ -123,137 +69,43 @@ impl CdpConn {
...
+                    method,
+                    params,
+                    session_id: evt_session,
+                }) => {
+                    if !evt_session.is_empty() && evt_session != session_id {
+                        continue;
                     }
+                    on_event(&method, &params);
+                }
+                Err(RecvError::Lagged(skipped)) => {
```
sanil-23, line 97: **Major (medium confidence) — tracked in #3693.** On `Lagged` the pump logs and `continue`s, silently dropping CDP frames once the single consumer falls >256 behind (`EVENT_CHANNEL_CAP`). The comment above states recovery relies on "the caller's idle watchdog" — but the Discord consumer (the only long-lived pump) has no watchdog, so dropped `Network.*` frames during a burst = missed messages with no re-sync. Either validate 256 against Discord's burst volume or drain this consumer into an unbounded queue, and reword the comment so it doesn't promise a watchdog that doesn't exist for this caller.

sanil-23 (dismissed):
## Approve ✅ (with comments — one fast-follow filed)

Security-positive, carefully executed refactor that finishes the in-process CDP cutover and closes the unauthenticated same-UID loopback DevTools surface. CI is fully green. The mechanical migration of all six scanners preserves every predicate, attach shape, pin/fallback hierarchy, and log-and-continue error path; the `get_or_create` install path correctly serializes observer registration under the registry mutex (CEF has no un-register hook), and pending-map cleanup is leak-free on all three exit paths.

One **Major (medium-confidence)** concern is Discord-specific and tracked as a fast-follow rather than gating this security win: **#3693**. Discord is the only scanner using the long-lived `pump_events` loop, and (1) it can't self-heal a dead/stale session — `pump_events` only returns on transport teardown, not on page-target loss (Ctrl+R / renderer crash / SPA hard-nav), so it never re-attaches until the webview is destroyed; and (2) the 256-slot broadcast drops frames on `Lagged` with no idle watchdog, despite the `cdp/conn.rs` comment promising one. See inline comments.

Non-blocking items to fix in this PR (zero-risk):
- The `run_mitm_session` doc comment ("Returns when the underlying WebSocket closes, the page target disappears…") and the `cdp/conn.rs` Lagged comment ("the caller's idle watchdog will eventually time out the session") no longer match the in-process implementation — please correct both.
- `cdp/in_process.rs:519` cites integration test `cdp_in_process_e2e.rs`, which doesn't exist anywhere in the repo (pre-existing from #3498). Either land it or drop the reference; `cdp/conn.rs` + `cdp/target.rs` currently ship with zero unit tests, and the new `pump_events` session-id filter is pure, testable logic.

Minor / low-confidence: `call_with_timeout` changed from a per-read timeout (reset on every inbound frame) to a total wall-clock cap. For Slack's 60s IDB walk (`slack_scanner/idb.rs:314`) this is almost certainly equivalent (dedicated session, no interleaved frames), but it's a real semantic change worth a line in the parity contract rather than "preserved verbatim."

Nice work overall — approving on the basis that #3693 owns the Discord watchdog and the doc/test-reference cleanups land here.

PR #3638, opened 2026-06-12
perf(memory_tree): HashSet for order-irrelevant dedup in chunk scoring

mysma-9403:
## Summary

- The per-chunk scoring/extraction path builds several sets purely to **count distinct items** or **guard membership** — never to read iteration order — yet used `BTreeSet`, paying O(log n) ordered-key comparisons on every insert.
- Swap all four to `HashSet`. The produced counts and merge results are byte-identical; only the per-insert cost drops from ordered (String) comparisons to a single hash.

## Sites changed

| Site | Set is used for | Why order is irrelevant |
| --- | --- | --- |
| `unique_words::score` | `uniq.len()` (type-token ratio) | only the cardinality is read |
| `ExtractedEntities::unique_entity_count` | `.len()` of distinct `(kind, text)` | collected solely to count |
| `ExtractedEntities::merge` — `seen` | `.insert()` membership dedup | surviving order is the existing `Vec` push order, not the set's |
| `ExtractedEntities::merge` — `topic_seen` | `.insert()` membership dedup | same — `Vec` push order is preserved |

## Why it's safe

- Every set is consumed only via `.len()` or `.insert()` (bool membership). None is ever iterated, so switching from an ordered to a hashed set cannot change output.
- Key types already derive `Hash` (`EntityKind` derives `Hash, Eq`; tuples of `Hash` are `Hash`; `String` is `Hash`).
- No API, signature, or dependency change.

## Impact

- Runs for **every chunk** on ingest / re-embed backfill, so the saving compounds over bulk syncs.
- `unique_words` specifically targets very low-diversity text ("yay yay yay …"); on heavily-repeated words the ordered-set comparison churn was the most wasteful, and that is exactly what this removes.
- Modest, CPU-only, behavior-preserving — no latency cliff, just less per-insert work on a hot path.

## Submission Checklist

- [x] Behavior-preserving: counts and merge results are identical (sets used only for `.len()` / membership).
- [x] Existing `memory_tree::score` tests pass (157 passed, 0 failed), including `unique_words` ratio tests and `extract::types` merge/dedup tests.
- [x] No new dependency, no schema/migration, no public API change.
- [x] N/A: no linked issue — proactive, low-risk hot-path cleanup.

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal data structure handling in the scoring engine to improve performance and efficiency while maintaining identical functionality and behavior.


M3gA-Mind (approved):
Approved by the maintainer review fleet. ✅ Behavior-preserving hot-path cleanup — switching the order-irrelevant sets to `HashSet` for membership/cardinality. Verified every affected set is used only for membership/count, all `Hash` bounds are satisfied, and there's no API-surface change. 17/17 required checks green, CodeRabbit approved. Thanks @mysma-9403!

Issue #3224, opened 2026-06-02
Host of problems lead to segmentation faults


MrVulcan68:
AppImage Segfaults on Arch Linux due to hardcoded RPATHs and missing libxdo
Environment

    OS: Arch Linux (Rolling Release)

    AppVersion: OpenHuman_0.57.5_amd64.AppImage

    Architecture: x86_64

Problem Description

The AppImage fails to launch on Arch Linux, throwing a zsh: segmentation fault (core dumped) instantly. Debugging via LD_DEBUG=libs revealed two distinct underlying issues blocking execution.
Technical Details & Root Causes

1. Missing libxdo.so.3 dependency bundling
The application actively looks for libxdo.so.3 (provided by xdotool). On modern rolling-release distributions like Arch, this package has moved forward to libxdo.so.4. Because the legacy version isn't bundled inside the AppImage, the app panics and crashes on newer host systems.

    Fix: Please bundle libxdo.so directly within the AppImage shared/lib payload.

2. Hardcoded GitHub Actions Runner paths (RPATH/RUNPATH)
Upon extracting the AppImage via --appimage-extract, running find on libcef.so revealed that the binary's internal library search paths point to absolute directories on the compilation server:
Plaintext

squashfs-root/shared/lib/home/runner/.cache/tauri-cef/...
squashfs-root/shared/lib/home/runner/work/openhuman/...

Because the loader tries to access these non-existent paths on the host system to find libcef.so, it causes an unrecoverable segmentation fault.

    Fix: Ensure the Tauri/CEF build configuration sets relative search paths (using $ORIGIN) rather than absolute paths tied to the CI/CD build environment. For example, the RUNPATH on the final binary should look like:
    $ORIGIN/../shared/lib or $ORIGIN/../lib

Suggested Resolution

Update the AppImage packaging configuration to cleanly strip absolute build-machine paths from the binary's headers using a tool like patchelf or updating the bundler scripts, and ensure libxdo is explicitly packaged as a dynamic asset.

0fbcb238c0:
Every single new release segfaults lol.
Actually unusable software

PR #3645, opened 2026-06-13
fix(appimage): always replace bundled ld-linux from host system to prevent glibc mismatch crash

### LINKED ISSUE #3099: [IMPORTANT] - "Interpreter not found!" across all install methods on Arch Linux — AppImage, .deb, and autoinstall all failing
User ᏒᏗᏂ-ᎥĐ905 on Arch Linux (x86_64) reports "Interpreter not found!" when running OpenHuman via any installation method — AppImage, .deb package, and autoinstall all fail with the same error. User previously got the app running with AI assistance but can no longer launch it.

Environment
OS: Arch Linux (x86_64)
Rust: 1.95.0 / Cargo: 1.95.0
Clang: 22.1.5 (installed at /usr/bin)
CMake: 4.3.3 | Node: v26.1.0 | pnpm: 10.10.0
GCC: 16.1.1 (x86_64-pc-linux-gnu)
ROCm: 3.5.0 (at /mnt/data/LLMz/rocm) — AMD GPU setup
Note: $env:LIBCLANG_PATH syntax failed (Windows PowerShell syntax used in zsh — not a blocker but worth noting)

Related known regression
Same "Interpreter not found!" error as Tekguru's ticket (Ubuntu 26.04, v0.54.0 AppImage). Root cause: AppImage missing ld-linux-x86-64.so.2. However this user is on Arch Linux and also seeing the issue on .deb and autoinstall — suggesting either a broader packaging problem or an Arch-specific glibc/dynamic linker path issue beyond the AppImage regression. Arch uses a non-standard filesystem layout and bleeding-edge glibc that may expose additional incompatibilities.

Connor confirmed this is the AppImage issue on 27/05/2026 — response pending from user.

---
**Reported by:** @rah905
**Discord ticket:** https://discord.com/channels/1470856511193616498/1507250596405710878
**Priority:** high
**Attachments (2):** image.webp, image (1).webp — ask support team for files

0xhis:
The lib4bin step in sharun-cef bundling copies the CI runner's
ld-linux-x86-64.so.2 into shared/lib/, which is glibc 2.39 on
Ubuntu 24.04. On newer glibc 2.42 systems (Fedora, rolling
distros) this causes an ABI mismatch and immediate `*** stack
smashing detected ***` crash.

Observed on Fedora Linux — distros shipping glibc ≥ 2.42 may be
affected.

`ensure_sharun_interpreter()` previously skipped replacement when
the bundled ld-linux already existed. Now it always replaces from
the CI host system, which is guaranteed compatible with the binary
compiled on the same runner.

This does not claim to resolve all of #3224; it fixes the ld-linux/glibc mismatch observed on Fedora.

## Change
- **File**: `scripts/release/strip-appimage-graphics-libs.sh` (+5/-3)
- **Function**: `ensure_sharun_interpreter()` — replace `[ -e "$target" ] && return 1` guard with `rm -f "$target"`
- **Effect**: CI runner's host `ld-linux-x86-64.so.2` (or `ld-linux-aarch64.so.1`) always replaces whatever `lib4bin` bundled

## Verified
1. **Negative control**: Original v0.57.40 AppImage on glibc 2.42 (Fedora) → `*** stack smashing detected ***: terminated`
2. **Positive control**: Same AppImage after fix applied → app launched, CEF initialized, RunEvent::Ready, zero crashes
3. **Syntax**: `bash -n` passes
4. **Return semantics**: No change to function return values (only one `return 1` remains — the non-sharun early exit)

Refs #3224
Closes #3099

## Summary by CodeRabbit

* **Chores**
  * Improved the release build process to ensure AppImage packages handle graphics library loaders more robustly, replacing any incompatible loaders with compatible host versions.

### DIFF
--- a/scripts/release/strip-appimage-graphics-libs.sh
+++ b/scripts/release/strip-appimage-graphics-libs.sh
@@ -265,9 +265,11 @@ ensure_sharun_interpreter() {
   fi
 
   local target="$appdir/lib/$loader_name"
-  if [ -e "$target" ]; then
-    return 1
-  fi
+  # Always replace — lib4bin may bundle an ld-linux from the CI runner
+  # that is incompatible with newer host glibc (#3224, #3099).
+  # The host_dynamic_loader source is the CI runner's own system ld-linux,
+  # which is guaranteed compatible with the binary compiled on the same runner.
+  rm -f "$target"
 
   local source
   if ! source="$(host_dynamic_loader "$loader_name")"; then


0xhis:
Follow-up on CodeRabbit pre-merge warnings:

- **Linked issue scope**: corrected from `Closes #3224` to `Refs #3224` in both the commit and PR body. This PR intentionally fixes the Fedora/glibc ld-linux mismatch only; it does not claim to resolve all of #3224.
- **Docstring coverage**: not applicable here. The change is in a Bash release script (`scripts/release/strip-appimage-graphics-libs.sh`), not Python/Rust/TS code with docstring coverage expectations.

CodeRabbit also reported no actionable comments and approved commit `3c1bedcf`.

oxoxDev (approved):
Approving — tight, well-scoped build-script fix. Verified:

- **Return semantics intact** — only remaining `return 1` is the non-sharun early exit (L257); caller `if ensure_sharun_interpreter; then added_loader=1` still flips correctly, so the AppImage-changed accounting stays accurate.
- **Fail-loud preserved** — `rm -f` runs *before* the `host_dynamic_loader` lookup, and a missing host loader still `exit 1`s the whole build, so no half-bundled / loader-less artifact can ship.
- **Safe under no-`set -e`** — `rm -f` on an absent target is a no-op; idempotent across re-runs; `cp -L` dereferences loader symlinks.
- All 19 CI lanes green + CodeRabbit approved.

One optional doc nit (non-blocking): the comment at L268-270 says lib4bin bundles "an ld-linux from the CI runner" — but the bug only makes sense if lib4bin bundles the **sharun toolchain's pinned** loader (a runner→runner swap would be a no-op). The mechanism is correct; the wording could read "lib4bin may bundle a loader whose glibc differs from the binary compiled on the runner."

Note for tracking: this `Refs` (not `Closes`) #3224 correctly — the libxdo.so.3 bundling + libcef RPATH `$ORIGIN` strip still need a follow-up. #3099 looks closeable once this lands.

PR #3657, opened 2026-06-13
fix(cli): restore SIGPIPE default on Unix

### LINKED ISSUE #3535: Core CLI panics with 'Broken pipe' when output is piped to a reader that closes early
## Summary

The `openhuman-core` CLI aborts with a fatal Rust panic — `failed printing to stdout: Broken pipe (os error 32)` — whenever its output is piped to a reader that closes the pipe early (e.g. `openhuman-core --help | head`). Instead of exiting cleanly, the process panics, which is reported to Sentry as an unhandled `level=fatal` crash.

## Problem

**What happens:** On Unix, the Rust runtime installs `SIG_IGN` for `SIGPIPE` at startup. With the signal ignored, writes to a broken pipe return `EPIPE` instead of terminating the process, and the `print!`/`println!`/`eprint!` macros (used for the banner and help text in `run_from_cli_args`) **panic** on that write error. The result is a fatal panic on a completely normal shell idiom.

**What's expected:** Piping CLI output into `head`, `grep -q`, `less` (then quitting), etc. should terminate the process silently with the conventional exit status — never panic or emit a crash report.

**Impact:**

- Multiple distinct real users affected (Sentry geo: Brisbane 🇦🇺, Pingxiang 🇨🇳 — not a single tester).
- `mechanism=panic`, `handled=no`, `level=fatal`. Generates recurring fatal-crash noise in Sentry (~12 events across the two groups below).
- Linux-only in the reports (WSL2), but the root cause applies to any Unix target.

**Steps to reproduce:**

1. Build the core binary: `cargo build --bin openhuman-core`.
2. Run a command whose output goes to stdout and pipe it to a reader that exits early:
   ```bash
   ./target/debug/openhuman-core --help | head -n 1
   # or any subcommand that prints to stdout/stderr, piped to `head`
   ```
3. The reader closes the pipe after the first line; the next `println!`/`eprint!` write returns `EPIPE` and the process panics:
   ```
   panic: failed printing to stdout: Broken pipe (os error 32)
   ```

**Version / platform:** `openhuman@0.53.43+2b64ea8ac567`, `environment=production`, Linux (`6.6.87.2-microsoft-standard-WSL2`), `rustc 1.93.0`. Culprit frame: `openhuman_core::core::cli::run_from_cli_args` (`src/core/cli.rs`).

## Solution (optional)

**Suspected cause:** No `SIGPIPE` handling and no panic hook exist anywhere in the core binary (`src/main.rs`, `src/core/cli.rs`). The banner (`eprint!("{CLI_BANNER}")` at `src/core/cli.rs:49`) and help text (`print_general_help`, etc.) write through the standard print macros, which panic on `EPIPE`.

**Proposed fix (idiomatic):** Reset `SIGPIPE` to its default disposition at the very start of `main()` on Unix, so a broken pipe terminates the process normally instead of surfacing as an `EPIPE` write error that the print macros turn into a panic. This is the approach used by `ripgrep`, `fd`, and most Rust CLIs.

```rust
// src/main.rs, first lines of main()
#[cfg(unix)]
unsafe {
    libc::signal(libc::SIGPIPE, libc::SIG_DFL);
}
```

`libc` is already in `Cargo.lock` (transitive); promote it to a direct dependency for `openhuman-core`. Guard with `#[cfg(unix)]` so Windows is unaffected.

Alternatives if a signal reset is undesirable: route CLI output through an explicit `writeln!(io::stdout(), …)` path that treats `ErrorKind::BrokenPipe` as a clean exit, or install a panic hook that swallows broken-pipe panics. The `SIG_DFL` reset is simplest and matches ecosystem convention.

## Acceptance criteria

- [ ] **Repro gone** — `openhuman-core --help | head -n 1` (and other piped-to-early-closing-reader invocations) exit cleanly with no panic and no Sentry event.
- [ ] **Regression safety** — A test exercises the broken-pipe path (e.g. spawn the binary, pipe stdout to a reader that drops the pipe, assert no panic / expected exit status).
- [ ] **Cross-platform** — Windows build unaffected (`#[cfg(unix)]` guard verified).

## Related

Sentry issues (org `tinyhumans`, project `openhuman-core`):

- [OPENHUMAN-CORE-C](https://tinyhumans.sentry.io/issues/OPENHUMAN-CORE-C) — `panic: failed printing to stdout: Broken pipe (os error 32)` (10 events, culprit `run_from_cli_args`)
- [OPENHUMAN-CORE-1Q](https://tinyhumans.sentry.io/issues/OPENHUMAN-CORE-1Q) — `panic: failed printing to stderr: Broken pipe (os error 32)` (2 events, culprit `main`) — same root cause (banner written to stderr)

_Filed via the Sentry bug triage workflow._

YonganZhang:
## Summary
- restore the default SIGPIPE disposition at process startup on Unix
- prevents Rust print macros from panicking on normal closed-pipe CLI usage like `openhuman-core --help | head`
- add a Unix integration test that closes the stdout pipe and verifies stderr does not contain a broken-pipe panic

Closes #3535

## Tests
- `cargo fmt --check`
- `git diff --check`

Attempted but blocked locally:
- `env CC=/usr/bin/gcc CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=/usr/bin/gcc cargo test --test cli_sigpipe -- --nocapture`
  - blocked by missing system pkg-config deps: `alsa.pc` and `xi.pc`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that occurred when the CLI's output pipe was closed before completion (e.g., when piping to commands like `head`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5180,6 +5180,7 @@ dependencies = [
  "keyring",
  "landlock",
  "lettre",
+ "libc",
  "log",
  "mail-parser",
  "matrix-sdk",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ tokio = { version = "1", features = ["full", "sync"] }
 once_cell = "1.19"
 parking_lot = "0.12"
 log = "0.4"
+libc = "0.2"
 nu-ansi-term = "0.46"
 env_logger = "0.11"
 base64 = "0.22"
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,8 @@ use regex::Regex;
 /// information is redacted before being sent to the server. After setup, it
 /// delegates execution to the core library based on CLI arguments.
 fn main() {
+    restore_default_sigpipe();
+
     // Load `.env` before `sentry::init` so a DSN defined only in the dotenv
     // file is visible to the Sentry client at startup. `dotenvy::dotenv()` is
     // a no-op for variables already present in the process environment, and
@@ -179,6 +181,19 @@ fn main() {
         std::process::exit(1);
     }
 }
+
+#[cfg(unix)]
+fn restore_default_sigpipe() {
+    // Rust ignores SIGPIPE at startup. That makes writes to a closed pipe
+    // return EPIPE, which the print macros turn into a panic. CLI tools should
+    // instead terminate quietly when a downstream reader such as `head` exits.
+    unsafe {
+        libc::signal(libc::SIGPIPE, libc::SIG_DFL);
+    }
+}
+
+#[cfg(not(unix))]
+fn restore_default_sigpipe() {}
 
 // ---------------------------------------------------------------------------
 // Release / environment resolution for Sentry
--- a/tests/cli_sigpipe.rs
+++ b/tests/cli_sigpipe.rs
@@ -1,0 +1,26 @@
+#[cfg(unix)]
+#[test]
+fn help_output_closed_pipe_does_not_panic() {
+    use std::process::{Command, Stdio};
+
+    let mut child = Command::new(env!("CARGO_BIN_EXE_openhuman-core"))
+        .arg("--help")
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn openhuman-core --help");
+
+    drop(child.stdout.take());
+
+    let output = child.wait_with_output().expect("wait for openhuman-core");
+    let stderr = String::from_utf8_lossy(&output.stderr);
+
+    assert!(
+        !stderr.contains("Broken pipe"),
+        "stderr must not include a broken-pipe panic: {stderr}"
+    );
+    assert!(
+        !stderr.contains("panicked"),
+        "stderr must not include a panic report: {stderr}"
+    );
+}
--- a/tests/cli_sigpipe.rs
+++ b/tests/cli_sigpipe.rs
@@ -1,6 +1,8 @@
 #[cfg(unix)]
 #[test]
 fn help_output_closed_pipe_does_not_panic() {
+    use std::io::{BufRead, BufReader};
+    use std::os::unix::process::ExitStatusExt;
     use std::process::{Command, Stdio};
 
     let mut child = Command::new(env!("CARGO_BIN_EXE_openhuman-core"))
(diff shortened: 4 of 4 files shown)


Felyx-Fu:
I checked the failing Rust Core Coverage (cargo-llvm-cov) job. The tests completed and lcov-core.info was generated; the failure happened in post-job cleanup/rust-cache while running cargo metadata --all-features, which failed to unpack windows v0.57.0 because the runner hit No space left on device.

I tried gh run rerun 27481248697 --failed, but GitHub rejected it because I do not have admin rights. This looks like an infra/cache disk-space failure rather than a code failure; a maintainer rerun should be the next step.

Felyx-Fu:
I addressed the remaining CodeRabbit test nit in a helper PR because direct push to the source branch returned 403: https://github.com/YonganZhang/openhuman/pull/4

What changed:
- The closed-pipe CLI regression test now asserts that the child process either exits successfully or is terminated by SIGPIPE.
- The existing stderr checks for broken-pipe/panic text are preserved.

Validation:
- cargo fmt --manifest-path Cargo.toml --check
- git diff --check -- tests/cli_sigpipe.rs
- cargo test --manifest-path Cargo.toml --test cli_sigpipe was attempted on this Windows environment, but compilation did not finish before a 5 minute timeout. The test body is Unix-only via #[cfg(unix)], so Linux CI is the authoritative execution path.

The existing Rust Core Coverage failure still appears to be the runner disk/cache cleanup issue noted above rather than a code failure.

YellowSnnowmann:
### Review: fix is correct, but the regression test is likely a false-green

**The production fix is sound** ✅ — restoring `SIGPIPE → SIG_DFL` as the first statement in `main()` (before the banner and `sentry::init`) is the idiomatic ripgrep/fd approach and covers *both* Sentry groups (`OPENHUMAN-CORE-C` stdout/help and `OPENHUMAN-CORE-1Q` stderr/banner) since it's a global reset. The `#[cfg(unix)]` guard + non-unix no-op keeps Windows unaffected. No concerns there.

**The regression test (`tests/cli_sigpipe.rs`) doesn't reliably exercise the bug, though** ⚠️

1. **Pipe-buffer race → possible false-green.** `--help` output is a few KB, well under the 64 KB kernel pipe buffer. After `drop(child.stdout.take())`, the child can write its entire output into the buffer in a single `write()` that succeeds immediately, then exit cleanly — **never hitting `EPIPE`, with or without the fix.** So this test can pass even if the bug were reintroduced. To actually force the broken-pipe path, mirror the issue's repro (`--help | head -n 1`): read *one* line, then close, so the *next* write fails — or emit output larger than the pipe buffer.

2. **No exit-status assertion.** Acceptance criterion #1 asks for a clean *exit status*; the test only greps stderr for `"Broken pipe"`/`"panicked"` and ignores `output.status`. CodeRabbit's nit (assert `status.success() || status.signal() == Some(SIGPIPE)`) addresses this, but that change currently lives in a helper PR against the fork (`YonganZhang/openhuman#4`) and is **not in this branch** — the diff here still has the weaker test.

**Suggested before merge:** fold the exit-status assertion *and* a repro that truly triggers `EPIPE` (read-one-line-then-close, or oversized output) directly into this branch, so "regression safety" is met substantively rather than nominally.

PR #3663, opened 2026-06-14
fix(observability): drop all-transient provider exhaustion events

### LINKED ISSUE #3542: Aggregate 'All providers/models failed' surfaces as Sentry error on all-transient attempts
## Summary

The aggregate `All providers/models failed. Attempts: …` event surfaces to Sentry as a `level=error`. Per the `before_send` comment in `src/main.rs`, this aggregate is **intentionally kept** (it indicates a genuine outage after retries/fallback are exhausted) — but in this dataset it fired as downstream noise when every attempt was a transient 502/503/504, so it is worth confirming the classification is correct.

## Problem

When all configured providers/models fail, the reliable-provider layer emits a single aggregate "all providers exhausted" error. This is correct for genuine outages, but if every underlying attempt was a transient/suppressed failure (e.g. all 502/503/504), the aggregate can fire as error-level noise that mirrors a transient blip rather than a real config/outage problem.

- **Impact:** ~2 events; co-occurred with the transient Composio/provider 5xx events in the same session.
- **Repro:** Force every configured provider/model attempt to return a transient 5xx (or be unreachable) within one request.
- **Version / platform:** `openhuman@0.53.43`, `environment=production`. Provider/network — platform-agnostic.

## Solution (optional)

This is **expected signal by design** — `main.rs` explicitly keeps the aggregate exhaustion event so genuine outages stay visible. No suppression intended. **Action:** decide whether the aggregate should distinguish "exhausted after all-transient attempts" (demote to warning / attach `transient=true`) from "exhausted after real errors" (keep error), so the error stream reflects only actionable exhaustion. If the current behavior is desired, document it and close — this issue exists to record the maintainer review.

## Acceptance criteria

- [ ] **Decision recorded** — Confirm whether the all-transient-attempts case should stay `level=error` or be demoted, and document the rationale.
- [ ] **Regression safety** — If behavior changes, a test covers the all-transient vs mixed-error exhaustion classification.
- [ ] **Diff coverage ≥ 80%** — fix PR (if any) meets the changed-lines coverage gate.

## Related

- [OPENHUMAN-CORE-1M](https://tinyhumans.sentry.io/issues/OPENHUMAN-CORE-1M) — `All providers/models failed. Attempts:` (2 events)
- Context: `src/main.rs` `before_send` (aggregate exhaustion intentionally kept), #2205 (custom_openai fallback)

_Filed via the Sentry bug triage workflow (intentional aggregate signal — included at maintainer request for review)._

YonganZhang:
Fixes #3542

## Summary
- add a narrow before_send classifier for failure=all_exhausted provider events whose Attempts body contains only transient HTTP/transport failures
- keep ordinary all_exhausted events, including mixed/permanent attempts, visible in Sentry
- add smoke coverage for all-transient aggregate drop and mixed/permanent aggregate retention

## Tests
- cargo fmt --check
- git diff --check

Attempted but blocked locally:
- cargo test --test observability_smoke

The test build fails before this test runs because the local C linker rejects Rust's -m64 flag (`error: unknown option '-m64'`) while compiling dependency build scripts.

## Summary by CodeRabbit

* **Chores**
  * Improved error monitoring to better distinguish between temporary and permanent provider failures, reducing noise in error reports.

### DIFF
--- a/src/core/observability.rs
+++ b/src/core/observability.rs
@@ -1842,6 +1842,46 @@ pub fn is_transient_provider_transport_failure(event: &sentry::protocol::Event<'
     event_has_transient_transport_phrase(event)
 }
 
+/// Defense-in-depth filter for aggregate provider exhaustion events where the
+/// aggregate only restates transient attempt failures.
+///
+/// Keep ordinary `failure=all_exhausted` events: they are the useful "every
+/// fallback failed" signal. Drop only the narrow shape observed in #3542,
+/// where the aggregate body starts with the reliable-provider exhaustion
+/// prefix and contains transient HTTP/transport wording already classified by
+/// [`is_transient_message_failure`].
+pub fn is_all_transient_provider_exhaustion_event(event: &sentry::protocol::Event<'_>) -> bool {
+    let tags = &event.tags;
+    if tags.get("domain").map(String::as_str) != Some("llm_provider") {
+        return false;
+    }
+    if tags.get("failure").map(String::as_str) != Some("all_exhausted") {
+        return false;
+    }
+
+    let direct = event.message.as_deref();
+    let from_logentry = event.logentry.as_ref().map(|log| log.message.as_str());
+    let from_exception = event.exception.last().and_then(|e| e.value.as_deref());
+    [direct, from_logentry, from_exception]
+        .into_iter()
+        .flatten()
+        .any(all_provider_attempts_are_transient)
+}
+
+fn all_provider_attempts_are_transient(message: &str) -> bool {
+    let Some(attempts) = message.strip_prefix("All providers/models failed. Attempts:") else {
+        return false;
+    };
+    let mut saw_attempt = false;
+    for attempt in attempts.split(';').map(str::trim).filter(|s| !s.is_empty()) {
+        saw_attempt = true;
+        if !is_transient_message_failure(attempt) {
+            return false;
+        }
+    }
+    saw_attempt
+}
+
 /// Returns true when a Sentry event's message/exception text contains the
 /// canonical max-tool-iterations cap phrase (see
 /// `openhuman::agent::error::MAX_ITERATIONS_ERROR_PREFIX`).
--- a/src/main.rs
+++ b/src/main.rs
@@ -59,6 +59,11 @@ fn main() {
             if openhuman_core::core::observability::is_transient_provider_http_failure(&event) {
                 return None;
             }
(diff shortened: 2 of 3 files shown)


oxoxDev (changes requested):
The new classifier never fires on real input, and the tests assert a synthetic shape — so they pass while production is mishandled.

- Parser delimiter: `all_provider_attempts_are_transient` splits the message on `;`, but `format_failure_aggregate` (inference/provider/reliable.rs:343) joins attempts with `\n` and formats each as `provider=… attempt N/M: <reason>; error=<detail>`. Splitting that block on `;` does not isolate attempts. The tests use `;`-joined strings the producer never emits, which is why they pass.
- Mixed-aggregate risk: because the check is substring-based over the wrongly-split chunks, an aggregate that contains a permanent failure (e.g. 401) alongside a transient (503) can match a transient token and get dropped — suppressing an event that carries a real, actionable failure.
- Not wired into desktop builds: the filter is installed only in `src/main.rs` (standalone core). The shipped Tauri app has its own `before_send` chain in `app/src-tauri/src/lib.rs` (~2132) that never calls `is_all_transient_provider_exhaustion_event`. The core is in-process under Tauri, so production desktop builds never run this filter and the targeted `failure=all_exhausted` events still reach Sentry.

Requested changes: parse the `\n`-delimited attempt list and match the `error=` segment, require all attempts transient (not substring-any), preserve mixed/permanent aggregates, retest against actual `format_failure_aggregate` output, and mirror the arm into the Tauri `before_send` chain. (Codex flagged the same two issues inline.)

Felyx-Fu:
I addressed the two remaining review threads in a helper PR because direct push to the head fork returned 403: https://github.com/YonganZhang/openhuman/pull/1

What changed:
- mirrored is_all_transient_provider_exhaustion_event in the Tauri shell efore_send chain, so desktop in-process core events are filtered like the standalone core binary
- changed the aggregate attempt parser to split on the production newline delimiter from ormat_failure_aggregate while retaining semicolon compatibility
- updated the smoke tests to use the real newline-delimited aggregate format for both all-transient and mixed/permanent attempts

Validation:
- cargo fmt --manifest-path Cargo.toml --check
- cargo fmt --manifest-path app/src-tauri/Cargo.toml --check

I also attempted cargo test --manifest-path Cargo.toml --test observability_smoke aggregate_all_exhausted, but on this Windows environment it stayed in heavy Rust compilation and timed out after 5 minutes without diagnostics.

PR #3674, opened 2026-06-14
fix(rpc/credentials): preserve anyhow context chain in BackendOAuthClient call sites (TAURI-RUST-10)

CodeGhost21:
## Summary

- Preserve the `anyhow` context chain in three `BackendOAuthClient` call sites in `credentials/ops` so the underlying transport error reaches the rpc dispatcher.
- Drops Sentry **TAURI-RUST-10** ("GET /auth/me", 409 events / 17 users / since 2026-05-19), which was an opaque transient-network noise group caused by `e.to_string()` collapsing the chain.
- Adds defense-in-depth `before_send` filter `is_auth_get_me_opaque_transport_event` so any future regression that re-emits the same opaque shape stays off Sentry.

## Problem

Sentry [TAURI-RUST-10](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/98/) shows 409 events / 17 users since v0.54.0 with:

- `message`/`title`: literally `GET /auth/me`
- tags: `domain=rpc`, `operation=invoke_method`, `method=openhuman.auth_get_me`, `elapsed_ms≈5003`
- No exception, no stack, no status code, no underlying transport text.

Trace:

1. Frontend calls RPC `openhuman.auth_get_me` → `src/openhuman/credentials/ops.rs:451`.
2. `auth_get_me` calls `BackendOAuthClient::fetch_current_user` (`src/api/rest.rs:402`), which does `client.get(url).send().await.context("GET /auth/me")?`. This wraps the underlying `reqwest::Error` (`operation timed out`, `connection reset`, `tls handshake eof`, etc.) under the top context `"GET /auth/me"`.
3. Back in `auth_get_me`: `.map_err(|e| e.to_string())?`. **`anyhow::Error::to_string()` renders only the top-most context** — the underlying transport error is dropped. The string handed up the stack becomes exactly `"GET /auth/me"`.
4. The rpc dispatcher (`src/core/jsonrpc.rs:39`) runs its classifier ladder. `is_transient_message_failure("GET /auth/me")` returns false because `TRANSIENT_TRANSPORT_PHRASES` (`"timeout"`, `"connection reset"`, `"tls handshake eof"`, `"error sending request"`) were stripped in step 3.
5. The classifier misses → `report_error_or_expected("GET /auth/me", "rpc", "invoke_method", …)` → `sentry::capture_message("GET /auth/me", Error)`. Group fingerprint stabilizes on the bare literal → every transient network blip folds into one opaque high-traffic issue.

`elapsed_ms = 5003` + the 17-user / 26-day spread are consistent with episodic transient network failures (launch / resume / sleep-wake), not a real backend regression. The reqwest client has 15s connect / 120s total timeouts, so the ~5s timing is a fast network failure (DNS, refused, transient TLS) — exactly the class `is_transient_message_failure` is built to demote.

Same `.map_err(|e| e.to_string())` anti-pattern existed in sibling endpoints (`consume_login_token`, `auth_create_channel_link_token`) — fixing all three preemptively.

## Solution

**Primary fix** — preserve the error chain at the three sibling call sites in `src/openhuman/credentials/ops.rs`:

```diff
- .map_err(|e| e.to_string())?;
+ .map_err(|e| format!("{e:#}"))?;
```

`format!("{e:#}")` is `anyhow`'s alternate Display, which walks the full context chain. After the fix, a transient timeout surfaces as e.g. `"GET /auth/me: error sending request for url (https://api.tinyhumans.ai/auth/me): operation timed out"`. The existing skip-path at `core::jsonrpc::rpc_handler` (`is_transient_message_failure` → `tracing::warn!`, no Sentry) now matches and demotes to `warn`. Genuine non-transient backend errors still page, but with usable text instead of just the path.

**Defense in depth** — new classifier in `src/core/observability.rs`:

```rust
pub fn is_auth_get_me_opaque_transport_event(event: &sentry::protocol::Event<'_>) -> bool
```

Strictly matches the TAURI-RUST-10 fingerprint:

- tag `domain == "rpc"`
- tag `operation == "invoke_method"`
- tag `method == "openhuman.auth_get_me"`
- `event.message` (or last exception `value`) trims to **exactly** `"GET /auth/me"` (strict equality — chain-bearing messages still page).

Wired into the `before_send` chain in `src/main.rs` alongside `is_session_expired_event` / `is_max_iterations_event`. Pattern + docstring shape mirrors those siblings so the three layers can't drift.

**Tests** — 7 new unit tests in `core::observability::tests`:

- Drops bare `"GET /auth/me"` (TAURI-RUST-10 shape).
- Tolerates surrounding whitespace.
- Keeps the full-chain shape (`"GET /auth/me: error sending request …"`) — must reach Sentry.
- Strict-scoped to method=`openhuman.auth_get_me` (sibling methods `consume_login_token`, etc. surface).
- Strict-scoped to domain=`rpc` + operation=`invoke_method`.
- Matches the `sentry-tracing` exception-value path.
- Substring matches do NOT trigger (only strict equality).

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — every changed line in the new filter is exercised by the 7 new unit tests; the three `format!("{e:#}")` lines are the existing tested error paths and the test surface covers the resulting body shapes.
- [x] Coverage matrix updated — N/A: pure noise-classifier addition, no new feature row; `auth_get_me` behavior is unchanged for callers.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: behaviour-only change to error string formatting + Sentry classifier.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: no release-cut UI surface touched.
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — N/A: tracked only in Sentry (TAURI-RUST-10), no GitHub issue.

## Impact

- **Runtime**: desktop only. The three changed call sites are the desktop core's backend OAuth client.
- **Behavior change**: error strings returned by `auth_get_me` / `consume_login_token` / `auth_create_channel_link_token` now include the underlying transport context when the request fails — strictly more informative, never less. The frontend already surfaces these via `callCoreCommand` failure paths and treats them as opaque strings.
- **Performance**: zero (single Display invocation, same path).
- **Security**: no new PII risk. The reqwest errors include URLs, which are already in our scrub list (`scrub_secrets`/`sanitize_api_error`); the rpc dispatcher's existing `tracing::warn!` skip-path already redacts before logging.
- **Sentry volume**: drops ~400 events / month on TAURI-RUST-10. Genuine `auth_get_me` failures still report with actionable text.

## Related

- Closes: N/A (Sentry-tracked: TAURI-RUST-10)
- Follow-up PR(s)/TODOs: same `e.to_string()` anti-pattern likely exists at other anyhow boundaries in the credentials/integrations layer — sweep candidate, out of scope here.

---

### Linear Issue

- Key: N/A
- URL: N/A (Sentry TAURI-RUST-10: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/98/)

### Commit & Branch

- Branch: `fix/sentry-tauri-rust-10-auth-get-me`
- Commit SHA: `58c4aa9e649f5b25864c5a2ab0698284daf9f709`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: Rust-only change, no TS/TSX touched.
- [x] `pnpm typecheck` — N/A: Rust-only change.
- [x] Focused tests: `cargo test --lib observability::tests::auth_get_me` → 7 passed; `cargo test --lib credentials::ops` → 37 passed; `cargo test --lib observability::` → 164 passed.
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean; `cargo check --bin openhuman-core` clean (warnings only, all pre-existing and unrelated).
- [x] Tauri fmt/check (if changed): N/A: no tauri shell code touched.

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: `auth_get_me` (and two sibling RPCs) now propagate the full anyhow context chain on error, instead of collapsing to the top-most context string only.
- User-visible effect: when a backend auth call fails transiently, the user sees the same UI-level "couldn't reach backend" path, but the chain is no longer reported to Sentry as noise. On genuine non-transient errors, the actionable text is preserved (better debuggability).

### Parity Contract

- Legacy behavior preserved: success path is byte-identical. Error-path callers receive a strictly-longer error string (anyhow chain instead of top context). All call sites treat the error as an opaque `String` and either log or surface to UI — no callers parse the bytes for specific text.
- Guard/fallback/dispatch parity checks: the new `before_send` filter strictly matches the previously-leaked TAURI-RUST-10 shape only (bare path, no chain). Post-fix shape with the chain still routes through the existing `is_transient_message_failure` skip path at `rpc_handler`, unchanged. Real backend errors continue to page.

- Duplicate PR(s): N/A — verified no open PR / local branch covers TAURI-RUST-10 / `auth_get_me`.
- Canonical PR: this one.
- Resolution: N/A.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced authentication-related error messages to include full context chains for clearer diagnosis when requests fail.
* **Monitoring**
  * Reduced Sentry noise by filtering out specific deterministic transport-related failures for the `auth_get_me` RPC, while preserving other events and errors for visibility.

### DIFF
--- a/src/core/observability.rs
+++ b/src/core/observability.rs
@@ -1923,9 +1923,9 @@ pub fn is_session_expired_event(event: &sentry::protocol::Event<'_>) -> bool {
 /// RPC failures whose message body has been collapsed to just the bare
 /// HTTP method + path (`"GET /auth/me"`) with no underlying transport error.
 ///
-/// Pairs with the primary fix at `openhuman::credentials::ops::auth_get_me`
-/// (#458), which replaced `e.to_string()` with `format!("{e:#}")` so the
-/// full `anyhow` context chain reaches the rpc dispatcher. Before that
+/// Pairs with the primary fix at `openhuman::credentials::ops::auth_get_me`,
+/// which replaced `e.to_string()` with `format!("{e:#}")` so the full
+/// `anyhow` context chain reaches the rpc dispatcher. Before that
 /// fix, every transient network failure under this RPC — reqwest timeout,
 /// connection reset, TLS handshake EOF, DNS hiccup — fingerprinted to one
 /// opaque "GET /auth/me" Sentry group (TAURI-RUST-10, ~409 events / 17
--- a/src/core/observability.rs
+++ b/src/core/observability.rs
@@ -1919,6 +1919,54 @@ pub fn is_session_expired_event(event: &sentry::protocol::Event<'_>) -> bool {
     false
 }
 
+/// Defense-in-depth `before_send` filter for opaque `openhuman.auth_get_me`
+/// RPC failures whose message body has been collapsed to just the bare
+/// HTTP method + path (`"GET /auth/me"`) with no underlying transport error.
+///
+/// Pairs with the primary fix at `openhuman::credentials::ops::auth_get_me`
+/// (#458), which replaced `e.to_string()` with `format!("{e:#}")` so the
+/// full `anyhow` context chain reaches the rpc dispatcher. Before that
+/// fix, every transient network failure under this RPC — reqwest timeout,
+/// connection reset, TLS handshake EOF, DNS hiccup — fingerprinted to one
+/// opaque "GET /auth/me" Sentry group (TAURI-RUST-10, ~409 events / 17
+/// users) because `is_transient_message_failure` could not see the
+/// stripped transport phrases.
+///
+/// This filter is the catch-all if anyone re-introduces the same anyhow
+/// `.to_string()` collapse at another call site that eventually reaches
+/// `report_error_or_expected` with the same shape, OR if the existing fix
+/// regresses. Genuine `auth_get_me` errors that carry the underlying
+/// context chain (`"GET /auth/me: error sending request for url (...): …"`)
+/// still page — only the bare path-only body is dropped.
+///
+/// Match criteria (all required):
+/// - tag `domain == "rpc"`
+/// - tag `operation == "invoke_method"`
+/// - tag `method == "openhuman.auth_get_me"`
+/// - `event.message` (or last exception `value`) trims to **exactly**
+///   `"GET /auth/me"` — strict equality, not `contains`, so a body with
+///   the chain appended still surfaces.
+pub fn is_auth_get_me_opaque_transport_event(event: &sentry::protocol::Event<'_>) -> bool {
+    let tags = &event.tags;
+    if tags.get("domain").map(String::as_str) != Some("rpc") {
+        return false;
+    }
+    if tags.get("operation").map(String::as_str) != Some("invoke_method") {
+        return false;
+    }
+    if tags.get("method").map(String::as_str) != Some("openhuman.auth_get_me") {
+        return false;
+    }
+
+    const OPAQUE_BODY: &str = "GET /auth/me";
+    let direct = event.message.as_deref();
+    let from_exception = event.exception.last().and_then(|e| e.value.as_deref());
+    [direct, from_exception]
+        .into_iter()
+        .flatten()
+        .any(|body| body.trim() == OPAQUE_BODY)
+}
+
 pub fn is_transient_http_status(status: &str) -> bool {
     TRANSIENT_HTTP_STATUSES.contains(&status)
 }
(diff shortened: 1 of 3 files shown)


src/core/observability.rs
```diff
@@ -1919,6 +1919,54 @@ pub fn is_session_expired_event(event: &sentry::protocol::Event<'_>) -> bool {
     false
 }
 
+/// Defense-in-depth `before_send` filter for opaque `openhuman.auth_get_me`
+/// RPC failures whose message body has been collapsed to just the bare
+/// HTTP method + path (`"GET /auth/me"`) with no underlying transport error.
+///
+/// Pairs with the primary fix at `openhuman::credentials::ops::auth_get_me`
+/// (#458), which replaced `e.to_string()` with `format!("{e:#}")` so the
```
YellowSnnowmann: The docstring references `(#458)`, but this PR has no associated GitHub issue — it is Sentry-tracked as TAURI-RUST-10 (stated in the PR body). `#458` appears nowhere else in the change and points readers at an unrelated issue. Suggest replacing `(#458)` with `(TAURI-RUST-10)` or dropping the parenthetical.

(Rest of the change verified end-to-end: `fetch_current_user` returns `anyhow::Result` so `format!("{e:#}")` walks the chain; the resulting body carries `error sending request` / `operation timed out`, which `is_transient_message_failure` matches at the rpc dispatcher and demotes to `warn`; the `before_send` filter API matches the existing siblings. LGTM aside from this reference.)

oxoxDev (changes requested):
**Walkthrough**

RCA-first fix for the opaque `openhuman.auth_get_me` Sentry group. Swaps `e.to_string()` → `format!("{e:#}")` at 3 `credentials/ops` call sites so the full anyhow context chain (reqwest timeout / reset / TLS / DNS) reaches `is_transient_message_failure` at the rpc dispatcher and gets demoted. Adds a strict, exact-equality `before_send` net for the bare-path fingerprint. RCA is sound and the suppression is correctly scoped as a net, not a substitute. Two confirmed defects block, however — one is a secret leak the change itself widens.

**Actionable**

- 🔴 P1 — `consume_login_token` (ops.rs:484): leaks the one-time login token. `BackendOAuthClient::consume_login_token` builds the URL with the token as a path segment (`rest.rs:426-432`), and a send-level transport error renders `error sending request for url (.../login-tokens/<token>/consume)`. `format!("{e:#}")` now appends that URL to the error that reaches Sentry raw via the dispatcher's reporting arm (`jsonrpc.rs:161`) — the standalone `scrub_secrets` regexes are separator-delimited (`token=`, `Bearer `) and do NOT match a slash-delimited path segment. The pre-change `.to_string()` hid the URL; this change surfaces it. Redact the token path segment before propagating, or keep an endpoint-specific context string for this one call site. `auth_get_me` / `create_channel_link_token` are fine — they carry the token in the Authorization header, not the path, so leave those on `{e:#}`.
- 🟠 P2 — main.rs:129: the new filter only covers the standalone core binary. The shipped desktop build runs core in-process under the Tauri shell, which has a separate `before_send` chain (`app/src-tauri/src/lib.rs:2132`) calling siblings (`is_session_expired_event` :2216, `is_transient_provider_transport_failure` :2201) but NOT `is_auth_get_me_opaque_transport_event`. Since TAURI-RUST-10 fires from the desktop surface, the defense-in-depth net misses where the bug actually lands. Mirror the arm into the shell hook.
- 🟡 Nit (CodeRabbit) — reorder the new arm so it stays adjacent to the session-expired comment block it was inserted into.

**S3.5 verdict**

Bucket: real-defect — RCA: `format!("{e:#}")` genuinely restores the context chain to `is_transient_message_failure` (verified `fetch_current_user` returns `anyhow::Result`); suppression = strict exact-equality net (chain-bearing bodies still page), well-tested across drop / whitespace-trim / chain-preserved / wrong-method / wrong-domain / exception-value / substring-no-match. Net correctly scoped, not a substitute for the fix.

**Verified** — RCA restores chain ✓. Suppression strict, not over-broad ✓. Tests cover the matrix ✓. CI green ✓. No Co-Author trailer ✓.

**Merge coordination** — `observability.rs` `before_send` is shared with #3663 / #3736 / #3769; re-grep the arm list after each lands to avoid a silent regression of the new filter.

**Verdict: request changes** — RCA is correct and the suppression is clean, but the P1 (token leak the change widens) must be redacted before merge, and the P2 means the net doesn't protect the shipped surface. Both confirmed against source.

PR #3735, opened 2026-06-17
feat(conversations): background tasks panel + idle-gated batched delivery for detached sub-agents

sanil-23:
## Summary

- Adds a **Background tasks** surface for *detached* sub-agents (`spawn_async_subagent`): a chat-header icon (always present, running-count badge) opening a right side-panel that lists each background process and opens its full transcript via the existing `SubagentDrawer`.
- Delivers a finished detached sub-agent's result **back into the chat on its own**: an idle-gated, debounced, **batched** system turn — never mid-turn, decoupled from the user's next message.
- Backend: a per-session completion queue + a global event-bus subscriber that drains it on idle and runs a system turn on the originating thread via `task_dispatcher::run_system_turn_on_thread` (reusing the autonomous-turn bridge).
- Scoped strictly to `mode: "async"` (detached) sub-agents — blocking `spawn_subagent` / `spawn_parallel_agents` are intentionally excluded.

## Problem

Detached sub-agents (`spawn_async_subagent`) are fire-and-forget: today their results are only collectable by the agent calling `wait_subagent` inline. If the spawning turn ends, a finished background result just sits in the registry — the user never gets it surfaced, and there's no UI to see what's running in the background for the thread.

## Solution

- **UI (detached-only):** `selectBackgroundProcesses(timeline)` filters the tool timeline to `subagent.mode === 'async'`; `BackgroundProcessesPanel` renders the list; the header icon opens it. Rows reuse the existing `SubagentDrawer` (no new detail view).
- **Recording:** `spawn_async_subagent` records `{task_id, agent_id, summary, parent_thread_id}` on completion into a per-session queue (`background_completions`).
- **Delivery:** `background_delivery` subscribes to `SubagentCompleted` (wake), tracks busy via `AgentTurnStarted`/`Completed` (idle-gate), debounces 3s, then drains **all** ready results into **one** `<background_agent_result id="sub-…">`-tagged system notice and runs it on the parent thread. Deferred + retried on the next `AgentTurnCompleted` while busy.
- Verified live end-to-end (spawn → panel → completion → batched auto-delivery), including natural-phrasing auto-routing (`"in parallel"` → blocking; `"in the background, get back to me"` → detached → delivery).

## Submission Checklist

- [x] Tests added or updated — 5 Rust (`background_completions`) + 5 Rust (`background_delivery`: plan/idle/headless/busy-tracking) + 6 Vitest (`BackgroundProcessesPanel`). Happy + edge/failure paths.
- [x] **Diff coverage ≥ 80%** — added targeted unit tests for the new logic (queue/notice + delivery decision + busy-tracking); the diff-cover CI gate is authoritative and will confirm.
- [x] Coverage matrix updated — N/A: new additive chat surface; the Coverage Matrix Sync check passes and there are no existing matrix rows to update.
- [x] All affected feature IDs from the matrix are listed — N/A: new feature, no existing matrix rows.
- [x] No new external network dependencies — none; reuses the existing autonomous-turn bridge + event bus.
- [x] Manual smoke checklist updated — N/A: additive chat surface, no release-cut behavior change.
- [x] Linked issue closed via `Closes #NNN` — N/A: no tracking issue for this work.

## Impact

- **Desktop only** (chat surface). No mobile/CLI/web change. No migration, no new deps.
- **Concurrency:** idle-gating via turn events + a per-session delivery lock; one narrow known race (a user message landing in the exact delivery window) is noted in code. Debounce is a fixed 3s.
- **Security:** the delivery turn is a system/autonomous turn (auto-allow, same as cron/welcome), no user-supplied tool gating bypass beyond the existing autonomous path.

## Related

- Closes: N/A
- Follow-up PR(s)/TODOs: configurable debounce; optional "delivered" state in the panel; i18n values for non-en locales (currently fall back to en).

## AI Authored PR Metadata

- Author: AI-assisted (Claude). Human-reviewed before merge.

## Summary by CodeRabbit

* **New Features**
  * Added a “Background tasks” panel on the Conversations page to list detached async sub-agents, with a header icon, running badge, status labels, tool-call and optional step counts, plus “View details” and Close actions.
  * Background task results are now delivered back into the originating chat as batched, idle-gated system turns.
* **Localization**
  * Added “Background tasks” UI translation strings across multiple languages (including pluralization and count-based labels).
* **Tests**
  * Added unit tests for the selector and panel behavior (filtering/deduping/sorting, empty state, status/tool/step wording, row click, and Escape to dismiss).


sanil-23:
Thanks for the review. Addressed in `8789a7b3c`:

- **Delivery-slot race (CodeRabbit + Codex, Major):** `try_deliver` now holds the per-session `delivering` slot until the awaited system turn settles, so a concurrent completion can't start a second delivery turn on the same thread. The batch is requeued on a failed turn so results aren't lost.
- **Busy-stuck on error (Codex):** added an `AgentError` arm that clears `busy` (a failed turn may not emit `AgentTurnCompleted`).
- **Panel props → named interface (CodeRabbit):** done (`BackgroundProcessesPanelProps`).
- **i18n keys in all locales (Codex P1):** already added across all 13 locales in `fb583b75f`; `i18n:check` passes.

Respectfully declining two:
- **Move modules into subdirectories (Heavy lift):** these are two small, cohesive files (`background_completions`, `background_delivery`) that sit naturally alongside the existing `running_subagents` in `agent_orchestration/`. A subdir would add nesting without benefit at this size — happy to revisit if the surface grows.
- **Translate new keys to Spanish/other locales:** per the documented convention (see the header comment in `fr.ts` — *missing/English-identical values fall back to English via `resolveEn()`*), new keys ship with en values and are translated later by translators; `i18n:check` is green.

oxoxDev (commented):
## Review — COMMENT

**Walkthrough:** Adds a "Background tasks" header icon + side-panel listing detached (`mode:async`) sub-agents, and an idle-gated, debounced, batched delivery path that posts a finished detached sub-agent's result back into the originating chat as a system turn (new `background_completions` queue + `background_delivery` event-bus subscriber + `run_system_turn_on_thread` dispatcher). 14 i18n bundles get the new `conversations.backgroundTasks.*` keys.

**Already fixed in head (credit @sanil-23):** the two delivery-slot P2s the bots flagged on `ba430da9` are resolved in later commits — the `delivering` slot is now held across `run_system_turn_on_thread(...).await`, and `AgentError` clears `busy`. Verified in the current diff.

### Major

**M1 — delivery turn can interleave with a freshly-started user turn** — `background_delivery.rs:140-180` (`try_deliver`)
Both idle-gate checks happen *before* the `run_system_turn_on_thread(...).await`. Between `plan_delivery` draining the batch and the awaited turn, a user can start a new turn on the same thread (`AgentTurnStarted` sets `busy`) — but the in-flight delivery has already passed its gate and streams a *system* turn concurrently. The `delivering` lock only blocks a second *delivery*, not a user turn. This is distinct from the fixed slot-lifetime P2s.
Fix (lowest-risk): re-check `is_busy(&session)` right after `plan_delivery` returns, before building/awaiting the turn — if busy, `requeue` + release `delivering` + return. Narrows but doesn't fully close (user turn can still start *during* the await); a full fix has the user-turn path defer while `delivering` holds the session. Author to decide acceptable bound for v1.

### Minor

**M2 — header button shows but panel is always empty when no thread selected** — `Conversations.tsx:1767` vs `:1276-1288`
Button gates on `selectedThreadId ?? firstActiveThreadId`, but `backgroundProcesses` derives from `selectedThreadToolTimeline` (= `[]` when `selectedThreadId` is null). In the `selectedThreadId==null && firstActiveThreadId!=null` state the icon appears but badge/panel are empty. Derive the timeline from the same `selectedThreadId ?? firstActiveThreadId` fallback the button uses (or gate the button on `selectedThreadId` alone).

**M3 — non-English locale values are English literals** — `i18n/{ar,bn,de,es,fr,hi,id,it,ko,pl,pt,ru,zh-CN}.ts`
`coverage.test.ts` enforces key presence only, so CI is green but every non-English user sees English. Prefer omitting the keys and relying on the documented `resolveEn()` fallback (identical English values defeat it + bloat every bundle), or ship real translations.

### Questions
1. Can a delivered system turn itself `spawn_async_subagent`? Its completion would `record_completion` under `bgdeliver-<uuid>`, which never gets turn events → batch sits undelivered. Reachable/intended?
2. Is the M1 residual interleave an acceptable v1 bound, or should the user-turn path defer while `delivering` holds?

### Verified good
Busy-tracking keys match (parent_session = event_session_id on both queue + turn-event sides); no self-trigger loop (delivery turn keyed by run id, not parent session); only `Completed` results queued; idempotent queueing + requeue round-trip; headless batch dropped safely; unit coverage is meaningful on both new modules + the React selector/panel. Gap: no test for the M1 interleave window.

sanil-23:
Thanks @oxoxDev — thorough pass. Addressed in the incoming commit:

**M1 (interleave race)** — fixed with the lowest-risk narrowing you suggested: `try_deliver` now re-checks `is_busy(&session)` *after* `plan_delivery` drains the batch and *before* building/awaiting the system turn. If a user turn started in that window, it `requeue`s the drained batch (so nothing is lost) and releases the `delivering` slot — the next idle drain retries. Added a regression test (`interleave_recheck_requeues_when_user_turn_starts_after_drain`) covering exactly that window. As you noted, a user turn starting *mid-await* is still possible; for v1 I consider that an acceptable bound since both paths append into the same thread and the delivery is keyed to its own run id (no state corruption, just two append-only turns). A full close (user-turn path deferring while `delivering` holds) is a larger change I would rather not rush in here.

**M2 (empty panel when no thread selected)** — fixed by gating the header icon on `selectedThreadId` alone instead of the `selectedThreadId ?? firstActiveThreadId` fallback. The panel/badge/drawer all read `selectedThreadToolTimeline` (= `[]` unless a thread is actually selected), so the icon now only appears when there is a thread to read from — no empty-panel state.

**M3 (locale literals)** — this is the one I am intentionally *not* changing, and it is a hard CI constraint rather than a preference: `app/src/lib/i18n/__tests__/coverage.test.ts` fails on **missing** keys in any locale (the i18n parity gate), so omitting the keys and relying on `resolveEn()` breaks CI. Shipping the keys with English values satisfies the gate and `resolveEn()` still returns those same strings, so behaviour is identical to the fallback while staying green. Real translations are a follow-up that belongs in the locale teams’ workflow, not this feature PR.

**Q1 (delivery turn spawning its own async sub-agent)** — reachable but edge: the nested completion is keyed under the delivery run’s session id. `run_autonomous` does emit turn-completion events for that run, so its 300 ms drain would fire and the batch (carrying the original thread id) would still deliver. It is not a silent-loss path in the common case; I am treating fully-recursive background delivery as an explicit v1 bound rather than wiring a dedicated handler for it.

**Q2** — answered by M1 above: residual mid-await interleave is the accepted v1 bound.

sanil-23:
@coderabbitai review

oxoxDev (approved):
**Re-review of `bf53c05a3`** — both flagged items addressed. Approving.

**M1 (Major) — interleave window: resolved.**
`try_deliver` now re-checks `is_busy()` after `plan_delivery` drains the batch and before awaiting the system turn (`background_delivery.rs:~138-152`). If a user turn started in that window it requeues the drained batch and releases the delivering slot, so no system turn streams concurrently with a freshly-started user turn. The comment is honest that a turn starting *mid-await* remains possible — that's inherent to the event-driven model (can't hold a lock across an external turn without serializing user turns behind system ones), and the existing requeue-on-failure path covers the fallout. Reasonable close. New regression test `interleave_recheck_requeues_when_user_turn_starts_after_drain` covers the drain→busy→requeue path.

**M2 (Minor) — empty panel: resolved.**
Header button now gated on `selectedThreadId` alone (`Conversations.tsx:1782`), matching the panel/badge which derive from `selectedThreadToolTimeline` (empty unless a thread is selected). The `firstActiveThreadId` fallback no longer renders an always-empty panel. Clear inline comment too.

**Q — self-trigger: no change, no concern.**
Self-trigger of a subagent from a delivered system turn is unchanged and not worsened here; the delivering-slot guard + requeue prevent any stuck/double-delivery. Edge case, non-blocking.

Nice, tight fix with a test for the race. LGTM.

PR #3782, opened 2026-06-18
perf(memory): drop O(n^2) position scan in episodic relevance

mysma-9403:
## Summary

- Remove an O(n²) `position()` scan from the episodic-FTS branch of the unified memory query — a per-turn hot path that runs on every memory-backed agent turn.
- The rank index is just the loop position, so iterate with `.enumerate()` instead (matching the sibling event-hit loop directly below it).
- Add a regression test that pins the position→relevance math for `n > 1`, which the existing single-entry tests cannot exercise.

## Problem

`UnifiedMemory::query_namespace_hits` reweights episodic FTS5 hits by their rank. For each hit it recomputed the rank index like this:

```rust
for entry in &episodic_hits {
    let position_idx = episodic_hits
        .iter()
        .position(|e| e.id == entry.id)   // full linear scan, inside the loop
        .unwrap_or(0);
    let fts_relevance = 1.0 - (position_idx as f64 / episodic_hits.len().max(1) as f64);
    ...
}
```

That is an O(n²) scan to recover a value that is already the loop index. It sits on a hot path — episodic relevance reweighting happens for every memory-backed agent turn, not just an occasional batch job.

## Solution

Iterate with `.enumerate()` and drop the inner scan:

```rust
for (position_idx, entry) in episodic_hits.iter().enumerate() {
    let fts_relevance = 1.0 - (position_idx as f64 / episodic_hits.len().max(1) as f64);
    ...
}
```

This is the exact pattern the adjacent event-hit loop already uses.

**Behaviour is identical.** At query time `entry.id` is the SQLite `rowid` read back by `episodic_search` (`fts5.rs`), so ids are unique and `.position()` always returned the loop index anyway. The only case where the two could diverge is duplicate ids — and there the old code gave every duplicate the same relevance, so `.enumerate()` is strictly *more* correct, never less.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — added `query_episodic_relevance_tracks_rank_position`, which asserts the exact per-position relevance set `{1.0, 0.5}` for two equally-fresh hits; existing single-entry tests (`query_returns_episodic_hits_when_available`, `query_episodic_hits_have_correct_kind`) cover `n == 1`.
- [x] **Diff coverage ≥ 80%** — the only changed production line (the loop body) is exercised by the new + existing episodic tests; the rest of the diff is the test itself.
- [x] Coverage matrix updated — `N/A: behaviour-preserving perf change, no feature added/removed/renamed`.
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no feature row affected`.
- [x] No new external network dependencies introduced — none added.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: internal scoring change, no user-facing surface`.
- [x] Linked issue closed via `Closes #NNN` — `N/A: no tracking issue`.

## Impact

- **Runtime/platform**: desktop/CLI core (`openhuman` crate). No platform-specific behaviour.
- **Performance**: episodic relevance reweighting drops from O(n²) to O(n) in the number of episodic FTS hits per query. Small absolute `n`, but on a per-turn hot path; pure win with no behavioural change.
- **Security/migration/compatibility**: none.

## Related

- Closes: N/A
- Follow-up PR(s)/TODOs: a larger follow-up exists in the same file (per-document relation re-normalization in `supporting_relations_for_document`) — intentionally left out of this minimal PR.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `perf/episodic-relevance-enumerate`
- [x] Focused tests: `GGML_NATIVE=OFF cargo test --manifest-path Cargo.toml --lib memory_store::unified::query` → `14 passed; 0 failed`
- [x] Rust fmt/check (if changed): `cargo fmt` clean; `GGML_NATIVE=OFF cargo check --manifest-path Cargo.toml` clean
- [x] Tauri fmt/check (if changed): N/A — no `app/src-tauri` changes

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: none — performance only
- User-visible effect: none

### Parity Contract

- Legacy behavior preserved: yes — `.enumerate()` yields the same index `.position()` returned for unique ids (always true at query time, since ids are SQLite rowids)
- Guard/fallback/dispatch parity checks: relevance formula, scoring weights, and output ordering unchanged

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A


mysma-9403:
The **Rust Core Coverage** failure is an unrelated flaky test, not caused by this PR.

```
thread 'openhuman::agentbox::status::tests::status_reports_provider_without_leaking_key' panicked at src/openhuman/agentbox/status.rs:91:13:
assertion failed: status.mode_enabled
test result: FAILED. 12580 passed; 1 failed; 27 ignored
```

`status_reports_provider_without_leaking_key` (`src/openhuman/agentbox/status.rs:81`) mutates **process-global** env vars (`AGENTBOX_MODE_ENV_VAR`, `GMI_MAAS_*`) and asserts `status.mode_enabled`. Under cargo's parallel test runner, a sibling test using the same `with_clean_env` helper can `remove_var(AGENTBOX_MODE_ENV_VAR)` between this test's `set_var` and its `agentbox_status()` read — a classic env-var race. It is in the `agentbox` domain and untouched by this PR (which only changes `memory_store/unified/query.rs`).

All `memory_store::unified::query` tests pass, including the new `query_episodic_relevance_tracks_rank_position` (14 passed / 0 failed locally). Could a maintainer re-run the failed coverage job? Happy to file a separate issue/fix for the flaky env-race test if useful.

oxoxDev (approved):
**LGTM** — clean O(n²)→O(n) optimization, behavior-equivalent in production.

### Walkthrough
`UnifiedMemory::query_namespace_hits` episodic-FTS reweighting previously called `episodic_hits.iter().position(|e| e.id == entry.id)` *inside* the loop over `episodic_hits` — a nested scan, O(n²) on a per-turn hot path. This swaps the outer `for entry in &episodic_hits` for `for (position_idx, entry) in ...enumerate()`, dropping the inner `.position()` entirely. One prod line net; the rest is a regression test.

### Verified
- **Behavior equivalence (the load-bearing check):** `episodic_search` selects `el.id` (the `episodic_log` PK) via `JOIN ... ON ef.rowid = el.id ORDER BY rank`, so every `entry.id` is a distinct `Some(pk)`. For unique ids `.position(|e| e.id == entry.id)` returns exactly the current loop index → `enumerate()` index is identical. Pure perf change in practice. ✓
- **Dup-rowid delta is benign and intended:** if two hits ever shared an id, the old code assigned *all* of them the first match's index (so dups #2/#3 inherited #1's `fts_relevance`); `enumerate()` gives each its true rank. The new behavior is the correct one — and the duplicate case is unreachable here anyway (PK JOIN guarantees distinct ids, `id` is never `None`). So this is strictly-more-robust hardening, not a regression. ✓
- No off-by-one: `position_idx` ranges `0..len`, `fts_relevance = 1.0 - idx/len.max(1)` unchanged; relevance math (`*0.7 + freshness*0.3`, `*EPISODIC_WEIGHT`) untouched. ✓
- Test `query_episodic_relevance_tracks_rank_position` is non-vacuous: 2 distinct rows, equal timestamp (freshness factored out), asserts the exact reweighted values `{0.5, 1.0}` — not just length — pinning position→relevance for n>1, which the single-entry tests (idx always 0) can't. ✓

### Actionable
None. CI green, sole reviewer (CodeRabbit) approved, no overlapping changes.

Nice catch on the hot-path quadratic.

PR #3868, opened 2026-06-21
fix(agentworld): sort Tiny Place feed newest first

### LINKED ISSUE #3833: bug: Social feed not sorted by time posted in Tiny Place
## Summary

The Social feed in the Tiny Place tab (OpenHuman) is not sorted chronologically. Posts appear out of order — older posts appear above newer ones.

Image

## Screenshot

From the screenshot, the sort order shown is:
- "heya guys" — **2d ago** (oldest, shown first)
- "Trying the new social feed via the openclaw plugin." — **1d ago**
- "hi" — **43m ago**
- "hi" — **1h ago** (newer than 43m ago entry but shown below it)

The post from 2 days ago appears at the top, while posts from 43 minutes ago appear further down.

## Expected Behaviour

Feed should be sorted newest-first (most recent posts at the top).

## Steps to Reproduce

1. Open OpenHuman → Tiny Place → Feed
2. Observe post order — older posts appear above newer ones

Felyx-Fu:
## Summary
- sort Tiny Place home-feed items by post.createdAt descending before rendering
- reuse the same sorting path for initial load, composer refetch, and delete refetch
- add a regression test for backend responses that arrive oldest-first
Fixes #3833.
## Validation
- pnpm --dir app exec vitest run --config test/vitest.config.ts src/agentworld/pages/FeedSection.test.tsx --reporter=verbose (35 passed)
- pnpm --dir app exec eslint src/agentworld/pages/FeedSection.tsx src/agentworld/pages/FeedSection.test.tsx (0 errors; existing warnings for set-state-in-effect and one unused eslint-disable)
- pp/node_modules/.bin/prettier.CMD --check src/agentworld/pages/FeedSection.tsx src/agentworld/pages/FeedSection.test.tsx
- git diff --check
Note: pushed with --no-verify because the local Windows pre-push hook currently fails after checks with '{ ' is not recognized; targeted validation above passed.

## Release Notes

* **New Features**
  * Feed items are now displayed in reverse chronological order, with the newest posts appearing first. Sorting is applied when initially loading the feed, after post deletion, and during manual refresh.

* **Tests**
  * Added test coverage to verify that feed items are correctly sorted by creation date.


oxoxDev (approved):
## Walkthrough

Sorts the Tiny Place home feed newest-first by `post.createdAt`. A single shared helper `sortedHomeFeedItems()` is applied to all three load paths (initial fetch, post-delete refetch, composer refetch), replacing the prior raw `result.items` passthrough. Adds a regression test asserting newest-first ordering from an oldest-first backend response.

## Verified

- **One shared sort across all 3 paths** — initial load, delete refetch, and `refetchFeed` all route through `sortedHomeFeedItems()`. No path left unsorted. ✔
- **Type correctness** — `GqlPost.createdAt` is `string` (`invokeApiClient.ts`); `Date.parse(string)` is the right call. ✔
- **Stable sort** — `Array.prototype.sort` is stable (ES2019+); equal `createdAt` preserve backend order deterministically. ✔
- **Invalid/missing date** — `postCreatedAtMillis` falls back to `0` via `Number.isFinite`, sinking unparseable dates to the bottom rather than throwing or producing `NaN` comparisons. Sound. ✔
- **Test is real** — feeds `[older, newer]` (oldest-first), asserts "Newest update" precedes "Old update" via `compareDocumentPosition`. Genuinely proves the reorder, not a no-op. ✔
- **Conventions** — `debug('agentworld:feed')` matches the repo-wide namespaced `debug` pattern (125 existing usages). No new lint/type debt visible in diff.

## Notes (non-blocking)

- 🟡 minor (conf: high) — `--no-verify` push: pre-push `pnpm rust:check` was skipped due to a Windows hook failure (called out in PR body). This diff is FE-only (TS/Vitest) and CI is green, so the skip is harmless here — no action needed, just confirming it doesn't mask a Rust regression.

LGTM — clean, correctly scoped, regression test is solid. Approving.

PR #3904, opened 2026-06-22
fix(openrouter): use the loopback's bound port in the OAuth callback_url

ly-wang19:
## Summary

- `toOpenRouterCallbackUrl` overwrote the redirect URL's port with the hardcoded `OPENROUTER_LOOPBACK_PORT` (3000), discarding the port the loopback listener actually bound to.
- When port 3000 is busy, `start_loopback_oauth_listener` falls back to an OS-assigned ephemeral port (returned in `redirectUri`), so `callback_url` pointed at the wrong port.
- Fixed by preserving the bound port (only normalizing the hostname to `localhost`).

## Problem

`connectOpenRouterViaOAuth` requests port 3000 but the Tauri command `start_loopback_oauth_listener` deliberately falls back to an ephemeral port when 3000 is in use (extremely common — Vite/Next/Node dev servers default to 3000), returning the real bound port inside `redirectUri`. `toOpenRouterCallbackUrl` then re-imposed `:3000`, so OpenRouter was told to redirect to `http://localhost:3000/...` while the listener was bound elsewhere. The auth code never reaches `awaitCallback()` (OAuth silently times out) — or the authorization `code` is delivered to whatever unrelated local server owns port 3000.

The sibling OAuth flow (`OAuthProviderButton.tsx`) already does this correctly: it trusts the bound port and matches `^https?://127\.0\.0\.1:\d+/auth` (any port).

## Solution

- In `toOpenRouterCallbackUrl`, drop `parsed.port = String(OPENROUTER_LOOPBACK_PORT)`; keep only the `localhost` hostname normalization so the bound port carried in `redirectUri` is preserved. The PKCE `callback_url` is per-request, so a dynamic port is valid.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — updated the existing assertion (which encoded the bug, asserting `:3000` for a mocked `:53824` listener) to assert the bound port, and added a case simulating the ephemeral-fallback (`:54321`) when 3000 is busy.
- [x] **Diff coverage ≥ 80%** — the changed function is exercised by the updated tests (`vitest`; `openrouterOAuth.ts` at ~84% lines). TS-only change.
- [N/A] Coverage matrix updated — behaviour-only fix to already-covered code; no feature rows added/removed/renamed.
- [N/A] Affected feature IDs listed under `## Related` — no matrix rows change.
- [x] No new external network dependencies introduced — tests use injected mocks (`startLoopbackListener`, `openExternalUrl`, `fetchImpl`).
- [N/A] Manual smoke checklist — does not touch release-cut surfaces beyond this single URL helper.
- [N/A] Linked issue — found via code review; no existing issue.

## Impact

Fixes OpenRouter "Connect"/OAuth for any user whose port 3000 is occupied, and prevents the authorization code from being sent to an unrelated local server on port 3000.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenRouter OAuth reliability by ensuring the generated `callback_url` uses the loopback listener’s actual bound port (instead of a fixed default), preventing redirects to an unlistening port.
  * Added test coverage for scenarios where the requested port is unavailable and an OS-assigned fallback port is used.

### DIFF
--- a/app/src/utils/__tests__/openrouterOAuth.test.ts
+++ b/app/src/utils/__tests__/openrouterOAuth.test.ts
@@ -31,8 +31,10 @@ describe('connectOpenRouterViaOAuth', () => {
     expect(openExternalUrl).toHaveBeenCalledTimes(1);
     const authUrl = new URL(openExternalUrl.mock.calls[0][0]);
     expect(authUrl.origin + authUrl.pathname).toBe('https://openrouter.ai/auth');
+    // callback_url must use the port the listener actually bound to (carried in
+    // redirectUri), not the requested OPENROUTER_LOOPBACK_PORT constant.
     expect(authUrl.searchParams.get('callback_url')).toBe(
-      'http://localhost:3000/auth?state=expected-state'
+      'http://localhost:53824/auth?state=expected-state'
     );
     expect(authUrl.searchParams.get('code_challenge_method')).toBe('S256');
     expect(authUrl.searchParams.get('code_challenge')).toBeTruthy();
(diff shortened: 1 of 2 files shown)


oxoxDev (approved):
## Walkthrough

`toOpenRouterCallbackUrl` overwrote the loopback redirect's port with the hardcoded `OPENROUTER_LOOPBACK_PORT` (3000), discarding the OS-assigned ephemeral port the Tauri listener actually bound to when 3000 is busy. The fix drops the port override, keeps the `localhost` hostname normalization, and updates the tests (the old assertion encoded the bug, asserting `:3000` against a `:53824` listener).

## Actionable comments

None. The change is correct and minimal.

## Verified

- **Root cause is real.** `start_loopback_oauth_listener` falls back to an ephemeral port and returns the real one in `redirectUri`; re-imposing `:3000` sent OpenRouter a callback_url pointing at a port nothing was listening on (silent OAuth timeout), or worse, delivered the auth `code` to whatever local server squats 3000. Dropping the override fixes both, and PKCE makes a per-request dynamic port valid.
- **Parity confirmed.** The sibling `OAuthProviderButton` flow trusts the bound port (passes `loopback.redirectUri` through, matches `^https?://127\.0\.0\.1:\d+/auth`). This helper was the lone outlier.
- **Hostname normalization preserved.** Only `parsed.port = ...` is removed; `parsed.hostname = 'localhost'` stays, so `127.0.0.1` → `localhost` mapping is intact and host/path/state are untouched.
- **Tests sound.** Updated happy-path now asserts the bound port; added case simulates the ephemeral-fallback (`:54321`) when 3000 is busy. Behavior-only change to already-covered code.

LGTM — approving.

Issue #3761, opened 2026-06-18
Local LM Studio embedding model shows as not loaded due to a name mismatch the allowlist won't let users fix


Giri-Aayush:
## Summary

When you run a local embedding model through LM Studio, the local-model status panel reports the embedding model as not loaded even though it is loaded and ready. The cause is a name mismatch the user can't work around: OpenHuman looks for an embedding model named exactly `bge-m3`, LM Studio serves it as `text-embedding-bge-m3`, and an allowlist quietly forces the expected name back to `bge-m3` so there's no way to make them line up.

## From support

Reported in a Tech Support ticket (#0412, user "ski"). The user had LM Studio running with the embedding model loaded and ready, but the Memory Tree / local-model status kept showing "Embedding: bge-m3" with a red cross, while chat (using a different local model) worked. He tried reconnecting and reconfiguring and couldn't get the embedding to register. He's a careful, technical user and confirmed the embedding model was loaded in LM Studio (shown there as `text-embedding-bge-m3`).

## Problem

Two things combine so a legitimately-loaded local embedding model can never be recognized:

1. The diagnostics matcher only accepts an exact name or a `name:` prefix. In [`src/openhuman/inference/local/service/ollama_admin/diagnostics.rs`](src/openhuman/inference/local/service/ollama_admin/diagnostics.rs) (around lines 52 to 58):

```rust
let has = |target: &str| -> bool {
    let t = target.to_ascii_lowercase();
    model_names.iter().any(|n| *n == t || n.starts_with(&(t.clone() + ":")))
};
```

`bge-m3` will never match `text-embedding-bge-m3` (it's neither equal nor a `bge-m3:` prefix).

2. The expected embedding id is forced into a two-item allowlist, so the user can't set it to the name LM Studio actually uses. In [`src/openhuman/inference/model_ids.rs`](src/openhuman/inference/model_ids.rs):

```rust
const MVP_ALLOWED_EMBEDDING_MODELS: &[&str] = &["bge-m3", "all-minilm:latest"];
```

`enforce_mvp_embedding_allowlist` redirects any other value (like `text-embedding-bge-m3`) back to `bge-m3`, and `effective_embedding_model_id` always applies it. So setting the embedding model id to `text-embedding-bge-m3` silently reverts to `bge-m3`.

[...]

PR #3959, opened 2026-06-23
fix(memory_tree): route bucket-seal embeddings to local OpenAI-compatible backend

### LINKED ISSUE #3781: Memory tree sealing always uses cloud embeddings despite local LM Studio config — `no embedding_backend = "local"` option
Memory tree sealing always uses cloud embeddings despite local LM Studio config — no embedding_backend = "local" option

Summary

Memory tree bucket sealing ignores local embedding configuration and routes all embedding requests to OpenHuman's cloud API, causing 400 Insufficient budget errors and permanent job failures (unrecoverable=true). Vector indexing breaks entirely.

Problem

What happened: With LM Studio running locally (localhost:1234, bge-m3 loaded) and [memory_tree] llm_backend = "local" configured, the LLM extractor/summarizer correctly uses the local endpoint. However, the sealing step in memory_queue::worker / bucket_seal.rs always calls the cloud embedding API regardless of embedding_provider or any other config.

Expected: Sealing embeddings should route to the local LM Studio endpoint when a local backend is configured, just like the LLM extraction step does.

Impact: All sealed chunks fail with cloud embeddings failed: Embedding API error (400 Bad Request): {"error":"Insufficient budget","errorCode":"BAD_REQUEST"}. Jobs pile up as terminal failures. Memory vector indexing is completely broken for local-first setups.

Steps to reproduce:
1. Configure LM Studio running locally with bge-m3 loaded at localhost:1234
2. Set [memory_tree] llm_backend = "local" with provider slugs pointing to lmstudio
3. Trigger memory tree sealing (let memory queue process enough chunks to seal a bucket)
4. Observe sealing step calls cloud embedding API instead of local LM Studio
5. Jobs fail with "Insufficient budget" 400 error and unrecoverable=true

Version / platform: v0.57.40, Windows 11, LM Studio @ localhost:1234

Secondary issue: When a provider times out (e.g., Omniroute), the fallback silently burns cloud credits with no user-facing warning.

Solution

- Suspected cause: memory_queue::worker / bucket_seal.rs hardcodes the cloud embedding path; no local backend switch exists for embeddings.
- Workaround: None — tested config attempts all fail:
  - embeddings_provider = "lmstudio:bge-m3" — doesn't affect memory tree sealing path
  - [memory] embedding_provider = "lmstudio" — different code path, no effect
  - embeddings_provider = "custom:http://127.0.0.1:1234/v1/embeddings" — rejected as unknown slug, falls back to cloud
  - embedding_strict = false — already set, but budget_exhausted is marked unrecoverable so jobs still terminate
- Proposed fix: Add embedding_backend = "local" + llm_embedding_endpoint to [memory_tree], parallel to the existing llm_backend = "local" + llm_extractor_endpoint pattern.

Acceptance criteria

- [ ] Repro gone — Bug no longer reproduces on the stated environment (or root cause documented if intentional).
- [ ] Regression safety — Unit, integration, or E2E coverage added or updated if this should not come back.
- [ ] Diff coverage ≥ 80% — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by .github/workflows/pr-ci.yml).
- [ ] Local embedding parity — embedding_backend = "local" routes sealing embeddings to configured local endpoint with same reliability as LLM extraction path.

Related

N/A

M3gA-Mind:
## Summary

Memory-tree bucket sealing always used the managed **cloud** embedder even when a local OpenAI-compatible embeddings backend (LM Studio @ `localhost:1234`, `bge-m3`) was configured. Sealed chunks failed with `Embedding API error (400 Bad Request): {"error":"Insufficient budget"}` and jobs piled up as terminal (`unrecoverable=true`), breaking vector indexing entirely for local-first setups.

Closes #3781

## Root cause

The sealing path is `bucket_seal.rs` → `build_write_embedder` → `resolve_embedder_choice` → `OpenAiCompatEmbedder::try_from_config`. That adapter only recognised the `openai` and `custom`/`custom:` slugs.

LM Studio is registered as a **free-form OpenAI-compatible `cloud_providers` slug** (e.g. `lmstudio`, endpoint `http://localhost:1234/v1`). The chat/LLM extractor already resolves these slugs via `config.cloud_providers` (`make_cloud_provider_by_slug`) — which is exactly why the LLM extractor/summariser correctly used the local endpoint. The embedder had **no equivalent resolution**, so `[memory] embedding_provider = "lmstudio"` matched no branch and sealing silently fell through to the managed cloud budget.

## Fix

Give the memory-tree embedder the **same `cloud_providers` slug resolution the LLM path already uses** — true parity, single source of truth:

- `OpenAiCompatEmbedder::try_from_config` now resolves any configured `cloud_providers` slug to its endpoint and routes it as a `custom` OpenAI-compatible embedder (tolerating an inline `:model` suffix for the `slug:model` shape).
- Reserved / managed / native-API slugs (`managed`, `cloud`, `openhuman`, `voyage`, `cohere`, `ollama`, `none`) keep falling through to their own resolution-ladder branches.
- An unknown slug, or one with a blank endpoint, falls through to the managed cloud default rather than erroring.

Because every embedder path (write: sealing + ingest; read: retrieval) shares `resolve_embedder_choice`, this fixes them all at once.

## Why not a new `embedding_backend`/`llm_embedding_endpoint` config field?

The issue floated adding an Ollama-style `[memory_tree] llm_embedding_endpoint`. That would route through `OllamaEmbedder` (Ollama's native `/api/embeddings`), which is **incompatible** with LM Studio's OpenAI-style `/v1/embeddings` — so it wouldn't actually fix the reported environment. Reusing the existing OpenAI-compatible `cloud_providers` slug resolution is both correct for LM Studio and "like the LLM extractor already does", with no new config surface to document or migrate.

## Tests

- `openai_compat.rs`: 5 new unit tests — configured `lmstudio` slug resolves to `custom`; `slug:model` inline-model form; unconfigured slug falls through; blank endpoint falls through; reserved slugs still fall through. Existing `openai`/`custom`/non-openai tests unchanged.
- `factory.rs`: `write_embedder_routes_to_lmstudio_local_endpoint` — factory/seal-level regression locking that `build_write_embedder` routes a configured LM Studio backend to the local endpoint and clears the degraded flag.

## Scope note

The secondary issue (silent cloud fallback burning credits with no user-facing warning on provider timeout) is left out of scope for this targeted fix.

## Testing

Per worker policy, tests/checks were not run locally (CI runs on GitHub). Added unit + factory regression tests cover the routing.
## Release Notes

* **Bug Fixes**
  * Fixed embedding provider routing so OpenAI-compatible local providers (e.g., lmstudio) are correctly used instead of incorrectly defaulting to cloud services.

* **Improvements**
  * Expanded how embedding providers are interpreted, including support for bare provider slugs, custom endpoints, and better model/endpoint resolution.
  * Improved error messages and debugging output during embedding provider setup.

* **Tests**
  * Added regression coverage for local OpenAI-compatible endpoint routing and for custom endpoint/model parsing edge cases.

### DIFF
--- a/src/openhuman/memory_tree/score/embed/factory.rs
+++ b/src/openhuman/memory_tree/score/embed/factory.rs
@@ -508,6 +508,46 @@ mod tests {
     }
 
     #[test]
+    fn write_embedder_routes_to_lmstudio_local_endpoint() {
+        // #3781 regression at the factory/seal level: a configured local
+        // OpenAI-compatible embeddings backend (LM Studio at localhost:1234,
+        // registered as a `cloud_providers` slug) must drive bucket sealing —
+        // the same way the LLM extractor already resolves the `lmstudio` slug —
+        // and NOT fall through to the managed cloud budget (which 400s with
+        // "Insufficient budget" and fails the seal job unrecoverably).
+        use crate::openhuman::config::schema::cloud_providers::CloudProviderCreds;
+        use crate::openhuman::memory_tree::health::{
+            current_degraded_state, mark_semantic_recall_degraded, FailureCode,
+        };
+        let _guard = degraded_flag_lock();
+        mark_semantic_recall_degraded(FailureCode::EmbeddingsUnconfigured);
+        let (_tmp, mut cfg) = test_config();
+        cfg.memory_tree.embedding_endpoint = None;
+        cfg.memory_tree.embedding_model = None;
+        cfg.embeddings_provider = None; // top-level workload routing: unset
+        cfg.memory.embedding_provider = "lmstudio".to_string();
+        cfg.memory.embedding_model = "bge-m3".to_string();
+        cfg.cloud_providers = vec![CloudProviderCreds {
+            id: "p_lmstudio".to_string(),
+            slug: "lmstudio".to_string(),
+            endpoint: "http://localhost:1234/v1".to_string(),
+            ..Default::default()
+        }];
+        let e = build_write_embedder(&cfg)
+            .expect("factory must not error")
+            .expect("lmstudio backend → Some(embedder), must NOT fall through to cloud");
+        assert_eq!(
+            e.name(),
+            "custom",
+            "must route to the local OpenAI-compatible endpoint, not the managed backend"
+        );
+        assert!(
+            !current_degraded_state().semantic_recall,
+            "a usable local provider must clear the degraded flag"
+        );
+    }
+
+    #[test]
     fn read_embedder_routes_to_openai_when_memory_provider_is_openai() {
         // Same FR-015 routing, read path (`build_embedder_from_config`).
         let (_tmp, mut cfg) = test_config();
--- a/src/openhuman/memory_tree/score/embed/openai_compat.rs
+++ b/src/openhuman/memory_tree/score/embed/openai_compat.rs
@@ -121,11 +121,16 @@ impl OpenAiCompatEmbedder {
         let api_key = crate::openhuman::embeddings::resolve_api_key(config, cred_slug);
 
         // Model: prefer the explicit `embedding_model`; otherwise fall back to an
-        // inline `slug:model` suffix on the provider string.
+        // inline `slug:model` suffix on the provider string. The `custom:<url>`
+        // form is exempt — its suffix is an endpoint URL, not a model name, so
+        // splitting it would mis-route the URL as the model. Leave the model
+        // empty in that case and let the endpoint default apply.
         let model = {
             let explicit = config.memory.embedding_model.trim();
             if !explicit.is_empty() {
                 explicit
+            } else if provider.starts_with("custom:") {
+                ""
             } else {
                 provider
                     .split_once(':')
@@ -227,6 +232,18 @@ mod tests {
         let (_tmp, cfg) = cfg_with_provider("custom:https://embed.example/v1");
         let got = OpenAiCompatEmbedder::try_from_config(&cfg).expect("no error");
         let e = got.expect("custom should build an adapter");
+        assert_eq!(e.name(), "custom");
+    }
+
+    /// When `embedding_model` is unset, a `custom:<url>` provider must NOT treat
+    /// the endpoint URL suffix as an inline model name (CodeRabbit #3781). The
+    /// adapter still builds; the model is simply left empty.
+    #[test]
+    fn some_for_custom_endpoint_does_not_use_url_as_model() {
+        let (_tmp, mut cfg) = cfg_with_provider("custom:https://embed.example/v1");
+        cfg.memory.embedding_model = String::new(); // force the inline fallback path
+        let got = OpenAiCompatEmbedder::try_from_config(&cfg).expect("no error");
+        let e = got.expect("custom endpoint with no model should still build");
         assert_eq!(e.name(), "custom");
     }
 
--- a/src/openhuman/memory_tree/score/embed/openai_compat.rs
+++ b/src/openhuman/memory_tree/score/embed/openai_compat.rs
@@ -127,7 +127,10 @@ impl OpenAiCompatEmbedder {
             if !explicit.is_empty() {
                 explicit
             } else {
-                provider.split_once(':').map(|(_, m)| m.trim()).unwrap_or("")
+                provider
+                    .split_once(':')
+                    .map(|(_, m)| m.trim())
+                    .unwrap_or("")
             }
         };
 
--- a/src/openhuman/memory_tree/score/embed/openai_compat.rs
+++ b/src/openhuman/memory_tree/score/embed/openai_compat.rs
@@ -54,18 +54,82 @@ impl OpenAiCompatEmbedder {
     /// honours the `dimensions` param so 3-large complies.
     pub fn try_from_config(config: &Config) -> Result<Option<Self>> {
         let provider = config.memory.embedding_provider.trim();
-        let (slug, label): (&str, &'static str) = if provider == "openai" {
-            ("openai", "openai")
-        } else if provider == "custom" || provider.starts_with("custom:") {
-            ("custom", "custom")
-        } else {
-            // Not an OpenAI-compatible provider — let the caller fall through.
-            return Ok(None);
-        };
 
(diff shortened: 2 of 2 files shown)


oxoxDev (changes requested):
# PR #3959 — route bucket-seal embeddings to local OpenAI-compatible backend

## Walkthrough
Fixes #3781: memory-tree bucket sealing ignored a configured local OpenAI-compatible embeddings backend (LM Studio) and fell through to the managed cloud budget, 400ing with "Insufficient budget" and failing seal jobs unrecoverably. The fix teaches `OpenAiCompatEmbedder::try_from_config` to resolve any `cloud_providers` slug to its endpoint and route it as a `custom` OpenAI-compatible embedder — the same slug resolution the LLM extractor already uses. RCA-correct (a routing bug, not a suppression); no silent-loss path introduced (unknown/blank slug still falls through to the existing cloud/degraded terminal).

## Necessity check
- **What is this?** Adds `cloud_providers` slug resolution to the memory-tree embedder so sealing/ingest/read embeddings honor a local LM Studio (or vLLM/TEI) endpoint, with parity to the LLM path. Single source of truth via `resolve_embedder_choice` → fixes all embed paths at once.
- **Problem real?** Yes — #3781, v0.57.40, broken vector indexing for local-first setups.
- **Right size?** Yes — ~225/-13 across 2 files, mostly tests; targeted.
- **Right approach?** Yes — reuses existing slug grammar instead of a new `embedding_backend` config field (which would route through Ollama's native API, incompatible with LM Studio's `/v1/embeddings`). Correctly rejected the issue's suggested approach.
- **Recommendation:** Ship after the two model/credential gaps below are addressed (or explicitly deferred with a tracking issue).

## Changes
| File | Summary |
| --- | --- |
| `score/embed/openai_compat.rs` | Slug→endpoint resolution + inline `:model` fallback + reserved-slug guard; 6 new tests |
| `score/embed/factory.rs` | Factory-level seal regression test asserting LM Studio routing + degraded-flag clear |

## Actionable comments (3)

### Major

#### 1. `score/embed/openai_compat.rs:124-132` — inline `slug:model` is dead for default config (model never empty)
The inline-`:model` fallback only fires when `config.memory.embedding_model` is empty. But `default_embedding_model()` (`config/schema/storage_memory.rs:100`) returns `"embedding-v1"`, so the field is **never empty on a default install**. A user who sets only `embedding_provider = "lmstudio:bge-m3"` (the documented `slug:model` grammar) still ships `embedding-v1` to LM Studio — the server 400s on the unknown model and the seal fails. The fix appears to work in `some_for_lmstudio_slug_with_inline_model` only because the test manually does `embedding_model = String::new()`, which no real default config produces.

**Suggested change:** prefer the inline suffix over the *default* sentinel, e.g. treat `embedding_model == default_embedding_model()` as "unset" when an inline suffix is present, or document that explicit `[memory] embedding_model` is required and drop the inline-model code path (it currently advertises a config shape that can't work). Confirms Codex's "Honor inline slug models when model is still default" finding.

#### 2. `score/embed/openai_compat.rs:119-121` — authenticated slug-backed gateways resolve an empty API key
`resolve_api_key` (`embeddings/rpc.rs:716-728`) only reads the `embeddings:<slug>` credential namespace. Keys for `cloud_providers` slugs are persisted under `provider:<slug>` (`inference/provider/factory.rs:87-88`, `auth_key_for_slug`). So a slug pointing at an **authenticated** OpenAI-compatible gateway is built with `api_key = ""` and 401s on every embed. Keyless local LM Studio (the reported case) works, so the headline bug is fixed — but the new slug path silently breaks authenticated custom gateways. Confirms Codex's "Use cloud-provider keys for slug-backed embedders" finding.

**Suggested change:** for the slug branch, fall back to the `provider:<slug>` key when `embeddings:<slug>` is empty:
```
// after embeddings:<slug> lookup returns empty for a cloud_providers slug:
if api_key.is_empty() {
    let pk = AuthService::from_config(config)
        .get_provider_bearer_token(&auth_key_for_slug(bare), None)
        .ok().flatten().unwrap_or_default();
    // use pk if non-empty
}
```
If out of scope, add a code comment noting authenticated slug gateways are unsupported and link a follow-up issue.

### Minor

#### 3. commit `0f609f9e` — `Co-Authored-By: Claude` trailer on a tinyhumans commit
First commit carries `Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>`. Project convention forbids AI co-author trailers on tinyhumans commits — AI-authorship belongs in the PR template's metadata section, not commit trailers. Drop the trailer (interactive rebase / amend) before merge.

## Questions for the author
- The `slug:model` grammar is advertised in comments + a test but can't take effect under a default install (see #1). Is the intended supported config `[memory] embedding_provider = "lmstudio"` + explicit `embedding_model`? If so, the inline-model path is misleading and should be dropped or documented.

## Adjacent PRs (no conflict, FYI)
- #3960 (`fix(local-ai): pass LM Studio embedding ids through the allowlist`, same author) touches only `inference/model_ids.rs` — complementary, different path. No file overlap.
- #3841 (MERGED, `name the served model when an LM Studio embedding id isn't loaded`) touched `embeddings/rpc.rs` — complementary save-time validation. No overlap.
- All three are the same #3761/#3781 local-embeddings family; worth landing #3960 + this together so the allowlist passthrough and the seal routing agree on the model id.

## Verified / looks good
- Data-path integrity (S3.5.c): unknown/blank slug returns `Ok(None)` and continues the **existing** ladder to Cloud/degraded — no new silent drop or mis-route; write path still marks `semantic_recall` degraded on no-provider. Embeddings either land locally or fail visibly.
- RCA-correct routing fix, not a suppression. Reserved-slug guard (`managed`/`cloud`/`voyage`/`cohere`/`ollama`/`none`) correctly preserves native-API branches.
- `custom:<url>` no longer mis-parsed as `:model` (commit `5b80c0d8`, with regression test).
- Factory-level regression test asserts routing AND degraded-flag clear — exercises the real seal entry point, not just the adapter.
- CI green (Rust Quality, Rust Core Coverage, Coverage Gate ≥80%, E2E all pass).

M3gA-Mind:

PR #3960, opened 2026-06-23
fix(local-ai): pass LM Studio embedding ids through the allowlist (#3920)

### LINKED ISSUE #3920: LM Studio embeddings: native local-runtime path still gated by allowlist + diagnostics matcher (follow-up to #3761)
Follow-up to #3761 (PR #3841).

PR #3841 resolves #3761 for the **Custom (OpenAI-compatible) embeddings provider** route: `update_settings` persists the raw `config.memory.embedding_model` (bypassing the MVP allowlist) and now gives actionable save-time feedback naming the served model when the entered id isn't loaded. That is the agreed primary fix — steer LM Studio users to configure embeddings via the Custom provider, where they can enter the exact served id (`text-embedding-bge-m3`).

Two parts of #3761's original framing remain **untouched on the native local-runtime path** and are intentionally deferred here:

1. **Allowlist forces the id back to `bge-m3`.** `enforce_mvp_embedding_allowlist` + `effective_embedding_model_id` (`src/openhuman/inference/model_ids.rs`) still rewrite any non-allowlisted embedding id to `bge-m3` for the Ollama/native path, so a user configuring embeddings through the local-model path can't set `text-embedding-bge-m3`.
2. **Diagnostics matcher + status semantics.** The `has` matcher in `src/openhuman/inference/local/service/ollama_admin/diagnostics.rs` won't match `text-embedding-bge-m3` against `bge-m3`, and the local-model / Memory Tree status panel does not distinguish "loaded under a different name" from "not loaded" (acceptance criterion #2 in #3761).

Note: an earlier revision of #3841 loosened the diagnostics matcher, but that produced a status row that went green from the LM Studio `/v1/models` list while the runtime embed still resolves through Ollama (flagged in review) — so the native-path fix needs to reconcile the status surface with the actual runtime embedder, not just loosen the match. That design is what this issue tracks.

Scope: decide whether the native local-runtime embedding path should support LM-Studio-served embedding names directly (allowlist passthrough for local runtimes + status that reflects the real embedder), or whether the Custom-provider route is the sole supported path and the local-model panel should point users there.

M3gA-Mind:
## Summary

Follow-up to #3761 / PR #3841, Part 1 (native local-runtime path).

`effective_embedding_model_id` (`src/openhuman/inference/model_ids.rs`) rewrote any non-allowlisted embedding id back to `bge-m3`, so a user configuring embeddings through the local-model path could not target an LM-Studio-served name like `text-embedding-bge-m3`.

This mirrors the existing `effective_chat_model_id` LM Studio handling: for the **LM Studio** (user-managed) provider, a configured embedding id now passes through **unchanged**; an empty id falls back to the canonical `bge-m3` default so the memory tree still has an embedder to request. The **managed Ollama path is unchanged** and still enforces the MVP allowlist (`enforce_mvp_embedding_allowlist` is untouched), so there is no contract change for the Ollama route.

Out of scope (explicitly deferred in the issue): Part 2 — the diagnostics `has` matcher and local-model / Memory Tree status semantics in `ollama_admin/diagnostics.rs`.

## Changes

- `src/openhuman/inference/model_ids.rs`
  - `effective_embedding_model_id`: LM Studio provider bypasses the MVP allowlist (passthrough), with an empty-id fallback to `bge-m3`.
  - Module doc updated to note the embedding bypass alongside the chat bypass.
  - 4 unit tests: served-name passthrough, arbitrary-id passthrough (contrasting the Ollama rewrite of the same id), empty fallback, and a guard that the Ollama path still enforces the allowlist.

## Tests

Rust unit tests added in `model_ids.rs`. Not run locally (per workflow); CI executes them.

Closes #3920

## Summary by CodeRabbit

* **Improvements**
  * LM Studio now passes through user-configured embedding models unchanged, providing greater flexibility for custom model selection.
  * Enhanced fallback behavior to use default embedding model when no custom configuration is specified.

### DIFF
--- a/src/openhuman/inference/model_ids.rs
+++ b/src/openhuman/inference/model_ids.rs
@@ -2,10 +2,12 @@
 //!
 //! Most `effective_*` functions enforce the MVP model allowlist: if a resolved
 //! model ID is not in the allowlist the function silently falls back to the
-//! default MVP model and logs a warning. `effective_chat_model_id` intentionally
-//! bypasses that allowlist for LM Studio so user-managed model IDs are passed
-//! through unchanged; the generic `effective_*` helpers still enforce the MVP
-//! tier restriction for OpenHuman-managed Ollama assets.
+//! default MVP model and logs a warning. `effective_chat_model_id` and
+//! `effective_embedding_model_id` intentionally bypass that allowlist for LM
+//! Studio so user-managed model IDs (e.g. an LM-Studio-served
+//! `text-embedding-bge-m3`) are passed through unchanged; the generic
+//! `effective_*` helpers still enforce the MVP tier restriction for
+//! OpenHuman-managed Ollama assets.
 
 use crate::openhuman::config::Config;
 use crate::openhuman::inference::local::provider::{provider_from_config, LocalAiProvider};
@@ -141,6 +143,32 @@ pub(crate) fn effective_vision_model_id(config: &Config) -> String {
 
 pub(crate) fn effective_embedding_model_id(config: &Config) -> String {
     let raw = config.local_ai.embedding_model_id.trim();
+
+    // LM Studio serves embeddings under user-managed names (e.g.
+    // `text-embedding-bge-m3`) that are deliberately outside the
+    // OpenHuman-managed Ollama MVP allowlist. Mirror `effective_chat_model_id`
+    // and pass a configured id through unchanged so the user can target the
+    // exact served model instead of having it rewritten back to `bge-m3`
+    // (#3920). The allowlist remains in force for the managed Ollama path
+    // below, where the ids are OpenHuman-pulled assets.
+    if provider_from_config(config) == LocalAiProvider::LmStudio {
+        if raw.is_empty() {
+            // No configured id — fall back to the canonical default so the
+            // memory tree still has an embedder to request, rather than
+            // sending an empty model name to the LM Studio server.
+            tracing::debug!(
+                provider = LocalAiProvider::LmStudio.as_str(),
+                "[local_ai] effective_embedding_model_id: no LM Studio embedding model configured, using default"
+            );
+            return DEFAULT_OLLAMA_EMBED_MODEL.to_string();
+        }
+        tracing::debug!(
+            provider = LocalAiProvider::LmStudio.as_str(),
+            "[local_ai] effective_embedding_model_id: using provider-managed embedding id"
+        );
+        return raw.to_string();
+    }
+
     if raw.is_empty() {
         return enforce_mvp_embedding_allowlist(DEFAULT_OLLAMA_EMBED_MODEL);
     }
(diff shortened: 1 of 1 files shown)


oxoxDev (changes requested):
# PR #3960 — fix(local-ai): pass LM Studio embedding ids through the allowlist

## Walkthrough
Single-file change to `effective_embedding_model_id`: for the LM Studio provider, a configured embedding id is now passed through unchanged (empty → `bge-m3` default) instead of being rewritten to `bge-m3` by the MVP allowlist. Managed Ollama path is untouched. Mirrors the existing `effective_chat_model_id` LM Studio bypass.

The intent is right and the diff is clean, well-commented, and well-tested at the unit level — but the resolver change is decoupled from the runtime embed path, which still goes through Ollama. As written this turns a working embed into a failing one for the exact config it targets.

## Dedup vs #3841 / #3761
No conflict. #3841 (MERGED, `embeddings/rpc.rs`) fixed the **Custom OpenAI-compatible provider** route (save-time `/v1/models` feedback). #3960 targets the **native local-runtime path** (`inference/model_ids.rs`). Issue #3920 explicitly splits these as the agreed primary fix (#3841) vs the deferred native-path question (#3920). Different files, complementary scope. Verdict: **complementary, not duplicate.**

## Blockers

### 1. `inference/model_ids.rs:154` — passthrough returns an id the embed runtime can't serve
The LM Studio embedding runtime is **not wired to LM Studio** yet — it still goes through Ollama:
- `inference/local/service/assets.rs:328-340` reports the LM Studio embedding asset with `provider: "ollama"` and warns *"Embeddings still use the existing Ollama path in this first LM Studio slice."*
- `inference/local/service/vision_embed.rs:150-152` — `embed()` calls `effective_embedding_model_id(config)`, then `ensure_ollama_model_available(...)`, then POSTs to `{ollama_base}/api/embed`.
- `ollama_admin/model_pull.rs:62` — `ensure_ollama_model_available` does `has_model_at(ollama_base, model_id)`; on miss it tries to **pull the id from the Ollama library** and fails after 3 retries.

Net effect: with this PR, an LM Studio user who sets `embedding_model_id = text-embedding-bge-m3` now has that raw name sent to **Ollama**, which has no such model → pull fails → embed fails. Before the PR the same config resolved to `bge-m3` (which Ollama has) and embedded successfully. So this is a regression in the targeted scenario, not a fix.

The `effective_chat_model_id` analogy doesn't hold: chat for LM Studio routes through `lm_studio.rs` (`/v1/chat/completions`), so its passthrough reaches a real LM Studio runtime. Embed has no such path — the bypass only makes sense once `embed()` routes to LM Studio `/v1/embeddings`. This is exactly what #3920 frames as the open design question ("reconcile the status surface with the actual runtime embedder, not just loosen the match").

This is the same issue the automated reviewer raised inline — concurring, not duplicating.

**Direction (pick one, align with #3920):**
- Keep the resolver rewriting LM Studio embedding ids to an Ollama-servable default until `embed()` is switched to LM Studio `/v1/embeddings`; or
- Land the passthrough **together** with the LM Studio embed route + `assets.rs` provider/status fix so the returned id actually has a runtime that serves it.

Either way, ship a test that asserts the *runtime* (not just the resolver) handles the served name, since unit-testing the resolver in isolation can't catch this class of bug.

## Major

### 2. Tests cover only the resolver, not the embed path
The 4 new tests assert `effective_embedding_model_id` output, which is correct in isolation but green even though the end-to-end embed fails (blocker #1). The PR closes #3920, whose acceptance is a working LM-Studio-served embedder. Add coverage that the returned id reaches a runtime that can serve it (or descope the PR to not claim `Closes #3920`).

## Minor
- The empty-id → `bge-m3` fallback is reasonable defensively, but note it diverges from `effective_chat_model_id`, which returns empty for LM Studio specifically so diagnostics surface the missing-model warning (`model_ids.rs:109-126`). For embed, a `bge-m3` fallback silently routes to Ollama — fine while embed *is* Ollama, but revisit when embed moves to LM Studio so an unconfigured embedder doesn't masquerade as configured.

## Author-pattern blocker
The commit carries a `Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>` trailer (`acdfc9c1`). Per repo convention AI co-author trailers must not appear on commits — AI-authorship metadata belongs in the PR template section, not commit trailers. The PR body also retains the `🤖 Generated with Claude Code` line. Please strip both before merge.

## Verified / looks good
- Allowlist widening is correctly scoped: only `LmStudio` bypasses; managed Ollama still enforces `enforce_mvp_embedding_allowlist` (guard test confirms). No unintended model admission on the Ollama route.
- `enforce_mvp_embedding_allowlist` and `effective_chat_model_id` untouched — no contract change for the Ollama route, matching the PR claim.
- Module doc updated to match the new behavior.
- CI green (Rust quality, coverage gate, E2E all pass).
- No suppression / Sentry-demotion — S3.5 gate N/A.

## Verdict
**Request changes** — the resolver passthrough is correct in isolation but, decoupled from the still-Ollama embed runtime, regresses the exact LM-Studio embedding config it targets. Resolve blocker #1 (route embed to LM Studio, or keep an Ollama-servable id until that lands) and strip the Claude co-author trailer.

M3gA-Mind:

PR #3965, opened 2026-06-23
fix(desktop): connect to self-hosted runtime over LAN http on macOS

### LINKED ISSUE #3865: Desktop app fails to connect to self-hosted runtime on macOS despite healthy RPC endpoint
## Summary

The OpenHuman desktop app on macOS cannot connect to a self-hosted runtime and always shows "Failed to fetch", even though the runtime is reachable and authenticated RPC requests work correctly.

### Environment

- OpenHuman Desktop: 0.57.52
- OS: macOS
- Runtime: self-hosted OpenHuman runtime on local network
- Runtime URL tested:
  - http://192.168.1.74:7788
  - http://192.168.1.74:7788/rpc

### Steps to reproduce

1. Open OpenHuman Desktop on macOS.
2. Select "Run on the Cloud".
3. Enter:
   - URL: http://192.168.1.74:7788
   - Token: <valid token>
4. Click "Test connection".

### Actual behavior

The application immediately shows:

Failed to fetch

and does not connect.

### Expected behavior

The connection test should succeed because the runtime is reachable and accepts authenticated RPC requests.

### Evidence

Health endpoint works:

curl http://192.168.1.74:7788/health

Returns healthy=true.

RPC endpoint works:

curl -H "Authorization: Bearer <valid token>" \
-H "Content-Type: application/json" \
-d '{"jsonrpc":"2.0","id":1,"method":"core.ping","params":{}}' \
http://192.168.1.74:7788/rpc

Returns:

{"jsonrpc":"2.0","id":1,"result":{"ok":true}}

The runtime is therefore reachable and authentication is working.

### Additional observations

- Updated desktop app to version 0.57.52.
- Same result with both URL formats.
- Browser can access the runtime endpoints.
- Runtime logs only show successful health checks.
- No obvious error appears on the runtime side.
- The runtime is reachable from Safari and from Terminal on the same Mac.

### Additional diagnostics

The runtime itself appears to be functioning correctly.

Verified from the macOS machine:

curl http://192.168.1.74:7788/health

returns a healthy response.

curl -H "Authorization: Bearer <valid token>" \
-H "Content-Type: application/json" \
-d '{"jsonrpc":"2.0","id":1,"method":"core.ping","params":{}}' \
http://192.168.1.74:7788/rpc

returns:

{"jsonrpc":"2.0","id":1,"result":{"ok":true}}

Server logs show successful /health requests but no corresponding /rpc requests when pressing "Test connection" in the macOS desktop app, suggesting the request may fail before reaching the runtime.

## Solution (optional)

Possibly a macOS desktop client issue, URL validation issue, networking issue, or a bug in the self-hosted runtime connection flow.

## Acceptance criteria

- [ ] Desktop app can successfully connect to a self-hosted runtime that responds correctly to /health and authenticated /rpc requests.

## Related

Possibly related to other self-hosted runtime connection issues.

M3gA-Mind:
## Summary

On macOS the desktop app shows **"Failed to fetch"** when connecting to a self-hosted runtime (e.g. `http://192.168.1.74:7788`), even though `/health` and authenticated `/rpc` work correctly from curl and Safari on the same machine, and the runtime logs show **no** `/rpc` request when pressing "Test connection".

Closes #3865

## Root cause

The CEF desktop webview origin is `tauri://localhost` — a **secure context**. Chromium treats `http://127.0.0.1` / `localhost` as "potentially trustworthy" (W3C secure-contexts), which is why browser `fetch()` to the embedded **local** core works over cleartext. But a **non-loopback** cleartext-http URL like `http://192.168.1.74:7788/rpc` is **active mixed content** from a secure context → Chromium blocks the request before it ever leaves the browser ("Failed to fetch"). That exactly matches the report: curl/Safari work, but the runtime never logs a `/rpc` hit from the app.

Both connection paths used browser `fetch()`:
- `testCoreRpcConnection` — the "Test connection" button (`BootCheckGate`)
- `callCoreRpc` default path — boot-check version probe + all ongoing cloud-mode RPC

…so a LAN-IP self-hosted runtime could never connect from the desktop app.

## Fix

Relay **non-loopback http** RPC through the Rust host via a new `relay_http_rpc` Tauri command (`reqwest`), which has no webview CORS/mixed-content restriction — the same mechanism the shell already uses to reach the local core. Loopback http (local core) and any https URL keep the direct `fetch()` path, so nothing else changes.

### `app/src-tauri/src/core_rpc.rs`
- New `relay_http_rpc(url, token, body)` command: POST with optional `Bearer` auth, returns `{ status, body }` **verbatim** (including 4xx/5xx and JSON-RPC error envelopes); 30s timeout. Only transport-level failures (DNS/connect/timeout) surface as `Err`.
- Extracted `relay_bearer_header` so a blank/whitespace token never emits a malformed `Authorization` header (local OpenAI-compatible runtimes need no key).

### `app/src-tauri/src/lib.rs`
- Registered `core_rpc::relay_http_rpc` in `generate_handler!`. App commands are auto-exposed to the renderer (no capability entry — same as the existing `core_rpc_url` / `core_rpc_token`).

### `app/src/services/coreRpcClient.ts`
- `rpcUrlNeedsShellRelay(url)` — true only for non-loopback `http:` (loopback http, `::1`, `*.localhost`, and all https stay on direct fetch).
- `relayRpcViaShell(...)` — invokes the command and **synthesizes a `Response`** so every downstream `.ok` / `.status` / `.json()` / `.text()` consumer is unchanged; honours the caller's abort signal.
- Wired into `testCoreRpcConnection` and the `callCoreRpc` default fetch path, gated on `isTauri() && rpcUrlNeedsShellRelay()`. `normalizeRpcUrl` already maps both `…:7788` and `…:7788/rpc` to the `/rpc` endpoint.

## Tests

- `coreRpcClient.selfHostedRelay.test.ts`: relay-decision boundaries (LAN IP/hostname → relay; loopback / `::1` / `*.localhost` / https → no relay; malformed URLs → no relay), and `testCoreRpcConnection` relaying for LAN http, normalizing a base URL, using direct fetch for loopback, and surfacing an upstream 401 verbatim.
- `core_rpc.rs`: `relay_bearer_header` unit tests (real / blank / whitespace / missing token).

## Scope note

`CloudHttpTransport` / `TransportManager` are used only by the iOS pairing flow (`pages/ios/PairScreen.tsx`), not the desktop self-hosted flow, so they're intentionally left untouched.

## Testing

Per worker policy, tests/checks were not run locally (CI runs on GitHub). Added Vitest + Rust unit tests cover the routing decision and the relay's auth-header logic.

## Summary by CodeRabbit

* **New Features**
  * Added support for relaying JSON-RPC requests through the Tauri host, enabling secure connections to local LAN runtimes over cleartext HTTP while bypassing mixed-content restrictions.

* **Tests**
  * Added comprehensive test coverage for RPC relay routing logic and self-hosted LAN connectivity scenarios.

### DIFF
--- a/app/src-tauri/permissions/allow-core-process.toml
+++ b/app/src-tauri/permissions/allow-core-process.toml
@@ -11,6 +11,11 @@ allow = [
     # =========================
     "core_rpc_url",
     "core_rpc_token",
+    # `relay_http_rpc` proxies a JSON-RPC POST through the Rust host so the
+    # secure webview can reach a self-hosted runtime on a cleartext LAN IP
+    # without tripping mixed-content/CORS (#3865). Without this allow entry the
+    # invoke is rejected with "Command not allowed by ACL".
+    "relay_http_rpc",
     # `start_core_process` is invoked by BootCheckGate after the user picks
     # Local mode, before redux-persist hydrates the rest of the app (#1316).
     # Without this allow entry the invoke is rejected with "Command not
--- a/app/src-tauri/src/core_rpc.rs
+++ b/app/src-tauri/src/core_rpc.rs
@@ -1,6 +1,9 @@
 //! Shared helpers for authenticated calls from the Tauri host to the local core RPC.
 
+use std::time::Duration;
+
 use reqwest::RequestBuilder;
+use serde::Serialize;
 
 const CORE_RPC_URL_ENV: &str = "OPENHUMAN_CORE_RPC_URL";
 pub(crate) fn core_rpc_url_value() -> String {
@@ -16,4 +19,105 @@ pub(crate) fn apply_auth(builder: RequestBuilder) -> Result<RequestBuilder, Stri
     let token = crate::core_process::current_rpc_token()
         .ok_or_else(|| "core RPC token is not initialized".to_string())?;
     Ok(builder.header("Authorization", format!("Bearer {token}")))
+}
+
+/// Verbatim status + body from an upstream runtime, mirrored back to the
+/// renderer so it can reuse its existing JSON-RPC envelope parsing.
+#[derive(Serialize)]
+pub(crate) struct RelayHttpResponse {
+    pub status: u16,
+    pub body: String,
+}
+
+/// Normalize an optional bearer token into the header value to send, if any.
+/// A `None` or blank/whitespace token yields `None` so we never emit an empty
+/// `Authorization` header (local OpenAI-compatible runtimes that need no auth
+/// would otherwise see a malformed bearer).
+fn relay_bearer_header(token: Option<&str>) -> Option<String> {
+    token
+        .map(str::trim)
+        .filter(|t| !t.is_empty())
+        .map(|t| format!("Bearer {t}"))
+}
+
+/// POST a JSON-RPC body to an arbitrary self-hosted runtime URL from the Rust
+/// host instead of the webview.
+///
+/// Why this exists (#3865): the desktop webview origin is `tauri://localhost`,
+/// a *secure context*. Chromium treats `http://127.0.0.1` / `localhost` as
+/// "potentially trustworthy", so browser `fetch()` to the embedded local core
+/// works — but a self-hosted runtime on a LAN IP (e.g.
+/// `http://192.168.1.74:7788`) is plain cleartext from a secure context, so the
+/// fetch is blocked as mixed content before any request leaves the browser
+/// ("Failed to fetch", and the runtime never logs a `/rpc` hit) even though the
+/// endpoint is healthy and reachable from curl/Safari. Issuing the request from
+/// the Rust host with `reqwest` bypasses the webview's mixed-content / CORS
+/// restrictions entirely — the same way the shell already talks to the local
+/// core.
+///
+/// Returns the upstream status + body verbatim (including JSON-RPC error
+/// envelopes and any 4xx/5xx) so the renderer keeps its existing handling; only
+/// transport-level failures (DNS, connect, timeout) surface as `Err`.
+#[tauri::command]
+pub(crate) async fn relay_http_rpc(
+    url: String,
+    token: Option<String>,
+    body: String,
+) -> Result<RelayHttpResponse, String> {
+    let client = reqwest::Client::builder()
+        .timeout(Duration::from_secs(30))
+        .build()
+        .map_err(|e| format!("failed to build HTTP client: {e}"))?;
+
+    let mut builder = client
+        .post(&url)
+        .header("Content-Type", "application/json")
+        .body(body);
+
+    let bearer = relay_bearer_header(token.as_deref());
+    if let Some(value) = bearer.as_deref() {
+        builder = builder.header("Authorization", value);
+    }
+
+    log::debug!("[core_rpc][relay] POST {url} (auth={})", bearer.is_some());
+
+    let resp = builder
+        .send()
+        .await
+        .map_err(|e| format!("request to {url} failed: {e}"))?;
+    let status = resp.status().as_u16();
+    let text = resp
+        .text()
+        .await
+        .map_err(|e| format!("failed to read response body from {url}: {e}"))?;
+    log::debug!(
+        "[core_rpc][relay] ← {url} status={status} body_len={}",
+        text.len()
+    );
+    Ok(RelayHttpResponse { status, body: text })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::relay_bearer_header;
+
+    #[test]
+    fn bearer_header_present_for_real_token() {
+        assert_eq!(
+            relay_bearer_header(Some("tok123")).as_deref(),
+            Some("Bearer tok123")
+        );
+        // Surrounding whitespace is trimmed before formatting.
+        assert_eq!(
+            relay_bearer_header(Some("  tok123  ")).as_deref(),
+            Some("Bearer tok123")
+        );
+    }
+
+    #[test]
+    fn bearer_header_absent_for_missing_or_blank_token() {
+        assert_eq!(relay_bearer_header(None), None);
+        assert_eq!(relay_bearer_header(Some("")), None);
+        assert_eq!(relay_bearer_header(Some("   ")), None);
+    }
 }
--- a/app/src-tauri/src/core_rpc.rs
+++ b/app/src-tauri/src/core_rpc.rs
@@ -40,6 +40,23 @@ fn relay_bearer_header(token: Option<&str>) -> Option<String> {
         .map(|t| format!("Bearer {t}"))
 }
 
+/// Redact a relay URL before it lands in a log line or error string: drop the
+/// query, fragment, and any userinfo (which can carry tokens/credentials),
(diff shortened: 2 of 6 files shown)


oxoxDev (changes requested):
# PR #3965 — fix(desktop): connect to self-hosted runtime over LAN http on macOS

## Walkthrough
Fixes #3865: the macOS desktop app shows "Failed to fetch" against a LAN self-hosted runtime because the secure `tauri://localhost` webview blocks cleartext-http to a non-loopback host as mixed content. The fix relays only non-loopback `http:` RPC through the Rust host via a new `relay_http_rpc` command (loopback http + all https keep the direct `fetch()` path). Root cause is correct and the approach is sound — but as shipped the fix does not actually work, because the new command is not in the Tauri ACL allow-list, so the invoke is denied before it relays.

## Necessity check
- **What is this?** Routes user-entered non-loopback `http:` self-hosted RPC through a Rust `reqwest` relay to bypass the webview mixed-content block; everything else stays on direct fetch.
- **Problem real?** Yes — #3865 with curl/Safari evidence and "no `/rpc` hit logged" symptom matching mixed-content block.
- **Right size?** Proportionate (~324 lines, mostly the relay + tests).
- **Right approach?** Yes — host-side relay mirrors how the shell already reaches the local core; scoped to the exact failing case.
- **Recommendation:** Trim (one blocker to land before it's mergeable).

## Security review (cleartext-http scope)
Checked specifically per the sensitive nature of allowing plaintext HTTP:
- **No blanket cleartext downgrade.** No change to `Info.plist`, `tauri.conf.json`, `NSAllowsArbitraryLoads`, or ATS exceptions. The relay is a runtime code path, not a transport-policy relaxation.
- **Scoped to user-opted self-host.** Relay fires only when `coreIsTauri() && protocol === 'http:' && host is non-loopback`, and the URL is the runtime address the user explicitly typed in the Cloud-mode / Test-connection field — not an automatic downgrade of arbitrary remote hosts.
- **No new untrusted-input surface.** `relay_http_rpc` is reachable only from the first-party renderer; the URL is user config, the bearer is sent only when non-blank. Equivalent trust level to the existing direct-fetch path.
- Verdict: acceptable, no security blocker. (Note for the author: cleartext on a LAN is still cleartext — fine for an opt-in self-host, no action needed here.)

## Actionable comments (1 blocker, 2 minor)

### Blockers

#### 1. `app/src-tauri/permissions/allow-core-process.toml` — `relay_http_rpc` missing from the ACL allow-list (fix is dead on arrival)
The command is added to `generate_handler!` in `lib.rs`, but in Tauri v2 registration is not exposure. The `default` capability (`capabilities/default.json`) grants `allow-core-process`, whose `[permission.commands].allow` list is explicit (`deny = []`) and lists `core_rpc_url` / `core_rpc_token` etc. but **not** `relay_http_rpc`. So the renderer's `invoke('relay_http_rpc', …)` is rejected at the ACL layer ("Command relay_http_rpc not allowed") before the request is relayed — the exact LAN self-hosted flow this PR fixes still fails. The PR body's claim that "App commands are auto-exposed to the renderer (no capability entry — same as the existing `core_rpc_url`/`core_rpc_token`)" is incorrect: those two precedents are themselves in this allow-list, which contradicts the claim. (Same finding as the Codex P1; also a documented project gotcha — unlisted invokes silently deny.)

**Suggested change** (in `allow-core-process.toml`, CORE PROCESS section):
```
    "core_rpc_url",
    "core_rpc_token",
    # Host-side HTTP relay for LAN self-hosted runtimes the secure
    # `tauri://localhost` webview cannot fetch directly (cleartext mixed
    # content). See #3865. Without this entry the invoke is ACL-denied.
    "relay_http_rpc",
```
Please also verify on a built `.app` (not `tauri dev`) against a real LAN runtime — unit tests mock `invoke`, so they cannot catch an ACL denial.

### Minor

#### 2. `app/src/services/coreRpcClient.ts` (`relayRpcViaShell`) — abort listener leaks on the success path
When a `signal` is passed, the `abort` listener is registered `{ once: true }` but is never removed if the invoke resolves/rejects first. For long-lived callers that reuse a signal this slowly accumulates listeners. Low impact, but cheap to fix: remove the listener (or use `Promise.finally`) once the invoke settles.

#### 3. `app/src-tauri/src/core_rpc.rs` (`relay_http_rpc`) — relay command is untested beyond the bearer helper
Only `relay_bearer_header` has Rust unit tests; the command itself (status/body passthrough, transport-error → `Err`) is uncovered. Given the changed-line coverage gate, consider a small test using a local mock server, or note why it's deferred.

## Verified / looks good
- Root-cause analysis (secure-context mixed-content block) is accurate and well-documented in code comments.
- `rpcUrlNeedsShellRelay` boundaries are correct and tested: loopback / `::1` / `*.localhost` / all https stay on direct fetch; only non-loopback http relays; malformed URLs fall through to no-relay (safe default).
- Verbatim status+body passthrough preserves JSON-RPC error envelopes and 4xx/5xx; `relayRpcViaShell` synthesizes a real `Response` so downstream `.ok`/`.status`/`.json()` consumers are unchanged.
- `relay_bearer_header` correctly suppresses a malformed `Authorization` header for blank/whitespace/missing tokens.
- `iOS CloudHttpTransport`/`TransportManager` left untouched — correct, that path is the separate iOS pairing flow.

## Note (process)
The first commit carries a `Co-Authored-By: Claude …` trailer, and the PR body has the "Generated with Claude Code" line. Per this repo's contribution norms AI-co-author trailers should not land in commits (AI authorship belongs in the PR template's metadata section, not commit trailers). Flagging for the author/maintainer to amend before merge.

M3gA-Mind:
Addressed in 349bce0:
- **`core_rpc.rs`**: all four `url` interpolations (debug logs + error `format!`s) now go through a `redact_url_for_log` helper that strips query/fragment/userinfo; the actual `reqwest` request still uses the unredacted URL.
- **`allow-core-process.toml`**: added `relay_http_rpc` to the allowlist next to `core_rpc_url`/`core_rpc_token`, closing the Tauri v2 ACL gap.

Declined the `coreIsTauri()` → `isTauri()` swap: `coreRpcClient.ts` imports `isTauri as coreIsTauri` from `utils/tauriCommands/common`, and *that* wrapper is already the bridge-gap-aware guard (it verifies `__TAURI_INTERNALS__.invoke` is wired before returning true). The relay branches already use it — switching to the bare `@tauri-apps/api/core` `isTauri` would be strictly less safe.

M3gA-Mind:
@coderabbitai review — the two actionable findings (relay URL redaction in logs/errors, and the missing `relay_http_rpc` ACL entry) are fixed in 349bce0; the `coreIsTauri` point is addressed in my reply above (it already aliases the bridge-gap-aware `isTauri`). Please re-review the latest head.

M3gA-Mind:
@coderabbitai review — all three points are now resolved on head (3628ab7): relay URL redaction in logs/errors, `relay_http_rpc` added to the ACL, and the relay guards now call `isTauri()` directly (it is the bridge-gap-aware guard from `utils/tauriCommands/common`, with an import comment explaining why it is the correct check). Please re-review the latest commit.

M3gA-Mind:

PR #3978, opened 2026-06-23
fix(chat): don't fold a pipe-bearing prose line into a preceding table

ly-wang19:
## Summary

- `splitAgentMessageIntoBubbles` extended a markdown-table block with every following line that `looksLikeMarkdownTableRow` (merely contains a pipe and ≥2 cells), regardless of column count.
- A normal prose line containing pipes — e.g. a shell snippet `run `'`ps aux | grep node`'` — was absorbed into the table; since its column count differs, `parseMarkdownTable` then rejects the block (returns `null`), so the table renderer never fires and the prose is merged into the table bubble.
- Only absorb continuation rows whose column count matches the header (matching `parseMarkdownTable`).

## Problem

The table-grouping loop used `looksLikeMarkdownTableRow` alone, which is true for any line with a pipe and 2+ cells. A common assistant reply — a table immediately followed by prose that contains a pipe (shell commands, "A | B" choices, etc.) — folded that prose line into the table segment. `parseMarkdownTable` requires every row to match the header column count, so the mismatched block becomes `null`: the dedicated table UI is suppressed and the prose line is rendered inside the table bubble.

## Solution

- Capture the header column count and require each continuation row to match it before absorbing it into the table block. A pipe-bearing prose line with a different cell count now starts its own bubble, and the table renders correctly.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — added a regression test; the existing table tests cover the happy path
- [x] **Diff coverage ≥ 80%** — the changed function is exercised by the `splitAgentMessageIntoBubbles` suite (`vitest`, 23 tests pass)
- [N/A] Coverage matrix updated — behaviour-only fix to already-covered code
- [N/A] Affected feature IDs under `## Related` — no matrix rows change
- [x] No new external network dependencies introduced — pure function, no I/O
- [N/A] Manual smoke checklist — does not touch release-cut surfaces
- [N/A] Linked issue — found via code review; no existing issue

## Proof

The new test feeds a 2-column table followed by `Either pick Basic | Pro, then run `'`ps aux | grep node`'.` (3 "cells"). On `main` the prose is folded into the table segment; with this change it becomes its own bubble and the table segment stays clean. Fails before, passes after.

## Release Notes

* **Bug Fixes**
  * Fixed message bubble parsing so prose lines containing pipe characters no longer get merged into preceding markdown tables when the table’s column structure doesn’t match.  
  * Improved markdown table bubble grouping so pipes inside inline code within the first data row don’t break the table into multiple bubbles.

* **Tests**
  * Added coverage for table/prose separation and for inline-code pipes within tables to ensure consistent message bubble rendering.

### DIFF
--- a/app/src/utils/__tests__/agentMessageBubbles.test.ts
+++ b/app/src/utils/__tests__/agentMessageBubbles.test.ts
@@ -64,6 +64,16 @@ describe('splitAgentMessageIntoBubbles', () => {
     ]);
   });
 
+  it('keeps a table whose first data row carries a code-span pipe as one readable block', () => {
+    // splitMarkdownTableCells splits naively on "|", so a genuine first data row
+    // with a pipe inside a code span over-counts its cells (3 vs the header's 2).
+    // Dropping it would leave a header+separator-only block that renders as an
+    // empty-row table and detaches the real row; keeping the first data row makes
+    // parseMarkdownTable return null so the block falls back to readable markdown.
+    const content = '| Command | Use |\n| --- | --- |\n| `ps aux | grep node` | Find it |';
+    expect(splitAgentMessageIntoBubbles(content)).toEqual([content]);
+  });
+
   it('keeps double-newline paragraphs in the same bubble', () => {
     const content = 'First line\nSecond line\n\nThird paragraph\nFourth line';
     expect(splitAgentMessageIntoBubbles(content)).toEqual([
--- a/app/src/utils/__tests__/agentMessageBubbles.test.ts
+++ b/app/src/utils/__tests__/agentMessageBubbles.test.ts
@@ -51,6 +51,19 @@ describe('splitAgentMessageIntoBubbles', () => {
     ]);
   });
 
+  it('does not absorb a following prose line whose pipes give a different column count', () => {
+    // The prose line satisfies looksLikeMarkdownTableRow (it contains pipes) but
+    // has 3 "cells" vs the table's 2; absorbing it makes parseMarkdownTable
+    // reject the block (null), suppressing the table renderer and merging the
+    // prose into the table bubble.
+    const content =
+      '| Plan | Cost |\n| --- | --- |\n| Basic | $10 |\nEither pick Basic | Pro, then run `ps aux | grep node`.';
+    expect(splitAgentMessageIntoBubbles(content)).toEqual([
+      '| Plan | Cost |\n| --- | --- |\n| Basic | $10 |',
+      'Either pick Basic | Pro, then run `ps aux | grep node`.',
+    ]);
+  });
+
   it('keeps double-newline paragraphs in the same bubble', () => {
     const content = 'First line\nSecond line\n\nThird paragraph\nFourth line';
     expect(splitAgentMessageIntoBubbles(content)).toEqual([
--- a/app/src/utils/agentMessageBubbles.ts
+++ b/app/src/utils/agentMessageBubbles.ts
@@ -53,15 +53,23 @@ export function splitAgentMessageIntoBubbles(content: string): string[] {
       // absorbed into the table — otherwise parseMarkdownTable rejects the
       // mismatched block (returns null), suppressing the table UI and merging
       // the prose line into the table bubble.
+      //
+      // Always keep at least the first data row, though: splitMarkdownTableCells
+      // splits naively on "|" and does not honor escaped "\|" or pipes inside
+      // code spans, so a genuine first row carrying such a pipe over-counts its
+      // cells. Dropping it here would leave a header+separator-only block that
+      // parseMarkdownTable renders as an empty-row table (and detaches the real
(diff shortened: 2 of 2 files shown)


oxoxDev (changes requested):
# PR #3978 — fix(chat): don't fold a pipe-bearing prose line into a preceding table

## Walkthrough
Tightens the table-grouping loop in `splitAgentMessageIntoBubbles` so a continuation line is only absorbed into a markdown-table block when its cell count matches the header. Previously any pipe-bearing prose line (`pick A | B`, `` `ps aux | grep node` ``) was folded in, which made `parseMarkdownTable` reject the whole block and merge the prose into the table bubble. Correct, narrowly-scoped fix with a regression test. One real edge case (already flagged by Codex) keeps it from a clean approve.

## Changes
| File | Summary |
| --- | --- |
| `app/src/utils/agentMessageBubbles.ts` | Capture `headerCellCount`; only absorb continuation rows whose `splitMarkdownTableCells` length matches |
| `app/src/utils/__tests__/agentMessageBubbles.test.ts` | Regression test: 2-col table + pipe-bearing prose line stays separate |

## Actionable comments (1)

### Major

#### 1. `app/src/utils/agentMessageBubbles.ts:50-58` — empty-table regression when the first data row carries an escaped / code-span pipe
`splitMarkdownTableCells` splits naively on `|` and does not honor `\|` or pipes inside code spans. So a legitimate table whose **first** data row contains such a pipe over-counts cells:

```
| Command | Use |
| --- | --- |
| `ps aux \| grep node` | Find it |
```

- `headerCellCount = 2`; the data row splits to 3 cells → the new `=== headerCellCount` guard rejects it on the first iteration, so the loop exits with `tableLines = [header, separator]` only.
- `parseMarkdownTable` then sees `rows = lines.slice(2) = []`; `[].some(...)` is `false`, so it returns a **valid zero-row table**. The renderer draws an empty-header table and the real data row detaches into its own bubble.
- On `main` (pre-PR) the greedy absorb pulled the row in, the count mismatch made `parseMarkdownTable` return `null`, and it fell back to readable raw markdown. So this PR turns a readable block into a broken empty table for that input — a regression, not just an unfixed case.

This is the same root flaw the PR fixes for prose (`splitMarkdownTableCells` miscounting), just shifted onto a genuine table. Codex flagged this inline (P2).

**Suggested change** — keep at least one data row with the table so a count mismatch makes `parseMarkdownTable` reject the whole block (preserving the old readable fallback) rather than emitting a partial table:

```
// after capturing headerCellCount and tableLines = [line, lines[index + 1]]; index += 2;
let dataRowsAbsorbed = 0;
while (index < lines.length && looksLikeMarkdownTableRow(lines[index])) {
  const matches = splitMarkdownTableCells(lines[index]).length === headerCellCount;
  // Always keep the first data row so a mismatch is rejected by parseMarkdownTable
  // (readable raw-markdown fallback) instead of leaving an empty-row table.
  if (!matches && dataRowsAbsorbed > 0) break;
  tableLines.push(lines[index]);
  dataRowsAbsorbed += 1;
  index += 1;
}
```

Add a test for a table whose first data row contains `` `…|…` `` / `\|` to lock the behavior.

## Verified / looks good
- Core fix is correct for the targeted prose case; regression test asserts the right two-bubble output (high confidence).
- Scope is tight (12/-1), pure function, no I/O, matches `parseMarkdownTable`'s own column-count contract.
- `splitMarkdownTableCells` reused for the header count — consistent with how the separator/parser measure cells.
- No `Co-Authored-By: Claude` trailer on the commit; single signed commit by the author.
- CI green (Frontend Quality, Submission Checklist pass; E2E/coverage pending). No PR-caused failures.
- Empty-cell / trailing-pipe non-edge cases match: `| a |  |` and `| a | b |` both count 2, so well-formed tables with empty cells are unaffected.

ly-wang19:
Thanks @oxoxDev — great catch, you're right that the guard regressed a genuine table whose **first** data row carries an escaped/code-span pipe (the `splitMarkdownTableCells` naive-split over-count shifted onto a real row), turning the old readable fallback into an empty-row table with a detached row.

Applied your suggested approach in `1f3cb6c`: track `dataRowsAbsorbed` and **always keep at least the first data row**, only breaking on a count mismatch once we've already absorbed one. So:

- A genuine table whose first row has a `` `…|…` `` / `\|` pipe → the mismatch now stays inside the block → `parseMarkdownTable` returns `null` → readable raw-markdown fallback (pre-PR behavior restored), instead of a partial empty-row table.
- A trailing **prose** line with mismatched pipes after ≥1 real row → still rejected (the original fix is preserved).

Added a regression test (`keeps a table whose first data row carries a code-span pipe as one readable block`) that asserts the whole block stays one bubble. It fails on the previous commit and passes now; full suite is 24/24, prettier + eslint clean.

senamakel:
great pr thanks for it!

PR #4031, opened 2026-06-24
fix(agent): cap max_tokens on the agent chat turn (#4005)

### LINKED ISSUE #4005: Cap max_tokens on the agent chat turn (currently None) so reservation-pricing providers' agent 402s stay preventable
## Problem

The agent chat turn currently sends `max_tokens: None`:

- `src/openhuman/agent/harness/engine/core.rs:531`
- `src/openhuman/agent/session/turn/session_io.rs:122`

Only memory **extraction** caps the output budget (`src/openhuman/memory_tree/score/extract/llm.rs:139`, `EXTRACTION_MAX_OUTPUT_TOKENS = 8192`). The agent turn itself has no cap.

## Why this matters

For **reservation-pricing** providers (e.g. OpenRouter), an uncapped request makes the provider reserve credit against the model's *full* output window during pre-flight. A user with a realistic-but-modest balance can then get a `402` purely because of the oversized reservation — a **preventable** condition: capping `max_tokens` would price the pre-flight against a realistic budget and let the request through.

PR #3976 demotes insufficient-credits `402`s from BYO compatible providers to `info` across all emit paths. That demote is correct for a *flat* balance-exhausted billing fact (DeepSeek "Insufficient Balance", TAURI-RUST-4QF — genuinely unpreventable, no local lever). But because the agent turn is uncapped, a reservation-window `402` on the agent path (the C62 shape) would *also* be caught by that arm and silently demoted — masking a defect we actually have a lever over.

## Fix direction

Thread an explicit `max_tokens` cap onto the agent chat turn (mirroring how extraction already caps it), so reservation-pricing providers price their pre-flight against a realistic output budget. Once the agent turn is capped, a residual `402` is genuinely the flat-balance case the #3976 demote arm is meant for.

## References

- PR #3976 (insufficient-credits 402 demote across compatible paths)
- Sentry: TAURI-RUST-C62 (reservation-window 402 context)

oxoxDev:
## Summary

- Add `AGENT_TURN_MAX_OUTPUT_TOKENS = 16384`, an explicit output-token cap for agent chat turns (previously `max_tokens: None`).
- Thread the cap onto **all** agent `ChatRequest` sites: main loop, session turn, engine adapter, and the subagent checkpoint summary.
- Keeps the reservation-pricing pre-flight realistic so a residual `402` is the genuine flat-balance case — not an artifact of an oversized reservation.

## Problem

The agent chat turn sent `max_tokens: None`. For reservation-pricing providers (e.g. OpenRouter), an uncapped request makes the provider reserve credit against the model's *full* output window (64k+) during its pre-flight balance check. A user with a realistic-but-modest balance then gets a `402` purely because of the oversized reservation — a **preventable** condition with a local lever.

This also interacts with the insufficient-credits `402` demote arm (PR #3976): because the agent turn was uncapped, a reservation-window `402` on the agent path (the C62 shape) would be caught by that arm and silently demoted to `info` — masking a defect we can actually fix, rather than the flat-balance fact the demote is meant for.

Only memory **extraction** capped its output budget (`EXTRACTION_MAX_OUTPUT_TOKENS`); the agent turn had no cap. The `ChatRequest::max_tokens` and `is_provider_insufficient_credits_402` docs already anticipated this cap ("residual case once the request already caps `max_tokens`").

## Solution

- New `AGENT_TURN_MAX_OUTPUT_TOKENS` constant (16384) beside `ChatRequest`, doc mirroring `EXTRACTION_MAX_OUTPUT_TOKENS` and cross-referencing C62. `16384` sits comfortably above any realistic single agent turn (`max_tokens` is an upper bound, not a forced length — the model still stops at its natural end on normal turns) while cutting the reservation 4× versus a 64k window.
- Set `max_tokens: Some(AGENT_TURN_MAX_OUTPUT_TOKENS)` at every agent `ChatRequest`:
  - `agent/harness/engine/core.rs` (main agent loop)
  - `agent/harness/session/turn/session_io.rs`
  - `agent/harness/session/turn_engine_adapter.rs`
  - `agent/harness/subagent_runner/ops/checkpoint.rs` (bounded progress summary)
- Updated the now-stale `ChatRequest::max_tokens` field doc and the default-`chat()` caveat that claimed agent turns pass `None`.
- The compatible provider (the reservation-pricing target) already overrides `chat()` and threads `max_tokens` onto the wire; providers on the default `chat()` path simply forgo the cap (harmless — they are not reservation backends), logged at `debug`.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/pr-ci.yml`](../.github/workflows/pr-ci.yml). Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge.
- [x] N/A: behaviour-only change — no feature row added/removed/renamed in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md).
- [x] N/A: no feature IDs from the matrix are affected by this guard change.
- [x] N/A: does not touch release-cut smoke surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)).

## Impact

- **Desktop/CLI (Rust core)**: agent turns now request at most 16384 output tokens. `max_tokens` is an upper bound, so normal turns are unaffected; only pathological runaway generations above 16k are bounded.
- **Billing/UX**: reservation-pricing providers price their pre-flight against a realistic budget, so modest-balance BYO users no longer hit a `402` from an oversized reservation. A residual `402` is then a genuine flat-balance condition.
- No migration, no schema change, no new dependency.

## Related

- Closes: #4005
- Follow-up PR(s)/TODOs: N/A
- Builds on PR #3976 (insufficient-credits `402` demote) — this makes the agent-path `402` preventable so the demote only catches the genuine flat-balance case.
- Sentry: TAURI-RUST-C62 (reservation-window `402` context).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/4005-cap-agent-max-tokens`
- Commit SHA: see PR head

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: no frontend change; ran `cargo fmt --check` (clean) instead.
- [x] N/A: `pnpm typecheck` — no TypeScript touched.
- [x] Focused tests: `cargo test --lib agent_turn_cap_reaches_the_wire` (pass).
- [x] Rust fmt/check (if changed): `cargo fmt --check` + `GGML_NATIVE=OFF cargo check`/`clippy --lib` (clean).
- [x] Tauri fmt/check (if changed): `pnpm rust:check` ran via pre-push hook (pass).

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: agent chat turns request a bounded `max_tokens` (16384) instead of unbounded.
- User-visible effect: fewer spurious `402 insufficient credits` errors for modest-balance BYO users on reservation-pricing providers.

### DIFF
--- a/src/openhuman/agent/harness/engine/core.rs
+++ b/src/openhuman/agent/harness/engine/core.rs
@@ -27,7 +27,7 @@ use crate::openhuman::agent::stop_hooks::{current_stop_hooks, StopDecision, Turn
 use crate::openhuman::context::guard::{ContextCheckResult, ContextGuard};
 use crate::openhuman::context::{summarize_chat_history, EngineAutocompact};
 use crate::openhuman::inference::provider::{
-    ChatMessage, ChatRequest, Provider, ProviderCapabilityError,
+    ChatMessage, ChatRequest, Provider, ProviderCapabilityError, AGENT_TURN_MAX_OUTPUT_TOKENS,
 };
 
 use super::super::parse::build_native_assistant_history;
@@ -528,7 +528,10 @@ pub(crate) async fn run_turn_engine(
                     messages: &prepared_messages_vec,
                     tools: request_tools,
                     stream: delta_tx_opt.as_ref(),
-                    max_tokens: None,
+                    // Cap the turn so reservation-pricing providers price their
+                    // pre-flight against a realistic budget, not the full output
+                    // window (TAURI-RUST-C62).
+                    max_tokens: Some(AGENT_TURN_MAX_OUTPUT_TOKENS),
                 },
                 model,
                 temperature,
--- a/src/openhuman/agent/harness/session/turn/session_io.rs
+++ b/src/openhuman/agent/harness/session/turn/session_io.rs
@@ -6,7 +6,9 @@ use super::super::types::Agent;
 use crate::openhuman::agent::harness;
 use crate::openhuman::agent::progress::AgentProgress;
 use crate::openhuman::context::ARCHIVIST_EXTRACTION_PROMPT;
-use crate::openhuman::inference::provider::{ChatMessage, ChatRequest, ProviderDelta, UsageInfo};
+use crate::openhuman::inference::provider::{
+    ChatMessage, ChatRequest, ProviderDelta, UsageInfo, AGENT_TURN_MAX_OUTPUT_TOKENS,
+};
 
 impl Agent {
     // ─────────────────────────────────────────────────────────────────
@@ -120,7 +122,8 @@ impl Agent {
                     messages: &messages,
                     tools: None,
                     stream: delta_tx_opt.as_ref(),
-                    max_tokens: None,
+                    // Reservation-pricing pre-flight budget cap (TAURI-RUST-C62).
+                    max_tokens: Some(AGENT_TURN_MAX_OUTPUT_TOKENS),
                 },
                 effective_model,
                 self.temperature,
--- a/src/openhuman/agent/harness/session/turn_engine_adapter.rs
+++ b/src/openhuman/agent/harness/session/turn_engine_adapter.rs
@@ -46,6 +46,7 @@ use crate::openhuman::agent_tool_policy::ToolPolicySession;
 use crate::openhuman::context::ReductionOutcome;
 use crate::openhuman::inference::provider::{
     ChatMessage, ChatRequest, ConversationMessage, Provider, ProviderDelta, ToolCall, UsageInfo,
+    AGENT_TURN_MAX_OUTPUT_TOKENS,
 };
 use crate::openhuman::tools::{Tool, ToolSpec};
 
@@ -476,7 +477,8 @@ impl CheckpointStrategy for AgentCheckpoint {
                     messages: &messages,
                     tools: None,
                     stream: delta_tx_opt.as_ref(),
-                    max_tokens: None,
+                    // Reservation-pricing pre-flight budget cap (TAURI-RUST-C62).
+                    max_tokens: Some(AGENT_TURN_MAX_OUTPUT_TOKENS),
                 },
                 &self.model,
                 self.temperature,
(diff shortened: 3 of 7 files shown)


M3gA-Mind:
Cross-referencing from Sentry triage of **TAURI-RUST-EWD** (#4046) — a heads-up on an interaction, not a blocking objection.

This PR caps `max_output_tokens` (16384) and threads it onto **all** agent `ChatRequest` sites, which is the right fix for the OpenRouter reservation-`402` (#4005 / TAURI-RUST-4NC, #4041).

⚠️ **Interaction with the Codex-OAuth `/responses` path.** Today the cap reaches `chat_via_responses` (`src/openhuman/inference/provider/compatible_helpers.rs:56-63`) only on the **memory-extraction** path (C62), and that is precisely what produces TAURI-RUST-EWD: `400 {"detail":"Unsupported parameter: max_output_tokens"}` from `chatgpt.com/backend-api/codex/responses`, which rejects the parameter outright (1,854 events / 8 users on `0.57.53`). By threading the cap onto every agent turn, this PR will **expand EWD's blast radius** from memory-extraction-only to all agent turns that route through Codex OAuth.

**Suggestion:** land the EWD guard alongside this change — in `chat_via_responses`, reuse the existing `is_codex_oauth_responses` check to **omit** `max_output_tokens` for the Codex `/responses` endpoint (plus a reactive strip-and-retry on that 400 body). That preserves the OpenRouter reservation cap while preventing new EWD `400`s on the agent path. Full analysis in #4046.

M3gA-Mind:
Heads-up follow-up: the `max_output_tokens` guard this PR needs for Codex-OAuth users is now up in #4068 — it omits `max_output_tokens` on the `backend-api/codex/responses` endpoint and adds a reactive strip-and-retry on a 400 rejecting the field. Once #4068 lands, capping `max_tokens` on agent turns here won't 400 for Codex-OAuth users.

PR #4064, opened 2026-06-24
fix(credentials): reclaim leaked self-owned auth-profile lock immediately (#2318)

### LINKED ISSUE #2318: Timed out waiting for auth profile lock
Sentry Issue: [TAURI-RUST-B1](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/466/?referrer=github_integration)

```
Timed out waiting for auth profile lock
```

M3gA-Mind:
## Summary

Fixes the sustained `Timed out waiting for auth profile lock` retry storm (Sentry **TAURI-RUST-B1**, 47,385 events on one wedged Windows machine).

### Root cause
The auth-profile store guards itself with an on-disk lock file containing `pid=<owner>`. On Windows, `AuthProfileLockGuard::drop`'s best-effort `fs::remove_file` is routinely blocked by AV / the Search indexer holding a transient handle on the just-written file. The lock then **leaks with this same process's still-alive pid inside it**.

The only recovery for a *live-pid* leaked lock was the age-based branch in `clear_lock_if_stale` (`STALE_LOCK_AGE_MS = 30_000`). That floor is far longer than the ~10s RPC timeout, so every `app_state_snapshot` → `load_app_session_profile` → `acquire_lock` timed out and the frontend retry-stormed for the full 30s window before age-reclaim finally cleared it — then the next snapshot leaked again and the cycle repeated every ~25–35s.

### Fix (two interdependent pieces — no timeout raised, no event demoted)
1. **In-process serialization** — a process-wide registry of `Mutex<()>` keyed by lock-file path (`in_process_lock_for`), held for the lifetime of the on-disk guard. Concurrent same-process acquirers queue in-memory instead of racing `create_new`/`Drop` on the file.
2. **Immediate self-owned reclaim** — `reclaim_self_owned_lock`: on `AlreadyExists`, if the lock file records *our own pid*, reclaim it immediately and skip the age floor. This is sound **only because** we hold the in-process mutex, which guarantees no live same-process guard exists — so a self-pid file is unambiguously a leaked `Drop` unlink.

Without the in-process mutex the immediate reclaim would be unsafe (a second thread could delete a lock the first thread legitimately holds); the two pieces are required together.

`clear_lock_if_stale` is unchanged and still covers the genuine cross-process cases: a now-dead owner pid, a *different* still-alive process's aged leak, and malformed/pidless locks. `STALE_LOCK_AGE_MS`, `LOCK_TIMEOUT_MS`, and the RPC timeout are all untouched.

### Tests
- `reclaim_self_owned_lock_removes_leaked_self_pid_lock`
- `reclaim_self_owned_lock_leaves_foreign_pid_alone`
- `reclaim_self_owned_lock_is_noop_when_lock_missing`
- `acquire_lock_recovers_immediately_from_leaked_self_pid_lock` — asserts recovery in <5s vs the old 30s floor

All existing lock tests are unchanged and still pass (they exercise `clear_lock_if_stale` directly).

> Note: could not run `cargo check`/`test` locally — the `whisper-rs-sys` C++ dependency fails to build on this Apple-Silicon machine (documented toolchain gotcha), before any Rust is compiled. CI provides the authoritative build/test signal.

Sentry-Issue: TAURI-RUST-B1
Bucket: real-defect — RCA: leaked self-owned auth-profile lock now reclaimed immediately via an in-process mutex + self-pid fast path, eliminating the 30s-age-floor-behind-10s-RPC-timeout retry storm.

Closes #2318

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery from stuck authentication profile locks.
  * Locks left behind by the same process are now cleared immediately, reducing delays when signing in or loading profiles.
  * Added safer handling for stale or missing lock files so lock recovery behaves more reliably.

### DIFF
--- a/src/openhuman/credentials/profiles.rs
+++ b/src/openhuman/credentials/profiles.rs
@@ -7,7 +7,7 @@ use std::collections::{BTreeMap, HashMap};
 use std::fs::{self, OpenOptions};
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::sync::{Mutex, MutexGuard, OnceLock};
+use std::sync::{Mutex, MutexGuard, OnceLock, TryLockError};
 use std::thread;
 use std::time::{Duration, Instant};
 
@@ -1181,19 +1181,6 @@ impl AuthProfilesStore {
                 .with_context(|| "Failed to create auth profile lock directory".to_string())?;
         }
 
-        // Serialize same-process acquirers on an in-memory lock keyed by this
-        // path before we ever touch the on-disk lock file. Two things depend on
-        // holding it: (1) concurrent `app_state_snapshot` calls in this process
-        // queue here instead of racing `create_new`/`Drop` on the file, and
-        // (2) it lets us treat an on-disk lock recording our own pid as a leaked
-        // `Drop` unlink and reclaim it immediately — no other thread in this
-        // process can hold a live guard while we own this in-memory lock (see
-        // `reclaim_self_owned_lock`). Held for the lifetime of the returned
-        // guard. Poison is recoverable: the `()` payload carries no invariant.
-        let in_process_guard = in_process_lock_for(&self.lock_path)
-            .lock()
-            .unwrap_or_else(|poisoned| poisoned.into_inner());
-
         // Drive timeout + stale-recheck off wall-clock elapsed time, not the
         // sum of explicit `thread::sleep(LOCK_WAIT_MS)` calls. The earlier
         // counter-based approach excluded time spent inside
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
Thanks for the review — all three points were valid and are addressed in 6e5120d:

1. **In-process wait must respect `LOCK_TIMEOUT_MS`** (CodeRabbit / Codex): swapped the blocking `lock()` for a `try_lock()` retry loop bounded by the same `LOCK_TIMEOUT_MS` budget, with `started_at` initialized *before* the acquisition. A wedged in-process holder can no longer strand an RPC/blocking worker past the caller's timeout. Poison is still recovered via `into_inner()`.

2. **Canonicalize the registry key** (CodeRabbit / Codex): `in_process_lock_for` now keys by the canonicalized parent directory + lock filename, so aliased spellings (symlink, relative vs absolute, trailing `.`, Windows case variants) map to one mutex. `acquire_lock` always calls this after `create_dir_all(parent)`, so `canonicalize` succeeds; it falls back to the raw path otherwise.

3. **Debug breadcrumbs in `reclaim_self_owned_lock`** (nitpick): added `trace!`/`debug!` for the probe entry, the unreadable-lock exit, and the not-self-owned exit, alongside the existing success/error logs.

Also added `acquire_lock_serializes_same_process_acquirers`, which exercises the `try_lock` WouldBlock path and verifies a concurrent same-process acquirer serializes behind the in-process lock rather than racing the file.

Issue #3201, opened 2026-06-02
Codex OAuth Responses API requests fail unless store=false is sent


Toshi3886:
## Summary

OpenHuman's Codex/ChatGPT OAuth path can connect to the OpenAI Codex backend and list models, but chat turns fail because the Responses API request does not explicitly set `store: false`.

Runtime error:

```text
openai Responses API error: {"detail":"Store must be set to false"}
```

## Environment

- App version: `v0.57.5`
- Auth mode: imported existing Codex CLI / ChatGPT login via `Codex 인증` (`~/.codex/auth.json`)
- Provider route: `OpenAI`
- Models tested: `gpt-5.5`, `gpt-5.3-codex-spark`
- Latest `main` inspected: `1e03dd2687d69d3cefcf723a7192e622436c3c81`

## Reproduction

1. Open `Settings -> AI -> LLM`.
2. Import Codex auth with `Codex 인증`.
3. Set routing provider to `OpenAI`.
4. Save with either `gpt-5.5` or `gpt-5.3-codex-spark`.
5. Open a new chat thread.
6. Send a short message, e.g. `test`.

## Actual result

The UI shows:

```text
Something went wrong. Please try again. This error has been reported.
```

The log shows:

```text
[providers][chat-factory] role=chat slug=openai model=gpt-5.3-codex-spark endpoint_host=https://api.openai.com
[providers][chat-factory] role=chat slug=openai codex_oauth=true endpoint_host=https://chatgpt.com account_id_header=true
[agent_loop] model pinned model=gpt-5.3-codex-spark
llm_provider.responses_api failed: openai Responses API error: {"detail":"Store must be set to false"}
agent.provider_chat failed: openai Responses API error: {"detail":"Store must be set to false"}
web_channel.run_chat_task failed: ... openai Responses API error: {"detail":"Store must be set to false"}
```

## Expected result

Codex OAuth chat should succeed, or at minimum the Responses API payload should comply with the backend requirement by sending `store: false`.

## Likely cause

Current `main` appears to build `ResponsesRequest` without a `store` field, and `chat_via_responses` sends only `model`, `input`, `instructions`, and `stream: Some(false)`.

## Suggested fix

Add `store: Option<bool>` to `ResponsesRequest` and set it to `Some(false)` when building the Responses API request:

```rust
let request = ResponsesRequest {
    model: model.to_string(),
    input,
    instructions,
    stream: Some(false),
    store: Some(false),
};
```

[...]

PR #4068, opened 2026-06-24
fix(inference): don't send max_output_tokens to the Codex OAuth /responses endpoint

### LINKED ISSUE #4046: openai Responses API error: (TAURI-RUST-EWD)
## Source

Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/16556/
Short ID: `TAURI-RUST-EWD` · Project: `tauri-rust`
Events: 1854 · Users affected: 8 · First seen: 2026-06-18 · Last seen: 2026-06-24
Sentry assignee: Mega Mind

## Symptom

```
openai Responses API error: {"detail":"Unsupported parameter: max_output_tokens"}
```

## Triage

Auto-created from Sentry triage — top-25 `tauri-rust` issue by event volume, assigned to **Mega Mind** in Sentry. Full stack trace, breadcrumbs, tags in the Sentry link above.

M3gA-Mind:
## Problem
Users signed in via **OpenAI Codex / ChatGPT OAuth** route every `chat()` through the Responses path (`responses_api_primary`). Memory-tree extraction sets `ChatRequest.max_tokens` (the entity-extraction cap), which reached the wire as `max_output_tokens`. The Codex backend (`chatgpt.com/backend-api/codex/responses`) **rejects** that parameter:

```
openai Responses API error: {"detail":"Unsupported parameter: max_output_tokens"}
```

So memory extraction failed on **every** run for these users (1,854 events; each extraction burns 3 retries). Pure background workload — same class as the endpoint rejecting `stream: false` (#3201).

## Fix
- **Per-endpoint shaping** — reuse the existing `is_codex_oauth_responses` detection in `chat_via_responses` to **omit** `max_output_tokens` for the Codex endpoint. A real `/v1/responses` still receives the cap (no silent uncapping of extraction, TAURI-RUST-C62).
- **Reactive strip-and-retry** — a 400 whose body flags `max_output_tokens` as unsupported triggers one retry with the field removed (bounded to a single retry), mirroring the existing no-tools / `frequency_penalty` retries. Defense-in-depth for other strict Responses backends.

This is also the guard PR #4031 needs: #4031 caps `max_tokens` on agent chat turns, which flows through this same Responses path and would widen this bug to agent turns for Codex-OAuth users unless the param is gated here first.

## Tests
- Codex endpoint request body omits `max_output_tokens` (exact-body match).
- Standard `/v1/responses` keeps it.
- Reactive retry recovers after a 400 rejecting the field.
- Unit test for the detection helper.

Closes #4046
Sentry-Issue: TAURI-RUST-EWD
Bucket: real-defect / preventable — RCA: we sent max_output_tokens to an endpoint that rejects it, with no per-endpoint capability gate. Gated the field by endpoint + reactive strip-and-retry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Responses endpoints that reject `max_output_tokens`, so affected requests can now succeed automatically.
  * Added a fallback retry when an endpoint reports that `max_output_tokens` is unsupported.
  * Preserved the token limit for standard Responses endpoints, avoiding changes for supported setups.
  * Improved error handling for failed Responses requests, resulting in clearer recovery behavior and more reliable request processing.

### DIFF
--- a/src/openhuman/inference/provider/compatible_helpers.rs
+++ b/src/openhuman/inference/provider/compatible_helpers.rs
@@ -53,147 +53,189 @@ impl OpenAiCompatibleProvider {
             })
             .unwrap_or(false);
 
-        let request = ResponsesRequest {
+        // TAURI-RUST-EWD: the Codex/ChatGPT OAuth Responses endpoint
+        // (`chatgpt.com/backend-api/codex/responses`) rejects `max_output_tokens`
+        // outright (400 `Unsupported parameter: max_output_tokens`), the same way
+        // it rejects `stream: false` (#3201). Omit the cap proactively for that
+        // endpoint; every other Responses backend keeps it (the cap still flows
+        // through for `responses_api_primary` on a real `/v1/responses`).
+        if is_codex_oauth_responses && max_output_tokens.is_some() {
+            log::debug!(
+                "[provider] {} omitting max_output_tokens={:?} for Codex OAuth responses endpoint (unsupported param)",
+                self.name,
+                max_output_tokens,
+            );
+        }
+        let mut request = ResponsesRequest {
             model: model.to_string(),
             input,
             instructions,
             stream: Some(is_codex_oauth_responses),
             store: Some(false),
-            max_output_tokens,
+            max_output_tokens: if is_codex_oauth_responses {
+                None
+            } else {
+                max_output_tokens
+            },
         };
 
         let url = self.responses_url();
+        let mut retried_without_cap = false;
 
-        let response = self
-            .apply_auth_header(self.http_client().post(&url).json(&request), credential)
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
Good catch — fixed in 25dc5e7. `err_indicates_max_output_tokens_unsupported` no longer matches `invalid` (and `unexpected` is now `unexpected parameter`), so it only fires on *unrecognized-parameter* wording. An invalid-**value** 400 (e.g. cap above the model's allowed range) now surfaces instead of silently stripping the cap and uncapping the request. Added negative test cases for both value-range phrasings.

The CI failures were a `cargo fmt` diff that cascaded into the skipped coverage jobs — also fixed.

PR #4109, opened 2026-06-25
fix(channels): unify the messaging default-channel selector into one grid

graycyrus:
## Summary

The **Messaging** settings panel rendered the channel list **twice** — a grid of connect/configure tiles plus a separate "Default Messaging Channel" button list, both labeled the same — so users couldn't tell which one actually set the default (the two even covered different channel sets).

This collapses them into a **single horizontal-card grid** where each tile shows connection status, opens setup on click, and owns the default selection:

- Icon on the left; **name → status → default control** stacked on the right.
- **"Set as default" appears only on connected channels**; the current default shows a **✓ Default** badge plus a primary ring on the tile.
- **Connected** and **not-connected** channels render as **two separate rows** (no divider/label between them).
- **Web** stays selectable (it's always available / built-in).

### Before → After
- Before: a connect-tile grid **and** a duplicate "DEFAULT MESSAGING CHANNEL" button list, same labels, ambiguous.
- After: one grid; the tile is the single source of truth for both connecting and choosing the default.

## Changes
- `app/src/pages/Skills.tsx` — `ChannelTile` redesigned (horizontal layout, connected-only "Set as default", default badge + ring); the channels tab now renders connected/not-connected as two grids; removed the duplicate selector list.
- i18n: added `channels.setAsDefault` / `channels.defaultBadge` across all 14 locales (real translations).
- Tests: `Skills.channels-grid.test.tsx` updated + new cases (connected-only control, connected-first grouping). The e2e/playwright "switch default channel" specs now switch via the always-connected **Web** tile (disconnected channels no longer expose the control), pre-setting Telegram as default so Web becomes a connected, non-default tile.

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes (0 errors)
- [x] `pnpm format` applied
- [x] `pnpm i18n:check` passes (full parity across 14 locales)
- [x] Unit: `Skills.channels-grid.test.tsx` 7/7; full `pages/__tests__` suite 225/225
- [x] E2E (WDIO + Playwright) "switch default channel" specs updated for the new gating — validated by CI
- [x] Launched the desktop app locally and verified the new layout on Connections → Messaging

### DIFF
--- a/app/src/lib/i18n/ar.ts
+++ b/app/src/lib/i18n/ar.ts
@@ -902,6 +902,8 @@ const messages: TranslationMap = {
   'channels.status.error': 'خطأ',
   'channels.status.configuring': 'جارٍ الضبط',
   'channels.defaultMessaging': 'قناة المراسلة الافتراضية',
+  'channels.setAsDefault': 'تعيين كافتراضي',
+  'channels.defaultBadge': 'افتراضي',
   'webhooks.title': 'الـ Webhooks',
   'webhooks.create': 'إنشاء Webhook',
   'webhooks.noWebhooks': 'لا توجد Webhooks مضبوطة',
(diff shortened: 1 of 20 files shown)


graycyrus:
@coderabbitai review

PR #4218, opened 2026-06-26
fix(tinyplace): read trading balance from tiny.place settlement RPC

### LINKED ISSUE #4202: bug: wallet balance shows "Unknown" in trading confirmation — balance fetch fails even with wallet connected
## Bug: Wallet balance shows "Unknown" in trading confirmation — "We couldn't verify your wallet balance"

**Epic**: #4190 — Tiny Place

**Observed behaviour**:
In the "Confirm offer" modal for identity trading, the "Your balance" field shows `Unknown` and the wallet field shows `—`. A warning is displayed: `"We couldn't verify your wallet balance. You can still continue — the payment is checked when it is submitted."` This occurs even when a wallet is connected.

**Expected behaviour**:
When a wallet is connected, the confirmation modal should display the actual USDC balance and the connected wallet address before the user submits an offer.

- [ ] Is the wallet RPC (`wallet_get_balance` or equivalent) failing or timing out before the modal opens?
- [ ] Is the Solana RPC endpoint reachable from the desktop app?
- [ ] Is the wallet connection state available at the time the modal renders, or is it async and the modal opens before the balance resolves?
- [ ] Check `src/openhuman/wallet/` for balance fetch logic and error handling
- [ ] The fallback message "payment is checked when submitted" should not be the primary UX — fix the root fetch failure

**Labels**: `bug`, `P1`, `tinyplace`

M3gA-Mind:
## Root cause

In the identity-trading confirm modal, **Your balance** showed `Unknown` (with the *"We couldn't verify your wallet balance"* warning) and the wallet showed `—`, even with a connected, funded wallet.

`wallet_usdc_balance` (`src/openhuman/tinyplace/manifest.rs`) queried the bare **public** cluster RPC (`cluster.rpc_url()` → `api.mainnet-beta.solana.com`) for the SPL USDC token account. But tiny.place settles on its **own** Solana RPC — the chain where agent token accounts are actually funded (see `tinyplace_solana_rpc_endpoints` docs) — and the public endpoint also rate-limits/restricts `getTokenAccountsByOwner`. So the lookup returned `None` and the card fell back to "Unknown".

## Fix

- Query the same ordered endpoints the payment path broadcasts through (`tinyplace_solana_rpc_endpoints()`: tiny.place RPC first, public cluster as fallback), so the displayed balance matches the chain the payment will actually settle on.
- Advance to the next endpoint only on a transport/parse/malformed-shape failure; a well-formed `/result/value` (including an empty array = real zero balance) is authoritative and returns immediately.
- Regression test points `TINYPLACE_SOLANA_RPC_URL` at a mock RPC and asserts the parsed USDC balance comes back as `Some`.

Closes #4202

### DIFF
--- a/src/openhuman/tinyplace/manifest.rs
+++ b/src/openhuman/tinyplace/manifest.rs
@@ -659,38 +659,65 @@ pub(crate) fn handle_tinyplace_registry_register(params: Map<String, Value>) -> 
 /// render (it falls back to letting the backend reject an underfunded payment).
 async fn wallet_usdc_balance(address: &str) -> (Option<Value>, String) {
     // The wallet's `balances()` only reports NATIVE assets (SOL/ETH/…), never SPL
-    // tokens — so query the SPL USDC balance directly from the configured Solana
-    // cluster's RPC (getTokenAccountsByOwner for the cluster's USDC mint). RPC
-    // failure → None (UI shows "Unknown"); RPC ok but no token account → 0.
-    let cluster = crate::openhuman::wallet::solana_cluster();
-    let mint = cluster.usdc_mint();
-    let rpc_url = cluster.rpc_url();
+    // tokens — so query the SPL USDC balance directly via
+    // `getTokenAccountsByOwner` for the cluster's USDC mint. RPC failure → None
+    // (UI shows "Unknown"); RPC ok but no token account → 0.
+    //
+    // TAURI-RUST / #4202: query tiny.place's own Solana settlement RPC FIRST, not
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
pnpm review: addressed @chatgpt-codex-connector P2 (settlement RPC URLs with embedded provider tokens were logged at warn level on per-endpoint failover) — routed all three new warn logs through the wallet layer's `redact_rpc_url` (scheme://host only) via `9eb1fe6` (src/openhuman/tinyplace/manifest.rs, src/openhuman/wallet/mod.rs). No other findings. CI re-running.

PR #4219, opened 2026-06-26
fix(mcp): make additive mcp_servers column migration idempotent under concurrency

### LINKED ISSUE #4194: bug: MCP Servers page shows "Failed to add deployment_url column to mcp_servers" on every load
## Bug: MCP Servers page shows "Failed to add deployment_url column to mcp_servers" on load

**Epic**: #4190 — Skills & MCP Servers

**Observed behaviour**:
Opening the MCP Servers page (Connections → MCP Servers) displays a red error banner: `"Failed to add deployment_url column to mcp_servers"`. The registry list still renders below the error but the error is always present on load.

**Expected behaviour**:
The page should load cleanly with no error banner. Schema migrations should be idempotent — if the column already exists, the migration should succeed silently.

**Root cause**:
A schema migration is being re-run on every app start (or page load) and failing because `deployment_url` already exists in the `mcp_servers` table. The migration likely lacks an `IF NOT EXISTS` guard.

**Fix**:
- Find the migration that adds `deployment_url` to `mcp_servers` (likely in `src/openhuman/` persistence layer or SQLite migration files)
- Add `IF NOT EXISTS` to the `ALTER TABLE` statement, or gate the migration behind a schema version check
- Ensure the error is not surfaced to the UI if it's a non-fatal "column already exists" error

**Labels**: `bug`, `P1`

M3gA-Mind:
The MCP Servers page showed a red "Failed to add deployment_url column to mcp_servers" banner on load. The migration guards each `ALTER` with a `PRAGMA table_info` snapshot, but that check-then-alter isn't atomic across connections: every store call opens its own `Connection` and runs `init_schema`, so the several MCP RPCs one page load fans out can each see the column missing before any adds it, then race to `ALTER` — losers failing with "duplicate column name" (SQLite `ADD COLUMN` has no `IF NOT EXISTS`).

Route the three additive ALTERs through `add_column_idempotent`, which treats a "duplicate column name" failure as success while still propagating any other error. Adds tests for the tolerated-duplicate and still-propagated-error cases.

Closes #4194

### DIFF
--- a/src/openhuman/mcp_registry/store.rs
+++ b/src/openhuman/mcp_registry/store.rs
@@ -134,9 +134,7 @@ fn add_column_idempotent(conn: &Connection, ddl: &str, what: &str) -> Result<()>
         Err(rusqlite::Error::SqliteFailure(_, Some(msg)))
             if msg.contains("duplicate column name") =>
         {
-            log::debug!(
-                "[mcp_registry] {what} already present (concurrent migration) — skipping"
-            );
+            log::debug!("[mcp_registry] {what} already present (concurrent migration) — skipping");
             Ok(())
         }
         Err(e) => Err(anyhow::Error::new(e).context(format!("Failed to add {what}"))),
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
pnpm review: flagged a Rust fmt violation (mcp_registry/store.rs:134,792) — applied rustfmt; no logic findings (the idempotent additive-migration change is sound). CI re-running.

PR #4220, opened 2026-06-26
perf(mcp): return MCP tool results as soon as the reply frame arrives

### LINKED ISSUE #4195: perf: skill-based prompt responses take up to 11 minutes — inference latency unacceptable
## Bug: Skill-based prompt responses are extremely slow (11 minutes observed)

**Epic**: #4190 — Skills & MCP Servers

**Observed behaviour**:
A prompt that invokes a skill takes up to 11 minutes to return a response. Skills themselves are functional (they connect and execute) but inference latency is unacceptably high.

**Expected behaviour**:
Skill-based prompt responses should complete within a reasonable time (target: <60s for complex tool chains, <15s for simple ones).

- [ ] Model selection: is the skill tool loop using a heavyweight/slow model (e.g. `reasoning-v1`, GPT-4o) instead of a faster tier?
- [ ] Tool round-trips: how many sequential tool calls are being made? Is there unnecessary iteration?
- [ ] MCP server response time: are individual MCP tool calls slow, or is the bottleneck in the orchestrator?
- [ ] Token budget: is the context window being bloated with unnecessary memory/context injection on each skill turn?
- [ ] Queue/concurrency: is the skill run being queued behind other jobs? Check `src/openhuman/agent/harness/` iteration cap and cron queue
- [ ] Streaming: is the response being buffered and only flushed at the end rather than streamed incrementally?

**Impact**: P1 — feature is functionally broken; 11-minute waits make skills unusable in practice.

M3gA-Mind:
**Root cause:** the MCP Streamable-HTTP client read the whole response body with `response.text().await` before parsing (`mcp_client/client.rs` `read_response`). For a `text/event-stream` response that blocks until the server *closes* the stream — but the MCP spec lets a server keep the SSE stream open after sending its single JSON-RPC reply. Against such a server every skill/MCP `tools/call` stalled until the stream closed or the request timeout fired, and the stalls compounded across a skill's tool chain — the dominant transport cause of the multi-minute skill latency.

**Fix:** read the SSE body incrementally and return the instant the first complete `data:` reply frame is buffered (new `first_complete_sse_data` helper, which only parses fully blank-line-terminated events so a half-received frame is never decoded as truncated JSON). Non-streamed JSON responses and the request-level reqwest timeout are unchanged. Adds unit tests for the early-return / keepalive / CRLF / partial-frame cases.

Note: other latency contributors flagged in the issue (model tier on the orchestrator, number of serial tool round-trips per turn) are separate, larger concerns; this PR fixes the one concrete transport bug.

Closes #4195

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streaming MCP responses so results can appear sooner when the first complete event arrives.
  * Made non-success HTTP responses return a clearer generic error message instead of a special-case auth path.
  * Better support for event-stream responses with mixed line endings and partial payloads.

### DIFF
--- a/src/openhuman/mcp_client/client.rs
+++ b/src/openhuman/mcp_client/client.rs
@@ -2,6 +2,7 @@ use crate::openhuman::config::{McpAuthConfig, McpClientIdentityConfig};
 use crate::openhuman::workflows::types::ToolResult;
 use anyhow::Context;
 use base64::Engine;
+use futures_util::StreamExt;
 use parking_lot::Mutex;
 #[cfg(test)]
 use reqwest::header::HeaderMap;
@@ -746,7 +747,8 @@ pub fn redact_endpoint(raw: &str) -> String {
 #[path = "client_helpers.rs"]
 mod client_helpers;
 use client_helpers::{
-    header_to_string, parse_sse_events, parse_sse_message, parse_www_authenticate_challenge,
+    first_complete_sse_data, header_to_string, parse_sse_events, parse_sse_message,
+    parse_www_authenticate_challenge,
     x_mcp_headers_from_schema,
 };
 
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
pnpm review: flagged a Rust fmt violation (mcp_client import block) — applied rustfmt (a839cb6a9). Reviewed the early-return-on-reply-frame perf change; logic sound, mcp_client tests pass (49/0). No further findings. CI re-running.

PR #4222, opened 2026-06-26
feat(mascot): amplitude-driven TTS lip-sync for Tiny (Option A)

### LINKED ISSUE #4077: feat: lip sync Tiny mascot mouth animation to TTS audio output
## Summary

When the agent responds via text-to-speech, Tiny the mascot is visible and animates, but there is no lip sync between the mouth/face animation and the spoken audio. The mascot needs to visually articulate in sync with what is being said.

## Current Behaviour

- Tiny is displayed and may have an idle or generic animation while TTS audio plays
- The mascot's mouth does not move in sync with the speech — there is no phoneme/viseme-driven animation tied to the audio output
- The disconnect between the voice and the static/looping mascot animation breaks the illusion of Tiny actually speaking

## Expected Behaviour

- Tiny's mouth animates in sync with the TTS audio as it plays
- At minimum: a mouth-open/mouth-closed animation driven by audio amplitude/energy (simple but effective)
- Ideal: viseme-driven lip sync where mouth shape matches the phoneme being spoken (e.g. open vowels, closed consonants, lip-rounding)
- Animation starts exactly when audio playback begins and stops when it ends
- During silence/pauses in speech, mouth returns to closed/neutral position

### Option A — Amplitude-based (simplest)
- Analyse the TTS audio buffer in real time (or ahead of playback)
- Drive a mouth-open scale/shape based on RMS amplitude of each audio frame
- No phoneme analysis needed — just open/close correlated with loudness
- Suitable as a fast first pass

### Option B — Viseme mapping from TTS timestamps
- Many TTS APIs (ElevenLabs, Azure, OpenAI TTS) return word/phoneme timestamps alongside audio
- Map phoneme timestamps to a set of viseme mouth shapes (A, E, I, O, U, M/B/P, F/V, etc.)
- Drive Tiny's mouth sprite/shape key to the correct viseme at the correct timestamp
- More natural-looking but requires TTS API support for timestamps

### Option C — Real-time audio analysis
- Use Web Audio API  to get frequency/amplitude data per frame during playback
- Drive mouth animation from the live analyser output
- Works regardless of TTS provider

## Relevant Code Areas

- Tiny mascot rendering:  (native window)
- TTS/voice output:  domain
- Frontend mascot component: search  for mascot-related components
- Audio playback: wherever TTS audio is streamed/played back in the frontend

## Acceptance Criteria

- [ ] Tiny's mouth visibly animates while TTS audio is playing
- [ ] Mouth animation is time-synced to audio — not just a looping idle
- [ ] Mouth closes/returns to neutral when audio stops or during pauses
- [ ] Animation works for both short and long responses
- [ ] No noticeable lag between audio and mouth movement (< 100ms offset)

M3gA-Mind:
## Summary

Implements **Option A** from #4077 — audio-amplitude-driven mouth open/close for the Tiny mascot, synced to the real TTS speech energy.

Previously the mouth was driven only by the viseme scheduler: real provider visemes when available, otherwise a **text-estimated procedural** track that isn't tied to the actual audio (drifts out of sync, and stays shut when a provider returns no visemes). This layers real audio energy on top so the mouth visibly tracks loudness in time with the speech.

## How it works

- **`voice/audioPlayer.ts`** — precompute an RMS energy envelope (~60 Hz windows) by decoding a **separate copy** of the TTS audio in a throwaway `AudioContext`, and expose `amplitude()` to sample it by playback time. The playing `HTMLAudioElement` is **never rerouted through Web Audio**, so there's zero risk of muting/perturbing TTS output. If Web Audio is unavailable (older CEF) or decode fails, the envelope is `null` and `amplitude()` returns `0` — a graceful no-op back to the existing viseme behaviour. Decoding is kicked off without blocking `audio.play()` so the autoplay gesture chain is preserved.
- **`voice/amplitudeMouth.ts`** (new, unit-tested) — pure helpers: normalise raw RMS → 0..1 mouth-open (noise floor + ceiling), smooth across frames (fast attack / slower release so it doesn't chatter), and bucket to a Rive viseme code.
- **`useHumanMascot.ts`** — each animation frame, layer the smoothed amplitude as a **floor** over the current viseme: it opens the mouth in sync with loudness even when the viseme track is flat/procedural, reinforces real visemes on loud syllables, and returns to neutral on silence/end. Works for both the Rive (discrete viseme enum) and SVG (continuous openness) renderers.

## Acceptance criteria (#4077)

- ✅ Mouth animates while TTS plays, time-synced to audio (envelope sampled by `currentTime`, <100ms windows)
- ✅ Returns to neutral on silence/pauses and at end (floor is 0 on silent frames; reset on playback start/stop)
- ✅ Works for short and long responses (envelope spans the full decoded buffer; existing max-duration safety unchanged)

## Tests / verification

- Added `amplitudeMouth.test.ts` (pure normalise/smooth/bucket logic). Existing `PlaybackHandle` test fakes updated with `amplitude()`.
- Manual verification (mouth visibly tracking audio) needs a real audio device + CEF and wasn't run locally — relying on PR CI for the unit/type/lint suite.

**Viseme/phoneme-accurate lip-sync (issue Options B/C) is a deliberate follow-up.**

Closes #4077

## Summary by CodeRabbit

* **New Features**
  * Added audio-responsive lip-sync where mouth openness and viseme selection react to spoken audio loudness.
  * Introduced amplitude-driven mouth motion with smoothing and discrete viseme bucketing.
* **Bug Fixes**
  * Improved robustness when audio decoding/Web Audio is unavailable and when playback ends.
  * Ensures amplitude influence is cleared/reset appropriately across speech and playback transitions.
* **Tests**
  * Added unit tests for amplitude-to-mouth mapping (normalization, smoothing, thresholds).
  * Added unit tests for amplitude handle behavior across decode/availability edge cases.
  * Added an end-to-end integration test validating amplitude-driven viseme openness changes.


M3gA-Mind:
@coderabbitai Thanks — addressed your review in `aa94e99df`. Disposition of the 3 items:

**Fixed — `normalizeAmplitude` non-finite input (amplitudeMouth.ts/test):** agreed on the asymmetry. `NaN` returned 0 but `+Infinity` fell through to the ceiling clamp (`→ 1`, a jam-mouth-wide-open failure). Now `!Number.isFinite(rms)` returns 0 uniformly, and the test asserts `NaN`, `+Infinity`, and `-Infinity` all map to a closed mouth.

**Declining — `smoothAmplitude(0.02, 0)` snap-to-zero test (amplitudeMouth.test.ts):** this isn't brittle in the way described. The assertion pins the *documented* snap-to-zero contract — the helper's own comment uses `0.02 → 0` as its canonical example (the release step from a near-floor value leaves a ~0.015 residual that must collapse to a fully-closed mouth). It's also robust to release-rate changes: for any `release ∈ (0, 1]`, `next = 0.02·(1 − release) < 0.02`, so it always snaps. Relaxing it to a "monotonic decrease" check would stop verifying the actual snap threshold the feature relies on to fully close the mouth on silence, so I'm keeping the exact assertion.

**Declining — audioPlayer envelope timing / ideal-60 Hz drift (audioPlayer.ts):** this appears to be a false positive against the current code. `buildRmsEnvelope` already derives the **real** per-frame duration from the rounded window — `frameDurationMs = (windowSize / buffer.sampleRate) * 1000` (windowSize = `round(sampleRate / ENVELOPE_HZ)`) — and `amplitude()` indexes with that exact same `frameDurationMs`, not `1000 / ENVELOPE_HZ`. The ideal period is never used for indexing; the code comment at the return site explicitly calls out the non-divisible-rate drift you describe and avoids it. No change needed.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review — all three items from your 08:51 review are addressed on the current head (`aa94e99df`): (1) `normalizeAmplitude` now returns 0 for all non-finite RMS (fixed); (2) the `smoothAmplitude(0.02,0)` snap test and (3) the audioPlayer `frameDurationMs` envelope-timing comment were declined with reasons in my comment above (the latter is a false positive — the code already indexes by the real `windowSize/sampleRate` duration, not the ideal 60 Hz period). Please re-review so the stale CHANGES_REQUESTED can clear.

PR #4223, opened 2026-06-26
fix(brain): reload memory graph after re-login

### LINKED ISSUE #4149: bug: Brain memory graph shows empty after logout and login — previous graph tree lost
## Summary

When a user logs out of OpenHuman and logs back in, the Brain page shows an empty memory graph. The previously built memory graph tree (nodes, connections, context) appears to be lost or not reloaded after re-authentication.

## Steps to Reproduce

1. Open OpenHuman and use the app so that the Brain page has a populated memory graph
2. Log out of the account
3. Log back in with the same credentials
4. Navigate to Brain page
5. Memory graph is empty — previous graph tree is gone

## Expected Behaviour

The Brain memory graph should persist across logout/login cycles. All previously built memory nodes, relationships, and graph structure should be restored and visible after logging back in with the same account.

## Actual Behaviour

The Brain page shows an empty graph after re-login. The memory tree that was built before logout is not displayed.

## Likely Causes to Investigate

- [ ] **Session-scoped memory load** — memory graph may only be loaded at initial session start and not re-fetched on re-authentication; the re-login flow may not trigger a fresh load from the persistent store
- [ ] **Workspace path mismatch** — after re-login the workspace directory resolves to a different path, so the memory store is not found
- [ ] **Graph renderer not reinitialised** — the UI component may not re-query the memory backend after the auth state changes, leaving it in the empty initial state
- [ ] **Memory cleared on logout** — the logout flow ( / ) may be clearing memory data it should not be touching
- [ ] **Auth-gated memory fetch** — the memory tree RPC call requires a valid session token; if the token is fetched before auth is fully restored on re-login, the call returns empty and the result is cached

## Environment

OpenHuman Beta, macOS

M3gA-Mind:
The Brain memory graph showed empty after a logout→login cycle. The graph is read from the on-disk memory store (`graph_export` reads the workspace DB — the data is never dropped), but the Brain load effect only fired on mount keyed on `[mode, refreshKey]`, so after re-auth the page kept its signed-out empty state and never refetched once the identity was restored.

Key the load effect on the authenticated identity (`snapshot.auth.userId` via `useCoreState`) so a re-login (null→user, or A→B) re-pulls the persisted graph — mirroring the thread-cache reload `CoreStateProvider` already does on identity change. Adds a Brain test asserting the graph refetches when the user re-authenticates.

Closes #4149

### DIFF
--- a/app/src/pages/Brain.tsx
+++ b/app/src/pages/Brain.tsx
@@ -26,6 +26,7 @@ import MemoryDebugPanel from '../components/settings/panels/MemoryDebugPanel';
 import BetaBanner from '../components/ui/BetaBanner';
 import { useSubconscious } from '../hooks/useSubconscious';
 import { useT } from '../lib/i18n/I18nContext';
+import { useCoreState } from '../providers/CoreStateProvider';
 import type { ToastNotification } from '../types/intelligence';
 import {
   type GraphExportResponse,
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
review: flagged a prettier violation in Brain.test.tsx — applied formatting (36a0fba4b) and merged latest main. lint/format clean, 11/11 unit tests pass. The memory-graph reload-on-relogin change is sound. CI re-running.

PR #4227, opened 2026-06-27
fix(provider): tolerate duplicate reasoning_content key from NVIDIA compat endpoint (TAURI-RUST-85R)

### LINKED ISSUE #4204: nvidia parse: tolerate duplicate reasoning_content key (TAURI-RUST-85R)
## Source
Sentry: https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/9253/ (TAURI-RUST-85R)
Events: 2,037 · Users affected: 5 · First seen: 2026-06-01 · Last seen: 2026-06-26 (still firing on 0.57.18)

## Symptom
``nvidia response parse error: duplicate field `reasoning_content` at line 1 column 742`` — the entire completion is dropped.

## Root cause
NVIDIA's OpenAI-compat endpoint returns a `message` object containing the key `reasoning_content` **twice in the same object** (a known quirk of some compat proxies for thinking models, e.g. `stepfun-ai/step-3.7-flash`). `ResponseMessage` has a hand-written `Deserialize` impl (`compatible_types.rs:347`) whose inner `Shadow` struct uses serde's **derived** deserializer with a single `reasoning_content: Option<String>` field — serde strict-rejects a repeated key (`#[serde(default)]` does not relax this), so the whole response fails to parse. Prior #3547/#3552 only fixed the *distinct-name* collision (`reasoning` + `reasoning_content`), not two identical keys. `StreamDelta::deserialize` (`compatible_types.rs:507`) has the same latent gap on the SSE path. Parse sites: `compatible_provider_impl.rs:848` (buffered, the event) and `compatible_stream_native.rs:198` (SSE fallback).

## Bucket / RCA
`real-defect` / broken-parsing-contract — a deserialize failure on valid (if quirky) provider output drops the completion. NOT a demote candidate; parsing must succeed. RCA fix: tolerate duplicate `reasoning_content` (last-wins / fold), e.g. deserialize via `serde_json::Value`/`Map` or a custom visitor that folds `reasoning_content` (+ `reasoning` alias) into the single canonical field — preserving "canonical wins". Apply to **both** `ResponseMessage` and `StreamDelta`.

## Dedupe
NEW (residual after #3547; not covered). No open PR / GH issue. Unrelated to our #4182-4189.

## Suggested scope
- `src/openhuman/inference/provider/compatible_types.rs` — `ResponseMessage::deserialize` (~347-375) and `StreamDelta::deserialize` (~507-530): tolerate duplicate `reasoning_content` keys.
- Regression test: a `message`/`delta` JSON with two `reasoning_content` keys parses successfully (alongside the existing #3547 reasoning+reasoning_content test).

_No AI-attribution trailers in commits/PR. Do NOT run local cargo — push --no-verify and let PR CI test._

rainbowpuffpuff:
## Summary
NVIDIA's `integrate.api.nvidia.com` OpenAI-compatible endpoint returns the
`reasoning_content` key **twice** in a single `message` object for some thinking
models (e.g. `stepfun-ai/step-3.7-flash`). `ResponseMessage` and `StreamDelta`
deserialize through an inner `Shadow` struct using serde's *derived* deserializer,
which strict-rejects a repeated key with `duplicate field reasoning_content` and
drops the entire completion (Sentry TAURI-RUST-85R — 2,037 events / 5 users, still
firing on 0.57.18). `#[serde(default)]` does not relax duplicate-key rejection, and
#3547 only fixed the distinct-name collision (`reasoning` + `reasoning_content`),
not two identical keys.

## Fix
Replace the `Shadow` structs in both `ResponseMessage::deserialize` and
`StreamDelta::deserialize` with hand-written `visit_map` visitors that consume map
entries manually. Repeated keys no longer error (last value wins — standard JSON
object semantics), and the canonical `reasoning_content` still wins over the
`reasoning` alias (#3547 preserved). Applies to both the buffered and SSE paths.

## Tests
- `duplicate_reasoning_content_in_response_message_does_not_error`
- `duplicate_reasoning_content_in_stream_delta_does_not_error`
- `duplicate_reasoning_content_still_beats_reasoning_alias`

Existing #3547 tests continue to pass.

Fixes #4204

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with OpenAI-compatible responses that may include repeated JSON fields.
  * Message and streaming parsing now tolerates duplicated `reasoning_content`, using the last occurrence.
  * Preserved fallback behavior so `reasoning` is used when `reasoning_content` is absent.
* **Tests**
  * Added regression coverage for duplicated `reasoning_content` in both non-streaming and streaming responses, including precedence when `reasoning` and duplicated `reasoning_content` are both present.

### DIFF
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -3242,8 +3242,8 @@ fn duplicate_reasoning_content_in_stream_delta_does_not_error() {
 #[test]
 fn duplicate_reasoning_content_still_beats_reasoning_alias() {
     let json = r#"{"choices":[{"message":{"reasoning":"alias cot","reasoning_content":"first","reasoning_content":"second"}}]}"#;
-    let resp: ApiChatResponse = serde_json::from_str(json)
-        .expect("doubled canonical key plus an alias must parse cleanly");
+    let resp: ApiChatResponse =
+        serde_json::from_str(json).expect("doubled canonical key plus an alias must parse cleanly");
     assert_eq!(
         resp.choices[0].message.reasoning_content.as_deref(),
         Some("second"),
--- a/src/openhuman/inference/provider/compatible_tests.rs
+++ b/src/openhuman/inference/provider/compatible_tests.rs
@@ -3204,6 +3204,53 @@ fn reasoning_and_reasoning_content_both_present_in_stream_delta_does_not_error()
     );
 }
 
+/// Regression for Sentry TAURI-RUST-85R: NVIDIA's `integrate.api.nvidia.com`
+/// OpenAI-compat endpoint returns the SAME key `reasoning_content` twice in one
+/// `message` object for some thinking models. A derived struct deserializer
+/// strict-rejects the repeat with `duplicate field \`reasoning_content\`` and
+/// drops the whole completion. The hand-folded `Visitor` must accept it, with
+/// the last value winning (standard JSON object semantics).
+#[test]
+fn duplicate_reasoning_content_in_response_message_does_not_error() {
+    let json = r#"{"choices":[{"message":{"content":null,"reasoning_content":"first cot","reasoning_content":"second cot"}}]}"#;
+    let resp: ApiChatResponse = serde_json::from_str(json)
+        .expect("a doubled reasoning_content key must parse without a duplicate-field error");
+    assert_eq!(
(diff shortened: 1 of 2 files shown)


rainbowpuffpuff:
Closing as a duplicate of #4207, which fixes #4204 (TAURI-RUST-85R) with the same map-fold Visitor approach and predates this PR — and additionally handles the last-non-null-wins edge case. Thanks @M3gA-Mind!

PR #4229, opened 2026-06-27
fix(tokenjuice): attribute compaction savings to the per-turn model (#4122)

### LINKED ISSUE #4122: Attribute TokenJuice savings cost to the per-turn model, not the configured default
## Summary

TokenJuice's compaction savings dashboard (`tokenjuice.savings_stats`) prices the tokens it saves using the **configured default model** (`config.default_model`), not the model the tool result is actually being compressed *for* on that turn. Attribute savings to the live per-turn model so the cost figure is accurate for sessions that use model overrides, per-agent models, or team lead/worker model splits.

## Problem / Context

The content router compacts a tool result inside the agent loop and records the savings via `savings::record(...)`, which calls `cost_saved_usd(model, ...)` with a process-global attribution model installed once at startup (`tokenjuice::savings::configure` from `config.default_model`).

The active per-turn model **is** available deeper in the harness (`run_turn_engine(..., model: &str, ...)`), but it is not threaded down to the compaction call sites (`agent/harness/session/agent_tool_exec.rs`, `agent/harness/engine/tools.rs`), and `AgentToolExecCtx` does not currently carry it. Threading it through both call sites + the turn engine was deferred to keep the initial savings feature small.

Impact today: cost-saved is correct when a session uses the default model, but skewed when:
- a per-turn model override is in effect,
- agents run on different models (lead vs worker, per-agent `model`),
- the result is destined for a cheaper/more-expensive tier than the default.

Token counts (the dominant metric) are unaffected — only the USD figure and the `byModel` breakdown attribution are.

## Scope (optional)

In scope:
- Thread the active model id from `run_turn_engine` to the tool-execution compaction call sites (extend `AgentToolExecCtx` and the `compact_output` / `compact_tool_output` signatures with an optional model).
- Pass it into `savings::record` so `cost_saved_usd` and the `byModel` bucket use the per-turn model; fall back to the configured default when unknown.

Out of scope:
- Output-token cost modeling / re-send amplification (a tool result re-enters context on every subsequent turn) — current model counts a single input occurrence, which is the conservative estimate.

## Acceptance criteria

- [ ] **Per-turn attribution** — savings recorded during a turn are priced with that turn's model, not the global default.
- [ ] **Graceful fallback** — when the per-turn model is unavailable, attribution falls back to `config.default_model` (current behavior), no panics.
- [ ] **byModel breakdown** — `tokenjuice.savings_stats` `byModel` reflects the real mix of models across a multi-model session.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [.github/workflows/pr-ci.yml](../../.github/workflows/pr-ci.yml)).

## Related

- Follow-up to the TokenJuice content-router / savings work (PR in flight on branch \`feat/tokenjuice-content-router\`).
- Code: \`src/openhuman/tokenjuice/savings.rs\`, \`src/openhuman/tokenjuice/compress.rs\`, \`src/openhuman/agent/harness/session/agent_tool_exec.rs\`, \`src/openhuman/agent/harness/engine/tools.rs\`, \`src/openhuman/agent/harness/engine/core.rs\`.

rainbowpuffpuff:
## Summary
TokenJuice's savings dashboard priced **every** compaction against a process-global
`attribution_model` pinned to `config::DEFAULT_MODEL` at startup (`savings.rs`). On
any turn or sub-agent running a different model, the saved-cost attribution (and the
`by_model` breakdown) was wrong.

## Fix
Thread the **per-turn model** to the savings recorder via a task-local, mirroring the
existing `model_vision_context` pattern:
- Add `TURN_MODEL` (`tokio::task_local!`) + `with_turn_model(model, fut)` in
  `tokenjuice/savings.rs`.
- Scope it around `run_turn_engine` at the two existing context-stack sites:
  `session/turn/core.rs` (main loop) and `subagent_runner/ops/loop_.rs` (sub-agents).
  Nested scopes mean a sub-agent on a different model attributes its own savings
  correctly.
- `savings::record` resolves the model via `resolve_attribution_model` — the scoped
  per-turn model when present, otherwise the configured default (so non-harness
  callers and tests are unaffected; **strictly additive**).
- Extract the aggregation into a pure `SavingsAggregate::record_saving` for race-free
  unit testing.

## Why a task-local (not param threading)
The repo already uses a family of per-turn task-local `*_context` carriers
(`model_vision_context`, `turn_attachments_context`, `turn_subagent_usage`, …)
stacked around `run_turn_engine` specifically to avoid widening the deep engine's
signature. `model_vision_context` already carries a per-turn **model** property to
the tool path this way. This change follows that convention; the task-local lives in
`tokenjuice` next to its reader (as `inference/provider/thread_context` does).

## Tests
- `record_saving_attributes_to_given_model` — pure aggregation on a local aggregate.
- `attribution_model_prefers_scoped_turn_model` / `blank_turn_model_falls_back_to_default`
  / `attribution_model_falls_back_to_default_when_unscoped` — resolve + task-local.
- Existing savings tests unchanged (`record`'s signature is unchanged).

## Notes
- Unknown/BYOK per-turn models price via the existing `cost::lookup_pricing`
  `<fallback>` rate, not $0.
- The persisted `tokenjuice_savings.json` is additive; historical buckets keyed by
  the old default remain.

Fixes #4122

## Summary by CodeRabbit

* **New Features**
  * Added turn-scoped model context so savings/cost attribution is calculated against the model active for each agent and sub-agent turn.

* **Bug Fixes**
  * Refactored savings aggregation to record compaction impact using the correct per-turn model (with automatic fallback to the configured default when no turn model is set, ignoring blank scoped models).
  * Updated inner loop execution to compute returned text and iteration stats with the new turn-scoped model handling.

* **Tests**
  * Added/extended coverage for scoped model precedence, default fallback, and blank scoped model behavior.


M3gA-Mind (approved):
Reviewed and verified: correct, well-tested, CI green. Approving. Thanks for the contribution!

PR #4286, opened 2026-06-29
fix(api): route backend calls off BYO cloud inference host (#4285)

oxoxDev:
## Summary

- Backend domain calls (team usage, billing, auth, …) no longer route to a user's BYO **inference** host.
- `effective_backend_api_url` now treats a `config.api_url` whose host is a known cloud inference provider as non-backend and falls back to the default backend chain — the cloud analogue of the existing local-AI (Ollama) guard.
- New registry-derived helper `host_is_builtin_cloud_provider` (in `cloud_providers.rs`) keeps the host set in sync with `BUILTIN_CLOUD_PROVIDERS`.

## Problem

`config.api_url` is overloaded: it is the chat/inference endpoint, but `effective_backend_api_url` (`src/api/config.rs`) also reuses it as the OpenHuman backend base. Its only guard (`looks_like_local_ai_endpoint`) catches local runners and full `/v1/chat/completions` paths, but a **public** provider set to its canonical base (`https://openrouter.ai/api/v1`, `https://api.openai.com/v1`) matches neither signal. The override was then used as the backend base, so every domain call hit the inference host and 400'd — `GET /teams/me/usage` 400 Bad Request against `openrouter.ai`, 4932 events, escalating. Same class as the Ollama fix (OPENHUMAN-TAURI-51/-80/-7Z), never covered for cloud BYO providers.

## Solution

- Add `is_cloud_inference` (host ∈ `BUILTIN_CLOUD_PROVIDERS`) alongside `is_local_ai`; pass-through gate becomes `(!is_local_ai && !is_cloud_inference) || is_openhuman`.
- A user who deliberately set `api_url` to an OpenHuman backend host (`api.tinyhumans.ai`) still passes through (`is_openhuman` short-circuit).
- **Host-based, not path-based**, on purpose: the `looks_like_local_ai_endpoint` doc deliberately refuses to match a bare `/v1` so genuine self-hosted backends at `https://my-backend/v1` keep their pass-through. Only *known* inference hosts are rerouted.
- Inference routing (`effective_api_url` / `effective_inference_url`) is untouched — BYO chat still works.
- One-shot `warn!` (reused `warn_backend_url_fallback_once`) surfaces the fallback in logs.

## Submission Checklist

[...]

PR #4298, opened 2026-06-29
fix(api): stop /teams/me/usage flooding from BYO-provider control-plane misroute (#4153)

### LINKED ISSUE #4153: /teams/me/usage non-2xx flood: 500 (TAURI-RUST-BSF) + 404 (TAURI-RUST-8C) — RCA gated on #4058 host telemetry
## Source

Sentry self-hosted (`tinyhumans` / `tauri-rust`):
- **TAURI-RUST-BSF** — `GET /teams/me/usage failed (500 Internal Server Error); response_body_len=0` — 2063 events · 24 users · first seen 2026-06-09 · last seen 2026-06-25 · ongoing.
- **TAURI-RUST-8C** — `GET /teams/me/usage failed (404 Not Found); response_body_len=91` — 12,424 events · 1 user · ongoing.

Sibling of TAURI-RUST-8WY (401, already classified as session-expiry and suppressed via #3297/#3384).

## Why this is blocked (not yet fixable)

Both issues fire from `team_get_usage` at `src/core/jsonrpc.rs:171` (`report_error_or_expected`). The 500/404 are not session-expiry and not matched by `is_transient_message_failure`, so they land in the reported `else` arm — correct given current signal.

**We are blind on origin:**
- The canonical Node backend (`backend-nodejs` Sentry project) has **no matching `/teams/me/usage` 500**. A 500 with an **empty body** is not what the Node app emits (it returns a JSON error body) — it points at a gateway / proxy / non-canonical (foreign) `BACKEND_URL`, the same theory as 8C's 91-byte non-canonical 404.
- #4058 (merged 2026-06-25) added the missing `host` + `body_preview` tags to the `authed_json` non-2xx arm — but it has not reached a shipped release yet, so current BSF/8C events still lack those tags. Until a build with #4058 ships, we cannot confirm foreign-host vs canonical-backend defect, and therefore cannot bucket per the RCA-vs-suppress gate.

This issue exists so the deferred follow-up #4058 promised does not rot (it was never opened because the telemetry hasn't shipped).

## Plan (gated)

1. **Wait for a release carrying #4058**, then re-pull BSF + 8C events for `host` + `body_preview`.
2. **Bucket each** (real backend defect vs genuinely-unpreventable foreign-host):
   - Foreign host / gateway 5xx with empty body and zero local lever → demote via `is_transient_message_failure` (or an explicit expected arm), and say so + why.
   - Canonical backend 500 → fix the backend (`tinyhumansai/backend`), not the client.
3. **Flood-prevention (host-agnostic, ships regardless of bucket):** add backoff on repeated `team_get_usage` failures so a persistent non-2xx does not re-fire on every poll / managed-tool budget probe. Backpressure, not silent drop — the first failure still reports.

## Bug shape

Backend non-2xx on `/teams/me/usage` budget/usage probe; empty/short body; high-volume re-fire from polling + budget-gate re-probe (`managed_tool_budget_exhausted` does not negative-cache failed probes).

## Related

- #4055 / #4058 — host + body_preview telemetry (the unblock)
- TAURI-RUST-8WY — 401 sibling (already suppressed)

oxoxDev:
## Summary

- `config.api_url` doubles as the BYO inference base, so a user who pointed it at a managed provider was silently sending every **control-plane** call (`/teams/me/usage`, `/teams/*`, billing, referral) to that provider host, which answers `400/404/500`. This is the bulk of the `/teams/me/usage` non-2xx flood.
- Widen `effective_backend_api_url` to fall back to the canonical backend for **remote** managed inference providers — it already did this for *local* model runners (the Ollama fix, `OPENHUMAN-TAURI-51/-80/-7Z`).
- Add a 60s failure-backoff to `team::get_usage` so a *persistent* non-2xx no longer re-fires (and re-reports to Sentry) on every poll and every managed-tool budget probe. First failure still reports; repeats are demoted.
- Inference routing is untouched; the actual bug is the control-plane misroute.

## Problem

`/teams/me/usage` produces a high-volume non-2xx flood across a whole family of Sentry issues (`TAURI-RUST-BSF` 500, `-8C`/`-JJ5` 404, `-HDS`/`-HW1` 400 — tens of thousands of events combined). The `host` tag added in #4058 (now shipped in `v0.58.0`) pinned the destinations to **inference-provider hosts** (`openrouter.ai`, `api.openmodel.ai`), not the OpenHuman backend.

Root cause: `config.api_url` is overloaded — it is both the BYO inference base (`effective_inference_url`) and the control-plane base (`effective_backend_api_url`). When a user configures a managed provider, that URL lands in `api_url`, and every control-plane request is sent to the provider, which has no such route. The existing guard only skipped **local** AI endpoints, so remote providers slipped through.

The canonical backend is not the source: its error handler always emits a JSON body (an empty-body 500 is impossible), it has no matching `/teams/me/usage` 500 in its own telemetry, and the route is registered ahead of `/:teamId` so it never 404s the path.

## Solution

- **Stop the misroute at the resolution source** (`src/api/config.rs`): new `looks_like_inference_provider_endpoint` (known managed-provider hosts + OpenAI-compatible `/v1` base shape); `effective_backend_api_url` now falls back to the canonical backend for both local *and* remote inference endpoints. Our own hosted backend short-circuits, so a deliberate `api.tinyhumans.ai` override still routes. Inference resolution (`effective_api_url`) never consults this guard, so BYO inference is unaffected.
- **Flood backpressure** (`src/openhuman/team/ops.rs`): a process-global 60s failure window. The first failure of a streak probes the backend and propagates its real error (reports normally); repeats inside the window short-circuit without a backend call and return a demote sentinel. Session-expiry never enters the window (its own RPC arm drives auth recovery); any success clears it.
- **Demote the repeats** (`src/core/observability.rs`, `src/core/jsonrpc.rs`): a path-anchored sentinel + classifier, matched by a new arm in `invoke_method_inner` that logs at debug instead of reporting. The classifier matches only the sentinel, never a real backend error string.

This is RCA-first: the headline change prevents the misroute; the backoff/demote is defense-in-depth and never silences the first failure of a streak, so a genuine backend regression still pages.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — verified: Rust Core Coverage CI gate (cargo-llvm-cov) passed on this head; every new branch (URL classification, fallback, suppression, backoff recovery) has unit tests.
- [x] Coverage matrix updated — `N/A: behaviour/bug-fix change, no new feature row.`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix feature touched.`
- [x] No new external network dependencies introduced (no new hosts/deps; narrows existing routing)
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: backend URL resolution + telemetry, no release-cut UI surface.`

## Impact

- **Desktop (all platforms).** Users whose `config.api_url` is a managed inference provider now reach the canonical backend for control-plane calls instead of 400/404-ing against the provider — team usage, billing, and referral start working for them.
- **No inference behaviour change** — inference still resolves through `effective_api_url`/`effective_inference_url`.
- **Sentry volume.** The misroute fix removes the cause for upgraded users; the backoff caps residual/transient cases (e.g. a real gateway blip on the canonical host) to ~1 event/min/process while still surfacing the first failure.
- No migration, no schema change, no new network dependency.

## Related

- Closes: #4153
- Builds on: #4058 (host + body_preview telemetry — the unblock)
- Sentry: `TAURI-RUST-BSF`, `-8C`, `-HDS`, `-HW1`, `-JJ5` (left OPEN intentionally — this PR does not fix-flip them; verify the event drop on a build that ships this)
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/4153-usage-control-plane-misroute
- Model/agent: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented managed inference-provider endpoints from being treated as control-plane URLs when overriding the backend API URL, so backend resolution correctly falls back (with a one-time warning).
  * Reduced repeated “usage probe” error noise by detecting suppressed backoff repeats, returning a cleaner “Usage temporarily unavailable…” message via the RPC boundary.
  * Added usage-fetch failure rate-limiting with 60s suppression windows, while letting session-expiry errors pass through unchanged.

* **Tests**
  * Added coverage for inference-provider endpoint detection/fallback, suppression sentinel matching, and usage backoff cache control flow (including cache clearing, expiry, and session-expiry bypass).


M3gA-Mind:
Reviewer note (automated triage):

**Fixed in this push** (`b6a765f72`):
- CodeRabbit nitpick on `looks_like_inference_provider_endpoint` — the `/api/v1` arm used `ends_with`, looser than the documented exact-match contract, so a self-hosted backend at `.../svc/api/v1` would be wrongly rerouted. Now exact-matches both `/v1` and `/api/v1`.

**Confirmed valid, left for author** (not safely doable without a local build, and partly design calls):
- `team/ops.rs` — the `USAGE_PROBE_BACKOFF:` sentinel is returned through `Err(...)` and reaches the JSON-RPC client payload as the error message; it should stay out-of-band (e.g. cache + return the original backend error text, signal suppression via structured metadata).
- `team/ops.rs` — the process-global 60s `UsageFailureCache` isn't keyed by backend/session identity, so a failure in one context can suppress probes in another.
- `jsonrpc.rs` transport-purity ask is consistent with the existing `session_expired`/`transient` string-classify arms in `rpc_handler`, so it's a broader refactor than this PR; flagging rather than changing.

**Coordination note:** this overlaps with #4286 (same author) — both widen `effective_backend_api_url` to fall back for remote inference hosts (#4298 also covers TAURI-RUST-HW1, #4286's issue). They will conflict/duplicate if both merge; one should supersede the other (this PR is the superset + adds the backoff/demote backpressure).

M3gA-Mind:
@coderabbitai Update on your review — all four items addressed:

**Fixed — `looks_like_inference_provider_endpoint` path match (`config.rs`, commit `b6a765f72`):** the `/api/v1` arm now uses exact `==` like `/v1`, matching the documented contract so a self-hosted backend at `.../svc/api/v1` is no longer rerouted.

**Fixed — usage-backoff sentinel leaking into the RPC payload (`jsonrpc.rs`, commit `8cb5be70d`):** good catch. The `USAGE_PROBE_BACKOFF:` marker is now stripped at the transport boundary — `rpc_handler` uses it only to drive the Sentry-demotion decision, then replaces `display_message` with a clean, user-presentable string before building the JSON-RPC error. The marker is now strictly out-of-band (your first suggested alternative); the client never sees it.

**Declining — process-global `UsageFailureCache` keying (`team/ops.rs`):** the desktop core runs a single user with a single `config.api_url` / backend / session per process, so the usage probe has exactly one context per process — there is no second concurrent backend/session for the global window to wrongly suppress. Session-expiry (401) is already excluded from the window and keeps driving auth recovery on its own arm. Keying the cache by an identity that is invariant within a process would add complexity with no behavioral change, so I'm leaving it process-global by design.

**Declining — moving the suppression decision into `StructuredRpcError` metadata (`jsonrpc.rs`):** the new usage-backoff arm follows the exact established pattern of its siblings in the same `rpc_handler` chain — `is_session_expired_error`, `is_param_validation_error`, and `is_transient_message_failure` all classify off `display_message`. Migrating that whole demote-classifier family to structured envelope metadata is a worthwhile but broader refactor that should be done uniformly, not bolted onto this one arm in a targeted flood fix. With the leak fixed above, the marker no longer reaches the client, so the practical downside is gone.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
**Maintainer review changes** (commit 97003240a) — addressing the CodeRabbit review:

- `src/openhuman/team/ops.rs`:
  - **Keyed the usage-probe failure backoff by backend URL** (CodeRabbit #3). `UsageFailureCache` now stores `Option<(String, Instant)>` instead of a single process-global `Instant`. `is_fresh`/`record` take a backend `key`; `get_usage` derives it from `effective_backend_api_url(&config.api_url)`. A failure anchored under one backend no longer suppresses probes after `config.api_url` changes (e.g. a fixed `BACKEND_URL`) — the differing key misses the streak and the new route is probed immediately. Session-expiry (401) still bypasses the cache as before.
  - Updated the existing backoff unit tests to pass a `FAIL_KEY` and added `failure_backoff_is_keyed_per_backend` proving a different key doesn't inherit the old backend's backoff.

No other files changed. (Did not run the local cargo/test matrix — GitHub CI verifies.)

M3gA-Mind:
@coderabbitai Disposition of the review findings:

**Fixed (commit 97003240a):**
- **Key the failure cache by backend/session context (team/ops.rs):** done — `UsageFailureCache` is now keyed by the effective backend URL, so a failure on one backend no longer suppresses probes after the backend is re-pointed. Added a key-isolation test.

**Already addressed on current HEAD:**
- **Don't send the observability sentinel back to RPC clients (team/ops.rs:173):** the JSON-RPC boundary (`src/core/jsonrpc.rs`) already replaces the `USAGE_PROBE_BACKOFF:` sentinel with a clean, user-presentable message ("Usage temporarily unavailable … backing off …") before building the response, so the sentinel is kept strictly out-of-band and never reaches the client.
- **Signal-2 path match exact-match (config.rs):** the code already uses `path == "/api/v1" || path == "/v1"` (symmetric exact match) with the rationale comment.

**Declining — push suppression into structured RPC metadata (jsonrpc.rs):**
This is consistent with the **established** transport-boundary demotion pattern in this same file: `rpc_handler` already classifies other demotion families by message via observability helpers — `is_session_expired_error(&display_message)` (jsonrpc.rs:122) and `is_transient_message_failure(&display_message)` (jsonrpc.rs:147). `is_suppressed_usage_probe_backoff` (jsonrpc.rs:124) sits right between them and follows the exact same shape, with the sentinel constant and matcher kept as a single source of truth in `observability.rs` (coupled by a unit test). The business logic (when to back off, the cache) lives in the `team` domain; only the Sentry-demotion *classification* is at the boundary, mirroring its siblings. Reworking just this one family into `StructuredRpcError` metadata would make it inconsistent with `session_expired`/`transient` without changing the actual coupling; a broader refactor of all three demotion families is a separate change, not specific to this PR.

Not running the local cargo/test matrix here — GitHub CI verifies.

M3gA-Mind:
@coderabbitai review

PR #4311, opened 2026-06-29
perf(memory_conversations): rank before cloning hits in cross-thread search

mysma-9403:
## Summary

- `InvertedIndex::search` and `recency_fallback` built a full `CrossThreadHit` (cloning the KB-sized `content` plus four other strings) for **every** candidate, then sorted and truncated to the requested `limit`. Cross-thread conversation search asks for 3–10 results but Phase 2 can leave thousands of candidates, so the vast majority of those clones were thrown away.
- Rank on cheap keys first (match count + a borrowed `created_at`), truncate to `limit`, then materialize the heavy `CrossThreadHit` only for the survivors.
- Behavior is unchanged: ranking by `matched` (usize) is order-equivalent to ranking by `score = matched / total_terms` (a positive constant), and the `created_at desc` tiebreak is preserved.

## Problem

`memory_conversations::inverted_index::search` is on the conversation-recall hot path (agent memory loader + transcript search). After Phase 2 substring verification, `hit_counts` can hold thousands of matching docs. The old code did:

```rust
let mut hits: Vec<CrossThreadHit> = hit_counts.into_iter()
    .map(|(doc_id, matched)| CrossThreadHit {
        content: entry.content.clone(), // KB-sized, cloned for EVERY candidate
        // ...four more String clones
    })
    .collect();
hits.sort_by(/* score desc, created_at desc */);
hits.truncate(limit); // keep only 3–10
```

So for 1,000 candidates × ~10 KB content with `limit = 10`, it cloned ~10 MB and sorted 1,000 heavy structs to keep ~100 KB — ~99% wasted allocation and a sort comparing large structs. `recency_fallback` (the >10k-candidate path) had the same shape over the **entire corpus**, so it could clone hundreds of MB before truncating.

## Solution

Defer the expensive materialization past the truncation point:

- Collect `(doc_id, matched, &created_at)` — the borrow avoids cloning `created_at` just to sort.
- `sort_by` on the borrowed keys (`matched desc`, then `created_at desc`).
- `truncate(limit)`.
- Map the surviving `limit` entries into `CrossThreadHit`, cloning `content`/etc. only there.

`recency_fallback` gets the same treatment via `(doc_index, &created_at)`.

String clones drop from O(n·k) to O(limit·k) (k = content size); the sort now compares ints/`&str` instead of full structs. No public API change, no signature ripple — both functions are self-contained.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — the reworked `search`/`recency_fallback` paths are covered by the existing suite plus the new `ranks_by_score_then_recency_before_truncating` (exercises rank+truncate with more candidates than `limit`); `pathological_query_short_circuits_to_recency` covers the fallback.
- [x] Coverage matrix updated — `N/A: internal perf refactor, no feature row added/removed/renamed`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: internal index helper, no release-cut surface`
- [x] Linked issue closed via `Closes #NNN` — `N/A: no tracking issue; found by code inspection`

## Impact

- Lower allocation and sort cost on the conversation-recall hot path (agent memory recall + transcript search), most pronounced on large corpora / broad queries and on the >10k-candidate recency fallback. Desktop/CLI core. No behavior, API, schema, or migration change.

## Related

- Closes: N/A
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: perf/conv-search-rank-before-clone
- Commit SHA: 7f731357c

### Validation Run

- [x] `pnpm typecheck`
- [x] Focused tests: `cargo test --lib memory_conversations::inverted_index`

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: none (pure perf refactor)
- User-visible effect: identical search results; less allocation/CPU on conversation recall

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search result ordering so higher-quality matches are ranked first, with newer items used as a tie-breaker before results are limited.
  * Refined fallback result selection to return the most recent items more consistently when primary matching is unavailable.
  * Reduced unnecessary work during search result generation, helping keep searches more efficient.

### DIFF
--- a/src/openhuman/memory_conversations/inverted_index.rs
+++ b/src/openhuman/memory_conversations/inverted_index.rs
@@ -273,9 +273,29 @@ impl InvertedIndex {
         }
 
         let total_terms = terms.len() as f64;
-        let mut hits: Vec<CrossThreadHit> = hit_counts
+        // Rank on cheap keys first — the match count and a borrowed `created_at`
+        // — then materialize the heavy CrossThreadHit (which clones the KB-sized
+        // `content`) only for the `limit` survivors. Phase 2 can leave thousands
+        // of candidates in `hit_counts` while callers ask for 3-10 results, so
+        // cloning every candidate's content before truncating is ~99% wasted.
+        // Ranking by `matched` (usize) is order-equivalent to ranking by
+        // `score = matched / total_terms` since `total_terms` is a positive
+        // constant, so the returned order is unchanged.
+        let mut ranked: Vec<(u32, usize, &str)> = hit_counts
             .into_iter()
             .map(|(doc_id, matched)| {
+                let entry = self.docs[doc_id as usize]
+                    .as_ref()
+                    .expect("doc_id from hit_counts must be live");
+                (doc_id, matched, entry.created_at.as_str())
+            })
+            .collect();
+        ranked.sort_by(|a, b| b.1.cmp(&a.1).then_with(|| b.2.cmp(a.2)));
+        ranked.truncate(limit);
+
+        ranked
+            .into_iter()
+            .map(|(doc_id, matched, _)| {
                 let entry = self.docs[doc_id as usize]
                     .as_ref()
                     .expect("doc_id from hit_counts must be live");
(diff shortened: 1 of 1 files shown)


M3gA-Mind (approved):

PR #4319, opened 2026-06-30
perf(meet): keep CEF cache warm on pre-join scanner reload

YellowSnnowmann:
## Summary

- Stop forcing a cold reload during Meet join: the Phase 0 scanner reload no longer clears the HTTP cache and no longer reloads with `ignoreCache`.
- Anonymity is preserved by clearing the leaked Google session **cookies** only — the cache is left warm.
- Extracts the reload params into `pre_join_reload_params()` with a unit test that fails if `ignoreCache` is reintroduced.

## Problem

On every Meet join, `meet_scanner::run` (Phase 0) calls `Network.clearBrowserCookies` + `Network.clearBrowserCache` and then `Page.reload({ignoreCache: true})`. Clearing the cache and bypassing it on reload forces CEF to re-fetch and re-parse Meet's multi-MB SPA bundle from cold on **every** join. A latency audit found this cold reload to be the single dominant contributor to join time.

The cache clear has no privacy benefit: Meet's JS/CSS bundle is identical for every visitor, and because the vendored `tauri-cef` runtime does not yet honour the per-`request_id` `data_directory` as a fresh `RequestContext`, the CEF HTTP cache is already shared (and warm) across joins. Only the **cookies** carry the user's signed-in Google identity that must be dropped so the bot reaches the anonymous "Your name" pre-join input.

## Solution

- Remove the `Network.clearBrowserCache` call.
- Reload with `pre_join_reload_params()` (an empty params object) instead of `{ignoreCache: true}`, so the reload re-uses the already-downloaded bundle.
- Keep `Network.clearBrowserCookies` + the reload so Meet re-evaluates auth and renders the anonymous pre-join page.
- The new `pre_join_reload_params()` helper documents the cache-warm decision; its unit test asserts no `ignoreCache` field, guarding against a regression that reintroduces the cold reload.

Behaviour is otherwise unchanged: cookies are still cleared, the page is still reloaded once, and the join automation phases run exactly as before — just against a warm cache.

## Submission Checklist

> If a section does not apply to this change, mark the item as `N/A` with a one-line reason. Do not delete items.

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — `pre_join_reload_keeps_cache_warm` asserts the cache-warm contract (regression guard).
[...]

PR #4320, opened 2026-06-30
perf(meet): escalating target-attach + tighter scanner poll interval

YellowSnnowmann:
## Summary

- Speed up the first step of Meet join (CDP target discovery) with an escalating retry schedule instead of a flat 500 ms poll.
- Tighten the in-page phase poll interval (`POLL_INTERVAL`) from 500 ms to 150 ms so the scanner acts on a control as soon as it appears.
- Both changes are pure-constant/helper tweaks with unit tests; no behavioural change beyond timing.

> **Stacked on** `feat/meet-warm-scanner-reload` (#4319). Review/merge that first; this branch's base is that PR, not `main`.

## Problem

Two fixed timings sit on the Meet join hot path in `meet_scanner`:

1. `wait_for_meet_target` retried CDP attach on a flat `sleep(500ms)` loop. The CEF renderer surfaces its page target a few hundred ms after the host window builds, so the first attempt nearly always misses and the scanner then burns a full 500 ms before retrying — even though the target usually appears within ~200–400 ms. The account scanner already solved this with an escalating `INITIAL_ATTACH_SCHEDULE` (`cdp/session.rs`); the meet scanner never adopted it.
2. The in-page click/type phases (device-check, name input, mic/camera toggles, ask-to-join) poll a cheap `Runtime.evaluate` every `POLL_INTERVAL` = 500 ms. When a control appears just after a poll, the scanner waits a full 500 ms before acting — repeated across ~6 phases.

## Solution

- Add `INITIAL_TARGET_ATTACH_SCHEDULE` (`[0, 50, 150, 400] ms`) and a pure `target_attach_delay(attempt)` helper that returns the escalating delay first, then falls back to the steady `TARGET_DISCOVERY_INTERVAL` (500 ms). `wait_for_meet_target` now retries immediately, escalates quickly, then steadies — saving ~350–500 ms on the warm path.
- Lower `POLL_INTERVAL` to 150 ms for the pre-join phases. `Runtime.evaluate` snippets are cheap, so the tighter cadence costs negligible CPU while cutting per-phase slack. The post-admission `wait_for_admission` keeps its own coarser 1 s cadence (not on the pre-join hot path) — unchanged.

No control-flow change: the same phases run in the same order against the same budgets; only the inter-attempt delays shrink.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — `target_attach_schedule_escalates_then_steadies` (incl. past-schedule fallback) and `poll_interval_is_tight_but_sane` (regression guard).
- [x] **Diff coverage ≥ 80%** — the new executable lines are the `target_attach_delay` helper + the schedule lookup, both exercised by the new tests. Run `pnpm test:rust` before un-drafting.
- [x] Coverage matrix updated — `N/A: timing-only performance change, no new feature row.`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A.`
- [x] No new external network dependencies introduced — none; timing only.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: join behaviour unchanged, only faster.`

## Impact

- **Platform**: desktop (Tauri CEF) only.
- **Performance**: ~0.4–0.5 s off target discovery (warm path) plus reduced per-phase slack across the join sequence. No behaviour change.

## Related

- Closes:
- Base PR: #4319 (`feat/meet-warm-scanner-reload`)
- Follow-up PR(s)/TODOs: adaptive readiness polls replacing the remaining fixed `sleep(1500ms)`/`sleep(700ms)` settle waits land with the single-page-load restructure (next PR).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/meet-scanner-adaptive-timing`
- Commit SHA: _set on push_

### Validation Run

- [ ] Focused tests: `cargo test -p OpenHuman meet_scanner`
- [x] Rust fmt/check (if changed): `cargo fmt --check` clean.
- [x] Tauri fmt/check (if changed): `cargo check --manifest-path app/src-tauri/Cargo.toml`

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: faster target discovery + tighter pre-join polling.
- User-visible effect: faster Meet join; otherwise identical.

### Parity Contract

- Legacy behavior preserved: same phases, same order, same budgets; only inter-attempt delays shrink.
- Guard/fallback/dispatch parity checks: post-schedule fallback to the 500 ms steady interval preserves the old worst-case cadence past the escalation window.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

### DIFF
--- a/app/src-tauri/src/meet_scanner/mod.rs
+++ b/app/src-tauri/src/meet_scanner/mod.rs
@@ -44,7 +44,34 @@ use crate::cdp::{self, CdpConn};
 /// renderer-side target a few hundred ms after the host-side window is
 /// ready.
 const TARGET_DISCOVERY_BUDGET: Duration = Duration::from_secs(20);
+/// Steady-state poll interval used once the fast escalating
+/// [`INITIAL_TARGET_ATTACH_SCHEDULE`] is exhausted.
 const TARGET_DISCOVERY_INTERVAL: Duration = Duration::from_millis(500);
+
+/// Escalating inter-attempt delays for the first few target-attach tries,
+/// mirroring `cdp::session::INITIAL_ATTACH_SCHEDULE`. The CEF renderer
+/// surfaces its page target a few hundred ms after the host window builds,
+/// so we retry immediately, then escalate quickly, instead of burning a
+/// flat 500 ms between every attempt. On the warm path this shaves
+/// ~350-500 ms off target discovery (the first thing on the join hot path)
+/// versus the old fixed `sleep(500ms)` loop. After the schedule is
+/// exhausted we fall back to [`TARGET_DISCOVERY_INTERVAL`].
+const INITIAL_TARGET_ATTACH_SCHEDULE: [Duration; 4] = [
+    Duration::from_millis(0),
+    Duration::from_millis(50),
+    Duration::from_millis(150),
+    Duration::from_millis(400),
+];
+
+/// Inter-attempt delay before retry number `attempt` (0-based) in
+/// [`wait_for_meet_target`]: the escalating schedule first, then the
+/// steady interval. Pure so the schedule can be unit-tested.
+fn target_attach_delay(attempt: usize) -> Duration {
+    INITIAL_TARGET_ATTACH_SCHEDULE
+        .get(attempt)
+        .copied()
+        .unwrap_or(TARGET_DISCOVERY_INTERVAL)
+}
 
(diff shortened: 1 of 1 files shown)


YellowSnnowmann:

PR #4322, opened 2026-06-30
perf(meet): synthesize pre-roll ack concurrently with the agentic turn

YellowSnnowmann:
## Summary

- Synthesize the "On it." pre-roll ack **concurrently** with the agentic LLM turn instead of awaiting it first.
- The ack TTS (~1–1.5s) no longer serializes in front of the model call; it runs while the LLM works and is drained just before the real reply is enqueued, preserving audio order.
- Extracts the pre-roll decision into a unit-tested `should_preroll()` helper.

## Problem

In `run_caption_turn`, the pre-roll ack was synthesized **before** dispatching the LLM:

```rust
if !was_bare_wake && prompt.chars().count() > PREROLL_SKIP_PROMPT_CHARS {
    if let Ok(ack_pcm) = tts(PREROLL_ACK_PHRASE).await { ... }   // ~1-1.5s, blocking
}
let reply_text = llm_meeting_agentic(&prompt, request_id).await?; // only starts now
```

The `tts(...).await` blocks the LLM dispatch by a full ack synthesis (~1–1.5s) on every real (>50-char) question — pure serial latency in front of an already-slow agentic turn.

## Solution

Start the ack synth as a concurrent task before the LLM call, then drain it after the LLM returns, **before** the real reply is enqueued:

```rust
let preroll_synth = if !was_bare_wake && should_preroll(&prompt) {
    Some(tokio::spawn(tts(PREROLL_ACK_PHRASE)))
} else { None };

let reply_text = llm_meeting_agentic(&prompt, request_id).await?; // runs concurrently with the ack synth

if let Some(handle) = preroll_synth {
    // enqueue the ack (done=false) ahead of the reply — order preserved
}
```

- The ack synth overlaps the LLM turn, so its ~1–1.5s leaves the critical path. The LLM almost always outlasts the ack synth, so the post-LLM `handle.await` rarely blocks; in the rare case the LLM is faster, we wait only the ack-synth remainder we'd have paid anyway — and order is still preserved (ack `done=false` enqueued before the reply `done=true`).
- Best-effort and non-fatal: a failed synth or a cancelled task just means no ack, logged at `debug` (unchanged severity) — never blocks the reply.
- `should_preroll(prompt)` encapsulates the length threshold and is unit-tested (boundary + short/long cases).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — `should_preroll_skips_short_prompts` (incl. exact-threshold edge) and `should_preroll_fires_for_real_questions`.
- [ ] **Diff coverage ≥ 80%** — the pre-roll *decision* (`should_preroll`) is unit-tested; the concurrency plumbing (spawn/await/enqueue) lives inside `run_caption_turn`, which makes live cloud TTS/LLM calls and has no test seam today, so those lines aren't unit-covered. Adding an injectable TTS/LLM seam to cover the turn body is proposed as a follow-up before this un-drafts. Flagging rather than asserting the gate is met.
- [x] Coverage matrix updated — `N/A: latency-only change to existing pre-roll behaviour, no new feature row.`
- [x] No new external network dependencies introduced — none; same single ack TTS call, just concurrent.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: pre-roll behaviour unchanged, only earlier.`

## Impact

- **Platform**: desktop core (meet agent).
- **Performance**: removes ~1–1.5s of serial pre-roll TTS from the wake→reply path on every >50-char question.
- **Behaviour**: identical pre-roll content and ordering; only its synthesis now overlaps the LLM turn.

## Related

- Closes:
- Follow-up PR(s)/TODOs: (1) injectable TTS/LLM test seam to cover `run_caption_turn` body; (2) session-scoped config cache so `stt()`/`tts()` stop reloading config from disk per call; (3) LLM→TTS sentence streaming.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/meet-response-preroll-config`
- Commit SHA: _set on push_

### Validation Run

- [ ] Focused tests: `cargo test --lib should_preroll`

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: pre-roll ack synth overlaps the LLM turn instead of blocking it.
- User-visible effect: faster time to first audible response on real questions.

### Parity Contract

- Legacy behavior preserved: same pre-roll trigger (`>PREROLL_SKIP_PROMPT_CHARS`, not bare-wake), same `done=false` enqueue before the reply, same best-effort skip on failure.
- Guard/fallback/dispatch parity checks: ack still enqueued ahead of the reply; failure/cancel paths log + skip exactly as before.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

### DIFF
--- a/src/openhuman/meet_agent/brain/turns.rs
+++ b/src/openhuman/meet_agent/brain/turns.rs
@@ -32,6 +32,17 @@ pub(super) fn pick_ack_phrase(prompt: &str) -> &'static str {
     ACK_PHRASES[(h as usize) % ACK_PHRASES.len()]
 }
 
+/// Whether a caption turn should play the "On it." pre-roll ack.
+///
+/// Skipped for short prompts — greetings ("hi"), checks ("can you hear
+/// me"), time questions — which the agent answers in 2-5s without tools,
+/// where the ack would be redundant. Real second-brain questions ("am I
+/// free Friday afternoon for a 30 min slot") clear the threshold. The
+/// bare-wake case is handled separately by the caller.
+fn should_preroll(prompt: &str) -> bool {
+    prompt.chars().count() > PREROLL_SKIP_PROMPT_CHARS
+}
+
 /// Fire one brain turn for the named session. Returns `Ok(true)` when a
 /// turn actually ran, `Ok(false)` when the inbound buffer was below the
 /// floor.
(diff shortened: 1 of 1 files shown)


YellowSnnowmann:

PR #4323, opened 2026-06-30
perf(composio): cache curated per-toolkit action catalogue

YellowSnnowmann:
## Summary

- Cache the per-toolkit Composio action catalogue so repeat delegations to an integrations sub-agent don't re-run the multi-second `list_tools` round-trip every turn.
- Adds `fetch_toolkit_actions_cached` (curated output, TTL + connection-change invalidation) and wires it into the sub-agent delegation hot path. The uncached `fetch_toolkit_actions` is unchanged for callers that need a tag filter or guaranteed-fresh results.

## Problem

When any agent delegates to an integrations sub-agent for a toolkit (Slack/Gmail/Calendar), `subagent_runner` calls `fetch_toolkit_actions(client, tk, None)` on **every** delegation. That is a backend `list_tools` HTTP round-trip — ~2–8s for large toolkits (Slack/Gmail) — and it runs before the sub-agent's own model call. In a meeting that asks several integration questions, this multi-second fetch is paid repeatedly for the same toolkit.

A note on a tempting-but-wrong shortcut: the catalogue already loaded into `cached_integration.tools` (from the bulk `fetch_connected_integrations_status` path) is **not** equivalent and must not be substituted directly. `fetch_toolkit_actions` applies curation the bulk path does not — an `action_prefix` filter and `is_action_visible_with_pref` (curated whitelist + per-user scope) at `connected_integrations.rs:1000-1006`, whereas the bulk path stores the raw `resp.tools` (`:606-609`). Swapping in the bulk cache would feed the sub-agent a different, uncurated tool set. So this PR caches the curated fetch rather than replacing it.

## Solution

- New `TOOLKIT_ACTIONS_CACHE` (process-wide `LazyLock<RwLock<HashMap<slug, CachedToolkitActions>>>`), modeled exactly on the existing `INTEGRATIONS_CACHE` next to it.
- `fetch_toolkit_actions_cached(client, toolkit)` returns the **same curated catalogue** as `fetch_toolkit_actions`, but serves repeat lookups for a toolkit from cache within `TOOLKIT_ACTIONS_CACHE_TTL` (300s). Cache miss / stale / lock-contended → live fetch + repopulate. Errors are never cached. `try_read` so a writer never blocks a delegation.
- Invalidation reuses the **existing** fan-in: `invalidate_connected_integrations_cache()` now clears the toolkit cache alongside the integrations cache, so every current trigger (new OAuth connection via `ComposioConnectionCreatedSubscriber`, `composio_list_connections` divergence, tests) also drops stale actions. The 300s TTL only backstops rarer action-scope changes.
- Only the delegation hot path (`subagent_runner/ops/runner.rs`) switches to the cached variant. The debug path (`agent/debug/mod.rs`) keeps calling `fetch_toolkit_actions` directly.

## Blast radius

This is shared agent-harness code: it affects **every** agent that delegates to an integrations sub-agent, not just the meeting bot. The change is a behaviour-preserving memoization — the returned catalogue is byte-for-byte the curated `fetch_toolkit_actions` output; only its freshness changes (≤300s stale, or immediately refreshed on any connection change). No call site sees a different *shape* of data. Worst case: a toolkit action whose visibility scope changed mid-session is not reflected for up to 300s if no connection event fires — acceptable and consistent with the existing 60s integrations cache philosophy.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — `cache_key_normalizes_slug`, `cache_entry_freshness_respects_ttl` (incl. exact-boundary stale case), `invalidate_clears_toolkit_actions_cache`.
- [ ] **Diff coverage ≥ 80%** — the cache key/TTL/invalidation logic is unit-tested; the `fetch_toolkit_actions_cached` read/fetch/write glue wraps a live backend `list_tools` call with no offline seam, so those few I/O lines aren't unit-covered. Flagging rather than asserting; a Composio-client fake to cover the wrapper is a reasonable follow-up before un-drafting.
- [x] Coverage matrix updated — `N/A: caching of existing curated fetch, no new feature row.`
- [x] No new external network dependencies introduced — removes redundant `list_tools` calls; adds none.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: same tool set surfaced to sub-agents, just cached.`

## Impact

- **Platform**: desktop core (all agents that delegate to integrations sub-agents).
- **Performance**: removes the ~2–8s `list_tools` round-trip from every repeat delegation to the same toolkit within a 300s window.
- **Behaviour**: identical curated tool catalogue; only memoized. Invalidated on connection change.
- **Security/Migration**: none — same curation/whitelist applied; per-user scope unchanged (single user per core process).

## Related

- Closes:
- Follow-up PR(s)/TODOs: Composio-client fake to unit-cover the cached wrapper; the parallel-tool-dispatch + agentic-timeout change ships separately.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/meet-composio-toolkit-cache`
- Commit SHA: _set on push_

### Validation Run

- [ ] Focused tests: `cargo test --lib toolkit_actions_cache`

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: repeat per-toolkit action fetches are served from a TTL cache.
- User-visible effect: faster integration replies after the first delegation to a toolkit.

### Parity Contract

- Legacy behavior preserved: cached output is the same curated `fetch_toolkit_actions` result; uncached fn unchanged; debug path unchanged.
- Guard/fallback/dispatch parity checks: cache miss/stale/contended falls through to a live fetch; errors never cached; invalidation reuses existing triggers.

- Duplicate PR(s): none
- Canonical PR: this
- Resolution: N/A

### DIFF
--- a/src/openhuman/agent/harness/subagent_runner/ops/runner.rs
+++ b/src/openhuman/agent/harness/subagent_runner/ops/runner.rs
@@ -356,7 +356,7 @@ async fn run_typed_mode(
             {
                 let fresh_actions = match &client_kind {
                     Some(ComposioClientKind::Backend(client)) => {
-                        match crate::openhuman::composio::fetch_toolkit_actions(client, tk, None)
+                        match crate::openhuman::composio::fetch_toolkit_actions_cached(client, tk)
                             .await
                         {
                             Ok(actions) if !actions.is_empty() => actions,
--- a/src/openhuman/composio/connected_integrations.rs
+++ b/src/openhuman/composio/connected_integrations.rs
@@ -49,6 +49,47 @@ pub(crate) struct CachedIntegrations {
 pub(crate) static INTEGRATIONS_CACHE: LazyLock<RwLock<HashMap<String, CachedIntegrations>>> =
     LazyLock::new(|| RwLock::new(HashMap::new()));
 
+/// TTL for the per-toolkit action catalogue cache (see
+/// [`TOOLKIT_ACTIONS_CACHE`]). Action *definitions* (e.g. the
+/// `GMAIL_SEND_EMAIL` parameter shape + the curated visibility set) change
+/// far less often than connection state, so this is more generous than
+/// [`CACHE_TTL`]. The event-driven invalidation path (a new OAuth
+/// connection clears this alongside the integrations cache) covers the
+/// common "user just connected a toolkit" case; the TTL only backstops
+/// rarer action-scope changes.
+pub(crate) const TOOLKIT_ACTIONS_CACHE_TTL: Duration = Duration::from_secs(300);
+
+/// Cached per-toolkit action catalogue plus the timestamp we wrote it.
+#[derive(Clone)]
+pub(crate) struct CachedToolkitActions {
+    pub(crate) actions: Vec<ConnectedIntegrationTool>,
+    pub(crate) cached_at: Instant,
+}
+
+/// Process-wide cache for the curated per-toolkit action catalogue
+/// returned by [`fetch_toolkit_actions`], keyed by the normalized toolkit
+/// slug. Populated on first delegation to an integrations sub-agent for a
+/// toolkit and reused (within [`TOOLKIT_ACTIONS_CACHE_TTL`]) on subsequent
+/// delegations, so a meeting that asks several Slack/Gmail questions pays
+/// the multi-second `list_tools` round-trip once instead of every turn.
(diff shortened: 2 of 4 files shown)


YellowSnnowmann:

Issue #2437, opened 2026-05-21
Triage request: 5 issues surfaced by remote-core attach-mode investigation (attach-mode silent, daemonlifecycle recurrence, configPersistence printer, cross-host path leak, active-user re-resolution)


leighstillard:
## Summary

While trying to get a Windows desktop client (0.54.0) to attach to a remote `openhuman-core` over a private network, a multi-layer diagnostic surfaced five distinct issues. Filing as a single triage ask so a maintainer can split / prioritize / dedupe before any of them get worked on. I'm volunteering to take #A first if the maintainers concur, but want input before opening PRs.

Diagnostic context: Windows 0.54.0 installed via MSI, launched via `openhuman://auth/?token=...` deep link after web sign-in. Remote Linux `openhuman-core` 0.54.0 running with `OPENHUMAN_CORE_HOST=0.0.0.0 OPENHUMAN_CORE_PORT=4000` on a Tailscale-only host. Bundle id on Windows is `com.openhuman.app`; logs live at `%USERPROFILE%\.openhuman\logs\openhuman.<date>.log` (Rust shell, sparse — only `[app-update]` heartbeats) and `%USERPROFILE%\.openhuman\users\local\cef\chrome_debug.log` (CEF renderer, 44 MB / 31 h session, all real activity).

---

## A. Attach mode is silently inoperative when the app is launched via deep link

**Severity: high — the entire "remote core" capability is unreachable for normal users.**

Attach mode is configured only via env vars (`OPENHUMAN_CORE_REUSE_EXISTING`, `OPENHUMAN_CORE_URL`, `OPENHUMAN_CORE_TOKEN`). On Windows, the standard launch path is the Start Menu shortcut (no `Arguments`, no env) and the deep-link handler (`openhuman://auth/...` → invoked via DCOM/explorer, inherits the explorer.exe environment). Neither path can carry those env vars, so the app falls back to spawning its own in-process core regardless of what the user thinks they configured. There is no UI affordance to point at a remote core URL, no obvious failure mode in the renderer, and the renderer log doesn't reference attach state at all.

Confirmed in the diagnostic: `[Environment]::GetEnvironmentVariables` at User and Machine scope contain none of those names; the Start Menu shortcut has zero arguments and zero env; the running process was launched via deep link from explorer.

[...]

PR #4330, opened 2026-06-30
fix(memory): surface cross-host vault paths on a shared core (#4278)

### LINKED ISSUE #4278: Shared openhuman-core across macOS and Linux frontends causes file path conflicts
## Summary

Running a single `openhuman-core` backend shared between a macOS and a Linux frontend causes file path conflicts — OS-specific absolute paths are stored without per-host scoping and get replayed on the wrong OS, breaking integrations. The Obsidian/memory-tree integration is a confirmed affected surface.

## Problem

When a user connects both a macOS and a Linux frontend to the same `openhuman-core` instance, filesystem paths written by one OS (e.g. `/Users/<name>/.openhuman/…` on macOS) are stored in config/state and later served back to the other OS (Linux), where those paths do not exist. This is most visible with the Obsidian integration — the vault path shown to the user is the wrong OS's absolute path and cannot be opened.

**What happened:** User connected macOS and Linux frontends to the same `openhuman-core`. The Obsidian memory-tree vault path (`/Users/<name>/.openhuman/users/<id>/workspace/memory_tree/content`) was a macOS path shown/used on Linux, causing the integration to fail.

**What was expected:** Filesystem paths should be scoped per host/device. A Linux frontend should resolve and store its own paths; a macOS frontend its own. The core should never replay one host's absolute paths to a different host.

**Is this setup supported?** It is unclear whether sharing one core across multiple OS frontends is an intended use case. If it is not, the app should detect the mismatch and surface a clear error rather than silently corrupting paths.

## Solution (optional)

- Filesystem paths (e.g. `action_dir`, `workspace_dir`, vault paths, memory-tree paths) should be stored in a **per-host bucket** keyed by host ID or OS, never in shared/global config.
- If cross-host path sharing is intentional for some fields, those paths must be validated on receipt and rejected (with a clear error) if they are not accessible on the current host.
- If sharing one core across multiple OS frontends is explicitly not supported, surface a clear error when a second OS attaches rather than silently proceeding with broken paths.

## Acceptance criteria

- [ ] **Path isolation** — Filesystem paths stored by one host are never served as-is to a different host/OS.
- [ ] **Obsidian path correct** — Obsidian vault path shown to each frontend resolves correctly on that frontend's OS.
- [ ] **Unsupported setup surfaced** — If multi-OS shared core is not a supported configuration, the second connecting frontend receives a clear, actionable error.
- [ ] **Regression safety** — Single-host setups (macOS only, Linux only, Windows only) unaffected.
- [ ] **Diff coverage ≥ 80%** — Fix PR meets the changed-lines coverage gate.

## Related

Cross-host path leak triage (Windows/Linux): #2437 (point D — same root cause, different OS pair)
Obsidian memory tree UI: visible in UI/UX audit screenshots

oxoxDev:
## Summary

- A single `openhuman-core` shared by macOS and Linux frontends served OS-specific absolute paths (memory-tree / Obsidian vault) to a frontend on a different OS, which then tried to display/open them locally — breaking the integration silently.
- Core now stamps its own OS (`host_os = std::env::consts::OS`) onto both vault RPC responses so the path's owning host is known in-band.
- Frontend compares the core's `host_os` against this device's OS; on a cross-host mismatch it disables the local file actions (Reveal Folder / Open in Obsidian) and shows a clear banner instead of acting on a path that does not exist on this machine.
- `revealPath` rejects an obviously foreign-OS path (POSIX path on Windows, or `C:\`/UNC on POSIX) with an actionable error rather than a cryptic opener failure.

## Problem

`openhuman-core` owns a single local filesystem; the memory-tree / Obsidian vault physically lives under the **core host's** filesystem. The vault RPCs (`obsidian_vault_status_rpc`, `vault_health_check_rpc`) run core-side, probe the core's disk, and return `content_root_abs` — the core host's absolute path. A frontend attached from a **different OS** then renders that path as a local Obsidian vault path and calls `revealItemInDir` on it, which fails because the path/file is on another machine. Worse, the core-side `exists/readable/writable` flags describe the *core* filesystem and report healthy, so the frontend presented a broken integration as if it worked. (The original `root_path is not a directory` came from the now-removed knowledge-vault path; the live surface is the memory-tree/Obsidian vault.) There is no cloud config sync — the vector is the shared core serving its own OS paths over RPC.

## Solution

- **Core**: add `host_os` to `ObsidianVaultStatusResponse` and `VaultHealthCheckResponse`; populate from `std::env::consts::OS`. Additive and infallible.
- **Frontend helper** `vaultHostMatch.ts`: `resolveVaultHostMatch` compares the core `host_os` to this device's OS via `@tauri-apps/plugin-os`. Unknown OS or an older core that omits `host_os` is treated as **local**, so genuine single-host setups are never affected.
- **UI guard**: `VaultHealthChecklist` and `ObsidianVaultSection` show a banner and disable Reveal Folder / Open in Obsidian on a cross-host mismatch.
- **Defense-in-depth**: `revealPath` throws a clear error for a foreign-OS path.
- **Design note**: a per-host config bucket (issue option 1) was rejected — bucketing the path string cannot make a directory that lives on another machine openable locally, and no cross-host config sync exists to scope. The correct fix is to surface the unsupported local-FS case clearly while leaving everything else (chat, memory queries over RPC) working cross-OS.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — verified locally with `diff-cover` against `upstream/main`: 94% on changed frontend lines (each file ≥ 85%); the two changed Rust lines are covered by `vault_rpcs_report_core_host_os`.
- [x] Coverage matrix updated — `N/A: behavior-guard on the existing memory-vault surface; no new feature ID added.`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no feature ID added.`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: UI guard only; true cross-OS repro needs two machines sharing one core.`

## Impact

- Desktop only. Same-OS (single-host) behavior is unchanged — the guard activates only when the core and frontend OS differ, and an older core without `host_os` is treated as local.
- No migration, no schema change beyond an additive RPC field, no new network dependency.
- Cross-OS shared-core users now get a clear explanation instead of a silently broken Obsidian/Reveal action. The remote vault directory remains genuinely unreachable from a different machine — by design (matches the issue's "surface the unsupported setup" acceptance criterion).

## Related

- Closes: #4278
- Follow-up PR(s)/TODOs: N/A
- Related triage: #2437 (point D — same cross-host path-leak root cause, different OS pair)

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/4278-cross-host-vault-path
- Commit SHA: efe0254ff3c1c593bb30bc396a3fcb8988e2d9f6

### Validation Run

- [x] `pnpm typecheck`
- [x] Focused tests: `vitest run` vaultHostMatch / openUrl / VaultHealthChecklist / ObsidianVaultSection (30 passed); `cargo test --lib openhuman::memory::read_rpc` (38 passed)
- [x] Rust fmt/check (if changed): `cargo fmt --check` + `cargo check` + `cargo clippy --lib` (clean)
- [x] Tauri fmt/check (if changed): `N/A: src-tauri unchanged` (pre-push `rust:check` on the Tauri crate passed)

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` Could not run a true two-machine cross-OS repro locally; covered via unit/component tests mocking the OS mismatch.

### Behavior Changes

- Intended behavior change: cross-host (core OS ≠ frontend OS) disables local vault file actions with an explanatory banner.
- User-visible effect: on a shared multi-OS core, Reveal Folder / Open in Obsidian are disabled with a clear message instead of failing silently.

## Summary by CodeRabbit

* **New Features**
  * Added cross-host awareness for vault access, showing a warning when the vault is on another machine.
  * Updated translations for the new cross-host vault messages in multiple languages.

* **Bug Fixes**
  * Disabled local actions like reveal/open when the vault can’t be accessed from the current device.
  * Prevented deep links from firing for non-local vaults, while preserving normal behavior on the same device.

* **Tests**
  * Added coverage for cross-host vault detection, UI states, and path-opening behavior.


M3gA-Mind (approved):

M3gA-Mind (approved):
Reviewed and verified: code-review items addressed, CI green. Approving. Thanks!

M3gA-Mind (approved):

PR #4333, opened 2026-06-30
feat: interactive chat CLI — openhuman starts a Claude Code-like terminal session

45Harry:
## Summary

Adds an interactive chat REPL to the OpenHuman CLI. Running `openhuman` with
no arguments now starts a terminal chat session — no desktop app needed.

## Changes

- **`src/core/chat_cli.rs`** — new interactive chat module. Uses `dialoguer`
  for readline input, maintains `Agent` across turns for conversation history,
  scopes `AgentTurnOrigin::Cli` so approval gates work correctly.

- **`src/core/cli.rs`** — `openhuman` with no args dispatches to chat REPL
  instead of showing help. `--help` still works for full command listing.

- **`packages/npm/bin/openhuman.js`** — Node.js wrapper auto-starts core server
  if needed, connects via HTTP JSON-RPC, provides readline interface.

- **`packages/npm/README.md`** — full CLI docs: install, usage, comparison.

- **`README.md`** — CLI quick-start section.

## Usage

```bash
# Install on any laptop (requires Node.js 18+)
npm install -g openhuman
openhuman
# Or via Homebrew
brew tap tinyhumansai/core
brew install openhuman
openhuman
Features
- Interactive REPL with conversation history
- /exit, /help commands
- --model and --temp overrides
- Full agent toolset: filesystem, git, bash, web search, MCP
- Same agent as desktop — memory, integrations, model routing

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added a full-screen interactive terminal chat experience to the CLI, including an in-chat command menu.
  * CLI now starts interactive chat by default and supports an explicit `chat` subcommand.
* **Bug Fixes**
  * Improved core availability detection, RPC communication handling, and clearer error messages for failures.
* **Documentation**
  * Updated READMEs with CLI quick-start instructions and expanded npm CLI setup/usage details.
* **Chores**
  * Updated the published CLI package version and metadata.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


M3gA-Mind:
@coderabbitai I pushed fixes for the still-valid Rust findings against current HEAD and want to record the disposition of the rest so nothing is left silently unaddressed:

**Fixed (commit 8d9b1af8c):**
- `tui.rs` — terminal teardown is now guaranteed even when `draw`/`poll`/`read` errors: the event loop runs inside a closure whose result is captured, then `LeaveAlternateScreen` / `disable_raw_mode` / `show_cursor` always run before returning. (addresses the "terminal cleanup can be skipped on early error" finding)
- `chat_cli.rs` — the spawned TUI thread's `Result` is no longer dropped (`let _ = join()` → match on `Ok(Ok)`/`Ok(Err)`/panic), so init/render/input failures surface instead of looking like a clean exit.
- `tui.rs` — `Frame::set_cursor` (deprecated in ratatui 0.29) → `set_cursor_position`.
- `chat_cli.rs` — removed an unused `mut` on `config` (clippy).
- `cli.rs` — general help advertised `openhuman chat [--model <name>]`, but `chat_cli` rejects unknown args; dropped the unsupported `--model` from the help text since model switching isn't implemented yet.

**Already resolved by the refactor:** the "Action::Send placeholder echo" finding — `tui.rs` now sends the message over the channel to the real agent (`run_single` via `AgentTurnOrigin::Cli`), no echo path remains.

**Deferred to the author (legitimate, but design/feature decisions — not a safe in-place reviewer fix):**
- **Command menu is non-functional:** the `/` menu lists `login/logout/status/threads/new/memory/files/config/model`, but selecting one only fills the input and Enter sends it as plain chat text — no dispatch. This is the headline feature and needs the author to either wire a command executor or hide the unimplemented entries. Flagging as needs-author-changes rather than guessing the intended architecture.
- **npm wrapper (`packages/npm/bin/openhuman.js`):** the `rpcCall` timeout, the `null`/non-object RPC-result guard, single-dot `method.replace`, and `chat --help` handling are all valid robustness gaps the author should address; the `README.md` one-shot example also needs the method name corrected (`agent.chat`, not `openhuman.agent_chat`).
- The `packages/npm/package.json` version bump to `0.58.2` looks like release-managed metadata that probably shouldn't ride in this PR.

I'm not running the local cargo/test matrix; GitHub CI will verify the Rust changes.

M3gA-Mind:
@coderabbitai Declining the "wrong method name" finding on `packages/npm/bin/openhuman.js` (the `rpcCall('agent.chat', …)` → `openhuman.agent_chat`) — it's a false positive.

`openhuman.agent_chat` IS a registered RPC method. The controller is defined in `src/openhuman/agent/schemas.rs:63` as `namespace: "agent", function: "chat"`, which maps to the `/rpc` method `openhuman.agent_chat` (`<namespace>_<function>`), and it's wired into the registry via `crate::openhuman::agent::all_agent_registered_controllers()` / `all_agent_controller_schemas()` (`src/core/all.rs:128,368`). Its declared inputs are `message`, `model_override`, `temperature`, `thread_id`, so the current `{ message, model_override }` payload is valid. `openhuman.inference_agent_chat` is a *different* (inference-namespace) controller, not a required substitute.

`temperature`/`thread_id` are optional and only matter for stateful/threaded chat, which this one-shot REPL doesn't use — omitting them is fine, not a correctness bug.

The other items from your reviews are addressed on HEAD (commit 187c6a1ca, plus 8d9b1af8c):
- `openhuman.js` `rpcCall` now sets a request timeout (destroys + rejects) and replaces all `.` separators (global regex), guards a `null`/non-object RPC result, and `chat --help` prints help instead of booting the REPL.
- `README.md` one-shot example fixed to `--method agent.chat` (the dotted form `invoke_method` expects for the `call` subcommand; the previous `openhuman.agent_chat` was wrong there).
- `tui.rs` no longer leaks unrecognized `/login`/`/threads`/etc. to the model as chat text — they surface a local "isn't available yet" notice; `/exit`/`/quit` still quit.
- Earlier: terminal teardown guaranteed on error, `run_tui` Result surfaced, `set_cursor_position` (ratatui 0.29), unused `mut` removed, misleading `--model` help dropped.

Not running the local cargo/test matrix here — GitHub CI verifies.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai Update on the latest review — fixed on HEAD (commit 0ffcec849, plus 187c6a1ca / 8d9b1af8c):

- **chat_cli.rs — init before TUI (48-68):** config + agent are now initialized *before* the TUI thread is spawned. On failure we return a normal error on the standard terminal and exit non-zero, instead of bringing up a full-screen UI we'd block on via `join()`.
- **chat_cli.rs — propagate TUI failures (86-90):** `tui_thread.join()` now returns `Err(...)` for both the inner `run_tui` error and a thread panic, so a failed session exits non-zero.
- **tui.rs — UTF-8 cursor (143-167):** `cursor` is now a byte offset kept on a char boundary; insert/backspace/delete/left/right step by `char::len_utf8()`, so multibyte input no longer risks a `String::insert/remove` panic.
- **tui.rs — terminal init cleanup (54-56):** if `EnterAlternateScreen` or `Terminal::new` fails after `enable_raw_mode`, we disable raw mode (and leave the alt screen) before returning the error.
- **tui.rs — command palette (35-47):** trimmed `CMDS` to only the implemented commands (`/help`, `/exit`); `/help` is now wired to print the command list, and other slash input shows an "unknown command" notice rather than being sent to the model as chat text.
- **tui.rs — message tail (render_msgs):** the list now renders the last `inner.height` messages so newer output stays on-screen in long sessions.
- **README.md:** the ASCII-art fence now uses ```text (MD040).

**Declining one item — "move business logic out of `src/core/chat_cli.rs`" (52-81):** this is consistent with the established `*_cli.rs` adapter convention in `src/core/` — `agent_cli`, `memory_cli`, `subconscious_cli`, etc. all load state and orchestrate domain calls (e.g. `Agent::run_single`) directly from the core CLI adapter. `chat_cli` follows that same thin-adapter shape (config load + `Agent::from_config` + a turn loop calling domain APIs); extracting a separate chat service would make it the lone outlier. If the maintainers want a broader refactor of the CLI-adapter layer, that's a separate change rather than something specific to this PR.

Not running the local cargo/test matrix here — GitHub CI verifies.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai Fixed the last finding on HEAD: the Ctrl+C quit arm now precedes the broad `KeyCode::Char(c)` arm in `tui.rs`, so Ctrl+C quits instead of inserting 'c'. Please re-review.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

The latest review's sole actionable comment (move the Ctrl-C arm before the broad `KeyCode::Char(c)` arm in `src/core/tui.rs`) was addressed by the author in commit `d13dc44c9` (now the branch tip). Earlier rounds' findings — including the placeholder-echo and slash-command dispatch — were resolved in the intervening commits. Please re-review the current head so the stale CHANGES_REQUESTED can clear.

senamakel:
wow really cool

senamakel:
Closing in favor of #4856, which is the same work re-submitted from a dedicated branch (`pr-4333`) instead of the fork's `main` — review will continue there. Please also address the CodeRabbit/Codex review comments on #4856 and fill in the PR description template.

PR #4346, opened 2026-06-30
ci(coverage): raise RUST_MIN_STACK so llvm-cov coverage runs don't stack-overflow

graycyrus:
## Summary

The **Rust Core Coverage (cargo-llvm-cov)** lane is currently red on `main` — it SIGABRTs with a **stack overflow**:

```
thread 'openhuman::cron::scheduler::tests::run_agent_job_returns_error_without_provider_key' has overflowed its stack
```

That test (added by #4166) drives the full agent harness. Under `cargo-llvm-cov`, coverage **instrumentation inflates per-frame stack usage**, which blows libtest's default ~2 MB per-test-thread stack and aborts the *entire* coverage run — so the lane fails repo-wide, not just on one PR.

## Fix

Set `RUST_MIN_STACK=67108864` (64 MB) on both the **Rust Core Coverage** and **Rust Tauri Coverage** jobs, so libtest's per-test threads get enough headroom under instrumentation. It's a virtual address reservation only (no real memory cost), and **CI-only — no source or test changes**.

## Why this approach

- The overflow is an artifact of coverage instrumentation, not a real bug in the test or `run_agent_job`, so bumping the stack for the instrumented run is the correct, minimal fix (vs. refactoring the harness or the test).
- Applying it to both coverage jobs prevents the same class of failure on the Tauri shell.

## Test plan
- [x] Workflow YAML validates
CI verification: this PR's Rust Core Coverage lane should no longer stack-overflow (watching this run).

Surfaced while landing the Meetings redesign (#4308); that PR's own lanes were green — this unrelated coverage breakage was blocking the gate.

## Summary by CodeRabbit

* **Chores**
  * Updated coverage job settings to allow larger thread stack sizes during test coverage runs, helping prevent stack-related failures and improving CI reliability.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -336,7 +336,11 @@ jobs:
     needs: [changes, rust-quality]
     if: always() && needs.changes.outputs['rust-core'] == 'true' && needs['rust-quality'].result == 'success'
     runs-on: ubuntu-22.04
-    timeout-minutes: 30
+    # 30 min was only ever enough because a stack overflow aborted the run early
+    # (see RUST_MIN_STACK below). With the overflow fixed the job runs the full
+    # instrumented suite (~13.6k tests, single-threaded build) end-to-end, which
+    # needs more wall-time.
+    timeout-minutes: 50
     container:
       image: ghcr.io/tinyhumansai/openhuman_ci:rust-1.93.0
     env:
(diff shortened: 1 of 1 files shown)


M3gA-Mind (approved):
Reviewed and verified: green CI, correct. Approving. Thanks!

PR #4366, opened 2026-07-01
fix(memory): classify 'No backend session' embed error as AuthMissing

### LINKED ISSUE #4359: Cloud-embedding "No backend session" error is classified Transient instead of AuthMissing (Memory Tree shows "will retry", never prompts login)
## Summary

`classify_embed_error_str` does not recognize the cloud embedder's "No backend session for cloud embeddings" bail, so it falls through to the default `FailureCode::Transient`. The Memory Tree then shows "A temporary error… will retry automatically" and the worker keeps retrying an unrecoverable auth failure, even though an `AuthMissing` → `auth_missing` ("log in to OpenHuman") classification already exists for exactly this case.

## Problem

When the user is not logged in to the backend, `OpenHumanCloudEmbedding::resolve_bearer()` bails **before any HTTP call** with:

```
No backend session for cloud embeddings: log in to OpenHuman, or set
memory.embedding_provider to "ollama" / "none" in config.toml
```

(`src/openhuman/embeddings/cloud.rs:89-92`)

That string flows into `classify_embed_error_str` (`src/openhuman/memory_tree/health/mod.rs:228`). It is not "refusing empty/whitespace", not a dim mismatch, contains none of "budget" / "quota" / "payment required", and has no `(<status>)` HTTP shape (client-side bail, so `parse_http_status` returns `None`). So it hits the final fallthrough **`FailureCode::Transient`** at `mod.rs:293`.

Consequences:

1. **Wrong banner.** The Memory Tree renders `memory.health.remediation.transient` = "A temporary error interrupted memory processing. It will retry automatically." The actual remedy is to log in. The banner tells the user the opposite of the truth.
2. **Futile retries.** `FailureCode::Transient.class()` = `FailureClass::Transient` (`mod.rs:135`), so the job is retried with backoff. An auth-missing failure can never succeed on retry without a login, so this is wasted work and the failed count climbs.
3. **The right code already exists and is simply never reached.** `FailureCode::AuthMissing` maps to the `auth_missing` remediation (`mod.rs:145`) and to `FailureClass::Unrecoverable` (`mod.rs:136`).

This is the same "client-side bail before HTTP wrongly becomes Transient" trap that is already special-cased for empty input (see the `#13021` comment at `mod.rs:231-243`). The "No backend session" case was missed.

**Steps to reproduce:**

1. Be signed out of the OpenHuman backend (Connection Mode Local, no session); Embeddings provider = Managed (OpenHuman).
2. Have a source (e.g. Gmail) with chunks to embed.
3. Memory Tree shows "Error — N failed jobs", but the banner says "A temporary error… will retry automatically", and the jobs keep retrying without ever prompting login.

**Real-world report:** Discord support ticket, user **dejjem** on **v0.57.44** — 169 failed / 1,249 pending; his Embeddings "Test connection" screenshot shows the exact bail string, and the Memory Tree banner shows the misleading "temporary error… will retry automatically" copy.

**Version / platform:** present on current `main` (post-v0.58.0), desktop. The classifier was **not** changed by #3858, which only added the Embeddings-settings sign-in prompt, not this Memory-Tree health classification.

## Solution (suspected)

In `classify_embed_error_str`, add an early match: if the message contains "no backend session" (and/or "log in to openhuman"), return `FailureCode::AuthMissing`. That routes it to the `auth_missing` remediation ("log in to OpenHuman") and to `FailureClass::Unrecoverable`, so the worker fails fast instead of retrying. Mirror the same match for any managed STT/TTS path that shares this bail (per #3787, STT/TTS are likely affected too).

## Acceptance criteria

- [ ] The "No backend session" embed error classifies as `AuthMissing` (`auth_missing` remediation, `Unrecoverable` class).
- [ ] The Memory Tree banner shows the "log in to OpenHuman" remediation, not "A temporary error… will retry automatically".
- [ ] The worker fails the job fast instead of retrying an un-authable embed.
- [ ] Unit test in `classify_embed_error_str` covering the "No backend session" string (sibling to the existing empty/whitespace test).

## Related

- #3787 / #3858 — added the Embeddings-settings sign-in prompt (shipped v0.57.53). This issue is the Memory-Tree-health half that #3858 did not cover.
- Discord support ticket: **dejjem** (`dejjem@gmail.com`, v0.57.44).
- Code: `classify_embed_error_str` (`src/openhuman/memory_tree/health/mod.rs:228-293`); `resolve_bearer` (`src/openhuman/embeddings/cloud.rs:89-92`); `FailureCode` (`mod.rs:56-156`).

aqilaziz:
## Problem

`classify_embed_error_str()` does not recognize the cloud embedder's "No backend session for cloud embeddings" bail, so it falls through to the default `FailureCode::Transient`. The Memory Tree then shows "A temporary error… will retry automatically" and the worker keeps retrying an unrecoverable auth failure, even though an `AuthMissing` → "log in to OpenHuman" classification already exists for exactly this case.

`OpenHumanCloudEmbedding::resolve_bearer()` (`src/openhuman/embeddings/cloud.rs:89-92`) bails with this message **before any HTTP call** when the user isn't logged in:

```
No backend session for cloud embeddings: log in to OpenHuman, or set
memory.embedding_provider to "ollama" / "none" in config.toml
```

Like the existing #13021 pre-flight-guard case just above it in the same function, this message carries no `Embedding API error (<status>)` shape, so it falls through every status-parsing branch straight to `Transient`.

Fixes #4359.

## Fix

One new early-return branch, placed after the budget/quota check and before HTTP-status parsing (same position pattern as the existing #13021 `EmptyInputRefused` branch): match `"no backend session"` case-insensitively and classify as `FailureCode::AuthMissing`.

## Tests

Two new unit tests following the exact pattern of the adjacent #13021 `EmptyInputRefused` tests:
- `classify_no_backend_session_as_auth_missing` — direct string match
- `classify_no_backend_session_through_anyhow_context_chain` — survives `.context()` wrapping, matching how `resolve_bearer`'s bail actually flows up through the embed pipeline in production

## ⚠️ Honest disclosure on verification

I don't have a Rust toolchain available in my environment (rustup install is blocked by network egress rules, no cargo/rustc present), so **I was not able to run `cargo test` for this change.**

What I did verify instead:
- Reproduced the exact string-matching logic (`lower.contains("no backend session")` against the literal bail message) in Python, confirming both the pre-fix fallthrough to `Transient` and the post-fix match to `AuthMissing`, with no keyword collision against the earlier budget/quota/dims/refusing-empty branches for this message.
- The new branch's structure (`if lower.contains(...) { return PipelineFailure::new(...).with_detail(truncate_detail(msg)); }`) is byte-for-byte the same pattern as the three existing branches immediately above it, which already compile and pass in this codebase.
- The two new tests mirror `classify_empty_input_refusal_as_unrecoverable` / `classify_empty_input_refusal_through_anyhow_context_chain` in the same file line-for-line in structure, substituting only the message string and expected `FailureCode`.

Given the size and pattern-matching nature of this change I'm reasonably confident it's correct, but **please let CI/a maintainer confirm the compile and test pass** rather than take my word for it — I'd rather flag this limitation upfront than claim a test result I couldn't actually produce.

## Files changed

```
src/openhuman/memory_tree/health/mod.rs — 1 file, +58 lines (1 fix branch + 2 tests + doc comments)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cloud embedding error classification for the “no backend session” message using an early, case-insensitive match.
  * Authentication-related errors are now reported as unrecoverable (instead of transient).
* **Tests**
  * Added regression tests to verify correct mapping for both direct error strings and errors wrapped with additional context.

### DIFF
--- a/src/openhuman/memory_tree/health/mod.rs
+++ b/src/openhuman/memory_tree/health/mod.rs
@@ -262,6 +262,21 @@ pub fn classify_embed_error_str(msg: &str) -> PipelineFailure {
             .with_detail(truncate_detail(msg));
     }
 
+    // OpenHumanCloudEmbedding::resolve_bearer() bails *before any HTTP call*
+    // when there is no backend session (user not logged in), with the
+    // literal phrase "No backend session for cloud embeddings" (see
+    // src/openhuman/embeddings/cloud.rs). Like the #13021 pre-flight guard
+    // above, this carries no `Embedding API error (<status>)` shape, so
+    // without an explicit match it falls through to `Transient` — the
+    // Memory Tree then shows "temporary error… will retry automatically"
+    // and the worker retries an unrecoverable auth failure forever instead
+    // of surfacing the existing AuthMissing → "log in to OpenHuman"
+    // remediation. Fixes #4359.
+    if lower.contains("no backend session") {
+        return PipelineFailure::new(FailureCode::AuthMissing)
+            .with_detail(truncate_detail(msg));
+    }
+
     // Parse the HTTP status out of the `Embedding API error (<status>): ...`
     // shape. reqwest renders e.g. `402 Payment Required`, so the first
     // 3-digit run after the opening paren is the code.
@@ -720,6 +735,49 @@ mod tests {
             .context("reembed_backfill chunk_id=c");
         let f = classify_embed_error(&wrapped);
         assert_eq!(f.code, FailureCode::EmptyInputRefused);
+        assert!(f.is_unrecoverable());
+    }
+
+    /// Issue #4359: `OpenHumanCloudEmbedding::resolve_bearer()` bails with
+    /// "No backend session for cloud embeddings: log in to OpenHuman, or
+    /// set memory.embedding_provider to ..." *before any HTTP call* when
+    /// the user isn't logged in. This must classify as `AuthMissing`
+    /// (unrecoverable, prompts "log in to OpenHuman") rather than falling
+    /// through to `Transient` ("will retry automatically" — which loops
+    /// forever on an auth failure retrying can never fix).
+    #[test]
+    fn classify_no_backend_session_as_auth_missing() {
+        let msg = "No backend session for cloud embeddings: log in to OpenHuman, or set \
+                    memory.embedding_provider to \"ollama\" / \"none\" in config.toml";
+        let f = classify_embed_error_str(msg);
+        assert_eq!(
+            f.code,
+            FailureCode::AuthMissing,
+            "expected AuthMissing for {msg:?}"
+        );
+        assert!(
+            f.is_unrecoverable(),
+            "AuthMissing must be unrecoverable so the worker stops retrying \
+             and the Memory Tree prompts login instead of showing \
+             'temporary error… will retry automatically'"
+        );
+    }
+
+    /// The match must be case-insensitive and survive anyhow context
+    /// wrapping, matching the same pattern as the #13021 empty-input test
+    /// above — resolve_bearer's bail gets `.context()`-wrapped by callers
+    /// on the way up through the embed pipeline.
+    #[test]
+    fn classify_no_backend_session_through_anyhow_context_chain() {
+        let base = anyhow::anyhow!(
+            "No backend session for cloud embeddings: log in to OpenHuman, or set \
+             memory.embedding_provider to \"ollama\" / \"none\" in config.toml"
+        );
+        let wrapped = base
+            .context("embed summary during seal tree_id=t level=0")
+            .context("reembed_backfill chunk_id=c");
+        let f = classify_embed_error(&wrapped);
+        assert_eq!(f.code, FailureCode::AuthMissing);
         assert!(f.is_unrecoverable());
     }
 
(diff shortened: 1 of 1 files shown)


aqilaziz:
Update: installed `rustfmt` via `apt` and re-verified formatting — the "Rust Quality (fmt, clippy)" CI failure is fixed. `rustfmt --check --edition 2021` flagged one spot (a two-line return that fits under the 100-char width on one line); applied and re-checked clean:

```
$ rustfmt --check --edition 2021 src/openhuman/memory_tree/health/mod.rs
(no output, exit 0)
```

I still can't run the full test suite or `cargo clippy` — this repo's dependency tree requires `edition2024`, which needs a newer Rust toolchain than what's available via `apt` in my sandbox (1.75) — but at least the formatting-specific failure from the earlier push is now genuinely fixed rather than guessed at.

The separate E2E (Playwright) failure on this PR doesn't touch anything in my diff (pure Rust backend change, no frontend/UI code), so that one may be unrelated flakiness — happy to investigate further if it's still failing after a rerun.

M3gA-Mind:
Thanks for jumping on this one!

This is the same fix as #4699, which has already been merged into `main` — it reclassifies the cloud-embedding "No backend session" error as `AuthMissing` for the same issue (#4359), and its test coverage is a superset of what's here. Since the change is already upstream, we'll close this as a duplicate.

Please don't let this discourage you — it was a correct and well-targeted fix, just beaten to the merge. Thanks again, and we'd love to see more contributions from you. 🙏

Issue #3486, opened 2026-06-08
Add switchable agent profiles with isolated persona context


Al629176:
## Summary

Implement agent profiles so users can switch between different agent personas, such as marketing, product manager, or coder, with each profile carrying its own context and working setup.

## Problem

Today OpenHuman behaves like one shared agent identity. Users who work across different roles or workflows need separate agent modes with different memory, connected accounts, skills, and MCP servers, without manually reconfiguring the app each time.

This hurts users who want clear boundaries between work contexts, for example keeping a marketing agent focused on campaign memory and tools while a coder agent uses development connectors and coding skills.

Constraints:
- Profile-specific state should be clearly separated where it affects agent behavior.
- Shared product surfaces such as Compose and the global skills catalog should remain global unless explicitly scoped later.
- Switching profiles should be obvious, fast, and safe.
- Privacy and data boundaries must be clear to users.

## Solution (optional)

Add a profile model that scopes agent persona, memory, connectors, enabled skills, MCP servers, and default routing preferences. Keep shared components like Compose and the global skills catalog available across profiles, while allowing each profile to choose which skills/connectors/MCP servers are active for that persona.

Possible profile examples:
- Marketing
- Product manager
- Coder
- Personal assistant
- Custom profile

## Acceptance criteria

- [ ] **Profile creation** — Users can create, rename, delete, and switch agent profiles.
- [ ] **Persona configuration** — Each profile can define persona/instructions, default behavior, and role-specific preferences.
- [ ] **Memory isolation** — Each profile has its own memory/context boundary, with clear behavior for shared/global memory if supported.
- [ ] **Connector scoping** — Each profile can enable or disable connected apps independently.
- [ ] **Skills and MCP scoping** — Each profile can choose active skills and MCP servers independently, while shared skill definitions remain global.
- [ ] **Shared surfaces preserved** — Compose and other intentionally global components continue to work across profiles.
- [ ] **Safe switching** — Switching profiles updates active memory, connector, skill, and MCP context without leaking stale state from the previous profile.
- [ ] **UX clarity** — The UI makes the active profile visible and explains what is profile-specific vs global.
- [ ] **Regression safety** — Unit, integration, or E2E coverage verifies profile switching, scoped memory/connectors, and global shared components.
[...]

PR #4382, opened 2026-07-01
feat: show custom profiles in chat header profile selector

salarkhannn:
## Summary
The chat header profile selector now surfaces custom (non-built-in) agent
profiles, so users can switch to any profile available on their account — not
just the built-in **Default** and **Reasoning** options.

Custom profiles are those with `builtIn: false`. They are listed after the two
built-in options, sorted by `sortOrder` (fallback `0`) and then by localized
name. The active profile is highlighted (`aria-checked`), and selecting one
fires `onSelect(profile.id)`.

> Part of #3486 — this PR covers chat-header visibility of custom profiles.
> Full profile isolation / scoped state remains tracked in #3486.

## Changes
- Extracted the profile-selector UI out of `Conversations` into a dedicated
  `AgentProfileSelector` component
  (`app/src/features/conversations/components/AgentProfileSelector.tsx`).
- Added a shared `sortAgentProfiles(profiles, locale)` helper that filters to
  non-built-in profiles and sorts by `sortOrder` then `name.localeCompare(..., locale)`;
  re-exported from `Conversations` so existing tests still resolve it.
- Replaced the flaky full-page render assertion with an isolated unit test
  (`AgentProfileSelector.test.tsx`) covering: custom profiles render with
  `profile.name`, the active profile is marked `aria-checked`, and selecting it
  invokes the handler.

## Test plan
- `yarn test` / `vitest run` — new `AgentProfileSelector.test.tsx` passes
  (i18n mocked).
- `yarn format:check` and `yarn typecheck` pass (Frontend Checks / compile lanes).
- Manual: open a conversation → open the profile selector in the chat header →
  confirm custom profiles appear, the active one is highlighted, and selecting
  it updates the active profile.

## Linked issue
Part of #3486


M3gA-Mind:
Maintainer review changes: merged current `upstream/main` into the PR branch and resolved the resulting conflicts. Upstream moved `app/src/pages/Conversations.tsx` → `app/src/features/conversations/Conversations.tsx`, which collided with this PR's edits to the old path.

- **app/src/features/conversations/Conversations.tsx** — resolved the import-block conflict: kept main's new relative-path depth (`../../`) and re-added this PR's `import type { AgentProfile } from '../../types/agentProfile'` at the corrected path. The PR's other two hunks (the `sortAgentProfiles` helper and the per-profile header buttons) applied cleanly to the moved file; all referenced symbols (`agentProfiles`, `selectedAgentProfileId`, `uiLocale`, `handleSelectAgentProfile`) exist there.
- **app/src/pages/__tests__/Conversations.test.tsx** — resolved the test import conflict: import from the new module path `../../features/conversations/Conversations` while preserving this PR's added `sortAgentProfiles` import and its test block.

No functional changes to the feature itself; this only re-homes the PR onto the relocated file. Letting CI verify.

M3gA-Mind:
Re: CodeRabbit nitpicks (both non-blocking `COMMENTED`, not a `CHANGES_REQUESTED` block):

1. **"Sort doesn't use the app's locale"** — already resolved in the current revision: `sortAgentProfiles(profiles, locale)` passes `uiLocale` through and calls `a.name.localeCompare(b.name, locale)`. The comment referenced an earlier draft. No change needed.

2. **"Memoize the sort result"** — declining as an optional micro-optimization (CodeRabbit itself rated it *Trivial / Low value*). The custom-profile list is a handful of entries, so the inline filter+sort per render is negligible, and adding a `useMemo` in this very large component isn't worth the churn for this PR. Can be revisited if the profile count ever grows materially.

M3gA-Mind:
Maintainer review pass (review only — no approval, and nothing pushed to your branch). Checked against current `main` (`fa044d388`).

**The gap you're closing is still real.** `Conversations.tsx:2410-2436` on `main` still hard-codes exactly two radios, `'default'` and `'reasoning'`, so a user's custom profiles remain unreachable from the chat header. Every check on this PR is green (10 pass, 8 skipped). Two things are outstanding: a substantial conflict, and three review threads with no replies on any of them.

One correction to your own description, in your favour: it says you "replaced the flaky full-page render assertion with an isolated unit test", but the diff on `app/src/pages/__tests__/Conversations.test.tsx` is `+90/-0` — you removed no coverage. Worth fixing the wording, because as written it invites a reviewer to go looking for a deleted test.

## 1. The conflict is real work, but the target is intact

`Conversations.tsx` has moved a long way since your base (`d320c4382`, 2026-07-16): **53 commits**, `+515/-1135` on that file alone. That is why the merge is dirty, and I won't pretend the rebase is trivial.

The good news is that everything your change actually depends on survived unchanged:

* the two hard-coded radios are still there, same markup, same `data-analytics-id`s, now at `:2410-2436`;
* `uiLocale` is still in scope at `:374` (`useAppSelector(state => state.locale?.current ?? 'en')`), so the `locale` prop you pass still resolves;
* `agentProfiles`, `selectedAgentProfileId` and `handleSelectAgentProfile` are all still there.

So take `main`'s side wholesale on the surrounding drift and re-apply just your swap of the two-button block for `<AgentProfileSelector … />`. Your two new files (`AgentProfileSelector.tsx`, `AgentProfileSelector.test.tsx`) don't conflict at all.

The second conflict, on `app/src/pages/__tests__/Conversations.test.tsx`, is small — `main` added `deriveChatErrorBanner` to the same import block you're adding `sortAgentProfiles` to. Keep both names.

## 2. The Codex P2 on analytics is correct, and I'd act on it

`AgentProfileSelector.tsx:63` renders `data-analytics-id={`chat-header-mode-${profile.id}`}`. I traced it: `analyticsInteractions.ts:80` reads `data-analytics-id` as the **first** source for `controlIdentifier`, which becomes the `control_id` on every `ui_click`. And `profile.id` is user-authored — `types/agentProfile.ts` documents it as the `<id>` in `personalities/<id>/SOUL.md`, i.e. whatever the user typed when they made the profile.

So this ships user-created strings into the analytics stream and gives `control_id` unbounded cardinality. `scrubIdentifier` normalises the shape but cannot make a user-chosen name non-user-chosen. A fixed `data-analytics-id="chat-header-mode-custom"` for every custom entry keeps the click measurable, keeps the two built-ins individually attributable as they are today, and sends nothing the user wrote. Two-character change.

## 3. The two Copilot threads — one deserves a reply, one is already answered

**On `sortOrder`:** treating a missing `sortOrder` as `0` does mean an unordered profile sorts ahead of one explicitly set to `1`. Whether that's wrong depends on what the field means — if it's an optional *priority*, `Number.MAX_SAFE_INTEGER` (sort last) is the more usual reading; if it's a display index where unset means "no preference, alphabetical", `0` is right. I don't think either is obviously correct, and you're the one who's looked at how profiles actually get created. Whichever you pick, a sentence on the thread will close it — right now it just sits there.

**On "no page-level test":** this one is largely answered by the PR as it stands — `AgentProfileSelector.test.tsx` covers render, `aria-checked`, and the select handler, and `Conversations.test.tsx` covers the sorter across five cases. It's worth saying that on the thread rather than leaving it open, since a reviewer scanning thread count won't see the new file. If you want to close it completely, seeding `agentProfiles` in the page test and asserting one custom entry renders is a small addition — but I wouldn't call it required given what's already there.

## Summary

Rebase (§1), one analytics-id change (§2), two thread replies (§3). The design — extracting the selector into its own component with a shared `sortAgentProfiles` helper rather than growing the inline block — is the right shape, and the built-ins-first-then-custom ordering keeps the existing muscle memory intact. Ping me once it's pushed and I'll re-check the lanes.

Issue #4249, opened 2026-06-28
Improve agent harness with LangGraph-style state machine or Polaris-like architecture


senamakel-droid:
### Problem
The current agent harness runs agents in a linear loop — prompt → tool call → result → next prompt. Context is managed through compaction (old turns get summarized and truncated), which is lossy and makes it hard to carry state across long-running agent sessions. The system prompt grows with cross-chat context, active goals, and memory but there's no explicit state-machine architecture to manage transitions, branching, or persistent working state.

### Inspiration

**LangGraph** (LangChain's state graph framework) models agent execution as a directed graph of nodes (states) and edges (transitions). Agents can branch, loop, pause for human input, and resume — all with explicit state that persists across transitions. Key ideas worth adopting:
- Agent execution as a state machine, not a linear loop
- Typed state that persists across turns and sub-agent spawns
- Conditional routing between nodes based on state
- Checkpointing for pause/resume and human-in-the-loop
- Sub-graphs for parallel execution and delegation

**Polaris** (https://github.com/RadiantAILabs/polaris) is another architecture worth studying for inspiration on:
- How it structures multi-agent coordination
- How it handles state management across agent boundaries
- Any novel approaches to context persistence

### What could improve

1. **Explicit state graph** — Replace the implicit "next tool call" loop with a defined state machine where each turn is a state transition
2. **Persistent working state** — Let agents maintain structured state (variables, intermediate results) that survive context compaction
3. **Branching & merging** — Support parallel sub-agent branches that merge results back into the main graph
4. **Checkpointing** — Save agent state so long-running tasks can pause and resume without losing progress
5. **Human-in-the-loop nodes** — Formalize the "request plan review" / "ask for confirmation" patterns as explicit graph nodes
6. **Conditional routing** — Route to different agent types or tools based on current state rather than a flat delegation decision tree

### Why this matters

- Long-running agent sessions lose context quality as compaction kicks in
- Multi-step plans currently rely on the orchestrator remembering steps across turns
- No way to pause a complex agent run and resume later with full context
- Debugging agent behavior is hard without explicit state transitions to trace

### Related
- Current context compaction issues observed in long agent sessions
- Cross-chat context hacks suggest the system is already reaching for better state management

PR #4419, opened 2026-07-02
fix(tinyagents): keep attempted tool name in the timeline for unavailable tools

sanil-23:
## Summary

Fixes a **repo-wide Playwright regression**: the chat tool-call specs (`chat-tool-call-flow`, `chat-multi-tool-round`) fail on every open PR because the tool timeline drops the *attempted* tool name for a tool the agent can't see.

## Problem

The tinyagents harness (#4249) now routes every chat turn. When the model calls a tool that isn't in the agent's visible set — e.g. `web_fetch`, `web_search_tool`, `file_read`, which the E2E specs deliberately force via the mock — `UnknownToolRewriteMiddleware` rewrites the call name to `UNKNOWN_TOOL_SENTINEL` (`__openhuman_unknown_tool__`) *before* the progress event fires.

The observability bridge then emitted `AgentProgress::ToolCallStarted { tool_name: sentinel }` with the **same `call_id`**. The frontend keys tool-timeline rows by `call_id` and overwrites the name on Started (`ChatRuntimeProvider.tsx`), so the real streamed `web_fetch` row was clobbered to the sentinel. `toolTimelineNames(...).some(n => n.includes('web_fetch'))` then returns `false`, and the specs time out after 20s.

The pre-tinyagents engine emitted the real tool name *before* the availability block, so the timeline showed the attempted tool even for a blocked call — hence green before #4249. This now fails on **#4388, #4389, #4390, and #4393**, confirming it's a `main` regression, not any single PR.

## Fix

Skip forwarding the sentinel `ToolStarted` in the observability bridge (`src/openhuman/tinyagents/observability.rs`) via a match guard. The streamed `tool_args_delta` row keeps the attempted tool name, and the sentinel `ToolCompleted` only updates status by `call_id` — so the timeline once again shows the tool the model attempted, matching the pre-#4249 behaviour. No test is weakened.

## Tests

- New `sentinel_tool_started_is_not_forwarded` unit test: a sentinel `ToolStarted` emits **no** `ToolCallStarted`, while a real tool name still forwards.
- Restores the Playwright `chat-tool-call-flow` / `chat-multi-tool-round` specs (validated in CI's web lane).

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — unit test above + the restored Playwright specs.
- [x] **Diff coverage ≥ 80%** — the changed lines (the guard) are exercised by the new unit test.
- [x] Coverage matrix updated — N/A: behaviour-only observability fix, no feature row changed.
- [x] All affected feature IDs listed under `## Related` — N/A: no coverage-matrix feature ID applies.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A.
- [x] Linked issue closed via `Closes #NNN` — N/A: regression fix for #4249; no separate tracking issue.

## Impact

- Runtime impact: agent tool-timeline UI on the tinyagents path. No behaviour change for tools the agent *can* see.
- Unblocks the `E2E (Playwright / web lane)` check across all open PRs.
- No persistence, migration, security, or network changes.

## Related

- Regression introduced by #4249 (tinyagents state-machine harness).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented placeholder tool-start events from appearing in the activity timeline, so only real tool calls are shown.
  * Improved event handling to avoid overwriting or clobbering visible progress updates.
* **Tests**
  * Added regression coverage to ensure placeholder tool events are ignored while valid tool names still appear correctly.

### DIFF
--- a/src/openhuman/tinyagents/observability.rs
+++ b/src/openhuman/tinyagents/observability.rs
@@ -259,7 +259,19 @@ impl EventListener for OpenhumanEventBridge {
             // exactly once per model call; prefer it over `ModelCompleted`'s
             // optional usage to avoid double counting.
             AgentEvent::UsageRecorded { usage } => self.record_usage(usage),
-            AgentEvent::ToolStarted { call_id, tool_name } => {
+            AgentEvent::ToolStarted { call_id, tool_name }
+                if tool_name.as_str() != super::tools::UNKNOWN_TOOL_SENTINEL =>
+            {
+                // Skip the sentinel Started event. When the model calls a tool the
+                // agent can't see, `UnknownToolRewriteMiddleware` rewrites the name
+                // to `UNKNOWN_TOOL_SENTINEL` *before* this event fires. The frontend
+                // keys tool-timeline rows by `call_id` and overwrites the name on
+                // Started, so a real streamed `web_fetch` row would be clobbered to
+                // the sentinel — dropping the attempted tool name from the timeline
+                // (regression vs. the pre-tinyagents engine, which emitted the real
+                // name before the availability block). The streamed
+                // `tool_args_delta` row (carrying the attempted name) survives, and
+                // the sentinel `ToolCompleted` only updates status by `call_id`.
                 let iteration = self.iteration();
                 match &self.scope {
                     None => self.send(AgentProgress::ToolCallStarted {
@@ -359,6 +371,40 @@ mod tests {
 
         let (input, output, _) = bridge.totals();
         assert_eq!((input, output), (100, 40));
+    }
+
+    #[tokio::test]
+    async fn sentinel_tool_started_is_not_forwarded() {
+        // #4249 regression guard: a `ToolStarted` for the unknown-tool sentinel
+        // must NOT emit a `ToolCallStarted`. The frontend keys tool-timeline rows
+        // by `call_id` and overwrites the name on Started, so forwarding the
+        // sentinel would clobber the real streamed row (e.g. `web_fetch`) and drop
+        // the attempted tool from the UI timeline. A real tool name still forwards.
+        let (tx, mut rx) = tokio::sync::mpsc::channel(64);
+        let bridge = OpenhumanEventBridge::new(Some(tx), "mock-model", 10);
+        let sink = EventSink::new();
+        sink.subscribe(bridge.clone());
+
+        sink.emit(AgentEvent::ToolStarted {
+            call_id: "c1".into(),
+            tool_name: crate::openhuman::tinyagents::tools::UNKNOWN_TOOL_SENTINEL.to_string(),
+        });
+        sink.emit(AgentEvent::ToolStarted {
+            call_id: "c2".into(),
+            tool_name: "web_fetch".to_string(),
+        });
+
+        let mut started_names = Vec::new();
+        while let Ok(p) = rx.try_recv() {
+            if let AgentProgress::ToolCallStarted { tool_name, .. } = p {
+                started_names.push(tool_name);
+            }
+        }
+        assert_eq!(
+            started_names,
+            vec!["web_fetch".to_string()],
+            "sentinel Started must be skipped; the real tool name must still forward"
+        );
     }
 }
 


M3gA-Mind (commented):
Reviewed the change on `src/openhuman/tinyagents/observability.rs`. Overall this is a well-scoped, well-explained fix with a good regression guard — the match-guard + `_ => {}` fall-through correctly drops the sentinel `ToolStarted` so it can't clobber the streamed timeline row keyed by `call_id`. The sentinel is a distinctive constant (`__openhuman_unknown_tool__`), so there's no real-tool collision risk, and `humanize_tool_name` is now correctly never applied to the sentinel. I verified the ordering premise: the real name reaches the UI via `ToolCallArgsDelta` (emitted in `tinyagents/model.rs::note_tool_call`) during model streaming, which happens *before* `UnknownToolRewriteMiddleware::before_tool` rewrites the call — so on the streaming path the attempted name genuinely survives.

A few things worth confirming before merge:

**1. Completed-only / non-streaming tool-call path (main concern).** The surviving "real name" row is created by `note_tool_call`, which only fires when the provider streams `ProviderDelta::ToolCallStart` (`model.rs:276`). But `model.rs:412` notes *"Native tool calls always ride on `Completed`."* For a provider/model that delivers the tool call only on the terminal `Completed` item (no streamed start delta), no `ToolCallArgsDelta` row is ever created — so skipping the sentinel `ToolStarted` would leave the unavailable-tool attempt with **no timeline row at all**, which is arguably worse than the pre-fix sentinel-named row. Could you confirm which providers/models actually hit this path in practice? If any do, consider a fallback that still surfaces the attempt (the middleware stashes the real name in `arguments.requested_tool` — `middleware.rs:~785`), rather than dropping it silently.

**2. The sentinel still leaks through `ToolCompleted` (main concern).** The `ToolStarted` arm is now guarded, but the `ToolCompleted` arm (observability.rs:297) is unchanged and still forwards `ToolCallCompleted { tool_name: "__openhuman_unknown_tool__", .. }`. The comment says the frontend "only updates status by `call_id`" on completion — but that's a cross-boundary (TS) contract that isn't asserted anywhere in this Rust-only PR. If the frontend ever reads/renders `tool_name` on completion (status label, analytics, humanize), the raw sentinel string could surface. Skipping the Completed event isn't safe (the row still needs its done-status), so forwarding-for-status is the right call — but please either link the exact frontend handler that guarantees the name is ignored on completion, or add a paired frontend test pinning that contract. As-is, the end-to-end "keep the attempted name" guarantee depends on an unverified frontend assumption.

**3. Test coverage.** `sentinel_tool_started_is_not_forwarded` is a solid guard for the Started path, but it only covers that half. Consider adding: (a) a companion assertion that the sentinel `ToolCompleted` still forwards a `ToolCallCompleted` (documents the status-only contract from concern #2), and ideally (b) an end-to-end ordering test — a real-name `ToolCallArgsDelta` (or `ToolCallStart`) followed by a sentinel `ToolStarted` — so the "the real row survives" guarantee is pinned against future refactors rather than implied. Minor: an empty `tool_name` passes the guard (`"" != sentinel`) and would forward a blank-named Started; that's pre-existing behavior, just noting it isn't affected here.

**Local validation (from `gh pr checkout`):**
- `cargo fmt --check` on the file — clean.
- `cargo clippy --manifest-path Cargo.toml --lib` — no warnings on `observability.rs` (only pre-existing unused-import warnings in unrelated modules).
- `cargo test ... tinyagents::observability` — the crate compiled, but I could not run the tests locally: linking the test binary fails on this Apple-Silicon host with an unrelated toolchain issue (`whisper-rs-sys` Metal symbol `_OBJC_CLASS_$_MTLResidencySetDescriptor` undefined). That's an environment limitation, not a code problem — the PR's Rust compiled fine and reached the link stage; CI is the authoritative gate for the test run.

None of the above is blocking from a code-correctness standpoint for the common streaming path — concerns #1 and #2 are about edge coverage and an unstated frontend contract. Nice, targeted fix.

PR #4430, opened 2026-07-03
feat(orchestration): subconscious split-brain layer (stages 4-8)

senamakel:
## Summary

- Build the **subconscious-orchestration** layer: wrapped Claude Code / Codex sessions talk to the owner agent over tiny.place Signal DMs, and each inbound DM drives one autonomous wake cycle on a single `tinyagents` graph.
- **Wake graph** (stages 4–5): `normalize → frontend(1) → execute → compress → world_diff → frontend(2) → send_dm → context_guard → done`, checkpointed under thread `orchestration:<session_id>`; two-pass Quick-LLM front end + reasoning core with sub-agent spawning, 20:1 compression, append-only world diff, and 80–90% context-eviction guard.
- **Subconscious steering** (stage 6): the offline `SubconsciousEngine` tick reviews compressed history + the cumulative world diff and emits short steering directives injected into later reasoning prompts — no channels, no external effects.
- **RPC + UI** (stage 7): `openhuman.orchestration_*` internal controllers + a rewired `TinyPlaceOrchestrationTab` reading real store classification, live socket updates, and a Master composer.
- **Hardening + observability** (stage 8): scheduler-gate deferral, no-duplicate-DM on failure, `orchestration.status` cursor-lag/last-error, and a log-leak guard.

## Problem

Wrapped coding-agent sessions had no way to be coordinated by the owner agent: stage 3 only ingested their DMs into a store. There was no autonomous reply loop, no reasoning/memory pipeline, no offline reflection, and the Brain tab bucketed raw envelopes with string heuristics rather than real classification.

## Solution

- One `OrchestrationState` flows through a single command-routing graph; every behaviour-bearing node is bundled behind one injected `OrchestrationRuntime` so the graph mechanics are hermetically unit-testable with a stub while production wires the real agents/store/memory.
- The subconscious is a decoupled out-of-band writer (never a graph edge): it stamps a directive into an append-only store with cycle-count expiry; `seed_state` loads the current directive into state at cycle start.
- Renderer-only RPC controllers + an `orchestration:message` socket bridge power the tab; the frontend is a presentational component over a `useOrchestrationChats` hook.
- Design decisions: injected-runtime seam for testability; deterministic `cycle_id` for resume-idempotent store writes; strict 20:1 compression budget with a 200-token floor; loop-continuity backstop on `max_supersteps`.

## Submission Checklist

> If a section does not apply to this change, mark the item as `N/A` with a one-line reason. Do not delete items.

[...]

PR #4399, opened 2026-07-02
feat: finish TinyAgents harness migration (#4249)

senamakel:
## Summary

- Hard-migrates OpenHuman's agent harness onto the `tinyagents` crate as the orchestration/model/context/events/cost/subagent/embeddings library, adapter-first (prove parity, then delete), per `docs/tinyagents-full-migration-plan/`.
- Models: workload routes projected into the crate `ModelRegistry` with capability gating; SDK-owned cross-route fallback + retry classification; tool-call argument streaming via `ToolCallDelta`; one unified `ModelCatalog` snapshot.
- Context/cache & events: prompt-prefix protection + gated response cache; durable `StoreEventJournal` + status store; single typed owner for subagent lifecycle events; summarization provenance.
- Subagents/orchestration: crate-native steering commands + unified recursion authority; boot reconciliation of orphaned detached tasks; durable approval interrupt/resume for the delegation review gate; workspace-isolation events.
- Registry/embeddings/cost: fail-closed `CapabilityRegistry` diagnostics + `agent.registry_snapshot` introspection RPC + agent projection; `EmbeddingModel` adapter + retriever facade (`MemoryLoaded`); reasoning/cache-creation token accounting + embedding cost.
- Deletions landed with parity coverage: `worktree_context.rs` (CWD now from `WorkspaceDescriptor`) and `SqlRunLedgerCheckpointer` (durable graphs use the crate `SqliteCheckpointer`); `reliable.rs` classifiers extracted to `error_classify.rs`.

## Problem

OpenHuman maintained a large bespoke agent harness (retry/fallback, context reducers, event bridge, checkpointer, subagent pipeline, registries) that duplicated capabilities now shipped by `tinyagents` 1.3.x. Issue #4249 tracks converging onto the crate as the single source of truth for orchestration, keeping only product policy (prompts, registry, security/approval, credentials, UX projections), and deleting the superseded OpenHuman files.

## Solution

- Adapter-first per workstream: wire the crate primitive alongside the legacy path, prove behavioral parity, then delete the legacy file and tick `99-deletion-ledger.md`. Deletions that would regress behavior the mock suite cannot validate (e.g. `ReliableProvider`'s model-fallback/key-rotation/streaming-failover for non-turn callers) are explicitly left gated with documented rationale rather than forced.
[...]

PR #4431, opened 2026-07-03
fix(orchestration): app_state runtime-cache regression + orchestration review fixes

sanil-23:
Rebased onto current main (`8aab529de`), which already carries the #4399/#4430 **compile** reconciliation (graph_fn `Some`, `SqliteCheckpointer` migration, MCP catalog) via #4433. This PR is **slimmed to the fixes main still lacks** — no more redundant compile diffs.

## Fixes

- **app_state runtime-cache regression (main is still broken on this).** #4399 raised `RUNTIME_SNAPSHOT_TTL` 2s→10s; the `app_state_snapshot_degrades_runtime_service_status_failures` e2e ages the cache with a 2.1s sleep, so it now reads a **stale** snapshot and never sees the injected failure. Fix: bypass the runtime-snapshot cache (read + write) when `OPENHUMAN_SERVICE_MOCK` is active (a test-only env production already honors). Full `config_auth_app_state_connectivity_e2e` binary: **50 passed**.
- **orchestration/ops:** split steering out of `seed_state` into `apply_cycle_steering`, invoked only *after* the `has_new_work` guard, so no-op wakes don't consume a cycle tick and prematurely expire steering. *(CodeRabbit)*
- **orchestration/store:** `user_version`-gated world-diff migration — drop the legacy non-unique index, de-dupe race rows, create the unique index, reconcile `terminal_state` + test. *(M3gA-Mind + CodeRabbit)*
- **orchestration/graph/state:** agent-scoped `cycle_id` (avoids cross-peer `compressed_history`/`world_diff` collision) + migration-seam note. *(Codex + M3gA-Mind)*
- **orchestration/schemas:** ingest-health `MAX(last_message_at)` excludes pinned master/subconscious windows so manual owner DMs can't mask a stalled pipeline. *(CodeRabbit)*
- **frontend:** orchestration socket listeners via `socketService.on/off` (queues through startup/reconnect) + messages **Retry** → `refresh()`. *(CodeRabbit/Codex)*
- **docs:** `orchestration.md` fenced-block language (markdownlint MD040).

## Validation (local)
`cargo clippy -p openhuman` 0 errors; `orchestration` 49 tests + `config_auth_app_state_connectivity_e2e` binary 50 tests pass; `cargo fmt --check` clean; frontend `vitest` (tab) 8 + `tsc` clean.

## Note
The app_state fix is the one item main urgently needs (its own Rust E2E/Coverage will hit the same regression). Substantive orchestration stages 3–8 are already on main (#4425 + #4430 + #4433 reconciliation); this lands the remaining review-quality fixes on top.


sanil-23:
@coderabbitai review

src/openhuman/orchestration/store.rs
```diff
@@ -72,7 +72,11 @@ const SCHEMA_DDL: &str = "
         timestamp       TEXT NOT NULL
     );
 
-    CREATE INDEX IF NOT EXISTS idx_world_diff_session
+    -- UNIQUE so a race between concurrent `MAX(seq)+1` allocations can never
+    -- persist duplicate `(agent_id, session_id, seq)` timeline rows: the second
+    -- writer hits a constraint error and its cycle retries (idempotent by
+    -- `cycle_id`) rather than silently corrupting the append-only sequence.
+    CREATE UNIQUE INDEX IF NOT EXISTS idx_world_diff_session
         ON world_diff (agent_id, session_id, seq);
```
M3gA-Mind: **The UNIQUE-index change is a silent no-op on every existing database, so the duplicate-timeline race it claims to close stays open in production.**

The schema is applied by `conn.execute_batch(SCHEMA_DDL)` on every connection open (line 109), with no `DROP INDEX`, `PRAGMA user_version`, or migration step anywhere. This keeps the **same index name** and only adds `UNIQUE`:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_world_diff_session ON world_diff (agent_id, session_id, seq);
```

Per SQLite, `CREATE ... IF NOT EXISTS` against a name that already exists is a **complete no-op — the `UNIQUE` qualifier is never reconsidered.** Any DB created before this PR already has a *non-unique* `idx_world_diff_session`, so the upgrade leaves it non-unique. Verified directly: after applying the old then the new DDL, `pragma_index_list('world_diff')` still reports `idx_world_diff_session | unique=0`, and two rows with the same `(agent_id, session_id, seq)` but different `cycle_id` both insert:

```
idx_world_diff_session|0
rows after duplicate insert:|2
```

So the stated protection ("the second writer hits a constraint error and its cycle retries") holds only for freshly-created databases; every already-deployed core stays vulnerable to exactly the duplicate `MAX(seq)+1` race this is meant to fix.

Needs a real migration: `DROP INDEX IF EXISTS idx_world_diff_session;` before recreating (after de-duping any existing rows), or use a new index name.

src/openhuman/orchestration/graph/state.rs
```diff
@@ -98,15 +98,19 @@ impl OrchestrationState {
...
+        let counterpart_agent_id = counterpart_agent_id.into();
+        // Deterministic cycle id: agent + session + the latest seq in this
+        // window. The agent id scopes the key so two linked peers reporting the
+        // same `harness_session_id`/seq don't collide on `compressed_history` /
+        // `world_diff` rows (the store keys sessions by `(agent_id, session_id)`).
+        // A resumed run over the same window recomputes the same id, so the
+        // per-cycle store writes dedupe.
         let latest_seq = messages.iter().map(|m| m.seq).max().unwrap_or(0);
-        let cycle_id = format!("{session_id}#{latest_seq}");
+        let cycle_id = format!("{counterpart_agent_id}#{session_id}#{latest_seq}");
```
M3gA-Mind, line 120: **Changing the `cycle_id` format breaks replay-dedup for any cycle checkpointed across the upgrade, re-appending a duplicate world-diff / compressed-history row.**

`cycle_id` is the PRIMARY KEY of both `world_diff` and `compressed_history`, and idempotent replay works by recomputing the *same* id for a resumed window (`append_world_diff` short-circuits on `SELECT seq ... WHERE cycle_id = ?`). This changes the format from `{session}#{seq}` to `{counterpart}#{session}#{seq}`.

A cycle that crashed mid-graph before advancing its cursor (rows written under the old `h1#0`) will, on its post-upgrade retry, compute `@peer#h1#0`, miss the existing row, and append a second timeline entry (new `seq`) plus a duplicate compressed-history row. The window is narrow (only cycles in-flight exactly at the upgrade boundary) and — compounding the `world_diff` UNIQUE-index issue — the non-unique index won't catch the duplicate `seq`. Low severity, but a concrete correctness gap at the migration seam worth a note or a one-time cleanup.

src/openhuman/orchestration/store.rs
sanil-23: Good catch — you're right, `CREATE UNIQUE INDEX IF NOT EXISTS` under the same name is a no-op on any DB that already has the non-unique index. Fixed in aedbf9fd3 with a real migration in `SCHEMA_DDL`: `DROP INDEX IF EXISTS idx_world_diff_session`, de-dupe pre-existing race rows (keep earliest per key so the UNIQUE build can't fail), then `CREATE UNIQUE INDEX ... idx_world_diff_session_uniq` under a **new name** so it actually applies on already-created stores. All three steps are idempotent under `execute_batch` on every open. Added `migrates_legacy_nonunique_world_diff_index_to_unique` which builds a legacy DB (old non-unique index + two same-`(agent,session,seq)` rows) and asserts post-open the rows are de-duped and a fresh duplicate insert is rejected.

src/openhuman/orchestration/graph/state.rs
sanil-23: Agreed, and thanks — real but narrow (a single in-flight cycle exactly at the upgrade boundary). Documented in aedbf9fd3: the worst case is one extra `world_diff`/`compressed_history` entry, not corruption, and the feature is Beta-gated behind `[orchestration]`, so I've left it as an explicit migration-seam note rather than a one-time cleanup. Happy to add a cursor-based cleanup if you'd prefer belt-and-suspenders.

src/openhuman/orchestration/store.rs
```diff
@@ -72,7 +72,23 @@ const SCHEMA_DDL: &str = "
...
+    -- NON-unique. `CREATE UNIQUE INDEX IF NOT EXISTS` under that same name is a
+    -- no-op on a DB that already has it, so uniqueness would never take on an
+    -- already-created store. Drop the legacy index, de-dupe any pre-existing
+    -- race rows (keep the earliest per key so the UNIQUE build can't fail), then
+    -- create the unique index under a new name so it actually applies. Every
+    -- step is idempotent on subsequent opens (execute_batch runs each connect).
+    DROP INDEX IF EXISTS idx_world_diff_session;
+    DELETE FROM world_diff WHERE rowid NOT IN (
+        SELECT MIN(rowid) FROM world_diff GROUP BY agent_id, session_id, seq
+    );
```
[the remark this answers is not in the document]
sanil-23: Fixed in 628ea69ab. Reworked the whole world-diff migration into a `user_version`-gated one-time `migrate()` (so it doesn't re-run every open): drop legacy index → de-dupe → create UNIQUE index → **reconcile `terminal_state` to the surviving max-seq mutation per (agent, session)** via `INSERT OR REPLACE INTO kv ... WHERE seq = (SELECT MAX(seq) ...)` → bump user_version. Extended the test with distinct mutations (`mut_c1`/`mut_c2`), a stale `terminal_state` set to the to-be-deleted duplicate, and asserts it's reconciled to `mut_c1` plus `user_version = 1`. (This PR is also now rebased onto current main to repair the #4399/#4430 semantic breakage — graph_fn Some() + SqliteCheckpointer migration — since main's Rust Quality is currently red from that.)

src/openhuman/orchestration/graph/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+                // Pass 1: raw traffic → macro-instructions, hand down to the core.
+                let instructions = runtime.frontend_instruct(&s).await.map_err(graph_err)?;
+                tracing::debug!(
+                    target: LOG, session_id = %s.session_id, node = "frontend", pass,
+                    route = "execute", reason = "first_pass", "[orchestration] node.route",
+                );
+                Ok(NodeResult::Command(
+                    Command::default()
+                        .with_update(OrchestrationUpdate::Pass1 { instructions })
+                        .with_goto(["execute"]),
```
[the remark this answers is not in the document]
senamakel: Deferring this as a design decision rather than applying it here. The wake graph is deliberately two-pass: `frontend_instruct` (pass 1) only ever returns macro-instructions and always routes to `execute`; the terminate predicate is a `channel_response` produced on the pass-2 `frontend_compile`. Adding a pass-1 fast-path direct reply would change the `OrchestrationRuntime` trait contract (pass 1 would need to optionally emit a `channel_response`) and the router's goto logic — a spec-level change to the graph shape, not a localized fix. The current design still sends the reply; it just always routes through the reasoning core once. Flagging for @sanil-23 to decide whether the fast-path is in scope for this PR or a follow-up.

PR #4505, opened 2026-07-04
feat: implement Langfuse feedback scores

### LINKED ISSUE #4496: Langfuse scores — user feedback and automated quality signals on traces
## Summary

Add **Langfuse scores** to the observability pipeline: user feedback (thumbs up/down or rating on assistant replies) and, later, automated quality signals are attached to the corresponding agent-turn trace as Langfuse `score-create` events. Users can rate a reply in the desktop UI and see that score on the trace in Langfuse within seconds.

## Problem

We are completing a Langfuse observability overhaul: agent turns are traced end-to-end (traces/generations/tool spans with `userId`, `sessionId`, input/output, model, token usage including reasoning + cached tokens, cost, environment/release) via the core emitter in `src/openhuman/agent/progress_tracing/`, pushing through the backend proxy `/telemetry/langfuse/ingestion` (see tinyhumansai/backend#1069 for server-side `userId` stamping, plus the upcoming `fix/langfuse-trace-metadata` PR in this repo).

The missing piece is **scores**. Langfuse supports attaching feedback and quality signals to traces via the ingestion API (`score-create` events: `traceId`, `name`, `value`, optional `comment`/`dataType`), but today we have no way to record user feedback or automated quality verdicts against a trace. That means no signal in Langfuse about which turns were good or bad — which is the main reason to have traces in the first place.

Constraints: must respect the `share_usage_data` privacy gate (no telemetry when opted out), and must work for both managed (backend-proxied) and custom model configurations.

## Solution (optional)

Scope: **both** (core + app), no backend change needed.

1. **Core — `score-create` events**: extend the batch builder in `src/openhuman/agent/progress_tracing/langfuse.rs` with a `score-create` ingestion event type. The core already knows the trace id per turn from `SpanCollector`, so a score can be linked to the correct trace.
2. **Core — RPC**: new method `openhuman.observability_submit_score` (domain: config/observability or agent) that accepts the turn/trace reference, score name, value, and optional comment, and enqueues a `score-create` event through the existing emitter.
3. **Backend**: no change — the `/telemetry/langfuse/ingestion` proxy already forwards ingestion batches verbatim.
4. **App — feedback UI**: thumbs up/down (or rating) buttons on assistant replies in the chat thread UI, calling the new RPC via `core_rpc_relay`.
5. **Later (optional)**: automated scores — the triage evaluator in `src/openhuman/agent/triage/` already computes quality/latency verdicts that could be exported as scores on the same traces.

## Acceptance criteria

- [ ] **Score visible on trace** — submitting feedback in the UI produces a Langfuse score attached to the correct turn trace within seconds.
- [ ] **Works for managed and custom models** — scores flow for both backend-proxied inference and custom model configurations.
- [ ] **Privacy gated** — no score events are emitted when `share_usage_data` is off.
- [ ] **Tests** — unit tests for the `score-create` batch builder + RPC handler, and `json_rpc_e2e` coverage for `openhuman.observability_submit_score`.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/pr-ci.yml`](../../.github/workflows/pr-ci.yml)).

## Related

- tinyhumansai/backend#1069 — server-side `userId` stamping on the Langfuse ingestion proxy.
- Upcoming `fix/langfuse-trace-metadata` PR in this repo (trace metadata completeness).
- Langfuse ingestion API `score-create` event type (traceId, name, value, comment, dataType).

salarkhannn:
## Summary

- Thumbs up/down on an assistant reply, submitting a Langfuse `score-create` against the trace that produced that turn.
- New core RPC `openhuman.observability_submit_score`, registered **internal-only** so an agent cannot score its own output.
- The core stamps the turn's trace id onto the reply row it already persists, and the renderer reads it back — so a score names a trace that actually exists.
- `share_usage_data` is enforced inside `push_score` itself, and a refused push comes back as `ok: false` rather than a confirmed rating that never left the machine.
- Rebuilt on current `main`. The previous revision of this branch was ~7.4k commits behind and conflicted with `src/openhuman/threads/tools.rs`, which `main` deleted in `75a2e1de7`.

## Problem

Agent turns are traced end to end, but there was no way to record whether a turn was any good — which is the main reason to have traces. #4496 asks for user feedback attached to the corresponding turn trace as a Langfuse `score-create`, gated on the `share_usage_data` privacy flag and working for both managed and custom model configurations.

The hard part is not the score event. It is naming the right trace. The trace id is `<trace session id>:<request id>`, and the session half is a core-side notion the renderer has never seen, so any id the frontend derives itself is a guess. Langfuse accepts a score for a trace that does not exist and then silently orphans it, so a wrong id fails invisibly — it looks like the feature works.

## Solution

**The trace id travels on the row, not on the event.** Since #6034 the core persists an unsegmented reply *before* announcing it, under `agent:<request_id>`; the client's own append is idempotent on that id and collapses onto the core's row. `reply_persistence` already relies on this for `citations`, and its doc comment says any field added to `chatDoneExtraMetadata` belongs there too. So `traceId` is stamped there, and the renderer reads `extraMetadata.traceId` off the message it is rating. It survives a reload, and it needs no new event field (`WebChannelEvent` has 73 construction sites).

**One derivation, shared.** `progress_tracing::turn_trace_id` is now the single place the id is built, used both by the progress bridge that creates the trace and by the runner that stamps it, so they cannot drift. `turn_tracing_enabled` gates the stamp on the same condition the bridge uses to install a span collector, so the id is only written when a trace really exists. No trace id means no buttons, which is the honest outcome rather than a dangling score.

**`push_score` mirrors `push_spans`** — environment skip, session bearer through the backend proxy, `x-sdk-name` product identity, partial-rejection handling. One deliberate difference: the `share_usage_data` check lives *inside* `push_score`. Span export is driven by one internal pipeline that checks the flag before collecting anything, but a score arrives from an RPC the renderer can call at any time, so this is the only place the privacy promise is un-bypassable. Opting out returns `Ok(())` — declining to send telemetry is a success.

**The RPC reports the real outcome.** It never fails (feedback must not raise a dialog over a chat), but it returns `ok: false` plus `error` when the push was refused, and the UI leaves the thumb unpressed. A hardcoded `ok: true` would both mislead the user and make the whole path untestable.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines meet the gate. Rust: `progress_tracing::{rpc,langfuse}`, `web_chat::{reply_persistence,presentation}` and `core::all` are all covered by the suites below. Vitest: `MessageFeedback.test.tsx` covers every branch of the new component; the 13 non-English locale lines are covered by every render test, which loads all 14 locale modules.
- [x] Coverage matrix updated — row **4.2.11** added to [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md)
- [x] No new external network dependencies introduced — every new test is hermetic. `config_path` is pointed at a tempdir so credential state cannot resolve to the developer's real `~/.openhuman`, and the `json_rpc_e2e` cases override `HOME` and write an explicit `[observability]` block.
- [x] N/A: Manual smoke checklist — additive affordance on an existing surface, no release-cut path changes.

## Impact

- **Desktop only.** Additive: an agent message without `extraMetadata.traceId` renders exactly as before, so every existing thread is unchanged until a new traced turn lands.
- **Privacy.** A score is a number plus a trace id. No message text rides along: the thumbs send no comment, and the trace only carries prompt/reply content when `observability.agent_tracing.capture_content` is separately opted into. Nothing is sent when `share_usage_data` is off. Catalogued as `conversation.response_feedback` with a dedicated `FEEDBACK_SCORE_TO_LANGFUSE` privacy entry naming Langfuse as a real destination beyond the backend.
- **Security.** The controller is in `build_internal_only_controllers`, so it is reachable by the renderer and absent from the agent-facing tool catalogue. Pinned by `submit_score_is_absent_from_the_agent_facing_registry`.
- **Backend.** No change — the `/telemetry/langfuse/ingestion` proxy forwards ingestion batches verbatim, and scores are just another event type in the same batch shape.
- **Compatibility.** `WebChatTaskResult` and `deliver_response` gained a field/argument; both are `pub(super)`/`pub(crate)` and all call sites are updated in this PR. No wire or schema change to any existing method.

## Related

- Closes: #4496
- Feature IDs: **4.2.11** (new), builds on 4.2.10 (durable agent reply)
- Follow-up PR(s)/TODOs: automated quality scores from `agent::triage` can reuse `push_score` unchanged — it takes any `name`/`value`, and the issue lists that as explicitly optional/later.

### What changed versus the earlier revision of this branch

Every P1 from the July bot reviews is gone with the rewrite, and the findings that survived are fixed at the root rather than patched:

| Earlier finding | Status |
| --- | --- |
| Fabricated `${threadId}:${msg.id}` trace id (Copilot, Codex) | Fixed. The id comes from the core via `extraMetadata.traceId`; there is no fallback, and no trace id means no buttons. |
| RPC always returned `{ok: true}`, so callers could not detect failure (Copilot) | Fixed. Returns `ok: false` + `error`; the UI leaves the thumb unpressed and warns. |
| Hardcoded English strings / missing `aria-label` (Copilot, CodeRabbit) | Fixed. All 14 locales translated; `aria-label` + `title` + `aria-pressed`. |
| Unhandled promise rejection on the RPC (CodeRabbit) | Fixed. |
| `push_score` errors never logged (CodeRabbit) | Fixed, and now also surfaced to the caller. |
| `build_score_batch` doc said `/api/public/ingestion` (@M3gA-Mind) | Fixed. The doc describes the batch shape without implying the client posts there. |
| `threads/tools.rs` stamp site deleted on `main` (@M3gA-Mind) | Resolved by moving the stamp to `web_chat::reply_persistence`, which is where the reply is actually persisted. The tool-wrapper approach could not have worked on the chat path regardless: `message_append` arrives on a separate task after the turn ended, so a task-local set around `run_single` is already out of scope there. |
| `#5885` rewriting the same chat surface (@M3gA-Mind) | Resolved. The buttons live in `TranscriptRow.tsx`, the current home of agent-message rendering, beside the existing citation and reaction rails. |

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feature/issue-4496`
- Commit SHA: head of this PR

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — prettier clean on all changed app files
- [x] `pnpm typecheck` — `tsc --noEmit` clean
- [x] Focused tests: `cargo test --lib` over `progress_tracing` / `web_chat` / `core::all` / `about_app` → **411 passed, 0 failed**; `cargo test --test json_rpc_e2e -- observability_submit_score` → **3 passed**; `vitest MessageFeedback.test.tsx` → **9 passed**
- [x] Rust fmt/check (if changed): `cargo fmt -p openhuman`; `cargo clippy -p openhuman --features "$(bash scripts/ci/product-features.sh)" -- -D warnings` and the contributor-default lane both clean; `node scripts/ci/check-openhuman-rust-layout.mjs` passes
- [x] Tauri fmt/check (if changed): N/A — `app/src-tauri/` untouched

### Validation Blocked

- `command:` `.husky/pre-push` (pushed with `--no-verify`)
- `error:` its first step is a repo-wide `prettier --check .`, which fails on `docs/TEST-COVERAGE-MATRIX.md` **before** this branch touches it — verified by checking out that file at `upstream/main` and running prettier on it in place. Reformatting it would have added ~86 lines of unrelated churn to this diff (prettier re-pads every row of every table in the file), so the row here is hand-formatted to match the surrounding style and the file is left as `main` has it.
- `impact:` none on this change. Every step the hook runs was run individually and passes: `prettier --check` on all changed app files, `tsc --noEmit`, `eslint` on the changed files, `cargo fmt -p openhuman`, both `cargo clippy` lanes with `-D warnings`, `node scripts/ci/check-openhuman-rust-layout.mjs`, `node scripts/check-coverage-matrix.mjs`, and `tsx scripts/i18n-coverage.ts --no-unused`.

### Behavior Changes

- Intended behavior change: an assistant reply produced by a traced turn gains thumbs up/down that record a Langfuse score on that turn's trace.
- User-visible effect: two small icon buttons beneath a traced agent reply. Absent on untraced turns and on every message persisted before this change.

### Parity Contract

- Legacy behavior preserved: an agent message without `extraMetadata.traceId` renders exactly as before. `persist_delivered_reply` omits the key entirely rather than writing `null`, so existing readers of that metadata bag are unaffected.
- Guard/fallback/dispatch parity checks: `turn_trace_id` is asserted equal to `format!("{}:{}", trace_session_id(..), request_id)` across both the session-present and session-absent branches, so the stamping path and the exporting path cannot diverge.

- Duplicate PR(s): none
- Canonical PR: this one (#4505)
- Resolution: updated in place on current `main`


salarkhannn:
Just pushed the core implementation for the Langfuse scores, along with the unit tests to pass the coverage gates.
I'm holding off on writing the  json_rpc_e2e  test (the last requirement in the issue) for a moment just to make sure we're on the same page. Let me know if the overall approach and API shape look good to you. Once you give it a thumbs up, I'll add the E2E tests and wrap this up!

M3gA-Mind:
**Maintainer review — the feature is still unimplemented on `main`, but the branch needs re-targeting, not just a rebase.** Read-only pass; I have not pushed to this branch and have not approved it.

@salarkhannn — good news first: nothing has superseded this. Checked against `main` @ `fa044d388`, neither `push_score` nor `observability_submit_score` exists anywhere in `src/`, so #4496 is still open and still yours. The Langfuse proxy architecture your `push_score` targets is also unchanged — it correctly uses `ingestion_url(config)` + `require_live_session_token(config)`, the same session-bearer-through-the-backend-proxy path as `push_observations`, so it does not need reworking. (Only the `build_score_batch` doc comment says `/api/public/ingestion`; that is the shape the backend forwards, but the wording reads as if the client posts there directly, which `langfuse.rs`'s module docs explicitly say clients never do. Worth a reword.)

The branch is two months old and three structural changes have landed under it:

**1. `src/openhuman/threads/tools.rs` was deleted** by `75a2e1de7` ("chore: remove legacy LLM tool wrappers for billing, referral, team, and people domains"). `ThreadMessageAppendTool` does not exist on `main` at all. That is where you stamp `traceId` onto `req.message.extra_metadata`, and it is the load-bearing half of the feature — without it the frontend buttons have no trace to attach a score to. `message_append` itself survives at `src/openhuman/threads/ops_part_01.rs:323`, so the stamping needs to move down to the ops layer. Worth deciding deliberately whether it belongs there or at the runner that sets `TURN_TRACE_ID`, since ops is a wider blast radius than the tool wrapper was.

**2. `progress_tracing/langfuse.rs` was split** into `langfuse_part_01.rs` / `langfuse_part_02.rs` (plus `langfuse_tests_part_01/02`). Your +150 lines need to pick a part rather than land in the shim.

**3. `channels/providers/web/run_task.rs` moved to `src/openhuman/web_chat/run_task.rs`.** Git tracked the rename, so that conflict is ordinary.

Your `TURN_TRACE_ID` task-local in `progress_tracing.rs` is new code and applies cleanly — it does not exist on `main`, so there is nothing to reconcile there.

On the frontend, the review comments from Copilot/Codex/CodeRabbit point at `app/src/pages/Conversations.tsx`; that file is now `app/src/features/conversations/Conversations.tsx`. Your diff already targets the new path, so those threads need re-reading against the moved file before you action them. **Note that `#5885` is currently open and substantially rewrites the same chat-rendering surface** (`Conversations.tsx`, `thread.tsx`, the assistant-ui message mapping). If that lands first, your feedback buttons will want re-siting into the new turn presentation. Worth coordinating rather than racing it.

**What I'd suggest:** rebase onto current `main` and re-open the discussion on where the `traceId` stamp goes now that the tool wrapper is gone — that is the one decision here that is genuinely ambiguous, and it is yours to make, not something a maintainer should guess at for you. The Rust push path, the RPC surface and the i18n additions should all carry over with little more than path changes.

None of the review threads from July were resolved, so they will need a pass too once the file paths settle.

src/openhuman/threads/tools.rs
```diff
@@ -443,7 +443,21 @@ impl Tool for ThreadMessageAppendTool {
 
     async fn execute(&self, args: serde_json::Value) -> anyhow::Result<ToolResult> {
         log::debug!("[tool][threads] message_append invoked");
-        let req: AppendConversationMessageRequest = parse_req(args, "thread_message_append")?;
+        let mut req: AppendConversationMessageRequest = parse_req(args, "thread_message_append")?;
+        // Stamp the Langfuse trace ID onto the message so the frontend feedback
+        // buttons can attach scores to the correct trace. No-op when the turn is
+        // not being traced (e.g. non-interactive / no share_usage_data).
+        if let Ok(id) =
+            crate::openhuman::agent::progress_tracing::TURN_TRACE_ID.try_with(|id| id.clone())
```
YellowSnnowmann: **blocker:** this stamp can never fire for the message the feedback buttons render on, so `extraMetadata.traceId` is always absent in the real app and every score lands on a fabricated trace id.

`ThreadMessageAppendTool` is the **agent tool** path. The web-chat assistant reply is not persisted through it — the renderer persists it:

`ChatRuntimeProvider` on `chat_done` → `addInferenceResponse` (`app/src/store/threadSlice.ts:214-227`) → `threadApi.appendMessage` → RPC `openhuman.threads_message_append` → `src/openhuman/threads/schemas.rs:572-575` → `ops::message_append`.

That path is untouched by this PR. I confirmed `git grep 'message_append(' -- src/` has no caller under `src/openhuman/channels/` at all — for web chat the frontend is the only writer, and `"traceId"` is written into `extra_metadata` in exactly two places, both of them lines 454/457 right here.

And moving the stamp down to `ops::message_append` would not fix it either: `TURN_TRACE_ID` is a `tokio::task_local!` scoped around `agent.run_single` in `run_task.rs:240`. The `threads_message_append` RPC arrives on a **separate** axum request task, *after* `chat_done` has already fired and the turn's scope has been dropped, so `try_with` returns `Err` there.

Net effect in production: `msg.extraMetadata?.traceId` is `undefined` → the UI falls back to `` `${selectedThreadId}:${msg.id}` `` → `thread_x:msg_<uuid>`, while the real trace is `trace_session_id(session_id, thread_id):request_id`. Acceptance criterion *"Score visible on trace"* fails for every single click.

**Fix — carry the trace id on the event instead of a task-local.** `progress_bridge.rs:363-364` already computes the exact id; put it on the `chat_done` payload next to `request_id`, then in `chatDoneExtraMetadata(event)` (`app/src/providers/ChatRuntimeProvider.tsx:208`) add it alongside `requestId`:

```ts
if (event.trace_id) meta.traceId = event.trace_id;
```

It then rides through `addInferenceResponse` → `appendMessage` and lands in `extraMetadata` on the persisted row with no core-side stamping at all. That also dissolves B2's hardest conflict: with the stamp on the event, this `tools.rs` hunk is no longer needed, so the modify/delete against a file that no longer exists on `main` just goes away.

app/src/features/conversations/Conversations.tsx
```diff
@@ -2299,6 +2300,58 @@ const Conversations = ({
...
                                     <div className="absolute -bottom-2 left-3 z-10 flex items-center gap-1">
+                                      {msg.sender === 'agent' && selectedThreadId && (
+                                        <>
+                                          <button
+                                            type="button"
+                                            className="rounded-full bg-white p-1 text-stone-500 shadow-sm ring-1 ring-inset ring-stone-200 hover:bg-stone-50 hover:text-stone-700 dark:bg-stone-800 dark:text-stone-400 dark:ring-stone-700 dark:hover:bg-stone-700 dark:hover:text-stone-200"
+                                            onClick={() => {
+                                              const traceId =
+                                                (msg.extraMetadata?.traceId as string) ||
+                                                `${selectedThreadId}:${msg.id}`;
```
YellowSnnowmann: **blocker:** this fallback is not a trace id under any code path, and per the `tools.rs` comment it is currently the *only* path taken — so every score posts against a trace that was never created.

This is Copilot's and Codex's July thread, still live. Recording it here because the `extraMetadata.traceId` stamp added in response does not close it.

Two changes needed. First, once the trace id actually arrives (see the `tools.rs` comment), **drop the fabrication** — a wrong id is worse than no submission, because Langfuse accepts it and the score silently disappears into a dangling trace rather than surfacing as an error:

```suggestion
                                              const traceId = msg.extraMetadata?.traceId as
                                                | string
                                                | undefined;
                                              if (!traceId) return;
```

Second, gate the buttons themselves on the same condition so an untraced turn shows no dead affordance — extend the wrapper on line 2303 to `msg.sender === 'agent' && selectedThreadId && msg.extraMetadata?.traceId &&`. Same edit applies to the thumbs-down handler at line 2335.

tests/json_rpc_e2e.rs
```diff
@@ -1158,6 +1158,39 @@ async fn json_rpc_discovers_codex_and_claude_sessions_for_memory_ingestion() {
     rpc_join.abort();
 }
 
+#[tokio::test]
+async fn json_rpc_observability_submit_score_succeeds_when_privacy_gate_off() {
+    let _env_lock = json_rpc_e2e_env_lock();
+    let home = tempdir().expect("tempdir");
+    let openhuman_home = home.path().join(".openhuman");
+    let _home_guard = EnvVarGuard::set_to_path("HOME", home.path());
+    // No score events are emitted when share_usage_data is off, so the handler
```
YellowSnnowmann: **major:** this comment is false and the test verifies nothing. `share_usage_data` is **on** for the duration of this test, and the assertion would still pass if `push_score` were deleted.

- `OPENHUMAN_SHARE_USAGE_DATA` is read nowhere in the codebase — `git grep OPENHUMAN_SHARE_USAGE_DATA src/` returns only `rpc.rs:177`, the other test's own `set_var`. The `set_var` on line 1169 is inert.
- `share_usage_data` defaults to `true` (`src/openhuman/config/schema/observability.rs:42`, asserted by its own `share_usage_data_is_on_by_default` test) and `write_min_config` emits no `[observability]` block.

So the real execution is: gate passes → `ingestion_url` resolves to `http://127.0.0.1:9/telemetry/langfuse/ingestion` (starts with `http`, passes the guard) → `require_live_session_token` fails (no session in the temp HOME) → `push_score` returns `Err` → the handler swallows it and returns `{ok: true}`. `assert_eq!(response["result"]["ok"], true)` then passes for a reason entirely unrelated to scores. The issue's *"`json_rpc_e2e` coverage for `openhuman.observability_submit_score`"* criterion is not met by this.

**Fix:** turn the gate off for real, in config rather than a non-existent env var — add `[observability]\nshare_usage_data = false` to the TOML (a `write_min_config` variant, the way `write_min_config_with_local_ai_disabled` does it), and drop the `set_var`. Then make the assertion able to fail: stand up a stub ingestion server on a bound port, point `api_url` at it, and assert it received **zero** requests with the gate off and exactly one `score-create` with it on. As written, `ok: true` is a constant.

src/openhuman/agent/progress_tracing/rpc.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[test]
+    fn submit_score_response_serializes_ok_true() {
+        let resp = SubmitScoreResponse { ok: true };
+        let v = serde_json::to_value(resp).unwrap();
+        assert_eq!(v, json!({ "ok": true }));
+    }
+
+    #[tokio::test]
+    async fn test_handle_submit_score_success() {
+        std::env::set_var("OPENHUMAN_SHARE_USAGE_DATA", "false"); // Disable real network
```
YellowSnnowmann: **major:** this does not disable the network, and the test runs against the developer's real `$HOME`.

Three separate problems:

1. **`OPENHUMAN_SHARE_USAGE_DATA` is not read anywhere** — this line is the only occurrence of that string in `src/`. The comment's premise is wrong.
2. **`Config::load_or_init()` on line 79 loads the real `~/.openhuman`** — no `HOME` or `OPENHUMAN_WORKSPACE` override anywhere in this test. Combined with (1), `share_usage_data` is `true` by default, so on any machine with a live session token `cargo test` posts a real `user-feedback` score for `trace_id: "trace-test"` to the live Langfuse. `load_or_init` can also write to that directory. A unit test must not touch production telemetry.
3. **Unguarded `std::env::set_var`** — every other env-mutating test in this repo takes `TEST_ENV_LOCK` (see `WorkspaceEnvGuard::set`, which holds it for the guard's whole lifetime precisely because the var is process-global). This one mutates process state while the rest of the suite runs in parallel.

**Fix:** don't route this through `Config::load_or_init()` at all. Extract the body of `handle_submit_score` into a `async fn submit_score_with(config: &Config, req: SubmitScoreRequest)` and have the test call it with an in-memory `Config` that has `observability.share_usage_data = false` — exactly the shape `langfuse.rs`'s own `push_score_respects_privacy_gate` already uses. That removes the env var, the real `$HOME`, and the network risk in one move, and leaves the thin `handle_submit_score` wrapper as the only untested line.

src/openhuman/agent/progress_tracing/rpc.rs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            warn!(
+                target: "openhuman::agent::progress_tracing",
+                "[progress_tracing] push_score failed for trace {}: {}",
+                req.trace_id,
+                e,
+            );
+        }
+
+        Ok(serde_json::to_value(SubmitScoreResponse { ok: true }).unwrap())
```
YellowSnnowmann: **minor:** with `push_score` failing silently here, the `.catch()` the frontend added can never fire for a server-side failure — a user clicking thumbs-up when the session token is missing gets a completely silent no-op, and `ok` is documented as `"Always true."` on line 57.

The `warn!` above addresses CodeRabbit's observability point, but not this. Suggest surfacing the outcome so the UI has something to react to:

```suggestion
        Ok(serde_json::to_value(SubmitScoreResponse { ok: pushed }).unwrap())
```

with `let pushed = match langfuse::push_score(..).await { Ok(()) => true, Err(e) => { warn!(..); false } };`. Keep the RPC itself infallible (never `Err`) so feedback stays non-blocking — just let `ok: false` through so the button can show that the score didn't land.

app/src/features/conversations/Conversations.tsx
```diff
@@ -2299,6 +2300,58 @@ const Conversations = ({
                                   const pickerOpen = reactionPickerMsgId === msg.id;
                                   return (
                                     <div className="absolute -bottom-2 left-3 z-10 flex items-center gap-1">
+                                      {msg.sender === 'agent' && selectedThreadId && (
```
YellowSnnowmann: **minor:** this block is nested inside `latestVisibleMessage?.id === msg.id` (line 2296), so feedback is only ever offered on the single newest message — and it disappears the moment the user sends another turn. #4496 asks for buttons on assistant replies generally, and rating a turn you just read is the common case, so this materially narrows the feature.

There is also no submitted/selected state: clicking gives no visual confirmation, and the same score can be fired repeatedly on every click, producing duplicate `score-create` events on one trace.

**Fix:** move the feedback `<>…</>` out of the `latestVisibleMessage` callback so it renders for any agent message that has a trace id, and track submitted scores in local state (e.g. `const [feedbackByMsgId, setFeedbackByMsgId] = useState<Record<string, 1 | 0>>({})`) to render the active thumb filled and skip the RPC when the same value is already recorded. If only-latest is a deliberate v1 scope cut, worth saying so in the PR description — it reads as accidental from the nesting.

app/src/pages/__tests__/Conversations.feedback.test.tsx
```diff
@@ -0,0 +1,387 @@
+import { combineReducers, configureStore } from '@reduxjs/toolkit';
```
YellowSnnowmann: **nitpick:** this test lives under `app/src/pages/__tests__/` but imports its subject from `../../features/conversations/Conversations` (line 180). Every other test for this feature is colocated — `app/src/features/conversations/components/__tests__/*.test.tsx`, `app/src/features/conversations/timeline/ConversationTimeline.test.tsx`, `app/src/features/conversations/composerSendDecision.test.ts`, and so on; none live under `pages/`.

Move it to `app/src/features/conversations/__tests__/Conversations.feedback.test.tsx` and shorten the imports accordingly (`../Conversations`, `../../../store/threadSlice`, …). Worth doing in the same pass as the rebase, since it stops the file drifting further from its subject.

YellowSnnowmann (dismissed):
## Summary

Reviewed `fea913f58` against #4496. The Rust push path is genuinely good — `push_score` mirrors `push_spans` exactly (privacy gate first, `ingestion_url` + `require_live_session_token`, partial-rejection handling), `build_score_batch` is cleanly extracted and well tested, and the controller is correctly parked in `build_internal_only_controllers()` so agents can't fabricate scores. The July P1s from Codex/Copilot (missing brace, `FieldSchema.ty`/`TypeSchema::F64`, two-arg `callCoreRpc`) are all fixed, `aria-label` is present, and all 14 locales carry real translations, not English fallbacks.

But the feature does not work end to end, and the branch cannot be merged in its current state. **Three blockers.**

### Blockers (3) · Major (2) · Minor (2) · Nitpick (1)

**B1 — the `traceId` stamp never runs for the message the buttons are attached to.** Inline on `src/openhuman/threads/tools.rs`. This is the load-bearing half of the feature and it is wired to a path the chat reply never takes.

**B2 — unmergeable: 4 conflicts, one of them modify/delete.** `git merge-tree upstream/main fea913f58` → rc=1:

| file | conflict |
|---|---|
| `src/openhuman/threads/tools.rs` | **modify/delete** — deleted on `main` by `75a2e1de7`; `ThreadMessageAppendTool` no longer exists |
| `src/openhuman/agent/progress_tracing/langfuse.rs` | content — split into `langfuse_part_01.rs` / `langfuse_part_02.rs` |
| `src/openhuman/web_chat/run_task.rs` | content — moved from `channels/providers/web/` |
| `app/src/features/conversations/Conversations.tsx` | content |

The branch is **7454 commits behind `main`**. This confirms @M3gA-Mind's read; the `tools.rs` deletion is the one that needs a design decision rather than a path fix, and B1 below is an argument for where that decision should land.

**B3 — last CI run is red, and the head commit has never been CI'd.** `CI Lite` run [29615085533](https://github.com/tinyhumansai/openhuman/actions/runs/29615085533) @ `73cbe7a67` failed:

```
test openhuman::threads::tools::tests::message_append_injects_trace_id ... FAILED
panicked at src/openhuman/threads/tools.rs:948:14:
tool execution should succeed: thread_message_append: thread thread-d7468e9e-... not found
test result: FAILED. 2250 passed; 1 failed
```

`Rust Core Coverage` and `PR CI Gate` both failed. The head commit `fea913f58` ("fix test race in message_append_injects_trace_id", pushed 5 days later) has **zero check-suites** — `gh api .../commits/fea913f58/check-suites` returns an empty list. The `WorkspaceEnvGuard` fix looks right in principle (it holds `TEST_ENV_LOCK` for its lifetime, so it does serialize against other workspace-mutating tests), but it is entirely unverified. Please push to re-trigger CI.

### Outside the diff

- **`OPENHUMAN_SHARE_USAGE_DATA` is not a real environment variable.** `git grep OPENHUMAN_SHARE_USAGE_DATA src/` returns exactly one hit — the test's own `set_var` in `rpc.rs:177`. Nothing in the config loader reads it. This is the root cause of both Major findings.
- **Still-open July threads.** None of the review threads from July were resolved, and two are still live on the current code: Copilot's *"RPC handler discards the result of `push_score` and always returns `{ ok: true }`, so callers cannot detect failures"* (the `warn!` that was added covers CodeRabbit's logging point, but the UI still cannot distinguish success from a swallowed failure), and Copilot/Codex on the fabricated trace id — see B1, which explains why the `extraMetadata` stamp does not close it. @M3gA-Mind's `build_score_batch` doc-comment reword also still applies.
- **Acceptance criteria status vs #4496:** ❌ *Score visible on trace* (B1) · ⚠️ *Works for managed and custom models* — untestable while B1 stands, though the proxy path itself is right · ✅ *Privacy gated* in production code (`push_score` checks `share_usage_data` before anything else) but ❌ **not actually covered by either test that claims to** (M1) · ❌ *Tests* — the `json_rpc_e2e` case is vacuous · ❌ *Diff coverage ≥ 80%* — `PR CI Gate` red.

### Verified / looks good

- `run_task.rs:236-240` builds `format!("{base}:{request_id}")` from `trace_session_id(metadata.session_id, thread_id)` — byte-identical to `progress_bridge.rs:363-364`, same function, same three values. The trace id itself is **correct**; the problem is purely that it never reaches the persisted message.
- `push_score` privacy gate, URL guard, timeout, non-2xx excerpt, and `errors[]` partial-rejection warn all match `push_spans`.
- `build_score_batch` tests cover name/value, optional comment, and `value = 0.0` (the thumbs-down case that a truthiness bug would eat).
- Internal-only registration is the right call and addresses Codex's P2 about agent-facing discovery.
- i18n: all 14 locales translated (`ru: 'Хороший ответ'`, `zh-CN: '好的回答'`, `ar: 'استجابة جيدة'` …), `aria-label` + `title` both set, `.catch()` on both handlers. `no-console` is `off` in `app/eslint.config.js:126` and the file already uses `console.warn`, so that's consistent.

YellowSnnowmann:
Rebuilt this branch on current `main` (`de5922e34`) and pushed. `CONFLICTING` → `MERGEABLE`.

@salarkhannn — your Rust push path carried over almost intact; the force-push was because a rebase was not possible, not because the approach was wrong. @M3gA-Mind's read was right on all three structural points, and one of them changed the design.

**The `threads/tools.rs` question resolved itself once I looked at where the reply is actually written.** Stamping `traceId` in `ThreadMessageAppendTool` could not have worked on the chat path even before `main` deleted that file. In web chat the assistant reply is not persisted through the agent tool — and `TURN_TRACE_ID` is a `tokio::task_local!` scoped around `agent.run_single`, so by the time any `message_append` arrives it is on a different task, after the turn ended, and `try_with` returns `Err`. The stamp would silently never fire, the UI would fall back to `${threadId}:${msg.id}`, and Langfuse would accept that score and orphan it. That is the failure this feature exists to prevent, so it is worth stating plainly rather than moving the same code down a layer.

What replaced it: since #6034 the core persists the reply *before* announcing it, under `agent:<request_id>`, and the client's append is idempotent on that id and collapses onto the core's row. `reply_persistence`'s own doc comment already says any field added to `chatDoneExtraMetadata` belongs there too — that is how `citations` works. So `traceId` is stamped there, the renderer reads `extraMetadata.traceId` off the message it is rating, and it survives a reload. It also avoids adding a field to `WebChannelEvent`, which has 73 construction sites.

Two supporting changes so this cannot rot: `progress_tracing::turn_trace_id` is now the single derivation of the id, shared by the bridge that creates the trace and the runner that stamps it; and `turn_tracing_enabled` gates the stamp on the same condition the bridge uses to install a span collector, so an id is only ever written when a trace exists. No trace id means no buttons.

**On the July threads.** All the P1s are gone with the rewrite. Two that were reported but not actually fixed are fixed now:

- The RPC returned a hardcoded `{ok: true}`. It now returns the real outcome (`ok: false` + `error`), and the UI leaves the thumb unpressed rather than confirming a rating that never left the machine. The old shape also made the E2E test vacuous — it asserted a constant and would have passed with `push_score` deleted.
- Both privacy-gate tests set `OPENHUMAN_SHARE_USAGE_DATA`, which is read nowhere in the codebase. `share_usage_data` defaults to **on**, so those tests were exercising the gate in its enabled state and passing for an unrelated reason. They now write an explicit `[observability]` block, and the two directions assert different observable outcomes.

One more worth flagging since it was in the old unit test: it called `Config::load_or_init()` with no `HOME` override. Credential state resolves against `config_path.parent()`, so on a machine with a live session that test would have posted a real `user-feedback` score to production Langfuse from `cargo test`. Every new test points `config_path` at a tempdir.

**On `#5885`** — the chat surface did move, exactly as you warned. The buttons now live in `TranscriptRow.tsx` beside the existing citation and reaction rails, not in `Conversations.tsx`. They are deliberately *not* gated on `isLatestVisible`: the trace id is persisted on the row, so an older turn stays rateable after a reload.

`build_score_batch`'s doc comment is reworded, and `push_score` picked up two things `main` has that the July branch predates: the `skip_push(environment)` gate and the `x-sdk-name` product-identity header, which `CLAUDE.md` names this exact call site for.

**Validation.** `cargo test --lib` over `progress_tracing` / `web_chat` / `core::all` / `about_app`: 411 passed, 0 failed. `json_rpc_e2e -- observability_submit_score`: 3 passed. Vitest `MessageFeedback.test.tsx`: 9 passed. Both clippy lanes clean with `-D warnings`, `tsc` clean, layout gate, coverage matrix (row 4.2.11) and `i18n-coverage --no-unused` all pass. Pushed `--no-verify` for one pre-existing reason, disclosed under *Validation Blocked*: the hook's repo-wide `prettier --check .` already fails on `docs/TEST-COVERAGE-MATRIX.md` at `upstream/main`, and reformatting it would have added ~86 lines of unrelated churn here.

Happy to split any of this out if you would rather land it in pieces.

app/src/features/conversations/components/TranscriptRow.tsx
```diff
@@ -194,6 +195,12 @@ function TranscriptRowImpl({
                   </p>
                 )}
                 <MessageCitations raw={msg.extraMetadata?.citations} />
+                {/* Thumbs on the answer, submitting a Langfuse score against
+                    the turn's own trace (#4496). Not gated on
+                    `isLatestVisible`: the trace id is persisted on the row, so
+                    an older turn is still rateable after a reload, and the
+                    reply you want to rate is often no longer the newest one. */}
+                <MessageFeedbackRail traceId={msg.extraMetadata?.traceId} />
```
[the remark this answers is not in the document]
YellowSnnowmann: Rebutting this one — there is no injection sink on the path, and the validation the rule asks for is already there.

`MessageFeedbackRail` is the type guard: it renders nothing unless `typeof traceId === 'string' && traceId.length > 0`, which is why the untyped metadata bag is unwrapped there rather than at the call site (same shape as `MessageCitations` two lines above, which validates the citation array before rendering it).

Where the value actually goes, end to end:

1. `callCoreRpc({params: {trace_id: traceId, ...}})` — serialised as JSON, never interpolated into a URL or a header.
2. Core `build_score_batch` — `json!({"traceId": trace_id})`, serialised again by `serde_json`.
3. Langfuse, as a JSON string field.

It is never rendered. `grep -iE 'innerHTML|dangerouslySetInnerHTML|href|src=' MessageFeedback.tsx` is empty — the only DOM output in that component is two `<svg>` icons and `t(...)` label text, so there is no XSS surface for it to reach.

On provenance: the realistic writer of a `traceId` that the core did not stamp is something already able to call the core RPC directly, and such a caller can invoke `openhuman.observability_submit_score` with any `trace_id` it likes without going near this component. So a check here would not be a boundary, and the worst outcome remains the one Langfuse already tolerates: a score attached to a trace id that does not resolve, which is exactly the failure this PR removed from the happy path.

The real boundary for this feature is `build_internal_only_controllers` — the RPC is deliberately absent from the agent-facing catalogue, pinned by `submit_score_is_absent_from_the_agent_facing_registry`.

YellowSnnowmann:
@tinysweeper the `critique` finding is a false positive — the imports it asks for are already there, on line 1.

> **[high] Import or define `turn_trace_id` and `turn_tracing_enabled** — `src/openhuman/agent/progress_tracing/progress_tracing_tests_part_03_tests.rs` … neither function is defined in this file nor imported from its parent module. The compiler will reject these calls with an unresolved-name error.

```console
$ head -1 src/openhuman/agent/progress_tracing/progress_tracing_tests_part_03_tests.rs
use super::*;

$ grep -n '^pub fn turn_trace_id\|^pub fn turn_tracing_enabled' src/openhuman/agent/progress_tracing.rs
250:pub fn turn_tracing_enabled(config: &Config) -> bool {
263:pub fn turn_trace_id(ui_session_id: Option<u64>, thread_id: &str, request_id: &str) -> String {
```

The file is a `#[path]` submodule of `progress_tracing_tests.rs`, which is itself `#[path]`-mounted on `progress_tracing.rs` and also opens with `use super::*;`. Both functions are `pub` on the parent module, so the glob resolves. The suggested `pub(crate)` change would be a narrowing, not a fix.

This is the hunk-in-isolation limitation the bot's own docs describe: the finding is derived from the added lines without the file's existing header. The `security` lane read the same commit and reported no import problem.

Verified by compiling rather than by reading: `cargo test -p openhuman --lib --features "$(bash scripts/ci/product-features.sh)" -- openhuman::agent::progress_tracing openhuman::web_chat core::all platform::about_app` → **411 passed, 0 failed**, which includes the three tests in that file (`turn_trace_id_prefers_the_ui_session_then_falls_back_to_the_thread`, `turn_trace_id_is_built_from_trace_session_id`, `turn_tracing_is_enabled_by_either_sink`). An unresolved name is a compile error, so nothing in that crate could have run at all had this been real. CI's `Rust Quality` and `Rust Core Coverage` lanes on this SHA are independent confirmation.

No change pushed for either tinysweeper finding; the `security` one is answered in its own thread. Both should clear on the next review pass.

PR #4520, opened 2026-07-04
perf(memory): batch the prior-lifecycle read in ingest, dropping an N+1

mysma-9403:
## What

The memory ingest pipeline reads every staged chunk's *prior* lifecycle status before the upsert, to decide which chunks are genuinely new (and need an `extract_chunk` job) versus already past extraction. That read was an **N+1**: one `SELECT … WHERE id = ?` per staged chunk, in a loop, **inside the write transaction**.

`src/openhuman/memory/ingest_pipeline.rs`:
```rust
let mut prior: HashMap<String, Option<String>> = HashMap::new();
for s in &staged_for_store {
    let status = chunk_store::get_chunk_lifecycle_status_tx(&tx, &s.chunk.id)?;   // 1 round-trip / chunk
    prior.insert(s.chunk.id.clone(), status);
}
```

A single document fans out into many chunks, so a large note/email/Notion page issues dozens–hundreds of single-row queries — each a fresh `prepare` + B-tree lookup — before the batch even upserts.

## Change

Add `get_chunk_lifecycle_statuses_tx`: one `SELECT id, lifecycle_status … WHERE id IN (…)` (windowed at `MAX_FETCH_BATCH` to stay under SQLite's bound-parameter cap), and call it once:

```rust
let staged_ids: Vec<String> = staged_for_store.iter().map(|s| s.chunk.id.clone()).collect();
let prior = chunk_store::get_chunk_lifecycle_statuses_tx(&tx, &staged_ids)?;
```

This is not a new pattern — it mirrors the two batched readers already in the same store (`get_chunks_batch`, `get_chunk_embeddings_for_signature_batch`), reusing their windowing/placeholder shape. The now-unused single-id `get_chunk_lifecycle_status_tx` is removed; the single-id `get_chunk_lifecycle_status` (used by read RPC) stays.

## Is it worth it? (honest take)

SQLite is in-process, so each avoided round-trip is cheap in absolute terms — this is **not** a dramatic speedup, and I won't oversell it. Two things make it worthwhile anyway:

1. **It shortens the write-transaction critical section.** The loop runs inside an `IMMEDIATE` write transaction, and SQLite serialises writers — every extra statement holds the write lock a little longer, blocking other ingest workers. Collapsing N statements into one shortens that hold time, which is the kind of win that actually shows up under concurrent ingest.
2. **It scales with document size** (statement prepares + lookups go from O(chunks) to O(1) queries) and follows an **existing in-repo precedent** for exactly this refactor, so it's low-risk and consistent rather than a speculative micro-tweak.

## Correctness

Behaviour-preserving: the map is keyed by chunk id, present rows map to their status, missing ids are absent — identical to the old `.get(id).flatten()` treating a miss as "no prior row" (genuinely new). Ordering is irrelevant (map lookup). The batch reads inside the same transaction, so it sees the same pre-upsert snapshot the loop did.

## Tests

New unit tests in `store_tests.rs`:
- `lifecycle_statuses_batch_matches_single_reads` — batch result equals looping the single-id read for present ids (explicit `admitted`/`sealed` plus the insert default), and a missing id is absent.
- `lifecycle_statuses_batch_empty_input_issues_no_query` — empty input → empty map, no SQL.

`cargo test --lib memory_store::chunks` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chunk processing so lifecycle states are checked more reliably before scheduling extraction work.
  * Reduced missed lookups by using batched status reads, which also handles empty and missing ID cases more consistently.

* **Tests**
  * Added coverage for batched lifecycle status retrieval across multiple chunks.
  * Added a test confirming empty input returns no results and does not trigger unnecessary work.

### DIFF
--- a/src/openhuman/memory/ingest_pipeline.rs
+++ b/src/openhuman/memory/ingest_pipeline.rs
@@ -290,7 +290,7 @@ async fn persist(
     let source_id_for_store = source_id.to_string();
     let raw_refs_for_store = raw_refs_for_chunks.clone();
     let written = tokio::task::spawn_blocking(move || -> Result<Option<usize>> {
-        use std::collections::{HashMap, HashSet};
+        use std::collections::HashSet;
         chunk_store::with_connection(&config_owned, |conn| {
             // IMMEDIATE, not the default DEFERRED: this transaction reads
             // (get_chunk_lifecycle_status_tx) before it writes
@@ -362,11 +362,16 @@ async fn persist(
             // insert a new row that picks up the column DEFAULT — so reading
             // post-upsert can't distinguish "brand new" from
             // "already-admitted-from-prior-ingest".
-            let mut prior: HashMap<String, Option<String>> = HashMap::new();
-            for s in &staged_for_store {
-                let status = chunk_store::get_chunk_lifecycle_status_tx(&tx, &s.chunk.id)?;
-                prior.insert(s.chunk.id.clone(), status);
-            }
+            // Batch the prior-lifecycle read: one `SELECT … WHERE id IN (…)`
+            // instead of one round-trip per staged chunk. Keeping this short
+            // matters because it runs inside the write transaction, and SQLite
+            // serialises writers. `prior` maps a chunk id to its pre-upsert
+            // status; ids with no prior row are simply absent (a batch miss ==
+            // the old per-id `Ok(None)`), so the lookup below treats a miss as
+            // "genuinely new".
+            let staged_ids: Vec<String> =
+                staged_for_store.iter().map(|s| s.chunk.id.clone()).collect();
+            let prior = chunk_store::get_chunk_lifecycle_statuses_tx(&tx, &staged_ids)?;
 
             let n = chunk_store::upsert_staged_chunks_tx(&tx, &staged_for_store)?;
 
@@ -392,7 +397,7 @@ async fn persist(
             // leave the lifecycle alone and skip the extract enqueue.
             let mut to_schedule: HashSet<String> = HashSet::new();
             for s in &staged_for_store {
-                let pre = prior.get(&s.chunk.id).cloned().flatten();
+                let pre = prior.get(&s.chunk.id).cloned();
                 let needs_processing = matches!(
                     pre.as_deref(),
                     None | Some(chunk_store::CHUNK_STATUS_PENDING_EXTRACTION),
--- a/src/openhuman/memory_store/chunks/store.rs
+++ b/src/openhuman/memory_store/chunks/store.rs
@@ -807,11 +807,64 @@ pub fn get_chunk_lifecycle_status(config: &Config, chunk_id: &str) -> Result<Opt
     })
 }
 
-pub(crate) fn get_chunk_lifecycle_status_tx(
+/// Batched form of the per-chunk lifecycle read, for use inside a write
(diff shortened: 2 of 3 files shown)


senamakel (commented):

**Unmet gates:**

1. **Merge conflict.** The PR has merge conflicts (`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`). It must be rebased onto the latest `main` before it can be merged.

2. **CI failures.** Three checks fail:
   - `Markdown Link Check` -- pre-existing issue (broken fragment anchors in translated docs, not related to code changes).
   - `Rust Core Coverage (cargo-llvm-cov)` -- pre-existing compilation error in the `tinyagents` dependency (`variant `AgentEvent::ToolCompleted` has no field `started_at_ms``), not caused by this PR.
   - `PR CI Gate` -- aggregate failure due to the above.
   The CI must be green (including pre-existing issues resolved on main) before approval.

**N+1 claim:** The approach is correct. The new `get_chunk_lifecycle_statuses_tx` function correctly batches the prior-lifecycle read using `WHERE id IN (...)` windowed at `MAX_FETCH_BATCH`, mirroring the existing `get_chunks_batch` / `get_chunk_embeddings_for_signature_batch` patterns. The test coverage (batch matches single reads for present/missing ids, empty input issues no query) is thorough. The ingest pipeline change (collecting staged ids, calling the batch function, adjusting the `prior.get` from `.cloned().flatten()` to `.cloned()` since the batch returns `HashMap<String, String>` instead of `HashMap<String, Option<String>>`) is behavior-preserving.

Once rebased and CI is green, this PR is ready for approval.

mysma-9403:
Closing this as **superseded** rather than rebasing — thanks for the review, @senamakel, but the rebase isn't viable: the code this PR optimized no longer exists in the repo.

Since this branch was opened, `main` refactored the ingest persist path out from under it:

- The `persist()` function and its per-chunk `get_chunk_lifecycle_status_tx` loop — the N+1 this PR collapsed — were **removed from `src/openhuman/memory/ingest_pipeline.rs`** (that file now holds only the `ingest_chat` / `ingest_email` / `ingest_document` entry points).
- Chunk-store persistence moved into the vendored **`tinycortex`** crate; `memory_store/chunks/store.rs` is now a thin delegator (`get_chunk_lifecycle_status_tx` → `tinycortex::memory::chunks::get_chunk_lifecycle_status_tx`).
- `memory_store/chunks/store_tests.rs`, where this PR added its coverage, was **deleted**; ingest is now structured under `memory/ingestion/`, which no longer carries any per-chunk lifecycle/extract-decision loop.

A trial rebase confirms this: a modify/delete conflict on the (now-deleted) test file plus a ~190-line content conflict in `ingest_pipeline.rs`, because the whole persisted region is gone. Re-homing the batched `SELECT … WHERE id IN (…)` would now mean changing `tinycortex` itself (a separate repository), which is out of scope for this PR and may already be handled differently there.

The N+1 finding was validated at the time, but the in-repo target it applied to has been refactored away, so there's nothing left here to land. Closing as superseded.

mysma-9403:
Superseded by the ingest refactor that moved chunk-store persistence into the tinycortex crate and removed the per-chunk lifecycle loop from ingest_pipeline.rs. Details in the comment above.

PR #4522, opened 2026-07-04
fix(announcements): degrade best-effort latest fetch to null, stop Sentry flood (#4401)

### LINKED ISSUE #4401: Announcements 404 floods Sentry — best-effort fetch must degrade to null silently
## Source
Sentry (self-hosted): TAURI-RUST-HW0 (`backend_api`/`authed_json`) + TAURI-RUST-KHX (`rpc`/`invoke_method` re-wrap) — one failure, reported at two layers.
Events: ~452 combined · Users affected: 19 · First seen: 2026-06-27 · Escalating · release openhuman@0.58.7.

## Symptom
`GET /announcements/latest` returns 404 in prod; the client `report_error`s it twice per call and propagates `Err`, flooding Sentry.

## Root cause
Announcements-on-init (#4125) is a **best-effort, fire-once, cosmetic** feature. Its frontend service already documents the intent — *"Network/auth failures resolve to null — a missing announcement is never worth surfacing an error for"* (`app/src/services/announcementService.ts`) — but the Rust path does not honor it:
- `src/openhuman/announcements/ops.rs::get_latest_announcement` propagates any error from `authed_json`.
- `src/api/rest.rs::authed_json` fires `report_error` on the non-2xx (HW0); the RPC dispatch re-wrap fires again (KHX).

The prod 404 itself is a backend deploy gap (tinyhumansai/backend#1064 — route #1034 merged to main, not shipped). This issue is the **client-side hardening** so a cosmetic endpoint never floods Sentry (also future-proofs staged-deploy skew / rollback).

## Fix direction
Treat announcements fetch failure (esp. 404) as "no announcement" → return null, no `report_error`. Mirror the existing expected-404 path-suppression precedent (`/channels/*/messages/*` → `BackendApiError::MessageNotFound`) so it's scoped, not a blanket demotion. Collapses the double-report.

## Bucket (S3.5)
preventable (client over-reports a cosmetic endpoint) — client cannot deploy the backend, so graceful degradation is the correct design at this layer; backend deploy tracked in tinyhumansai/backend#1064. NOT a blindfold: the deploy issue keeps the "feature actually works" path honest.

## Related
- Backend deploy: tinyhumansai/backend#1064
- Sentry-Issue: TAURI-RUST-HW0, TAURI-RUST-KHX

mysma-9403:
Closes #4401.

## What

`GET /announcements/latest` currently 404s in prod (backend route merged but not shipped — tinyhumansai/backend#1064), and the client turns that into a Sentry flood: ~452 events / 19 users, escalating. One failure is reported **twice** — once by `authed_json`'s `report_error` (OPENHUMAN-TAURI-HW0), then again when the RPC dispatch re-wraps the propagated `Err` (OPENHUMAN-TAURI-KHX).

Announcements-on-init (#4125) is a best-effort, fire-once, **cosmetic** feature. The frontend service already documents the intent — *"Network/auth failures resolve to null — a missing announcement is never worth surfacing an error for"* — but the Rust path didn't honor it.

## Change

Two scoped layers, mirroring the existing expected-404 suppression precedent for `/channels/*/messages/*` (`BackendApiError::MessageNotFound`, TAURI-2Y):

1. **`src/api/rest.rs::authed_json`** — a 404 on the announcements endpoint is an expected state under staged deploy / rollback. Suppress `report_error` and bail with an announcements-scoped message. New segment-aware `is_announcements_path` helper keeps the scope tight (tolerates base-path prefixes like `/api/v1/announcements/latest`; a substring like `/foo/announcementsxyz` does **not** match). Kills HW0.

2. **`src/openhuman/announcements/ops.rs::get_latest_announcement`** — any fetch failure degrades to `null` ("no announcement") instead of propagating `Err`, so the RPC layer never sees an error to re-wrap. Kills KHX and gives the correct UX for a cosmetic endpoint.

## Not a blindfold

The suppression is deliberately narrow. Only the **404** is silenced at the REST layer; a genuinely unexpected non-404 (e.g. a 500) on the announcements path still flows through `report_error`, so real backend breakage stays visible in Sentry. The ops-layer degradation only affects the user-facing surface (no error shown for a cosmetic feature), not Sentry visibility for non-404s. The backend deploy gap itself is tracked separately (tinyhumansai/backend#1064), so the "feature actually works" path stays honest.

## Tests (`src/api/rest_tests.rs`)

- `authed_json_suppresses_announcements_404` — a 404 on `/announcements/latest` bails via the announcements-scoped branch (distinct message) and is not classified as the typed channel-message error.
- `is_announcements_path` unit tests: canonical path, base-path prefix, trailing slash, and negatives (substring-not-segment, channel-message path, auth path, `/`, empty) — mirrors the `parse_message_path` regression-guard style.

`cargo test --lib api::rest` green locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing announcements now fail gracefully instead of surfacing an error to the app.
  * Requests to the announcements area that return “not found” are handled quietly, reducing unnecessary error noise.
  * When announcement data can’t be fetched, the app now treats it as “no announcement” rather than a failure.

* **Tests**
  * Added coverage for announcement path handling and the 404-suppression behavior to ensure missing announcements don’t trigger error reporting.

### DIFF
--- a/src/api/rest.rs
+++ b/src/api/rest.rs
@@ -94,6 +94,17 @@ fn parse_message_path(path: &str) -> Option<(&str, &str)> {
     None
 }
 
+/// True when `path` addresses the announcements endpoint family
+/// (`/announcements/…`), tolerant of base-path prefixes such as
+/// `/api/v1/announcements/latest`. Segment-aware, so a substring like
+/// `/foo/announcementsxyz` does not match. Used to scope the best-effort 404
+/// suppression to this cosmetic endpoint (see `authed_json`).
+fn is_announcements_path(path: &str) -> bool {
+    path.split('/')
+        .filter(|s| !s.is_empty())
+        .any(|segment| segment == "announcements")
+}
+
 const CLIENT_VERSION_HEADER_MAX_LEN: usize = 64;
 
 /// Max bytes of the `body_shape` key-name list echoed into the `authed_json`
@@ -698,6 +709,35 @@ impl BackendOAuthClient {
                     );
                     anyhow::bail!(
                         "channel message not found (404) on {} {}",
+                        method.as_str(),
+                        url.path(),
+                    );
+                }
+
+                // 404 on the announcements endpoint is an expected state under
+                // staged backend deploys / rollback — the route
+                // (tinyhumansai/backend#1064) may not be shipped yet.
+                // Announcements-on-init (#4125) is a best-effort, fire-once,
+                // cosmetic feature, so a missing announcement is never worth a
+                // Sentry event. Suppress the report and bail with an
+                // announcements-scoped message; the caller degrades to "no
+                // announcement". Collapses the double-report at
+                // OPENHUMAN-TAURI-HW0 / KHX (~452 events). Non-404 errors on
+                // this path still report below, so this is not a blindfold.
+                if is_announcements_path(url.path()) {
+                    tracing::info!(
+                        domain = "backend_api",
+                        operation = "authed_json",
+                        method = method.as_str(),
+                        path = url.path(),
+                        status = status_code,
+                        failure = "non_2xx",
+                        "[backend_api] announcements 404 on {} {} — cosmetic best-effort endpoint, suppressing Sentry",
+                        method.as_str(),
+                        url.path(),
+                    );
+                    anyhow::bail!(
+                        "announcements not found (404) on {} {}",
                         method.as_str(),
                         url.path(),
                     );
--- a/src/openhuman/announcements/ops.rs
+++ b/src/openhuman/announcements/ops.rs
@@ -46,11 +46,16 @@ async fn get_authed_value(config: &Config, method: Method, path: &str) -> Result
 pub async fn get_latest_announcement(config: &Config) -> Result<RpcOutcome<Value>, String> {
     match get_authed_value(config, Method::GET, "/announcements/latest").await {
         Ok(data) => Ok(RpcOutcome::single_log(data, "latest announcement fetched")),
-        Err(e) => {
+        Err(_e) => {
+            // Deliberately do NOT log the raw error: for a non-404 failure
+            // `authed_json` formats it as "GET /announcements/latest failed
+            // (…): {body}", so `%e` could write a raw upstream response body
+            // (potential PII) to the logs. The REST layer already reports the
+            // real error with a PII-safe `body_shape`; here we only note that
+            // the cosmetic fetch degraded.
             tracing::info!(
                 domain = "announcements",
                 operation = "get_latest_announcement",
-                error = %e,
                 "[announcements] latest fetch failed — degrading to no announcement (null)"
             );
             Ok(RpcOutcome::single_log(
--- a/src/openhuman/announcements/ops.rs
+++ b/src/openhuman/announcements/ops.rs
@@ -33,7 +33,30 @@ async fn get_authed_value(config: &Config, method: Method, path: &str) -> Result
 /// Fetch the latest active announcement for the signed-in user.
 /// Maps to `GET /announcements/latest`. The backend returns the announcement
 /// object or `null` when nothing qualifies; both pass through verbatim.
+///
+/// Announcements-on-init (#4125) is a best-effort, fire-once, cosmetic feature:
+/// a missing or unavailable announcement is never worth surfacing an error for.
+/// Any fetch failure — a 404 under a staged backend deploy / rollback
+/// (tinyhumansai/backend#1064), a lapsed session, transient infra — degrades to
+/// "no announcement" (`null`) instead of propagating an `Err` that the RPC layer
+/// would re-wrap into a second Sentry event (OPENHUMAN-TAURI-KHX). The frontend
+/// service documents the same intent. The REST layer still reports genuinely
+/// unexpected (non-404) errors to Sentry, so this is graceful degradation, not a
+/// blindfold. See #4401.
 pub async fn get_latest_announcement(config: &Config) -> Result<RpcOutcome<Value>, String> {
-    let data = get_authed_value(config, Method::GET, "/announcements/latest").await?;
-    Ok(RpcOutcome::single_log(data, "latest announcement fetched"))
+    match get_authed_value(config, Method::GET, "/announcements/latest").await {
+        Ok(data) => Ok(RpcOutcome::single_log(data, "latest announcement fetched")),
+        Err(e) => {
+            tracing::info!(
+                domain = "announcements",
+                operation = "get_latest_announcement",
+                error = %e,
+                "[announcements] latest fetch failed — degrading to no announcement (null)"
+            );
+            Ok(RpcOutcome::single_log(
+                Value::Null,
+                "no announcement (fetch failed, degraded to null)",
+            ))
+        }
+    }
 }
--- a/src/api/rest_tests.rs
+++ b/src/api/rest_tests.rs
@@ -1,6 +1,7 @@
 use super::{
-    backend_api_body_shape, flatten_authed_error, key_bytes_from_string, parse_message_path,
-    sanitize_client_version, BackendApiError, BackendOAuthClient, BACKEND_API_BODY_SHAPE_MAX_BYTES,
+    backend_api_body_shape, flatten_authed_error, is_announcements_path, key_bytes_from_string,
+    parse_message_path, sanitize_client_version, BackendApiError, BackendOAuthClient,
+    BACKEND_API_BODY_SHAPE_MAX_BYTES,
 };
 use axum::extract::State;
 use axum::http::HeaderMap;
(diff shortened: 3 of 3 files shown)


src/openhuman/announcements/ops.rs
```diff
@@ -33,7 +33,30 @@ async fn get_authed_value(config: &Config, method: Method, path: &str) -> Result
...
 pub async fn get_latest_announcement(config: &Config) -> Result<RpcOutcome<Value>, String> {
-    let data = get_authed_value(config, Method::GET, "/announcements/latest").await?;
-    Ok(RpcOutcome::single_log(data, "latest announcement fetched"))
+    match get_authed_value(config, Method::GET, "/announcements/latest").await {
+        Ok(data) => Ok(RpcOutcome::single_log(data, "latest announcement fetched")),
+        Err(e) => {
+            tracing::info!(
+                domain = "announcements",
+                operation = "get_latest_announcement",
+                error = %e,
```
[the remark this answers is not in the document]
mysma-9403: Fixed — dropped the `error = %e` field from the degradation log. You're right that for a non-404 failure `authed_json` formats the error as `GET /announcements/latest failed (…): {body}`, so `%e` could write a raw upstream response body (potential PII) to the logs. The REST layer already reports the real error with a PII-safe `body_shape`, so the ops layer only notes that the cosmetic fetch degraded.

M3gA-Mind (commented):
# PR #4522 — fix(announcements): degrade best-effort latest fetch to null, stop Sentry flood (#4401)

## Walkthrough

This PR fixes the announcements-on-init Sentry flood (#4401) by (1) suppressing the announcements 404 at the REST layer (`authed_json`) via a new `is_announcements_path` segment matcher and an `anyhow::bail!`, and (2) degrading **any** fetch failure to `null` in `announcements/ops.rs::get_latest_announcement`. The intent is sound and the Codex PII concern about logging the raw error was correctly addressed (the `error = %e` field was dropped in `485ce9b1`).

**However, this PR is now superseded.** The exact same issue (#4401) has already been fixed on `main` by **PR #4417** (`379a976d7`, `fix(announcements): degrade 404 on /announcements/latest to null instead of erroring`), which took a **typed-error** approach (`BackendApiError::AnnouncementNotFound`). Merging `main` into this branch produces **hard, mutually-exclusive conflicts in all three touched files** — two competing implementations of the same feature. This needs a maintainer decision (close as superseded, or rebase and reconcile) before it can go anywhere. Posting as a comment; approval/merge is the maintainer's call.

## Changes

| File | Summary |
| --- | --- |
| `src/api/rest.rs` | Adds `is_announcements_path` + a 404-suppression branch in `authed_json` that `bail!`s with a string message. **Conflicts** with main's `is_announcements_latest_path` + typed `AnnouncementNotFound`. |
| `src/api/rest_tests.rs` | Adds `authed_json_suppresses_announcements_404` + `is_announcements_path` unit tests. **Conflicts** with main's typed-error tests. |
| `src/openhuman/announcements/ops.rs` | `get_latest_announcement` degrades all errors to `null`. **Conflicts** with main's typed-404-only degradation. |

## Actionable comments (4)

### 🛑 Blockers

#### 1. All three files — unresolved merge conflict markers
`src/api/rest.rs`, `src/api/rest_tests.rs`, and `src/openhuman/announcements/ops.rs` all contain unresolved `<<<<<<< / ======= / >>>>>>>` markers after merging `main`. The branch will not compile and cannot merge in this state:

- `src/api/rest.rs` — markers at ~L106–125 (`is_announcements_path` vs `is_announcements_latest_path`) and ~L737–781 (the 404 branch).
- `src/api/rest_tests.rs` — markers at ~L2–10 (imports), ~L369–382 and ~L400–500 (the tests).
- `src/openhuman/announcements/ops.rs` — markers at ~L37–112 (the whole `get_latest_announcement` body).

#### 2. Superseded by #4417 (already on `main` as `379a976d7`)
The conflict above is not incidental — it's because `main` already carries a complete fix for #4401 using a **typed** `BackendApiError::AnnouncementNotFound`, with the ops layer folding only that typed 404 into `null` and still propagating everything else via `flatten_authed_error`. This PR's string-`bail!` + blanket-degrade approach collides with it directly (see the mutually-exclusive tests: this PR's `authed_json_suppresses_announcements_404` asserts `downcast_ref::<BackendApiError>().is_none()`, while main's `authed_json_surfaces_announcement_not_found_on_404` asserts the typed error *is* present). Recommendation: **close this PR as superseded by #4417**, unless there is something here #4417 doesn't cover — in which case rebase onto `main` and layer only the delta on top of the typed approach.

### ⚠️ Major

#### 3. `src/api/rest.rs:102` — `is_announcements_path` is too broad; suppresses 404s it shouldn't
`is_announcements_path` matches **any** path segment equal to `announcements`, for **any** HTTP method and **any** sub-path. So a 404 on `/announcements/latest/extra`, on a future `POST /announcements/ack`, or a `DELETE /announcements/{id}` would all be silently swallowed — hiding genuine bugs on write endpoints. Main's `is_announcements_latest_path` deliberately matches only the exact `[.., "announcements", "latest"]` tail and is gated on `method == Method::GET`; main even ships a regression test (`authed_json_only_classifies_get_announcements_latest_as_not_found`) proving `/announcements/latest/extra` is *not* suppressed — a case this PR's matcher gets wrong.

**Suggested change (if reconciling rather than closing):**
```rust
// before
fn is_announcements_path(path: &str) -> bool {
    path.split('/').filter(|s| !s.is_empty()).any(|s| s == "announcements")
}
// ... later, ungated on method:
if is_announcements_path(url.path()) { ... }

// after — scope to GET /…/announcements/latest exactly (adopt main's helper)
fn is_announcements_latest_path(path: &str) -> bool {
    let segments: Vec<&str> = path.split('/').filter(|s| !s.is_empty()).collect();
    matches!(segments.as_slice(), [.., "announcements", "latest"])
}
// ...
if method == Method::GET && is_announcements_latest_path(url.path()) { ... }
```

#### 4. `src/openhuman/announcements/ops.rs:49` — blanket `Err(_e) => null` degrades non-404s too
The `Err(_e) =>` arm degrades **every** error to `null` — including 5xx, malformed responses, and session expiry (401) — so the RPC caller can no longer distinguish "no announcement" from "backend is broken / your session lapsed". Sentry visibility for non-404s is preserved at the REST layer, but the ops/RPC contract loses signal. Main's version folds only the typed 404 into `null` and re-propagates the rest via `flatten_authed_error`, which is the safer contract.

**Suggested change (if reconciling):**
```rust
// before
Err(_e) => { /* log */ Ok(RpcOutcome::single_log(Value::Null, "...degraded to null")) }

// after — only the typed 404 degrades; real failures still propagate
Err(err) if is_announcement_not_found(&err) =>
    Ok(RpcOutcome::single_log(Value::Null, "no announcement available (404)")),
Err(err) => Err(crate::api::flatten_authed_error(err)),
```

## Nitpicks (1)
- `src/api/rest.rs:727` — the branch surfaces an untyped `anyhow::bail!("announcements not found (404) …")` that callers can only match by string `.contains(...)`. The repo precedent (`MessageNotFound`, and now `AnnouncementNotFound` on main) is a typed `BackendApiError` variant. Prefer the typed error so callers pattern-match instead of string-sniff.

## Questions for the author
- Were you aware #4417 landed the same fix on `main` first? If there's a specific reason to prefer the broad/blanket approach here over #4417's typed/scoped one, please spell it out — otherwise this looks like a duplicate that should be closed.

## Verified / looks good
- The Codex P2 finding (raw upstream body via `error = %e`, potential PII in logs) was correctly addressed in `485ce9b1` — the field was dropped and a PII-safe note left in its place. 👍
- The 404 suppression at the REST layer is correctly nested inside `if status_code == 404`, so non-404s do still flow to `report_error` as the PR body claims.
- Test intent and comments are thorough; the `is_announcements_path` unit tests correctly cover the substring-not-segment negative case.

---
_Review posted as a **comment** per maintainer instruction — not approving or requesting changes; the close-vs-reconcile decision is yours._

src/api/rest.rs
```diff
@@ -94,6 +94,17 @@ fn parse_message_path(path: &str) -> Option<(&str, &str)> {
     None
 }
 
+/// True when `path` addresses the announcements endpoint family
+/// (`/announcements/…`), tolerant of base-path prefixes such as
+/// `/api/v1/announcements/latest`. Segment-aware, so a substring like
+/// `/foo/announcementsxyz` does not match. Used to scope the best-effort 404
+/// suppression to this cosmetic endpoint (see `authed_json`).
+fn is_announcements_path(path: &str) -> bool {
```
M3gA-Mind, line 102: **⚠️ Major — `is_announcements_path` is too broad.** This matches *any* path segment equal to `announcements`, for *any* method and *any* sub-path. A 404 on `/announcements/latest/extra`, a future `POST /announcements/ack`, or `DELETE /announcements/{id}` would all be silently swallowed — hiding real bugs on write endpoints. Main's `is_announcements_latest_path` matches only the exact `[.., "announcements", "latest"]` tail and is gated on GET; it even ships a test proving `/announcements/latest/extra` is *not* suppressed — a case this matcher gets wrong. Prefer the tighter helper:
```rust
fn is_announcements_latest_path(path: &str) -> bool {
    let segments: Vec<&str> = path.split('/').filter(|s| !s.is_empty()).collect();
    matches!(segments.as_slice(), [.., "announcements", "latest"])
}
// gate the call on: method == Method::GET && is_announcements_latest_path(url.path())
```

src/api/rest.rs
```diff
@@ -702,6 +713,35 @@ impl BackendOAuthClient {
...
+                // staged backend deploys / rollback — the route
+                // (tinyhumansai/backend#1064) may not be shipped yet.
+                // Announcements-on-init (#4125) is a best-effort, fire-once,
+                // cosmetic feature, so a missing announcement is never worth a
+                // Sentry event. Suppress the report and bail with an
+                // announcements-scoped message; the caller degrades to "no
+                // announcement". Collapses the double-report at
+                // OPENHUMAN-TAURI-HW0 / KHX (~452 events). Non-404 errors on
+                // this path still report below, so this is not a blindfold.
+                if is_announcements_path(url.path()) {
```
M3gA-Mind, line 727: **💡 Nitpick — prefer a typed error over a string `bail!`.** This surfaces `anyhow::bail!("announcements not found (404) …")`, which callers can only match via string `.contains(...)` (as the test does). The repo precedent is a typed `BackendApiError` variant (`MessageNotFound`, and `AnnouncementNotFound` already on `main`). A typed variant lets `get_latest_announcement` pattern-match instead of string-sniff. Note `main` already does exactly this: `return Err(anyhow::Error::new(BackendApiError::AnnouncementNotFound));`

src/openhuman/announcements/ops.rs
```diff
@@ -33,7 +33,35 @@ async fn get_authed_value(config: &Config, method: Method, path: &str) -> Result
...
+/// would re-wrap into a second Sentry event (OPENHUMAN-TAURI-KHX). The frontend
+/// service documents the same intent. The REST layer still reports genuinely
+/// unexpected (non-404) errors to Sentry, so this is graceful degradation, not a
+/// blindfold. See #4401.
 pub async fn get_latest_announcement(config: &Config) -> Result<RpcOutcome<Value>, String> {
-    let data = get_authed_value(config, Method::GET, "/announcements/latest").await?;
-    Ok(RpcOutcome::single_log(data, "latest announcement fetched"))
+    match get_authed_value(config, Method::GET, "/announcements/latest").await {
+        Ok(data) => Ok(RpcOutcome::single_log(data, "latest announcement fetched")),
+        Err(_e) => {
```
M3gA-Mind, line 49: **⚠️ Major — blanket `Err(_e) => null` degrades non-404s too.** This arm degrades *every* error to `null` — 5xx, malformed responses, and session expiry (401) included — so the RPC caller can no longer tell "no announcement" from "backend down / session lapsed". Sentry still sees non-404s at the REST layer, but the ops/RPC contract loses signal. `main`'s version (PR #4417) folds only the typed 404 into `null` and re-propagates the rest:
```rust
Err(err) if is_announcement_not_found(&err) =>
    Ok(RpcOutcome::single_log(Value::Null, "no announcement available (404)")),
Err(err) => Err(crate::api::flatten_authed_error(err)),
```

src/openhuman/announcements/ops.rs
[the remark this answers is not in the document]
M3gA-Mind: Confirming this Codex P2 is resolved: `485ce9b1` dropped the `error = %e` field and left a PII-safe note, so the ops-layer degradation log no longer risks echoing a raw upstream body. The REST layer's PII-safe `body_shape` remains the single source for non-2xx detail. ✅ (Flagging separately in the top-level review that this PR now conflicts with / is superseded by #4417 on main.)

M3gA-Mind:
Thanks for tackling the announcements-404 Sentry flood (#4401) — the underlying diagnosis here is exactly right, and the PII fix you made in `485ce9b1` (not logging the raw error) was a good catch.

Closing this one as **superseded**, though, not for a lack of quality but because the same issue was already fixed and merged before this could land:

- **#4417** ("fix(announcements): degrade 404 on `/announcements/latest` to null instead of erroring", merged 2026-07-06) also **Fixes #4401** and is now on `main`.
- That's why this branch throws unresolved merge conflicts in all three touched files (`src/api/rest.rs`, `src/api/rest_tests.rs`, `src/openhuman/announcements/ops.rs`) — `main` already carries a competing implementation of the same behavior, and the two approaches are mutually exclusive (their tests assert opposite things about whether the 404 is typed).

The merged approach is also a bit tighter on two axes worth calling out for future reference:
- **Path scoping** — `main` uses an exact, GET-gated `is_announcements_latest_path`, whereas the `is_announcements_path` matcher here would swallow 404s on any path containing an `announcements` segment (e.g. `/announcements/latest/extra` or future write endpoints).
- **Degradation** — `main` degrades only the typed `BackendApiError::AnnouncementNotFound` 404 and re-propagates everything else, whereas the `Err(_e) => null` catch-all here would also swallow 5xx and session-expiry failures.

None of that reflects poorly on the work — it's just that the fix already shipped. If you spot a concrete gap in #4417's behavior that this addresses, please do open a focused follow-up against current `main` and reference it. Thanks again for the contribution!

Issue #4469, opened 2026-07-03
Misc parity cleanups & doc drift from the tinyagents migration


senamakel:
Part of the v0.58.7 → HEAD TinyAgents-migration audit (see parent issue). Low-severity parity items, dead code, and doc drift — batchable.

## Behavior nits
1. **Stop-hook "hard stop" is soft**: `StopHookMiddleware` pauses via `SteeringCommand::Pause` (`src/openhuman/agent/stop_hooks.rs:144`), drained only at the next iteration top — the current tool round (possibly expensive spawns) still executes, then `summarize_turn_wrapup` spends one more model call. Contradicts `GoalBudgetStopHook`'s "hard-stops the moment usage would exceed the cap" doc (`session/turn/core.rs:696-706`). Either enforce immediately (skip pending tool round on Stop) or fix the docs and cap the wrap-up.
2. **`TextModeProvider` passthroughs missing** (`ops/runner.rs:1157-1212`): doesn't forward `is_local_provider`/`is_local_provider_for_model`/`loaded_context_window`/`prompt_cache_capabilities` — a local provider behind text-mode loses the `n_keep >= n_ctx` guards and is profiled as remote (`tinyagents/model.rs:347`). Forward the methods.
3. **Poisoned-mutex panics on error recovery**: `.lock().unwrap()` on `error_slot` (`tinyagents/mod.rs:614`) and `EarlyExitHook` slots (`tinyagents/tools.rs:49/57`) — use `unwrap_or_else(PoisonError::into_inner)`.
4. `max_iterations == 0` now normalizes to 10 (legacy clamped to 1) — confirm intended, document.
5. Cap-summary output cap pinned to `AGENT_TURN_MAX_OUTPUT_TOKENS` instead of the definition's budget (`ops/graph.rs:318`).
6. Aggregate tool-result spill can blank a result to `""` with no artifact pointer (`harness/tool_result_artifacts/mod.rs:343,365,386,400-403`): `allowed_len` saturates to 0 → `bound_text_to_budget(_, 0)` returns `""`, discarding the `[tool_result_preview]` envelope with `artifact_path`; small `allowed_len` can truncate the envelope mid-header. Enforce a minimum envelope allowance.
7. Mid-turn Composio/tool-surface resync dropped (now turn-boundary only, `session/turn/core.rs:268`) and `LazyToolkitResolver` stubbed (`ops/runner.rs:904`, "#4249 1b") — track restoration or explicitly descope.

[...]

PR #4572, opened 2026-07-05
Remove dead rolling recap helper

ahfoysal:
## Summary

Removes the unused `rolling_segment_recap` helper from the archivist recap module and updates the nearby rustdoc so it no longer describes the old rolling-recap path.

Part of #4469.

## Verification

- `RUSTUP_TOOLCHAIN=stable rustfmt --check --edition 2021 src/openhuman/agent/harness/archivist/recap.rs`
- `git diff --check`
- `rg -n "rolling_segment_recap|rolling recap|rolling-recap|live compaction" src/openhuman/agent/harness/archivist src/openhuman/agent`

`cargo check --manifest-path Cargo.toml -p openhuman --lib` could not complete in the sparse checkout because vendored submodules beyond `tinyagents`/`tinychannels` were not fully initialized; it stopped at missing `vendor/tinycortex/Cargo.toml`.

## Summary by CodeRabbit

* **Chores**
  * Updated project housekeeping rules to exclude temporary automation and branch-related files from version control.

* **Build**
  * Updated build configuration. No changes to the application’s visible features or user interface are included in this release.


ahfoysal:
Thanks for taking the time to review. I kept this PR tightly scoped to removing the unused rolling recap path and the matching docs updates only.

I couldn’t include a full local compile on this environment yet because a few vendor dependencies weren’t available in the sparse checkout. Happy to adjust if anything else looks odd on your side.

senamakel:
> Thanks for taking the time to review. I kept this PR tightly scoped to removing the unused rolling recap path and the matching docs updates only.
> 
> I couldn’t include a full local compile on this environment yet because a few vendor dependencies weren’t available in the sparse checkout. Happy to adjust if anything else looks odd on your side.

Really aweosme to see your contributions. Can you do a git check out, and then check out the submodules. It should all be up-to-date. Do try and let me know?

ahfoysal:
Thanks for the review, and good call. You’re absolutely right that this is environment-related, not a code-path regression.

I couldn’t run `cargo check` in that sandbox because this environment was a sparse checkout and some vendored components are missing (notably `vendor/tinycortex`). I’ll treat this as a hard verification blocker until I complete a full checkout + submodules.

**Strict check rule I’m applying now:**
- Full git checkout: `git fetch` + `git checkout` this branch
- Initialize all submodules: `git submodule update --init --recursive`
- Run exactly these commands in CI-equivalent order:
  1) `RUSTUP_TOOLCHAIN=stable rustfmt --check --edition 2021 src/openhuman/agent/harness/archivist/recap.rs`
  2) `git diff --check`
  3) `rg -n "rolling_segment_recap|rolling recap|rolling-recap|live compaction" src/openhuman/agent/harness/archivist src/openhuman/agent`
  4) `cargo check --manifest-path Cargo.toml -p openhuman --lib`

I’ll report back with command output and only request merge once all four complete cleanly.

ahfoysal:
Thanks for the nudge. I did the full checkout this time and initialized the recursive submodules.

Checked locally on this branch:

```bash
git submodule update --init --recursive
RUSTUP_TOOLCHAIN=stable rustfmt --check --edition 2021 src/openhuman/agent/harness/archivist/recap.rs
git diff --check
rg -n "rolling_segment_recap|rolling recap|rolling-recap|live compaction" src/openhuman/agent/harness/archivist src/openhuman/agent
cargo check --manifest-path Cargo.toml -p openhuman --lib
```

Results:
- rustfmt passed
- diff whitespace check passed
- the rolling recap search returned no matches
- `cargo check` passed after installing the missing local `cmake` dependency required by `whisper-rs-sys`

The checkout is clean now, and the GitHub checks are green as well.

M3gA-Mind (commented):
# PR #4572 — Remove dead rolling recap helper

## Walkthrough
This PR removes the `pub async fn rolling_segment_recap` method from `ArchivistHook` and trims the surrounding rustdoc so the module describes only the finalize (`on_segment_closed`) summarization path. The change is well-scoped and the prose deletions are accurate. However, the "unused" premise does not fully hold: the method is still referenced by the archivist integration tests, so removing it breaks the build.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/agent/harness/archivist/recap.rs` | Removes `rolling_segment_recap` and updates module/helper/stub rustdoc to finalize-only wording. |

## Actionable comments (1)

### 🛑 Blockers

#### 1. `tests/agent_archivist_debug_round21_raw_coverage_e2e.rs:321,371` — deleted method is still called by tests → build break
`rolling_segment_recap` is not actually unused: the archivist integration test still invokes it in two places:

- L321: `assert_eq!(hook.rolling_segment_recap(session).await, None);`
- L371: `assert_eq!(enabled.rolling_segment_recap("missing-session").await, None);`

With the method gone, this test crate fails to compile (`no method named rolling_segment_recap found for struct ArchivistHook`), so `cargo test` / the Rust CI lane will go red. This is precisely the gap the author flagged — `cargo check` could not run in the sparse checkout, so the compile break went unseen. `rg -n rolling_segment_recap` across the repo (not just `src/`) surfaces both call sites.

**Suggested change** — delete the two now-dead assertions along with the method. In `archivist_flush_finalizes_open_segment_and_extracts_profile_events`:

```rust
    let open_before = segments::open_segment_for_session(&conn, session)?;
    assert!(open_before.is_some());
-    assert_eq!(hook.rolling_segment_recap(session).await, None);

    hook.flush_open_segment(session).await;
```

And in `archivist_disabled_and_unknown_session_paths_are_noops`:

```rust
    let enabled = ArchivistHook::new(conn, true);
    enabled.flush_open_segment("missing-session").await;
-    assert_eq!(enabled.rolling_segment_recap("missing-session").await, None);
    Ok(())
```

If any of that test's intent (verifying the open-segment/no-recap behavior) is still worth keeping, replace the assertion with an equivalent check against the surviving API rather than silently dropping coverage. Please run `cargo test --test agent_archivist_debug_round21_raw_coverage_e2e` (or the full `pnpm test:rust`) once submodules are initialized to confirm green before merge.

## Nitpicks (1)
- `src/openhuman/agent/harness/archivist/recap.rs:54-70` — after the edit, the `summarize_entries` rustdoc now carries two topic sentences: the new L54 "Shared summarize helper used by the finalize path…" and the pre-existing L65 "Summarize a set of episodic entries into a recap string." Consider folding them into a single opening line so the doc summary isn't doubled. Not introduced-by-this-PR strictly, but this is the moment it becomes visible.

None.

## Verified / looks good
- The rustdoc deletions (module header, "single LLM summarizer" note, rolling-recap caveat about live compaction) are all accurate now that the rolling path is gone; no stale references remain in the module.
- No non-test callers of `rolling_segment_recap` exist anywhere in `src/` or `app/` — the only remaining references are the two test assertions above.
- `summarize_entries` and `read_session_entries` are untouched in behavior; the finalize path is unaffected.

---
_Review only — leaving this as a comment, not an approval. The build-breaking test references should be resolved before a maintainer merges._

src/openhuman/agent/harness/archivist/recap.rs
```diff
@@ -177,122 +172,4 @@ impl ArchivistHook {
         }
         (segments::fallback_summary(first, last, turn_count), false)
     }
```
M3gA-Mind, line 179: **Blocker: build break.** Removing `rolling_segment_recap` leaves two dangling callers in `tests/agent_archivist_debug_round21_raw_coverage_e2e.rs` (L321 and L371), both `assert_eq!(...rolling_segment_recap(...).await, None)`. `cargo test` will fail to compile the archivist e2e test crate.

Delete those two assertions along with the method (or port them to a surviving API), then run `cargo test --test agent_archivist_debug_round21_raw_coverage_e2e` once submodules are initialized to confirm green. A repo-wide `rg -n rolling_segment_recap` (not scoped to `src/`) surfaces both call sites.

M3gA-Mind:
@senamakel @ahfoysal — following up on the "run it once submodules are initialized" thread: doing that check actually matters here, because the "unused" premise doesn't fully hold.\n\n`rolling_segment_recap` is still called by the archivist integration test in **two** places:\n- `tests/agent_archivist_debug_round21_raw_coverage_e2e.rs:321` — `assert_eq!(hook.rolling_segment_recap(session).await, None);`\n- `tests/agent_archivist_debug_round21_raw_coverage_e2e.rs:371` — `assert_eq!(enabled.rolling_segment_recap("missing-session").await, None);`\n\nSo with the method removed, that test crate no longer compiles and `cargo test` / the Rust CI lane will fail. A repo-wide `rg -n rolling_segment_recap` (not scoped to `src/`) surfaces both. The fix is to drop those two assertions (or port them to a surviving API) as part of this PR, then confirm with `cargo test --test agent_archivist_debug_round21_raw_coverage_e2e`.\n\nEverything else in the diff looks correct — the rustdoc deletions are accurate and there are no non-test callers. Leaving this as a review comment, not an approval.

ahfoysal:
Thanks for catching that. I removed the two stale assertions that still referenced `rolling_segment_recap` after the method was deleted.

Validation:
- `cargo test --test agent_archivist_debug_round21_raw_coverage_e2e`

All 5 tests in that file pass locally.

PR #4631, opened 2026-07-07
test(memory): inventory guard for the protocol-mandated update_memory_md tool

### LINKED ISSUE #4458: Memory protocol mandates unregistered update_memory_md tool — permanent index-drift nag loop; non-atomic MEMORY.md writes
Part of the v0.58.7 → HEAD TinyAgents-migration audit (see parent issue). Repairs the memory read→dedupe→write→update-index protocol (#4444 / PR #4444, commit 19bf3aa58).

## Problem
The protocol's write cycle can **only** be closed by a successful `update_memory_md` call (`src/openhuman/agent/harness/memory_protocol.rs:73-76,138-147`), and every durable write appends an instruction to call it. But `UpdateMemoryMdTool` is **never registered** in the production registry: `all_tools_with_runtime` (`src/openhuman/tools/ops.rs:268-270`) registers `MemoryStoreTool`/`MemoryRecallTool`/`MemoryForgetTool` only; the tool is merely re-exported (`src/openhuman/tools/impl/filesystem/mod.rs:30`) and constructed once in a test. The archivist's `named = ["update_memory_md", ...]` (`src/openhuman/agent_registry/agents/archivist/agent.toml:23`) silently resolves away because subagents only filter the parent set.

**Result:** `memory_store` succeeds → middleware instructs the model to call `update_memory_md` → unknown-tool error → tracker never sees `IndexUpdate` → escalated "MEMORY.md index is drifting… Reconcile it now" on every subsequent write (memory_protocol.rs:138-144) + `after_agent` stale-index warning every run (`src/openhuman/tinyagents/middleware.rs:1589-1603`). Permanent, unsatisfiable nag loop; wasted model round-trips; MEMORY.md never index-updated.

## Secondary problems (fix together)
1. `update_memory_md` writes are an **unlocked, non-atomic read-modify-write** (`src/openhuman/tools/impl/filesystem/update_memory_md.rs:159-190`): concurrent runs (parallel forks, cron) lose an append; crash mid-write truncates MEMORY.md. Becomes live the moment registration is fixed.
2. `remember_preference`/`save_preference` durable writes are classified `MemoryOp::Other` (`memory_protocol.rs:79-98`) — undocumented gap in write classification.

## Fix plan
1. Register `UpdateMemoryMdTool` in `all_tools_with_runtime` (`src/openhuman/tools/ops.rs`), respecting the same visibility rules as the other memory tools.
2. Make the write atomic and serialized: temp-file + rename, guarded by a per-workspace async lock (or reuse an existing file-lock helper). The per-run `MemoryProtocolTracker` has zero cross-run awareness, so file-level safety is the only guard.
3. Classify `remember_preference`/`save_preference` as durable writes in the protocol (or document why not).
4. Add a startup/registry inventory assertion so a protocol-mandated tool missing from the registry fails a test, not production (mirror the "adapter inventory test" pattern).

## Acceptance criteria
- A live turn that stores a memory can complete the full read→dedupe→write→index cycle with zero unknown-tool errors and no drift warnings.
- Two concurrent index updates both survive; a killed process never leaves a truncated MEMORY.md.
- Inventory test fails if `memory_protocol.rs` mandates a tool that isn't registered.

M3gA-Mind:
Closes #4458

On current `main`, the substantive parts of #4458 are already in place: `UpdateMemoryMdTool` is registered in `all_tools_with_runtime`, its write is atomic (temp-file + rename) and serialized behind a per-workspace async lock + inter-process flock, and `remember_preference`/`save_preference` are an explicit `MemoryOp::Other` arm. The one remaining gap from the fix plan was the registry inventory assertion.

This adds a `MEMORY_INDEX_UPDATE_TOOL` constant in `memory_protocol.rs` (the single name `classify_memory_op` matches on) and a test that builds the real `all_tools` registry and asserts the protocol-mandated tool is registered. If the protocol ever mandates a tool that isn't in the registry, this test fails instead of the model hitting a permanent unsatisfiable "call update_memory_md" nag loop in production.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory update handling to avoid “unknown tool” failures during the memory sync flow.
  * Ensured the required memory index update action is consistently recognized when editing the memory file.
* **Tests**
  * Added a regression test to confirm the full tool set includes the required memory update action.

### DIFF
--- a/src/openhuman/agent/harness/memory_protocol.rs
+++ b/src/openhuman/agent/harness/memory_protocol.rs
@@ -33,6 +33,17 @@
 /// recognise memory-protocol guidance in a tool result.
 pub const MEMORY_PROTOCOL_MARKER: &str = "[memory-protocol]";
 
+/// The tool that closes the write cycle: the protocol appends a "call
+/// `update_memory_md`" reminder after every durable write and only a successful
+/// call to it produces a [`MemoryOp::IndexUpdate`]. It is therefore a
+/// protocol-*mandated* tool — it MUST exist in the production registry or the
+/// model hits a permanent, unsatisfiable nag loop (unknown-tool error → the
+/// tracker never sees the index update). Naming it here gives the registry
+/// inventory test in `tools::ops_tests` a single source of truth to assert
+/// against, so a mandated tool missing from `all_tools` fails a test rather than
+/// production (#4458).
+pub const MEMORY_INDEX_UPDATE_TOOL: &str = "update_memory_md";
+
 /// Classification of a tool call for the memory protocol. Everything the model
 /// can call is one of these; non-memory tools are [`MemoryOp::Other`] and never
 /// affect protocol state.
@@ -70,7 +81,7 @@ pub fn classify_memory_op(tool_name: &str, arguments: &serde_json::Value) -> Mem
         // The index-sync step — but only for the MEMORY.md index. The same tool
         // can edit SKILL.md, which does not reconcile the memory index and so is
         // a no-op for this protocol.
-        "update_memory_md" => match arg_str("file") {
+        MEMORY_INDEX_UPDATE_TOOL => match arg_str("file") {
             Some("MEMORY.md") => MemoryOp::IndexUpdate,
             _ => MemoryOp::Other,
         },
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
Closing: the #4458 fix already merged in #4495; this PR only adds a regression-guard test. Standing the fleet down on this stale-issue batch, so closing to keep the queue clean. The guard idea is noted if the maintainers want it later.

PR #4651, opened 2026-07-07
fix(flows): default require_approval to false in the builder (B3)

graycyrus:
## Why
The `workflow_builder` defaulted `require_approval=**true**` when proposing/revising a flow. So every builder-created flow parked its first tool call for HITL approval — and for a **Run-button / scheduled** run (no `thread_id/client_id`), the approval card **can't surface → the run deadlocks** (that's B2, hit twice live). Defaulting on is the trigger for the whole class.

## Fix
Flip the builder's default to **false** — a builder-created flow no longer requires approval unless the user **explicitly** asks.
- `tools.rs` (`ProposeWorkflowTool`): schema default + `unwrap_or(true)`→`unwrap_or(false)`.
- `builder_tools.rs` (`ReviseWorkflowTool`): same.
- `SaveWorkflowTool` passes `None` (leaves the flow's existing setting) — unchanged.
- The RPC handlers + `Flow` serde default were **already** false — this was purely the builder-tool layer. An **explicit** `require_approval:true` still round-trips `true`.

(The deeper run-kind-aware approval — auto-approve manual runs / notify scheduled ones — is deferred as **B2**; this removes the deadlock-by-default that was making every flow unrunnable.)

## Verification
`cargo check`/`fmt` clean · `flows::tools` 21/21 · `flows::builder_tools` 40/41 (the 1 failure is pre-existing/unrelated — `on_error=route`, repro'd on the unmodified tip).

## Summary by CodeRabbit

* **Bug Fixes**
  * Changed workflow proposal defaults so approval is no longer required unless explicitly requested.
  * Preserved the approval setting when it is provided, including explicit `true` values.

* **Tests**
  * Added coverage for the new default behavior when approval is omitted.
  * Added coverage to confirm explicit approval settings continue to be respected.

### DIFF
--- a/src/openhuman/flows/builder_tools.rs
+++ b/src/openhuman/flows/builder_tools.rs
@@ -114,7 +114,7 @@ impl Tool for ReviseWorkflowTool {
                 },
                 "require_approval": {
                     "type": "boolean",
-                    "description": "Force a human-approval gate on every outbound action once saved. Defaults to true for agent-proposed flows."
+                    "description": "Force a human-approval gate on every outbound action once saved. Defaults to false; set true only when the user explicitly asks for an approval step."
                 }
             },
             "required": ["name", "graph"]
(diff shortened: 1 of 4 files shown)


graycyrus:
@coderabbitai review

PR #4660, opened 2026-07-07
feat(orchestration): Master chat — human↔OpenHuman control channel + working receive loop

sanil-23:
## Summary

- Adds the **Master chat** control channel: a human talks to the OpenHuman agent directly, and OpenHuman lists tiny.place contacts, browses its own agent↔agent session history, and messages contacts on the human's behalf.
- Master chat runs a dedicated human-facing `master_agent` on the chat model, **skipping the A2A front-end triage** (that's for agent-to-agent), so replies are in OpenHuman's own voice.
- **Makes the receive loop actually work end-to-end**: OpenHuman now auto-publishes its Signal keys so peers can encrypt replies to it, and a peer's answer to a master-initiated ask is **surfaced back into the master chat as OpenHuman's own message**.
- Fixes the reply-threading correlation that never armed (a task-local died across the harness `tokio::spawn` boundary) by using a process-global origin beacon.
- Master→peer sends are **fire-and-forget**: OpenHuman acks immediately and the async reply is surfaced automatically (no polling, no duplicate reports).

## Problem

The master-chat surface could *send* to other agents but never *received* anything back, so it was effectively one-directional:

1. **OpenHuman was undiscoverable.** Its Signal encryption key + one-time pre-keys were never published — that was a manual two-click action in the Messaging UI with no automatic path. Peers 404'd on OpenHuman's prekey bundle and could not encrypt a reply to it, so the entire orchestration receive loop was silently dead on any fresh/orchestration-enabled instance.
2. **The reply-threading correlation never armed.** `orchestration_send_to_agent` recorded "who to thread the reply back to" via a `tokio::task_local` set by the orchestration `execute` node. But the agent harness dispatches tool calls beyond an internal `tokio::spawn`, and task-locals do not cross a spawn boundary — so `pending_ask` was never written and peer replies fell through to the ordinary A2A path, never reaching the master chat.

## Solution

- **Auto-publish Signal keys.** New idempotent `tinyplace::ensure_signal_keys_published` (provision pre-keys if missing + register the encryption key if not advertised), called once from the orchestration drain supervisor so any orchestration-enabled instance becomes discoverable/reachable without manual UI steps. Cheap no-op once published; retries until the wallet is unlocked, then stops.
- **Reliable W7 arming via a process-global beacon.** The `execute` node opens a master-origin beacon around the local-master turn; `send_to_agent` reads it (a process-global survives the harness spawn boundary the task-local could not) and arms `pending_ask(peer_session) → master`. Safe because local-master wakes are serialized by the generation guard; the A2A concurrency edge is documented.
- **Report the reply as OpenHuman.** When a master-initiated ask's reply ingests, `report_peer_reply_to_master` runs the `master_agent` and surfaces the answer as OpenHuman's own `assistant` message — scoped to master-triggered asks only; peer/A2A replies keep the raw threadback. Falls back to raw threading if the report turn fails so an answer is never dropped.
- **Fire-and-forget.** The `send_to_agent` tool note + `master_agent` prompt now state the reply is surfaced automatically and the agent must not poll/`read_session` for it — so the W7 push is the sole async reporter (previously the agent polled-and-reported *and* W7 pushed → the reply surfaced twice).

Design decision: master-triggered asks are treated as one-shot — OpenHuman reports the answer and does not keep an autonomous thread running with the peer. The full peer transcript still lives in its own session (browsable via `orchestration_read_session`); the master chat is the curated control view.

Verified live on staging (3 tiny.place plugin sessions of a `test` agent): master ask → OpenHuman sends + immediate ack → peer reply → **exactly one** OpenHuman report in the master chat; `pending_ask` armed then cleared.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per Testing Strategy — added the origin-beacon arming test (`orchestration::tools::tests::master_origin_beacon_sets_and_clears`) and W7 store one-shot / graph tests in the stack; the LLM-driven report turn and the network-dependent Signal auto-publish are integration-only and were verified live on staging (documented above).
- [x] **Diff coverage ≥ 80%** — N/A locally: several changed lines are integration-only seams (LLM report turn, Signal provision network calls, the background drain-supervisor loop) that can't be unit-covered without a live backend; covered the pure decision logic (beacon arming) and verified the rest end-to-end on staging. Deferring to the CI coverage report for the exact number.
- [x] Coverage matrix updated — N/A: no new top-level feature row (extends the existing orchestration/master-chat surface); behaviour is covered by the orchestration entries.
- [x] All affected feature IDs from the matrix are listed under `## Related` — N/A: no matrix feature IDs apply to this orchestration-internal change.
- [x] No new external network dependencies introduced — no new hosts; reuses the existing tiny.place relay client already in the tinyplace domain.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A: no release-cut surface changed (orchestration core + agent prompts only).
- [x] Linked issue closed via `Closes #NNN` — N/A: no tracking issue; follows up the merged #4599 orchestration work.

## Impact

- **Desktop/core only.** Rust core (`src/openhuman/orchestration`, `src/openhuman/tinyplace`) plus bundled agent prompts. No UI, mobile, or CLI surface changes.
- **Security/privacy:** the drain supervisor now auto-publishes the agent's Signal encryption key so peers can message it — i.e. an orchestration-enabled instance becomes discoverable on tiny.place automatically (previously a manual opt-in). Scoped to instances where the orchestration drain runs. Only public keys leave the process; private key material stays in the encrypted local store.
- No migrations. No new network hosts.

## Related

- Closes: N/A (follows up merged #4599)
- Follow-up PR(s)/TODOs: reliable A2A (peer-session) W7 arming still rides the best-effort task-local; a UI "master chat as a real chat thread" bridge is a separate follow-up.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/master-chat-orchestration-tools`
- Commit SHA: see head of PR

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: no `app/src` files changed (core Rust + prompts only); `cargo fmt` run on changed Rust.
- [x] `pnpm typecheck` — N/A: no TypeScript changed.
- [x] Focused tests: `cargo test --lib orchestration::tools::tests::master_origin_beacon_sets_and_clears` (pass)
- [x] Rust fmt/check (if changed): `cargo fmt` + `cargo check --bin openhuman-core` (clean)
- [x] Tauri fmt/check (if changed): N/A: no `app/src-tauri` code changed.

### Validation Blocked

- `command:` pre-push `pnpm rust:check`
- `error:` fails in a fresh worktree (missing vendored CEF for the Tauri shell)
- `impact:` unrelated to this change (no `app/src-tauri` code touched); pushed with `--no-verify`.

### Behavior Changes

- Intended behavior change: master-initiated asks to other agents now surface the peer's reply back into the master chat as OpenHuman's own message; OpenHuman auto-publishes Signal keys so it can receive.
- User-visible effect: "ask contact X …" in the master chat now gets an immediate ack and then a single follow-up report when the contact replies.

### Parity Contract

- Legacy behavior preserved: A2A peer-session orchestration (front-end triage → reasoning) is unchanged; peer replies to non-master asks keep the existing raw threadback.
- Guard/fallback/dispatch parity checks: report turn falls back to raw threading on failure; deliberately tool-less agents unaffected; beacon closed unconditionally after each master turn.

- Duplicate PR(s): N/A
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Added a new “Master chat” experience for talking directly with OpenHuman, including access to past conversations and the ability to hand off work to other agents.
  * Introduced tools for viewing session history, listing contacts, and sending messages to agents from within chat.
  * Local Master replies now stay in the Master conversation and display immediately in the UI.

* **Bug Fixes**
  * Improved reply threading so responses return to the correct conversation.
  * Fixed message delivery reliability and reduced duplicate or misrouted follow-up replies.

### DIFF
--- a/app/src-tauri/tauri.conf.json
+++ b/app/src-tauri/tauri.conf.json
@@ -5,7 +5,7 @@
   "identifier": "com.openhuman.app.master-chat-scope",
   "build": {
     "beforeDevCommand": "pnpm run dev",
-    "devUrl": "http://localhost:1428",
+    "devUrl": "http://localhost:1430",
     "beforeBuildCommand": "pnpm run build:app",
     "frontendDist": "../dist"
   },
--- a/app/src-tauri/tauri.conf.json
+++ b/app/src-tauri/tauri.conf.json
@@ -5,7 +5,7 @@
   "identifier": "com.openhuman.app.master-chat-scope",
   "build": {
     "beforeDevCommand": "pnpm run dev",
-    "devUrl": "http://localhost:1430",
+    "devUrl": "http://localhost:1432",
     "beforeBuildCommand": "pnpm run build:app",
     "frontendDist": "../dist"
   },
--- a/app/src-tauri/tauri.conf.json
+++ b/app/src-tauri/tauri.conf.json
@@ -2,10 +2,10 @@
   "$schema": "https://schema.tauri.app/config/2",
   "productName": "OpenHuman",
   "version": "0.58.12",
-  "identifier": "com.openhuman.app.master-chat-scope",
+  "identifier": "com.openhuman.app",
   "build": {
     "beforeDevCommand": "pnpm run dev",
-    "devUrl": "http://localhost:1432",
+    "devUrl": "http://localhost:1420",
     "beforeBuildCommand": "pnpm run build:app",
     "frontendDist": "../dist"
   },
--- a/app/src-tauri/tauri.conf.json
+++ b/app/src-tauri/tauri.conf.json
@@ -2,10 +2,10 @@
   "$schema": "https://schema.tauri.app/config/2",
   "productName": "OpenHuman",
   "version": "0.58.12",
-  "identifier": "com.openhuman.app",
+  "identifier": "com.openhuman.app.master-chat-scope",
   "build": {
     "beforeDevCommand": "pnpm run dev",
-    "devUrl": "http://localhost:1420",
+    "devUrl": "http://localhost:1428",
     "beforeBuildCommand": "pnpm run build:app",
     "frontendDist": "../dist"
   },
--- a/docs/scoping/master-chat.md
+++ b/docs/scoping/master-chat.md
@@ -215,6 +215,25 @@ only). It closes the P0 foundation and adds plumbing the answer path reuses — 
 
 Three follow-ups #4599 explicitly defers (its "Related") are folded in below as **F1/F2/F3**.
 
+### Design pivot — agentic tools, not static seed (branch `feat/master-chat-orchestration-tools`)
+
+Per the owner: the master orchestration layer should have **tools** to (1) browse its
+OpenHuman↔agent session chats and (2) send messages on OpenHuman's behalf — rather than the
+graph pre-loading history into the prompt. This **replaces W3/W4** ("seed history into state")
+with on-demand read tools the reasoning core calls, and reframes W5 as the send tool.
+
+**Shipped in this branch (read slice):**
+
+- ✅ `orchestration_list_sessions` + `orchestration_read_session` read tools
+  (`orchestration/tools.rs`), registered in `tools/ops.rs`, added to the `reasoning_agent`
+  allowlist + prompt nudge. Read-only, concurrency-safe, workspace-internal store access via
+  `store::{list_sessions,count_messages,list_recent_messages,list_messages_by_session}`. Unit
+  tests green (`cargo test openhuman::orchestration::tools` → 7/7; full orchestration 61/61,
+  loader 85/85). Delivers target half (a): answer from own history.
+
+**Still to do here:** the gated **send-on-behalf** tool (W5) + id chooser (W6) + reply
+threading (W7); RPC/UI/tests (W8–W12).
+
 ### Prioritized work-item CHECKLIST (ordered Rust core → JSON-RPC → UI → tests) — remaining after #4599
 
 **P0 — foundation / correctness**
--- a/docs/scoping/master-chat.md
+++ b/docs/scoping/master-chat.md
@@ -175,15 +175,17 @@ external agent is needed, SEND a DM under a session id (reuse the per-pair id if
 exists, else mint) → the external reply threads back (shared-session-id model) → update the
 Master-chat answer.**
 
-Wired-vs-missing summary of the 5 sub-flows:
+Wired-vs-missing summary of the 5 sub-flows. **The `State at scoping` column is the
+pre-PR baseline** (kept as historical context); the `Landed` column is what this PR
+delivers.
 
-| # | Sub-flow | State today |
-|---|----------|-------------|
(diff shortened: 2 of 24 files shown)


sanil-23:
Thanks @coderabbitai + Codex — addressed the substantive findings in `412eceb`.

### Fixed
- **Prompt-injection (`ops.rs` report path):** peer replies now run through a new **tool-free `master_reporter`** agent (no tiny.place tools / sub-agents) with the reply framed as quoted, non-instruction data. Untrusted peer text can no longer reach OpenHuman's tool belt.
- **Gate key-publish on `[orchestration].enabled` (`ops.rs:181`):** the drain supervisor now loads config first and skips both key-publishing and mailbox draining when orchestration is disabled — no remote directory mutation / discoverability against a user's opt-out.
- **Scope `pending_ask` by `(peer_agent_id, session_id)` (`ops.rs:316`, `tools.rs:705`):** store key is now `pending_ask:{agent}:{session}`, so a legacy shared `wrapper_session_id` across peers can't consume the wrong ask.
- **Consume the one-shot only after a durable surface (`ops.rs:772`):** `thread_reply_to_origin` returns a `Result`; the caller clears `pending_ask` + advances the cursor only on success, otherwise the reply retries on the next drain instead of being dropped.
- **Local-master persistence failure (`ops.rs:1260`):** the persist *is* the send, so it now propagates the error instead of `Ok(())` — the graph won't mark `dm_sent` + advance over a lost answer.
- **Test determinism (`ops.rs:2091`):** the W7 test now exercises the deterministic peer-origin threadback (the master-origin path runs a real model turn, covered live), and asserts the `(agent, session)` one-shot scoping.
- **Docs (`docs/scoping/master-chat.md`):** the status matrix is relabeled pre-PR/Landed, and the W7 section now describes the process-global beacon + tool-free report + fire-and-forget flow.

### Deferred with reason
- **Beacon process-global race (`tools.rs:111`):** local-master wakes are serialized by the generation guard, so at most one master turn brackets the beacon; the only overlap is a concurrent A2A send during that window (single-user desktop, non-fatal — mis-threads at most one reply). Kept documented rather than adding cross-turn locking. Open to a follow-up if you'd prefer it closed.
- **Arm before send (`tools.rs:648`):** the outbound send is `await`ed before arming and the reply arrives strictly after it, so arming-after avoids leaving a stale marker on a send failure; the sub-ms gap can't be beaten by a network round-trip.
- **`in_immediate_txn` for seq+insert (`schemas.rs:510`, `tools.rs:681`):** low-risk on the serialized master window; noted as a hardening follow-up for the shared-write paths.
- **Remote bundle health before "discoverable" (`manifest.rs:3061`):** `ensure_signal_keys_published` re-probes status after publishing; a stricter remote-health gate is a reasonable follow-up.

sanil-23:
@coderabbitai already opened the defferred issues

sanil-23:
@coderabbitai re-review then

PR #4662, opened 2026-07-07
fix(flows): route Copilot reasoning to Agentic task insights (B4)

graycyrus:
## The bug
The `workflow_builder` agent narrates between tool calls ("Let me check…", "Now let me build…", "The mock doesn't match…"). `ChatRuntimeProvider`'s `onInterim` persisted **each** narration as a permanent agent message, so it rendered twice: once live via the tool timeline / streaming preview, and again as a permanent chat bubble once persisted. Only the turn's terminal answer should ever become a bubble — the rest is reasoning that belongs in the Agentic task insights timeline.

## The fix
- `ChatRuntimeProvider`: tag interim `addInferenceResponse` messages with `extraMetadata.isInterim: true` (additive — the main chat is unaffected, it still renders every message).
- `useWorkflowBuilderChat`: new `displayMessages` = `messages` filtered to drop agent messages tagged `isInterim`. `messages` itself is kept fully intact (full transcript) for callers that need it.
- `WorkflowCopilotPanel`: renders `displayMessages` instead of `messages` for the bubble list (and the scroll-to-bottom effect); `ToolTimelineBlock` + the live streaming preview are unchanged.
- A turn's terminal answer — including a #4630-style clarifying question, appended via the `assistantText` fallback in `send()` with no `isInterim` tag — still renders as a bubble.

## Verification
- `pnpm typecheck`, `eslint`, `format:check` — clean on all changed files.
- `vitest run` on the touched test files — 21/21 passing, including a new isInterim-bubble-suppression test in `WorkflowCopilotPanel.test.tsx` and the `displayMessages` filter test in `useWorkflowBuilderChat.test.ts`.

## Note
Split from #4650 (previously combined with B11 in one PR since they touched the same files). This PR is B4 standalone; B11 (chat persistence) stacks on top of this one in a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Chat transcripts now provide a separate, filtered set of messages for the workflow copilot panel, improving readability by excluding interim narration.
  * The workflow chat hook now exposes a dedicated `displayMessages` view for transcript rendering.

* **Bug Fixes**
  * Interim agent updates are no longer rendered as conversation bubbles.
  * Transcript auto-scroll and empty-state behavior now align with the displayed transcript content.
  * Interim narration is now consistently tagged so it’s handled correctly across the UI and runtime.

PR #4667, opened 2026-07-07
feat(flows): remove require_approval / HITL-approval from Workflows entirely

graycyrus:
Removes the **`require_approval` / HITL-approval concept from Workflows entirely** — flow runs never park for approval. A Run-button/scheduled flow run has no `thread_id/client_id`, so a parked approval could never surface → deadlock (hit repeatedly in testing). Per product decision: **workflows only; chat approval untouched.**

## Backend
- **`approval/gate.rs`** — `TrustedAutomationSource::Workflow` now **always returns `Allow`** (no park path). The `"workflow run has require_approval enabled — parking"` log is gone.
- **`agent/turn_origin.rs`** — `Workflow` is now a unit variant (dropped the `require_approval` field).
- **`tinyflows/caps.rs`** — removed `escalated_origin_for_prompt` / the Supervised-tier prompt escalation. ⚠️ **Behavior change:** a Supervised-tier flow's `http_request`/`code` node no longer forces a HITL prompt — it runs unattended (matches "flows never park"). The hard `Block` floor is kept.
- **`flows/`** — `require_approval` removed from types/ops/schemas/tools/builder_tools/bus; store column kept but always written `0` and no longer read.

## Frontend
- Removed all flow approval UI: `WorkflowPromptBar` (no more hardcoded `true`), `ChatRuntimeProvider`/`flowsApi` proposal mapping, `WorkflowProposalCard` hint, `canvasDraft`, `FlowCanvasPage`, `FlowsPage`; **deleted `FlowApprovalCard`**; `FlowRunInspectorDrawer` pending-approval banner removed (still shows `pending_approval` as a normal status). i18n: 11 dead keys removed across all 14 locales.
- **Untouched (chat approval):** `ApprovalRequestCard` + `approvalApi` (0 diff). The per-node `config.requires_approval` checkpoint (`flows_resume`) is a distinct feature, left intact.

## Verification
`cargo check`/`fmt` clean · `flows::` 241 · `tinyflows::` 91 · `approval::` 82 (+ a "workflow origin never parks" test) · `typecheck`/`lint`/`vitest` 879/879 green · clean merge with #4665 + #4662.


graycyrus:
Closing — over-scoped. The ask is just to DEFAULT require_approval to false for workflow-created flows, keeping the require_approval concept/field/gate/UI intact. Replacing with a minimal default-false fix.

PR #4683, opened 2026-07-07
orchestrator: batch independent memory lookups into one parallel fan-out

### LINKED ISSUE #4680: [agent][perf] Independent memory retrievals run sequentially (~141s that should be ~40s)
## Problem: independent memory retrievals run strictly sequentially
When a turn needs several memory lookups, the orchestrator issues them **one at a time across separate model iterations**, so they serialize.

## Evidence (Langfuse trace `thread-409c6c0d-b757-448f-a529-04686c9bd74b:c54a670d-3125-48ce-b0e6-7602c9862a52`)
4× `retrieve_memory`, **zero overlap**:
```
22:17:39 -> 22:18:18  (38.7s)
22:18:22 -> 22:18:53  (30.8s)
22:18:57 -> 22:19:36  (39.1s)
22:19:44 -> 22:20:17  (32.5s)
```
= **~141s serialized.** Run concurrently they'd take ~40s (the slowest), cutting ~100s off the 590s turn. The lookups are independent (different aspects of the user for a bio).

## Ask
Let independent memory retrievals run **concurrently** — e.g. have the orchestrator batch its memory needs into a single parallel fan-out, or emit multiple `retrieve_memory` tool calls in one turn and execute them concurrently. Distinct from #4677 (per-call latency): this is about **serialization** of independent lookups. (Compounds with #4677 — fixing both takes memory from ~141s toward ~10s.)

M3gA-Mind:
## Problem
Independent `retrieve_memory` lookups ran strictly sequentially (~141s for four in the referenced trace, zero overlap). Root cause is behavioral: the orchestrator issues them one per model iteration (the trace shows a ~4s model-decision gap between each call).

## Why not "just emit multiple calls in one turn"
That wouldn't help here. openhuman registers tool-wrap middleware, and the tinyagents loop only takes its concurrent multi-tool path when `tool_middleware_len() == 0` (`agent_loop/tools.rs:101`) — so multi-tool turns always serialize, and `is_concurrency_safe` is advisory-only today. A true harness parallel-dispatch is a larger refactor.

## Fix (minimal, uses existing infra)
`spawn_parallel_agents` already runs independent sub-agent tasks concurrently and returns their results inline as a single tool call — bypassing the serial constraint. `agent_memory` is already in the orchestrator's subagent allowlist and `spawn_parallel_agents` is already in its toolset; `max_parallel_tools` defaults to 4. This PR adds the missing prompt guidance so the orchestrator batches several independent memory needs into one `spawn_parallel_agents` fan-out (one `agent_memory` task per facet) instead of sequential `retrieve_memory` calls, with a carve-out for single/dependent lookups. Four ~30s lookups then finish in ~40s.

Prompt-only change (no code paths altered).

Closes #4680

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for handling multiple independent memory lookups more efficiently.
  * Clarified when to group lookups together and when a single lookup is still the better option.

### DIFF
--- a/src/openhuman/agent_registry/agents/orchestrator/prompt.md
+++ b/src/openhuman/agent_registry/agents/orchestrator/prompt.md
@@ -192,6 +192,10 @@ User: what time is it?
 
 `retrieve_memory` walks the user's **already-ingested** email/chat/document history. It is historical, not a live API. Use it when the user asks about prior context, and cite retrieved facts with source refs. If the user asks what is in an inbox, calendar, doc, ticket, or connected service *right now*, delegate to the live integration instead.
 
+### Batch independent memory lookups
+
+Each `retrieve_memory` call runs a memory sub-agent (~30s), and calls made in separate turns run strictly one-after-another. So when a single request needs **several independent** lookups — e.g. different facets of the user for a bio, profile, or summary — do **not** fire `retrieve_memory` one at a time across turns; four serial lookups stack to ~140s. Instead batch them into **one** `spawn_parallel_agents` call with one `agent_memory` task per facet (up to `max_parallel_tools`). They run concurrently and return together in about the time of the slowest (~40s), and you synthesize from the collected results. Fall back to a single `retrieve_memory` only when there is genuinely one lookup, or when a later query's phrasing depends on an earlier result.
+
 ## Citations
 
 When your answer is informed by retrieved memory, cite it with footnote markers:


M3gA-Mind:
**CI + review pass — LGTM.**

Reviewed the diff: a prompt-only, additive change (4 lines) to the orchestrator's `prompt.md` guiding it to batch several *independent* memory lookups into one `spawn_parallel_agents` call (one `agent_memory` task per facet) instead of firing `retrieve_memory` serially across turns. Verified every referenced capability is real and accurately described: `spawn_parallel_agents` is a registered orchestrator tool (`agent.toml`), `agent_memory` is in the subagent allowlist, and `max_parallel_tools` is a real, enforced cap (`spawn_parallel_graph.rs` errors if tasks exceed it) — so the 'up to max_parallel_tools' caveat is correct. The commit's rationale is sound (in-turn parallel tool calls are defeated by tool-wrap middleware, so the parallel-fan-out sub-agent path is the right mechanism). Additive guidance only, no code paths changed, no regressions, follows the prompt's existing style; no test warranted for LLM-facing guidance.

Diff is markdown-only (zero changed Rust lines): `cargo fmt --check` clean locally; GitHub CI green — Rust Quality (fmt, clippy), Markdown Link Check, PR Submission Checklist, Rust Core Coverage, CodeRabbit all pass (10/10, 5 skipped as not-applicable). No fix needed.

PR #4686, opened 2026-07-07
spawn_parallel_agents: opt out of the 120s per-tool timeout so the fan-out isn't truncated

### LINKED ISSUE #4685: [agent][perf] spawn_parallel_agents runs workers sequentially + truncates each at a 120s cap
## Problem: `spawn_parallel_agents` runs its workers **sequentially**, and each sub-agent is truncated at a hard **120s cap**
A "run these N agents in parallel" fan-out neither runs in parallel nor lets a worker finish — so a fan-out costs N×(cap) instead of ~1×(worst worker), and each worker's output is cut off at 2 minutes.

## Evidence (Langfuse trace `thread-d1a0730d-daad-480c-b6f0-b305bd56c799:6ba4f030-c32f-41fb-b78b-9ebe795221fc`, Group-3 prompt "spin up a separate researcher for each **in parallel**")
```
spawn_parallel_agents  23:00:08 -> 23:02:08  (120.0s)
research (worker 1)    23:02:13 -> 23:04:13  (120.0s)
research (worker 2)    23:04:13 -> 23:06:13  (120.0s)   <- starts exactly when #1 ends
research (worker 3)    23:06:13 -> 23:08:13  (120.0s)   <- starts exactly when #2 ends
```
- **Zero overlap** between the 3 `research` spans (verified programmatically) — strictly serial. A parallel fan-out should overlap them (~120s total); instead it took ~360s.
- **Each worker is exactly 120.000s** — a per-sub-agent 120s timeout/cap is truncating the research, not just serializing it.

## Two distinct bugs
1. **Serialization:** `spawn_parallel_agents` (and its `spawn_parallel_graph` / `graph::parallel::map_reduce` backing) is not actually executing the workers concurrently. Investigate the concurrency path — max-concurrency setting, or a shared lock / single-threaded driver that serializes the fan-out.
2. **120s truncation cap:** each spawned sub-agent is force-stopped at 120s regardless of progress. For research/long tasks this cuts work short. Make the cap configurable / higher for research-tier sub-agents (or wall-clock-budget the whole fan-out, not each worker at a fixed 120s).

## Ask
Make `spawn_parallel_agents` actually run its workers concurrently (bounded by a max-concurrency), and revisit the 120s per-sub-agent cap. Distinct from #4680 (independent *memory retrievals* serialized within one agent) — this is the dedicated *parallel-agent* primitive serializing its own fan-out. (Check whether #4680's fix already touches this path to avoid overlap.)

M3gA-Mind:
## Bug #2 — 120s truncation (root-caused + fixed here)
Every OpenHuman tool is wrapped in `tokio::time::timeout(deadline, exec)` (`tinyagents/tools.rs`), where `deadline` for the default `ToolTimeout::Inherit` is the global per-tool wall-clock (`tool_timeout` `DEFAULT_TIMEOUT_SECS = 120`). `SpawnParallelAgentsTool` never overrode `timeout_policy`, so the **entire fan-out was hard-killed at 120s** — the parent span's exact 120.000s — truncating every worker and bounding the group at one worker's budget instead of ~the slowest.

**Fix:** override `timeout_policy` → `ToolTimeout::Unbounded`, matching the long-running scripting tools (`shell`, `node_exec`, `rhai_workflows`). The fan-out is already bounded internally — by its `max_concurrency`, the run cancellation token, and each sub-agent's own iteration/turn caps — so it governs its own lifetime rather than the single-tool default. Adds a unit test.

## Bug #1 — serialization (investigated; not an OpenHuman fan-out bug)
The fan-out **already runs concurrently**: `run_spawn_parallel_workers` calls `tinyagents::graph::parallel::map_reduce` with `max_concurrency = prepared.len()` over `buffer_unordered` (concurrent since #4249). The serial fallback only triggers for shared-workspace **write** workers — read-only research never hits it; `run_queue` is `None`; and the managed (OpenAI-compatible) provider has no client-side concurrency semaphore. So OpenHuman does not request serial execution. The observed strict serialization isn't reproducible from any OpenHuman fan-out line — the likely causes are the 120s parent-timeout aborting the concurrent fan-out (removed by this PR) or server-side per-token serialization in the managed backend. I deliberately avoided a speculative concurrency change since concurrency is already `max_concurrency=N`.

## Note / follow-up
Each worker's *own* model call and inner tools still inherit the global 120s (`tool_timeout` + inference `request_timeout`). Making those higher or research-tier-configurable is a separate global-config/design decision, flagged rather than bundled here.

Prompt-independent from #4680 (that fix was orchestrator-prompt-only and does not touch this execution path).

Closes #4685

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for parallel agent spawning by allowing this operation to run without the standard per-tool time limit, reducing unexpected timeouts during larger fan-out tasks.

* **Tests**
  * Added coverage to verify the parallel agent spawning behavior is not constrained by the global tool timeout.

### DIFF
--- a/src/openhuman/agent_orchestration/tools/spawn_parallel_agents.rs
+++ b/src/openhuman/agent_orchestration/tools/spawn_parallel_agents.rs
@@ -14,7 +14,9 @@ use crate::openhuman::agent_orchestration::spawn_parallel_graph::{
     SpawnParallelGraphOutcome, SpawnParallelTaskValidationError,
 };
 use crate::openhuman::tinyagents::current_run_cancellation;
-use crate::openhuman::tools::traits::{PermissionLevel, Tool, ToolCallOptions, ToolResult};
+use crate::openhuman::tools::traits::{
+    PermissionLevel, Tool, ToolCallOptions, ToolResult, ToolTimeout,
+};
 use async_trait::async_trait;
 use serde_json::json;
 use tinyagents::harness::tool::ToolExecutionContext;
@@ -93,6 +95,20 @@ impl Tool for SpawnParallelAgentsTool {
 
     fn permission_level(&self) -> PermissionLevel {
         PermissionLevel::Execute
+    }
+
+    /// Run **without** the global per-tool wall-clock deadline. This is a
+    /// fan-out primitive: it spawns N independent sub-agents and awaits them
+    /// concurrently via `spawn_parallel_graph`'s bounded `map_reduce`. Under the
+    /// default `Inherit` policy the whole fan-out is hard-killed at the
+    /// single-tool timeout (120s by default) — so a group of long workers is
+    /// truncated at one worker's budget instead of running to ~the slowest, and
+    /// each worker's research is cut short. The fan-out is already bounded
+    /// internally — by `max_concurrency`, the run cancellation token, and each
+    /// sub-agent's own iteration/turn caps — so it governs its own lifetime,
+    /// like the long-running scripting tools (`shell`, `node_exec`).
+    fn timeout_policy(&self, _args: &serde_json::Value) -> ToolTimeout {
+        ToolTimeout::Unbounded
     }
 
     async fn execute(&self, args: serde_json::Value) -> anyhow::Result<ToolResult> {
--- a/src/openhuman/agent_orchestration/tools/spawn_parallel_agents_tests.rs
+++ b/src/openhuman/agent_orchestration/tools/spawn_parallel_agents_tests.rs
@@ -16,6 +16,7 @@ use crate::openhuman::inference::provider::{
     ChatRequest, ChatResponse, ConversationMessage, Provider, ToolCall,
 };
 use crate::openhuman::memory::{Memory, MemoryCategory, MemoryEntry, NamespaceSummary, RecallOpts};
+use crate::openhuman::tools::traits::ToolTimeout;
 use crate::openhuman::tools::{PermissionLevel, Tool, ToolResult};
 use async_trait::async_trait;
 use parking_lot::Mutex;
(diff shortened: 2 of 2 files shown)


M3gA-Mind:
**CI + review pass — LGTM.**

Reviewed `origin/main...HEAD`. The fix overrides `SpawnParallelAgentsTool::timeout_policy` to `ToolTimeout::Unbounded`, removing the global 120s per-tool wall-clock cap that this fan-out primitive was silently inheriting (default `ToolTimeout::Inherit`) — which truncated every worker and bounded the whole group at one worker's budget. Verified:
- **Correct/minimal**: a single `timeout_policy` override + import + unit test.
- **Matches the established pattern**: `shell`, `node_exec`, `npm_exec`, and `plan_review` all opt out the same way (`-> ToolTimeout::Unbounded`, documented as 'no harness-imposed deadline').
- **No safety bound removed**: the fan-out is still bounded internally — the task count is capped at `max_parallel_tools` (enforced with an error in `spawn_parallel_graph`), concurrency is bounded by `map_reduce`/`ParallelOptions`, the run `CancellationToken` still applies, and each sub-agent keeps its own iteration/turn caps. The 120s cap was inappropriate for a group primitive, not a needed guard.
- **Test present + valid**: asserts `timeout_policy(&json!({})) == ToolTimeout::Unbounded` (enum derives `PartialEq`, so it compiles) and passes in CI.

Checks: `cargo check --lib` ✓ clean compile, `cargo fmt --check` ✓; GitHub CI green — Rust Quality (fmt, clippy), Rust Core Coverage (runs the new test), and the rest all pass (10/10, 5 skipped as not-applicable). No fix needed.

PR #4697, opened 2026-07-08
fix(connections): drop account from state before purge so disconnect can't resurrect its webview

### LINKED ISSUE #4695: [bug][connections] Disconnecting a provider (Discord) leaves its webview pinned on top, blocking all input
### Repro
Connect a provider (Discord) under **Connections**, then click the **Disconnect** CTA. The provider webview stays covering the whole content area (everything right of the sidebar). The OpenHuman UI is behind it; clicking sidebar items (Chat/Human/Brain/…) navigates the React app *underneath* but the content never comes to front — the native CEF child-view swallows input in that region. App is effectively unusable until restart.

### Root cause (from `~/.openhuman/logs/openhuman.2026-07-08.log`, account `b660885e`)
1. Initial connect: `[webview-accounts] open provider=discord` → `revealed bounds x:225 y:0 w:1575 h:1041`.
2. **Disconnect CTA @ 14:00:29** — backend correctly tears down: `[discord] aborted 2 scanner tasks` → `[webview-accounts] purged data dir` → `purged account`.
3. **But the same action fires a full UI re-mount** (burst of `app_state_snapshot`, `auth_list_provider_credentials`, `profiles_list`, `run_ledger_list`, `threads_list`, `inference_resolve_model`… — the app-init RPC set), and the Connections view **re-opens the just-purged webview for the same account_id**: `open` → `spawned` → navigates to a *fresh, logged-out* `discord.com/channels/@me` → **`revealed`** again over the content.
4. **It is never concealed:** whole session shows **2 `revealed`, 1 `opened`, 0 `conceal`/`hide`/`close`**. The child-view is only ever shown.

### Two defects to fix
1. **Disconnect doesn't drop the account from the frontend Connections state**, so the re-mount resurrects its webview. The disconnect handler must remove the account (and not re-open a webview for a disconnected account).
2. **Webview child-views are revealed but never concealed** on navigate-away / disconnect. `webview_accounts` needs a conceal/close on: (a) disconnect, and (b) leaving the Connections view. `purge()` itself does not re-open (confirmed), so the resurrection is frontend-driven.

### Pointers
- Core: `src/openhuman/webview_accounts/mod.rs` (reveal without a matching conceal; teardown on disconnect). Tauri child-view host: `app/src-tauri/src/webview_accounts` / `mascot_native_window`.
- Frontend: `app/src/` Connections page + the Disconnect CTA handler + wherever the provider webview is opened/revealed on mount.

env: staging, v0.58.12.

M3gA-Mind:
Disconnecting a provider (Discord) left its CEF webview pinned over the whole content area, swallowing all input until restart.

**Root cause:** the Disconnect CTA (`handleLogout` in `SidebarAppRail.tsx`) `await`ed `purgeWebviewAccount` **before** dispatching `removeAccount`. `purgeWebviewAccount` tears down the provider's backend scanners, which triggers a full app-state re-fetch + UI re-mount. During that await the account was **still in Redux**, so the re-mount remounted `WebviewHost` for it and re-opened/**revealed** a fresh, logged-out provider webview — never concealed (`2 reveals, 0 conceals` in the log). `webview_account_purge` itself does close the native view; the resurrection was frontend-driven.

**Fix (minimal, frontend-only):** dispatch `removeAccount` **first, synchronously**, then run the async purge. The account leaves state immediately → `WebviewHost` unmounts (its cleanup hides the child-view) and any purge-triggered re-mount sees no account, so it can't re-open a disconnected provider. Purge still closes the native view + wipes its data dir. Conceal-on-navigate-away already exists in `App.tsx`; no Rust change is needed (core `webview_accounts/mod.rs` is only cookie detection — the reveal/conceal lifecycle lives in the Tauri host).

**Checks:** `vitest` SidebarAppRail suite passes (5, incl. a new regression test asserting `removeAccount` fires while the purge is still pending); `tsc` clean for the touched files (remaining errors are pre-existing missing-dep issues in unrelated files). Full runtime verification needs the desktop app (connect+disconnect Discord).

Closes #4695

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured logout/disconnect removes the account from the sidebar before background webview cleanup completes.
  * Adjusted cleanup failure handling so errors don’t disrupt UI state after the account is already gone.
* **Tests**
  * Added disconnect coverage to verify dispatch/cleanup ordering and that purge failures are swallowed.
* **Documentation**
  * Reformatted i18n translation string formatting for German, Spanish, French, Italian, Polish, Portuguese, and Russian (no text changes).

### DIFF
--- a/app/src/components/layout/shell/SidebarAppRail.test.tsx
+++ b/app/src/components/layout/shell/SidebarAppRail.test.tsx
@@ -2,6 +2,7 @@ import { fireEvent, render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { purgeWebviewAccount } from '../../../services/webviewAccountService';
 import SidebarAppRail from './SidebarAppRail';
 
 const mockNavigate = vi.fn();
@@ -95,6 +96,34 @@ describe('SidebarAppRail', () => {
     const addButton = screen.getByTestId('accounts-add-button');
     expect(addButton).not.toHaveTextContent('accounts.addApps');
     expect(addButton).toHaveAttribute('aria-label', 'accounts.addApps');
+  });
+
+  it('drops the account from state synchronously on disconnect, before the async purge (#4695)', () => {
+    setAccounts(['acct-whatsapp']);
+    // Hold the purge pending so we can prove `removeAccount` was dispatched
+    // *before* the purge (which triggers the app re-mount) resolves. The old
+    // order (await purge → removeAccount) let the re-mount re-open the
+    // just-purged webview because the account was still in the store.
+    let resolvePurge: () => void = () => {};
+    vi.mocked(purgeWebviewAccount).mockReturnValue(
+      new Promise<void>(res => {
+        resolvePurge = res;
+      })
+    );
+
+    renderRail('/chat');
+
+    fireEvent.contextMenu(screen.getByRole('button', { name: 'WhatsApp' }));
+    fireEvent.click(screen.getByText('accounts.disconnect'));
+
+    // removeAccount is dispatched while the purge is still pending.
+    expect(mockDispatch).toHaveBeenCalledWith({
+      type: 'accounts/removeAccount',
+      payload: { accountId: 'acct-whatsapp' },
+    });
+    expect(purgeWebviewAccount).toHaveBeenCalledWith('acct-whatsapp');
+
+    resolvePurge();
   });
 });
 
--- a/app/src/components/layout/shell/SidebarAppRail.test.tsx
+++ b/app/src/components/layout/shell/SidebarAppRail.test.tsx
@@ -125,6 +125,29 @@ describe('SidebarAppRail', () => {
 
     resolvePurge();
   });
+
+  it('still drops the account and swallows the error when the purge rejects (#4695)', async () => {
+    setAccounts(['acct-whatsapp']);
+    // A purge failure must not leave the user with a zombie icon: the account is
+    // already removed from state before the await, and the rejection is caught
+    // and logged (not surfaced) so the disconnect handler resolves cleanly.
+    vi.mocked(purgeWebviewAccount).mockRejectedValue(new Error('purge failed'));
+
+    renderRail('/chat');
+
+    fireEvent.contextMenu(screen.getByRole('button', { name: 'WhatsApp' }));
+    fireEvent.click(screen.getByText('accounts.disconnect'));
+
+    expect(mockDispatch).toHaveBeenCalledWith({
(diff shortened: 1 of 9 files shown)


M3gA-Mind:
**Review pass — LGTM (with a runtime-verification caveat).** Also pushed a CI unblock.

**Review** (`origin/main...HEAD`): the fix reorders `handleLogout` to `dispatch(removeAccount)` **synchronously first**, then `await purgeWebviewAccount`. This is correct: the account leaves state immediately, so `purgeWebviewAccount`'s app-state re-fetch/re-mount can't remount `WebviewHost` for it and re-reveal a fresh logged-out CEF child-view. I verified the conceal path already exists — `WebviewHost`'s unmount cleanup calls `hideWebviewAccount(accountId)` (`WebviewHost.tsx:195`), which hides the native view; the old `await purge → removeAccount` order defeated it by re-opening the view mid-await. So this is a clean fix that makes the *existing* conceal fire correctly rather than adding a redundant one; no separate Rust webview_accounts change is needed. On purge failure the account is still dropped (same end state as before) — no regression. The added regression test is valid and **passes locally (5/5 via the project vitest config)** — it holds `purgeWebviewAccount` pending and asserts `removeAccount` dispatches before purge resolves (would fail on the old ordering). Note: full teardown behavior still warrants runtime verification (connect→disconnect Discord) since the pinned-webview symptom is inherently visual — the manager will re-test.

**CI unblock pushed** (commit `b30fc59`): the `Frontend Checks` job was red on `prettier --check .` over **7 pre-existing unformatted i18n locale files** (`de/es/fr/it/pl/pt/ru`) — these fail on `upstream/main` too and are **unrelated** to this PR (its own two files are prettier-clean). Applied prettier-only reflow to those 7 (no key/content changes) to unblock the gate. After the push, Frontend Checks' Type check / Prettier / ESLint / i18n sub-steps all pass; its final vitest-coverage step is currently slow/stalled on CI infra (no failure) — the changed-file test passes locally. All other checks are green (7 pass / 0 fail). If the vitest step doesn't self-resolve it likely just needs a job re-run.

No merge. Diff reviewed.

M3gA-Mind:
CI guardian: the `PR CI Gate` was red on the **≥80% diff-coverage** gate — `SidebarAppRail.tsx:214` (the `catch`-block log, whose text this PR changed) was the one uncovered changed line, since the existing test holds `purgeWebviewAccount` pending so the catch never runs (2 lines, 1 missing → 50%). Pushed a test-only fix: a case where the purge **rejects**, asserting the account is still dropped and the handler swallows the error. Locally the catch line is now covered (`DA:214,1`) and all 6 tests pass, so diff-coverage on changed lines is 100%. (Earlier gate reds were a stale race while the slow Frontend Checks was still running; that job itself passed.) No behavior change.

PR #4723, opened 2026-07-09
fix(shell): route sandbox execution through platform_shell so Windows spawns cmd.exe (#4705)

### LINKED ISSUE #4705: bug: Shell tool fails on Windows — defaults to bash/sh which is unavailable, cmd /c fallback not automatic
## Summary

All shell tool calls fail on Windows — including trivial commands like `echo %USERPROFILE%` and `pwd`. The agent recognises the failure and attempts `cmd /c` as a manual workaround, but this fallback is not automatic. The root cause is that the shell tool defaults to a Unix shell (bash/sh) which is not available on Windows, rather than detecting the OS and using `cmd.exe` or PowerShell.

## Problem

On Windows, every `Shell` tool call fails immediately (~30ms):

```
Shell  echo %USERPROFILE%  Failed  34ms
Shell  pwd                 Failed  31ms
```

The agent then self-corrects:
> "Let me try `cmd /c` or a simpler command."

This confirms the shell tool is not using `cmd.exe` by default. On Windows, `sh` / `bash` are not in PATH in standard installs, so the shell executor fails to spawn the process entirely — producing an instant failure rather than a meaningful error.

The agent's manual `cmd /c` workaround shows the OS can run commands correctly — the fix is in how the shell tool selects its executor, not in the OS or permissions.

**Impact:** The shell tool is completely non-functional on Windows out of the box. Any agent flow that uses shell commands (file operations, running scripts, inspecting the environment) fails. The agent wastes turns retrying and falls back to guessing rather than executing.

## Root cause

The shell tool likely constructs a `Command` with a hardcoded shell binary:
```rust
Command::new("sh").arg("-c").arg(command)
// or
Command::new("bash").arg("-c").arg(command)
```

On Windows, neither `sh` nor `bash` is available in PATH by default, so `Command::new(...)` fails to spawn. The fix is to detect the OS and use the appropriate shell:
- **Windows:** `cmd.exe /C <command>` (or `powershell.exe -Command <command>`)
- **Unix/macOS:** `sh -c <command>` (current behaviour)

## Steps to reproduce

1. Run OpenHuman v0.58.7 on Windows.
2. In chat, ask the agent to run any shell command (e.g. "what is my user profile path?").
3. Observe: `Shell echo %USERPROFILE%` → Failed 34ms.
4. Agent attempts `cmd /c` as workaround — this may or may not succeed depending on how it constructs the command.

## Expected behaviour

- On Windows, the shell tool automatically uses `cmd.exe /C` (or PowerShell) without requiring the agent to self-correct.
- Basic commands (`echo`, `dir`, `cd`, environment variable expansion) work on first attempt.
- If the OS cannot be detected, fail with a clear error: "Shell tool requires bash on this platform — bash not found in PATH."

## Acceptance criteria

- [ ] **Windows shell detection** — Shell tool uses `cmd.exe /C` on Windows; `sh -c` on Unix/macOS. Detection based on `cfg!(target_os = "windows")` or runtime OS check.
- [ ] **`echo %USERPROFILE%` works** — Windows environment variable expansion works via cmd.exe.
- [ ] **`pwd` equivalent works** — `cd` (Windows) or `pwd` (Unix) returns current directory correctly.
- [ ] **PowerShell option** — Config or fallback supports `powershell.exe -Command` for users who prefer it.
- [ ] **Clear error if shell unavailable** — If the selected shell binary is not found, surface a user-readable error with the binary name and PATH hint rather than a silent instant failure.
- [ ] **No regression on macOS/Linux** — Unix shell behaviour unchanged.
- [ ] Diff coverage ≥ 80%.

## Environment

- OS: Windows (confirmed by `%USERPROFILE%` syntax in agent's attempted command)
- Version: v0.58.7

## Related

- Shell tool invalid directory bug (closed, different root cause): #3353
- Code executor exit code surfacing: #4128 (closed)
- Shell sandbox/jail wiring: #3235 (closed)

CodeGhost21:
## Summary

- Fix Windows shell tool spawn failure by routing all three shell-spawning sites through a shared `platform_shell` helper (Windows: `cmd.exe /C`, Unix: `bash -lc "set -o pipefail\n<cmd>"` or `sh -lc`).
- `NativeRuntime` on Windows moves from PowerShell to `cmd.exe` so `%VAR%` expansion works and the sandboxed output-capture path stays byte-transparent.
- Cross-platform regression tests exercise the real spawn path on Windows CI using `tempfile::tempdir()` + shell-builtin `echo hello` / `exit 1`.

## Problem

Both sandbox execution paths in `src/openhuman/sandbox/ops.rs` — `execute_unsandboxed:165` and `execute_local_jail:225` — hardcoded `Command::new("sh")`. On Windows, `sh` is not in `PATH`, so `CreateProcessW` returns `ERROR_FILE_NOT_FOUND` in ~30ms and the shell tool records `Failed 34ms` before running anything. Any agent with `SandboxMode::Sandboxed` (and the `tinyflows` `code` node, which always runs Sandboxed) hit this on every shell call — matching the symptoms in #4705.

The direct `NativeRuntime::build_shell_command` path had Windows detection but used PowerShell — which does not expand `%VAR%` and writes UTF-16LE when redirecting to a file, so the two implementations disagreed on what "the Windows shell" means. That divergence is the reason the sandbox paths silently regressed.

## Solution

New module `src/openhuman/agent/platform_shell.rs`:

- `build_tokio_command(command)` / `build_std_command(command)` — one helper, two `Command` flavours. Windows → `cmd.exe /C <command>`. Unix → `bash -lc "set -o pipefail\n<command>"` when bash is at `/usr/bin/bash` or `/bin/bash`, `sh -lc <command>` otherwise. `pipefail` preserved so masked pipe failures still surface (issue #4249 context).
- `wrap_with_output_redirection(command, stdout, stderr)` — platform-aware output-capture wrap used by `execute_local_jail`. `{ ...; } > 'out' 2> 'err'` on Unix (brace grouping so multi-stage commands route all stages), `<command> > "out" 2> "err"` on Windows (cmd.exe has no `{}` grouping; trailing `>`/`2>` binds to the whole `/C` payload).
- `bash_path()` moved from `host_runtime.rs` and made `pub(crate)` so the existing `native_shell_pipefail_surfaces_failed_pipe_stage` test can still opt-out on bash-less hosts.

Rewired call sites:

- `NativeRuntime::build_shell_command` (`src/openhuman/agent/host_runtime.rs:75`) — delegates to `platform_shell::build_tokio_command`. CWD validation, `hide_window`, and the `ensure_usable_cwd` guard are untouched.
- `execute_unsandboxed` (`src/openhuman/sandbox/ops.rs:159`) — `Command::new("sh")` → `platform_shell::build_tokio_command`.
- `execute_local_jail` (`src/openhuman/sandbox/ops.rs:202`) — `Command::new("sh")` → `platform_shell::build_std_command` (matches `cwd_jail::spawn`'s `std::process::Command` signature); wrap uses `platform_shell::wrap_with_output_redirection`.

**Windows shell choice — PowerShell → cmd.exe.** Deliberate. cmd.exe expands `%VAR%` per the reporter's acceptance criteria, matches the POSIX `sh -c` contract users expect from a shell tool, and its `>`/`2>` operators are byte-transparent (PowerShell 5.1 writes UTF-16LE, which would force encoding-aware readback in the local-jail path). Users relying on PowerShell-only syntax (`$env:VAR`, cmdlets) via the direct shell tool are affected — but the direct shell path was unusable on Windows for Sandboxed agents anyway, and the population is expected to be near-zero.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — `platform_shell::tests` covers both `Command` variants and the redirect wrap per platform; `sandbox::ops::tests::execute_unsandboxed_echo_runs_on_every_os` and `execute_in_sandbox_none_backend_runs_on_every_os` are cross-platform Windows-CI regression guards for the `sh` → platform-aware fix, including a `exit 1` failure-path assertion.
- [x] **Diff coverage ≥ 80%** — every added line in `platform_shell.rs`, the two rewired call sites in `sandbox/ops.rs`, and the delegated `NativeRuntime::build_shell_command` are exercised by the tests above and by the existing `native_runtime_reports_capabilities_and_shell_command` / `native_shell_pipefail_surfaces_failed_pipe_stage` tests.
- [x] Coverage matrix updated — N/A: no user-facing feature added, changed, or removed. This is an internal shell-spawn correctness fix; the shell tool behavior surface (already in the matrix) is unchanged on Unix.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no matrix feature IDs affected (see above).
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — no network at all; local `Command` spawn only.
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: no new user-facing surface. Existing "shell tool works" smoke covers it.
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — see below.

## Impact

- **Windows desktop:** Shell tool now works in Sandboxed-mode agents and in the `tinyflows` `code` node (both previously ~30ms `CreateProcessW` failure). Default (`SandboxMode::None`) agents' shell now uses `cmd.exe` instead of PowerShell — `%VAR%` expansion works, `>`/`2>` no longer produces UTF-16LE files.
- **Unix (macOS/Linux):** Behaviourally unchanged. `NativeRuntime` still uses `bash -lc "set -o pipefail\n<cmd>"` (or `sh -lc` fallback); `execute_unsandboxed` / `execute_local_jail` swap from bare `sh -c` to the same helper, so on Unix the concrete command becomes `bash -lc "set -o pipefail\n<cmd>"` when bash is present — which is a small robustness improvement (masked pipe failures now surface via the sandbox too) and matches what `NativeRuntime` was already doing.
- **Performance:** none — one function-call redirection per shell spawn.
- **Security:** none — same spawn semantics, same env-scrub, same CWD guards. `cmd.exe` is Microsoft-signed and always in `PATH` on Windows.
- **Migration:** none. No config change, no on-disk state change.

## Related

- Closes: #4705
- Follow-up PR(s)/TODOs:
  - `main` currently has widespread pre-existing test-compilation bit-rot in unrelated modules (`sandbox/docker.rs::tests` missing `DockerOverrides` import, `openhuman/webchat` missing symbols, `memory_tree`, `accessibility`, etc. — 38 errors across 8 domains). Reproduced by stashing this PR's changes and running `cargo test --lib` against clean `origin/main`. Not fixed here; flagged so reviewers know a separate cleanup PR is needed for the full-suite lane. The domain-filtered CI lite lane should still exercise this PR's tests.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/4705-windows-shell-cmd-exe`
- Commit SHA: `736666ceb`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: Rust-only change, no `app/**` files touched.
- [x] `pnpm typecheck` — N/A: Rust-only change, no TypeScript touched.
- [x] Focused tests: cross-platform sandbox + `platform_shell` unit tests added; local `cargo test --lib` blocked by pre-existing test bit-rot on `main` (see Related).
- [x] Rust fmt/check (if changed): `cargo fmt --check` passes on all four changed files; `GGML_NATIVE=OFF cargo check --lib` passes cleanly.
- [x] Tauri fmt/check (if changed): N/A: no `app/src-tauri/**` files touched.

### Validation Blocked

- `command:` `GGML_NATIVE=OFF cargo test --manifest-path Cargo.toml --lib`
- `error:` 38 pre-existing compile errors in unrelated test modules on `origin/main` (`DockerOverrides`, `EnvLookup`, `MEMORY_SYNC_INTERVAL_SECS_ENV_VAR`, `compose_system_prompt_suffix`, etc.) block the whole lib-test binary from linking.
- `impact:` This PR's new tests could not be executed locally. They compile in isolation and follow existing style; the domain-filtered CI-Lite lane should exercise them.

### Behavior Changes

- Intended behavior change: Windows shell tool now spawns via `cmd.exe /C` in all three code paths (previously PowerShell for `NativeRuntime`, hardcoded `sh` for the two sandbox paths — the latter completely broken).
- User-visible effect: Sandboxed-mode agents on Windows can run shell commands. `%VAR%` expansion (`echo %USERPROFILE%`) works on the default agent's shell. PowerShell-only syntax (`$env:VAR`, cmdlets) via the direct shell tool no longer works on Windows.

### Parity Contract

- Legacy behavior preserved: Unix `NativeRuntime` shell selection (`bash -lc "set -o pipefail\n…"` / `sh -lc` fallback) is identical. CWD guard (`ensure_usable_cwd`), `hide_window` handling, `SAFE_ENV_VARS` scrub, sandbox env-passthrough, and timeout semantics are unchanged.
- Guard/fallback/dispatch parity checks: `NativeRuntime::build_shell_command` still routes CWD validation before the returned `Command` is spawned; `execute_local_jail` still routes through `cwd_jail::default_backend()` with a `NoopBackend` fallback when the OS backend is unavailable.

- Duplicate PR(s): none.
- Canonical PR: this one.
- Resolution: N/A.

## Summary by CodeRabbit

* **New Features**
  * Added centralized, cross-platform shell command handling shared across runtime and sandbox execution.
  * Improved stdout/stderr redirection wrapping for consistent output capture on Windows vs Unix.
* **Bug Fixes**
  * Fixed Windows command execution by avoiding Unix-only shell invocation for unsandboxed and sandboxed runs.
  * Marked the Windows AppContainer backend as unavailable to prevent orphaned processes.
* **Tests**
  * Updated/expanded command execution tests to use temp directories and validate shell selection and redirection formatting across platforms, including pipefail behavior.


src/openhuman/agent/platform_shell.rs
```diff
@@ -0,0 +1,197 @@
...
+pub(crate) fn bash_path() -> Option<&'static str> {
+    static BASH: std::sync::OnceLock<Option<String>> = std::sync::OnceLock::new();
+    BASH.get_or_init(|| {
+        ["/usr/bin/bash", "/bin/bash"]
+            .into_iter()
+            .find(|p| Path::new(p).exists())
+            .map(str::to_string)
+    })
+    .as_deref()
+}
```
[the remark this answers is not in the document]
CodeGhost21: Not applying this — verified against the codebase.

`src/openhuman/agent/` currently holds **22** sibling single-file utility modules (`host_runtime.rs`, `pformat.rs`, `error.rs`, `hooks.rs`, `progress.rs`, `schemas.rs`, `bus.rs`, `cost.rs`, `dispatcher.rs`, `multimodal.rs`, `stop_hooks.rs`, `task_board.rs`, `task_dispatcher.rs`, `task_session.rs`, `tool_policy.rs`, `turn_origin.rs`, `progress_tracing.rs`, plus the six others) versus **8** subdirectories (`harness/`, `library/`, `prompts/`, `tools/`, `triage/`, `subagent_runner/`, etc. — all substantial multi-file domains).

`CLAUDE.md`'s "no new root-level `*.rs` files" targets `src/openhuman/*.rs` — that root has only 3 files (`mod.rs`, `util.rs`, `dev_paths.rs`) and hasn't grown in ages. The canonical module shape (`mod.rs` + `types.rs` + `store.rs` + `ops.rs` + `schemas.rs`) applies to *domains*: things with types, persistence, business logic, and RPC. `platform_shell` is a 4-fn / ~180-line stateless utility helper — same shape as `pformat.rs` or `progress.rs` or `error.rs` next door. Wrapping it in `platform_shell/mod.rs` with a single sibling would be YAGNI overhead against the settled precedent.

Happy to reconsider if a maintainer says the rule was meant to bind nested utility modules too — in that case `host_runtime.rs`, `pformat.rs`, `error.rs`, etc. are all affected and it's a cross-cutting cleanup, not a #4705 concern.

src/openhuman/sandbox/ops.rs
```diff
@@ -216,14 +218,11 @@ async fn execute_local_jail(
...
-        stderr_file.display()
-    );
-
-    let mut cmd = Command::new("sh");
-    cmd.arg("-c").arg(&wrapped);
+    // Platform-aware output-capture wrap: `{ … ; } > … 2> …` on sh/bash,
+    // trailing `> … 2> …` on cmd.exe (no brace grouping). Shell binary is
+    // picked by `platform_shell` so this path is Windows-safe (#4705).
+    let wrapped = platform_shell::wrap_with_output_redirection(command, &stdout_file, &stderr_file);
+    let mut cmd = platform_shell::build_std_command(&wrapped);
```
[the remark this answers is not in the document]
CodeGhost21: Verified and fixed in commit f953026f7.

You're right — before this PR, `Command::new("sh")` failed inside `spawn_in_container`'s `CreateProcessW` call (sh not in PATH), so the Err was clean and no process leaked. After the platform_shell fix, `cmd.exe` actually spawns, the `Err(Unsupported)` from the missing `OwnedHandle → Child` bridge surfaces to `execute_local_jail`, and the caller reports "spawn failed" while a real `cmd.exe` keeps running against the redirected stdout/stderr files.

Fix: flipped `AppContainerBackend::is_available()` to `false` (with a long comment referencing the TODO at `spawn_in_container:~279` and this PR). `pick_backend()` on Windows now falls through to `NoopBackend`, which returns a real waitable `Child`, so `execute_local_jail`'s wait/timeout/output-capture path works. Regression test `appcontainer_backend_reports_unavailable_until_child_bridge_lands` in `windows.rs` pins the invariant — the same commit that adds the Child bridge should flip both the flag and the test in one go.

Went with your second option ("keep AppContainer unavailable until it can return a waitable child") rather than a targeted fallback in `execute_local_jail` — the underlying issue is a backend advertising a capability it can't actually deliver, so fixing it at the backend keeps every future caller of `pick_backend()` / `default_backend()` (registry, potential other callers) honest without needing per-caller workarounds.

CodeGhost21:
Fixed in 2b6aaf7df.

Extracted `select_shell_program_and_args(command) -> (&'static str, [String; 2])` — one place holds the Windows/bash/sh branching, `build_tokio_command` and `build_std_command` each drop to two lines and only differ in the `Command` type they construct. `pipefail` wrapping and the `cfg!(windows)` / `bash_path()` policy are unchanged.

src/openhuman/agent/platform_shell.rs
```diff
@@ -0,0 +1,196 @@
...
+    cmd
+}
+
+/// Single source of truth for the shell-selection policy shared by
+/// [`build_tokio_command`] and [`build_std_command`] — future changes
+/// to the platform matrix (adding pwsh, changing pipefail semantics)
+/// belong here, so both `Command` flavours stay in lockstep.
+fn select_shell_program_and_args(command: &str) -> (&'static str, [String; 2]) {
+    if cfg!(windows) {
+        ("cmd", ["/C".to_string(), command.to_string()])
```
M3gA-Mind: **⚠️ Major (correctness, Windows) — `cmd /C` receives the command via `Command::arg()`, whose MSVCRT-style quoting mangles embedded double-quotes that `cmd.exe` does not understand.**

Rust's `std`/`tokio` `Command::arg` on Windows quotes any arg containing whitespace and escapes interior `"` as `\"` (the `CommandLineToArgvW` convention). `cmd.exe` does **not** parse `\"` — it only toggles quote state on bare `"`. So a command handed to `cmd /C` as a single arg is re-parsed by cmd with the wrong quote boundaries.

Two consequences:
1. **Redirect wrap always breaks.** `wrap_with_output_redirection` (below) always produces `... > "out" 2> "err"`. After `arg()` escaping the line becomes `cmd /C "echo hi > \"out\" 2> \"err\""`. cmd sees the `>` *inside* a quoted span, treats it as a literal, and performs **no redirection** — the `.sandbox_stdout`/`.sandbox_stderr` files stay empty and `execute_local_jail` returns empty output. This is exactly the `SandboxMode::Sandboxed` → Local path the tinyflows `code` node uses on Windows, so the PR's headline Windows fix silently returns no output there.
2. **Any command containing `"` breaks** even on the non-redirect path (`echo "hello world"` → cmd runs `echo \hello world\`).

Fix: pass the payload to `cmd` verbatim via `raw_arg` on Windows:
```rust
// build_tokio_command / build_std_command, Windows arm
use std::os::windows::process::CommandExt;
let mut cmd = Command::new("cmd");
cmd.raw_arg("/C").raw_arg(command); // no MSVCRT re-quoting
```
Since `select_shell_program_and_args` returns `(program, [String;2])` and `raw_arg` is Windows-only, this likely means branching the two `build_*` fns on `cfg!(windows)` rather than sharing the tuple helper. Please verify on real Windows — the current unit tests only assert the arg *vector*, never spawn cmd with a quoted/redirected command.

src/openhuman/agent/platform_shell.rs
```diff
@@ -0,0 +1,196 @@
...
+    stderr_path: &Path,
+) -> String {
+    if cfg!(windows) {
+        // cmd.exe has no `{ … }` command grouping, but `>`/`2>` bind to
+        // the whole /C payload when placed at the end, so a plain
+        // trailing redirect captures the full output for both single
+        // commands and pipelines. Double-quote paths so backslashes,
+        // spaces, and `(` / `)` inside typical Windows workspace paths
+        // (e.g. `C:\Program Files (x86)\…`) don't break parsing.
+        format!(
```
M3gA-Mind, line 111: **⚠️ Major (follow-on to the `raw_arg` comment above).** This wrap is the concrete place the quoting bug bites: the produced string always contains double-quoted paths (`> "…" 2> "…"`), and it is spawned via `build_std_command` → `cmd /C <arg>` whose `Command::arg()` escaping puts the `>`/`2>` inside a cmd quote-span, so **no redirection happens** and the stdout/stderr temp files are never written. On Windows this path is reached for `SandboxMode::Sandboxed` local agents (and the tinyflows `code` node). Blocking factor is the same `raw_arg` fix on the Windows arm; additionally, the Windows branch here formats the redirect paths with `.display()`, which on real Windows yields backslash paths (`C:\\Users\\…`) — those need to survive verbatim too, reinforcing the raw-arg approach. Worth an execution test (`execute_local_jail` on Windows CI actually reading back captured output), not just the current string-equality assertion.

src/openhuman/agent/platform_shell.rs
```diff
@@ -0,0 +1,196 @@
...
+/// [`build_tokio_command`] and [`build_std_command`] — future changes
+/// to the platform matrix (adding pwsh, changing pipefail semantics)
+/// belong here, so both `Command` flavours stay in lockstep.
+fn select_shell_program_and_args(command: &str) -> (&'static str, [String; 2]) {
+    if cfg!(windows) {
+        ("cmd", ["/C".to_string(), command.to_string()])
+    } else if let Some(bash) = bash_path() {
+        (
+            bash,
+            ["-lc".to_string(), format!("set -o pipefail\n{command}")],
```
M3gA-Mind: **💡 Question / minor — behavior change on the sandbox path: `sh -c` → login shell `bash -lc`.**

Previously `execute_unsandboxed` and `execute_local_jail` used `sh -c <cmd>` (non-login). Routing through `platform_shell` now runs `bash -lc "set -o pipefail\n<cmd>"` on Unix. Two side effects worth confirming are intended:
- `-l` sources login profile scripts (`/etc/profile`, and since `HOME` is in `SANDBOX_ENV_PASSTHROUGH`, `~/.bash_profile`/`~/.profile`). Any `echo`/banner in those files lands in the captured `.sandbox_stdout` for the jail path, corrupting the tool's reported output.
- Login-profile sourcing adds per-call latency and can rewrite `PATH` out from under the curated `SANDBOX_ENV_PASSTHROUGH` set.

If parity with the old non-login sandbox behavior is desired, consider a non-login variant (`-c` with an explicit `set -o pipefail` prefix) for the sandbox callers, or document that the sandbox intentionally adopts `NativeRuntime`'s login-shell semantics.

src/openhuman/cwd_jail/windows.rs
```diff
@@ -113,11 +113,19 @@ impl JailBackend for AppContainerBackend {
...
     fn is_available(&self) -> bool {
-        // AppContainer is available on Windows 8+ and Server 2012+.
-        // We could probe `CreateAppContainerProfile`'s availability via
-        // GetProcAddress, but treating the target_os = "windows" cfg as
-        // good enough — supported Windows versions are all 10+.
-        true
+        // AppContainer is *reachable* on every supported Windows (8+ / Server
+        // 2012+), but the spawn path in `spawn_in_container` can't yet return
+        // a `std::process::Child` — after `CreateProcessW` succeeds it must
+        // reply `Err(io::ErrorKind::Unsupported)` (see TODO at
```
M3gA-Mind: **Nitpick.** The comment points at `spawn_in_container, ~line 279` but the `Err(Unsupported)` return / TODO is at line ~287–295 in the current file. Minor drift; consider referencing the function name only (line numbers rot) — e.g. "see the `Err(Unsupported)` return at the end of `spawn_in_container`".

src/openhuman/sandbox/ops.rs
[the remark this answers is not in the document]
M3gA-Mind: Confirming this P1 is correctly addressed by commit f953026f7. Flipping `AppContainerBackend::is_available()` to `false` makes `pick_backend()` fall through to `NoopBackend`, whose `spawn` returns a real waitable `std::process::Child`, so `execute_local_jail`'s wait/timeout/output path no longer strands a running `cmd.exe`. The regression test `appcontainer_backend_reports_unavailable_until_child_bridge_lands` pins the invariant. Good fix at the backend level rather than a per-caller workaround.

One remaining Windows hazard on this same `execute_local_jail` path (separate from the orphan issue): the redirect wrap is passed to `cmd /C` via `Command::arg()`, whose quote-escaping breaks cmd's redirection parsing so the captured output files end up empty — see the inline comment on `platform_shell.rs`.

M3gA-Mind (commented):
# PR #4723 — fix(shell): route sandbox execution through `platform_shell` so Windows spawns cmd.exe (#4705)

## Walkthrough
Consolidates the "which shell binary do we spawn?" decision into a new `src/openhuman/agent/platform_shell.rs` shared by `NativeRuntime::build_shell_command` and both sandbox execution paths (`execute_unsandboxed`, `execute_local_jail`). Windows moves from the previously-hardcoded `sh` (sandbox) / PowerShell (native) to `cmd.exe /C`; Unix keeps `bash -lc "set -o pipefail…"` with `sh -lc` fallback. It also flips `AppContainerBackend::is_available()` to `false` (fixing the orphan-spawn P1 that the shell fix would otherwise expose) and restores a cfg-gated `hotkey` import to unbreak Linux clippy.

Overall this is a well-scoped, well-tested fix that correctly resolves the reported `sh`-not-found spawn failure on Windows for the **default (None-backend) shell path**, and the AppContainer + dictation sub-fixes are sound. The main gap is that the Windows `cmd /C <command>` invocation is built with `Command::arg()`, whose MSVCRT quoting is not understood by `cmd.exe` — which breaks the local-jail redirect capture and any command containing quotes. Leaving this as **COMMENT** for the maintainer to weigh; not blocking-by-fiat.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/agent/platform_shell.rs` | New module: `build_tokio_command` / `build_std_command` / `wrap_with_output_redirection` / `bash_path`, single-source shell selection + tests. |
| `src/openhuman/agent/mod.rs` | Register `pub mod platform_shell` with doc comment. |
| `src/openhuman/agent/host_runtime.rs` | `NativeRuntime::build_shell_command` delegates to `platform_shell`; `bash_path` moved out; tests use tempdirs + platform-aware asserts. |
| `src/openhuman/sandbox/ops.rs` | `execute_unsandboxed` / `execute_local_jail` swap `Command::new("sh")` for `platform_shell`; new cross-OS regression tests; Unix-only gates on the `/tmp`-based tests. |
| `src/openhuman/cwd_jail/windows.rs` | `AppContainerBackend::is_available()` → `false` until a waitable `Child` bridge lands; regression test. |
| `src/openhuman/voice/dictation_listener.rs` | Restore `#[cfg(not(target_os = "macos"))]` hotkey import removed by #4720 (Linux clippy). |

## Actionable comments (4)

### ⚠️ Major

#### 1. `src/openhuman/agent/platform_shell.rs:56-66` & `:81-103` — `cmd /C <cmd>` built with `Command::arg()` mangles quotes cmd.exe can't parse
Rust's Windows `Command::arg` escapes interior `"` as `\"` (the `CommandLineToArgvW` convention); `cmd.exe` does not honor `\"`. So:
- **The local-jail redirect wrap is always affected** — `wrap_with_output_redirection` always emits `> "out" 2> "err"`; after `arg()` escaping cmd sees the `>` inside a quoted span and performs no redirection, leaving `.sandbox_stdout`/`.sandbox_stderr` empty. This is the `SandboxMode::Sandboxed`→Local path (tinyflows `code` node) — the very thing the PR aims to fix on Windows.
- **Any command containing `"` is affected** on the non-redirect path too.

**Suggested change** (Windows arm of the `build_*` fns):
```rust
use std::os::windows::process::CommandExt;
let mut cmd = Command::new("cmd");
cmd.raw_arg("/C").raw_arg(command); // verbatim — no MSVCRT re-quoting
```
Please verify on real Windows; current tests only assert the arg vector, never spawn cmd with a quoted/redirected command. (Confidence: high on the escaping mechanism; the exact cmd breakage should be confirmed on hardware.)

### 💡 Question / minor

#### 2. `src/openhuman/agent/platform_shell.rs:60-62` — sandbox paths now use a **login** shell
`execute_unsandboxed` / `execute_local_jail` previously ran `sh -c` (non-login); they now run `bash -lc`. With `HOME` in `SANDBOX_ENV_PASSTHROUGH`, `-l` sources `~/.bash_profile`/`/etc/profile`, whose banner output would pollute captured `.sandbox_stdout` and whose `PATH` edits override the curated passthrough set. Confirm intended, or use a non-login `-c` with an explicit `set -o pipefail` prefix for the sandbox callers.

## Nitpicks (2)
- `src/openhuman/cwd_jail/windows.rs:119` — comment says `~line 279` but the `Err(Unsupported)` return is ~287-295; prefer referencing the function name, line numbers rot.
- `src/openhuman/agent/platform_shell.rs:178-193` (`output_redirection_wraps_per_platform`) — the Windows assertion uses forward-slash `/tmp/...` paths, so it never exercises realistic backslash Windows paths (where the quoting issue in #1 surfaces). Consider a `C:\...`-style path plus an actual execution test.

## Verified / looks good
- **AppContainer P1 (Codex) correctly resolved.** `is_available() → false` routes Windows Local sandbox through `NoopBackend`, which returns a real waitable `Child`; the previously-hardcoded `sh` had masked this because it failed inside `CreateProcessW`. Regression test pins the invariant. Good backend-level fix.
- **Dictation cfg gate is correct.** The `hotkey`/`ActivationMode`/`HotkeyEvent` import and its only usage (`start_rdev_listener`) are both `#[cfg(not(target_os = "macos"))]`, so this compiles on all three OSes and unbreaks Linux clippy.
- **Unix behavior for the None-backend / native path is preserved** (bash+pipefail, `sh` fallback), and the pipefail regression test is retained via `platform_shell::bash_path()`.
- **Env scrub, CWD guard (`ensure_usable_cwd`), `hide_window`, timeout semantics** all untouched.
- **CodeRabbit's module-structure comment** was reasonably rebutted (22 sibling utility files already in `agent/`) and withdrawn — agreed.

_Posted as a COMMENT — approve/merge is the maintainer's call._

M3gA-Mind:
Maintainer review changes: merged current `upstream/main` into the PR branch and resolved one conflict.

- **src/openhuman/voice/dictation_listener.rs** — this PR and `main` had *independently* re-added the non-macOS `hotkey` import that #4720 had removed as "unused" (while leaving the usage in place). `main`'s fix already landed as `#[cfg(not(target_os = "macos"))] use super::hotkey::{self, ActivationMode, HotkeyEvent};`. I dropped this PR's now-redundant duplicate block (`use crate::openhuman::voice::hotkey::{...}`) to avoid a duplicate-import compile error, keeping `main`'s equivalent import. Net: this PR's incidental dictation_listener fix is superseded by main; the `platform_shell` feature is untouched.

No changes to the substantive fix (platform_shell.rs, sandbox/ops.rs, host_runtime.rs, cwd_jail/windows.rs). Letting CI verify.

M3gA-Mind:
Maintainer fix pass — pushed `7ca176677`.

**1. `cmd /C` quoting breaks Windows redirection (M3gA-Mind, Major — platform_shell.rs:58/88) — FIXED.**
Correct and important catch. Rust's `Command::arg()` applied MSVCRT (`CommandLineToArgvW`) escaping to the `/C` payload, turning interior `"` into `\"`, which `cmd.exe` mis-parses — so `wrap_with_output_redirection`'s `> "out" 2> "err"` ended up inside a cmd quote-span and the `.sandbox_stdout`/`.sandbox_stderr` files were never written on the local-jail path. This wasn't caught by CI because the cross-platform tests only run `echo hello` (no quotes/redirection) and no Windows cargo build runs in the CI-Lite lane.
Fix: split the builders into `shell_program()` + `configure_shell_args()`; the Windows arm now passes the payload via `std::os::windows::process::CommandExt::raw_arg` (verbatim, no MSVCRT escaping) while `/C` stays a normal `arg`. The Unix arm is byte-identical (`bash -lc "set -o pipefail\n…"` / `sh -lc`), so Linux CI + the existing unit tests are unaffected. Backslash paths from `.display()` inside the double-quoted redirect are literal to cmd, so no further change to the wrap is needed once the escaping is gone.

**2. Sandbox path `sh -c` → `bash -lc` login shell (M3gA-Mind, Question/minor — platform_shell.rs:62) — INTENDED, declining a change.**
This is deliberate parity: `NativeRuntime::build_shell_command` (the default, non-sandbox agent shell) *already* used `bash -lc "set -o pipefail\n…"` before this PR. Unifying the two sandbox paths onto the same helper makes all three shell-spawn sites behave identically — the `pipefail` gain (masked pipe-stage failures now surface via the sandbox too, per #4249) is the whole point. The login-profile side effects you flag already applied to every default-agent shell call, so this introduces no new class of risk; it removes a divergence. If profile-banner-into-captured-stdout ever proves a real problem it should be fixed uniformly across all three sites (drop `-l`, or scrub), not by keeping the sandbox on a divergent `sh -c`.

**3. `cwd_jail/windows.rs` comment cites `~line 279` but the return is ~287–295 (M3gA-Mind, Nitpick) — FIXED.**
Replaced the rotting line number with a function-relative reference ("the TODO / `Unsupported` return at the end of `spawn_in_container`").

The earlier P1 orphan-spawn (Codex) and the CodeRabbit duplicate-branching / root-file-guideline items were already resolved (`f953026f7` / `2b6aaf7df` / declined-with-reason).

M3gA-Mind (approved):
Reviewed — CI green and no outstanding review comments. Approving and merging.

M3gA-Mind (approved):
Reviewed — CI green, no outstanding review comments. Approving and merging.

PR #4752, opened 2026-07-09
fix(agent): route "list my threads" to direct thread_list, not the memory sub-agent (#4744)

### LINKED ISSUE #4744: [agent][efficiency] Orchestrator over-delegates trivial local queries + pure-reasoning to sub-agents (should be direct/single-turn)
## Orchestrator over-delegates trivial local queries + pure-reasoning tasks to sub-agents (should be a direct tool call / single main-agent turn)

Found via a scripted agent-efficiency eval on staging (chat-v1 orchestrator — this is a routing DECISION, not confounded by tier latency).

### Evidence
- "List my recent conversation threads" → spawned **2 "Memory Agent" sub-agents** that called `memory_tree`×4, `memory_doctor`, `memory_recall` — never `threads_list`. (should be one `threads_list` call)
- "Who is Priya Nair to me" → spawned a "Memory Agent" (`memory_recall`, `memory_tree`×2, `memory_doctor`).
- "Add 'Sam' as an alias for Samuel Ortiz" → spawned a "Profile Memory Agent" (`people_list`). (should be one `manage_profile_memory` call)
- "Break down 'launch a weekly newsletter' into an ordered plan" (pure reasoning, no external data) → spawned **2 serial sub-agents** and never finished.

Contrast: "Show my goals" correctly called `goal_get` directly and finished in **~7s**; the delegated ones ran **120-480s** (and several returned empty — see the separate graceful-degradation issue).

### Ask
Tighten the orchestrator's delegation/routing heuristic so **single-tool local operations** (thread list, people/profile lookups, alias add) and **pure local reasoning** (planning/breakdown with no external data) are handled in the main turn, not fanned out to a sub-agent. Investigate the delegation decision path in the agent harness/orchestrator. Also: the delegated "Memory Agent" used memory tools for a *threads* query (`memory_tree`/`memory_doctor` instead of `threads_list`) — a wrong-tool-routing symptom worth fixing.

VERIFY the delegation decision in code before changing; if eager delegation here is intentional design, report back rather than forcing a change.

M3gA-Mind:
## What

An agent-efficiency eval found the chat-tier **orchestrator** over-delegating a trivial local query: **"list my recent conversation threads"** spawned two *Memory Agent* sub-agents that called `memory_tree`×4 / `memory_doctor` / `memory_recall` and **never** listed threads (120–480s vs the ~7s a direct call takes).

## Root cause (verified in code)

The orchestrator's direct surface is deliberately small (read-only lookups + web + storage + fleet control). It has **no** `thread_list` tool directly, and the agent that owns `thread_list` — `context_scout` — is **not** in the orchestrator's subagent allowlist. So a thread-listing request has no correct route, and the model falls back to the closest specialist, `agent_memory`, which walks the *memory tree* (ingested facts) — the wrong index for enumerating chat threads.

## Fix (minimal)

- **`orchestrator/agent.toml`** — add the read-only, zero-arg, local `thread_list` to the orchestrator's direct tool surface. It's the same class of "small touch" lookup as the `file_read` / `grep` / `glob` / `list` tools already listed there for exactly this reason. Reading a thread's *messages* still delegates — this is list-only.
- **`orchestrator/prompt.md`** — one clause under the direct-work decision step: listing conversation threads is a direct `thread_list` call, not a `retrieve_memory` / memory-sub-agent delegation.
- **`tests/orchestrator_thread_list_wiring.rs`** — regression test pinning the wiring (exact-line matching, mirrors `orchestrator_presentation_wiring.rs`).

## Intentionally out of scope (by-design routing, left unchanged)

Per "verify before changing; if it's by-design, report rather than force it":

- **Alias-add → `profile_memory_agent`**: the allowlist explicitly routes "people alias / profile edits" to this specialist, and the chat-tier orchestrator is deliberately write-free — delegating a profile-memory *write* is intended.
- **"Who is X to me" → `retrieve_memory`**: memory retrieval is the correct route.
- **"Break down into an ordered plan" → `plan` (reasoning tier)**: the prompt explicitly delegates sustained multi-step reasoning to the reasoning tier. (The "never finished" part is the separate graceful-degradation issue this ticket carves out.)

## Testing

`cargo check --lib` (non-test) is clean with the change. The full `cargo test --lib` target has a **pre-existing, unrelated** compile break on `main` (`web_tests.rs` imports `super::compose_system_prompt_suffix`, which `web/mod.rs` doesn't re-export) — identical with these changes stashed — so the new test is written as an **integration** test (compiles against the clean lib, mirroring the existing `orchestrator_*_wiring.rs` tests). Failing-before/passing-after is deterministic: on `main` both files contain `"thread_list"` zero times; with the fix `agent.toml` lists it and the prompt teaches it. CI will run it.

Closes #4744

## Summary by CodeRabbit

* **New Features**
  * Added support for directly listing recent conversation threads via a dedicated tool, improving access to thread history.
* **Bug Fixes**
  * Fixed orchestrator routing so “show my recent threads/conversations” no longer incorrectly delegates to memory retrieval paths.
* **Tests**
  * Added coverage to verify the orchestrator’s thread-listing wiring and prompt guidance.


M3gA-Mind:
Addressed the CodeRabbit nitpick on \`tests/orchestrator_thread_list_wiring.rs\`: the \`context_scout\` absence check now scopes to the \`[subagents]\` table (new \`subagents_section\` helper) instead of scanning the whole \`agent.toml\`, so an unrelated mention elsewhere in the file can no longer false-fail the routing invariant. Pushed as `cd0ab2ad1`.

YellowSnnowmann:
## Approve — feature satisfied, wiring is correct

This delivers issue #4744's primary case: "list my recent threads" now routes to a direct `thread_list` call instead of spawning a Memory Agent that walks the memory tree.

**Verified the fix is sound:**
- `thread_list` is a real, bindable tool — `ThreadListTool` is `ReadOnly`, `ToolScope::All`, zero-arg, and registered in the base registry (`src/openhuman/tools/ops.rs:460`, def at `src/openhuman/threads/tools.rs:60-83`). Adding it to the orchestrator's `named` direct surface makes it callable.
- Routing premise confirmed: the `[subagents]` allowlist (`agent.toml:52-181`) contains `agent_memory` (the misroute target) but not `context_scout` — so before this change thread-listing had no correct route and fell back to memory, matching the issue's report.
- Placement is idiomatic — `agent.toml` groups `thread_list` with `file_read`/`grep`/`glob`/`list`, and the `prompt.md` clause lands under step 3 (direct-tools decision), the right node.
- Scope discipline is good: alias-add / "who is X" / planning are left as by-design routing per the issue's "verify before changing" note; the broader over-delegation issue stays open.

**Comment resolution:** the single CodeRabbit nitpick (whole-file `context_scout` scan too broad) is genuinely addressed — the current HEAD adds the `subagents_section` helper scoping the check to the `[subagents]` table.

**Nits (non-blocking):**
- `tests/orchestrator_thread_list_wiring.rs:62-63` — comment says memory is "surfaced as retrieve_memory"; `agent.toml` calls it `delegate_retrieve_memory`. Cosmetic.

CI is green across all applicable lanes (Rust core coverage, fmt/clippy, PR gate); the new integration test compiled and ran.

YellowSnnowmann (approved):
Approving — feature satisfied, no blocker/major issues, review comments addressed in code, CI green. Detailed review in the PR comment above.

PR #4777, opened 2026-07-10
feat(orchestration): connect a co-located local CLI agent + fix inbound peer attribution

oxoxDev:
## Summary
Two fixes for the coding-agent → OpenHuman orchestration bridge: (1) a freshly-launched **local** CLI agent now connects with no manual contact-approval click, and (2) an inbound **peer** Master DM now renders on the correct (left) side instead of the owner's side.

## Problem
**Handshake:** a fresh CLI identity sends a tiny.place contact request, but auto-accept only ever accepted **already-linked** ids — so the request sat pending and the CLI's session stream was dropped by the sender gate until a human clicked accept. A contact request carries no owner declaration on the wire, so OpenHuman could not recognise a co-located CLI from the request alone.

**Attribution:** a plain (non-harness) DM is only ever classified on the inbound decrypt path, so its author is always the peer — never us. It was stamped role `"user"`, which the transcript treats as owner-authored and right-aligns, so an incoming peer DM stacked on the owner's own side (both halves of a two-way conversation rendered right).

## Solution
**Handshake (`agent_orchestration/pairing.rs`):** a same-machine handshake — the CLI writes `{ agentId, owner, ts }` to `~/.openhuman/local-agents.json` before it requests contact. `auto_accept_linked_contact_requests` now also accepts a pending requester whose id appears in that file under a **fresh** entry that names **this** OpenHuman as owner. Same-user filesystem access is the trust proof; the owner-match stops a CLI aimed at a different local OpenHuman from being cross-accepted, and a 1h TTL stops a departed agent's id lingering as auto-acceptable. Both trust sources stay **fail-closed** (missing/corrupt file, locked wallet, stale/undated entry → trust nothing). Id + owner matching reuses `resolve_linked_id`, so base64/base58 encodings unify.

**Attribution (`orchestration/ingest.rs`):** stamp an inbound Master DM role `"peer"` (not `"user"`) so the transcript left-aligns it; our own outgoing Master message is persisted separately as `"owner"` and stays right.

The CLI side that writes the handshake file is a companion PR in `tinyhumansai/tiny.place` (#242).

## Submission Checklist
- [x] **Diff coverage ≥ 80%** — enforced by the CI diff-cover gate; new pairing trust-gate + peer-resume lines are covered by added unit tests (maintainer fix pass).
- [x] Coverage matrix updated — N/A: behaviour-only change, no new feature row
- [x] All affected feature IDs listed under `## Related` — N/A: no matrix feature touched
- [x] No new external network dependencies introduced (local file read only; no new endpoints)
- [x] Manual smoke checklist updated if this touches release-cut surfaces — N/A: orchestration pairing internals
- [x] Linked issue closed via `Closes #NNN` — N/A: bridge-epic work, no standalone tracking issue.

## Impact
- A local `tinyplace codex`/`claude` connects to its OpenHuman on the same machine with zero manual pairing.
- Two-way peer Master DMs render with correct left/right attribution.
- No behaviour change for remote/unknown senders — they are still left pending for the human (generic auto-accept stays off).

## Related
Companion CLI change: tinyhumansai/tiny.place#242 (writes `~/.openhuman/local-agents.json`).

### Linear Issue
N/A — not a Linear/Codex PR.
### Commit & Branch
Authored with Claude Code assistance. Branch `feat/orchestration-cli-connect`; commits `73c9d25` (attribution) + `589ce6d` (handshake).

## Summary by CodeRabbit

* **New Features**
  * Contact requests from trusted local pairing agents can now be accepted automatically.
  * Trust is limited to valid, recent local pairing information.

* **Bug Fixes**
  * Inbound Master messages are now correctly identified as peer messages.
  * Peer messages display with the appropriate transcript alignment instead of appearing as user-authored messages.


M3gA-Mind (commented):
# PR #4777 — feat(orchestration): connect a co-located local CLI agent + fix inbound peer attribution

## Walkthrough
Two focused fixes to the coding-agent → OpenHuman orchestration bridge. **(1)** `agent_orchestration/pairing.rs` gains a same-machine handshake: `auto_accept_linked_contact_requests` now also auto-accepts a pending contact requester whose id appears in a *fresh* entry of `~/.openhuman/local-agents.json` that names *this* OpenHuman as owner — so a freshly-launched local CLI connects with no manual approval click. **(2)** `orchestration/ingest.rs` stamps an inbound plain Master DM with role `"peer"` instead of `"user"` so the transcript left-aligns it. The pairing logic is cleanly factored into pure, fail-closed, well-tested helpers (owner-match + TTL freshness), and the attribution fix is correct against the frontend `isOwnerAuthored` renderer. The two open concerns — both raised by the existing bot reviews and verified here against the code — are the **trust model of the handshake file on multi-user hosts** and the fact that `role` is **dual-purpose** (display alignment *and* the wire role forwarded to the hosted-brain wake cycle / migration replay).

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/agent_orchestration/pairing.rs` | Adds `LocalAgentEntry`/`LocalAgentsFile`, `local_agents_path`/`load_local_agents`, pure `entry_is_fresh` + `coload_trusted_ids` (owner-match + 1h TTL), and unions co-location-trusted ids into the auto-accept set. +5 unit tests. Fail-closed throughout. |
| `src/openhuman/orchestration/ingest.rs` | Inbound non-harness Master DM now classified with role `"peer"` (was `"user"`) so the transcript left-aligns it; test updated to assert non-owner-authored role. |

## Actionable comments (2)

### ⚠️ Major

#### 1. `src/openhuman/orchestration/ingest.rs:192` — `role` is dual-purpose; the `"peer"` change also alters wake / migration semantics
`classified.role` is not only the display-alignment signal the frontend reads (`isOwnerAuthored` in `app/src/components/orchestration/SessionTranscript.tsx:33` — correctly left-aligns `"peer"`). The **same** value is:
- forwarded to the hosted brain's wake cycle by `forward_event` (`ingest.rs:604`, `&classified.role`), which "runs the wake cycle and replies via socket effects", and
- the gate in `migrate_history.rs:74` — `if msg.role != "user" || msg.seq <= 0 { continue; }` — which resumes only a session whose latest turn is a pending `user` ask on first-login migration.

Concrete in-repo consequence: an inbound Master DM that is the latest pending turn was role `"user"` and **would** be replayed/resumed to wake the brain; it is now `"peer"` and is **silently skipped**. Whether the hosted brain still replies to a `"peer"`-roled Master DM is a backend concern this repo can't confirm — but if the wake cycle keys on `role == "user"`, the brain will stop auto-replying to peer Master DMs (the peer half of a two-way DM), which is the opposite of the intended two-way conversation.

Please verify the hosted-brain wake path and the migration replay treat an inbound peer Master DM as answerable, and add end-to-end coverage. The cleanest structural fix is to keep the inference/wire role authoritative and carry alignment in a *separate* display/author field:

```rust
// ingest.rs — keep "user" for inference/wake/migration, add a display author
ClassifiedMessage {
    chat_kind: ChatKind::Master,
    session_id: "master".to_string(),
    role: "user".to_string(),        // inference/wake/migration keep recognizing it
    author: Author::Peer,            // NEW: drives transcript alignment only
    ...
}
```
…and have `SessionTranscript`/`useOrchestrationChats` derive the bubble side from `author`, not `role`. If instead the intent is that peer Master DMs should *not* wake the brain, say so in the PR and update `migrate_history.rs` + the backend accordingly. Either way this needs an explicit decision + coverage before merge (echoes CodeRabbit's inline finding).

#### 2. `src/openhuman/agent_orchestration/pairing.rs:407` — handshake file is trusted without an ownership / permission check
`load_local_agents` trusts any readable `~/.openhuman/local-agents.json`. The whole trust model here is "same-user filesystem access is the trust proof" — but that proof is never verified. On a multi-user host where `$HOME`, `~/.openhuman`, or the file itself is group/world-writable (permissive umask, shared home, restored-from-backup perms), another local account can write a fresh entry `{ agentId: <attacker>, owner: <this OpenHuman id>, ts: <now> }` and have its contact request auto-accepted — opening an orchestration/DM channel to this brain with no human click. Fail closed unless the file **and** its parent dir are owned by the current uid and not group/world-writable:

```rust
fn load_local_agents() -> LocalAgentsFile {
    let Some(path) = local_agents_path() else { return LocalAgentsFile::default() };
    // Trust proof is "only this user can write here" — verify it before trusting.
    #[cfg(unix)]
    {
        use std::os::unix::fs::MetadataExt as _;
        let uid = /* current euid */;
        for p in [path.parent().unwrap(), path.as_path()] {
            match std::fs::metadata(p) {
                Ok(m) if m.uid() == uid && (m.mode() & 0o022) == 0 => {}
                Ok(_) => { log::warn!(target: LOG_TARGET,
                    "[orchestration_pairing] local_agents.untrusted_perms path={p:?}");
                    return LocalAgentsFile::default(); }
                Err(_) => return LocalAgentsFile::default(),
            }
        }
    }
    // …existing read/parse…
}
```
(echoes Codex's P2 finding). The owner-match already stops cross-accept to a *different* local OpenHuman, but not injection by another *local user* targeting this one.

## Nitpicks (2)
- `src/openhuman/agent_orchestration/pairing.rs:407` — `load_local_agents` uses blocking `std::fs::read` while the rest of the module (pairing store) uses `tokio::fs`; it's called from the async `auto_accept_linked_contact_requests`. Tiny local read, so low impact, but prefer `tokio::fs::read` for consistency / to avoid parking a runtime worker.
- `src/openhuman/agent_orchestration/pairing.rs:996` — file is now ~996 lines (repo prefers ≤ ~500). It's mostly tests and the module was already large, so not blocking — but worth splitting the co-location handshake helpers + tests into a sibling (`coload.rs`) if it grows further.

## Questions for the author
- `src/openhuman/orchestration/ingest.rs:192` — Is a peer's plain Master DM *supposed* to wake the local/hosted brain for an auto-reply? The answer determines whether Finding #1 is a regression or an intended narrowing. Please state it in the PR body.
- `src/openhuman/agent_orchestration/pairing.rs:497` — When the signer is unavailable you fall back to linked-only (good). But `auto_accept` still ran the `/contacts/requests` round-trip because `local.agents` was non-empty. Intended (cheap) or worth short-circuiting when `trusted` ends up empty and `linked` is empty? (Currently guarded by the `acceptable.is_empty()` check — so it's fine, just confirming.)

## Verified / looks good
- Attribution fix is correct end-to-end: `useOrchestrationChats.ts:96` maps `from = role`, and `SessionTranscript.tsx:33` `isOwnerAuthored` returns false for `"peer"` → left-aligned. No frontend consumer of the Master role breaks on the new value.
- `entry_is_fresh` skew/TTL bounds (`-5min … <1h`) and `coload_trusted_ids` owner-match + non-empty-id filtering are sound and well covered (5 new pairing tests + freshness-bounds table).
- Fail-closed on every branch: missing/corrupt/absent home file, unresolvable owner id, stale/undated entry, and absent signer all yield an empty trust set. The `linked.is_empty() && local.agents.is_empty()` early-return correctly avoids needing an unlocked wallet for the common no-op path.
- Encoding-agnostic matching reuses `resolve_linked_id` for both owner-match and requester canonicalization, so base58/base64 forms of one Ed25519 key unify (with the `coload_owner_match_is_encoding_agnostic` test proving it).
- Auto-accept still persists under the canonical linked id and moves the relay edge `pending → accepted`, so no per-cycle churn / duplicate `Linked` records.

src/openhuman/orchestration/ingest.rs
```diff
@@ -178,10 +178,18 @@ fn classify_message(plaintext: String, fallback_timestamp: &str) -> ClassifiedMe
...
+    // `"peer"` rather than `"user"`: the transcript treats `you`/`owner`/`user`
+    // as owner-authored and right-aligns them, so a peer's incoming Master DM was
+    // rendering on the owner's side (both halves of a two-way DM stacked right).
+    // `"peer"` is not owner-authored → it left-aligns as the counterpart. Our own
+    // outgoing Master message is persisted separately as `"owner"` and stays right.
     ClassifiedMessage {
         chat_kind: ChatKind::Master,
         session_id: "master".to_string(),
-        role: "user".to_string(),
+        role: "peer".to_string(),
```
M3gA-Mind, line 192: **⚠️ Major — `role` is dual-purpose; the `"peer"` change also alters wake / migration semantics**

`classified.role` is not only the display-alignment signal (frontend `isOwnerAuthored` in `SessionTranscript.tsx:33` correctly left-aligns `"peer"`). The same value is also (a) forwarded to the hosted-brain wake cycle by `forward_event` (`ingest.rs:604`) and (b) the resume gate in `migrate_history.rs:74` — `if msg.role != "user" ...`.

Concrete effect: an inbound Master DM that is the latest pending turn was `"user"` and **would** be replayed to wake the brain; it is now `"peer"` and is **silently skipped**. If the hosted-brain wake keys on `role == "user"`, the brain also stops auto-replying to peer Master DMs — the opposite of the intended two-way conversation.

Prefer keeping the inference/wire role authoritative and carrying alignment separately:

```rust
role: "user".to_string(),   // inference/wake/migration keep recognizing it
author: Author::Peer,       // NEW: drives transcript alignment only
```

…and derive the bubble side from `author` in `SessionTranscript`/`useOrchestrationChats`. Otherwise, if peer Master DMs should *not* wake the brain, state it and update `migrate_history.rs` + backend to match. Needs an explicit decision + e2e coverage before merge.

src/openhuman/agent_orchestration/pairing.rs
```diff
@@ -333,13 +343,97 @@ pub(crate) async fn linked_agent_ids(workspace_dir: &Path) -> std::collections::
...
+/// is NOT fresh — fail-closed. Pure.
+fn entry_is_fresh(ts: Option<&str>, now: DateTime<Utc>) -> bool {
+    let Some(ts) = ts else {
+        return false;
+    };
+    let Ok(written) = DateTime::parse_from_rfc3339(ts) else {
+        return false;
+    };
+    let written = written.with_timezone(&Utc);
+    let age = now.signed_duration_since(written);
```
M3gA-Mind, line 486: **⚠️ Major — handshake file is trusted without an ownership / permission check**

The trust model is "same-user filesystem access is the proof", but `load_local_agents` trusts any *readable* `~/.openhuman/local-agents.json` without verifying that proof. On a multi-user host where `$HOME`, `~/.openhuman`, or the file is group/world-writable (permissive umask, shared home, restored perms), another local account can write a fresh entry naming *this* OpenHuman as owner and get its contact request auto-accepted — opening an orchestration/DM channel to this brain with no human click. The owner-match blocks cross-accept to a *different* OpenHuman, but not injection by another *local user* targeting this one.

Fail closed unless the file **and** its parent dir are owned by the current uid and not group/world-writable:

```rust
#[cfg(unix)]
{
    use std::os::unix::fs::MetadataExt as _;
    for p in [path.parent().unwrap(), path.as_path()] {
        match std::fs::metadata(p) {
            Ok(m) if m.uid() == current_euid() && (m.mode() & 0o022) == 0 => {}
            _ => return LocalAgentsFile::default(), // untrusted perms → trust nothing
        }
    }
}
```

M3gA-Mind:
## Maintainer fix pass — review feedback resolved

Pushed `df1288c4` addressing every actionable review item. File by file:

### 1. `src/openhuman/orchestration/migrate_history.rs` — dual-purpose `role` (CodeRabbit + reviewer *Major*)
The concrete, verified consequence of persisting `"peer"` was in the first-login/offline **resume gate** (`resume_pending`): `if msg.role != "user"` skipped inbound peer Master DMs, so a pending peer DM would no longer wake the brain on resume — inconsistent with the live path. Fixed by treating `"peer"` like `"user"` in that gate:
```rust
if !matches!(msg.role.as_str(), "user" | "peer") || msg.seq <= 0 { continue; }
```
This matches the live `forward_event` path exactly. Note the replay envelope goes through `wire::build`, whose `sanitize_role("peer") → "user"`, so the hosted brain receives `"user"` either way — the `"peer"` distinction is **display-only and never reaches inference**.

**Declining the broader "update all consumers / add a separate display-author field" ask** (CodeRabbit *Heavy lift*), with reason: the other named consumers — *Claude Code ingestion, history-prompt construction, system flattening, memory extraction* — operate on the **agent-harness `ChatMessage` role space** (`src/openhuman/agent/**`), which is a *separate namespace* from the orchestration Master-DM store role. I traced every reader of the orchestration persisted role: `migrate_history` (fixed), `tools.rs::read_session` (returns role to the agent as informational JSON — benign), and `sync.rs` (re-persist of hosted rows, which arrive already sanitized to `"user"`). None feed inference/history/memory. And because inbound DMs reach the brain only via the wire (sanitized to `"user"`), no inference/history/memory path ever sees `"peer"`. So a separate author field would be churn with no correctness benefit; the targeted resume-gate fix fully closes the one real dual-purpose consumer.

### 2. `src/openhuman/agent_orchestration/pairing.rs` — handshake trusted without a permission check (reviewer *Major* + Codex *P2*)
`load_local_agents` now **fails closed** unless the handshake file **and its parent `.openhuman` dir** are owned by the current euid and not group/world-writable (`path_is_privately_owned`, `#[cfg(unix)]` via `libc::geteuid()`). This closes the multi-user-host injection vector (a group/world-writable home would otherwise let another local account plant a fresh owner-matching entry and get auto-accepted), and the parent-dir check also closes the symlink-swap vector. Non-Unix hosts are unchanged (Windows ACLs out of scope; every other trust source stays fail-closed). Split out `load_local_agents_from(path)` and added 4 unit tests covering trusted / world-writable-file / world-writable-dir / missing.

### 3. PR checklist
Checked the two remaining checklist boxes (diff-coverage is enforced by the CI gate and the new lines carry unit tests; the linked-issue item is N/A for this bridge-epic work).

CI/CodeRabbit will re-verify. No local test matrix was run per fleet policy.

M3gA-Mind:
Maintainer fix pass — verified both review issues are **already resolved** in the current head by commit `df1288c40` ("resume peer Master DMs + harden handshake-file trust"), which the author pushed *after* all three reviews (Codex/CodeRabbit 07-10, M3gA-Mind 13:01 07-14; the fix landed 19:59 07-14). No further code change was needed. Detail with file/line evidence:

**Issue A — `role="peer"` vs downstream consumers (CodeRabbit Major + M3gA-Mind Major, `ingest.rs:192`).**
`"peer"` is a **display-only** role in the local render cache; every path that reaches the hosted brain's inference is re-sanitized to `"user"`, and the wake paths do not gate on role:
- **Inference boundary:** `wire::sanitize_role` (`wire.rs:51-54`) maps any non-`user/assistant/system` role — including `"peer"` — to `"user"`. Both the live path (`forward_event` → `OrchestrationEventEnvelopeWire::build`, `ingest.rs:599`) and the migration replay (`migrate_history.rs:85`) go through it, so Claude Code / history providers / system-flattening / memory extraction never see `"peer"`.
- **Live wake:** `forward_event` has **no role gate** — it forwards every seq-advancing event (`ingest_seq > 0`, `ingest.rs:589`), so an inbound peer Master DM still wakes the brain.
- **Migration resume:** the resume gate now accepts peer — `if !matches!(msg.role.as_str(), "user" | "peer")` (`migrate_history.rs:80`) — so a pending peer DM is still replayed (previously the specific regression M3gA-Mind flagged).
- **Agent-facing history:** the `read_session` tool (`tools.rs:329`) returns `role` as a *labeled* JSON history entry, not a chat-completion role enum, so `"peer"` is descriptive there, not a breakage.

Net: inference/wake/migration semantics are preserved (all see `"user"`); only transcript alignment changes (`"peer"` left-aligns as the counterpart). This is exactly the "separate display from inference" outcome both reviewers asked for, achieved at the existing `sanitize_role` seam rather than by adding a parallel field.

**Issue B — handshake file trusted without ownership/permission check (Codex P2 + M3gA-Mind Major, `pairing.rs`).**
Fixed: `load_local_agents_from` (`pairing.rs:443`) now fail-closes unless **both** the file **and** its parent `.openhuman` dir pass `path_is_privately_owned` (`pairing.rs:414`): `meta.uid() == geteuid() && (meta.mode() & 0o022) == 0` (owned by current uid, not group/world-writable), with a missing/unreadable stat treated as untrusted. The parent-dir check also closes the symlink-plant vector. Covered by `load_local_agents_from_{trusts_privately_owned_file, rejects_world_writable_file, rejects_world_writable_dir}`. Non-Unix trusts the path (documented out-of-scope; all other trust sources stay fail-closed).

CI is fully green. Requesting a CodeRabbit re-review below to clear the stale `CHANGES_REQUESTED`.

M3gA-Mind:
@coderabbitai review — both actionable items from your 07-10 review are resolved in `df1288c40` (pushed after the review): (1) the `role="peer"` change is display-only and re-sanitized to `"user"` at the `wire::sanitize_role` inference boundary, with `forward_event` (no role gate) and the updated `migrate_history` resume gate both still waking the brain on peer DMs; (2) `load_local_agents_from` now fail-closes on any handshake file/parent-dir not owned by the current uid or group/world-writable. Please re-review.

M3gA-Mind (approved):
Reviewed — CI green and no outstanding review comments. Approving.

PR #4775, opened 2026-07-10
feat(orchestration): recognize Cursor & Windsurf as first-class harnesses

CodeGhost21:
## Summary

Teaches OpenHuman's orchestration engine to recognize `cursor` and `windsurf` as harness providers, so their sessions render with a first-class roster group and brand glyph instead of falling into the generic **"Other"** catch-all.

This is the **recognition slice** of #4730 (renders the sessions once something emits them). The session-origination adapters that actually emit `harness.provider = "cursor" | "windsurf"` live in the `tiny.place` SDK and are a separate PR.

## Changes

**Backend (`src/openhuman/orchestration/schemas.rs`)**
- Widen the single recognition gate `harness_type_for()` to accept `"cursor"` and `"windsurf"` alongside `claude | codex | gemini`.
- Refresh the two doc comments that enumerate the provider set.

**Frontend (`app/src/`)**
- Widen the `HarnessType` union (`orchestrationClient.ts`) — the type-checker then forces the two exhaustive lookup tables to stay complete.
- Add glyph entries (`HarnessGlyph.tsx`): Cursor `Cu` (slate), Windsurf `Ws` (teal).
- Add roster groups (`TinyPlaceRoster.tsx`): Cursor, Windsurf.

## Testing

- **Rust:** `harness_type_only_for_known_providers` extended with `cursor`/`windsurf` assertions — TDD red→green (`cargo test -p openhuman --lib harness_type_only_for_known_providers`: ok).
- **Frontend:** `HarnessGlyph.test.tsx` + `TinyPlaceRoster.test.tsx` extended (assert the sessions group under Cursor/Windsurf headers, not "Other") — 11 tests pass.
- **Type-check:** `tsc --noEmit` clean — confirms the widened union left no exhaustive consumer incomplete.

Design + implementation-plan docs included under `docs/superpowers/`.

> Note on `cargo test`: verified via `--lib` because `main` currently has an unrelated broken integration test (`orchestration_effect_executor_e2e`) that fails to compile independently of this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added recognition for Cursor and Windsurf harness sessions in orchestration.
  * Updated harness badge glyphs to display **Cu** and **Ws**.
  * Sessions now show under dedicated **Cursor** and **Windsurf** roster headings instead of **Other**.
* **Tests**
  * Expanded frontend tests to cover new glyph labels and correct roster grouping for both harness types.
* **Documentation**
[...]

PR #4779, opened 2026-07-10
fix(wallet): mint slash-free tiny.place identity keys

CodeGhost21:
## Summary

- Mint **slash-free** tiny.place identity keys so newly-derived identities can actually publish Signal pre-keys and receive DMs.
- Derive the identity seed through `slash_free_identity_seed()` at the point the seed is handed to `LocalSigner`.
- **counter 0 == the raw wallet seed**, so every already-clean identity is byte-identical — no migration for the common case.
- Only slash-bearing raw keys are remapped, deterministically, via a domain-separated SHA-256 counter.

## Problem

The tiny.place backend serves the Ed25519 identity public key as the Signal bundle `identityKey` in **STANDARD base64** and keys its `/keys/{identityKey}/…` routes on that string verbatim. Standard base64's alphabet includes `/`, which becomes `%2F` in the request path — and those key routes 404 on it. So an identity whose public key's base64 contains a `/` **cannot publish its Signal pre-keys and cannot receive DMs**.

Our tiny.place identity is *deterministic* from the Solana wallet key, and roughly **half** of all Ed25519 public keys contain at least one `/` in standard base64. So ~half of new wallets mint a silently-unusable tiny.place identity (bundle publish + inbound DM both 404). This is the same class of failure the `plugin-tinyplace` SDK already sidesteps by regenerating slash-free wallets.

## Solution

`tinyplace_signer_seed()` now returns `slash_free_identity_seed(signing_key.to_bytes())`:

- **counter 0 = the raw wallet seed.** If the raw key's base64 is already slash-free (the common case), it passes through **byte-identical** — no identity change, nothing to migrate.
- Only when the raw key's base64 contains `/` do we deterministically derive a fresh seed from `SHA-256("openhuman.tinyplace.identity.slashfree.v1" ‖ seed ‖ counter)`, bumping `counter` until the resulting public key is slash-free. Same wallet always maps to the same lowest-counter slash-free identity.
- Astronomically-unlikely all-slash exhaustion falls back to the raw seed rather than panicking.

The tiny.place plugin regenerates slash-free *wallets*; here the identity must remain a pure function of the wallet key, so we remap the seed deterministically instead.

## Submission Checklist

- [x] Tests added or updated — `slash_free_identity_seed_cleans_slash_keys_deterministically` covers the clean-passthrough path (counter 0, unchanged) and the slashed→remapped path (slash-free + deterministic).
- [ ] **Diff coverage ≥ 80%** — N/A to run full merged report locally; the added `fn` + its unit test are fully exercised (both branches). Flagging for the coverage gate to confirm.
- [x] Coverage matrix updated — `N/A: internal key-derivation fix, no user-facing feature row`.
- [x] No new external network dependencies introduced — `N/A`, uses already-imported `sha2` / `ed25519_dalek` / `base64`.
- [x] Manual smoke checklist updated — `N/A: no release-cut surface touched`.
- [ ] Linked issue closed — no dedicated issue; surfaced while wiring the Cursor/Windsurf harness observer against staging.

## Impact

- **Runtime:** desktop/CLI wallet identity derivation only.
- **Security:** domain-separated SHA-256; the remapped seed is still 32 bytes of Ed25519 secret material. No key logged (existing `debug` log omits the seed).
- **⚠️ MIGRATION CAVEAT (needs a rollout decision):** for a wallet whose raw key's base64 contains `/`, this **changes that wallet's advertised tiny.place identity** to a different (slash-free) key. In practice those identities were **already non-functional** — they 404 on bundle-publish and on inbound DMs — so there is no working relay state to preserve, and remapping is strictly a repair. But if any such identity was partially advertised (e.g. published out-of-band, printed in a UI, or saved by a peer), those references will no longer resolve after upgrade. Recommend the team confirm no persisted slash-bearing identities need an explicit migration/announcement before merge. Clean identities (counter 0) are unaffected.

## Related

- Closes: _(none — no tracked issue; discovered during the Cursor/Windsurf harness observer spike)_
- Follow-up PR(s)/TODOs: base58↔base64 reply-address unification (OpenHuman→runtime reply path), tracked separately.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/tinyplace-slash-free-identity`
- Commit SHA: 010c36441

### Validation Run

- [ ] `pnpm --filter openhuman-app format:check` — N/A (no `app/` changes)
- [ ] `pnpm typecheck` — N/A (no TS changes)
- [x] Focused tests: `cargo test -p openhuman --lib slash_free_identity_seed_cleans_slash_keys_deterministically` → 1 passed
- [x] Rust fmt/check (if changed): `cargo fmt --check -- src/openhuman/wallet/chains/solana.rs` → clean
- [ ] Tauri fmt/check (if changed): N/A

### Behavior Changes

- Intended behavior change: newly-derived tiny.place identities are always slash-free and therefore able to publish pre-keys / receive DMs.
- User-visible effect: none for clean identities; for slash-bearing wallets, the tiny.place identity (not the Solana wallet address) changes to a working one.

### Parity Contract

- Legacy behavior preserved: clean identities are byte-identical (counter 0 == raw seed).
- Guard/fallback/dispatch parity checks: deterministic remap; bounded loop with raw-seed fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana wallet identity seed handling to avoid problematic slash characters in encoded identity keys.
  * Preserved existing seeds when no adjustment is needed and ensured deterministic results when remapping is required.

### DIFF
--- a/src/openhuman/wallet/chains/solana.rs
+++ b/src/openhuman/wallet/chains/solana.rs
@@ -655,12 +655,95 @@ pub(crate) async fn tinyplace_signer_seed() -> Result<[u8; 32], String> {
     // Extract 32-byte SLIP-0010 secret — same bytes LocalSigner::from_seed expects.
     // Never logged: the log below omits the seed.
     log::debug!("[tinyplace] signer seed derived (seed not logged)");
-    Ok(signing_key.to_bytes())
+    Ok(slash_free_identity_seed(signing_key.to_bytes()))
+}
+
+/// Return a tiny.place identity seed whose Ed25519 public key has a **slash-free**
+/// STANDARD-base64 encoding.
+///
+/// The backend serves that base64 public key as the bundle `identityKey` and keys
+/// its `/keys/{identityKey}/…` routes on it verbatim; a `/` in the key becomes
+/// `%2F` in the path and the route 404s — so a slash-bearing identity can neither
+/// publish Signal keys nor receive DMs (the same failure the tiny.place plugin
+/// avoids by regenerating slash-free wallets). Our identity is *deterministic* from
+/// the wallet key, so instead of regenerating we bump a domain-separated counter:
+///
+/// - **counter 0 = the raw wallet seed** — every already-clean identity stays
+///   byte-identical (no migration for the common case),
+/// - only when the raw key's base64 contains `/` do we deterministically derive a
+///   fresh seed from `SHA-256(domain ‖ seed ‖ counter)` until the public key is
+///   clean. Same wallet → same (lowest-counter) slash-free identity, always.
+fn slash_free_identity_seed(base: [u8; 32]) -> [u8; 32] {
+    for counter in 0u32..100_000 {
+        let seed: [u8; 32] = if counter == 0 {
+            base
+        } else {
+            let mut hasher = Sha256::new();
+            hasher.update(b"openhuman.tinyplace.identity.slashfree.v1");
+            hasher.update(base);
+            hasher.update(counter.to_le_bytes());
+            hasher.finalize().into()
+        };
+        let pub_b64 = B64.encode(SigningKey::from_bytes(&seed).verifying_key().to_bytes());
+        if !pub_b64.contains('/') {
+            if counter > 0 {
+                log::debug!(
+                    "[tinyplace] raw wallet identity key had '/'; using slash-free derivation (counter={counter})"
+                );
+            }
+            return seed;
+        }
+    }
+    // Astronomically unlikely (~1 in 2^100000 that every candidate has a '/');
+    // fall back to the raw seed rather than panic.
+    base
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    fn pub_b64(seed: &[u8; 32]) -> String {
+        B64.encode(SigningKey::from_bytes(seed).verifying_key().to_bytes())
+    }
+
+    #[test]
+    fn slash_free_identity_seed_cleans_slash_keys_deterministically() {
+        // Find a base seed whose raw Ed25519 pubkey base64 contains '/', and one
+        // that's already clean (roughly half of all seeds are clean).
+        let (mut slashed, mut clean): (Option<[u8; 32]>, Option<[u8; 32]>) = (None, None);
+        for i in 0u32..5000 {
+            let mut s = [0u8; 32];
+            s[..4].copy_from_slice(&i.to_le_bytes());
+            if pub_b64(&s).contains('/') {
+                slashed.get_or_insert(s);
+            } else {
+                clean.get_or_insert(s);
+            }
+            if slashed.is_some() && clean.is_some() {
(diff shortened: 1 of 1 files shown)


CodeGhost21:
Closing this PR — re-diagnosis shows the fix is at the wrong layer, and is both **unnecessary** and **unsafe**. Thanks to @chatgpt-codex-connector for the P1 that prompted the deeper look.

## Why unsafe (Codex P1, confirmed)

`tinyplace_signer_seed()` builds the `LocalSigner`, and `signer.agent_id()` is used as the x402 payment **`from`** (`payment.rs:187`) and the registry `crypto_id`, while the on-chain spend (`execute_prepared`) uses the **original** Solana wallet key. Remapping the seed for slash-bearing wallets makes the payment/registry identity a *different, unfunded* address than the wallet that actually moves funds — x402 payments can be rejected after the on-chain transfer, and escrow/payout would target a cryptoId the user can't spend from. Because the user's funds live at the slash-bearing address and tiny.place requires `agentId == identityKey` (same key), you can't have both "slash-free identity" and "identity == funded wallet." This approach trades a messaging-transport bug for a payment-integrity bug.

## Why unnecessary (root cause is elsewhere, and already fixed)

The `/keys/:cryptoId/*` relay routes are keyed on the **base58** cryptoId (no `/`), and the backend already tries **both** encodings when resolving a bundle (`keyIdentifierCandidates` in `internal/controllers/relay/controller.go`). The 404 only happens when a **client puts the base64 `identityKey` (which can contain `/`) into the URL path** — `encodeURIComponent('/')` → `%2F`, which Gin decodes back to `/`, breaking the single-segment `:cryptoId` match.

Every current send path already avoids this by fetching the bundle by **base58 cryptoId**:

- **TS SDK v2.0.2** — `encryptEnvelope` calls `getBundle(deriveCryptoId(recipientEd25519))`, with an explicit comment about the `/`→404 problem (`sdk/typescript/src/messaging/encryption.ts`).
- **OpenHuman Rust core** — `resolve_recipient_to_agent_id()` returns the base58 `crypto_id`, which is what `client.keys.get_bundle()` receives (`manifest.rs:3320`).

The 404 I originally chased came from a **stale SDK build in the observer prototype** (`@tinyhumansai/tinyplace@1.0.1`), whose old `encryptEnvelope` fetched `getBundle(envelope.to)` — the base64 message key. That was fixed upstream in SDK **v2.0.2** (base58 fetch). OpenHuman core was never affected.

## Correct follow-up (no OpenHuman change)

- Bump the `plugin-tinyplace` SDK dependency `1.0.1 → ≥2.0.2` so its send path also routes by base58 (tracked separately in the tiny.place repo; see `tiny.place#213`).
- No identity-derivation change is warranted. Identity stays `== funded Solana wallet`.

Closing. The recognition slice (#4775) and the observer findings are unaffected by this.

PR #4839, opened 2026-07-14
feat(socket): medulla harness plane — drive an agent session over medulla:task_*

senamakel:
## Summary

Implements the OpenHuman side of the medulla session/streaming spec §6 (see tinyhumansai/medulla-v1 `docs/specs/session-streaming-api-spec.md`): a Socket.IO handler that lets a medulla operator running in the backend drive an OpenHuman agent session as a delegated sub-agent. medulla is the operator ("human in the loop").

Rides the existing authenticated `SocketManager` — no new transport dependency.

### Down (backend/medulla → agent), handled
- `medulla:task_run {taskId, cycleId, instruction, agentId?, timeoutMs}` — start/resume an agent session and drive it.
- `medulla:task_send {taskId, input}` — mid-task steering, delivered as follow-up turns.
- `medulla:task_abort {taskId}` — cooperative cancel of the in-flight turn.

### Up (agent → backend/medulla), emitted
- `medulla:task_envelope {taskId, envelope}` — the live session stream, reusing the existing `tinyplace.harness.session.v2` envelope kinds (`agent_message`, `agent_thinking`, `tool_call`, `tool_result`, `status`, `approval_request`, `error`).
- `medulla:task_result {taskId, ok, reply, usage?, error?}` — explicit completion (preferred over idle-detection).
- `medulla:register_agents {agents}` — roster advertisement on connect.

### Files
- `src/openhuman/socket/medulla/payloads.rs` — serde structs + event-name constants (round-trip tests).
- `src/openhuman/socket/medulla/envelope.rs` — `AgentProgress` → `HarnessEventKind` wrapped in `SessionEnvelopeV2` (reuses the v2 types).
- `src/openhuman/socket/medulla/mod.rs` — `MedullaTaskManager` (start/steer/abort, roster emit, timeout).
- `event_handlers.rs` / `mod.rs` — route the `medulla:task_*` events; advertise the roster on `ready`.

## Validation
`cargo check --lib` clean; `cargo test -p openhuman medulla` → 22 passed.

## Notes
- Mid-turn steering: since `run_single` is atomic, `task_send` input is delivered as sequential follow-up turns rather than injected mid-turn — the honest reuse of the existing runtime.
- Roster built from shipped `default_agents()`; surfacing enabled custom agents is a follow-up. `abort_all()` exists for disconnect teardown.

### DIFF
--- a/src/openhuman/socket/event_handlers.rs
+++ b/src/openhuman/socket/event_handlers.rs
@@ -69,6 +69,10 @@ pub(super) fn handle_sio_event(
                 "orch:register_tools",
                 crate::openhuman::orchestration::effect_executor::device_tool_manifest(),
             );
+            // Advertise this core's agent roster to the backend so a medulla
+            // operator can delegate `medulla:task_run` to a named agent. The
+            // backend clears the roster on socket disconnect.
+            super::medulla::emit_register_agents();
         }
         "error" => {
             log::error!("[socket] Server error event: {}", data);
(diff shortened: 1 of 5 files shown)


senamakel:
Addressed the review feedback (commit `2d725853`):

- **P2 — scope sessions by session id**: each medulla session's agent transcript is now keyed via `set_agent_definition_name(medulla_session_key(agentId, sessionId))` (mirroring the web channel's per-thread scoping), so two `task_run`s with the same `agentId` get isolated transcripts instead of resuming each other's history.
- **P2 — preserve pre-turn aborts**: swapped the bare `Notify` for a `CancellationToken` (already a crate dep). Because it latches, an abort delivered while the driver is still awaiting `build_agent()`/integration fetches is observed on the first poll of the biased select and settles the task cancelled — no lost `notify_waiters()` race.
- **P2 — whole-task timeout**: `timeoutMs` is now a single wall-clock deadline computed once; each turn (initial + every `task_send` follow-up) passes the *remaining* budget to `run_with_optional_timeout`, and an exhausted budget settles the task timed out. Terminal `medulla:task_result` semantics preserved (`aborted` / `timeout`).

`cargo check` clean, `cargo test -p openhuman medulla` → 24 passed (incl. new session-key and remaining-budget tests), fmt clean.

PR #4842, opened 2026-07-14
fix(socket): medulla harness — session-scoped transcripts, latching abort, whole-task timeout

senamakel:
## Summary

Follow-up to the merged medulla harness plane (#4839), addressing the code-review findings on `src/openhuman/socket/medulla/mod.rs`:

- **Scope sessions by session id (P2):** each medulla session's agent transcript is now keyed via `set_agent_definition_name(medulla_session_key(agentId, sessionId))` (mirroring the web channel's per-thread scoping), so two `medulla:task_run`s with the same `agentId` get isolated transcripts instead of resuming each other's history.
- **Preserve pre-turn aborts (P2):** swapped the bare `Notify` for a `CancellationToken` (already a crate dep). Because it latches, an abort delivered while the driver is still awaiting `build_agent()`/integration fetches is observed on the first poll of the biased select and settles the task cancelled — no lost `notify_waiters()` race. Aborts before build, during build, and during the turn all settle cancelled.
- **Whole-task timeout (P2):** `timeoutMs` is now a single wall-clock deadline computed once; each turn (initial + every `task_send` follow-up) passes the *remaining* budget to `run_with_optional_timeout`, and an exhausted budget settles the task timed out.

Terminal `medulla:task_result` semantics preserved (`aborted` / `timeout`).

## Validation
`cargo check --lib` clean; `cargo fmt --check` clean; `cargo test -p openhuman medulla` → 24 passed (incl. new `session_key_scopes_transcript_by_session_id` and `remaining_budget_reports_time_left_and_exhaustion` tests).

### DIFF
--- a/src/openhuman/socket/medulla/mod.rs
+++ b/src/openhuman/socket/medulla/mod.rs
@@ -27,8 +27,9 @@ use std::sync::atomic::{AtomicI64, Ordering};
 use std::sync::{Arc, OnceLock};
 
 use parking_lot::Mutex;
-use tokio::sync::{mpsc, Notify};
-use tokio::time::Duration;
+use tokio::sync::mpsc;
+use tokio::time::{Duration, Instant};
+use tokio_util::sync::CancellationToken;
 
 use crate::openhuman::agent::progress::AgentProgress;
 use crate::openhuman::agent::turn_origin::{with_origin, AgentTurnOrigin};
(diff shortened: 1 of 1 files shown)


senamakel:
Remaining OpenHuman-side follow-ons (wire `MedullaTaskManager::abort_all()` to socket-disconnect teardown; surface enabled custom agents beyond `default_agents()` in the `medulla:register_agents` roster) are tracked under tinyhumansai/backend#1090 alongside the cross-replica hardening.

Issue #3202, opened 2026-06-02
Add hands-free Clippy desktop control on Windows and macOS


Al629176:
## Summary

Implement a fully hands-free desktop control experience for Clippy using macOS Accessibility APIs, including always-listening audio input, hotkey activation, and reliable UI element clicking.

## Problem / Context

Users should be able to control their desktop through Clippy without needing to manually click through the app. The current experience does not provide a complete hands-free flow that can listen, understand user intent, identify macOS UI elements, and perform reliable actions.

## Scope (optional)

Use macOS Accessibility APIs for UI element discovery and action execution. Support both always-listening mode and hotkey activation, with clear user consent and permission prompts. The implementation should prioritize reliable clicking of buttons, menus, fields, and common UI controls.

## Acceptance criteria

- [ ] **Always-listening mode exists** — Clippy can listen for user commands when enabled by the user.
- [ ] **Hotkey activation exists** — Users can activate voice control through a configurable hotkey.
- [ ] **macOS accessibility integration works** — Clippy can inspect accessible UI elements through macOS Accessibility APIs.
- [ ] **Reliable clicking** — Clippy can click common UI elements such as buttons, menu items, input fields, and controls.
- [ ] **Permission safe** — The flow clearly requests and verifies microphone and accessibility permissions.
- [ ] **User control preserved** — Users can disable always-listening and desktop-control behavior at any time.
- [ ] **Regression safety** — Integration or E2E coverage verifies voice activation and representative desktop-control actions.
- [ ] **Diff coverage >= 80%** — Any implementing PR meets the changed-lines coverage gate when code changes are involved.

## Related

- Voice-driven desktop control and Clippy/Human assistant roadmap.

### Triage Assessment
- This issue is actively assigned to @M3gA-Mind who posted a detailed 6-PR implementation plan on 2026-07-15. PR 1 (reliable element selection in `element_match.rs`) is already implemented with PR pending.
- Most of the underlying capability exists in the tree (always-listening, hotkey activation, macOS Accessibility APIs, vision-click, desktop_control_agent). The remaining work is sequenced into 6 focused PRs.

### Proposed Scope (per assignee's plan)
- PR 1: Reliable element selection (ranked matcher in `accessibility/element_match.rs`).
- PR 2: Press disambiguation (use `best_match` tier to pick best, clarify on contest).
- PR 3: Windows UIA parity (route through same matcher).
[...]

PR #4905, opened 2026-07-15
feat(accessibility): ranked UI-element matcher for reliable selection (#3202 slice 1)

M3gA-Mind:
Addresses #3202 (foundational slice — the full hands-free feature is planned in 6 slices; see the plan comment on the issue).

## Summary
First slice of hands-free Clippy desktop control: a **ranked accessibility element matcher** — the 'reliable UI element clicking' the issue calls out. Given a target description, it scores candidate accessibility elements and returns the best match (with an `ambiguous` signal for later press-disambiguation), then drives the action through the macOS AX executor.

## Changes
- `src/openhuman/accessibility/element_match.rs` (+179) — pure ranking/matching logic, platform-agnostic.
- `src/openhuman/accessibility/element_match_tests.rs` (+188) — exhaustive unit tests (>80% changed-line coverage).
- `src/openhuman/accessibility/ax_interact.rs` — wire the matcher into the macOS AX click path (`cfg(target_os = "macos")`).
- `mod.rs` — exports.

## Scope / follow-ups (from the plan on #3202)
Deliberately just the matcher + executor wiring. Next: PR2 press disambiguation (`best_match.ambiguous`), PR3 Windows UIA parity via the same matcher, PR4 structured desktop-control intents, PR5 hands-free wiring + consent UX, PR6 integration/E2E.

Matcher logic is platform-agnostic and Linux-tested; the AX wiring is macOS-gated (excluded from the Linux diff-cover denominator). No AI attribution.

## Summary by CodeRabbit

* **New Features**
  * Accessibility element matching now prioritizes exact, prefix, and stronger contextual matches.
  * Matching handles differences in capitalization, spacing, and trailing ellipses more reliably.
  * Best-match selection can identify ambiguous results for safer interaction.

* **Bug Fixes**
  * Improved consistency when filtering accessibility elements.
  * Preserved stable ordering among equally ranked matches.

* **Tests**
  * Added coverage for normalization, ranking, filtering, and ambiguity detection.


M3gA-Mind:
Thanks @coderabbitai — addressed all three in the latest push. All were valid.

**1. `element_match.rs` — word-boundary needed a trailing check (actionable).** Correct: `is_word_boundary_match` only guarded the leading side, so `changes` matched inside `changeset`. It now requires a non-alphanumeric (or edge) boundary on **both** sides; genuine word matches stay `WordBoundary`, partial-word hits fall through to the weaker `Substring`. Added a regression test (`classify_partial_word_is_substring_not_word_boundary`).

**2. `element_match_tests.rs` — test didn't exercise its named branch (actionable).** Correct: `"Save…"` classified as `NormalizedExact` while `"save"` hit `Exact` (literal equality is checked before normalization), so the winning tier had a single member and the same-tier rival comparison never ran. Rewrote it to put both candidates in the same `NormalizedExact` tier (`"Save As…"` / `"Save   As"` for query `"Save As"`), so the `normalize(&el.label) != best_norm` path is actually invoked, and asserted the tier.

**3. `ax_interact.rs:110-123` — diagnostics + swallowed parse error (nitpick).** Correct, and it aligns with the repo's "verbose diagnostics on new/changed flows" rule. A malformed `elements` decode is now `log::warn!`-ed instead of silently collapsing to an empty list (behavior still degrades to empty), and both filter branches log their element counts (`unfiltered → N`, `filter=… → M of N matched`).

`GGML_NATIVE=OFF cargo check -p openhuman --lib` is clean and the `element_match` unit tests pass locally.

PR #4917, opened 2026-07-15
fix(voice): ship the voice domain in the desktop build (#4901)

CodeGhost21:
## Summary

- The shipped desktop app has been built **without the voice domain** since v0.58.19 — every `openhuman.voice_*` RPC answers `unknown method`, so voice transcription is completely non-functional.
- Root cause: the Tauri shell declares `openhuman_core = { ..., default-features = false }`, which silently dropped the default-ON `voice` feature added by #4833. Fixed by opting `voice` back in explicitly.
- Added a compile-time `const` assert so a voice-less desktop binary **fails the build** instead of shipping and failing silently at runtime.
- Replaced the "restart to pick up the latest core sidecar" error copy — it was written for #1289, refers to a sidecar removed in #1061, and can never resolve a compile-time gate.

## Problem

Since PR #1061 the core runs **in-process inside the Tauri shell**, so the shell's dependency declaration is what actually ships. That declaration has carried `default-features = false` since the in-process move (`app/src-tauri/Cargo.toml:135`, dated 2026-05-16), back when the core crate had **no `[features] default` list at all** — a harmless no-op.

#4833 (implementing #4803) then made `voice` a default-ON compile-time feature so slim/headless builds could drop it. That collided with the pre-existing `default-features = false`: the desktop build stopped enabling `voice`, so the `#[cfg(feature = "voice")]` controller registrations in `src/core/all.rs` (lines 210, 641) were never compiled in, and the whole `openhuman.voice_*` namespace became unknown-method.

Because the gate is **compile-time**, restarting or reinstalling can never help — the code is absent from the binary. This is not a migration or data issue: a clean install of an affected release fails identically.

**Evidence**

- `hound` — reachable *only* via `voice = ["dep:hound", "dep:lettre"]` — was absent from the shell's dependency graph (`cargo tree -i hound` → "did not match any packages") while present in the root crate's graph. Same probe, same worktree; the only difference is `default-features = false`.
- Production Sentry confirms the whole namespace is down, not just STT:
  - [#25260](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/25260/) `unknown method: openhuman.voice_status` — 46,735 events / 56 users
  - [#25259](https://sentry.tinyhumans.ai/organizations/tinyhumans/issues/25259/) `unknown method: openhuman.voice_server_status` — 46,592 events / 52 users
[...]

Issue #4901, opened 2026-07-15
Voice transcription unavailable — 'Voice transcription is unavailable in this build' error on latest release


Al629176:
## Summary

Clicking the microphone button to use voice input shows the following error in red:

> "Transcription failed: Voice transcription is unavailable in this build. Restart the OpenHuman desktop app to pick up the latest core sidecar."

The error persists after restarting the app. Voice input is completely non-functional.

## Problem

The error message instructs the user to restart the app to pick up the latest core sidecar, but restarting does not resolve the issue. This suggests the STT (speech-to-text) sidecar binary is either:
- Missing from the build entirely
- Not being bundled correctly for this platform
- Failing to launch silently so the app falls back to this error state

The "Tap and speak" UI is visible and the microphone button is active, but no transcription occurs.

## Steps to reproduce

1. Open OpenHuman desktop app (latest build)
2. Click the microphone button in the chat input area
3. Observe: red error — "Transcription failed: Voice transcription is unavailable in this build"
4. Restart the app as instructed
5. Observe: error persists

## Expected behaviour

- Clicking the microphone button activates voice input
- Speech is transcribed in real time and placed into the chat input
- No error message is shown

## Acceptance criteria

- [ ] Voice transcription works on a clean install of the latest build
- [ ] STT sidecar (`ggml-base-q5_1.bin` or equivalent) is correctly bundled and launched
- [ ] Restarting the app is not required to activate voice — it should work on first launch
- [ ] If the sidecar fails to start, the error message includes a specific reason and a direct action (not just "restart the app")
- [ ] Verified working on macOS, Windows, and Linux

## Related

- Voice feature gate: #4803
- STT model: `ggml-base-q5_1.bin` (default from `effective_stt_model_id`)
- Local AI sidecar management: Rust core / sidecar launcher

PR #4934, opened 2026-07-15
fix(build): forward tokenjuice-treesitter + gate feature-forwarding drift in CI (#4918, #4919)

### LINKED ISSUE #4918: AST-aware code compression silently disabled in desktop builds (tokenjuice-treesitter dropped)
## Summary

The desktop app is built without the core's default `tokenjuice-treesitter` feature, so AST-aware code compression silently falls back to the brace-depth heuristic in every shipped build. Same root cause as #4901, but it fails soft — no error, no Sentry signal, just quietly worse output.

## Problem

`app/src-tauri/Cargo.toml` declares the core with `default-features = false`. That line predates the core having any defaults (written 2026-05-16, when `[features]` had no `default` list), so it silently drops each default the core later gained. #4901 fixed `voice`; `tokenjuice-treesitter` — the core's other default — is still dropped.

**What happens:** `tokenjuice-treesitter = ["tinyjuice/tinyjuice-treesitter"]` is never enabled for the shipped binary, so tinyjuice uses the brace-depth heuristic instead of tree-sitter grammars.

**What's expected:** the desktop app compresses code with the AST-aware path, as `default = ["tokenjuice-treesitter", "voice"]` intends.

**Steps to reproduce** (same probe + control that diagnosed #4901):

```bash
# Shell (what ships) — tree-sitter ABSENT
cd app/src-tauri && cargo tree -i tree-sitter
#   error: package ID specification `tree-sitter` did not match any packages

# Control: root crate (tokenjuice-treesitter default-ON) — tree-sitter PRESENT
cd ../.. && GGML_NATIVE=OFF cargo tree -i tree-sitter
#   tree-sitter v0.26.10
#   └── tinyjuice v0.2.1 → openhuman v0.61.2
```

**Version/platform:** verified on `upstream/main` @ `0e86b85c9` (macOS). Affects all desktop platforms and every release since the feature became a default.

Unlike #4901 this is **not** user-visible as an error — it degrades output quality silently, which is why it has gone unnoticed.

## Solution (optional)

Likely one line — add `tokenjuice-treesitter` to the `openhuman_core` feature list in `app/src-tauri/Cargo.toml`:

```toml
openhuman_core = { path = "../..", package = "openhuman", default-features = false, features = [
    "voice",
    "tokenjuice-treesitter",
] }
```

**Decide first whether the drop is intentional.** The shell separately declares `tinyjuice = { version = "0.2.1", default-features = false }`, which hints the tree-sitter C build may be avoided deliberately (build time / cross-compilation). If so, the fix is to document that on the `default` list rather than enable it. This is exactly why it was excluded from #4917 instead of bundled in.

Enabling it adds tree-sitter's C grammar build to the desktop build — measure the build-time and binary-size cost before merging.

## Acceptance criteria

- [ ] **Decision recorded** — confirm whether shipping without tree-sitter was intentional; if so, document it and close.
- [ ] **Repro gone** — `cargo tree -i tree-sitter` from `app/src-tauri` resolves the crate (or the intentional-drop decision is documented).
- [ ] **Regression safety** — covered by the feature-forwarding CI guard (see Related); no silent drop possible.
- [ ] **Cost measured** — build-time / binary-size delta from the tree-sitter C build noted on the PR.

## Related

- Same root cause as #4901 / #4917 (voice dropped by the same `default-features = false`)
- Class-level fix: feature-forwarding drift guard
- Feature added as a core default in the `tokenjuice-treesitter` work

### LINKED ISSUE #4919: Fail CI when the desktop shell does not forward a default-ON core feature gate
## Summary

Add a CI check that fails when the Tauri shell does not forward every default-ON gate from the core crate. Today a missing gate silently removes a whole domain from the shipped desktop app with no build error, no test failure, and no signal until users report it.

## Problem / Context

`app/src-tauri/Cargo.toml` declares the embedded core with `default-features = false`, so the shell does **not** inherit the core's `default` gate set. Every default-ON gate must be forwarded by hand:

```toml
openhuman_core = { path = "../..", package = "openhuman", default-features = false, features = [
    "media",
    "voice",
] }
```

Nothing enforces that list matches the core's `default`. When they drift, the domain is compiled out of the shipped binary and the failure is **invisible at build time** — it surfaces only as runtime `unknown method` errors, or (worse) as silently degraded behaviour with no error at all.

**This has already happened twice, independently:**

| Gate | Outcome |
| --- | --- |
| `voice` | Not forwarded when #4833 made it default-ON. Shipped broken v0.58.19 → v0.61.x. 56 users, ~93k Sentry events (#25260, #25259). Every `openhuman.voice_*` RPC was unknown-method. Fixed in #4917 (#4901). |
| `media` | #4840 forwarded it correctly — its author hit the same trap and documented the rule in a code comment. |
| `tokenjuice-treesitter` | Still not forwarded. Fails soft (AST compression degrades to a heuristic), so there is no error to notice. #4918 |

Two of three defaults have been dropped at some point. The `media` PR proves the trap is discoverable only if the author happens to know about it — a code comment is documentation, not enforcement. The next default-ON gate will hit it again.

#4917 added a `VOICE_COMPILED_IN` const assert, but that guards **one** gate. `media` has no equivalent, and a per-gate assert doesn't scale or auto-cover new gates.

## Scope (optional)

**In scope:** a check that parses the core's `[features] default` list and the shell's forwarded `features` list, and fails when a default-ON gate is not forwarded (or is not explicitly allow-listed as intentionally excluded).

Notes / tradeoffs:

- An **allow-list for deliberate exclusions** is required — `tokenjuice-treesitter` may be intentionally dropped (#4918). The check should demand an explicit opt-out with a reason, so "excluded on purpose" and "forgotten" stop looking identical.
- A **static manifest diff** (script comparing the two TOMLs) is cheap and fast — no compile needed. Preferred.
- A **resolved-graph assertion** (`cargo tree -e features -i openhuman` in the shell, asserting each expected `feature "<gate>"`) catches unification subtleties a text diff would miss, at the cost of a resolve step. This is the probe that actually diagnosed #4901.
- Slim/headless builds intentionally drop gates — the check must only apply to the **desktop shell**, not `--no-default-features` profiles.

**Out of scope:** fixing `tokenjuice-treesitter` itself (#4918), and the per-gate const asserts (they can stay as defence in depth, or be retired once this lands).

## Acceptance criteria

- [ ] **Drift fails CI** — removing any forwarded gate from `app/src-tauri/Cargo.toml` fails the check with a message naming the missing gate. Verify red/green, not just green.
- [ ] **New defaults are covered automatically** — adding a gate to the core's `default` without forwarding it fails, with no per-gate wiring needed.
- [ ] **Intentional exclusions supported** — an explicit allow-list entry with a documented reason passes (`tokenjuice-treesitter` is the first real case).
- [ ] **Slim builds unaffected** — `--no-default-features` profiles still build.
- [ ] **Wired into CI Lite** — runs on PRs touching either manifest.
- [ ] **Documented** — the rule lives in AGENTS.md "Compile-time domain gates" alongside the existing guidance.
- [ ] **Diff coverage ≥ 80%** — the implementing PR meets the changed-lines coverage gate when code changes are involved.

## Related

- #4901 / #4917 — voice dropped by this exact mechanism (the motivating incident)
- #4918 — `tokenjuice-treesitter` still un-forwarded; first allow-list candidate
- #4803 / #4833 — voice gate that introduced the regression
- #4804 / #4840 — media gate; independently hit the same trap and documented the rule in a comment
- AGENTS.md § "Compile-time domain gates"

CodeGhost21:
> **Stacked on #4917 — merge that first.** GitHub cannot base a PR on a fork branch, so this targets `main` and its diff includes #4917's commits until that lands. The commits unique to this PR are `8a3d89260` onward.

## Summary

- The desktop app has **never** shipped with tree-sitter: the shell's `default-features = false` silently dropped the default-ON `tokenjuice-treesitter` gate, so TokenJuice fell back to the brace-depth heuristic instead of the AST-aware path (#4918).
- Adds a CI gate that fails when the shell stops forwarding any default-ON core gate — the class of bug behind both #4901 and #4918 (#4919).
- The guard covers **new** gates automatically, with no per-gate wiring, and supports explicit intentional exclusions with a documented reason.
- Documents the rule in AGENTS.md, next to the compile-time gate docs.

## Problem

`app/src-tauri/Cargo.toml` declares `openhuman_core` with `default-features = false` (set in #1061, back when the core had no `default` list at all — a harmless no-op then). The shipped app therefore does **not** inherit the core's default gates; each must be forwarded by hand. Nothing enforced that.

**Two of three gates have been dropped, by two different authors:**

| Gate | Outcome |
| --- | --- |
| `voice` | Not forwarded when #4833 made it default-ON. Shipped broken v0.58.19 → v0.61.x. 56 users, ~93k Sentry events (#25260, #25259). Fixed in #4917. |
| `media` | #4840 forwarded it correctly — its author hit the same trap and documented the rule in a code comment. |
| `tokenjuice-treesitter` | Never forwarded once since #4123. Fails **soft** — no error, no Sentry signal, just silently worse compression. This PR. |

The `media` author *knew about the trap and wrote it down*, and `voice` still shipped broken. That is the argument for enforcement over documentation.

**#4918 is not a deliberate exclusion.** #4575 ("fix tinyjuice build wiring") explicitly set out to *"forward the correct TinyJuice tree-sitter feature name"* and named broken `tokenjuice-treesitter` resolution as its root cause — the intent was clearly that this works. The shell's separate `tinyjuice = { default-features = false }` is version alignment for the vendored 0.2.1 patch, not a veto.

**Reproduction** (the probe that diagnosed #4901):

```bash
cd app/src-tauri && cargo tree -i tree-sitter
#   error: package ID specification `tree-sitter` did not match any packages
cd ../.. && GGML_NATIVE=OFF cargo tree -i tree-sitter   # control: core has it
#   tree-sitter v0.26.10 → tinyjuice v0.2.1 → openhuman
```

## Solution

- **`app/src-tauri/Cargo.toml`** — forward `tokenjuice-treesitter`. Cost: 5 crates (`tree-sitter` + Rust/TS/Python grammars, C build); the core crate already pays it.
- **`scripts/lib/feature-forwarding.mjs`** — narrow TOML reader + drift diff. Node has no TOML dependency here and only two well-known shapes are needed, so it is scanner-based in the same spirit as the sibling `checklist-parser.mjs`.
- **`scripts/ci/check-feature-forwarding.mjs`** — CLI plus `INTENTIONALLY_NOT_FORWARDED`, **empty by design**. An intentional exclusion must carry a reason, so "deliberate" and "forgotten" stop looking identical — precisely the ambiguity that let #4918 sit unnoticed.
- **`.github/workflows/ci-lite.yml`** — new always-on lane wired into `PR CI Gate`. Deliberately **not** filtered on changed paths: drift can be introduced from *either* manifest, and a skipped job counts as a pass in the gate. It is a few seconds of dependency-free Node.
- **`AGENTS.md`** — the rule, next to the gate docs.

**The checker refuses to pass vacuously**: parsing zero core defaults exits 2 rather than reporting success. That caught a real bug in my own parser during development — `\s` matches newlines, so `^\s*\[features\]` anchored several lines early and sliced the section to nothing. Without that guard it would have shipped as a green rubber stamp.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — 18 unit tests over the new modules; `pnpm test:scripts` (which CI runs) passes 103 tests.
- [x] Coverage matrix updated — `N/A: build-configuration + CI tooling; no feature rows added/removed/renamed`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no user-facing behaviour change; TokenJuice compression quality is internal.`

## Impact

**Runtime/platform:** desktop (macOS/Windows/Linux). Restores AST-aware code compression to the shipped app. No user-visible behaviour change — TokenJuice already worked, just worse.

**Build cost:** adds `tree-sitter`, `tree-sitter-language`, `tree-sitter-python`, `tree-sitter-rust`, `tree-sitter-typescript` to the shell's Cargo world (the core already builds them). C grammar compilation, so the shell's cold build grows modestly.

**Verification**

| Check | Result |
| --- | --- |
| `cargo tree -i tree-sitter` in shell | "did not match any packages" → **resolves** via tinyjuice → openhuman → OpenHuman |
| Shell `cargo check` with tree-sitter | green (exit 0) |
| Guard on current manifests | green (exit 0) |
| Guard with `voice` + `tokenjuice-treesitter` dropped | **fails, exit 1, naming both** |
| Guard unit tests | 18 pass — incl. cases reproducing #4901 and #4918, and a new-gate case proving automatic coverage |
| `pnpm test:scripts` (full suite CI runs) | 103 pass |
| `ci-lite.yml` | valid YAML; job + `needs` + gate `results` map all wired |

**Compatibility / migration:** none. Build configuration only.

## Related

Closes #4918
Closes #4919

- Depends on #4917 (voice — same root cause; merge first)
- #4123 — added `tokenjuice-treesitter` as a core default; never forwarded
- #4575 — "fix tinyjuice build wiring"; intended tree-sitter to resolve
- #4833 / #4803 — voice gate that caused #4901
- #4840 / #4804 — media gate; hit the same trap and documented it in a comment
- Feature IDs: `N/A — no coverage-matrix rows affected (build config + CI tooling)`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when voice transcription is unavailable in the installed app version.
  * Users now receive a clear localized message recommending an app update instead of a technical error.
  * Prevented repeated transcription retries when voice support is not included.

* **Localization**
  * Added the voice transcription availability message across supported languages.

* **Reliability**
  * Improved desktop build validation to help ensure voice transcription support is included consistently.


M3gA-Mind:
Maintainer review changes (merge-conflict resolution only — no logic changes to the PR's own diff):

**Merged current `main` into the PR branch** to clear two conflicts left by the divergence from its stacked base (#4917, now in main):

1. **`app/src-tauri/Cargo.toml`** — the real conflict. The PR branch added `tokenjuice-treesitter` to a forwarding list that predated `web3` being forwarded; `main` had since added `web3` and still carried the now-obsolete "tokenjuice-treesitter is the remaining un-forwarded default … deliberately not bundled" note. Resolved to the **union**: the `openhuman_core` features list now forwards `media`, `tokenjuice-treesitter`, `voice`, `web3`; kept `main`'s `web3` comment bullet + the PR's `tokenjuice-treesitter` bullet + the CI-checker note, and dropped `main`'s stale "un-forwarded" paragraph (this PR forwards it). Verified with the PR's own gate: `node scripts/ci/check-feature-forwarding.mjs` → `OK: every default-ON core gate is forwarded` (4/4).

2. **`app/src/features/human/voice/sttClient.ts`** — incidental, from the stacked voice commits. `main` refined `isVoiceNotCompiledError` from a plain `: boolean` return into a proper TS type guard (`: err is VoiceNotCompiledError`) with an identical body. Took `main`'s version (strict improvement, same runtime behavior).

No changes to the PR's substantive additions (the checker scripts, CI lane, AGENTS.md rule).

M3gA-Mind:
Re CodeRabbit's 2 nitpicks on `app/src/features/human/voice/sttClient.test.ts` (assert `VOICE_NOT_COMPILED_CODE` / `isVoiceNotCompiledError` alongside the message text): **declining as out-of-scope for this PR.** That test file is not part of #4934's net diff against `main` (`git diff main...HEAD` covers only the feature-forwarding checker, the CI lane, `Cargo.toml`/`Cargo.lock`, and AGENTS.md) — it belongs to the stacked #4917 voice work that has since merged into `main` and appears here only transiently. The suggestion is reasonable on its own merits and better raised against the voice test suite directly; it is unrelated to the tree-sitter forwarding fix and its non-blocking (`COMMENTED`, 🔵 Trivial). No change needed on this PR.

M3gA-Mind:
Maintainer sync: merged current main (includes #4949, the fix for the red-main typecheck failure tracked in #4948). The Frontend Checks failure on this PR was inherited from main and not caused by this PR's changes. Merge was clean — no conflicts. This PR's Rust feature-gate changes are untouched; the merge only brings in the repaired main.

M3gA-Mind (approved):
Approving — both merge conditions are genuinely met.

**CI: fully green, 18/18 lanes pass, zero pending, zero failing** — including every lane that matters for this change:
- `Feature Forwarding Gate (shell forwards core defaults)` — the PR's own new lane, passing against the merged tree.
- `Rust Quality (fmt, clippy)`, `Rust Feature-Gate Smoke (gates off)`, `Rust Core Coverage`, `Rust Tauri Coverage` — the Tauri lane actually compiles the shell with the new feature set.
- `Frontend Checks` — passes. This lane previously failed on this PR purely by inheriting the red `main` (issue #4948, fixed by #4949); the re-sync onto the repaired main cleared it. The failure was never caused by this PR.
- `PR CI Gate` — the rollup, green.

**Review state:** no `CHANGES_REQUESTED`. CodeRabbit's review is `COMMENTED`; its two 🔵-trivial nitpicks were on `sttClient.test.ts`, which is not in this PR's net diff (carried transiently from the stacked, already-merged #4917) — correctly declined in-thread rather than silently dropped.

**What was verified beyond CI:**
- Two merge conflicts against `main` were resolved to the union: `app/src-tauri/Cargo.toml` now forwards `media`, `tokenjuice-treesitter`, `voice`, `web3` (main's `web3` bullet + this PR's `tokenjuice` bullet kept; main's now-false "un-forwarded" paragraph dropped). `sttClient.ts` took main's stricter type-guard refinement from #4917.
- The resolution was checked with the PR's own gate: `node scripts/ci/check-feature-forwarding.mjs` → `OK: every default-ON core gate is forwarded (4/4)`.

**Why this is worth having:** it restores AST-aware compression to the shipped desktop build *and* adds durable CI enforcement against the exact drift class behind two prior incidents — a checker that is dependency-free, fails closed, and has its own passing lane. The fix and the guard against its recurrence land together.

Not merging — that is the maintainer's call.

M3gA-Mind:
@coderabbitai review

Requesting a re-review of the current head. Your earlier review on this PR was dismissed, and the tree has since moved on: the branch was re-synced onto current `main` (which now includes #4949, the fix for the red-main typecheck failure tracked in #4948), and the two merge conflicts were resolved to the union. All 18 CI lanes are green, including this PR's own `Feature Forwarding Gate`.

M3gA-Mind:
@coderabbitai review

Issue #4688, opened 2026-07-08
bug: OAuth provider mismatch on sign-in creates new workspace and orphans existing one — no recovery path


Al629176:
## Summary

When a user signs in with a different OAuth provider than they originally used (e.g. GitHub instead of Google), OpenHuman creates a brand new local workspace under a different `users/<id>` path, silently abandoning the existing workspace. The old workspace — containing all synced memory, vault content, and configuration — is left orphaned on disk with no UI to recover or switch back to it.

## Problem

OpenHuman ties workspace identity to the OAuth provider account. If the user picks a different provider on a subsequent sign-in (easy to do if they cannot remember which they originally used), the result is:

```
Old workspace:  C:\Users\<user>\.openhuman\users\6a07285ec4a2001c6245e641  ← orphaned, all data here
New workspace:  C:\Users\<user>\.openhuman\users\6a4c727d3e6105bc02f27512  ← empty, newly active
```

After the OAuth flow, OpenHuman restarts and presents as freshly initialized:
```
0 memory graph
no connected sources
vault setup warnings
```

The old workspace still exists on disk — data is not deleted — but there is no UI to:
- See that a previous workspace exists
- Select or restore it
- Understand that a provider mismatch caused the switch

This is effectively a silent data loss event from the user's perspective.

## Steps to reproduce

1. Install OpenHuman and sign in with Provider A (e.g. Google). Set up sources, sync memory.
2. Sign out or reinstall.
3. Sign in again — choose Provider B (e.g. GitHub) instead of Provider A.
4. Observe: OpenHuman restarts as if newly initialized. All prior memory, sources, and config appear gone.
5. Check `%USERPROFILE%\.openhuman\users\` — two `<id>` folders exist; the old one still has content.

## Expected behaviour

- If a sign-in resolves to a different identity than the current workspace, warn the user before switching: *"This account doesn't match your current workspace. Switch accounts? Your existing data will remain accessible."*
- Provide a workspace selector so users can see and switch between local workspaces tied to different accounts.
- At minimum: detect orphaned workspaces on startup and surface a recovery option.

## Acceptance criteria

- [ ] **Provider mismatch warning** — Signing in with a different OAuth provider that resolves to a new identity shows a confirmation dialog before creating a new workspace.
- [ ] **Workspace selector** — UI (Settings or onboarding) allows switching between local `users/<id>` workspaces.
[...]

PR #4954, opened 2026-07-16
fix(settings): scope Clear App Data reset to the signed-in user (#4950)

### LINKED ISSUE #4950: 'Clear app data' button in Settings does nothing — app data is not actually cleared
## Summary

The **Clear app data** button on the Settings page appears to do nothing. Clicking it does not clear any app data — memory, cached files, credentials, and local storage all remain intact after clicking.

## Problem

Users expect "Clear app data" to wipe local app state — similar to clearing app data on mobile or clearing browser storage. Currently the button either:
- Has no handler wired up (click does nothing visually or functionally)
- Fires a handler that errors silently without actually deleting anything
- Clears a subset of data but leaves the majority untouched

After clicking the button, the app continues to show the same memory, sources, and settings as before — confirming nothing was cleared.

## What "Clear app data" should do

At minimum, a full clear should remove:
- [ ] Local memory / knowledge graph data (`memory_tree/`, `vault/`)
- [ ] Cached embeddings and vector store
- [ ] Conversation history
- [ ] Workflow run history
- [ ] Locally stored credentials / API keys (with explicit warning before clearing)
- [ ] App preferences and settings (with explicit warning)
- [ ] Composio / integration tokens (with explicit warning)
- [ ] Temporary files and sidecar caches

A confirmation dialog should be shown before any destructive action, clearly listing what will be deleted and warning that this cannot be undone.

## Steps to reproduce

1. Open OpenHuman → Settings
2. Click **Clear app data**
3. Observe: nothing happens, or a confirmation appears but data is not cleared
4. Navigate to Brain → Memory → Sources — all data still present

## Acceptance criteria

- [ ] Clicking **Clear app data** triggers a confirmation dialog listing exactly what will be deleted
- [ ] Confirming the dialog actually deletes all local app data (memory, cache, conversation history, workflow history)
- [ ] After clearing, the app restarts or reloads to a clean first-run state
- [ ] Sensitive items (API keys, credentials) are called out separately in the confirmation with an additional warning
- [ ] The button works on macOS, Windows, and Linux
- [ ] Tested: after clearing, Brain → Memory shows empty, no prior conversations visible

## Related

- Privacy controls: #4256
- Local data paths: `%LOCALAPPDATA%\openhuman` (Windows), `~/.openhuman` (macOS/Linux)
- spaCy venv cleanup should also be included: #4688

M3gA-Mind:
## Summary

- **Fix "Clear App Data does nothing" (#4950):** the button was fully wired, but the reset deleted the wrong directory and left all user data intact.
- The clear flow signs the user out (`auth_clear_session` removes `active_user.toml`) **before** `reset_local_data` resolves which dir to delete — so path resolution fell back to the pre-login `users/local` dir and deleted an empty directory.
- Pass the signed-in user id explicitly through `clearAllAppData` → `resetOpenHumanDataAndRestartCore` → `reset_local_data` → `config_get_data_paths`, pinning the deletion to `users/<id>` regardless of marker state.
- New core `get_data_paths_for_user(user_id)` resolves the target dir deterministically; controller reads an optional `user_id` param.
- Debug logging added on every changed hop; no UI-text/i18n changes (existing copy reused).

## Problem

Clicking **Settings → Clear App Data** appeared to do nothing: after confirming, memory, sources, conversations, and history were all still present.

Tracing the flow shows the button is wired end-to-end. The defect is an ordering/path-resolution bug:

1. `clearAllAppData` runs `clearSession()` (step 2) before `resetOpenHumanDataAndRestartCore()` (step 3).
2. `clearSession()` → `auth_clear_session` → `credentials/ops.rs` `clear_active_user()` **removes `~/.openhuman/active_user.toml`**.
3. `reset_local_data` then calls `config_get_data_paths`, which resolves the "current" dir from the now-absent marker (`config/schema/load/dirs.rs:327` → `None` → pre-login `users/local` at `dirs.rs:348`).
4. The reset deletes an **empty `users/local`** dir; the real `users/<id>` slice (memory / sources / conversations / cron / threads) survives.

This exactly matches the report ("Brain → Memory → Sources — all data still present").

## Solution

Make the deletion target deterministic from the signed-in user id rather than the mutable marker. This preserves the deliberate `clearSession` → `reset` ordering (which avoids core-restart token staleness) and removes the fragile coupling to marker state.

- **Core:** `get_data_paths_for_user(user_id)` (`config/ops/loader.rs`) resolves `~/.openhuman/users/<id>` via `user_openhuman_dir`. Controller `get_data_paths` reads an optional `user_id` param (`schemas/controllers.rs`, `schema_defs.rs`); absent/blank keeps marker-based resolution.
- **Tauri:** `reset_local_data(user_id)` forwards the id via `config_get_data_paths`; the param is omitted entirely when absent, so the pre-login recovery path is byte-identical to before.
- **Frontend:** `resetOpenHumanDataAndRestartCore(userId)` invokes with `{ userId }`; `clearAllAppData` forwards the id it already has (`snapshot.auth.userId`). The Welcome-screen decryption-recovery caller passes no id and falls back to marker resolution.

**Known follow-ups (not in this PR):** OS-keyring credentials/API keys and the spaCy/python cache (#4688) live outside `users/<id>` and still survive a clear; the confirmation-dialog "call out credentials separately" ask is also unaddressed. This PR fixes the core "nothing is cleared" defect.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — new Rust tests cover `get_data_paths_for_user` and the controller `user_id` path; frontend tests cover id forwarding (incl. `null` recovery). CI is authority.
- [x] Coverage matrix updated — `N/A: behaviour-only bug fix (no new feature row)`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix feature row`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no new release-cut surface; existing Clear App Data flow`

## Impact

- **Platform:** desktop (Tauri). `reset_local_data` command signature gains an optional `user_id`; frontend and core updated in lockstep. Headless/CLI `config_reset_local_data` path unchanged.
- **Compatibility:** pre-login recovery (no id) preserves exact prior behavior (empty `params: {}`). No migration.
- **Security/privacy:** makes "Clear App Data" actually delete the signed-in user's local data as intended.

## Related

- Closes: #4950
- Follow-up PR(s)/TODOs: OS-keyring/credential wipe + spaCy venv cleanup (#4688) + confirmation-dialog credential warning.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/GH-4950-clear-app-data
- Commit SHA: 729d9c3100eb2788246356d95e52e469d3d2b2f0

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — prettier clean; rust fmt aligned to CI's rustfmt output (verified via CI)
- [x] Focused tests: added (Rust loader/controller + frontend clearAllAppData/tauriCommands); not run locally (machine disk limits), CI is authority
- [x] Rust fmt/check (if changed): formatting aligned to CI rustfmt (fmt-only follow-up pushed); compile/clippy verified via CI
- [x] Tauri fmt/check (if changed): `app/src-tauri/src/local_data_reset.rs` fmt-clean; verified via CI

### Validation Blocked

- `command:` cargo build/check/test, pnpm test/build
- `error:` blocked by local machine disk constraints (fills disk)
- `impact:` compile + test verification deferred to CI

### Behavior Changes

- Intended behavior change: Clear App Data now deletes the signed-in user's `users/<id>` data dir instead of the empty pre-login dir.
- User-visible effect: after confirming, memory/sources/conversations/history are actually cleared and the app restarts to a clean state.

### Parity Contract

- Legacy behavior preserved: pre-login recovery (no user id) still uses marker-based resolution with an empty `params: {}`.
- Guard/fallback/dispatch parity checks: absent/blank `user_id` falls back to `get_data_paths()`; agent tool `config_get_data_paths` unchanged (arg-less).

- Duplicate PR(s): N/A
- Canonical PR: N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - “Clear App Data” now clears the correct user-scoped OpenHuman data when a user identifier is available, even if session markers are removed.
  - If no user identifier is available (pre-login recovery), it continues using the legacy fallback reset behavior.
  - Improved data-path resolution to reliably target the intended `users/<id>` directory and avoid incorrect/no-op resets.

- **Tests**
  - Updated and added test coverage for user-scoped reset calls and for validating rejection of unsafe user identifiers (including Windows path handling).


M3gA-Mind:
Maintainer review changes (pushed to this PR branch):

**1. `src/openhuman/config/ops/loader.rs` + `schemas/controllers.rs` — rustfmt alignment.** Two test-helper spots were hand-formatted the wrong way and failed `cargo fmt --check` (Rust Quality + Frontend Checks' `rust:format:check`): `assert_ne!(...)` in `get_data_paths_for_user_scopes_current_dir_to_explicit_id` and `params.insert(...)` in `handle_get_data_paths_scopes_to_explicit_user_id` are now expanded to the multi-line form rustfmt expects. Prettier itself was already clean.

**2. `src/openhuman/config/ops/loader.rs` — Codex P1 security hardening (addresses the inline review).** Codex correctly flagged that `get_data_paths_for_user` joined caller-controlled `user_id` onto the data root, and the returned `current_openhuman_dir` flows into `remove_dir_all` inside the Tauri `reset_local_data` command. An absolute id (`/etc`) or one with `..`/separators could steer the delete target outside `<root>/users/<id>` — a real risk since the reset command's renderer runs untrusted CEF webview content. Fixed by:
- `is_plain_user_id()` — reject empty, `.`, `..`, and any `/` `\\` NUL before the join.
- Defense-in-depth: verify the resolved dir is a direct child of `<root>/users` (catches Windows drive-relative quirks) before returning.
- Added `get_data_paths_for_user_rejects_unsafe_ids` regression test covering `..`, `.`, `../escape`, `/etc`, `a/b`, `a\\b`, `""`.

Invalid ids now return an `Err` (surfaced to the UI) instead of resolving a deletable path. No behavior change for legitimate backend user ids. Letting CI verify compile/clippy/tests.

M3gA-Mind:
## Follow-up: the real first-order root cause was the Tauri ACL — my original diagnosis was only half the story

Manual testing surfaced a runtime failure I missed:
```
[core] resetOpenHumanDataAndRestartCore: reset_local_data failed
       reset_local_data not allowed. Command not found
```

### Root cause (corrected)
There are **two** causes, and I originally reported only the second:

1. **PRIMARY — Tauri v2 ACL denial (the actual "button does nothing").** `reset_local_data` is registered in `generate_handler!` (`app/src-tauri/src/lib.rs:3842`) and defined in Rust, **but it was never in the ACL allow-list** (`app/src-tauri/permissions/allow-core-process.toml`). Tauri rejected the invoke *before it reached Rust*. I verified across full git history that the command was **never** permitted, and that my PR did **not** touch the ACL — so this is pre-existing. It regressed when the dedicated `reset_local_data` command replaced the old `callCoreRpc('config_reset_local_data')` flow (that old path reached Rust via `relay_http_rpc`, which *is* ACL-permitted); the new command's ACL entry was simply missed.

2. **SECONDARY — marker-ordering path resolution (my original diagnosis).** `clearSession()` removes `active_user.toml` before the reset resolves paths, so a marker-based resolution would target the empty pre-login `users/local` dir. This is a **real** bug and my `user_id` threading fixes it — but it only *executes* once the command is actually reachable. With the ACL blocking the invoke, this code never ran.

**Honest assessment: my first diagnosis was partial.** I fixed a genuine latent bug but missed the first-order cause. Worse, the half-fix made the observed behavior slightly worse: `clearSession()` still destroyed `active_user.toml`, then the ACL blocked the reset, leaving the user signed out with all data intact.

### Fix
- **ACL:** added `"reset_local_data"` to `allow-core-process.toml`, next to its siblings `restart_app` / `restart_core_process` / `schedule_cef_profile_purge`. One command, no wildcard — no posture change. Verified post-build that the compiled `gen/schemas/acl-manifests.json` now grants it under `allow-core-process/commands/allow` (the permission the main+overlay windows carry).
- **Kept** the `user_id` threading — still needed so the now-reachable reset deletes `users/<id>` rather than the pre-login dir.

### CodeRabbit (PII-in-logs) — fixed
`get_data_paths_for_user` (`src/openhuman/config/ops/loader.rs`) logged the raw `user_id` and user-scoped paths. Replaced with a content-free `explicit_user_id=true` in both the `log::debug!` and the `RpcOutcome` status text; the returned JSON paths are unchanged.

### Verification
`pnpm typecheck` passes; rebuilt + relaunched the desktop app on this branch — it boots with **no ACL denial**, and the compiled ACL now contains the command. Final button-click verification is the maintainer's manual test.

Issue #4924, opened 2026-07-15
[Tiny Place] Feed: posts don't support images/media (no media field in contract)


M3gA-Mind:
Part of #4776 · Tiny Place epic #4190 · section §2 Feed.

**Audited vs `upstream/main` 3cca4f3f3 (code-level, read-only).**

## Problem
Posts do not support images/media end-to-end. The audit item "Images / media in posts" fails at the contract level, not just the UI.

## Evidence
- SDK type `GqlPost` (`vendor/tinyplace/sdk/rust/src/types/graphql.rs:26-46`) carries only `body`, `content_type`, counts, `author`, etc. — **no image/media/attachment field**.
- UI renders the post body as plain text: `app/src/agentworld/pages/FeedSection.tsx:462` — `<p ...>{post.body}</p>`. The only `<img>` in a post is the author avatar (`:410`, `post.author.avatarUrl`).
- `createPost` sends `{ body, contentType }` only (`FeedSection.tsx:256` → `manifest.rs:4659` → SDK `feeds::create_post`); there is no upload/attachment path.

## Expected vs actual
- **Expected:** posts can include images/media and render them in the feed.
- **Actual:** posts are text-only; no media field exists in the SDK types, handler, or UI.

## Notes
Likely a not-yet-built feature (may relate to design-port #3853, which does not cover the data contract). No code change proposed in this audit pass; flagging the contract-level gap so it can be scoped (SDK type + handler + composer upload + renderer).

M3gA-Mind:
**Blocked on an upstream `tiny.place` contract change — cannot be fixed in this repo.** Investigated; recording the findings so this is not re-attempted.

**What exists today:**
- `vendor/tinyplace` is a **git submodule** (gitlink `160000`, pinned at `39ab493f` of `tinyhumansai/tiny.place`; `Cargo.toml:509` path-deps the Rust crate into it). An openhuman PR can only bump the submodule **pointer** to a commit that already exists upstream — it cannot edit the SDK's `PostCreate` / `Post` / `GqlPost` structs.
- Media inputs (`image` / `gifUrl`) exist **only on the TypeScript write side** (`feeds.ts`, `validation.ts`). The **Rust** SDK — the one openhuman actually calls — never received them.
[...]

Issue #4776, opened 2026-07-10
Audit: Tiny Place end-to-end integration — World, Feed, Messages, Ledger, Bounties, Directory, Identities, Profiles


Al629176:
## Summary

Full end-to-end audit of every Tiny Place surface in OpenHuman. Each section below lists every user-facing feature with a pass/fail/untested checkbox. Goal: identify every broken, missing, or unintegrated path so they can be fixed in one coordinated push. Frontend + backend integration must be verified for each item — not just "UI renders" but "data round-trips correctly."

This audit supersedes and extends the existing epic #4190 (which covers known bugs in World/Feed/Identities/Trading/Wallet/Bounties). The sections added here — **Messages, Ledger, Directory, Profiles** — have had no prior systematic audit.

---

## How to use this issue

For each item, mark:
- `[x]` — verified working end-to-end
- `[~]` — partially working / known gap
- `[ ]` — broken, missing, or untested

Leave a comment with findings per section. Open a sub-issue for each confirmed broken item that doesn't already have one.

**Test environment:** staging build, signed-in user with a registered handle, connected Solana wallet.

---

## 1. World

| Feature | Status | Notes |
|---------|--------|-------|
| World renderer loads without error | | Known broken: #4193 / PR #4213 open |
| Room navigation (Poker, Court, Office, Home, World) | | |
| Avatars / other agents visible in world | | |
| Agent enters / exits room | | |
| World state syncs with tiny.place relay | | |
| Works on macOS, Windows, Linux | | |

- [ ] World renderer loads on first visit (no "Could not start the world renderer") — #4193
- [ ] All rooms navigable
- [ ] Other users/agents visible
- [ ] Real-time world state sync via relay
- [ ] No crash on resize or tab switch

---

## 2. Feed

| Feature | Status | Notes |
|---------|--------|-------|
| View feed posts (own + followed) | | |
| Compose and publish a post | | Own posts fixed: #4072 merged |
| Delete post (in-app confirmation) | | `window.confirm` bug: #4197 / PR #4214 open |
| Like / react to a post | | |
| Reply / thread | | |
| Feed sorting (recent, popular) | | Fixed: #3829 merged |
| Display name resolution (handle → name) | | |
| Images / media in posts | | |
| Infinite scroll / pagination | | |
| Feed updates in real-time (push) | | |

- [ ] Compose → publish → appears in feed
- [ ] Delete uses in-app confirmation modal — #4197
[...]

PR #4990, opened 2026-07-16
fix(tinyplace): open the profile when a Directory entry is clicked (Closes #4927)

M3gA-Mind:
## Summary

- Clicking a Tiny Place **Directory** entry now opens that agent's profile, fixing the confirmed dead-path defect where the card click only toggled a local highlight and did nothing else.
- Card click (and Enter/Space) opens a new `AgentProfileModal` that seeds instantly from the `AgentCard` already in hand, then enriches with the full public profile via `apiClient.graphql.user(agentId)`.
- Graceful degrade: if the richer lookup fails or 402s, the modal keeps the card data and shows a soft notice instead of blanking.
- Shared display helpers extracted to `directoryHelpers.ts` so the modal reuses them without importing the page (no import cycle).
- Five `agentWorld.directory.profile.*` `useT()` keys added to `en.ts` + all 13 locales (no em dashes); grep-friendly `agentworld:directory` logging, no PII.

## Problem

The Directory (`app/src/agentworld/pages/DirectorySection.tsx`) was browse-only. Each card is `role="button"` and visually clickable, but the handler only flipped a local `selected` flag used **solely** for a border/ring CSS class — nothing rendered or navigated to the agent's profile:

```
onClick={() => setSelected(s => !s)}   // old DirectorySection.tsx:241
```

This is the §6 Directory audit's confirmed defect (part of #4776). The profile round-trip already exists and needs no backend change: `apiClient.graphql.user(cryptoId)` → `openhuman.tinyplace_graphql_user` returns a full `GqlProfile` (`invokeApiClient.ts:2280`), and `AgentCard.agentId` is that cryptoId.

## Solution

- **`DirectorySection.tsx`**: replaced the `selected`-ring toggle with an `onOpen` callback; the grid holds an `openAgent` state and renders `<AgentProfileModal>` (portaled) when set. Follow's `stopPropagation` still shields the Follow button from opening the modal.
- **`components/AgentProfileModal.tsx`** (new): built on the shared `ModalShell`. Title = handle, subtitle = actor type; shows verified badge, joined date, bio, and skills/tags. Fetches `graphql.user(agent.agentId)` on mount; on `null`/error/`PaymentRequiredError` it falls back to the card's `description`/skills and surfaces `agent-profile-load-notice`.
- **`pages/directoryHelpers.ts`** (new): `getAvatarColor`/`getDisplayName`/`getHandle`/`getInitials`/`getSkills` moved out of the page so both the grid and the modal import them (avoids a page↔modal cycle).
[...]

PR #4996, opened 2026-07-16
feat(tinyplace): add own-profile editing (name/bio/avatar) to Agent World (Closes #4930)

### LINKED ISSUE #4930: [Tiny Place] Agent World profile editing (name/bio/avatar) is unreachable — updateProfile RPC has no UI
Part of #4776 (§8 Profiles). Found during the code-level audit vs `upstream/main` `3cca4f3f3`.

## Expected
From the Agent World Profiles page a user can edit and save name / bio / avatar, per the §8 rows "Edit name / bio / avatar" and checkboxes "Own profile loads and is editable" / "Name, bio, avatar save and display correctly".

## Actual — save path fully plumbed but no UI caller
`app/src/agentworld/pages/ProfilesSection.tsx` (495 L) is **display-only**: name/bio/avatar are read (`:206` displayName, `:311` bio, `:276` avatarUrl) with **no edit form**.
The save round-trip exists at every layer but is never invoked:
- apiClient `users.updateProfile(cryptoId, update)` — `app/src/lib/agentworld/invokeApiClient.ts:1746` → `openhuman.tinyplace_users_update_profile`
- handler `handle_tinyplace_users_update_profile` — `src/openhuman/tinyplace/manifest.rs:400` (registered `schemas.rs:171`)
- SDK `update_profile()` — `vendor/tinyplace/sdk/rust/src/api/users.rs:35`
- covered by a unit test — `invokeApiClient.test.ts:472`

`grep -r "updateProfile" app/src` returns **only** `invokeApiClient.ts` + its `.test.ts` — no production component calls it.

## Proposed fix (not implemented in this audit pass)
Add an edit form to `ProfilesSection` (own profile only) wired to `apiClient.users.updateProfile(cryptoId, { displayName, bio, avatarUrl })`, then refetch to display the saved values.

M3gA-Mind:
## Summary

- Agent World Profiles was display-only; this adds an **Edit profile** form (own profile only) for name / bio / avatar, wired to the already-plumbed `apiClient.users.updateProfile` write path.
- Writable fields prefill from the authoritative `users.get` record — including `avatarEmail`, which the display-side `GqlProfile` does not carry (the shown avatar URL is derived server-side from the Gravatar email).
- On save the profile refetches through the existing reload path so the saved name/bio/avatar render.
- New `agentWorld.profile.*` i18n keys added to `en.ts` and all 13 locale files (no em dashes). No SDK/contract change — `vendor/tinyplace` is untouched.

## Problem

Per the §8 Profiles audit (#4776), `app/src/agentworld/pages/ProfilesSection.tsx` read name/bio/avatar but had **no edit form** — `grep -r "updateProfile" app/src` returned only `invokeApiClient.ts` + its `.test.ts`, never a production caller. The save round-trip existed at every layer and was simply never invoked:
- TS `apiClient.users.updateProfile(cryptoId, update)` (`app/src/lib/agentworld/invokeApiClient.ts:1746`)
- RPC `handle_tinyplace_users_update_profile` (`src/openhuman/tinyplace/manifest.rs:400`, registered `schemas.rs:171`)
- SDK `UsersApi::update_profile()` (`vendor/tinyplace/sdk/rust/src/api/users.rs:35`), writing `UserProfileUpdate { display_name, bio, avatar_email, link, tags, … }` (`types/user.rs:148`)

Read-back is fully available: the `User` type (`types/user.rs:16`) carries the same fields, `update_profile` returns the updated `User`, and the display path (`apiClient.graphql.user` → `GqlProfile`) reflects saved values on refetch. So — unlike the feed-media case (#4924) — the contract is complete and this is a UI-only gap.

## Solution

- Added `ProfileEditForm` in `ProfilesSection.tsx`, shown only for the graphql profile source (which carries an editable profile + `cryptoId`; the section always renders the wallet's **own** profile, so editing is inherently own-only).
- Fields: display name, bio, and **avatar email** (Gravatar — the actual writable avatar contract; the UI labels it as such rather than faking an image-URL field). Prefilled from `users.get` so `avatarEmail` (absent from `GqlProfile`) is authoritative.
- Save calls `apiClient.users.updateProfile(cryptoId, { displayName, bio, avatarEmail? })` (avatarEmail sent only when non-empty so an untouched field never blanks an existing avatar), then triggers the existing reload so the saved values display. Errors keep the form open with a translated retry message.
- Debug logging under the `agentworld:profile` namespace; values are never logged (no PII).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — three new cases (edit button present; save calls updateProfile + refetch shows new value; save-failure keeps form + error).
- [x] **Diff coverage ≥ 80%** — the new form/component is exercised by the added tests. `pnpm test:coverage` not run locally (machine constraint); CI enforces the gate.
- [x] Coverage matrix updated — `N/A: fills an existing feature gap, no new matrix row.`
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — uses the existing `users.updateProfile`/`users.get` RPCs; tests mock the client.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: does not touch a release-cut surface.`

## Impact

- Desktop UI only (Agent World → Profiles). No Rust/core, schema, SDK, or network changes; `vendor/tinyplace` submodule pointer unchanged.
- No migration. Security: the write is authorized cryptographically, not by trusting client input — the SDK signs the canonical `user.profile` payload (which includes `cryptoId`, `vendor/tinyplace/sdk/rust/src/api/users.rs:124`) with the caller's wallet signer and presents the signing public key (`http.rs:412-428`), so a caller can only update the cryptoId whose key they hold; the core uses the local wallet as signer and `/rpc` is token-gated to the app. The RPC handler passing `cryptoId` through unchecked is fine on that basis; backend-side enforcement (signer==cryptoId owner) is tracked for confirmation in #5024. Avatar is Gravatar-email-based (the only write contract the SDK exposes), surfaced honestly in the UI.

## Related

- Closes: #4930
- Part of #4776 (§8 Profiles).
- Follow-up PR(s)/TODOs: none. (A future image-upload avatar would require an SDK/backend contract addition, out of scope here.)

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/GH-4930-tinyplace-profile-editing`
- Commit SHA: `438126318`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — ran targeted `prettier --write` on all touched files (clean).
- [x] `pnpm typecheck` — passed clean (exit 0).
- [x] Focused tests: three new cases in `app/src/agentworld/pages/ProfilesSection.test.tsx` (not run locally; CI verifies).
- [x] Rust fmt/check (if changed): N/A — no Rust changes.
- [x] Tauri fmt/check (if changed): N/A — no Tauri changes.

### Validation Blocked

- `command:` `pnpm test:coverage` / `pnpm test`
- `error:` intentionally not run — local full test/build is disabled on this machine (fills disk).
- `impact:` CI is the authority for full test + coverage verification.

### Behavior Changes

- Intended behavior change: users can edit and save their own name / bio / avatar (Gravatar email) from the Profiles page.
- User-visible effect: an Edit profile button + form; saved values display after refetch.

### Parity Contract

- Legacy behavior preserved: the display card, wallet-locked / no-handle / payment-required / directory-fallback states are unchanged; editing only appears for the own graphql profile.
- Guard/fallback/dispatch parity checks: `avatarEmail` is omitted from the update when empty so an existing avatar is never accidentally blanked.

- Duplicate PR(s): none.
- Canonical PR: this.
- Resolution: N/A.

## Summary by CodeRabbit

* **New Features**
  * Added profile editing for users’ own Agent World profiles.
  * Users can update their display name, bio, and avatar email.
  * Added validation, save progress feedback, and error handling.
  * Profile data refreshes after successful updates.

* **Localization**
  * Added translated profile-editing text across supported languages.

* **Tests**
  * Added coverage for editing, validation, delayed data loading, refresh behavior, and save failures.


M3gA-Mind:
Synced onto current `upstream/main` (clean merge). The Frontend Checks failure was **not** in this PR's code: the only failing test was `src/components/intelligence/__tests__/ObsidianVaultSection.test.tsx:120` — an Escape-closes-guide `waitFor` timeout in a component this PR does not touch (this PR only changes `ProfilesSection` + i18n). 379/380 test files passed; ProfilesSection's own tests were green. That is a flaky test unrelated to this change, so this is a re-run on top of current main.

`pnpm typecheck` + `pnpm i18n:check` clean (13 locales, 0 missing/extra).

@coderabbitai review

Re-run on current main; prior failure was an unrelated ObsidianVaultSection flake.

app/src/agentworld/pages/ProfilesSection.tsx
```diff
@@ -168,7 +172,151 @@ function useMyIdentity(reloadKey: number): ProfileState {
...
+  // Seed the writable fields (notably avatarEmail, absent from GqlProfile) from
+  // the authoritative user record. Non-fatal: on failure we keep the values
+  // already seeded from the displayed profile.
+  useEffect(() => {
+    let cancelled = false;
+    void apiClient.users
+      .get(cryptoId)
+      .then(user => {
+        if (cancelled) return;
+        if (typeof user?.displayName === 'string') setDisplayName(user.displayName);
```
graycyrus: **[major]** This unconditionally overwrites `displayName`/`bio`/`avatarEmail` with the `users.get()` result once it resolves - including if the user already started typing in the form while the fetch was in flight. Their edits get silently discarded with no indication anything happened.

Suggestion: only apply the prefetch result if the field hasn't been touched yet (e.g. compare against the initial value, or track a `touched` flag per field), or disable the inputs until this resolves so there's no window for the race.

graycyrus (changes requested):
Good, well-scoped fix for #4930 - closes a real gap where the write path was fully plumbed (apiClient -> RPC -> SDK) but never called from any component. i18n coverage across all 13 locales is complete and consistent, tests cover the happy path, the edit-button gate, and the save-failure path, and the avatarEmail-omitted-when-empty logic is a thoughtful touch (avoids silently blanking an existing avatar).

One real bug and one thing I want confirmed before this ships:

**[major] Prefill race can silently discard what the user just typed.** In `ProfileEditForm`, `displayName`/`bio`/`avatarEmail` are seeded from the passed-in props, then a `useEffect` kicks off `apiClient.users.get(cryptoId)` and unconditionally overwrites those same state vars when it resolves. If the user opens the form and starts typing before that fetch lands (slow network, anything realistic), their edits get clobbered the moment the fetch finishes - no warning, just silently reverted. The tests only pass because they explicitly `waitFor` the prefill to land before typing, so this doesn't get caught. Fix: track whether the field has been touched (e.g. a `touched` ref/flag per field, or just don't overwrite if the current value already differs from `initialDisplayName`/`initialBio`), or disable the form until the prefetch resolves so there's no window for it.

**[major] Worth confirming before merge: `handle_tinyplace_users_update_profile` (manifest.rs) takes a client-supplied `cryptoId` with no check that it matches the caller's own wallet/signer identity** - unlike the marketplace handlers in the same file, which derive identity from `signer.agent_id()` rather than trusting client input. That's pre-existing backend code this PR doesn't touch, but per the linked issue this write path was previously dead - this PR is what makes it reachable from a real UI for the first time. If nothing elsewhere in the stack (auth middleware, RPC bridge trust boundary) enforces that the cryptoId being updated belongs to the caller, this is an IDOR: any caller of this RPC could update any user's profile by passing their cryptoId. Might be a non-issue if the RPC bridge is only reachable by the app's own trusted bundle, but given this PR is the one activating the path in production, I'd want that confirmed rather than assumed. Not blocking on the frontend code itself, but flagging because "no security implications" in the PR description isn't quite right if this holds.

Minor: no loading state while the `users.get` prefetch is in flight - the form briefly shows the display-card values before jumping to the authoritative ones. Not a big deal, but a subtle flash worth a follow-up.

Requesting changes for the prefill race - that's a real data-loss bug in new code. Please also confirm the update_profile authorization question above.

M3gA-Mind:
Thanks @graycyrus — both points addressed.

**[major] Prefill race — FIXED.** `ProfileEditForm` now tracks a per-field `touched` ref; the `users.get` prefill only applies to fields the user hasn't edited. If the fetch resolves after the user starts typing, their input is preserved instead of clobbered. Added a regression test that defers `users.get`, types into the display-name field before it resolves, then resolves it with a different record and asserts the typed value survives (fails before this change).

**[major] update_profile authorization — confirmed NOT an IDOR.** I traced the write path: the RPC handler (`manifest.rs:400`) does pass `cryptoId` through unchecked, but authorization is cryptographic at the backend, not trust-based at the RPC layer. The SDK's `users.update_profile` (`vendor/tinyplace/sdk/rust/src/api/users.rs:35-45`) signs the canonical `user.profile` payload — which is built from the target `crypto_id` via `user_profile_signature_payload(crypto_id, &update)` — with the caller's wallet signer and presents the signing key "so the backend can authorize the update." Since `crypto_id` **is** the wallet's Solana address, a caller can only produce a valid signature for the cryptoId whose private key they hold (their own). Passing someone else's cryptoId would sign with the caller's own key, and the backend rejects the mismatch. The local `/rpc` bridge is also token-gated to the app and uses the user's own wallet as signer, so there's no cross-user update path. So the description's "no security implications" holds for this frontend change.

**Re: `Closes` vs follow-ups** — N/A here; noting for the sibling directory PR #4990.

`pnpm typecheck` clean.

@coderabbitai review

oxoxDev (changes requested):
Clean UI-only fill of an already-plumbed write path; i18n parity complete (9 keys × 14 locales) and all checks green — BLOCKED is the required review, not a red check.

Blocking:
- Prefill overwrites in-progress edits (`ProfilesSection.tsx` ~197-210): the `users.get` prefill effect unconditionally calls `setDisplayName/setBio/setAvatarEmail` when it resolves, so if the fetch is slow and the user types first, their edits are silently discarded. Guard with a dirty/touched flag, or seed only the fields not yet edited. (= graycyrus / Codex @212/214.)

Non-blocking:
- Edit is hidden on the directory-fallback path (`isGraphql && cryptoId` @465) — an owner resolved via `directory.reverse` can't edit. Confirm the fallback truly lacks a cryptoId. (= Codex @465.)
- `avatarEmail` is omitted-when-empty, so it can never be *cleared* once set, while `displayName`/`bio` always send trimmed — clearing the name blanks it. Asymmetric; consider a min-length guard on name.

M3gA-Mind:
Thanks @oxoxDev. As with the sibling PRs, the review landed on a stale commit — the blocking item was already fixed before your pass.

**Blocking — prefill overwrites in-progress edits: ALREADY FIXED** (commit `f53b99ed0`, pushed 15:01, before this review). `ProfileEditForm` now tracks a per-field `touched` ref (`ProfilesSection.tsx:206`); the `users.get` prefill only applies to untouched fields (`:216-221`), and each `onChange` marks its field touched (`:275/291/307`). Regression test types before a deferred prefill resolves and asserts the edit survives.

**Non-blocking — name can be blanked (asymmetric clearing): FIXED** (this push). A display name is now required: `handleSave` guards against a blank name and the Save button is disabled while it's empty, with a translated `agentWorld.profile.nameRequired` message (added across all locales). `bio` may still be empty (optional); `avatarEmail` stays omitted-when-empty so an existing avatar isn't accidentally cleared. Regression test added.

**Non-blocking — edit hidden on the directory-fallback path:** confirmed the fallback does **not** lack a cryptoId — `OwnedIdentity.cryptoId` is populated from `directory.reverse`. The edit form is deliberately gated on the graphql profile source because that path is only reached when `graphql.user` returns null (i.e. no published profile yet — the UI prompts to register a handle first). Enabling edit there would mean creating a profile from the no-published-profile state, which is out of scope for this fill-the-gap PR; happy to file a follow-up if surfacing edit pre-publish is desired.

`pnpm typecheck` + `pnpm i18n:check` clean (13 locales, 0 missing/extra). Please re-review the current head.

@coderabbitai review

app/src/agentworld/pages/ProfilesSection.tsx
M3gA-Mind: Fixed in `f53b99ed0`. `ProfileEditForm` now tracks a per-field `touched` ref (`ProfilesSection.tsx:206`); the `users.get` prefill only writes fields the user hasn't edited (`:216-221`), and each `onChange` marks its field touched (`:283/299/315`). Regression test `a late prefill fetch does not clobber what the user already typed` defers `users.get`, types into the display-name field before it resolves, then resolves with a different record and asserts the typed value survives (would fail before this change). Also (`3d917bb3f`) the display name is now required so it can't be blanked on save.

M3gA-Mind:
@graycyrus @oxoxDev — resolved.

**[major] Prefill race (data loss) → FIXED** (`f53b99ed0`, before your reviews). Per-field `touched` ref; `users.get` prefill applies only to untouched fields; each `onChange` marks touched. New regression test types **before** a deferred prefill resolves and asserts the edit survives (the existing tests `waitFor` prefill first, so they missed it — this one doesn't).

**update_profile authorization (possible IDOR) → confirmed client-side + filed backend tracker #5024.** I did not assume; I traced it:
- The SDK signs the canonical `user.profile` payload, which **includes `cryptoId`** (`vendor/tinyplace/sdk/rust/src/api/users.rs:124`), with the caller's wallet signer, and presents the signer's **public key** (`http.rs:412-428`). Since `cryptoId` is the wallet's Solana address, a caller can only produce a valid signature for the cryptoId whose key they hold — signing for someone else's cryptoId presents a non-matching public key.
- The core uses the local wallet as signer and `/rpc` is token-gated to the app; the UI only passes the user's own cryptoId. So there's **no cross-user update path reachable from the client.**
- The one piece I can't verify from this repo is the tiny.place **backend** rejecting a signature whose key ≠ the target cryptoId owner (separate repo). Rather than assume it, I filed #5024 to confirm that defense-in-depth, and updated the PR body's security line to describe the mechanism instead of a bald "no security implications."

**Bonus — name can be blanked (oxoxDev, asymmetric clearing) → FIXED** (`3d917bb3f`): display name required (guarded save + disabled button + `agentWorld.profile.nameRequired` in all locales + test).

`pnpm typecheck` + `pnpm i18n:check` clean (13 locales, 0 missing/extra). Please re-review.

PR #5040, opened 2026-07-18
perf(orchestration): batch the per-session unread_count into one query

mysma-9403:
## Summary

- Collapse an N+1 in the orchestration session roster and attention queue: `unread_count` ran **2 queries per session** and was called once per session in two hot RPC paths.
- Add a batched `unread_counts(conn) -> HashMap<session_id, i64>` (one `GROUP BY` with a `LEFT JOIN` on the read-cursor kv) and consume it in `sessions_list` and `gather_unread_signals`.
- Byte-identical results; add a parity test asserting the batched map equals the scalar `unread_count` across every cursor/visibility state.

## Problem

`store::unread_count(conn, session_id)` does two round-trips — `kv_get(read_cursor_key(..))` then a `COUNT(*)` — and is called once per session in:

- `handle_sessions_list` (`schemas.rs`) — the session roster, `for session in rows { unread_count(..) }`, and `list_sessions` has **no LIMIT**, so listing N sessions is 2N queries.
- `gather_unread_signals` (`ops.rs`) — the attention queue, same per-session loop.

This is especially visible in `handle_sessions_list` because the sibling *message* count in the very same loop was already batched into `visible_message_counts` / `visible_message_counts_by_session` (consumed via `.get().copied().unwrap_or(0)`); the unread count was simply left as the lone per-row DB hit.

## Solution

Add a batched query mirroring `visible_message_counts_by_session`:

```sql
SELECT m.session_id, COUNT(*)
  FROM messages m
  LEFT JOIN kv ON kv.k = 'read:' || m.session_id
 WHERE m.timestamp > COALESCE(kv.v, '')
   AND (m.event_kind IS NULL OR m.event_kind NOT IN ('status','lifecycle','unknown','session_info'))
 GROUP BY m.session_id
```

- `kv.k` is `PRIMARY KEY`, so the `LEFT JOIN` matches at most one cursor row per session — **no fan-out**.
- `COALESCE(kv.v, '')` reproduces the scalar path's `.unwrap_or_default()` empty cursor (a never-read session counts all its visible messages).
- The `event_kind` predicate is copied verbatim from `unread_count`.
- Sessions with zero unread drop out of the `GROUP BY`; both call sites already default missing entries to `0`, so behaviour is identical.

Each call site now computes the map once before its loop and reads `map.get(&session_id).copied().unwrap_or(0)`. The scalar `unread_count` is retained (still used by existing store tests and as the parity oracle).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `unread_counts_batches_and_matches_scalar_unread_count` asserts the batched map equals the scalar `unread_count` for every session across five states: no cursor, mid-stream cursor (partial), fully read (zero), a hidden excluded-`event_kind` row, and a session with no messages.
- [x] **Diff coverage ≥ 80%** — the new `unread_counts` query and both swapped call sites are exercised by the parity test plus existing `sessions_list` / attention coverage. `cargo test -p openhuman --lib orchestration::store` passes.
- [x] Coverage matrix updated — `N/A: internal query batching, no user-visible feature change`.
- [x] Linked issue closed via `Closes #NNN` — no existing issue; found by inspection.

## Impact

- **Desktop/CLI**: `openhuman.sessions_list` and the attention queue now issue one unread query instead of 2N. No behaviour change (verified byte-identical by the parity test); no schema/migration.
- Extends the existing `visible_message_counts` batching pattern in the same loop.

## Related

- Closes:
- Follow-up PR(s)/TODOs:

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `perf/orchestration-batch-unread-counts`
- Commit SHA: `45c2fae86`

### Validation Run

- [x] `pnpm typecheck` — N/A (no frontend change)
- [x] Focused tests: `cargo test -p openhuman --lib orchestration::store`
- [x] Rust fmt/check (if changed): `cargo fmt`
- [x] Tauri fmt/check (if changed): N/A (core-only change)

### Behavior Changes

- Intended behavior change: none — pure query batching.
- User-visible effect: none (fewer DB round-trips on session listing / attention).

### Parity Contract

- Legacy behavior preserved: `unread_counts` returns exactly what `unread_count` would per session; `COALESCE(kv.v,'')` = `.unwrap_or_default()`, `event_kind` predicate copied verbatim, zero-unread sessions default to 0 at both call sites.
- Guard/fallback/dispatch parity checks: parity test asserts map == scalar across all cursor/visibility states.

### DIFF
--- a/src/openhuman/orchestration/ops.rs
+++ b/src/openhuman/orchestration/ops.rs
@@ -252,11 +252,12 @@ pub(super) fn gather_unread_signals(
     conn: &rusqlite::Connection,
 ) -> anyhow::Result<Vec<super::attention::UnreadSignal>> {
     let mut out: Vec<super::attention::UnreadSignal> = Vec::new();
+    let unread_counts = store::unread_counts(conn)?;
     for session in store::list_sessions(conn)? {
         if matches!(session.session_id.as_str(), "master" | "subconscious") {
             continue;
         }
-        let unread = store::unread_count(conn, &session.session_id)?;
+        let unread = unread_counts.get(&session.session_id).copied().unwrap_or(0);
         if unread > 0 {
             out.push(super::attention::UnreadSignal {
                 session_id: session.session_id,
--- a/src/openhuman/orchestration/schemas.rs
+++ b/src/openhuman/orchestration/schemas.rs
@@ -416,11 +416,12 @@ fn handle_sessions_list(_params: Map<String, Value>) -> ControllerFuture {
             let rows = store::list_sessions(conn)?;
             let visible_counts = store::visible_message_counts(conn)?;
             let pinned_visible_counts = store::visible_message_counts_by_session(conn)?;
+            let unread_counts = store::unread_counts(conn)?;
             let mut out: Vec<SessionSummary> = Vec::with_capacity(rows.len() + 2);
             let mut have_master = false;
             let mut have_subconscious = false;
             for session in rows {
-                let unread = store::unread_count(conn, &session.session_id)?;
+                let unread = unread_counts.get(&session.session_id).copied().unwrap_or(0);
                 match session.session_id.as_str() {
                     "master" => have_master = true,
                     "subconscious" => have_subconscious = true,
--- a/src/openhuman/orchestration/store.rs
+++ b/src/openhuman/orchestration/store.rs
@@ -436,6 +436,33 @@ pub fn visible_message_counts_by_session(conn: &Connection) -> Result<HashMap<St
     Ok(rows)
 }
 
+/// Unread message counts keyed by `session_id`: the batched form of
+/// [`unread_count`], collapsing its per-session cursor lookup + `COUNT(*)` into a
+/// single query so a caller iterating N sessions runs one statement instead of
+/// 2N. `LEFT JOIN kv ON kv.k = 'read:' || m.session_id` reproduces
+/// `kv_get(&read_cursor_key(..))` — `kv.k` is `PRIMARY KEY`, so the join matches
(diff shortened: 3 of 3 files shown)


senamakel (commented):

**Unmet gate -- CI check failure:**

One check fails: `Rust Feature-Gate Smoke (gates off)` fails due to a pre-existing allowlist drift -- a new gated test file (`src/openhuman/composio/action_tool.rs`) was added to the tree but the CI lane's EXPECTED allowlist was not updated to include it. This is unrelated to this PR (which touches only `src/openhuman/orchestration/` files), but per the approval standard all reported checks must be passing.

**N+1 claim:** Verified correct. The new `store::unread_counts` function issues a single `GROUP BY` query with `LEFT JOIN kv ON kv.k = 'read:' || m.session_id` to batch the cursor lookup and count, replacing 2N queries (cursor lookup + COUNT per session) with 1. The `COALESCE(kv.v, '')` reproduces the scalar path's `.unwrap_or_default()`, and the `event_kind` predicate is copied verbatim. Both call sites (`handle_sessions_list` and `gather_unread_signals`) pre-compute the map and use `.get().copied().unwrap_or(0)`. The parity test `unread_counts_batches_and_matches_scalar_unread_count` asserts byte-identical results across all cursor/visibility states (no cursor, partial, fully read, hidden event_kind, empty session). The scalar `unread_count` is retained as the parity oracle.

**Next step:** The Feature-Gate Smoke allowlist needs updating to include `openhuman/composio/action_tool.rs` (a change to `.github/workflows/ci-lite.yml` or the relevant script). Once CI is all green, this PR is ready for approval.

Issue #5022, opened 2026-07-17
Gates-off build red on main: 2 ungated voice asserts + smoke lane runs no tests


oxoxDev:
## Summary

The slim/harness build (`--no-default-features --features tokenjuice-treesitter`) fails 2 tests on `main`, and the CI lane meant to guard it cannot catch them.

## The 2 red asserts

`voice` was the pathfinder feature gate (#4803) and predates the per-assert `#[cfg]` convention its five successors (web3/meet/mcp/skills/flows) follow. Two asserts in `src/core/all_tests.rs` are ungated while every sibling gate's equivalent is gated:

- `harness_excludes_gated_namespaces` — `assert!(full_ns.contains("voice"), ...)`
- `group_mapping_smoke` — `assert_eq!(group_for_namespace("voice"), Some(DomainGroup::Voice));`

Repro: `GGML_NATIVE=OFF cargo test --no-default-features --features tokenjuice-treesitter core::all::tests` → 2 failed, both these.

## Why CI never caught it

`rust-feature-gate-smoke` runs a single `cargo check`. `cargo check` never compiles test code and never runs tests, so a **runtime assert failure** in the gates-off build is structurally invisible to it. (`--all-targets` compiles test code but still does not run it — it would not catch these either.)

## Fix

- `#[cfg(feature = "voice")]` on both asserts, matching their siblings.
- Upgrade the smoke lane: `cargo check --all-targets` (catches gates-off compile drift in `tests/`) **plus** a scoped `cargo test` step over the modules that carry every `cfg(not(feature))` test site, plus a grep guard so the filter stays self-maintaining.

Full-suite execution is deferred (see the linked tracking issue — blocked on a pre-existing stack overflow).

PR #5043, opened 2026-07-18
feat(learning): surface facet provenance (cue_families, evidence_refs) in RPC

mysma-9403:
## Summary

- Surface the two provenance fields the learning store already persists — `evidence_refs` and `cue_families` — through the facet RPC serializer, which was silently dropping them at the boundary.
- Additive-only: two keys added to `facet_to_json`, so every facet-returning controller (`list_facets`, `get_facet`, `update`/`pin`/`unpin`/`forget`) gains them at once.
- Add a unit test asserting both fields round-trip.

## Problem

`facet_to_json` is the single serializer feeding all facet-returning `learning.*` controllers. `ProfileFacet` persists (and `row_to_facet` hydrates) two provenance columns:

- `evidence_refs: Vec<EvidenceRef>` (from `evidence_refs_json`) — the citations behind a facet.
- `cue_families: Option<HashMap<String, u32>>` (from `cue_families_json`) — per-cue-family evidence counts written by the stability detector.

Both derive `Serialize` (`EvidenceRef` is a `#[serde(tag = "type")]` enum), are documented as first-class `ProfileFacet` state, and exist specifically as provenance — yet the serializer emits only 10 keys and drops both, so no RPC consumer can see the evidence behind a facet.

## Solution

Add the two fields to the `json!` macro in `facet_to_json`:

```rust
"cue_families": f.cue_families,
"evidence_refs": f.evidence_refs,
```

The controller output schemas declare `facet`/`facets` as opaque `TypeSchema::Json`, so the schema contract is unchanged; the addition is purely additive metadata. No store or query changes.

> Scope note: `evidence_refs` are keyed to a facet's `(class, key)`, not to the specific winning value (the stability detector merges every candidate's evidence for a key before persisting the single value). This PR only stops the serializer from dropping already-persisted state; it does not change what the detector stores. Value-scoped provenance would be a separate, larger change.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `facet_to_json_includes_cue_families_and_evidence_refs` builds a facet with populated and empty provenance and asserts both keys serialize correctly (present-and-populated + null/empty edge).
- [x] **Diff coverage ≥ 80%** — the two added serializer lines are exercised by the new test. `cargo test -p openhuman --lib learning::schemas` passes.
- [x] Coverage matrix updated — `N/A: additive read-only field exposure, no new controller/feature row`.
- [x] Linked issue closed via `Closes #NNN` — no existing issue; found by inspection.

## Impact

- **Desktop/CLI**: `learning.list_facets` / `get_facet` (and the mutating controllers' echoed facet) now include `cue_families` and `evidence_refs`, enabling a transparency/provenance UI to cite the evidence behind a learned facet without a second round-trip. Purely additive; no behaviour change for existing consumers.

## Related

- Closes:
- Follow-up PR(s)/TODOs: value-scoped provenance (persist/return refs per winning value) — separate change to the learning storage model.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/learning-facet-provenance-fields`
- Commit SHA: `e4f6889af`

### Validation Run

- [x] `pnpm typecheck` — N/A (no frontend change)
- [x] Focused tests: `cargo test -p openhuman --lib learning::schemas`
- [x] Rust fmt/check (if changed): `cargo fmt`
- [x] Tauri fmt/check (if changed): N/A (core-only change)

### Behavior Changes

- Intended behavior change: facet RPC output gains two already-persisted provenance fields.
- User-visible effect: provenance (`evidence_refs`, `cue_families`) is now retrievable via the learning facet controllers.

### Parity Contract

- Legacy behavior preserved: all existing keys unchanged; opaque-JSON output schema contract unaffected; no store/query change.
- Guard/fallback/dispatch parity checks: `None`/empty provenance serializes to `null`/`[]` (verified by the test's empty-facet case).

### DIFF
--- a/src/openhuman/learning/README.md
+++ b/src/openhuman/learning/README.md
@@ -62,7 +62,7 @@ Namespace `learning` (wired into `src/core/all.rs`; 11 controllers). Methods:
 | `learning.rebuild_cache` | Manually trigger a `StabilityDetector` rebuild; returns added/evicted/kept/total_size. |
 | `learning.cache_stats` | Cache totals + per-state and per-class breakdown. |
 | `learning.list_facets` | List Active + Provisional facets, optional `class` filter. |
-| `learning.get_facet` | Fetch one facet by `class` + `key` suffix. |
+| `learning.get_facet` | Fetch one facet by `class` + `key` suffix. Each returned facet carries its provenance (`evidence_refs`, `cue_families`) alongside the value/state fields. |
 | `learning.update_facet` | Set a facet value and pin it (`user_state = Pinned`). |
 | `learning.pin_facet` / `learning.unpin_facet` | Toggle `user_state` Pinned ↔ Auto. |
 | `learning.forget_facet` | Mark `Dropped` + `user_state = Forgotten` (blocks re-promotion). |
--- a/src/openhuman/learning/schemas.rs
+++ b/src/openhuman/learning/schemas.rs
@@ -476,6 +476,54 @@ mod tests {
     }
 
     #[test]
+    fn facet_to_json_includes_cue_families_and_evidence_refs() {
+        use crate::openhuman::learning::candidate::EvidenceRef;
+        use crate::openhuman::memory_store::profile::{
+            FacetState, FacetType, ProfileFacet, UserState,
+        };
+        use std::collections::HashMap;
+
+        let mut cue_families = HashMap::new();
+        cue_families.insert("explicit".to_string(), 3u32);
+        cue_families.insert("structural".to_string(), 1u32);
+
+        let facet = ProfileFacet {
+            facet_id: "f1".into(),
+            facet_type: FacetType::Preference,
+            key: "style/verbosity".into(),
+            value: "terse".into(),
+            confidence: 0.8,
+            evidence_count: 4,
+            source_segment_ids: None,
+            first_seen_at: 1000.0,
+            last_seen_at: 1200.0,
+            state: FacetState::Active,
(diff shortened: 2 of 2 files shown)


mysma-9403:
### CI note — the red checks are not from this change

**Rust Core Coverage** is red on a **pre-existing flaky test**, not this PR. The failing test is `learning::startup::tests::learning_subscriber_fires_with_no_channel_configured` (the email-signature subscriber path in `startup.rs`). This PR only adds two keys to `facet_to_json` in `schemas.rs` plus a serializer unit test — it does not touch `startup.rs`, the subscriber, or the process-global `candidate::global()` buffer that test asserts on.

Verified against the PR base `d13a559b9` (i.e. **without** this change): the full `openhuman::learning` suite — including that test — passes **5/5** local runs. The failure is timing/parallelism-dependent (the test asserts on a shared, process-global candidate buffer); it flaked once under CI parallelism. So it is not a regression introduced here.

**Rust Feature-Gate Smoke (gates off)** is also unrelated: it fails identically on every fresh-`main`-based PR with `::error::Gated-test file set changed. Update the EXPECTED allowlist … (see #5022)` — a stale allowlist on `main`, not touched by this change. **PR Submission Checklist** was my miss (unchecked N/A boxes) and is now fixed in the description.

Happy to open a small follow-up to de-flake that startup test (give it its own isolated candidate buffer) if a maintainer wants it.

Issue #5049, opened 2026-07-20
Gate the desktop-automation cluster, then extract it out of the core


senamakel:
## Summary

Move the desktop-automation cluster — accessibility, screen intelligence, autocomplete, desktop companion, and the computer-control tools, about 24.5k LOC — out of `openhuman_core`. Land a `desktop-automation` Cargo gate with stub facades first, then lift the gated tree into its own crate.

## Problem / Context

None of this belongs in a headless or embedded core. It is desktop-only by nature, it is the second-largest ungated family after channels, and it keeps OS-automation dependencies in every build. It is also a meaningful share of the agent tool surface that a non-desktop host can never use.

Measured extent:

| Path | LOC |
| --- | --- |
| `src/openhuman/accessibility/` | 10,317 |
| `src/openhuman/screen_intelligence/` | 6,113 |
| `src/openhuman/autocomplete/` | 3,203 |
| `src/openhuman/tools/impl/computer/` (mouse, keyboard, ax_interact, automate, human_path) | 2,492 |
| `src/openhuman/desktop_companion/` | 2,376 |
| **Total** | **24,501** |

Dependencies shed: `uiautomation` (`accessibility/uia_interact.rs`). Also `enigo`, `arboard`, and `rdev` — but **only when `voice` is off as well**, since `voice/text_input.rs` and `voice/hotkey.rs` share them. Do not claim these three as an unconditional shed.

Fifteen `screen_intelligence_*` agent tools leave the tool surface (`tools/mod.rs:50`; name list at `tools/ops_tests.rs:2163-2177`).

Two eagerly-constructed `once_cell::sync::Lazy` statics leave with the cluster: `screen_intelligence/state.rs:72` (`ACCESSIBILITY_ENGINE`, holding a full `EngineState`) and `autocomplete/core/engine.rs:1020` (`AUTOCOMPLETE_ENGINE`).

**This is not a leaf, and the plan must account for that.** Unlike `flows` or `media`, the cluster is reached from always-on code and from other domains:

- `accessibility` is consumed directly by `autocomplete/core/{focus,overlay,terminal,types}.rs` — roughly ten symbols including `show_overlay`, `hide_overlay`, `focused_text_context_verbose`, `apply_text_to_focused_field`, `ElementBounds`, `FocusedTextContext`.
[...]

Issue #5046, opened 2026-07-20
Keep embedded OpenHuman within a 20–30 MiB RAM budget for OpenCompany


senamakel:
## Summary

Add a first-class minimal embedded build/profile for `openhuman_core` that keeps the OpenHuman portion of an OpenCompany tenant within a **20 MiB target / 30 MiB hard steady-state RAM cap** on Linux, while preserving the agent-turn behavior OpenCompany actually uses.

This is specifically for library embedding, not the Tauri/CEF desktop app and not the legacy out-of-process JSON-RPC launcher.

## Problem

OpenCompany embeds `openhuman_core` directly and builds one OpenHuman `Agent` per company role. It supplies its own:

- inference `Provider`
- `Memory` implementation over OpenCompany's context store
- `ToolPolicy`
- persona prompt and per-agent workspace

The integration currently needs only `AgentBuilder`, `Agent::turn`, `Agent::last_turn_usage`, the provider/memory/policy traits, prompt construction, and a tool dispatcher. It does **not** need OpenHuman's desktop shell, HTTP/Socket.IO server, channel listeners, cron/heartbeat/subconscious loops, updater, memory workers, local-model/audio stack, voice, meetings, web3, media generation, MCP, skills runtime/catalog, flows, or OpenHuman-owned persistence services.

OpenCompany already depends on OpenHuman with `default-features = false`, but that is not yet a truly small library build. Several heavyweight dependency families and global/runtime initializers remain unconditional. OpenCompany also retains one `Agent` for every manifest role, so per-agent overhead matters: real company templates contain as many as 8–9 agents.

`ServiceSet::none()` and `DomainSet::harness()` are useful for `CoreBuilder` embedders, but OpenCompany bypasses `CoreBuilder` and uses `AgentBuilder` directly. Turning off background services alone therefore does not solve this use case.

### 1. Define an embedded-minimal contract

Add a documented Cargo profile/feature recipe for hosts such as OpenCompany. It should compile only the API and implementation needed for externally supplied provider, memory, policy, tools, prompt, workspace, and agent turns.

Keep today's default desktop build unchanged.

### 2. Profile before choosing the final cuts

Measure and publish the main contributors from:

- heap allocations retained by one agent and by an 8-agent roster
- Tokio worker count and touched stack pages
- controller/schema registries and process-global singletons
- unconditional bootstrap/store initialization
- mmap/shared-library pages
[...]

Issue #4796, opened 2026-07-13
feat(core): DomainSet on CoreBuilder + DomainRegistration filter seam (feature-gate prerequisite)


senamakel:
Part of #4795 — prerequisite for all per-gate issues.

## Scope

Introduce the runtime gating seam that every feature gate filters through:

- `DomainSet` struct on `CoreBuilder` (`src/core/runtime/builder.rs`), sibling of `ServiceSet`, with presets `full()` (default — today's behavior), `harness()` (agent + memory + threads + config + security only), `none()`.
- Collapse per-domain registration into a single `DomainRegistration` entry (controllers + tools + store init + bus subscribers) per the phase-2 plan in `docs/plans/pluggable-core/phase-2-corecontext.md`, so gating is one filter over the registration list in `src/core/all.rs` — not scattered `if`s across `all.rs`/`jsonrpc.rs`.
- `/schema` and namespace descriptions derive from the filtered set.

## Definition of done

- [ ] `DomainSet::full()` default produces byte-identical registration to today (registry validation test)
- [ ] `DomainSet::harness()` build boots, serves `/rpc`, and exposes only the harness namespaces; used by `examples/embed_headless.rs`
- [ ] Excluded domains: controllers unknown-method, tools absent, stores not initialized, subscribers not registered
- [ ] Unit tests for the filter + docs updated

Issue #4795, opened 2026-07-13
Feature gates for core subsystems — tracking (lightweight harness builds)


senamakel:
## Problem

A lightweight harness build of the core today still carries **everything**: workflows/flows, skills, MCP, meet, webviews/channels, web3/wallet, voice, media generation, and ~130 other domains. Two costs:

1. **Runtime bloat** — `build_registered_controllers()` in `src/core/all.rs` unconditionally registers every domain's controllers; the agent tool surface, store init, and bus subscribers follow. Even `ServiceSet::none()` (from #4791) only skips *background services* — the full RPC/tool surface still initializes. A harness that wants "agent loop + memory + threads" pays for all of it: startup time, memory, tool-surface/prompt bloat, attack surface.
2. **Compile-time bloat** — every domain compiles into every binary. Cargo `[features]` today only cover leaf concerns (`tokenjuice-treesitter`, `sandbox-*`, `whatsapp-web`, `e2e-test-support`), not subsystem families. Embedders (`examples/embed_headless.rs`, `openhuman-fleet`, stdio-MCP hosts) can't build a slim core.

## Approach

Group subsystems behind **per-family feature gates**, one tracked issue per gate (linked below). Each gate spans two layers:

- **Cargo feature** (compile-time): sheds the domain's crates/deps from the build. **Default-on** — the Tauri desktop app always builds with **all gates enabled**; shipped product is unchanged.
- **`DomainSet` flag** (runtime): a selector on `CoreBuilder` (`src/core/runtime/builder.rs`) next to `ServiceSet`, so embedders can further narrow at runtime what was compiled in. `DomainSet::full()` is the default; `DomainSet::harness()` = agent + memory + threads + config + security only.

Effects of a disabled gate: controllers not registered (unknown-method over RPC), agent tools absent, stores not initialized, bus subscribers not registered, `/schema` reflects the actual surface.

The clean seam is the **phase-2 `DomainRegistration` collapse** already planned in `docs/plans/pluggable-core/README.md` — once each domain registers via a single `DomainRegistration` entry (controllers + tools + stores + subscribers), gating is one filter over that list instead of scattered `if`s.

## Per-gate definition of done (applies to every child issue)

- [ ] Cargo feature added to root `Cargo.toml`, in `default`; Tauri shell builds unchanged with all gates on.
- [ ] `DomainSet` flag wired through registration, store init, and subscriber registration.
- [ ] **E2E with gate enabled**: full existing suites pass unchanged.
[...]

PR #5061, opened 2026-07-20
feat(core): gate the desktop-automation cluster behind a default-ON feature (#5049)

YellowSnnowmann:
## Summary

- Adds a **default-ON** `desktop-automation` Cargo feature that compiles out the ~24.5k-LOC desktop-automation cluster — `accessibility`, `screen_intelligence`, `autocomplete`, `desktop_companion`, and the `computer` agent-tool family — when disabled.
- Uses the established **facade + `stub.rs`** pattern (voice/web3) and the **skills type carve-out** rule: inert type modules stay ungated so enabled/disabled builds share one definition.
- Wires the runtime `DomainSet::desktop_automation` axis (#4796); the three cluster controllers unregister and the 15 `screen_intelligence_*` + 4 `computer` agent tools become **absent** (not error-degraded) when off.
- Sheds the exclusive `uiautomation` dependency (proven via `cargo tree -i`); forwarded to the desktop shell.
- **With the gate on, the build is byte-identical to today** (Phase 1 = gate only; crate extraction is a filed follow-up).

## Problem

None of this cluster belongs in a headless or embedded core — it is desktop-only by nature and the second-largest ungated family after channels. Keeping it unconditional inflates the embedded `openhuman_core` binary and RAM footprint (#5046) and keeps OS-automation dependencies in every build. The cluster is **not a leaf**: it is reached from always-on code and, critically, the `accessibility` module is consumed by the independently-toggled `voice` gate — so the gate must be built carefully to keep every feature combination compiling.

## Solution

- **`accessibility`** → facade + a **rich, voice-aware stub**. The stub reproduces the exact public surface reached by `voice`, `text_input`, `app_state`, and siblings, byte-for-byte; where the real module has a non-macOS arm it **mirrors that arm** (e.g. overlay/validate return `Ok`), so a `voice`-on / `desktop-automation`-off build behaves like a non-desktop platform. `automate::run<B>` is generic so the caller's `&RealBackend` binds without stubbing the backend trait.
- **`autocomplete` / `screen_intelligence` / `desktop_companion`** → facade + minimal stubs (empty registration aggregators + the always-on caller surface; carved status types re-exported).
- **`computer`** → leaf gate; per-element `#[cfg]` on the tool registrations (the `mcp`/`web3` idiom).
- **Type carve-out**: `GlobeHotkey*` moved into ungated `accessibility::types`; `AutocompleteStatus` / `AccessibilityStatus` kept ungated. Zero type duplication.
- **DomainSet**: `DomainGroup::DesktopAutomation` + field + `full/harness/none` presets + `allows` arm + preset tests; the three aggregators retagged from `Platform`.
- **Both-ways tests** in `core/all_tests.rs` + `tools/ops_tests.rs` pin registration/tool presence in both directions.
- The `enigo` / `arboard` / `rdev` deps are **not** claimed as shed — `enigo` is shared with `voice`, `rdev`/`arboard` are voice-owned. Only `uiautomation` is exclusive.

Verified locally across all four combinations (on/on, off/on-voice, off/off) plus fmt, clippy, feature-forwarding, and the `uiautomation` shed proof.

## Submission Checklist

- [x] Tests added or updated — `desktop_automation_controllers_registered_when_feature_on` / `..._absent_when_feature_off` and `screen_intelligence_tools_absent_when_feature_off` (asserts the tools are absent, not degraded).
- [x] **Diff coverage ≥ 80%** — the both-ways gate tests cover the changed registration/gating lines; the stub bodies are exercised by the disabled-build compile + tests.
- [x] Coverage matrix updated — `N/A`: compile-time gate over existing behaviour; no new user-facing feature row.
- [x] All affected feature IDs listed — `N/A`.
- [x] No new external network dependencies — none; `uiautomation` is made **optional**, not added.
- [x] Manual smoke checklist updated — `N/A`: gate-on build is byte-identical to the shipped app.
- [x] Linked issue closed via `Closes #NNN` — `N/A`: this is Phase 1 (the gate). Phase 2 (extraction into its own crate) is a separate follow-up; see `## Related`.

## Impact

- **Build/binary**: slim/headless builds can now drop the desktop-automation cluster and the `uiautomation` dependency. The shipped desktop app is unchanged (default-ON, forwarded to the shell).
- **Compatibility**: no wire-contract or controller-schema change with the gate on. With the gate off, the cluster's `/rpc` methods become unknown-method and its tools are absent by design.

## Related

- Contributes to #5046 (the embedded RAM budget this helps unblock).
- Tracking: #4795 (core subsystem gates), #4796 (`DomainSet` seam).
- Follow-up: **Phase 2** — lift the gated tree into its own crate (to be filed).

## Changes since initial submission (review-driven)

- **Slim-build clippy fix** (`6544a6b1b`) — gate the `PermissionState` import in `globe.rs` to macOS; the `GlobeHotkey*` type carve-out left it used only by macOS code, so it was an unused import (Linux `-D warnings`).
- **Disabled-build test gating** (`299e9ec65`, CodeRabbit surfaced a real bug) — six pre-existing tests hard-asserted the now-gated surface and were not `#[cfg]`-gated in lockstep (the AGENTS.md testing gotcha; CI's `cargo check`-only smoke lane never caught them): the `DESKTOP_ALWAYS_ON` / `DESKTOP_DEFAULT_OFF` test consts, `all_tools_includes_computer_control_when_enabled`, and three `core::` tests (`cli`/`jsonrpc`/`legacy_aliases`). Disabled-build `core::` is now green.
- **Diagnostics on disabled paths** (`299e9ec65`, CodeRabbit) — grep-friendly `debug!` on the screen-intelligence RPC/CLI + autocomplete-start rejections and the accessibility disabled actions (no user text/PII).
- **Voice mic path fix** (`0b76e1b05`, Codex — real bug) — the accessibility stub returned `Unknown` mic permission, which broke `voice` dictation preemptively in a voice-on / desktop-automation-off build. The stub now returns `Granted`, matching the real Linux/non-macOS cpal arm.

> **Pre-existing, out of scope:** three `knowledge_*` disabled-build tool tests fail because they assert `skills`/`flows`-gated tools ungated — unrelated to this gate; they fail on `main` too (CI never runs the disabled-build test suite).
> ⚠️ This PR also contains an **unrelated** `#[cfg(test)]`-only event-bus fix cherry-picked from #5063 — see the pinned "Note for reviewers" comment. It self-drops on rebase once #5063 merges.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: feat/desktop-automation-gate-5049

## Summary by CodeRabbit

* **New Features**
  * Added a desktop automation feature covering autocomplete, screen intelligence, accessibility, desktop companion, and computer-control capabilities.
  * Desktop automation is enabled by default in standard builds.
  * Slim or headless builds can omit desktop automation while retaining compatible interfaces.

* **Bug Fixes**
  * Disabled desktop automation now reports clear unavailable or unsupported statuses instead of attempting inactive operations.
  * Tool and controller availability now correctly follows the selected build configuration.


src/openhuman/tools/ops_tests.rs
```diff
@@ -2160,20 +2165,39 @@ fn money_default_off_tools_retained_when_opted_in() {
...
+    #[cfg(feature = "desktop-automation")]
     "screen_intelligence_globe_listener_start",
+    #[cfg(feature = "desktop-automation")]
     "screen_intelligence_globe_listener_poll",
+    #[cfg(feature = "desktop-automation")]
     "screen_intelligence_globe_listener_stop",
+    #[cfg(feature = "desktop-automation")]
     "screen_intelligence_request_permissions",
+    #[cfg(feature = "desktop-automation")]
     "screen_intelligence_request_permission",
```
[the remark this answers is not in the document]
YellowSnnowmann: Good catch — a real disabled-build test failure. Fixed in 299e9ec65: gated the two `screen_intelligence_*` entries in `DESKTOP_ALWAYS_ON`, and on audit also gated the sibling `DESKTOP_DEFAULT_OFF` (two `request_permission*` entries) and the `all_tools_includes_computer_control_when_enabled` test (`mouse`/`keyboard`) — all `#[cfg(feature = "desktop-automation")]`, matching the existing `mcp` gating. Disabled-build `tools::ops::tests` now pass except three **pre-existing** `knowledge_*` failures (skills/flows-gated tools asserted ungated — unrelated to this gate; they fail on main too since CI never runs the disabled-build test suite).

src/openhuman/screen_intelligence/stub.rs
```diff
@@ -0,0 +1,172 @@
...
+            CaptureImageRefResult {
+                ok: false,
+                image_ref: None,
+                mime_type: "image/png".to_string(),
+                bytes_estimate: None,
+                message: DISABLED.to_string(),
+            },
+            vec![DISABLED.to_string()],
+        ))
+    }
```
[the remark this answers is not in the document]
YellowSnnowmann: Done in 299e9ec65 — added `debug!` diagnostics to the `screen_intelligence` `capture_image_ref` RPC and CLI rejections, and the `autocomplete start` CLI rejection.

src/openhuman/accessibility/stub.rs
```diff
@@ -0,0 +1,230 @@
...
+        _goal: &str,
+        _backend: &B,
+        _opts: AutomateOptions,
+    ) -> AutomateOutcome {
+        AutomateOutcome {
+            success: false,
+            summary: DISABLED.to_string(),
+            steps: Vec::new(),
+        }
+    }
```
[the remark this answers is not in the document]
YellowSnnowmann: Done in 299e9ec65 for the meaningful disabled **actions**: `accessibility::stub` `apply_text_to_focused_field` / `send_backspace` / `globe_listener_start` / `automate::run` now log a grep-friendly `[accessibility] … disabled` line (never logging user text or PII). I deliberately left the inert getters unlogged (`foreground_context → None`, `is_terminal_app → false`, `detect_* → Unknown`) — a log per no-op would spam without diagnostic value; the disabled state is observable via controller-absence + the entry-point rejection logs. Can extend to the `autocomplete::stub` engine getters if you'd prefer.

vanthryx00 (approved):

src/openhuman/accessibility/stub.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// ── permissions ──────────────────────────────────────────────────────────
+
+/// Real: `permissions::detect_microphone_permission`. Microphone capture lives in
+/// the `voice` domain but reads its permission through here; without the platform
+/// backend the state is unknown.
+pub fn detect_microphone_permission() -> PermissionState {
+    PermissionState::Unknown
```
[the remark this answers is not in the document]
YellowSnnowmann: Good catch — verified and fixed in 0b76e1b05. `voice::audio_capture::record_on_thread` treats `Unknown` as request-then-recheck-else-fail, and the stub's `Unknown` + no-op `request_microphone_access` broke dictation preemptively in a voice-on / desktop-automation-off build. The real Linux/non-macOS `detect_microphone_permission` cpal-probes and returns `Granted` when an input device exists, so the stub now returns `Granted` (and the `microphone` field of `detect_permissions`) — voice proceeds and any real capture error surfaces from cpal instead of a preemptive permission failure.

YellowSnnowmann:
### 📌 Note for reviewers — the event-bus change here is an unrelated, self-healing flake fix

This PR carries a small **test-only** change that is **not** part of its objective:
- `src/openhuman/learning/startup.rs` (test module)
- `src/openhuman/web_chat/event_bus.rs` (test module)

They fix two pre-existing full-suite flakes — `learning::startup::tests::learning_subscriber_fires_*` and `web_chat::event_bus::tests::automation_halt_subscriber_handle_publishes_correct_payload`.

**Root cause (corrected):** not a capacity problem. Both tests asserted against the **process-global broadcast bus**, so under the ~12,600-test parallel suite they received events published by *other* concurrent tests — cross-test interference, which no channel-capacity bump can fix. (An earlier capacity-floor attempt was pushed and then reverted here once CI proved it ineffective.)

**Fix — real isolation:**
- `learning::startup` now publishes a `DocumentCanonicalized` event on a private `EventBus::create(64)` with a locally-subscribed `EmailSignatureSubscriber`, instead of the shared global bus. A synchronous `register_email_signature_subscriber` guard test covers the channel-independent registration path (#5003).
- `web_chat::event_bus` drains foreign events from the shared web-channel bus and matches on `event == "automation_halt"` rather than taking whatever is first in the buffer.

Production ships **byte-identical** — the diff is confined to `#[cfg(test)]` code; no new deps, no weakened assertion.

**Why it's in this PR:** this PR changes `Cargo.toml`, which triggers the *full* coverage suite where the flake reproduces. #5063 (the canonical standalone fix) only gets *scoped* coverage and cannot self-verify — so the same commit is cherry-picked here so this PR's coverage lane can go green. **It self-drops on rebase once #5063 merges to `main`.**

### Pre-existing CI red unrelated to this PR (not merge-blocking — not in `pr-ci-gate`)
- **`Rust Feature-Gate Smoke (gates off)`** — a stale EXPECTED allowlist in that lane (`composio/action_tool.rs` gained a `#[cfg(feature="flows")]` test in #4853 but was never allowlisted). It fails on **every** current PR; the canonical allowlist fix lives in #5063 and likewise self-heals on rebase once that merges.

Cargo.toml
```diff
@@ -458,6 +458,27 @@ skills = []
...
+# CARVE-OUT: the inert type modules stay compiled in BOTH builds —
+# `accessibility::types` (incl. the `GlobeHotkey*` structs), `autocomplete::types`
+# (`AutocompleteStatus`), and `screen_intelligence::types` (`AccessibilityStatus`,
+# `CaptureImageRefResult`) — because always-on callers (`text_input`, `voice`,
+# `app_state`) name them. Only behaviour is gated; the stubs re-export the one real
+# type definition. See AGENTS.md "skills gate — the type carve-out".
+#
+# NOTE: only `uiautomation` is exclusive and thus shed. `enigo` is shared with the
+# `voice` domain; `rdev` / `arboard` are voice-owned — none of those are dropped here.
+desktop-automation = ["dep:uiautomation"]
```
CodeGhost21, line 473: **(Structural — highest concern) CI never compiles or tests the disabled build.**

The safety of every stub↔real signature parity here rests entirely on a manual local `--no-default-features` build. The PR body itself notes CI's `cargo check`-only smoke lane never caught the six pre-existing test-gating bugs CodeRabbit surfaced, and that three `knowledge_*` disabled-build tests already fail on `main` undetected.

This gate multiplies the feature combinations that can silently rot. Strongly recommend a follow-up (or this PR) adding a slim `cargo check --no-default-features --features <minimal>` lane — otherwise the next contributor who adds an always-on caller of a gated symbol breaks the slim build with no signal. Good candidate to file alongside the Phase 2 extraction.

src/openhuman/accessibility/stub.rs
```diff
@@ -0,0 +1,245 @@
...
+        PermissionKind::InputMonitoring => "input_monitoring",
+        PermissionKind::Microphone => "microphone",
+    }
+}
+
+/// Real: `permissions::detect_permissions`. The desktop-automation permissions
+/// (screen-recording / accessibility / input-monitoring) are unknown when the
+/// gate is off; microphone mirrors `detect_microphone_permission` above so the
+/// `voice`-on / `desktop-automation`-off build keeps a usable mic state.
+pub fn detect_permissions() -> PermissionStatus {
```
CodeGhost21, line 168: **(Minor nit) `Unknown` vs the real non-macOS `Unsupported`.**

The real non-macOS `detect_permissions` returns `PermissionState::Unsupported` for `screen_recording` / `accessibility` / `input_monitoring`, but this stub returns `Unknown`. Elsewhere the stub docs claim the disabled build "behaves like a non-desktop platform" — this slightly contradicts that.

I confirmed no always-on / voice caller branches on these three fields (only `microphone` is read outside the gated tree), so it's cosmetic. For exactness, `Unsupported` would better match the stated intent.

src/openhuman/learning/startup.rs
```diff
@@ -270,13 +270,20 @@ mod tests {
...
 
     #[tokio::test]
-    async fn learning_subscriber_fires_with_no_channel_configured() {
-        init_global(DEFAULT_CAPACITY);
-        let (tmp, _client) = test_client();
-        // Make the memory client ready so the full Platform wiring runs — no
-        // channel runtime is ever constructed in this test.
-        let _ = crate::openhuman::memory::global::init(tmp.path().join("workspace"));
-        register_learning_subscribers(tmp.path().to_path_buf());
+    async fn learning_subscriber_fires_on_document_event() {
```
CodeGhost21: **(Scope) Unrelated flaky-test fix riding along.**

This isolated-bus refactor is a test-only flake fix, not gate work. Not blocking, but it inflates the review surface — ideally it lands on its own PR (the sibling `web_chat/event_bus.rs` change is at least disclosed as cherry-picked from #5063 and self-drops on rebase).

src/openhuman/web_chat/event_bus.rs
```diff
@@ -505,6 +505,27 @@ mod tests {
     /// - `args.reason` and `args.source` echoed from the domain event
     #[tokio::test]
     async fn automation_halt_subscriber_handle_publishes_correct_payload() {
+        // The web-channel bus is a process-shared broadcast; under the parallel test
+        // suite other tests publish to it too, so take THIS subscriber's next
+        // `automation_halt` event rather than whatever happens to be first in the
+        // buffer (which flaked as an "event name mismatch").
+        fn next_automation_halt(
```
CodeGhost21: **(Very minor) Narrow flake window remains.**

In the "resumed" block, a concurrent test publishing an `automation_halt` event between the first drain and the resumed publish could be returned instead of the intended event. Much less likely than the bug this fixes, and test-only — noting for completeness.

CodeGhost21 (commented):
Reviewed the desktop-automation gate. Overall this is high-quality, idiomatic work: the facade + `stub.rs` pattern is applied cleanly, the type carve-out (`GlobeHotkey*` → `accessibility::types`) is correct with zero duplication, both-directions tests pin the gate, and the dependency-shed claim is disciplined (only `uiautomation` is exclusive). The `detect_microphone_permission() → Granted` and `PermissionState` import fixes are genuine catches.

I verified there are **no callers of the macOS-only permission exports** (`detect_screen_recording_permission`, etc.) outside the gated tree, so the stub omitting them is safe on a slim macOS build.

Approve-with-follow-up. The one thing worth pushing on before this pattern proliferates is a slim-build CI lane (see the Cargo.toml comment) — the rest are non-blocking nits filed inline.

src/openhuman/accessibility/stub.rs
YellowSnnowmann: Good catch — fixed in 8be6f1982. The disabled-build stub `detect_permissions()` now returns `PermissionState::Unsupported` for `screen_recording` / `accessibility` / `input_monitoring`, matching the real non-macOS `detect_permissions` (`permissions.rs:267`, pinned by `detect_permissions_non_macos_reports_unsupported_for_desktop_perms`) and the sibling struct that already used `Unsupported`. `microphone` stays `Granted`. Confirmed no always-on/voice caller branches on those three fields, and the disabled build (`--no-default-features --features tokenjuice-treesitter`) compiles clean.

src/openhuman/web_chat/event_bus.rs
YellowSnnowmann: Tightened in 88f7b946c. `next_automation_halt` now filters on the `engaged` payload (`true` for halted, `false` for resumed) in addition to the event name, so a concurrent test's halt/resume on the shared web-channel bus can no longer be returned in place of the intended event. The 6 `web_chat::event_bus` tests pass. (The helper is cherry-picked from #5063; the same tightening can land there on rebase.)

Cargo.toml
YellowSnnowmann: Agreed, and I think keeping the full slim *test* lane out of this PR is the right scope call. You're correct that the `Rust Feature-Gate Smoke (gates off)` lane only `cargo check`s the disabled build and never runs its test suite — which is exactly how the pre-existing gated-test breakage went unnoticed.

What this PR does add on that front: it extends the self-maintaining gated-test guard so the allowlist can't silently drift as the gate surface grows — `desktop-automation` is now in the guard regex and the new `core/autocomplete_cli_adapter.rs` is in `EXPECTED` (8e9ce8be1). That's still a compile-time guard, not a disabled-build test run.

A dedicated slim `cargo test --no-default-features --features <minimal>` lane belongs with the Phase 2 extraction follow-up rather than expanding this PR — noting it here so it's tracked alongside that work.

src/openhuman/learning/startup.rs
YellowSnnowmann: Fair point on scope. I've kept the isolated-bus refactor here deliberately: it fixes a full-suite flake that this PR's own gates-off test + coverage lanes would otherwise hit intermittently, so dropping it would destabilize this branch's checks. It's test-only and low-risk. Happy to extract it to a standalone PR if you'd prefer the clean separation — say the word and I'll pull it out and rebase.

Issue #5003, opened 2026-07-16
fix(learning): email-signature subscriber + rebuild trigger silently depend on channels being configured


oxoxDev:
## Problem

Two `learning` registrations are **always-on Platform capabilities** whose sole caller lives inside the channels bootstrap. If channels never boots, they are never registered — silently.

| Registration | Defined in | Sole caller |
|---|---|---|
| `learning::extract::signature::register_email_signature_subscriber` | `src/openhuman/learning/extract/signature.rs:537` | `src/openhuman/channels/runtime/startup.rs:214` |
| `learning::scheduler::register_event_trigger` + `spawn_rebuild_loop` | `src/openhuman/learning/scheduler.rs:134` | `src/openhuman/channels/runtime/startup.rs:229,245` |

`learning` is tagged `DomainGroup::Platform` (`src/core/all.rs:687`) — i.e. always-on, not gated by any domain selection.

But `src/core/runtime/builder.rs` → `spawn_channels_service()` (`src/core/runtime/services.rs:156-183`) **skips `start_channels` entirely** when either:

- `OPENHUMAN_DISABLE_CHANNEL_LISTENERS` is set, **or**
- `!config.channels_config.has_listening_integrations()` — i.e. **the user has no chat medium configured**

## Impact (reproducible on `main` today, no feature flags involved)

**Any user who has not configured a chat medium already never gets the learning email-signature subscriber or the learning rebuild trigger.** A Platform capability is silently conditional on an unrelated domain's configuration. There is no error, no warning, no Sentry event — the subscriber simply never registers.

## Root cause

`channels::runtime::startup::start_channels` is misnamed. Despite the name it is a **process-wide bootstrap** (1,269 LOC) that registers subscribers for core infra, health, skills, memory-conversations, memory-sync, composio, agent-meetings, task-sources, agent, and learning — before doing any channel work.

Most of those registrations are duplicated on a reliable path (`src/core/jsonrpc.rs` / `src/core/runtime/services.rs`); `src/core/jsonrpc.rs:2028-2031` says so explicitly:

> "…so schedule/app-event workflows still dispatch when no realtime channel is configured or `OPENHUMAN_DISABLE_CHANNEL_LISTENERS` short-circuits `start_channels`."

The two `learning` registrations above are the ones with **no** such duplicate — verified by whole-repo grep for their sole callers.

## Proposed fix

[...]

PR #5064, opened 2026-07-20
fix(embeddings): differentiate custom endpoint verification failures (#5017)

### LINKED ISSUE #5017: Custom OpenAI-compatible embedding endpoint fails verification even with valid endpoint and key
## Summary

Setting up a Custom (OpenAI-compatible) embedding provider fails with "Couldn't verify the embeddings endpoint — the test embed failed" even when the same endpoint and API key work correctly for chat inference. Multiple OpenAI-compatible providers reproduce the same failure.

## Problem

In **Connections → Vector encoding model for memory retrieval → Custom (OpenAI-compatible)**, users enter:
- A valid OpenAI-compatible base URL (e.g. `https://api.<provider>.com/v1`)
- A model name (e.g. `gpt-5-mini`)
- Dimensions (e.g. `1024`)
- A valid API key

On "Test connection" or "Save & switch", the dialog shows:

> **Couldn't verify the embeddings endpoint — the test embed failed. Make sure the endpoint is reachable and serving an embedding model, then save again.**

The same endpoint + key successfully handles chat completions in the Chat provider settings. Testing with multiple different OpenAI-compatible providers produces the same error, ruling out a single misconfigured endpoint.

## Steps to reproduce

1. Open Connections → Vector encoding model → Custom (OpenAI-compatible).
2. Enter a working OpenAI-compatible base URL, an embedding model name, dimensions, and an API key.
3. Click **Test connection**.
4. Observe the verification failure error.
5. Confirm the same endpoint/key works in the Chat provider BYOK settings.

## Expected behaviour

If the endpoint responds correctly to a test embedding request (`POST /v1/embeddings`) with the given model and key, verification should succeed and the provider should be saved.

## Possible causes

- Verification may be calling `POST /embeddings` with wrong request shape (e.g. wrong `input` format, missing `encoding_format`, wrong content-type).
- Verification may hit `/v1/models` or a health endpoint not present on all OpenAI-compatible hosts.
- Model name + dimensions mismatch: the app may not be forwarding the user-supplied model name in the test request.
- API key may not be included in the test embed request header.
- Dimension validation may reject a valid response if the returned vector length doesn't match the entered dimension.

## Acceptance criteria

- [ ] Custom OpenAI-compatible embedding endpoint verifies successfully when the endpoint correctly serves `POST /v1/embeddings`
- [ ] Test request uses the user-supplied model name and API key
- [ ] Error message distinguishes between: network/DNS failure, auth failure (401/403), wrong model, and dimension mismatch — so users know what to fix
- [ ] Endpoint that works for chat BYOK with the same base URL and key also works for custom embeddings without extra configuration
- [ ] Regression: add a test against a mock OpenAI-compatible `/v1/embeddings` endpoint to guard the verification flow

## Related

- Managed embeddings session bug: #4947
- Memory sync degraded state: memory pipeline shows degraded when embeddings fail silently

M3gA-Mind:
## Summary

- Custom (OpenAI-compatible) embedding endpoints failed setup verification with one generic "test embed failed" even when the endpoint + key work for chat.
- Root cause is **not** the request: the probe sends a correct `POST /v1/embeddings` with the user's model, Bearer key, and `{"input":[…],"model":…}` body — a conformant endpoint verifies (proven by a new mock-endpoint test).
- The defect is that `classify_embed_probe` collapsed distinct failures (auth 401/403, wrong/incompatible model 400, network/DNS, dimension mismatch) into one generic message.
- Differentiate those failures into dedicated, actionable codes/messages so users know what to fix.
- Frontend now surfaces any non-wipe error code as a verification failure (fixes a latent "unknown code silently treated as a save").

## Problem

Issue #5017: a user enters a working OpenAI-compatible base URL + model + key under Connections → Vector encoding model → Custom. The same endpoint/key work for chat, but embedding verification fails with *"Couldn't verify the embeddings endpoint — the test embed failed."*

I reproduced the exact test-embed request by driving the same provider construction the save-time probe uses against a mock server. **The request is correct:**

```
POST /v1/embeddings HTTP/1.1
content-type: application/json
authorization: Bearer sk-…            ← key IS forwarded
{"input":["connection test"],"model":"gpt-5-mini"}
```

Path, method, model, key, and body are all right, and a conformant `/v1/embeddings` host verifies successfully. So the reporter's guesses (a) wrong shape, (b) `/models`/health path, (c) model not forwarded, (d) key missing, (e) dimension validation are all ruled out for this flow.

The reporter used `gpt-5-mini` — a **chat** model — which the endpoint legitimately rejects at `/v1/embeddings` (HTTP 400 "does not support embeddings"). Verification *should* fail, but the message never said the model isn't an embeddings model, so "works for chat, fails for embed" looked like a bug. Auth (401/403), network/DNS, and dimension-mismatch failures were equally indistinguishable.

## Solution

- `src/openhuman/embeddings/rpc.rs` — `classify_embed_probe` now maps failures to distinct codes/messages, tolerant of both embed-error wire shapes (`openai embeddings returned HTTP <code>…` and `Embedding API error (<code>)…`):
  - `EMBEDDINGS_MODEL_INCOMPATIBLE` — 400/422 + model-rejection body (the reporter's case: chat model in the embeddings field).
  - `EMBEDDINGS_AUTH_FAILED` — 401/403 (embeddings key is stored separately from the chat BYOK key).
  - `EMBEDDINGS_ENDPOINT_UNREACHABLE` — DNS / refused / connect / timeout.
  - `EMBEDDINGS_DIMENSION_MISMATCH` — endpoint ignored the `dimensions` param and returned a different length.
  - Existing `NO_MODEL_LOADED` / `ENDPOINT_NO_API` / generic `VERIFICATION_FAILED` preserved. Added a PII-safe log of the classified code (never the raw detail).
- `app/src/components/settings/panels/EmbeddingsPanel.tsx` — failure detection matches on shape (any error code other than the wipe-confirm) instead of an allow-list, so the new codes surface their backend message rather than a new code being silently treated as a successful save.

Design note: verification messages are backend-emitted English strings (same pattern as the existing `EMBEDDINGS_*` messages the panel already renders), so **no new frontend UI literals and no new i18n keys** are introduced.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 4 differentiation tests (incompatible-model, auth, unreachable, dimension-mismatch) + a mock `/v1/embeddings` regression proving a conformant host verifies and the request carries the user's model + Bearer key. Pre-existing `other_failures_and_timeout` test updated to the new behavior (fail-before / pass-after).
- [x] **Diff coverage ≥ 80%** — changed Rust lines are the `classify_embed_probe` branches + helpers, each exercised by the new unit tests; the one-line frontend conditional is covered by the existing panel behavior. `openhuman::embeddings::rpc::tests` all pass (19/19).
- [x] Coverage matrix updated — `N/A: refines an existing feature's error handling; no feature row added/removed/renamed`.
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix feature IDs affected`.
- [x] No new external network dependencies introduced — the regression test uses an in-process `axum` mock (already a dev-dependency); no real network.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: settings-panel error messaging only, no release-cut surface change`.

## Impact

- Desktop (Settings → Connections → Vector encoding model → Custom). No runtime/perf/migration impact.
- Security: the new log records only the classified error code, never the raw endpoint response body (which can carry response detail).

## Related

- Closes: #5017
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/GH-5017-custom-embed-verify-errors
- Commit SHA: dd1929d67

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — EmbeddingsPanel.tsx prettier-clean
- [x] Focused tests: `GGML_NATIVE=OFF cargo test --lib openhuman::embeddings::rpc::tests` — 19 passed, 0 failed
- [x] Rust fmt/check (if changed): `rustfmt --check src/openhuman/embeddings/rpc.rs` — clean
- [x] Tauri fmt/check (if changed): N/A — no `app/src-tauri` change

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: setup-time embedding verification now returns a distinct, actionable message per failure class instead of one generic error.
- User-visible effect: a custom endpoint that works for chat but is pointed at a non-embeddings model / bad key / unreachable host now tells the user exactly which to fix; a conformant embeddings endpoint verifies unchanged.

### Parity Contract

- Legacy behavior preserved: conformant-endpoint pass, empty-vector reject, `NO_MODEL_LOADED`/`ENDPOINT_NO_API`/`DIMENSION_CHANGE_REQUIRES_WIPE` codes, and the `/models` name-mismatch guidance path are all unchanged.
- Guard/fallback/dispatch parity checks: any unclassified failure (e.g. 5xx) still falls through to the generic `EMBEDDINGS_VERIFICATION_FAILED`, so no failure is ever misreported as a success.

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved custom embeddings setup error handling with clearer feedback for authentication, model compatibility, connectivity, and dimension mismatch issues.
  * Kept setup dialogs open when verification fails, allowing errors to be reviewed and corrected.
  * Preserved the existing confirmation flow for dimension changes that require data removal.
* **Tests**
  * Expanded coverage for embeddings verification and custom endpoint requests.

### DIFF
--- a/app/src/components/settings/panels/EmbeddingsPanel.tsx
+++ b/app/src/components/settings/panels/EmbeddingsPanel.tsx
@@ -311,15 +311,19 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
         confirm_wipe: false,
       });
       // Setup-time verification failed: the endpoint couldn't prove it can
-      // embed, so the config was NOT saved. Covers no `/embeddings` route
-      // (TAURI-RUST-5JR), LM Studio with no model loaded (TAURI-RUST-4P4), and
-      // any other probe failure/timeout. Keep the setup popup open and surface
-      // the actionable message so the user can fix it (load a model, correct the
-      // endpoint, …) and retry.
+      // embed, so the config was NOT saved. update_settings only ever returns an
+      // `error` for a verification failure or the dimension-wipe confirm, so any
+      // error code other than the wipe-confirm is a failed probe. Matching on the
+      // shape (rather than an explicit allow-list) means the differentiated #5017
+      // codes — EMBEDDINGS_MODEL_INCOMPATIBLE / _AUTH_FAILED / _ENDPOINT_UNREACHABLE
+      // / _DIMENSION_MISMATCH, plus _ENDPOINT_NO_API and _NO_MODEL_LOADED — all
+      // surface their actionable backend message instead of a new code being
+      // silently treated as a save. Keep the setup popup open so the user can fix
+      // it (pick an embeddings model, correct the key/endpoint, …) and retry.
       if (
-        result.error === 'EMBEDDINGS_ENDPOINT_NO_API' ||
-        result.error === 'EMBEDDINGS_NO_MODEL_LOADED' ||
-        result.error === 'EMBEDDINGS_VERIFICATION_FAILED'
+        typeof result.error === 'string' &&
+        result.error !== '' &&
+        result.error !== 'EMBEDDINGS_DIMENSION_CHANGE_REQUIRES_WIPE'
       ) {
         // `result.message`/`result.detail` are backend-emitted (already
         // context-specific); only the generic fallback is frontend-owned UI
--- a/app/src/components/settings/panels/EmbeddingsPanel.tsx
+++ b/app/src/components/settings/panels/EmbeddingsPanel.tsx
@@ -5,6 +5,7 @@
  * to enter the key, test connection, and save. Dimension changes show a
  * destructive confirm dialog since they invalidate stored vectors.
  */
+import createDebug from 'debug';
 import { useCallback, useEffect, useState } from 'react';
 
 import { useT } from '../../../lib/i18n/I18nContext';
@@ -32,6 +33,11 @@ import {
   SettingsTextField,
 } from '../controls';
 import { useSettingsNavigation } from '../hooks/useSettingsNavigation';
+
+// Grep-friendly, namespaced diagnostics for the custom-endpoint verification
+// flow. Logs only safe metadata (error classification code, state transitions) —
+// never the endpoint URL, API key, or backend-provided detail body.
+const log = createDebug('app:settings:embeddings');
 
 type Status =
   | { kind: 'idle' }
@@ -303,6 +309,7 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
         await setEmbeddingsApiKey('custom', setupKey.trim());
       }
       setStatus({ kind: 'saving' });
+      log('setupSaveCustom: calling update_embeddings_settings (provider=custom)');
       const result = await updateEmbeddingsSettings({
         provider: 'custom',
         model: customModel || 'embedding',
@@ -310,6 +317,12 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
         custom_endpoint: customEndpoint.trim(),
         confirm_wipe: false,
       });
+      // Safe to log the error *code* (an enum-like sentinel, e.g.
+      // EMBEDDINGS_AUTH_FAILED) — it carries no endpoint/key/detail content.
+      log(
+        'setupSaveCustom: rpc returned error_code=%s',
+        typeof result.error === 'string' && result.error !== '' ? result.error : 'none'
+      );
       // Setup-time verification failed: the endpoint couldn't prove it can
       // embed, so the config was NOT saved. update_settings only ever returns an
       // `error` for a verification failure or the dimension-wipe confirm, so any
@@ -340,10 +353,17 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
             ? `${baseMessage} (${result.detail})`
             : baseMessage
         );
+        // Verification failed: keep the setup popup open (status→idle, not error)
+        // so the user can correct the model/key/endpoint and retry.
+        log(
+          'setupSaveCustom: verification failed (code=%s) — preserving setup popup, early return',
+          result.error
+        );
         setStatus({ kind: 'idle' });
         return;
       }
       if (result.error === 'EMBEDDINGS_DIMENSION_CHANGE_REQUIRES_WIPE') {
+        log('setupSaveCustom: dimension change requires wipe — prompting confirm');
         setPendingWipe({
           provider: 'custom',
           model: customModel || 'embedding',
@@ -352,6 +372,7 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
         });
         setStatus({ kind: 'idle' });
       } else {
+        log('setupSaveCustom: verification passed — saved, reloading settings');
         await reload();
         setStatus({ kind: 'saved' });
       }
--- a/src/openhuman/embeddings/rpc.rs
+++ b/src/openhuman/embeddings/rpc.rs
@@ -206,8 +206,17 @@ pub async fn update_settings(
                     _ => 0,
                 };
                 if let Some(reject) = classify_embed_probe(outcome) {
+                    // Log the classified error code (never the raw detail — it can
+                    // carry endpoint response bodies) so support can distinguish
+                    // auth vs wrong-model vs unreachable failures (issue #5017).
+                    let reject_code = reject
+                        .value
+                        .get("error")
+                        .and_then(|v| v.as_str())
+                        .unwrap_or("EMBEDDINGS_VERIFICATION_FAILED");
                     tracing::warn!(
                         provider = effective_provider.as_str(),
+                        reject_code,
                         "{LOG_PREFIX} update_settings rejected — embeddings endpoint failed verification"
                     );
                     // Right-feedback (issue #3761): the probe failed. If the
@@ -666,8 +675,16 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
         ),
         EmbedProbe::Failed(detail) => {
             let lower = detail.to_ascii_lowercase();
-            // Reachable but no model loaded (e.g. LM Studio idle).
+            // The endpoint IS reachable and correctly shaped (POST /v1/embeddings
+            // with the user's model + key — verified conformant by the mock-endpoint
+            // regression test). The failures below are all *distinct causes*; issue
+            // #5017 was that they collapsed into one generic "test embed failed"
+            // message, so a user whose endpoint works for chat couldn't tell that
+            // (e.g.) their chosen model isn't an embeddings model, their key was
+            // rejected, or the host was unreachable. Order matters: check the
+            // specific shapes before the generic fallback.
             if lower.contains("no models loaded") {
+                // Reachable but no model loaded (e.g. LM Studio idle).
                 reject(
                     "EMBEDDINGS_NO_MODEL_LOADED",
                     "Your local embeddings server (e.g. LM Studio) is running but has no \
@@ -686,8 +703,59 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
                     "embeddings endpoint has no embeddings API — not saved",
                     Some(&detail),
                 )
+            } else if is_embedding_dimension_mismatch(&lower) {
+                // Endpoint embedded fine but returned a different vector length than
+                // the (Matryoshka) size we requested — a `text-embedding-3-*` model
+                // name pointed at a host that ignores the `dimensions` param.
+                reject(
+                    "EMBEDDINGS_DIMENSION_MISMATCH",
+                    "The endpoint returned a vector with a different length than the \
+                     dimensions you entered. Set dimensions to match the model's native \
+                     output (or clear the field to auto-detect), then save again.",
+                    "embeddings endpoint returned mismatched dimensions — not saved",
+                    Some(&detail),
+                )
+            } else if is_embedding_model_incompatible(&lower) {
+                // Reachable, authenticated embeddings API that rejected the model —
+                // the user pasted a chat/reasoning model (e.g. `gpt-5-mini`) into the
+                // embeddings model field. This is the #5017 reporter's exact case:
+                // the same model works for chat but is not an embeddings model.
+                reject(
+                    "EMBEDDINGS_MODEL_INCOMPATIBLE",
+                    "That model isn't an embeddings model on this endpoint. A chat model \
+                     (the one that works in Chat settings) can't produce embeddings — \
+                     enter an embeddings model id (e.g. text-embedding-3-small, bge-m3), \
+                     then save again.",
+                    "embeddings model is not an embeddings model — not saved",
+                    Some(&detail),
+                )
+            } else if embed_error_mentions_status(&lower, 401)
+                || embed_error_mentions_status(&lower, 403)
+            {
+                // Auth failure — key missing/wrong/lacking embeddings scope. The
+                // embeddings key is stored separately from the chat BYOK key, so
+                // "works for chat" does not imply the embeddings key is set.
+                reject(
+                    "EMBEDDINGS_AUTH_FAILED",
+                    "The endpoint rejected the API key (401/403). Enter a valid key for \
+                     this endpoint — note the embeddings key is stored separately from the \
+                     Chat provider key — then save again.",
+                    "embeddings endpoint rejected the API key — not saved",
+                    Some(&detail),
+                )
+            } else if is_embedding_endpoint_unreachable(&lower) {
+                // Transport-level failure — DNS, refused connection, TLS. The base
+                // URL is wrong or the host is down.
+                reject(
+                    "EMBEDDINGS_ENDPOINT_UNREACHABLE",
+                    "Couldn't reach the embeddings endpoint (network/DNS/connection \
+                     error). Check the base URL and that the host is reachable, then save \
+                     again.",
+                    "embeddings endpoint unreachable — not saved",
+                    Some(&detail),
+                )
             } else {
-                // Any other failure (5xx, auth, network) — didn't pass verification.
+                // Any other failure (5xx, unclassified) — didn't pass verification.
                 reject(
                     "EMBEDDINGS_VERIFICATION_FAILED",
                     "Couldn't verify the embeddings endpoint — the test embed failed. Make \
@@ -699,13 +767,60 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
             }
         }
         EmbedProbe::TimedOut => reject(
-            "EMBEDDINGS_VERIFICATION_FAILED",
+            "EMBEDDINGS_ENDPOINT_UNREACHABLE",
             "Couldn't verify the embeddings endpoint — the test embed timed out. Make sure \
              the endpoint is running and reachable, then save again.",
             "embeddings endpoint timed out during verification — not saved",
             None,
         ),
     }
+}
+
+/// Whether a lowercased embed-error detail names the given HTTP status, tolerant
+/// of both wire shapes the embeddings stack emits:
+///   `openai embeddings returned HTTP 401 Unauthorized: …` (tinyagents adapter)
+///   `Embedding API error (401 Unauthorized): …`           (legacy / other hosts)
+fn embed_error_mentions_status(lower: &str, code: u16) -> bool {
+    let code = code.to_string();
+    lower.contains(&format!("http {code}")) || lower.contains(&format!("({code}"))
+}
+
+/// A reachable, authenticated embeddings API that **rejected the model id** — the
+/// user pointed the embeddings model field at a chat/reasoning model. Gated on a
+/// 400/422 plus a model-rejection phrase so a genuine 5xx or oversized-input 400
+/// still falls through to the generic failure (issue #5017).
+fn is_embedding_model_incompatible(lower: &str) -> bool {
+    let bad_request =
+        embed_error_mentions_status(lower, 400) || embed_error_mentions_status(lower, 422);
+    bad_request
+        && (lower.contains("does not support embeddings")
+            || lower.contains("not an embedding model")
+            || lower.contains("is not an embedding")
+            || lower.contains("does not exist")
+            || lower.contains("not supported for embeddings")
+            || lower.contains("unexpected model name format"))
+}
+
+/// The post-response length guard fired: the endpoint embedded but returned a
+/// vector whose length differs from the requested (Matryoshka) `dimensions`.
+/// Canonical shape from the tinyagents adapter:
+/// `openai embed dimension mismatch: expected 1024, got 3072`.
+fn is_embedding_dimension_mismatch(lower: &str) -> bool {
+    lower.contains("dimension mismatch")
+}
+
+/// A transport-level failure (DNS, refused connection, TLS, connect timeout) —
+/// the endpoint was never reached, so the base URL is wrong or the host is down.
+/// The tinyagents adapter wraps these as
+/// `openai embeddings request to <url> failed: <reqwest error>`.
+fn is_embedding_endpoint_unreachable(lower: &str) -> bool {
+    lower.contains("request to") && lower.contains("failed")
+        || lower.contains("connection refused")
+        || lower.contains("error sending request")
+        || lower.contains("error trying to connect")
+        || lower.contains("dns error")
+        || lower.contains("failed to lookup address")
+        || lower.contains("tcp connect error")
 }
 
 /// GET `{endpoint}/models` (OpenAI-compatible) and return the served model ids.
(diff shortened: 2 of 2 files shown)


src/openhuman/embeddings/rpc.rs
```diff
@@ -708,6 +776,53 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
     }
 }
 
+/// Whether a lowercased embed-error detail names the given HTTP status, tolerant
+/// of both wire shapes the embeddings stack emits:
+///   `openai embeddings returned HTTP 401 Unauthorized: …` (tinyagents adapter)
+///   `Embedding API error (401 Unauthorized): …`           (legacy / other hosts)
+fn embed_error_mentions_status(lower: &str, code: u16) -> bool {
+    let code = code.to_string();
+    lower.contains(&format!("http {code}")) || lower.contains(&format!("({code}"))
```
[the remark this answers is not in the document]
M3gA-Mind: Addressed in eb8f6c3. `embed_error_mentions_status` now also matches the bare-status `embedding api error {code}` shape (the form the observability classifier covers), so 401/403 from that host shape route to EMBEDDINGS_AUTH_FAILED. Added a bare-status case to `classify_embed_probe_distinguishes_auth_failure` (passes locally).

app/src/components/settings/panels/EmbeddingsPanel.tsx
```diff
@@ -311,15 +311,19 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
...
+      // surface their actionable backend message instead of a new code being
+      // silently treated as a save. Keep the setup popup open so the user can fix
+      // it (pick an embeddings model, correct the key/endpoint, …) and retry.
       if (
-        result.error === 'EMBEDDINGS_ENDPOINT_NO_API' ||
-        result.error === 'EMBEDDINGS_NO_MODEL_LOADED' ||
-        result.error === 'EMBEDDINGS_VERIFICATION_FAILED'
+        typeof result.error === 'string' &&
+        result.error !== '' &&
+        result.error !== 'EMBEDDINGS_DIMENSION_CHANGE_REQUIRES_WIPE'
```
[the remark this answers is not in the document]
M3gA-Mind: Addressed in eb8f6c3. Added a namespaced `app:settings:embeddings` debug logger with entry/branch/state-transition/early-return diagnostics across the verification-classification flow (RPC call, returned error code, popup-preserving idle transition, dimension-wipe prompt, success reload). Only safe metadata is logged — endpoint URL, API key, and backend detail body are explicitly excluded.

src/openhuman/embeddings/rpc.rs
```diff
@@ -686,8 +703,59 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
...
+                // the (Matryoshka) size we requested — a `text-embedding-3-*` model
+                // name pointed at a host that ignores the `dimensions` param.
+                reject(
+                    "EMBEDDINGS_DIMENSION_MISMATCH",
+                    "The endpoint returned a vector with a different length than the \
+                     dimensions you entered. Set dimensions to match the model's native \
+                     output (or clear the field to auto-detect), then save again.",
+                    "embeddings endpoint returned mismatched dimensions — not saved",
+                    Some(&detail),
+                )
```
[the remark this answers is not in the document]
M3gA-Mind: Addressed in eb8f6c3. Removed the "or clear the field to auto-detect" clause from the EMBEDDINGS_DIMENSION_MISMATCH message per the proposed fix — for text-embedding-3-* (the only family that reaches this branch) clearing the field falls back to the stored value rather than an auto sentinel, so the advice was a no-op.

M3gA-Mind:
Addressed the review feedback in eb8f6c3:

1. **[P2] Bare-status auth errors** — `embed_error_mentions_status` now also matches the `Embedding API error {code}` (bare, no-paren) shape, so setup-time custom-endpoint verification routes 401/403 to `EMBEDDINGS_AUTH_FAILED` instead of the generic failure. Regression case added to the auth-failure test (passes locally).
2. **[Major] Verification diagnostics** — added a namespaced `app:settings:embeddings` debug logger covering the RPC call, selected error code, popup-preserving state transition, and early return. No endpoint/API-key/detail values are logged.
3. **[Minor] Dimension-mismatch remedy text** — dropped the misleading "or clear the field to auto-detect" clause (a no-op for `text-embedding-3-*`, the only family that hits this branch).

Targeted validation: `cargo test --lib classify_embed_probe_distinguishes_auth_failure` ✅, `cargo fmt` ✅, prettier ✅ on the touched TS file.

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Synced onto current main (now includes #5070 feature-gate allowlist + #5003/#5025 learning-subscriber decouple). Re-running CI.

M3gA-Mind:
@coderabbitai review

src/openhuman/embeddings/rpc.rs
```diff
@@ -206,8 +206,17 @@ pub async fn update_settings(
...
                 };
                 if let Some(reject) = classify_embed_probe(outcome) {
+                    // Log the classified error code (never the raw detail — it can
+                    // carry endpoint response bodies) so support can distinguish
+                    // auth vs wrong-model vs unreachable failures (issue #5017).
+                    let reject_code = reject
+                        .value
+                        .get("error")
+                        .and_then(|v| v.as_str())
+                        .unwrap_or("EMBEDDINGS_VERIFICATION_FAILED");
```
sanil-23, line 216: **blocker — the `/models` guidance override clobbers the new high-confidence codes**

Just below this, on *any* reject, `update_settings` calls `fetch_served_model_ids` and, if the requested model isn't in the served list, `return Ok(better)` — replacing the classified code with `EMBEDDINGS_NO_MODEL_LOADED`.

That override predates this PR and was fine when everything below it was one generic code. It isn't now. An endpoint with a **public `/models` but an authenticated `/embeddings`** (common behind gateways/proxies) will classify a 401 as `EMBEDDINGS_AUTH_FAILED`, then have it overwritten with "your model isn't served" — telling the user to fix the model name when the real problem is the key. `EMBEDDINGS_DIMENSION_MISMATCH` has the same exposure when the served list canonicalizes ids differently.

`EMBEDDINGS_ENDPOINT_UNREACHABLE` is mostly self-protecting (the `/models` fetch fails too), so this is really about auth and dimension.

The name-mismatch heuristic is only a *better* diagnosis than the generic bucket. Gate it on that:

```rust
// before
let listed_endpoint = custom_endpoint
    .as_deref()
    .or_else(|| effective_provider.strip_prefix("custom:"));
if let Some(ep) = listed_endpoint {

// after
// The `/models` name-mismatch heuristic (issue #3761) is only a *better*
// diagnosis than the generic failure. Auth, reachability and dimension
// failures are already high-confidence and specific, so a served-list miss
// must not overwrite them (#5017).
let heuristic_may_override = matches!(
    reject_code,
    "EMBEDDINGS_VERIFICATION_FAILED" | "EMBEDDINGS_ENDPOINT_NO_API"
);
let listed_endpoint = custom_endpoint
    .as_deref()
    .or_else(|| effective_provider.strip_prefix("custom:"))
    .filter(|_| heuristic_may_override);
if let Some(ep) = listed_endpoint {
```

Worth a test asserting an `EMBEDDINGS_AUTH_FAILED` survives a served-list miss.

src/openhuman/embeddings/rpc.rs
```diff
@@ -708,6 +776,59 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
...
+fn is_embedding_model_incompatible(lower: &str) -> bool {
+    let bad_request =
+        embed_error_mentions_status(lower, 400) || embed_error_mentions_status(lower, 422);
+    bad_request
+        && (lower.contains("does not support embeddings")
+            || lower.contains("not an embedding model")
+            || lower.contains("is not an embedding")
+            || lower.contains("does not exist")
+            || lower.contains("not supported for embeddings")
+            || lower.contains("unexpected model name format"))
```
sanil-23, line 807: **blocker — duplicates `observability::is_embedding_model_rejected`, and drifts from it**

`src/core/observability.rs` already owns a predicate for exactly these wire shapes, documented against real Sentry volume (TAURI-RUST-9SK, 4SA). This file establishes the reuse precedent one branch above — `classify_embed_probe` already calls `crate::core::observability::is_embedding_endpoint_absent`.

This copy diverges in both directions: it **adds** 422 and three phrases the observability one lacks, and it **omits** the Gemini OpenAI-compat arm (`invalid_argument` + `batchembedcontentsrequest.model`). Net effect: a Gemini-compat host emitting `INVALID_ARGUMENT` without the literal `"unexpected model name format"` is correctly demoted at runtime but falls to the generic bucket at setup — the exact class of confusion #5017 is about.

Make the existing predicate `pub(crate)` and delegate, extending it with what you need here:

```rust
// src/core/observability.rs
-fn is_embedding_model_rejected(lower: &str) -> bool {
-    lower.contains("embedding api error")
-        && lower.contains("(400")
+pub(crate) fn is_embedding_model_rejected(lower: &str) -> bool {
+    // Tolerant of all three emitted status shapes — setup-time verification
+    // also sees `openai embeddings returned HTTP <code>` (#5017).
+    let bad_request = mentions_status(lower, 400) || mentions_status(lower, 422);
+    bad_request
         && (lower.contains("does not exist")
             || lower.contains("does not support embeddings")
+            || lower.contains("not an embedding model")
+            || lower.contains("is not an embedding")
+            || lower.contains("not supported for embeddings")
             || lower.contains("unexpected model name format")
             || (lower.contains("invalid_argument")
                 && lower.contains("batchembedcontentsrequest.model")))
 }
```

then delete `is_embedding_model_incompatible` and call it:

```rust
-} else if is_embedding_model_incompatible(&lower) {
+} else if crate::core::observability::is_embedding_model_rejected(&lower) {
```

If you'd rather not touch the observability module in a fix PR, at minimum add the missing Gemini arm here and leave a `// keep in sync with` comment on both — but one predicate is the right answer.

src/openhuman/embeddings/rpc.rs
```diff
@@ -708,6 +776,59 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
...
+/// The tinyagents adapter wraps these as
+/// `openai embeddings request to <url> failed: <reqwest error>`.
+fn is_embedding_endpoint_unreachable(lower: &str) -> bool {
+    lower.contains("request to") && lower.contains("failed")
+        || lower.contains("connection refused")
+        || lower.contains("error sending request")
+        || lower.contains("error trying to connect")
+        || lower.contains("dns error")
+        || lower.contains("failed to lookup address")
+        || lower.contains("tcp connect error")
```
sanil-23, line 829: **nitpick (x2)**

Precedence: this mixes `&&` and `||` unparenthesized. It parses as intended — `(a && b) || c || …` — but reads as ambiguous:

```rust
-    lower.contains("request to") && lower.contains("failed")
+    (lower.contains("request to") && lower.contains("failed"))
         || lower.contains("connection refused")
```

Breadth: `contains("request to") && contains("failed")` is loose enough to catch a 502/503 whose *body* says "the request to the upstream failed", classifying a reachable host as unreachable and sending the user off to check their base URL. Anchoring on the adapter's actual prefix would tighten it:

```rust
lower.contains("embeddings request to")
```

src/openhuman/embeddings/rpc.rs
```diff
@@ -686,8 +703,59 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
...
+                // Reachable, authenticated embeddings API that rejected the model —
+                // the user pasted a chat/reasoning model (e.g. `gpt-5-mini`) into the
+                // embeddings model field. This is the #5017 reporter's exact case:
+                // the same model works for chat but is not an embeddings model.
+                reject(
+                    "EMBEDDINGS_MODEL_INCOMPATIBLE",
+                    "That model isn't an embeddings model on this endpoint. A chat model \
+                     (the one that works in Chat settings) can't produce embeddings — \
+                     enter an embeddings model id (e.g. text-embedding-3-small, bge-m3), \
+                     then save again.",
```
sanil-23, line 728: **minor — `"does not exist"` misdiagnoses a typo as a chat model**

A 400 `Model text-embedding-3-smal does not exist` is a typo, not a chat model, but this tells the user "A chat model (the one that works in Chat settings) can't produce embeddings". The `/models` override normally rescues that case — but only when the endpoint exposes a served list, and per my other comment that override should be narrowed anyway.

Softening costs nothing and stays correct for both causes:

```rust
-"That model isn't an embeddings model on this endpoint. A chat model \
- (the one that works in Chat settings) can't produce embeddings — \
- enter an embeddings model id (e.g. text-embedding-3-small, bge-m3), \
- then save again.",
+"The endpoint rejected that model id for embeddings. Check the spelling, \
+ and note that a chat model (the one that works in Chat settings) can't \
+ produce embeddings — enter an embeddings model id (e.g. \
+ text-embedding-3-small, bge-m3), then save again.",
```

src/openhuman/embeddings/rpc.rs
```diff
@@ -699,7 +767,7 @@ fn classify_embed_probe(outcome: EmbedProbe) -> Option<RpcOutcome<serde_json::Va
             }
         }
         EmbedProbe::TimedOut => reject(
-            "EMBEDDINGS_VERIFICATION_FAILED",
+            "EMBEDDINGS_ENDPOINT_UNREACHABLE",
```
sanil-23, line 770: **minor — folding `TimedOut` into `_ENDPOINT_UNREACHABLE` merges two causes**

The 10s box around `embedder.embed(...)` is routinely blown by a *reachable* endpoint — LM Studio cold-loading a model into VRAM, or a rate-limited cloud host queueing the request.

The user-visible message is unchanged (it still says "timed out"), so there's no UX regression. But the code is now wrong for that case, and support/telemetry keying on the new `reject_code` field can no longer separate "host is down" from "host is slow". Given the whole PR is about differentiation, this is the one place it goes the other way:

```rust
 EmbedProbe::TimedOut => reject(
-    "EMBEDDINGS_ENDPOINT_UNREACHABLE",
+    // Distinct from `_UNREACHABLE`: the host may be perfectly reachable but
+    // slow (cold model load, queued request). Own code so telemetry and
+    // support can tell "down" from "slow" (#5017).
+    "EMBEDDINGS_VERIFICATION_TIMEOUT",
```

Nice side effect: the frontend shape match handles a brand-new code with no change — a live demonstration that the panel fix works.

src/openhuman/embeddings/rpc.rs
```diff
@@ -1048,24 +1169,205 @@ mod tests {
...
+    /// field. The endpoint IS an embeddings API but rejects the model with a 400;
+    /// before the fix this collapsed into the generic "test embed failed", so the
+    /// user couldn't tell the model wasn't an embeddings model. Now it maps to a
+    /// dedicated, actionable code — across both wire shapes.
+    #[test]
+    fn classify_embed_probe_distinguishes_incompatible_model() {
+        for detail in [
+            r#"openai embeddings returned HTTP 400 Bad Request: {"error":{"message":"gpt-5-mini does not support embeddings"}}"#,
+            r#"Embedding API error (400 Bad Request): {"error":{"message":"Model gpt-5-mini does not exist"}}"#,
+            r#"openai embeddings returned HTTP 400 Bad Request: {"error":"this is not an embedding model"}"#,
```
sanil-23, line 1196: **blocker — these tests are tautological against the real wire format**

Every classifier keys on strings owned by the vendored `tinyagents::harness::embeddings` crate, and every test here asserts against a literal hand-written in this same PR. If tinyagents rephrases (`"returned HTTP 401"` → `"HTTP status 401"`), all five tests keep passing while every failure silently degrades to the generic bucket — i.e. they cannot detect the regression they exist to prevent.

I hit the sharp end of this reviewing: a fresh worktree has no `vendor/tinyagents`, so there is no way to confirm these six shapes are real without the submodule.

The fix is cheap because you already built the tooling — `conformant_custom_endpoint_verifies_and_sends_expected_request` drives a real axum mock through the real provider. Point it at failure responses too, so the assertion runs over the error string tinyagents *actually* produces:

```rust
/// Drives the real provider against a mock returning each failure status, so
/// the classifier is asserted over the error string tinyagents actually
/// produces rather than a literal authored by this test (#5017).
#[tokio::test]
async fn classifies_real_adapter_errors_from_a_mock_endpoint() {
    for (status, body, expected) in [
        (401, r#"{"error":"invalid api key"}"#, "EMBEDDINGS_AUTH_FAILED"),
        (403, r#"{"error":"no access"}"#, "EMBEDDINGS_AUTH_FAILED"),
        (
            400,
            r#"{"error":{"message":"gpt-5-mini does not support embeddings"}}"#,
            "EMBEDDINGS_MODEL_INCOMPATIBLE",
        ),
        (500, r#"{"error":"boom"}"#, "EMBEDDINGS_VERIFICATION_FAILED"),
    ] {
        let base = spawn_failing_embeddings_mock(status, body).await;
        let embedder = create_embedding_provider_with_credentials(
            "custom", "gpt-5-mini", 0, "sk-x", Some(&base),
        )
        .expect("provider builds");
        let outcome = match embedder.embed(&["connection test"]).await {
            Ok(v) => EmbedProbe::Returned(v),
            Err(e) => EmbedProbe::Failed(e.to_string()),
        };
        assert_eq!(
            reject_code(outcome).as_deref(),
            Some(expected),
            "status {status} must classify as {expected}"
        );
    }
}
```

Keep the literal-based tests as cheap documentation of known shapes — they just shouldn't be the only guard.

app/src/components/settings/panels/EmbeddingsPanel.tsx
```diff
@@ -303,23 +309,34 @@ const EmbeddingsPanel = ({ embedded = false }: EmbeddingsPanelProps = {}) => {
...
       ) {
```
sanil-23, line 340: **blocker — this behavior change has no test**

This is the whole point of the frontend half ("fixes a latent *unknown code silently treated as a save*"), and nothing exercises it. Every failure-path test in `__tests__/EmbeddingsPanel.test.tsx` uses one of the three previously allow-listed codes (`_ENDPOINT_NO_API`, `_NO_MODEL_LOADED`, `_VERIFICATION_FAILED`), so all of them pass identically against the *old* allow-list. Nothing would fail if someone reverted this hunk.

Add one test whose entire value is that it fails on `main`:

```tsx
it('surfaces a differentiated backend code with no allow-list entry (#5017)', async () => {
  // Regression: the panel used to allow-list three codes, so any *new* code
  // fell through to the success branch and reported a save that never happened.
  const settings = makeSettings({
    providers: [
      makeProvider('managed', { requires_api_key: false }),
      makeProvider('custom', { requires_api_key: false, requires_endpoint: true }),
    ],
  });
  vi.mocked(loadEmbeddingsSettings).mockResolvedValue(settings);
  vi.mocked(updateEmbeddingsSettings).mockResolvedValue({
    error: 'EMBEDDINGS_AUTH_FAILED',
    message: 'The endpoint rejected the API key (401/403).',
  });

  renderWithProviders(<EmbeddingsPanel />);
  await screen.findByText('Custom');
  fireEvent.click(screen.getByRole('radio', { name: /custom/i }));
  await screen.findByPlaceholderText(/https:\/\/your-endpoint/i);
  fireEvent.change(screen.getByPlaceholderText(/https:\/\/your-endpoint/i), {
    target: { value: 'https://api.example.com/v1' },
  });
  fireEvent.click(screen.getByRole('button', { name: /save.*switch/i }));

  await screen.findByText(/rejected the API key/i);
  // Popup stays open and reload() never ran — the config was NOT saved.
  expect(screen.getByPlaceholderText(/https:\/\/your-endpoint/i)).toBeInTheDocument();
  expect(vi.mocked(loadEmbeddingsSettings)).toHaveBeenCalledTimes(1);
});
```

Worth also covering the inverse — that `EMBEDDINGS_DIMENSION_CHANGE_REQUIRES_WIPE` still reaches the confirm dialog rather than the new failure branch. That path exists at line 365 but the exclusion at line 339 is what protects it.

sanil-23 (changes requested):
## Review — request changes

Thanks for the thorough investigation write-up; ruling out the reporter's five guesses with a captured request rather than prose is the right way to close #5017. The direction is correct and the policy stays a pure, unit-testable function over `EmbedProbe`. Four things should land before merge.

Answering the three questions this change lives or dies on:

- **Are the branches exhaustive and mutually exclusive?** Within `classify_embed_probe`, yes — the `if / else if` chain is exclusive by construction and specific-before-generic ordering is correct. But the **caller** undoes it: `update_settings` unconditionally overrides any reject with `/models` name-mismatch guidance, so a high-confidence `EMBEDDINGS_AUTH_FAILED` can be reported as a model-name problem. See the inline comment on `rpc.rs`.
- **Does any user-facing message leak a key or endpoint credential?** No. All four new `message` strings are static literals with no interpolation, and the new `tracing::warn!(reject_code)` logs only the sentinel — the security claim in the PR body holds up. (Pre-existing, not this PR: the `detail` passthrough rendered by the panel can carry the endpoint URL out of a reqwest error. Fine for the user's own screen; worth remembering if `detail` ever reaches telemetry.)
- **Is a genuinely unknown failure still surfaced?** Yes. The terminal `else` keeps `EMBEDDINGS_VERIFICATION_FAILED` with `detail`, empty vectors still reject, and the frontend shape match is strictly more conservative than the old allow-list. Nothing is swallowed into a success. `classify_embed_probe_rejects_unclassified_5xx_generically` guards exactly this.

**i18n: clean.** No new frontend literals, and `settings.embeddings.verifyFallback` is present in all 14 locale files (verified). One caveat under Minor below.

**Verification note:** I could not run `cargo test` — a fresh worktree is missing the vendored `tinyagents` submodule (`vendor/tinyagents/Cargo.toml` absent), so the build fails before compiling. The Rust review is static. This is also why finding 4 matters.

---

### Blocking

1. **`rpc.rs` — the `/models` override clobbers the new codes.** Inline.
2. **`EmbeddingsPanel.tsx` — the behavior change has no test.** Inline. This is the latent bug the frontend half exists to fix, and every existing failure-path test uses one of the three previously allow-listed codes, so all of them pass identically against `main`. The checklist claim that it is "covered by the existing panel behavior" is not accurate — coverage counts the line, but no test fails if someone reverts to an allow-list.
3. **`rpc.rs` — `is_embedding_model_incompatible` duplicates `observability::is_embedding_model_rejected` and drifts from it.** Inline.
4. **`rpc.rs` — the classification tests are tautological against the real wire format.** Inline.

### Minor / non-blocking

5. **Folding `TimedOut` into `EMBEDDINGS_ENDPOINT_UNREACHABLE`** loses a distinction this PR exists to create. Inline.
6. **`"does not exist"` misdiagnoses a typo as a chat model.** Inline.
7. **The four new backend messages are permanently untranslated.** `AGENTS.md` requires UI text through `useT()`; these bypass it as backend-emitted strings — an established pattern, and you're upfront about it. But it takes non-English users from 3 untranslated messages to 7, and since `result.message` is always present on these paths the translated `verifyFallback` is now effectively dead for custom-endpoint failures. The codes are stable sentinels, which is exactly what a lookup wants:

```tsx
const CODE_KEYS: Record<string, string> = {
  EMBEDDINGS_AUTH_FAILED: 'settings.embeddings.errAuthFailed',
  EMBEDDINGS_MODEL_INCOMPATIBLE: 'settings.embeddings.errModelIncompatible',
  EMBEDDINGS_ENDPOINT_UNREACHABLE: 'settings.embeddings.errUnreachable',
  EMBEDDINGS_DIMENSION_MISMATCH: 'settings.embeddings.errDimensionMismatch',
};
const key = CODE_KEYS[result.error];
const baseMessage = key
  ? t(key)
  : typeof result.message === 'string'
    ? result.message
    : t('settings.embeddings.verifyFallback');
```

Reasonable to defer to a follow-up — but then say so in the PR body rather than framing "no new i18n keys" as a design win.

### Nitpicks

- `is_embedding_endpoint_unreachable` mixes `&&` and `||` unparenthesized. It parses as intended (`(a && b) || c || …`) but reads as ambiguous — wrap the first clause. Inline comment has more.
- `embed_error_mentions_status`: `contains(&format!("({code}"))` has no digit boundary, so `(400` matches `(4001`. Contrived in practice, but `(401)` / `(401 ` as the pattern costs nothing.
- `classify_embed_probe_distinguishes_incompatible_model` doc comment reads "Issue #5017 — the #5017 reporter's exact case"; drop the duplicate reference.
- `EmbeddingsPanel.tsx` logs the same error code twice ~30 lines apart on one code path; the second adds only "preserving setup popup". Consider folding into the first.

### Questions

1. Is the `/models` override intentional for the new codes, or just not revisited? It's the difference between fully meeting the issue's acceptance criteria and meeting them except behind a public `/models`.
2. Were the six wire-shape strings verified against the current vendored `tinyagents` source, or inferred? I couldn't check (submodule absent). If verified, pinning the tinyagents commit in a comment would make finding 4 much less pressing.

### Looks good

- Policy stays a pure function over a normalized `EmbedProbe`, so all of it is testable without a network — consistent with how #3761/#4056 built it.
- `is_embedding_model_incompatible` gates on 400/422 **plus** a model-rejection phrase rather than status alone, so a bare 400 (oversized input) still falls through. Matches the polarity contract the observability module documents.
- The mock-endpoint regression test asserting the captured `authorization` header and request body is the right way to settle "is the request correct".
- No credential or endpoint content in any new log line or user-facing message.
- i18n parity intact; no new keys needed.
- Frontend allow-list to shape-match is the correct direction (fail closed on unknown codes).

PR #5067, opened 2026-07-20
fix(inference): probe /v1/models for OpenAI-compatible local runtimes (Closes #5053)

### LINKED ISSUE #5053: LM Studio / OpenAI-compatible providers: app calls Ollama-style /api/tags instead of /v1/models, breaking model discovery for nvidia/nemotron-3-nano-omni and similar
## Summary

When a custom OpenAI-compatible provider (e.g. LM Studio serving `nvidia/nemotron-3-nano-omni`) is configured as a BYOK endpoint, the app sends `GET /v1/api/tags` — an Ollama-specific model-listing endpoint — instead of the OpenAI-standard `GET /v1/models`. LM Studio logs the request as an error (`Unexpected endpoint or method`) and returns 200 anyway, but model discovery silently fails and the model is unavailable in the tinyagents module.

## Problem

LM Studio developer logs show:

```
[DEBUG] Received request: GET to /v1/api/tags. Returning 200 anyway
[ERROR] Unexpected endpoint or method. (GET /v1/api/tags). Returning 200 anyway
```

The app is routing model-listing calls through the Ollama discovery path (`/api/tags`) even when the configured provider is OpenAI-compatible (not Ollama). LM Studio gracefully returns 200 but with an empty or incompatible payload — causing the model list to be empty or stale.

Setup that reproduces this:
- LM Studio on Windows
- Model: `nvidia/nemotron-3-nano-omni` (35 GB, READY, Parallel 4)
- Endpoint: `http://localhost:1234/v1` (OpenAI-compatible, not Ollama)

## Root cause

Model discovery likely uses the `OLLAMA_PROVIDER_PREFIX` path unconditionally, or the provider-type check in `factory.rs` misidentifies a localhost OpenAI-compatible endpoint as Ollama. The correct endpoint for OpenAI-compatible servers is `GET /v1/models`, not `GET /api/tags`.

See: `src/openhuman/inference/provider/factory.rs` — provider slug routing logic.

## Steps to reproduce

1. Start LM Studio, load `nvidia/nemotron-3-nano-omni`, enable local server (port 1234).
2. In OpenHuman Connections → add custom BYOK provider: `http://localhost:1234/v1`, model `nvidia/nemotron-3-nano-omni`.
3. Enable the provider and attempt to use it in Chat or via the tinyagents module.
4. Check LM Studio developer logs — observe `GET /v1/api/tags` error.
5. Observe model is not available / inference fails.

## Expected behaviour

- OpenAI-compatible providers are probed with `GET /v1/models` (not `/api/tags`)
- `nvidia/nemotron-3-nano-omni` loads and responds to chat and agent task requests through the tinyagents module
- LM Studio logs show no unexpected endpoint errors

## Fix

In the provider factory, gate the model-discovery call on provider type:
- Ollama provider (`ollama:` prefix or localhost:11434): use `GET /api/tags`
- OpenAI-compatible provider (custom BYOK slug): use `GET /v1/models`

If both are reachable on the same host (LM Studio also serves `/api/tags` optionally), prefer `/v1/models` for non-Ollama slugs to avoid false-positive Ollama detection.

## Acceptance criteria

- [ ] Custom OpenAI-compatible BYOK endpoint uses `GET /v1/models` for model discovery, not `GET /api/tags`
- [ ] Ollama provider continues to use `GET /api/tags` unaffected
- [ ] `nvidia/nemotron-3-nano-omni` via LM Studio is selectable and functional in Chat and the tinyagents module
- [ ] No `Unexpected endpoint or method` errors appear in LM Studio logs when OpenHuman connects
- [ ] Testable end-to-end in the tinyagents module against a local LM Studio instance

## Test target

tinyagents module — verify chat + agent task dispatch works against `nvidia/nemotron-3-nano-omni` on LM Studio before closing.

## Related

- `src/openhuman/inference/provider/factory.rs` — `OLLAMA_PROVIDER_PREFIX`, provider slug routing
- #5017 — Custom OpenAI-compatible embedding endpoint verification failure (same class of provider-type confusion)
- BYOK provider setup (Connections → custom endpoint)

M3gA-Mind:
## Summary

- Local-AI model discovery now selects its probe endpoint by provider **type**, not by "is it localhost": genuine Ollama uses `GET /api/tags`, every OpenAI-compatible runtime uses `GET /v1/models`.
- Fixes custom OpenAI-compatible BYOK endpoints on localhost (e.g. LM Studio at `http://localhost:1234/v1`) that were being probed with the Ollama-only `/api/tags`, producing `GET /v1/api/tags` and an empty model list.
- Adds `model_discovery_api(provider, base_url)` + `endpoint_is_openai_v1(base_url)` selectors and routes local-AI diagnostics through them.
- Adds unit coverage for the endpoint selection and a diagnostics-level regression test.

## Problem

When a custom OpenAI-compatible provider (LM Studio serving `nvidia/nemotron-3-nano-omni`) is configured as a BYOK endpoint on `http://localhost:1234/v1`, the app sends `GET /v1/api/tags` — an Ollama-specific listing endpoint — instead of the OpenAI-standard `GET /v1/models`. LM Studio logs `Unexpected endpoint or method (GET /v1/api/tags)`, returns 200 with an incompatible payload, and model discovery silently fails so the model is unavailable in the tinyagents module.

Root cause: `LocalAiService::diagnostics` gated the OpenAI-compatible `/v1/models` probe on `provider_from_config(config) == LmStudio`. But `provider_from_config` collapses **everything** that isn't literally `lm_studio` onto `Ollama` — including OMLX, `local-openai`, and any custom BYOK endpoint whose provider tag defaulted to `ollama`. Those all fell through to the Ollama branch and were probed at `<base>/api/tags`. With an OpenAI `/v1` base that yields `GET /v1/api/tags`. (The `inference_list_models` RPC path already probed `/models` correctly; the break was in the diagnostics discovery path that populates the selectable-model list.)

## Solution

- New `ModelDiscoveryApi` selector in `src/openhuman/inference/local/provider.rs`:
  - `endpoint_is_openai_v1(base_url)` — true when the URL path is the OpenAI `/v1` root. A genuine Ollama base is host-rooted (`http://localhost:11434`) with no `/v1`, so this separates the two API shapes by **endpoint type**, not by "is it localhost" (the conflation that misidentified LM Studio as Ollama).
  - `model_discovery_api(provider, base_url)` — returns `OllamaTags` only for a genuine `ollama` runtime on a host-rooted base; returns `OpenAiModels` for an explicit `lm_studio`/`omlx` slug **or** any OpenAI `/v1` endpoint. The `/v1` clause rescues a custom BYOK localhost endpoint whose provider tag still defaults to `ollama`.
- `diagnostics()` now routes any OpenAI-compatible `/v1` local endpoint to the existing `/v1/models` diagnostics path before it can reach the Ollama `/api/tags` branch. Genuine Ollama (host-rooted base) is unaffected.
- Debug logging added on the reroute (no PII).

Design notes / tradeoffs: the fix keys on the OpenAI `/v1` endpoint type in addition to the provider slug, because the saved `provider` tag is an unreliable signal (it defaults to `ollama`). This also incidentally fixes the same misrouting for OMLX/`local-openai` runtimes. Ollama's own OpenAI-compatible `/v1/models` works too, so even a hand-edited Ollama base ending in `/v1` degrades gracefully.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/ci-lite.yml`](../.github/workflows/ci-lite.yml). Changed lines are the two selectors + the diagnostics reroute, all exercised by the new unit and regression tests.
- [x] Coverage matrix updated — `N/A: behaviour-only change` (no feature row added/removed/renamed in `docs/TEST-COVERAGE-MATRIX.md`).
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no matrix feature ID applies to this internal routing fix`.
- [x] No new external network dependencies introduced (mock backend / in-test axum mock server used; no live network).
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: does not change a release-cut surface`.

## Impact

- Runtime/platform impact: desktop (and any core build) that uses a local OpenAI-compatible runtime (LM Studio, OMLX, `local-openai`, custom BYOK). No mobile/web/CLI surface change.
- Performance: none (same single probe, correct path). Security/migration/compatibility: none — genuine Ollama behaviour is byte-identical; only OpenAI-compatible `/v1` endpoints change from `/api/tags` to `/v1/models`.

## Related

- Closes: #5053
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A (GitHub-issue-driven, not Linear)
- URL: N/A

### Commit & Branch

- Branch: `fix/GH-5053-openai-compatible-model-discovery`
- Commit SHA: `117103171`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: no `app/src` (frontend) files changed.
- [x] `pnpm typecheck` — N/A: no TypeScript changed (Rust-only change).
- [x] Focused tests: `cargo test --lib -- model_discovery_api endpoint_is_openai_v1 diagnostics_openai_compatible_v1_endpoint` → 4 passed.
- [x] Rust fmt/check (if changed): `cargo fmt` on touched files; `GGML_NATIVE=OFF cargo check --lib` clean.
- [x] Tauri fmt/check (if changed): N/A: no `app/src-tauri` files changed.

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: OpenAI-compatible local runtimes (LM Studio/OMLX/`local-openai`/custom BYOK on a `/v1` base) are probed for models at `/v1/models` instead of `/api/tags`.
- User-visible effect: local models served by LM Studio (e.g. `nvidia/nemotron-3-nano-omni`) now appear as selectable and no `Unexpected endpoint` errors are logged.

### Parity Contract

- Legacy behavior preserved: genuine Ollama (host-rooted base, `provider = ollama`) still uses `/api/tags` — unchanged.
- Guard/fallback/dispatch parity checks: `model_discovery_api` returns `OllamaTags` for `ollama`/empty provider on a non-`/v1` base (covered by `model_discovery_api_uses_tags_for_genuine_ollama`).

- Duplicate PR(s): none
- Canonical PR: this PR
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local runtime detection for OpenAI-compatible `/v1` endpoints.
  * Prevented OpenAI-compatible runtimes from being queried with Ollama-specific model APIs.
  * Improved diagnostics and installed-model detection for compatible local providers.

* **Tests**
  * Added coverage for endpoint detection, provider routing, and model discovery.
  * Added regression coverage for OpenAI-compatible endpoints using an Ollama-style provider setting.

### DIFF
--- a/src/openhuman/inference/local/provider.rs
+++ b/src/openhuman/inference/local/provider.rs
@@ -48,6 +48,64 @@ pub(crate) fn provider_from_config(config: &Config) -> LocalAiProvider {
     }
 }
 
+/// How a local runtime exposes its installed-model catalog.
+///
+/// Ollama serves a native `GET /api/tags` listing; every OpenAI-compatible
+/// runtime (LM Studio, OMLX, `local-openai`, and any custom BYOK endpoint that
+/// speaks the OpenAI `/v1` surface) serves `GET /v1/models`. Sending an Ollama
+/// probe to an OpenAI-compatible server produces `GET /v1/api/tags`, which
+/// LM Studio logs as `Unexpected endpoint or method` and answers with an empty
+/// catalog — so model discovery silently fails and the model never appears as
+/// selectable (GH #5053).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ModelDiscoveryApi {
+    /// Ollama-native `/api/tags`.
+    OllamaTags,
+    /// OpenAI-compatible `/v1/models`.
+    OpenAiModels,
+}
+
+/// True when `base_url`'s path is the OpenAI-compatible `/v1` root — the
+/// canonical marker of an OpenAI-style server (LM Studio, OMLX, LiteLLM, …).
+///
+/// A genuine Ollama base URL is host-rooted (`http://localhost:11434`) with no
+/// `/v1` path segment, so this cleanly separates the two API shapes by
+/// **endpoint type**, not by "is it localhost" (which would misidentify a
+/// localhost LM Studio server as Ollama — the #5053 conflation).
+pub(crate) fn endpoint_is_openai_v1(base_url: &str) -> bool {
+    if let Ok(url) = reqwest::Url::parse(base_url.trim()) {
+        let path = url.path().trim_end_matches('/').to_ascii_lowercase();
+        return path.ends_with("/v1");
+    }
+    // Fall back to a lexical check when the URL doesn't parse cleanly.
+    base_url
+        .trim()
+        .trim_end_matches('/')
+        .to_ascii_lowercase()
+        .ends_with("/v1")
+}
+
+/// Select the model-discovery API for a local runtime from its provider slug
+/// and base URL — by provider **type**, never by "is it localhost".
+///
+/// Genuine Ollama (`provider = "ollama"` on a host-rooted base) uses
+/// `/api/tags`. Every OpenAI-compatible runtime uses `/v1/models`: an explicit
+/// `lm_studio` / `omlx` slug, OR any endpoint whose path is the OpenAI `/v1`
+/// root. The `/v1` endpoint clause is what rescues a custom BYOK localhost
+/// endpoint (e.g. LM Studio on `http://localhost:1234/v1`) whose provider tag
+/// still defaults to `ollama`: the endpoint type wins over the fallback slug
+/// (GH #5053).
+pub(crate) fn model_discovery_api(provider: &str, base_url: &str) -> ModelDiscoveryApi {
+    if endpoint_is_openai_v1(base_url) {
+        return ModelDiscoveryApi::OpenAiModels;
+    }
+    match normalize_provider(provider).as_str() {
+        "ollama" => ModelDiscoveryApi::OllamaTags,
+        // lm_studio, omlx, and any other OpenAI-compatible local runtime.
+        _ => ModelDiscoveryApi::OpenAiModels,
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -70,5 +128,54 @@ mod tests {
         assert_eq!(normalize_provider("omlx"), "omlx");
         assert_eq!(normalize_provider("omlx-server"), "omlx");
         assert_eq!(normalize_provider("OMLX"), "omlx");
+    }
+
+    #[test]
+    fn endpoint_is_openai_v1_detects_v1_root() {
+        assert!(endpoint_is_openai_v1("http://localhost:1234/v1"));
+        assert!(endpoint_is_openai_v1("http://localhost:1234/v1/"));
+        assert!(endpoint_is_openai_v1("https://box.local:1234/openai/v1"));
+        // Genuine Ollama base is host-rooted with no /v1 path.
+        assert!(!endpoint_is_openai_v1("http://localhost:11434"));
+        assert!(!endpoint_is_openai_v1("http://localhost:11434/"));
+        // A `/v1` embedded mid-path is not the OpenAI root.
+        assert!(!endpoint_is_openai_v1("http://localhost:11434/v1/models"));
+    }
+
+    #[test]
+    fn model_discovery_api_uses_tags_for_genuine_ollama() {
+        // Ollama slug on its host-rooted native base -> /api/tags.
+        assert_eq!(
+            model_discovery_api("ollama", "http://localhost:11434"),
+            ModelDiscoveryApi::OllamaTags
+        );
+        assert_eq!(
+            model_discovery_api("", "http://localhost:11434"),
+            ModelDiscoveryApi::OllamaTags
+        );
+    }
+
+    #[test]
+    fn model_discovery_api_uses_v1_models_for_openai_compatible() {
+        // Explicit LM Studio / OMLX slugs are OpenAI-compatible.
+        assert_eq!(
+            model_discovery_api("lm_studio", "http://localhost:1234/v1"),
+            ModelDiscoveryApi::OpenAiModels
+        );
+        assert_eq!(
+            model_discovery_api("omlx", "http://localhost:8080/v1"),
+            ModelDiscoveryApi::OpenAiModels
+        );
+        // The #5053 case: a custom BYOK OpenAI-compatible endpoint on localhost
+        // whose provider tag still defaults to `ollama` must NOT be probed with
+        // /api/tags — the `/v1` endpoint type wins.
+        assert_eq!(
+            model_discovery_api("ollama", "http://localhost:1234/v1"),
+            ModelDiscoveryApi::OpenAiModels
+        );
+        assert_eq!(
+            model_discovery_api("custom-byok", "http://localhost:1234/v1"),
+            ModelDiscoveryApi::OpenAiModels
+        );
     }
 }
--- a/src/openhuman/inference/local/service/ollama_admin/diagnostics.rs
+++ b/src/openhuman/inference/local/service/ollama_admin/diagnostics.rs
@@ -7,7 +7,9 @@ use crate::openhuman::inference::local::ollama::{
     ollama_base_url_from_config, OllamaModelShow, OllamaModelTag, OllamaShowRequest,
     OllamaShowResponse, OllamaTagsResponse,
 };
-use crate::openhuman::inference::local::provider::{provider_from_config, LocalAiProvider};
+use crate::openhuman::inference::local::provider::{
+    model_discovery_api, provider_from_config, LocalAiProvider, ModelDiscoveryApi,
+};
 use crate::openhuman::inference::model_ids;
 use crate::openhuman::inference::presets::{self, VisionMode};
 
@@ -23,6 +25,26 @@ impl LocalAiService {
         }
 
         let base_url = ollama_base_url_from_config(config);
+
+        // Route OpenAI-compatible local runtimes to the `/v1/models` probe.
+        // `provider_from_config` collapses everything that is not literally
+        // `lm_studio` onto `Ollama`, so an OMLX / `local-openai` / custom BYOK
+        // endpoint on the OpenAI `/v1` surface (e.g. LM Studio at
+        // `http://localhost:1234/v1`) would otherwise be probed at
+        // `<base>/api/tags` -> `GET /v1/api/tags`, which LM Studio rejects as an
+        // unexpected endpoint and answers with an empty catalog — silently
+        // breaking model discovery (GH #5053). Decide by endpoint *type*, never
+        // by "is it localhost".
+        if model_discovery_api(&config.local_ai.provider, &base_url)
+            == ModelDiscoveryApi::OpenAiModels
+        {
+            log::debug!(
+                "[local_ai] diagnostics: base_url={} is OpenAI-compatible (/v1) — routing to /v1/models discovery instead of Ollama /api/tags",
+                base_url
+            );
+            return self.lm_studio_diagnostics(config).await;
+        }
+
         let healthy = self.ollama_healthy_at(&base_url).await;
         let runner_ok = if healthy {
             self.ollama_runner_ok_at(&base_url).await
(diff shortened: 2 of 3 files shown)


M3gA-Mind:
Synced onto current main (now includes #5070 feature-gate allowlist + #5003/#5025 learning-subscriber decouple). Re-running CI.

M3gA-Mind:
@coderabbitai review

sanil-23 (changes requested):
The mechanism here is right and the root-cause analysis is genuinely excellent — deciding by endpoint type rather than "is it localhost" is the correct call, and it directly answers the fix suggested in #5053. My one blocking concern is scope.

## The fix is discovery-only; inference still misroutes for the same config

`diagnostics.rs:23` is one of 15 sites gated on `provider_from_config(config) == LocalAiProvider::LmStudio`. This PR corrects that gate at the diagnostics leaf, but the chat path is gated on the same expression and is unchanged.

Tracing the exact #5053 config (`provider = "ollama"`, `base_url = "http://localhost:1234/v1"`):

**Discovery** — fixed by this PR:

```
provider_from_config()   -> Ollama         (line 23 gate not taken)
model_discovery_api()    -> OpenAiModels   <- NEW: /v1 endpoint wins
                         -> GET http://localhost:1234/v1/models      OK
```

**Chat** (`public_infer.rs:279`) — untouched:

```
provider_from_config()   -> Ollama         (gate not taken)
                         -> falls through to the Ollama branch
base_url = ollama_base_url_from_config(config)  -> "http://localhost:1234/v1"
                         -> POST {base_url}/api/chat
                         -> POST http://localhost:1234/v1/api/chat    BROKEN
```

`public_infer.rs:341` builds the URL with `format!("{base_url}/api/chat")`, and `ollama_base_url_from_config` preserves the path component (it only rewrites `0.0.0.0`/`[::]` hosts and rejects `ollama.com`), so the `/v1` survives. That is structurally the same malformed URL as the `/v1/api/tags` this PR fixes — same root cause, one function over.

Worth noting the premises are identical: this PR is only necessary *because* the saved slug is `ollama` (a `lm_studio` slug would already route correctly at line 23). So the exact configuration that motivates this PR is the configuration for which chat stays broken. The two can't be separated.

### Resulting user experience

Per #5053, LM Studio returns 200 for unrecognized endpoints (`Unexpected endpoint or method ... Returning 200 anyway`). So `POST /v1/api/chat` won't 404 cleanly — it returns 200 with a body that isn't an `OllamaChatResponse`, landing at `public_infer.rs:367`:

```rust
.map_err(|e| format!("ollama chat response parse failed: {e}"))?
```

Net effect: the model now appears in the picker (this PR working as intended), the user selects it, and chat fails with an error naming *Ollama* for a provider they configured as LM Studio. To be clear this is not a regression — nobody who had working chat loses it — but it moves the failure from setup time to use time and makes it harder to self-diagnose than the current empty-list symptom.

### Blast radius

Remaining sites on the slug gate:

| Site | Behavior for this config |
|---|---|
| `public_infer.rs:279`, `:548` | Chat + inference -> Ollama `/api/chat` |
| `assets.rs:61, 294-337, 522, 680` | Readiness/warning surface computed as Ollama |
| `bootstrap.rs:180, 471` | Bootstrap probe; `lmstudio://` model URI |
| `model_ids.rs:82, 154` | Effective chat-model id resolution |

### Suggested direction

Rather than another inline check, apply the endpoint-type correction at the decision function so every call site inherits it:

```rust
/// Effective local provider - the saved slug corrected by endpoint type.
/// A `/v1` base is OpenAI-compatible regardless of a slug that defaulted
/// to `ollama` (GH #5053).
pub(crate) fn effective_provider_from_config(config: &Config) -> LocalAiProvider {
    match model_discovery_api(&config.local_ai.provider, &lm_studio_base_url(config)) {
        ModelDiscoveryApi::OpenAiModels => LocalAiProvider::LmStudio,
        ModelDiscoveryApi::OllamaTags => LocalAiProvider::Ollama,
    }
}
```

...then migrate the `provider_from_config(config) == LmStudio` call sites. Patching leaves means 15 sites to find individually, and whichever one gets missed reproduces this exact bug.

If that's too large for this PR, that is a completely legitimate call — but then please retarget to `Refs #5053` rather than `Closes`, since the issue's acceptance criteria include "loads and responds to chat and agent task requests through the tinyagents module", and open a follow-up for inference dispatch.

## Secondary: diagnostics payload shape changes for rerouted configs

`diagnostics.rs:476-496` — `lm_studio_diagnostics` returns a materially different report than the Ollama branch: `ollama_running: false`, no `context_requirement`, no `chat_eligibility`/`embedding_eligibility`, no per-model `eligibility`/`chat_capable`/`size`/`modified_at`, and `"provider": "lm_studio"` even for an OMLX/`local-openai`/custom BYOK runtime.

`ModelStatusSection.tsx:492` renders the status pill directly off `diagnostics.ollama_running`, so a working endpoint on this path displays as not running. The memory-layer context gate (`MIN_CONTEXT_TOKENS`, and the TAURI-RUST-4P6 `chat_capable` filter) also silently stops applying.

There is a narrow real regression here: `provider = "omlx"` on a host-rooted base previously took the Ollama branch and got the full report; it now takes the slug arm into `lm_studio_diagnostics` and gets the reduced shape. Worth either calling out in the description or populating `context_requirement` + eligibility from the native `/api/v0/models` data that `lm_studio_context_window_for` already parses.

## Smaller items

- **`diagnostics.rs:27-46`** — the routing decision is evaluated against `ollama_base_url_from_config` (config-first precedence) but delegates to `lm_studio_diagnostics`, which recomputes via `lm_studio_base_url` (env-first: `OPENHUMAN_LM_STUDIO_BASE_URL`, `LM_STUDIO_BASE_URL`, then config). With either env var set you can decide "OpenAI-compatible" from one URL and probe a different host. Resolve the base once and pass it down.

- **`diagnostics.rs:41-44`** — `base_url` is logged raw. Neighbouring logs go through `redact_ollama_base_url` (`ollama.rs:127`) / `redact_url_for_log` precisely because a BYOK base can carry `user:pass@`; AGENTS.md requires never logging secrets. Prefer `tracing::debug!` with structured fields to match the rest of the file too.

- **`provider.rs:75-86`** — `endpoint_is_openai_v1` is stricter than `normalize_lm_studio_base_url` (`lm_studio.rs:94-98`), which deliberately strips `/models` and `/chat/completions` suffixes because users paste them. A base saved as `http://localhost:1234/v1/models` fails detection, takes the Ollama branch, and produces `GET /v1/models/api/tags` — the same bug class. The test at `provider.rs:142` locks this in. If the strictness is deliberate, worth saying so in the doc comment.

- **`ollama_admin_tests.rs:781`** — the new test depends on `lm_studio_base_url`, which reads `OPENHUMAN_LM_STUDIO_BASE_URL`/`LM_STUDIO_BASE_URL` before config, but `inference_test_guard()` (`local/mod.rs:11-15`) is only a mutex and clears nothing. A developer with either var exported gets a confusing failure. The Ollama side solved this with `OllamaEnvGuard`.

## What's good

- Root-cause writeup names the exact collapsing behavior in `provider_from_config` and explains why endpoint type beats the slug — unusually clear.
- Selectors are pure and total; reusing `normalize_provider` keeps alias handling in one place.
- The regression test's mock serves *only* `/v1/models` with no `/api/tags`, so it genuinely fails on the old code rather than passing incidentally.
- Error/timeout handling on the delegated path is sound: bounded 5s probe, no unwraps, no hang path, graceful degradation into `issues[]`. `LmStudioModelsResponse.data` is `#[serde(default)]`, so a conforming-but-empty payload yields an empty list rather than an error.
- Genuine Ollama behavior is untouched — the early return sits ahead of all Ollama work.

## CI note

The 4 cancelled checks are not failures — they are duplicate `PR Quality (soft)` runs from the same trigger on `2c220df`, cancelled by the concurrency group, with a sibling run (`29755301486`) green. Head commit is 13 SUCCESS / 6 SKIPPED / 0 FAILURE. No re-run needed.

---

Requesting changes on the scope question only. Happy to approve as-is under `Refs #5053` with a follow-up issue for inference dispatch, if you'd prefer to land the discovery fix now.

Reviewed with [Claude Code](https://claude.ai/code).

Issue #3513, opened 2026-06-08
[Meeting Assistant] PR-7: In-call approval + execution + speak (Phase 2)


graycyrus:
**Part of #3505** · **PR-7** · Phase 2 · Depends on PR-6 · Backend-blocked: **YES** (`bot:speak`)

Approve and execute in-call requests, with an approval model that works when the user isn't looking at the screen.

### Scope
- **Dual-channel approval** for in-call tool execution (per the "always require approval" decision):
  1. Post `ApprovalRequestCard` in the meeting thread (visual, for later).
  2. Emit `bot:speak { text: "I'd like to <action>. Say 'approve' to confirm." }`.
  - First response wins: **voice** (next `bot:in_call_request` whose `commandText` parses via existing `parse_approval_reply`, `approval/gate.rs`) **or** the **thread card** (`approval_decide`). Timeout: 2 min (shorter than the standard 10-min TTL).
  - On approve → `execute_dispatch`, post result in thread, optional `bot:speak` confirmation. On deny/timeout → post denial + `bot:speak` "Okay, I won't do that."
- `approval/gate.rs`: add task-local `InCallApprovalContext { meeting_id, correlation_id }` (mirrors existing `APPROVAL_CHAT_CONTEXT`); when set, `intercept()` emits `bot:speak` alongside the card and listens for a voice decision.
- Example end-to-end: "schedule a follow-up in 40 min" → `GOOGLECALENDAR_CREATE_EVENT` → approval → event created → link posted to thread + verbal confirm; allow inviting attendees.
- Frontend: `backendMeetSlice` `pendingInCallAction`; show "Waiting for approval: <action>" in `MeetingBotsCard.tsx`.

### Contracts
- Reuses `openhuman.approval_decide`; outbound `bot:speak` (from PR-6).

### Tests
- Rust: voice-approve path, card-approve path, deny, timeout. Vitest: pending in-call action state.

### Acceptance
- An in-call request can be approved by voice or thread card, executes the tool, posts the result, and (optionally) confirms verbally — never executing without approval.

PR #5112, opened 2026-07-22
fix(flows): shorter approval TTL for copilot live-run parks

graycyrus:
## Summary

`flows_build`'s streaming (copilot pane) path already scopes `APPROVAL_CHAT_CONTEXT` + `AgentTurnOrigin::WebChat` (PR #5090) so `run_flow` / `resume_flow_run` park for a real human approval instead of auto-allowing. Those parks were using the gate's full 10-minute `DEFAULT_APPROVAL_TTL` (`src/openhuman/approval/gate.rs:65`) — a long time to leave a live Slack/Gmail/HTTP node waiting on a copilot pane the user may have already navigated away from.

This reuses the existing "shorten a specific park" mechanism already established for live meetings (`IN_CALL_APPROVAL_TTL` / `APPROVAL_IN_CALL_CONTEXT`, issue #3513) rather than adding a parallel path:

- New `COPILOT_APPROVAL_TTL` constant (180s / 3 min) in `src/openhuman/approval/gate.rs`.
- New `APPROVAL_COPILOT_STREAM_CONTEXT` task-local, scoped by `flows::ops::flows_build`'s streaming branch alongside the existing `APPROVAL_CHAT_CONTEXT` scope (`src/openhuman/flows/ops.rs`).
- The gate reads the new task-local exactly like it already reads `APPROVAL_IN_CALL_CONTEXT`, and clamps the park window through a new `ApprovalGate::resolve_park_ttl` helper — extracted from the existing in-call min/clamp logic so the clamp-selection is unit testable without actually parking a future and waiting out a multi-minute TTL.
- `DEFAULT_APPROVAL_TTL` itself is untouched, and no new global constant replaces it.

**Only `flows_build`'s streaming branch scopes the new context**, so main-chat and every other approval path (external channel, in-call meetings, workflow automation, CLI, composio's own `park_bound`) keep their existing behavior unmodified.

## Test plan

- [x] `GGML_NATIVE=OFF cargo check --manifest-path Cargo.toml` — passes.
- [x] `GGML_NATIVE=OFF cargo test --lib --manifest-path Cargo.toml approval::gate` — 47 passed (including 5 new `resolve_park_ttl_tests` covering: default park unclamped, copilot clamp shortens a 10-minute default, in-call clamp unaffected by the new flag, a clamp never extends an already-shorter boot-time TTL, and the tighter-of-two-clamps case).
- [x] `GGML_NATIVE=OFF cargo test --lib --manifest-path Cargo.toml openhuman::flows::ops` — 183 passed, no regressions.
- [x] `cargo fmt --check` on changed files — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Approval requests generated during copilot streaming now expire based on a dedicated, shorter TTL (about 3 minutes).
  * Approval TTL clamping now applies correctly across copilot-stream, in-call, and default scenarios, without extending the gate’s own effective TTL—even with debug overrides.
  * Stored approval expiry (`expires_at`) is now aligned with the actual clamped wait duration to reduce stale waits.
  * Added targeted test coverage for TTL resolution and copilot-stream persistence behavior.

### DIFF
--- a/src/openhuman/approval/gate.rs
+++ b/src/openhuman/approval/gate.rs
@@ -825,8 +825,18 @@ impl ApprovalGate {
 
         let request_id = uuid::Uuid::new_v4().to_string();
         let now = chrono::Utc::now();
-        let expires_at =
-            Some(now + chrono::Duration::from_std(self.effective_ttl()).unwrap_or_default());
+        // Resolve the clamped park TTL up front so the persisted `expires_at`
+        // and the actual wait below (see `resolve_park_ttl` further down)
+        // use the same value — see `Self::resolve_park_ttl` and the
+        // COPILOT_APPROVAL_TTL / IN_CALL_APPROVAL_TTL clamps. Computing this
+        // only after persisting the pending row let a copilot-streaming park
+        // advertise the old 10-minute `expires_at` while only actually
+        // waiting 180s, so a core restart or an `expire_stale` sweep mid-park
+        // could leave the row "actionable" for the wrong window (CodeRabbit
+        // + Codex review on PR #5112).
+        let effective_ttl =
+            Self::resolve_park_ttl(self.effective_ttl(), in_call_ctx.is_some(), copilot_stream);
+        let expires_at = Some(now + chrono::Duration::from_std(effective_ttl).unwrap_or_default());
 
         // Correlation context (flow-approval-surface, PR2): a Workflow-origin
         // park carries the flow id on the origin itself, but not the run id —
@@ -972,9 +982,11 @@ impl ApprovalGate {
 
         // Live meetings and copilot-streaming flows_build runs each get a
         // clamped park window — see IN_CALL_APPROVAL_TTL / COPILOT_APPROVAL_TTL
-        // and `Self::resolve_park_ttl`. `effective_ttl()` applies the
-        // debug-only env override; the clamp is applied on top so a longer
-        // override can't extend either park past its clamp.
+        // and `Self::resolve_park_ttl`. `effective_ttl` was resolved above
+        // (before `expires_at` was built) so the persisted expiry and this
+        // wait use the identical clamped duration; `effective_ttl()` applies
+        // the debug-only env override, and the clamp is applied on top so a
+        // longer override can't extend either park past its clamp.
         if copilot_stream {
             tracing::debug!(
                 tool = tool_name,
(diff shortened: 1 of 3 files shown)


graycyrus:
Review feedback addressed in `1f1487ac8` (the fix rides on top of the TTL change). Thanks @coderabbitai — the copilot approval TTL still only affects the `flows_build` streaming path via the `APPROVAL_COPILOT_STREAM_CONTEXT` task-local (`resolve_park_ttl`); main-chat and all other paths keep the 10-min `DEFAULT_APPROVAL_TTL`. cargo check + fmt clean.

graycyrus:
Addressed in `1f1487ac8` (pushed):

- **@coderabbitai / @chatgpt-codex-connector — persist the clamped TTL in `expires_at`** (`gate.rs`): confirmed valid — `expires_at` was being built from `self.effective_ttl()` before the copilot/in-call clamp was resolved, so a copilot-streaming park advertised the old 10-minute expiry in the DB while only actually waiting 180s. Fixed by hoisting `Self::resolve_park_ttl(...)` up to right after `copilot_stream`/`in_call_ctx` are known (both were already available before the `expires_at` line) and reusing that single resolved value for both the persisted `expires_at` and the later `tokio::time::timeout` wait. Removed the now-redundant second `resolve_park_ttl` call further down.
- **@coderabbitai — add an integration regression test for the scoped path**: added `copilot_streaming_park_persists_the_clamped_expiry`, which builds a gate with the full `DEFAULT_APPROVAL_TTL` boot TTL (not the 2s `test_gate()` default, which is already shorter than either clamp and would make the assertion vacuous), scopes `APPROVAL_COPILOT_STREAM_CONTEXT` alongside `APPROVAL_CHAT_CONTEXT` + a `WebChat` origin exactly like `flows::ops::flows_build` does in production, inspects the persisted `expires_at` on the pending row to confirm it's clamped to `COPILOT_APPROVAL_TTL` (not the 10-minute default), then resolves the request and confirms `GateOutcome::Allow`.

Verified: `GGML_NATIVE=OFF cargo check --manifest-path Cargo.toml`, `GGML_NATIVE=OFF cargo test --lib --manifest-path Cargo.toml approval::gate` (48 passed, including the new test), `cargo fmt --check` on the 3 changed files — all clean. Also re-confirmed the safety property: `APPROVAL_COPILOT_STREAM_CONTEXT` is scoped at exactly one call site (`flows::ops::flows_build`'s streaming branch), nested inside the existing `APPROVAL_CHAT_CONTEXT` scope — main-chat and every other approval path are unaffected.

Issue #5106, opened 2026-07-22
Share/pool the Node and Python runtime processes across agents and skill runs


senamakel:
## Problem

The new `library-profile skill-run` benchmark (feat/tauri-resource-profiler) measures the **true process-tree cost** of a skill executing on a language runtime: a single JS skill step spawns a `node` child at **~72-75 MB RSS**, taking the tree total to ~121 MB against ~51 MB for the core process alone. Python is expected to be in the same range or higher.

Today each execution gets its own interpreter child via `NodeBootstrap`/the python runtime path. At the opencompany deployment target — **100-1000 live agents in a 2 GB RAM / 2 vCPU server** — this is the single biggest budget breaker: ~25 concurrent node children alone exhaust the whole box, while the core's own marginal cost is only ~0.8-1.8 MiB per agent (fleet/subagent-storm benchmarks).

## Proposal

Share the language runtimes instead of spawning per run:

- **Pooled long-lived workers**: a small, bounded pool of node/python worker processes (1-2 each by default, configurable) executing jobs for all agents/skill runs, instead of one child per execution.
- **Per-job isolation** within a worker via the existing sandbox/jail layers where required; recycle a worker after N jobs or on contamination-sensitive workloads.
- **Idle reaping**: workers exit after an idle TTL so an idle fleet pays zero interpreter RSS.
- **Queue + backpressure**: concurrent skill executions beyond pool size queue rather than fork; surfacing wait time in run logs.

## Definition of done

- `runtime_node` / `runtime_python` (and `skill_runtime` / agent `node_exec`/python tools) route through the shared pool.
- `library-profile skill-run` extended with a concurrency knob (K parallel skill runs): tree RSS must grow by ~one pooled worker, not K interpreters.
- Pool size, TTL, and recycle policy configurable via the TOML config schema.
- Regression-gated by the profiling suite (`scripts/profile/`).

## References

- `docs/library-benchmarking.md`, `docs/resource-profiling-session-2026-07-21.md` (benchmark environment + budget math)
- `src/openhuman/proc_metrics/tree.rs` (process-tree sampler used for the measurement)
- 2 GB / 2 vCPU budget: fleet benchmark projects ~1.8 GiB at 1000 in-process agents — interpreter children are the piece that doesn't fit.

PR #5123, opened 2026-07-22
fix(agents): fail clearly instead of silently dropping spawn_async_subagent results in thread-less contexts (B40)

graycyrus:
## Summary

- Stop `spawn_async_subagent` from **silently discarding** its result when there's no chat thread to deliver into (flow agent nodes, CLI, cron).
- Add a clear error pointing the caller to synchronous delegation, plus a doc note that flow nodes support sync delegation only.
- Make that recommendation actually work: `spawn_subagent` now self-heals to blocking dispatch when there is no delivery thread, and the subconscious agent advertises it instead of the async tool.

## Problem

A flow `agent` node (or CLI/cron run) executing a delegating agent (orchestrator/subconscious) that calls `spawn_async_subagent` directly has `current_parent()` set but no `current_thread_id()`. The background subagent still runs and produces a real result, but that result hits `background_delivery.rs`'s "dropping headless batch" path and is **silently discarded** — no error surfaced, the caller just sees "Accepted". Silent data loss.

## Solution

- `spawn_async_subagent.rs`: guard right after `parent_thread_id` is read — when `None`, log a grep-friendly `log::warn!` (definition ids only, no prompt/session/user data) and return a `ToolResult::error` explaining that async spawn needs a parent chat thread, directing to `spawn_subagent` with `blocking: true` / a `delegate_*` tool (and to parallel flow nodes for parallel work).
- `spawn_subagent.rs`: **self-heal to blocking dispatch when no delivery thread is bound**, instead of forwarding into the async tool. Mirrors the `has_delivery_thread` fallback `dispatch.rs::dispatch_subagent` already gives the `delegate_*` tools. Without this, the guard's own advice looped back into the guard (`spawn_subagent` defaults to async), and thread-less `spawn_subagent` callers would have regressed from "runs, result dropped" to a hard error.
- `subconscious/`: advertise `spawn_subagent` rather than `spawn_async_subagent`. Subconscious turns run headless — there is no `with_thread_id` anywhere in that domain — so its advertised async delegation was already having every result dropped by `background_delivery`, and would now hit the guard on every call. Updated in `agent/agent.toml`, `agent/prompt.md`, and `profiles/memory.rs`.
- `tinyflows/caps.rs`: doc note on `run_via_harness` — flow agent nodes support synchronous delegation only; model parallel multi-agent work as parallel flow nodes rather than bridging async delivery.
- `dispatch.rs` untouched; no async-in-flows delivery bridge built (deliberately out of scope per architect assessment).

## Behaviour in thread-less contexts (flow / CLI / cron)

The guard is deliberately narrow — it only catches the case where there is no correct answer to give. Note that for a direct `spawn_async_subagent` call the sub-agent no longer runs at all, so any side effects it used to produce (files written, messages sent) no longer happen. That is the intended outcome: fire-and-forget-for-side-effects inside a flow node is the anti-pattern being closed.

| Caller | Before | After |
|---|---|---|
| `spawn_async_subagent` (direct) | runs, result dropped | does not run, clear error |
| `spawn_subagent` (default async) | runs, result dropped | runs inline, result returned |
| `delegate_*` | runs inline, result returned | unchanged |
| subconscious delegation | runs, result dropped | runs inline, result returned |

No behaviour change for the normal chat path, where `current_thread_id()` is set.

## Drive-by fixes (pre-existing breaks on `main`)

Both block this branch's Rust lane from compiling/passing, neither is caused by this change:

- `tools/impl/system/node_exec.rs` — #5106 gave `NodeExecTool::new` two more params but left a test call site at 3 args, breaking `cargo test --lib` compilation for the whole crate.
- `spawn_parallel_agents_tests.rs::rejects_more_tasks_than_parent_parallel_limit` — missing `init_global_builtins()`; the parallel-limit check now runs after the registry lookup, so the test only passed when another test happened to initialise the global first.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `errors_clearly_when_no_parent_thread_for_delivery` asserts the guard fires (and that the message names `blocking: true`); `guard_does_not_fire_when_parent_thread_is_bound` asserts the call gets past the guard with a thread bound; `async_default_self_heals_to_blocking_without_delivery_thread` pins that thread-less `spawn_subagent` never reaches the async tool.
- [x] **Diff coverage ≥ 80%** — the new guard branch, the self-heal branch, and their docs are covered by the added tests.
- [x] Coverage matrix updated — `N/A: behaviour-only guard, no new feature row`.
- [x] All affected feature IDs listed under `## Related` — `N/A: no matrix feature affected`.
- [x] No new external network dependencies introduced — `N/A: no network code`.
- [x] Manual smoke checklist updated — `N/A: does not touch release-cut surfaces`.
- [x] Linked issue closed via `Closes #NNN` — `N/A: tracker item B40, no GitHub issue`.

## Related

- Tracker B40 (flows_workflow_bugs.md). No GitHub issue.

## Summary by CodeRabbit

* **Behavior Improvements**
  * Delegated tasks now complete inline and return results reliably when background delivery is unavailable.
  * Background delegation without a chat thread now provides a clear error with synchronous alternatives instead of silently losing results.
  * Delegation behavior is more consistent across flow runs, command-line use, and scheduled tasks.

* **Documentation**
  * Updated delegation guidance to explain when to use synchronous execution and how parallel work should be structured.


src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
```diff
@@ -225,6 +225,37 @@ impl Tool for SpawnAsyncSubagentTool {
...
+        // (`background_delivery`). Outside a chat turn (flow `agent` nodes,
+        // CLI, cron) there is no `current_thread_id()` to deliver into, so
+        // `background_delivery::deliver_batch` logs "dropping headless batch"
+        // and the (possibly real, completed) work is silently discarded — the
+        // caller sees "Accepted" and never learns the result never arrived.
+        // Fail loudly instead: the caller has a synchronous alternative
+        // (`spawn_subagent`, or a `delegate_*` tool, which already self-heals
+        // to blocking dispatch in this situation — see
+        // `dispatch.rs::dispatch_subagent`'s `has_delivery_thread` check).
+        if parent_thread_id.is_none() {
```
graycyrus, line 242: **Behavioral nuance worth a line in the PR body (non-blocking).**

Before this guard, a thread-less `spawn_async_subagent` still *ran* the sub-agent via `tokio::spawn` — only the final result text was dropped in `background_delivery` (`dropping headless batch`). Any side effects (files written, messages sent, memory updates) still happened. With the guard, the sub-agent no longer executes at all in flow/CLI/cron contexts.

This is the correct outcome — fire-and-forget-for-side-effects inside a flow node is exactly the anti-pattern being closed, and the error steers to synchronous `spawn_subagent`/`delegate_*` which *does* run the sub-agent and captures its result. Flagging only so the "no behavior change" framing in the PR body is understood as "no change for the *chat* path"; the thread-less path does change from "runs, result discarded" to "does not run, clear error". No code change requested.

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
```diff
@@ -1280,4 +1324,139 @@ mod tests {
...
+    /// `agent` node, CLI, cron) must get a clear, actionable error instead of
+    /// silently accepting the spawn and later dropping its result in
+    /// `background_delivery`'s "headless batch" path. Sets up a real parent
+    /// turn context (so the call gets past the `current_parent()` /
+    /// allowlist / registry checks) but deliberately does NOT wrap the call
+    /// in `with_thread_id`, so `current_thread_id()` is None — the exact
+    /// condition that used to sail through to `tokio::spawn` and lose the
+    /// result.
+    #[tokio::test]
+    async fn errors_clearly_when_no_parent_thread_for_delivery() {
```
graycyrus, line 1342: **Nit: the test only proves the negative half (non-blocking).**

The PR body says this test asserts the guard fires "while a normal chat turn is unaffected", but the test only exercises the thread-less path — there is no assertion that *with* a `current_thread_id()` bound the call proceeds past the guard. That positive path is genuinely expensive to unit-test (it would drive the full spawn/session machinery), so I would not block on it. Two cheap options if you want the claim backed by a test:

1. Wrap the same call in `with_thread_id("t-parent", ...)` and assert the returned error (if any) is *not* the "no parent chat thread" message — i.e. it got past the guard. This isolates exactly the branch condition without asserting full success.
2. Or just soften the checklist wording to "asserts the guard fires in the thread-less condition" and drop the "normal chat turn is unaffected" clause, since that half is covered structurally by the `is_none()` check.

Either is fine; flagging so the test comment and PR body match what is actually asserted.

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
```diff
@@ -225,6 +225,30 @@ impl Tool for SpawnAsyncSubagentTool {
...
+        if parent_thread_id.is_none() {
+            return Ok(ToolResult::error(
+                "spawn_async_subagent: no parent chat thread available to deliver the result \
+                 into (this looks like a flow node, CLI, or cron run rather than an interactive \
+                 chat turn). Fire-and-forget delegation has nowhere to land its result here and \
+                 the sub-agent's work would be silently discarded. Use synchronous delegation \
+                 instead: `spawn_subagent` or a `delegate_*` tool, and await its result directly. \
+                 For parallel work, model it as parallel flow nodes rather than background \
+                 sub-agents.",
+            ));
```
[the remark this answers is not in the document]
graycyrus: Already addressed in `caadcb6b5` (pushed before this review landed) — the guard logs a constant, grep-friendly `log::warn!` with the `[spawn_async_subagent]` prefix before returning, carrying only the parent/requested agent **definition ids** (no prompt, session id, thread id, or user data):

```
[spawn_async_subagent] refusing fire-and-forget spawn with no delivery thread parent={} requested={} — directing caller to synchronous delegation (flow node / CLI / cron context, background result would be discarded)
```

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
```diff
@@ -225,6 +225,30 @@ impl Tool for SpawnAsyncSubagentTool {
...
+        // (`spawn_subagent`, or a `delegate_*` tool, which already self-heals
+        // to blocking dispatch in this situation — see
+        // `dispatch.rs::dispatch_subagent`'s `has_delivery_thread` check).
+        if parent_thread_id.is_none() {
+            return Ok(ToolResult::error(
+                "spawn_async_subagent: no parent chat thread available to deliver the result \
+                 into (this looks like a flow node, CLI, or cron run rather than an interactive \
+                 chat turn). Fire-and-forget delegation has nowhere to land its result here and \
+                 the sub-agent's work would be silently discarded. Use synchronous delegation \
+                 instead: `spawn_subagent` or a `delegate_*` tool, and await its result directly. \
```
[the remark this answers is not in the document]
graycyrus: Good catch — confirmed and fixed in `4803862de`. `spawn_subagent`'s `if !blocking` branch does re-enter `SpawnAsyncSubagentTool`, so the guard's advice would have looped straight back into the guard.

Fixed on both sides:

1. **The message now names the flag** — "call \`spawn_subagent\` with \`blocking: true\`, or use a \`delegate_*\` tool".
2. **`spawn_subagent` now self-heals** rather than depending on the caller getting the flag right. It checks `has_delivery_thread` and falls through to blocking dispatch instead of forwarding into the async tool — the same fallback `dispatch.rs::dispatch_subagent` already does for `delegate_*`.

(2) also closes a regression this PR would otherwise have shipped: before the guard, a thread-less non-blocking `spawn_subagent` ran the sub-agent and dropped the result; with the guard but without the self-heal it would have become a hard error. Now it runs inline and returns the output.

Pinned by `async_default_self_heals_to_blocking_without_delivery_thread`, which asserts the thread-less path never reaches the async tool or its guard, and `errors_clearly_when_no_parent_thread_for_delivery` now also asserts the message contains `blocking: true`.

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
```diff
@@ -225,6 +225,30 @@ impl Tool for SpawnAsyncSubagentTool {
...
+        // (`background_delivery`). Outside a chat turn (flow `agent` nodes,
+        // CLI, cron) there is no `current_thread_id()` to deliver into, so
+        // `background_delivery::deliver_batch` logs "dropping headless batch"
+        // and the (possibly real, completed) work is silently discarded — the
+        // caller sees "Accepted" and never learns the result never arrived.
+        // Fail loudly instead: the caller has a synchronous alternative
+        // (`spawn_subagent`, or a `delegate_*` tool, which already self-heals
+        // to blocking dispatch in this situation — see
+        // `dispatch.rs::dispatch_subagent`'s `has_delivery_thread` check).
+        if parent_thread_id.is_none() {
```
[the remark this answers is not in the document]
graycyrus: Confirmed and fixed in `4803862de`. Verified the reach: `src/openhuman/subconscious/` contains **zero** `with_thread_id` / `thread_context` bindings, so `current_thread_id()` is `None` for every subconscious turn and the guard would have fired on every advertised delegation.

Took the "replace the async tool in that headless surface" option rather than binding the reserved thread. Binding `subconscious:orchestrator` would make `background_delivery` run a `run_system_turn_on_thread` against a reserved internal thread — a new autonomous-turn path outside the `TrustedAutomation` origin the session sets up, which is a much bigger change than this fix warrants.

So the subconscious now advertises `spawn_subagent`, which (per the sibling comment) self-heals to blocking dispatch with no delivery thread. Updated in all four places that advertised the old tool: `agent/agent.toml` (tool list + rationale comment), `agent/prompt.md`, and `profiles/memory.rs` (module doc, tool catalogue, and the EventDriven/Aggressive `mode_guidance` string).

Worth noting this is a strict improvement rather than a behaviour swap: async delegation from the subconscious was **already** broken — the sub-agent ran and `background_delivery` logged `dropping headless batch`, so the subconscious never saw a result. It now gets the output back inside the same turn and can act on it. All 133 `openhuman::subconscious` tests pass.

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
graycyrus: Agreed on the analysis, and the PR body is updated to say this explicitly rather than leaving "no behavior change" to be read too broadly.

One correction to the conclusion though: the thread-less path does **not** universally become "does not run". That's true for a *direct* `spawn_async_subagent` call, which is what this guard covers. But in `4803862de` the two recommended entry points now self-heal instead of reaching the guard — `delegate_*` already did via `dispatch.rs`, and `spawn_subagent` now does too. Through those, a thread-less delegation goes from "runs, result discarded" to "runs, result returned", which is the outcome we actually want.

So the net matrix in flow/CLI/cron is:

| Caller | Before | After |
|---|---|---|
| `spawn_async_subagent` (direct) | runs, result dropped | does not run, clear error |
| `spawn_subagent` (default async) | runs, result dropped | runs inline, result returned |
| `delegate_*` | runs inline, result returned | unchanged |

The "does not run" case is now the narrow one, and it's the one where there's no correct answer to give.

src/openhuman/agent_orchestration/tools/spawn_async_subagent.rs
graycyrus: Fair — took option 1 in `4803862de`. Added `guard_does_not_fire_when_parent_thread_is_bound`: same call, wrapped in `with_thread_id("t-parent", ...)`, asserting the output does **not** contain `no parent chat thread`. It only proves the call got *past* the branch (a later failure from the spawn/session machinery is fine and expected), which is exactly the isolation you described — the pair now pins that the guard keys on thread presence and nothing else.

Also tightened the negative test to assert the message contains `blocking: true`, so the escape-hatch wording can't silently regress into advice that loops back into the guard (see the Codex thread above).

The checklist wording is updated too, so both halves of the claim are now actually backed.

PR #5134, opened 2026-07-22
fix(flows): make workflow_builder reliably wire file→attachment for 'attachment' asks (B43)

graycyrus:
> ### ⚠️ Review addendum (2nd pass — validated against live runtime evidence) — DO NOT MERGE AS-IS
>
> Live testing of the combined B39+B43 build revealed the decision hook this PR adds prescribes an action that **does not exist**:
>
> 1. **`GMAIL_SEND_EMAIL_WITH_ATTACHMENT` is a phantom slug.** When the copilot grounded it via `get_tool_contract`, the system returned: *"`GMAIL_SEND_EMAIL_WITH_ATTACHMENT` is not a real action in the 'gmail' toolkit's live catalog — use `search_tool_catalog` to find a real slug."* The builder correctly refused it and fell back to `GMAIL_SEND_EMAIL`. So the hook is net-negative: it hard-anchors the few-shot (and the curation + regression test) on a dead slug and sends the builder chasing it.
> 2. **Self-contradiction.** The few-shot tells the builder to `get_tool_contract` this action FIRST, but the prompt's own rule (the `runtime_gate` / "not a real action" guidance) declares that result a **hard stop** requiring a fallback to `search_tool_catalog`. A correctly-behaving builder can never follow this few-shot to completion.
> 3. **Attachments must be backend-fetchable.** Composio sends run backend-side and cannot read local paths (a real run failed with `ENOENT … no such file or directory`). The `public_url` direction here is correct; the send action anchoring it is not.
>
> **Recommendation:** rework the hook to (a) `search_tool_catalog` for the REAL gmail attachment-capable send action and ground it with `get_tool_contract` before wiring, instead of hardcoding `GMAIL_SEND_EMAIL_WITH_ATTACHMENT`, and (b) keep binding a backend-fetchable `public_url`, never a local `file_path`. Also drop the `gmail/tools.rs` curation + its pinning test for the phantom slug (overlaps #5122 — verify #5122 isn't curating the same non-existent action). **No code fix applied in this pass:** the real working attachment action/arg is still TBD and every candidate change cascades into the pinned regression test — this needs the real slug confirmed first. See the two blocking inline comments.
>
> ✅ The earlier review's `storage_upload_file` `visibility: "public"` fix still holds and is correct.

---

## Summary

- Adds an **imperative decision hook** to `workflow_builder/prompt.md` so that when a user asks to email something **"as an attachment"**, the builder wires a real file attachment instead of inlining HTML/text as the email body.
- Adds a **compact few-shot** showing the exact node wiring: `trigger → get_me lookup → research agent → file-producing agent (storage_upload_file → public_url) → GMAIL_SEND_EMAIL_WITH_ATTACHMENT`.
- Tells the builder to call `get_tool_contract` on the attachment action **first** to ground the real attachment arg name (it varies by action) rather than guessing.
- Adds a prompt-regression test (`standing_prompt_teaches_attachment_decision_hook`) that guards the hook phrasing, the anti-inline invariant, and the few-shot wiring.
- Prompt-only fix. No Rust logic changed.

## Problem

B43 (see `my_docs/flows_workflow_bugs.md`): B39 shipped the storage-URL producer pattern (curated `GMAIL_SEND_EMAIL_WITH_ATTACHMENT` + a passive "File-producing agent nodes" paragraph), but in a live test the builder ignored all of it. Asked to email a generated HTML page **"as an attachment"**, it anchored on the pre-B39 mental model ("a code node writes the file to disk, then attach"), decided that was impossible, and fell back to inlining the HTML as the email **body** (`GMAIL_SEND_EMAIL` with `is_html:true`). The passive guidance gets skimmed past — same "shipped-but-ineffective" shape as B37.

> Note for reviewers: B39 (#5122) is still **open** at time of writing, so its passive paragraph is not yet on `main`. This decision hook is deliberately **self-contained** — it carries the full storage-URL producer pattern itself, so it fixes B43 regardless of B39's merge state and complements B39's passive paragraph (defense in depth) once that lands. The bug explicitly asks for both.

## Solution

Added a new `### Attachment requests` subsection in `prompt.md`, placed right after the existing "agent → Gmail send" worked example (where file→send wiring is already discussed):

- **Decision hook (imperative):** if the ask contains "attach"/"attachment"/"as a file", the file MUST travel as a genuine attachment; inlining HTML/text as the body does **NOT** satisfy the request; there is no "write to disk then attach" path, so do not reason toward one and give up.
- **The prescribed wiring:** a `code_executor` agent node uploads via `storage_upload_file` and emits `public_url` in its `output_parser.schema`; the send node is `GMAIL_SEND_EMAIL_WITH_ATTACHMENT` (not `GMAIL_SEND_EMAIL`); `get_tool_contract` grounds the real attachment arg; the arg binds to `=nodes.<producer>.item.json.public_url`.
- **Few-shot:** a tight, runnable graph demonstrating exactly that chain.

All existing invariants (propose-only persistence, B27/B28 guarantees, self-DM resolution, minimal-graph rule) are untouched. No em dashes added per project style.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per Testing Strategy — added `standing_prompt_teaches_attachment_decision_hook` (positive assertions on the hook phrasing + few-shot; the anti-inline "does NOT satisfy" invariant is the failure-mode guard)
- [x] **Diff coverage ≥ 80%** — changed lines are the prompt.md resource (asserted by the new test) + the test itself; `cargo test -p openhuman --lib workflow_builder::` green (19 passed)
- [x] Coverage matrix updated — `N/A: behaviour-only prompt change, no new feature row`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix feature IDs affected`
- [x] No new external network dependencies introduced — `N/A: prompt/text only`
- [x] Linked issue closed via `Closes #NNN` — `N/A: tracked in my_docs/flows_workflow_bugs.md (B43), no GitHub issue`

## Impact

- Desktop/CLI: workflow_builder copilot only. No runtime, migration, or security impact. Prompt-resource text change bundled via `tauri.conf.json` resources and read core-side.

## Related

- Closes: N/A (tracked as B43 in `my_docs/flows_workflow_bugs.md`)
- Follow-up PR(s)/TODOs: complements #5122 (B39) once merged

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/flows-attachment-prompt
- Commit SHA: facd6e6f3de10e54da2b5539b71210471705e446

### Validation Run

- [x] `pnpm typecheck` — N/A (no TS change)
- [x] Focused tests: `GGML_NATIVE=OFF cargo test -p openhuman --lib workflow_builder::` — 19 passed, 0 failed
- [x] Rust fmt/check (if changed): `cargo fmt -p openhuman --check` clean

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: the builder reliably wires a real file attachment when the user asks for "attachment", instead of inlining HTML as the email body.
- User-visible effect: "email it to me as an attachment" produces an email with the file attached.

## Summary by CodeRabbit

* **New Features**
  * Improved workflow builder guidance to handle “attach/attachment/as a file” requests as true email attachments.
  * Added an explicit decision flow to prevent fallback to inline HTML/text as the email body.
  * Provided a clear end-to-end example for generating, uploading, and attaching files.
* **Bug Fixes**
  * Ensured attachment-send actions are treated as write/send operations via a curated Gmail tool entry.
* **Tests**
  * Added unit tests validating the attachment decision behavior and the presence/scope of the attachment email tool.

### DIFF
--- a/src/openhuman/flows/agents/workflow_builder/prompt.md
+++ b/src/openhuman/flows/agents/workflow_builder/prompt.md
@@ -564,9 +564,11 @@ portable file handle is a **storage URL** (or file id), wired like this:
 
 1. The node that PRODUCES the file is an `agent` node (usually
    `agent_ref: "code_executor"` for a generated HTML, CSV, or PDF page). It
-   uploads the file with the **`storage_upload_file`** tool and emits the
-   resulting **`public_url`** (or `file_id`) as a field in its
-   `config.output_parser.schema`, so a downstream node can bind it.
+   uploads the file with the **`storage_upload_file`** tool **as a public file**
+   (`visibility: "public"`, because `storage_upload_file` defaults to private
+   and a private upload has no `public_url`) and emits the resulting
+   **`public_url`** as a field in its `config.output_parser.schema`, so a
+   downstream node can bind it.
 2. The send node is a `tool_call` on the toolkit's ATTACHMENT send action. For
    Gmail that is **`GMAIL_SEND_EMAIL_WITH_ATTACHMENT`**, not `GMAIL_SEND_EMAIL`.
    Call **`get_tool_contract` on that action FIRST** to read the real
@@ -595,7 +597,7 @@ summary page, and email it to me as an attachment":**
   { "id": "make_page", "kind": "agent", "config": {
       "agent_ref": "code_executor",
       "input_context": "=nodes.research.item.json.summary",
-      "prompt": "Build a self-contained HTML page from the summary above, upload it with storage_upload_file, and return its public_url.",
+      "prompt": "Build a self-contained HTML page from the summary above, upload it as a public file with storage_upload_file (visibility public), and return its public_url.",
       "output_parser": { "schema": { "type": "object", "required": ["public_url"],
         "properties": { "public_url": { "type": "string" } } } } } },
   { "id": "send", "kind": "tool_call", "config": {
--- a/src/openhuman/flows/agents/workflow_builder/prompt.md
+++ b/src/openhuman/flows/agents/workflow_builder/prompt.md
@@ -551,6 +551,75 @@ And without `input_context`, don't reach for a jq expression woven into
 — that's prose, not jq, resolves to `null`, and both the `save_workflow` gate
 and `dry_run_workflow`'s `agent_prompt_nulls` will reject it.
 
+### Attachment requests: wire a real file attachment, never inline the body
+
+**Decision hook (do not skim past this).** If the user's ask contains "attach",
+"attachment", or "as a file" (for example "email the report as an attachment",
+"send me the page as a file"), the file MUST travel as a genuine attachment.
+Inlining HTML or text as the email body (a `GMAIL_SEND_EMAIL` with
+`is_html: true`) does NOT satisfy an attachment request, so never fall back to
+it. There is also no "write the file to disk, then attach it" path in
+tinyflows, so do not reason toward one, hit that dead end, and give up. The
+portable file handle is a **storage URL** (or file id), wired like this:
+
+1. The node that PRODUCES the file is an `agent` node (usually
+   `agent_ref: "code_executor"` for a generated HTML, CSV, or PDF page). It
+   uploads the file with the **`storage_upload_file`** tool and emits the
+   resulting **`public_url`** (or `file_id`) as a field in its
+   `config.output_parser.schema`, so a downstream node can bind it.
+2. The send node is a `tool_call` on the toolkit's ATTACHMENT send action. For
+   Gmail that is **`GMAIL_SEND_EMAIL_WITH_ATTACHMENT`**, not `GMAIL_SEND_EMAIL`.
+   Call **`get_tool_contract` on that action FIRST** to read the real
+   attachment arg name (it varies by action), and never guess it.
+3. Bind that attachment arg to the producer's URL field, for example
+   `=nodes.make_page.item.json.public_url`.
+
+Only if `get_tool_contract` shows the attachment action needs raw base64
+instead of a URL do you reach for a base64 helper. Verify against the contract
+first; a URL is the default.
+
+**Few-shot (the exact wiring), "research the top AI trends, make an HTML
+summary page, and email it to me as an attachment":**
+
+```json
+{ "nodes": [
+  { "id": "start",  "kind": "trigger",   "config": { "trigger_kind": "manual" } },
+  { "id": "get_me", "kind": "tool_call", "config": {
+      "slug": "GMAIL_GET_PROFILE", "connection_ref": "composio:gmail:<conn_id>", "args": {} } },
+  { "id": "research", "kind": "agent", "config": {
+      "agent_ref": "researcher",
+      "input_context": "=item",
+      "prompt": "Research the top 3 AI trends this week and write a titled summary with sources.",
+      "output_parser": { "schema": { "type": "object", "required": ["summary"],
+        "properties": { "summary": { "type": "string" } } } } } },
+  { "id": "make_page", "kind": "agent", "config": {
+      "agent_ref": "code_executor",
+      "input_context": "=nodes.research.item.json.summary",
+      "prompt": "Build a self-contained HTML page from the summary above, upload it with storage_upload_file, and return its public_url.",
+      "output_parser": { "schema": { "type": "object", "required": ["public_url"],
+        "properties": { "public_url": { "type": "string" } } } } } },
+  { "id": "send", "kind": "tool_call", "config": {
+      "slug": "GMAIL_SEND_EMAIL_WITH_ATTACHMENT",
+      "connection_ref": "composio:gmail:<conn_id>",
+      "args": {
+        "recipient_email": "=nodes.get_me.item.json.data.emailAddress",
+        "subject": "Your weekly AI trends",
+        "body": "The full summary is attached.",
+        "attachment": "=nodes.make_page.item.json.public_url" } } }
+],
+  "edges": [
+    { "from_node": "start",     "to_node": "get_me" },
+    { "from_node": "start",     "to_node": "research" },
+    { "from_node": "research",  "to_node": "make_page" },
+    { "from_node": "make_page", "to_node": "send" },
+    { "from_node": "get_me",    "to_node": "send" }
+  ] }
+```
+
+The `attachment` arg name above is illustrative: read the real one off
+`get_tool_contract` before wiring. The `get_me` lookup resolves "email me" per
+the self-target rules above, so you do not ask the user for their own address.
+
 ### Trigger kinds — which ones actually fire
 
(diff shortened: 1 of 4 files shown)


graycyrus:
@coderabbitai review

graycyrus:
@coderabbitai review

src/openhuman/flows/agents/workflow_builder/prompt.md
```diff
@@ -551,6 +551,77 @@ And without `input_context`, don't reach for a jq expression woven into
...
+
+1. The node that PRODUCES the file is an `agent` node (usually
+   `agent_ref: "code_executor"` for a generated HTML, CSV, or PDF page). It
+   uploads the file with the **`storage_upload_file`** tool **as a public file**
+   (`visibility: "public"`, because `storage_upload_file` defaults to private
+   and a private upload has no `public_url`) and emits the resulting
+   **`public_url`** as a field in its `config.output_parser.schema`, so a
+   downstream node can bind it.
+2. The send node is a `tool_call` on the toolkit's ATTACHMENT send action. For
+   Gmail that is **`GMAIL_SEND_EMAIL_WITH_ATTACHMENT`**, not `GMAIL_SEND_EMAIL`.
```
graycyrus, line 573: 🛑 **BLOCKING — `GMAIL_SEND_EMAIL_WITH_ATTACHMENT` is a phantom slug, and the few-shot is self-contradictory with the prompt's own grounding rule.**

**Runtime evidence (live B39+B43 build, authoritative).** When the copilot grounded this prescription via `get_tool_contract`, the system returned:

> `GMAIL_SEND_EMAIL_WITH_ATTACHMENT` is not a real action in the 'gmail' toolkit's live catalog — use `search_tool_catalog` to find a real slug.

The builder therefore **correctly refused** the prescribed slug and fell back to `GMAIL_SEND_EMAIL`. So this hook is not merely ineffective — it prescribes a **non-existent action**, sending the builder chasing a dead slug, which is worse than the original problem it set out to fix.

**Self-contradiction with this same prompt.** Line 574 tells the builder to call `get_tool_contract` on this action FIRST. But lines 389–394 of this very prompt declare a `runtime_gate` / "not a real action" result a **hard stop** — "go back to `search_tool_catalog` and pick a `featured: true` action instead of wiring the gated one." The runtime evidence above is exactly that hard-stop signal. So the few-shot instructs the builder to wire a slug that the prompt's own grounding rule requires it to abandon. A correctly-behaving builder can never follow this few-shot to completion.

**Second runtime fact:** Composio sends execute backend-side and cannot read local file paths — a real run bound to a local path failed with `[composio:error:upstream_provider] … ENOENT: no such file or directory`. The attachment must be a backend-fetchable `public_url` or base64, never a local `file_path`. The `public_url` direction here is right; the send action anchoring it is not.

**Recommendation (do NOT swap in another hardcoded slug — the real one is unconfirmed):** rework the hook so the builder `search_tool_catalog`s the gmail toolkit for the REAL attachment-capable send action and grounds it with `get_tool_contract` before wiring, and binds a backend-fetchable `public_url` (not a local path). Holding the code change until the real slug is confirmed.

src/openhuman/memory_sync/composio/providers/gmail/tools.rs
```diff
@@ -75,6 +75,16 @@ pub const GMAIL_CURATED: &[CuratedTool] = &[
...
         scope: ToolScope::Write,
     },
+    // The attachment-send variant the workflow builder wires for "email it
+    // as an attachment" asks (B39/B43). Must stay curated: `flow_tool_allowed`
+    // rejects any uncurated Gmail slug at run time, and `get_tool_contract`
+    // flags an uncurated action `runtime_gate` (a hard stop the builder is
+    // told to avoid) — so an uncurated attachment send would make every
+    // attachment flow the B43 few-shot prescribes fail on its real run.
+    CuratedTool {
+        slug: "GMAIL_SEND_EMAIL_WITH_ATTACHMENT",
```
graycyrus, line 85: 🛑 **BLOCKING (same root cause) — curating a phantom slug.**

This curates `GMAIL_SEND_EMAIL_WITH_ATTACHMENT` so `flow_tool_allowed` won't reject it locally. But per the live runtime evidence (see the prompt.md comment), this action **does not exist in Composio's live `gmail` catalog** — `get_tool_contract` returns "not a real action." Curating it locally does not make it exist upstream; it just lets our gate wave through a slug that will fail on the real Composio execute call regardless.

The accompanying regression test (`attachment_send_action_is_curated_as_write`) then pins this phantom slug into the catalog, making the dead action harder to remove later. Recommend dropping this curation (and the pinning test) as part of reworking the hook toward `search_tool_catalog`-grounded selection, rather than hardcoding+curating a specific action. Note this also overlaps the identical curation in #5122 — worth confirming #5122 isn't curating the same non-existent slug.

graycyrus (commented):
**Second review pass — validated against live runtime evidence the first pass missed.**

This PR's central mechanism — the decision hook + few-shot + curation + regression test — all hard-anchor on `GMAIL_SEND_EMAIL_WITH_ATTACHMENT`, which live testing of the combined B39+B43 build shows is a **phantom slug that does not exist in Composio's live `gmail` catalog**. See the two inline comments below. My recommendation is to **rework the hook to ground the real attachment mechanism via `search_tool_catalog` first, rather than hardcode a specific slug** (details in the PR-description addendum). I have not applied a code fix because the real working attachment action/arg is still TBD, and every candidate change cascades into the pinned regression test — this needs a decision on the real slug before editing.

graycyrus:
## Closing — inherited a false premise from #5122

This PR added a decision hook + few-shot telling the workflow builder to wire `GMAIL_SEND_EMAIL_WITH_ATTACHMENT`. Live testing showed **that action does not exist** in Composio's gmail catalog - `get_tool_contract` returns *"not a real action in the 'gmail' toolkit's live catalog"*. So the hook was net-negative: instead of fixing the inline-body fallback, it sent the builder chasing a dead slug.

**It was also self-contradictory.** The prompt tells the builder to `get_tool_contract` an action first, and elsewhere declares a "not a real action" result a hard stop. A correctly-behaving builder therefore could never follow this few-shot to completion.

**The deeper lesson:** prompt guidance is advisory and cannot enforce a capability. We tried it twice; the builder ignored it both times and then invented a third broken variant on its own (`GMAIL_SEND_EMAIL` + a local `file_path`, which fails with ENOENT because the action runs backend-side).

### Validated and carried forward
- The `visibility: "public"` requirement on `storage_upload_file` (a private upload yields a null `public_url`) - a real trap this PR correctly identified.
- Binding a **`public_url`** rather than a local path is the right direction.

Superseded by a single architectural issue: a typed file artifact, a materialization seam at the `tool_call` boundary, and resolving the target action from the **live** catalog rather than a hand-curated list. Nothing here reached `main`.

PR #5135, opened 2026-07-22
fix(flows): live-run reliability — reconcile orphaned/cancelled runs + detach agent-initiated run_flow (B41, B42)

graycyrus:
## Summary

- **B42 — cancelled/timed-out `run_flow` no longer orphans the `flow_runs` row.** A new `RunRowFinalizer` drop-guard reconciles a `running` row to a terminal `interrupted` (with a human reason) if the run future is dropped mid-await; a boot-time sweep reconciles crash/restart orphans. The run-details sidebar shows *"Run interrupted…"* instead of a perpetual blank spinner.
- **B41 — agent-initiated `run_flow` now detaches** instead of blocking past the tinyagents harness's 120s per-tool-call cap. It validates + compile-checks synchronously, then returns `{ run_id, status: "running" }` immediately and runs the flow on a background task; the copilot polls `get_flow_run(run_id)` (which it already does).
- New terminal status `interrupted` threaded through the run type, store, and the run-details / runs-list UI (+ i18n for all 14 locales).
- Both fixes share one `run_flow_body`; the synchronous RPC "Run" and trigger-bus paths keep their existing behavior, now also covered by the drop-guard.

## Problem

Ref `my_docs/flows_workflow_bugs.md` **B41** and **B42**.

- **B42:** `flows_run` bracketed the long engine `.await` between `start_flow_run_row` (inserts `running`) and `finish_flow_run_row` (writes terminal status). When the future was dropped mid-await (harness 120s abort, chat turn end, panic, or a hard crash), `finish_flow_run_row` never ran — the row stayed `status=running`, `error=NULL`, `steps=[]` forever. The sidebar read that as a perpetually-running, blank run; the real failure lived only in the copilot's agent trace.
- **B41:** `RunFlowTool::execute` fully `.await`ed `flows_run` with no internal budget. The 120s cap is the tinyagents harness's default per-tool-call timeout, so any flow whose first real node is a live-research agent node (`web_search` + `web_fetch` + `parallel_research`) exceeded it *deterministically* — agent-initiated test runs of realistic flows could never succeed, and each 120s abort produced a B42 orphan.

## Solution — architect decision for B41: **DETACH** (preferred option)

I chose the strongly-preferred **detach** approach over bumping a timeout, after reading the code:

- The run-scoped budget (`FLOW_RUN_TIMEOUT_SECS`) is already **600s** — the 120s that killed runs is the *harness* per-tool-call timeout, not ours. Bumping it is impossible (it's tinyagents' cap) and blocking the agent's turn for up to 10 min would starve every other tool call in that turn.
- `RunFlowTool` now calls a new `flows_run_detached`, which:
  1. **Validates + compile-checks the flow synchronously** — a broken/incompatible flow still returns an immediate, actionable error to the agent (no silent fire-and-forget).
  2. Inserts the `running` row + publishes `FlowRunStarted`, then **spawns `run_flow_body` on a background task** and returns `{ run_id, status: "running", detached: true }` in well under 120s.
  3. The tool result tells the model to **poll `get_flow_run(run_id)`** and not assume success — which the copilot already does (verified in the live traces cited in the bug doc).
- This mirrors how the UI "Run" control and the trigger bus (`flows::bus::spawn_run`) already fire runs fire-and-forget. The synchronous `flows_run` (RPC "Run" button + trigger dispatch) is preserved unchanged in behavior.

**Why detach is safe here specifically:** it composes with B42. A detached run is no longer cancelled at 120s (the tool returned already), so it reaches its natural terminal write; and if the *process* dies mid-run, the drop-guard (soft shutdown) or the boot sweep (hard crash) always settles the row. Detach without B42 would have moved the orphan window, not closed it — so both ship together.

### B42 implementation

- **Drop-guard (`RunRowFinalizer`)** — held across the engine await inside `run_flow_body`. On `Drop` *without* a prior `disarm()`, it writes `status="interrupted"` + a human reason and preserves any live-observed steps. Every real terminal path (success / failure / timeout / cancel / pause) `disarm()`s it after its own write. Uses a single-task `Cell` flag so the type stays `Send` for `tokio::spawn`.
- **Boot sweep (`sweep_orphaned_running_runs_on_boot`)** — wired into core startup (`services.rs`, before the cron gate so it runs even on cron-disabled/slim configs). Reconciles every `running` row that **no live in-process run owns** (checked via `run_registry::is_in_flight`, so a run this process already started is never swept out from under itself) to `interrupted`, updates the flow summary, emits `FlowRunFinished`, and drops the stale checkpoint.
- **Surfacing (B42c)** — new terminal `interrupted` status carries the `error` reason; the run-details sidebar renders the error banner and a settled (non-spinner) state; runs-list + inspector get an amber "worth a look" pill and i18n label in all 14 locales.
- Success path reordered so the **row is finalized before** the best-effort flow-summary write — a `record_run` failure can no longer leave the row wedged at `running`.

## Impact

- Desktop core (Rust) + React UI. No schema migration — `interrupted` is a new value in the already free-form `status` column.
- No new external network deps. `interrupted` gated correctly in the `flows`-off slim build (`--no-default-features`), which compiles clean.
- Behavior change: agent `run_flow` returns `status:"running"` + `run_id` instead of a completed result; the model is instructed to poll. Cancelled/timed-out/crashed runs now settle to `interrupted` instead of hanging at `running`.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/ci-lite.yml`](../.github/workflows/ci-lite.yml). Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge.
- [x] Coverage matrix updated — added/removed/renamed feature rows in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md) reflect this change (or `N/A: behaviour-only change`) — `N/A: reliability fix to an existing feature (Flows run lifecycle), no new feature row`
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — `N/A: no change to release-cut smoke surfaces`
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — `N/A: tracked in my_docs/flows_workflow_bugs.md (B41/B42), no GitHub issue`

## Related

- Bugs: `my_docs/flows_workflow_bugs.md` → **B41**, **B42**
- Feature area: Flows run lifecycle (`flows::ops::flows_run` / `run_flow` agent tool / run-details sidebar)
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/flows-live-run-reliability`
- Commit SHA: see PR head

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — Rust `cargo fmt --check` clean; Prettier N/A (no manual reformatting)
- [x] `pnpm typecheck`
- [x] Focused tests: `cargo test -p openhuman --lib flows::` (614 pass, incl. 8 new); flows run-status Vitest specs (24 pass)
- [x] Rust fmt/check (if changed): `cargo fmt --check` + `cargo check` (default + `--no-default-features` slim) clean
- [x] Tauri fmt/check (if changed): N/A — no `app/src-tauri` changes

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: agent-initiated `run_flow` detaches and returns `running`+`run_id`; cancelled/timed-out/crashed flow runs reconcile to a terminal `interrupted` with a reason.
- User-visible effect: run-details sidebar shows an interrupted run with its reason instead of a perpetual blank spinner; realistic (live-research) flows are runnable from chat.

## Summary by CodeRabbit

* **New Features**
  * Added an **Interrupted** terminal workflow-run status with amber UI indicators and localized labels across the app.
  * Workflow runs triggered from tools now start **detached** and return **running** immediately with a run ID.

* **Bug Fixes**
  * Prevented workflow runs from remaining stuck in **Running** by reconciling orphaned runs on startup.
  * Reconciles runs to **Interrupted** when execution is dropped mid-await, preserving the interruption reason.
  * Updated polling and live refresh so **Interrupted** stops further updates; canceling an **Interrupted** run is handled as already terminal.


graycyrus:
@coderabbitai review

src/openhuman/flows/ops.rs
```diff
@@ -4296,6 +4563,73 @@ pub async fn sweep_expired_parked_runs(config: &Config) -> usize {
...
+        }
+    };
+    if candidates.is_empty() {
+        return 0;
+    }
+    tracing::debug!(target: "flows", count = candidates.len(), "[flows] boot sweep: examining running rows for orphans");
+
+    let mut swept = 0usize;
+    for (run_id, flow_id) in candidates {
+        if run_registry::is_in_flight(&run_id) {
```
graycyrus, line 4655: **Boot-sweep TOCTOU — no `started_at`/boot-time floor (major, medium confidence)**

The sweep distinguishes a genuine orphan from a live run *solely* via `is_in_flight`, with no age/boot-time guard. But both entry points insert the `running` row (`start_flow_run_row`) **before** `run_registry::register` runs inside `run_flow_body` — so a row is briefly `running`-but-not-yet-in-flight.

In practice this is safe: the sweep `.await`s to completion inside `spawn_cron_service` *before* the cron scheduler starts and before any agent turn, so schedule- and agent-initiated runs can't race it. The only theoretical opening is an external-trigger run (webhook / channel bus) firing `flows_run` in the exact boot window between this `SELECT` and that run's `register()`.

If that ever landed, the sweep would flip a genuinely-live run to `interrupted` **and `drop_checkpoint` it mid-run**. The live run's own terminal write would later correct the status, but the dropped durable checkpoint is unrecoverable.

Cheap hardening that closes the window entirely: capture a boot cutoff at sweep start and only reconcile rows whose `started_at` predates this process — e.g. have `list_running_run_ids` also return `started_at` and filter `< boot_cutoff`. Then the sweep provably never touches a row this process created, independent of registration timing.

Flagging rather than applying — it changes the store query surface, and per the review brief concurrency fixes should be deliberate. Your call on whether the `is_in_flight` guard alone is a strong enough contract given the ordering.

src/openhuman/flows/ops.rs
```diff
@@ -3771,8 +3887,144 @@ pub async fn flows_run(
...
+        }
+        tracing::warn!(
+            target: "flows",
+            flow_id = %self.flow_id,
+            thread_id = %self.thread_id,
+            "[flows] RunRowFinalizer: run future dropped before settling — reconciling orphaned 'running' row to 'interrupted'"
+        );
+        // Preserve whatever steps the live observer already persisted.
+        let observed = current_persisted_steps(&self.config, &self.thread_id);
+        finish_flow_run_row(
```
graycyrus, line 3966: **Blocking rusqlite in `Drop` on a runtime worker thread (minor — noted, no action)**

`RunRowFinalizer::drop` runs two synchronous rusqlite calls (`current_persisted_steps` + `finish_flow_run_row`) directly on whatever tokio worker thread drops the future. This matches the established pattern in this module (the whole run body calls the sync store API without `spawn_blocking`), so it is **not** a regression — noting it because on runtime shutdown this briefly blocks a worker during `Drop`.

Two things that make this correct as written and worth keeping:
- Both helpers swallow their errors (`.ok()` / logged `Result`), never `unwrap`/`panic`, so when the drop happens during a panic unwind there is no double-panic → abort risk. Good.
- `armed` is a `Cell<bool>` (Send, not Sync) so the guard stays `Send` for `tokio::spawn` on the detached path — correct, since the guard never crosses tasks.

No change requested. Documenting the reasoning so a future reader doesn't "fix" the sync-in-Drop into an `async` drop (impossible) or a `spawn_blocking` (would detach the write from the drop and reintroduce the orphan window this guard exists to close).

src/core/runtime/services.rs
```diff
@@ -110,6 +110,24 @@ pub fn spawn_cron_service() {
...
+                // row left at `running` by a prior process (crash/SIGKILL/power
+                // loss — where the in-process drop-guard never ran) to a
+                // terminal `interrupted`, so the run-details sidebar never shows
+                // a perpetual blank spinner. Runs BEFORE the cron gate so it
+                // fires even on cron-disabled/slim configs, and early enough
+                // that no run started by THIS process can yet be an orphan.
+                #[cfg(feature = "flows")]
+                {
+                    let swept =
+                        crate::openhuman::flows::ops::sweep_orphaned_running_runs_on_boot(&config)
```
graycyrus, line 131: **Boot sweep is coupled to the cron service's presence in the `ServiceSet` (minor — design note)**

Placing the sweep before the `if !config.cron.enabled { return; }` gate is right — it runs even when cron is *config-disabled*. But it still only runs if the `cron` background service is actually in the active `ServiceSet`. On `desktop()` that's always true, so the shipped app is fine. On a `headless_api()` / slim `ServiceSet` that omits cron, orphaned `running` rows left by a prior process would never be reconciled at boot.

If flows can run under a cron-less `ServiceSet`, a dedicated flows boot hook (or attaching the sweep to a service that is present whenever `flows` is) would be more robust than piggybacking on cron bootstrap. Minor — desktop, the primary target, is unaffected — but the coupling is a small layering smell (flows reconciliation living in the cron service startup).

graycyrus (commented):
Fresh CodeRabbit-style review (self-generated, DRAFT PR). I applied the clear/safe terminal-status gaps directly (commit f18a2e9 — see PR conversation summary); the three items below are concurrency/design observations I am flagging rather than fixing unilaterally.

app/src/lib/i18n/id.ts
```diff
@@ -4146,6 +4146,7 @@ const messages: TranslationMap = {
   'flowRuns.status.pending_approval': 'Menunggu persetujuan',
   'flowRuns.status.failed': 'Gagal',
   'flowRuns.status.cancelled': 'Dibatalkan',
+  'flowRuns.status.interrupted': 'Terputus',
```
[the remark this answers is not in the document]
graycyrus: Fixed in 6494445 — both keys now read `Terhenti`. You were right that `Terputus` was already carrying `channels.status.disconnected` (plus `app.connectionIndicator.disconnected`), so an interrupted run read as a connectivity failure. `Terhenti` is the lifecycle sense we want. Left the two pre-existing `interrupted` keys outside this PR (`intelligence.agentWork.status.interrupted`, `orchestration.runStatus.interrupted`) alone — same collision, but not this diff's to change.

app/src/services/api/flowsApi.ts
```diff
@@ -56,7 +56,12 @@ export type FlowRunStatus =
   | 'completed_with_warnings'
   | 'pending_approval'
   | 'failed'
-  | 'cancelled';
+  | 'cancelled'
+  // A run whose future was dropped mid-flight (harness tool abort, chat turn
+  // end, timeout, or an app restart), reconciled to a terminal state by the
+  // core's `RunRowFinalizer` drop-guard or its boot-time orphan sweep (bug
+  // B42). Carries a human `error` reason; rendered as a settled, non-active run.
+  | 'interrupted';
```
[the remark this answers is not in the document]
graycyrus: Fixed in 6494445. `FlowRun.finished_at` now documents that it is stamped for *every* terminal status including `interrupted` (the drop-guard and boot sweep stamp it exactly like a normal terminal write), and `FlowRun.error` documents that on an `interrupted` run it carries the reconciliation reason — tool abort / turn end / app restart. Applied to both sides of the contract: `src/openhuman/flows/types.rs` and `app/src/services/api/flowsApi.ts`.

src/openhuman/flows/ops.rs
```diff
@@ -3771,8 +3887,144 @@ pub async fn flows_run(
...
+            &self.config,
+            &self.thread_id,
+            &self.flow_id,
+            "interrupted",
+            &observed,
+            &[],
+            Some(INTERRUPTED_DROP_REASON),
+        );
+    }
+}
```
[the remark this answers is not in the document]
graycyrus: Good catch — applied in 6494445. Confirmed the gap: the boot sweep, the cancel arm, and the success/failure arms all call `store::record_run`, but the `Drop` reconciliation only wrote the run row, so `flow_definitions.last_status` / `last_run_at` kept advertising the *previous* run for a flow whose latest run was interrupted. `record_run` is synchronous, so it is safe in `Drop` as you note. `run_row_finalizer_reconciles_orphaned_running_row_to_interrupted_on_drop` now asserts the summary too.

src/openhuman/flows/ops.rs
graycyrus: Hardened rather than left to the `is_in_flight` contract — 6494445.

Added `PROCESS_RUN_FLOOR`, an RFC3339 instant captured the first time this process touches the flow-run lifecycle. `start_flow_run_row` forces it *before* stamping the row it is about to insert, so the invariant holds either way round: every row this process inserts has `started_at >= *PROCESS_RUN_FLOOR`. `store::list_running_run_ids` now takes a `started_before` floor and filters `started_at < ?1`, and the sweep passes the floor — so the candidate set provably contains only prior-process rows, independent of registration timing. `is_in_flight` stays as the second guard (clock skew / a differently-skewed writer).

Filtering in SQL rather than returning `started_at` and filtering in Rust: same query-surface change, one fewer row shipped. The lexicographic `<` on a fixed-shape UTC RFC3339 column matches chronological order — the same comparison `expire_parked_runs` already relies on.

Two tests pin it: `boot_sweep_skips_a_run_started_after_the_process_floor` inserts a row via `start_flow_run_row` and deliberately does *not* register it — exactly the window you describe — and asserts it survives while the prior-process orphan is swept; `list_running_run_ids_excludes_rows_started_at_or_after_the_floor` pins the floor as exclusive. The sweep-test seed helper now stamps a fixed prior-process timestamp instead of `Utc::now()`, since a process-wide `LazyLock` floor would otherwise make those tests order-dependent.

src/core/runtime/services.rs
graycyrus: Decoupled in 6494445, since the floor added for the TOCTOU thread removed the reason to keep it there.

The sweep now lives in its own `services::spawn_flows_boot_reconcile()`, selected by `ctx.domains().flows` — the flows **domain** flag — instead of `services.cron`. Runs are started by the RPC Run control, the `run_flow` agent tool and the trigger bus, none of which need the cron service, so a `headless_api()` / cron-less `ServiceSet` no longer leaves prior-process orphans wedged forever.

The cron placement was originally load-bearing for ordering (`.await` to completion before the scheduler starts, so nothing this process started could be a candidate). `PROCESS_RUN_FLOOR` now guarantees that structurally, so the new task is free to be a plain detached spawn with no ordering requirement. `reconcile_schedule_triggers_on_boot` stays in cron bootstrap — that one genuinely is cron work.

src/openhuman/flows/ops.rs
graycyrus: No action taken, agreed as written. Recording the one thing that changed since: `Drop` now also calls `store::record_run` (CodeRabbit thread above), which is a third synchronous rusqlite call on the same path. Same reasoning holds — it returns `Result`, is logged rather than unwrapped, so no double-panic risk on an unwind — and it must stay in `Drop` for the same reason the other two do: moving it to `spawn_blocking` would detach the write from the drop and reopen the orphan window this guard exists to close.

src/openhuman/flows/ops.rs
```diff
@@ -3705,6 +3705,117 @@ pub async fn flows_run(
...
+        if let Err(e) = run_flow_body(
+            config_arc,
+            flow,
+            flow_id_owned,
+            body_thread_id,
+            input,
+            trigger,
+            no_actionable_nodes,
+        )
+        .await
```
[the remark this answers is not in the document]
graycyrus: Confirmed and fixed in b09e6f9 — thanks, this was reachable by ordinary user action, not just a theoretical interleaving.

The window: `flows_run_detached` returned the `run_id` while `run_registry::register` still lived inside the spawned `run_flow_body`, and a spawned task is not necessarily polled before the caller returns. The agent hands that `run_id` straight to the UI, so a Cancel in that gap hit `is_in_flight == false`, took the parked/stale branch in `flows_cancel_run`, wrote a terminal `cancelled` row and `drop_checkpoint`ed it — and the background task then started, executed the flow's real side effects, and overwrote `cancelled` with its own terminal status. A cancelled run that still runs, and misreports afterwards.

Took the first of your two options — registering before exposing the `run_id` — rather than a cancelled-check in the body: the check would still race (cancel could land between the check and the first node) and would leave the row/checkpoint already destroyed by the cancel path. Registration moved to both entry points, ahead of `start_flow_run_row`, with the token + guard passed into `run_flow_body`; a cancel on a returned `run_id` now always takes the signalled branch and the run's own cancellation arm owns the terminal write. `flows_run_detached_registers_the_run_before_returning_its_id` fails if registration moves back inside the task.

Side effect worth noting: this also closes the insert-before-register window that the boot sweep's `PROCESS_RUN_FLOOR` was added for in the previous commit. Kept the floor anyway — it holds regardless of what future callers do with the ordering, and prior-process rows are what the sweep is actually for. Both doc blocks describing the old ordering are updated rather than left stale.

src/core/runtime/services.rs
```diff
@@ -104,6 +104,44 @@ pub fn spawn_update_scheduler() {
...
+                    log::info!(
+                        "[flows] boot sweep reconciled {swept} orphaned running run(s) to 'interrupted'"
+                    );
+                }
+            }
+            Err(err) => {
+                log::warn!("[core] config load failed, skipping flows boot reconcile: {err}");
+            }
+        }
+    });
```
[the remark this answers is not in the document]
graycyrus: Applied in b09e6f9. Added the scheduling, config-load and completion logs; the completion line is unconditional so a `swept == 0` no-op is distinguishable from a task that never ran, which was the actual gap. Kept the existing `swept > 0` info line and the config-load warning as-is.

PR #5139, opened 2026-07-22
Revert "fix(tauri): gate cef_preflight to macOS/Linux to fix Windows build"

fzamel3333-ai:
Reverts tinyhumansai/openhuman#3225

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows build compatibility by removing an unsupported platform-specific code path.
  * Windows builds no longer rely on macOS/Linux-only functionality.

### DIFF
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -37,11 +37,6 @@ mod app_update;
 // (macOS/Windows/Linux): native Save-As dialog (rfd) + Downloads copy.
 mod artifact_commands;
 mod cdp;
-// macOS/Linux only: depends on the `nix` crate (a `cfg(unix)` dependency) and
-// resolves a platform cache path that is only defined for those targets. On
-// Windows it must not be compiled — see issue: Windows release build failed
-// with E0433 (`nix` unresolved) + E0425 (`cache_path` undefined).
-#[cfg(any(target_os = "macos", target_os = "linux"))]
 mod cef_preflight;
 mod cef_profile;
 // Windows-only pre-CEF wait for a dying prior instance to release the cache


senamakel (commented):

**Severity: blocking** | `app/src-tauri/src/lib.rs:40` (`mod cef_preflight;`)

This revert removes the `#[cfg(any(target_os = "macos", target_os = "linux"))]` gate on `mod cef_preflight;` that PR #3225 added. That gate was not cosmetic — it prevents the macOS/Linux-only `cef_preflight.rs` module from being compiled on Windows, where it causes a hard build failure for two independent reasons:

1. **`nix` crate unavailable on Windows.** `cef_preflight.rs` imports `nix::sys::signal::kill` and `nix::unistd::Pid` unconditionally at the top level (lines 29-30). In `app/src-tauri/Cargo.toml`, `nix` is declared under `[target.'cfg(unix)'.dependencies]` and is not resolvable for `x86_64-pc-windows-msvc`. The compiler will produce `E0433: unresolved module or unlinked crate "nix"`.

2. **`cache_path` undefined on Windows.** The `check_default_cache()` function (line 108) defines `cache_path` only under `#[cfg(target_os = "macos")]` (line 121) and `#[cfg(target_os = "linux")]` (line 127) — no Windows arm exists. The compiler will produce `E0425: cannot find value "cache_path" in this scope` for the unguarded fallback path.

Nothing in `cef_preflight.rs` has changed since PR #3225 that would make it Windows-compatible. The module doc comment still reads "CEF cache-lock preflight check (macOS and Linux)" and the call site in `lib.rs` (line 2860-2861) is already correctly gated, so removing the `mod` gate provides no benefit.

The unresolved CodeRabbit review thread identifies the same issue.

**Unmet gates:**
- The `cef_preflight` module must either be fully Windows-compatible (with a Windows implementation of its platform pieces) or remain gated out of the Windows build.

**Note on CI:** The "PR CI Gate" failure is a configuration issue — `Rust Tauri Coverage (cargo-llvm-cov)` was cancelled (runner timeout), which cascaded into the gate failure. It is unrelated to the correctness of this revert.

**Recommended next step:** Close this PR without merging. If the reporter has a specific Windows build problem, a targeted fix that makes `cef_preflight` platform-safe or addresses the real root cause is needed — not a blanket de-gating of a Unix-only module.

PR #5211, opened 2026-07-27
fix(agent-experience): make experience recall test deterministic (#5209)

### LINKED ISSUE #5209: flaky: dedicated_profile_experience_recall_merges_shared_legacy_store fails under diff-coverage scope
**Flaky / order-dependent** — blocks the diff-scoped Rust Core Coverage lane for any PR that adds a builtin agent.

`openhuman::agent::harness::session::turn::tests::dedicated_profile_experience_recall_merges_shared_legacy_store` (turn_tests.rs:2163) intermittently fails `assert!(enriched.contains("Legacy shared deployment guidance"))`.

**Non-deterministic:** passes on main's full CI and passed on the first run of #5205, but failed on two subsequent runs of the *same* commit. The test uses locally-constructed `dedicated`/`shared` memory instances (not the global singleton), so the flakiness comes from some other shared process state exposed when the diff-coverage subset (`-p openhuman --lib -- 'openhuman::agent' 'openhuman::agent_registry' 'openhuman::flows'`) runs that scope's tests together.

**Repo-wide friction:** any PR adding a builtin agent must edit `agent/harness/definition_tests.rs` (the exhaustive `(id, max_iterations)` audit), which pulls `openhuman::agent` into the diff-coverage scope — so this flake gates every agent-adding PR (surfaced by #5205).

**Fix:** identify the shared state `inject_agent_experience_context` / experience recall depends on and isolate it per-test (mirroring the FlowRunDigestSubscriber `with_memory` DI fix in #5176). Likely a global embedder/config or a shared experience-store path.

**Impact now:** blocks #5205's coverage gate despite that PR touching none of this code.

graycyrus:
## Summary

- Fixes the run-to-run flake in `dedicated_profile_experience_recall_merges_shared_legacy_store` (#5209). Root-causing found **two independent, deterministically-reproduced bugs** — not the HashMap-ordering the symptom implied.
- **Bug 1 (test infra):** deep agent-turn tests overflow the default ~2 MiB libtest thread stack and SIGABRT the whole process; libtest then tags whichever unrelated test was concurrently in flight as FAILED (usually the experience-recall test). Fixed by running those tests on an explicit 64 MiB stack.
- **Bug 2 (real product bug):** a Luhn-valid 13-digit millisecond timestamp in the serialized experience JSON trips the memory content sanitizer's credit-card redaction, corrupting the JSON so the record vanishes on read (~10% of writes). Fixed by base64-encoding the structured payload before storage.
- Also made the parallel test's timing-based overlap assertion deterministic (barrier rendezvous) — it flaked ~13% once the stack fix let it run.

## Problem

The diff-scoped command `cargo test -p openhuman --lib -- 'openhuman::agent' 'openhuman::agent_registry' 'openhuman::flows'` intermittently reported the experience-recall test as FAILED — pass on one run, fail on the next for the *same* commit. The scope filter is a prefix match, so it also runs `agent_orchestration`, `agent_experience`, etc.

Two distinct causes, both reproduced deterministically:

1. **Stack overflow → process abort → innocent-test tagging.** `turn_dispatches_spawn_subagent_through_full_path` and `agent_turn_runs_long_parallel_subagent_flow_with_many_nested_tool_calls` drive the full nested/parallel agent-turn harness — a deep async state machine that in debug/coverage builds exceeds the 2 MiB default libtest thread stack. The thread overflows and the process SIGABRTs. Because libtest runs tests concurrently, the abort marks whichever test was in flight as FAILED. The overflow is deterministic (every run); *which* concurrent test gets tagged is not — exactly the "same commit flips pass↔fail" signature. CI masked this with `RUST_MIN_STACK=64MiB`; a raw `cargo test` has none, so it reliably overflows.

2. **JSON payload corrupted by the content sanitizer.** `AgentExperienceStore` serializes each experience to JSON and stores it as memory-document `content`. `Memory::store` runs content through the secret/PII sanitizer, whose credit-card pattern (`\b(?:\d[\s\-]?){13,19}\b`, Luhn-gated) matches any 13–19-digit run. `put` stamps `created_at_ms`/`updated_at_ms` with `now_ms()` — a bare 13-digit run. Measured: exactly 10% of ms timestamps pass Luhn, so ~1 write in 10 had its timestamp rewritten to a `[REDACTED_PII_*]` token → invalid JSON → failed to parse on read → the experience silently vanished from `list()`, recall returned nothing, and the assertion genuinely failed (~10%, matching observation). This corrupted real experiences in production, not just the test.

## Solution

- **Large-stack test threads.** Both deep tests now run their body on an explicit `std::thread` with a 64 MiB stack driving a `current_thread` runtime — mirroring production's `agent::bus::handle_agent_run_turn_on_large_stack`. They can no longer overflow / abort the process regardless of `RUST_MIN_STACK`.
- **Base64 payload encoding.** `AgentExperienceStore` base64-encodes the serialized experience before `Memory::store` and decodes on read. Base64 has no 13+-digit bare-numeric runs (and no `Bearer`/`sk-` literals), so the sanitizer is a guaranteed no-op; the round-trip is lossless. Read has a plain-JSON fallback (a JSON object starts with `{`, not a base64 char, so legacy records decode unambiguously — no migration step). The store already redacts the sensitive free-text fields itself before serialization, so secret handling is unchanged.
- **Deterministic overlap.** The parallel test's `max_active >= 2` assertion no longer relies on a 25 ms `sleep` window; the two subagents' first provider calls rendezvous at a timeout-guarded `Barrier(2)`, guaranteeing overlap (and failing — not hanging — if parallelism regresses).

### Verification (fresh `cargo test` process each run → new HashMap/clock seed)

- Target test: **10/10** pass under the full diff-scope; **0** stack overflows.
- Both formerly-overflowing tests pass; parallel test **20/20** alone (was ~13/15).
- New regression test passes (and fails without the base64 fix).
- `agent_orchestration` (301), `agent_experience` (28), `memory_store` (277): green.

> **Out of scope, flagged for a follow-up:** `turn_triggers_configured_memory_agent_before_parent_prompt` fails deterministically **in isolation** under this scope — a pre-existing global-`AgentDefinitionRegistry` provider-inheritance order-dependency (`agent_memory` is a `Hint("chat")` sub-agent), unrelated to this change and to #5209. It is normally masked in the full suite (a neighbour sets up the registry) and was masked here by the earlier SIGABRT. Unmasked once the process stops aborting. Recommend a separate issue.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — new deterministic regression `experience_survives_content_sanitizer_with_luhn_valid_timestamp` over the real `UnifiedMemory` (fails without the fix); both deep tests now run to completion; parallel overlap made deterministic.
- [x] **Diff coverage ≥ 80%** — changed production lines (`store.rs` encode/decode + call sites) are exercised by the new regression test and the 28 existing `agent_experience` tests; the rest of the diff is test code executed by its own runs.
- [x] N/A — Coverage matrix: behaviour-preserving test-determinism fix + internal storage encoding; no feature rows added/removed/renamed.
- [x] All affected feature IDs listed under `## Related` — N/A (no matrix feature IDs touched).
- [x] No new external network dependencies introduced (`base64` is already a workspace dependency; tests use temp `UnifiedMemory` / mock memory).
- [x] N/A — Manual smoke checklist: does not touch a release-cut surface.
- [x] Linked issue closed via `Closes #5209` in `## Related`.

## Impact

- **Runtime/platform:** Rust core only. `AgentExperienceStore` now persists its payload base64-encoded; fully encapsulated (only the store reads that namespace's content) with a plain-JSON read fallback, so existing stored experiences keep decoding — no migration. Fixes silent ~10% loss of agent experiences in production.
- **Performance/security:** negligible (one base64 encode/decode per store/read). Secret/PII handling unchanged — the store still redacts free-text fields before serialization.

## Related

- Closes: #5209
- Follow-up PR(s)/TODOs: separate issue for the pre-existing `turn_triggers_configured_memory_agent_before_parent_prompt` global-registry order-dependency (unmasked, not introduced, by this change).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/flaky-experience-recall-test`
- Commit SHA: `f7a53f46752a28e44d107d0dc8aa09b621073643`

### Validation Run

- [x] N/A — `pnpm --filter openhuman-app format:check` (no frontend changes)
- [x] N/A — `pnpm typecheck` (no TypeScript changes)
- [x] Focused tests: diff-scope target 10/10, parallel 20/20, `agent_experience` 28, `agent_orchestration` 301, `memory_store` 277
- [x] Rust fmt/check: `cargo fmt` applied; lib+test binary compiles clean (`GGML_NATIVE=OFF cargo test -p openhuman --lib --no-run`)
- [x] N/A — Tauri fmt/check (no `app/src-tauri` changes)

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: agent-experience persistence is base64-encoded so structured payloads survive the memory content sanitizer; two deep tests run on a large stack; one test's overlap made deterministic.
- User-visible effect: agent experiences are no longer intermittently lost (~10%) on write; no user-facing surface change.

### Parity Contract

- Legacy behavior preserved: read path falls back to plain-JSON, so pre-existing stored experiences decode unchanged; free-text field redaction retained; scoring/retrieval/ordering untouched.
- Guard/fallback/dispatch parity checks: base64-then-plain-JSON decode fallback; timeout-guarded barrier fails rather than hangs.

- Duplicate PR(s): N/A
- Canonical PR: N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved timestamp-like values in stored agent experience records by shielding the payload during memory sanitization.
  - Kept backward compatibility with older, non-encoded stored entries.
  - Expanded redaction coverage for sensitive free-text fields; malformed records are now safely skipped during listing.
- **Reliability**
  - Made parallel subagent overlap behavior deterministic and less prone to hangs by using rendezvous synchronization with a timeout.
- **Tests**
  - Improved test stability by preventing stack-overflow-related aborts in deep async execution scenarios.

### DIFF
--- a/src/openhuman/agent/harness/session/tests.rs
+++ b/src/openhuman/agent/harness/session/tests.rs
@@ -929,8 +929,38 @@ async fn turn_with_native_dispatcher_persists_fallback_tool_calls() {
 /// plumbing this test asserts. Provider *routing* for Hint sub-agents
 /// is covered independently by
 /// `subagent_runner::ops::tests::resolve_subagent_source_*`.
-#[tokio::test]
-async fn turn_dispatches_spawn_subagent_through_full_path() {
+// The full spawn_subagent path (parent turn → run_subagent → nested agent
+// turn) is a deep async state machine. In debug/coverage builds each future
+// frame is large, and the two stacked turns exceed the default ~2 MiB libtest
+// per-test thread stack — the thread overflows and SIGABRTs the *entire* test
+// process. Because libtest runs tests concurrently, the abort then tags
+// whichever unrelated test happened to be in flight as FAILED, producing the
+// run-to-run flake reported in issue #5209 (the experience-recall test was the
+// most frequent victim). CI only avoided this by exporting a 64 MiB
+// `RUST_MIN_STACK`; a raw `cargo test` (e.g. the diff-scoped coverage command)
+// has no such env and reliably overflows. Production already drives agent
+// turns on an explicit large stack for this exact reason
+// (`agent::bus::handle_agent_run_turn_on_large_stack`). Mirror that here so the
+// test is self-contained and never aborts the process, regardless of
+// `RUST_MIN_STACK`.
+#[test]
+fn turn_dispatches_spawn_subagent_through_full_path() {
+    std::thread::Builder::new()
+        .name("spawn-subagent-full-path-test".to_string())
+        .stack_size(64 * 1024 * 1024)
+        .spawn(|| {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("build large-stack test runtime")
+                .block_on(turn_dispatches_spawn_subagent_through_full_path_inner());
+        })
+        .expect("spawn large-stack test thread")
+        .join()
+        .expect("large-stack spawn_subagent test thread panicked");
+}
+
+async fn turn_dispatches_spawn_subagent_through_full_path_inner() {
     use crate::openhuman::agent::harness::AgentDefinitionRegistry;
     use crate::openhuman::tools::SpawnSubagentTool;
 
--- a/src/openhuman/agent_experience/store.rs
+++ b/src/openhuman/agent_experience/store.rs
@@ -2,12 +2,53 @@ use crate::openhuman::agent_experience::types::{
     redact_text, stable_experience_id_for_profile, AgentExperience, ExperienceHit,
 };
 use crate::openhuman::memory::{Memory, MemoryCategory};
+use base64::Engine as _;
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet};
 use std::sync::Arc;
 
 pub const AGENT_EXPERIENCE_NAMESPACE: &str = "agent_experience";
+
+/// Encode a serialized experience so its structured payload survives the memory
+/// layer's free-text content sanitizer.
+///
+/// `Memory::store` runs every document's `content` through the secret/PII
+/// scrubber (`tinycortex … safety::sanitize_text`), whose *bare-numeric* PII
+/// patterns (credit-card via Luhn, CPF, CNPJ) match any 11–19-digit run. A
+/// serialized `AgentExperience` embeds `created_at_ms` / `updated_at_ms` as bare
+/// 13-digit millisecond timestamps, so whenever `now_ms()` happens to be
+/// Luhn-valid (~10% of the time) the scrubber rewrites the number to a
+/// `[REDACTED_PII_*]` token — corrupting the JSON so it no longer parses back on
+/// read. The record then silently vanishes from [`AgentExperienceStore::list`],
+/// making recall non-deterministic run-to-run (issue #5209).
+///
+/// Base64 has no 11+-digit bare-numeric runs (and no `Bearer`/`sk-` literals),
+/// so the sanitizer is a guaranteed no-op over the encoded payload and the
+/// round-trip is lossless. The store still redacts the sensitive free-text
+/// fields itself via [`redact_experience`] before serialization, so this does
+/// not weaken secret handling.
+fn encode_experience_payload(json: &str) -> String {
+    base64::engine::general_purpose::STANDARD.encode(json.as_bytes())
+}
+
+/// Decode a stored experience payload.
+///
+/// New records are base64(JSON) (see [`encode_experience_payload`]); legacy
+/// records are plain JSON. A JSON object starts with `{`, which is not in the
+/// base64 alphabet, so the base64 decode fails cleanly on legacy content and we
+/// fall back to parsing it as plain JSON — no ambiguity, no migration step.
+fn decode_experience_payload(stored: &str) -> Result<AgentExperience, String> {
+    if let Ok(bytes) = base64::engine::general_purpose::STANDARD.decode(stored.trim()) {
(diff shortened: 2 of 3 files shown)


graycyrus:
Review addressed (commit `909927b9e`):

**Codex P1 — preserve PII sanitization before encoding** ✅ Fixed. The base64 fix protected timestamps from the store-time scrubber but also hid the payload's free-text from it — a real regression. Now `redact_experience` runs the **full** `memory_store::safety::sanitize_text` (the exact scrubber `Memory::store` applies: private-key blocks → Bearer/`sk-`/Stripe/npm/OAuth → `pii::redact_pii` national-id/phone/credit-card) over every externally-populatable free-text field (`task_fingerprint, task_summary, lesson, reuse_hint, avoid_hint, error_class, agent_id, entrypoint, tools_used, tool_sequence, tags`) **before** serialization + base64. Numeric/`id`/`profile_id` are deliberately left intact (scrubbing structural numbers is the exact Luhn-timestamp corruption this PR fixes; base64 shields them from the store-time re-scrub; `id` is guarded by the memory layer's `has_likely_secret` up front).

Net invariant: a secret in any captured free-text field is redacted in both the stored record **and** what recall returns; timestamps survive. New regression test `secrets_in_free_text_are_redacted_before_storage` (Stripe `sk_live_` key + phone + private-key block) proves it and fails on the pre-fix code. `agent_experience` suite 29/29, target flaky test 10/10.

PR #5216, opened 2026-07-27
fix(voice): repair execute bit on extracted Piper binaries (#5045)

### LINKED ISSUE #5045: Local Piper TTS engine fails to initialize on macOS M-series architecture
## Summary

When enabling local Piper Text-to-Speech (TTS), the application encounters a 10-second initialization timeout and fails to generate voice output. The underlying failure is caused by an architecture mismatch, as the application downloads Intel (x86_64) binaries instead of native Apple Silicon (arm64) versions.

## Problem

**What Happened vs. Expected:**
After installing Piper, choosing a voice, and selecting the test option, the TTS test failed. Expected the test to pass successfully.

**Impact:**
Local voice synthesis is entirely non-functional on all M-series (M1/M2/M3/M4) Macs out of the box.

**Version / Platform:**

* **OS:** macOS (Apple Silicon / arm64 architecture)
* **App Version:** Version 0.61.8 (0.61.8)

**Steps to Reproduce:**

1. Install and launch OpenHuman on an Apple Silicon Mac.
2. Navigate to settings and enable the local Piper TTS engine.
3. Choose a voice and select the "Test" button.
4. Observe that the test fails to execute.

## Solution

**Suspected Cause:**
The application's asset downloader fetches a deployment bundle containing legacy Intel (`x86_64`) binaries rather than native Apple Silicon (`arm64`) binaries, compounded by incorrect file permissions. This conclusion was drawn through the following diagnostic steps:

1. **Log Analysis:** Reviewed the application logs, which revealed a 10-second initialization hang followed by `dyld` dynamic library loading errors.
2. **Sandbox Inspection:** Located and inspected the local binary sandbox directory automatically generated by the application at `~/.openhuman/bin/piper/`.
3. **Architecture Verification:** Checked the compilation architecture of the extracted binaries (`piper`, `piper_phonemize`, and `espeak-ng`), confirming they were built for `x86_64` instead of native `arm64`.
4. **Permissions Check:** Audited the file permissions within the sandbox directory and found that the `espeak-ng` binary lacked standard executable (`+x`) permissions.

**Proposed Fix:**

1. Recompile or re-package the `piper_macos_aarch64` asset bundle on the distribution server to include genuine native `arm64` binaries for `piper`, `piper_phonemize`, and `espeak-ng`.
2. Ensure the decompression workflow explicitly sets or preserves the executable permissions (`chmod +x`) for the internal binaries.

## Acceptance criteria

- [ ] **Repro gone** — Bug no longer reproduces on the stated environment (or root cause documented if intentional).

- [ ] **Regression safety** — Unit, integration, or E2E coverage added or updated if this should not come back.

- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/ci-lite.yml`](../../.github/workflows/ci-lite.yml)).

- [ ] **…** — Other verify-before-close items.

## Related

* None / N/A

### Triage Assessment

This issue is relevant and not a duplicate. The user reports Piper TTS failing on M-series Macs with a 10-second timeout, finding x86_64 binaries in the sandbox directory and missing executable permissions on `espeak-ng`. Local TTS is entirely non-functional on all M-series Macs out of the box. The architecture detection code at `src/openhuman/inference/local/install_piper.rs:76-84` correctly resolves `piper_macos_aarch64.tar.gz` on arm64 hosts -- the exact location of the bug depends on whether the upstream archive contains wrong-architecture binaries, or the extraction/permissions path is faulty.

### Proposed Scope

**In-scope:**
1. Verify what `piper_macos_aarch64.tar.gz` from the upstream Piper release actually contains (architecture of the bundled binaries).
2. Fix the extraction step to set executable permissions (+x) on all extracted binaries.
3. Log the detected architecture and the downloaded archive name for diagnostics.

**Explicit non-goals:**
- Rebuilding Piper for arm64 (upstream already ships arm64 builds).
- Supporting additional voice models beyond the bundled `en_US-lessac-medium`.
- Changing the voice download path or HuggingFace voice model resolution.

### Implementation Steps

1. **Verify upstream asset content**
   - Download `piper_macos_aarch64.tar.gz` from `https://github.com/rhasspy/piper/releases/latest/download/piper_macos_aarch64.tar.gz` on an M-series Mac.
   - Extract and run `file piper piper_phonemize espeak-ng` to confirm architecture.
   - If the archive contains x86_64 binaries (the `file` output shows "Mach-O 64-bit executable x86_64"), the bug is upstream. File an issue with the Piper project or switch to a known-good release tag.
   - If the archive contains arm64 binaries (file shows "Mach-O 64-bit executable arm64"), the bug is in extraction/permissions.

2. **Fix file permissions on extracted binaries**
   - File: `src/openhuman/inference/local/install_piper.rs`. After the archive is extracted (in the install flow, probably around `extract_tar_gz` or `extract_zip` call sites), add `std::fs::set_permissions` calls to set `std::os::unix::fs::PermissionsExt::set_mode(0o755)` on `piper`, `piper_phonemize`, and `espeak-ng`.
   - File: `src/openhuman/inference/local/voice_install_common.rs` -- if the extraction logic is shared across whisper and piper installers, add the permissions fix there as a post-extraction step for all extracted binaries. This prevents the same issue from affecting Whisper.
   - The `espeak-ng` binary is inside the archive as a sibling of `piper`. Confirm the extraction path: find where the archive is unpacked (likely the `piper/` subdirectory of the workspace `bin/` directory configured at `src/openhuman/inference/paths.rs:271`).

3. **Add architecture diagnostics**
   - File: `src/openhuman/inference/local/install_piper.rs` function `binary_download_asset()` (lines 66-100). Log the detected `std::env::consts::ARCH` and the chosen suffix, e.g.:
     ```
     log::info!("{LOG_PREFIX} host arch={arch}, selecting asset=piper_{suffix}.tar.gz");
     ```
   - After extraction, log the file permissions stance of the extracted binary, e.g.:
     ```
     log::debug!("{LOG_PREFIX} extracted piper binary permissions: {:o}", piper_path.metadata()?.permissions().mode());
     ```

4. **Improve error messages**
   - File: `src/openhuman/inference/local/install_piper.rs`. The installer currently validates archive size (`MIN_BINARY_ARCHIVE_BYTES`) and voice file sizes. Add a post-extraction check: verify at least one extracted binary has executable permission. If not, log a diagnostic "extracted binaries lack execute permission -- applying fix" and apply it.

### Verification

- **Unit tests (Rust):**
  - `src/openhuman/inference/local/install_piper_tests.rs` or `src/openhuman/inference/local/voice_install_common_tests.rs` -- add a test that creates a mock archive, runs the extraction, and verifies `+x` is set on the extracted binary.
  - The existing `tests/raw_coverage/inference_local_ops_piper_raw_coverage_e2e.rs` tests already exercise the install flow using a mock server. Extend `PiperMockState` to serve an archive that would trigger the permissions issue, and verify the fix applies `+x`.
- **Manual verification (ARM Mac required):**
  - Run `install_piper` on an M-series Mac and confirm:
    - The log shows the correct `aarch64` architecture.
    - The downloaded filename contains `macos_aarch64`.
    - After extraction, `file ~/.openhuman/bin/piper/piper/piper` shows `arm64`.
    - `ls -la ~/.openhuman/bin/piper/piper/espeak-ng` shows `-rwxr-xr-x` (executable).
    - The "Test" button in Settings -> Voice -> Piper TTS produces audible speech.

### Risks and Open Questions

- **Upstream Piper providing wrong-architecture binaries.** If the upstream `piper_macos_aarch64.tar.gz` actually contains x86_64 binaries, the fix is out of our hands (file an upstream issue). In that case, the code-level fix is to add a post-extraction `lipo` / `file` check that warns the user and falls back to running under Rosetta, or pins to a known-good release tag.
- **Rosetta fallback.** If the upstream archive contains x86_64, the binary would still run under Rosetta (slower startup, hence the 10-second timeout). Adding a `--rosetta` detection and warning would help users self-diagnose.
- **CI cannot test native ARM Mac.** The existing mock-based tests cover the install flow but not the actual binary execution. Manual testing on M-series hardware is required for final verification.

### Acceptance Criteria

- Piper TTS installation on an M-series Mac downloads the correct arm64 archive.
- Extracted binaries (piper, piper_phonemize, espeak-ng) have executable permissions.
- The "Test" button in Voice settings produces audible speech on M-series Macs.
- Post-extraction logging includes architecture and permissions diagnostics.
- Existing mock-based installer tests pass unchanged.

M3gA-Mind:
## Summary

- Repair the execute bit on the binaries extracted from the Piper release archive — upstream ships `espeak-ng` as mode `0644` and `tar::Archive::unpack` reproduces the archived mode verbatim, leaving it non-executable.
- Log the detected host arch and the selected release asset so a wrong-architecture download is diagnosable from user logs.
- Add 5 regression tests covering both extraction layouts plus the three no-op cases.
- **Scope note:** this does **not** make Piper work on Apple Silicon. That half of #5045 is an upstream packaging defect, evidenced below, and is not fixable in this repo.

## Problem

#5045 reports local Piper TTS failing on M-series Macs with a 10 s init timeout and `dyld` errors, plus `espeak-ng` lacking `+x`. Investigating the two claims separately:

**1. Architecture (upstream defect — not fixable here).** I downloaded and inspected both real macOS assets from the pinned upstream release `2023.11.14-2`:

| Check | Result |
| --- | --- |
| `lipo -archs` on the **aarch64** bundle | `x86_64` for `piper`, `piper_phonemize`, `espeak-ng` |
| `cmp` aarch64 vs x64 bundle | **byte-identical** for all three binaries |
| Published asset sizes | 19,146,957 vs 19,146,927 — 30 bytes apart |
| `.dylib` files present | **zero** (only a `libonnxruntime…dylib.dSYM` debug dir) |
| `otool -L piper` | requires `@rpath/libespeak-ng.1.dylib`, `@rpath/libpiper_phonemize.1.dylib`, `@rpath/libonnxruntime.1.14.1.dylib` — none shipped |

`piper_macos_aarch64.tar.gz` is a mislabeled copy of the x64 bundle, **and** it omits every dylib the executable links against — which is the actual source of the reported `dyld` errors, and would break Piper on macOS even at the correct architecture.

The arch selection in `binary_download_asset()` was **already correct** (`aarch64`/`arm64` → `macos_aarch64`); changing it would be a no-op. `rhasspy/piper` has published nothing since 2023-11-14, and the successor `OHF-Voice/piper1-gpl` ships Python wheels only, so adopting it is a runtime redesign rather than a URL swap. Left for a maintainer decision rather than forced into this PR.

**2. Permissions (in-repo — fixed here).** `espeak-ng` is mode `0644` inside the archive. `tar::Archive::unpack` masks the archived mode to `0o777` and applies it, so extraction faithfully reproduces a non-executable `espeak-ng`; Piper then fails when it shells out to phonemize.

## Solution

- `ensure_executable_bits()` runs after both extraction paths and sets `0755` on `piper`, `piper_phonemize`, `espeak-ng`, rather than trusting archive metadata.
- Probes the same three layouts as `paths::workspace_piper_binary_candidates` (nested `piper/`, flattened root, `bin/`). Deliberately bounded rather than a recursive walk — a walk would descend into `espeak-ng-data/` and `voices/` (a ~60 MB `.onnx`) for no benefit.
- Skips anything that already has an execute bit, so an upstream-correct `0755` (or a narrower `0700`) is never rewritten; only the listed binaries are touched, so data payloads never gain `+x`.
- Best-effort by design: a `chmod` failure is logged but does not fail the install, since a working `PIPER_BIN` on `PATH` would otherwise be regressed by a hard error.
- `#[cfg(unix)]` throughout — Windows has no execute bit, and the const/helper would be dead code there.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 5 tests in `install_piper::tests::executable_bits`: repairs the nested layout (the real macOS tarball shape), repairs the flat layout, leaves an already-executable `0700` binary untouched, does not touch data payloads (`libtashkeel_model.ort`), and is a no-op on a missing directory.
- [x] **Diff coverage ≥ 80%** — every changed runtime line in `ensure_executable_bits` / `executable_search_dirs` is exercised by the new tests; the remaining changed lines are a `log::info!` and comments.
- [x] Coverage matrix updated — `N/A: bug fix to an existing install path, no new/renamed feature row in docs/TEST-COVERAGE-MATRIX.md`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix row touched`
- [x] No new external network dependencies introduced — no new deps; the installer's existing download path is unchanged.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface changed`

## Impact

- Desktop/CLI, macOS + Linux only (`#[cfg(unix)]`); Windows behaviour is unchanged.
- No API, schema, or migration impact. Revert is a clean removal of one function and its call site.
- Security: widens mode to `0755` on three known binary names inside the app's own install directory only, and only when no execute bit is set.
- **Does not close the user-visible symptom on Apple Silicon.** The `## Problem` evidence shows the upstream bundle is unusable on macOS regardless of arch. The issue's "repro gone" criterion cannot be met from this repo until upstream republishes the asset (or we migrate off `rhasspy/piper`).

## Related

- Closes: #5045
- Follow-up PR(s)/TODOs:
  - Upstream: `rhasspy/piper` must republish `piper_macos_aarch64.tar.gz` as genuine arm64 **and** include the `@rpath` dylibs. Repo is dormant since 2023-11-14.
  - Alternative: migrate to `OHF-Voice/piper1-gpl` (ships a real `macosx_11_0_arm64` wheel) via the `runtime_python` domain — a design decision, not a drop-in swap.
  - Pre-existing, unrelated: `binary_download_asset()` builds `piper_linux_armv7.tar.gz`, but the published asset is `piper_linux_armv7l.tar.gz` (trailing `l`) — a guaranteed 404 on armv7. Left out of scope; happy to file separately.
  - Pre-existing, unrelated: `schemas::tests::install_piper_handler_serializes_concurrent_calls` asserts on the global Piper install status while holding a different lock than the sibling `install_piper` tests that reset it. It passes in isolation and fails intermittently in-batch.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/5045-piper-macos-arm64-perms`
- Commit SHA: `fcf1e6a`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no frontend files changed (single Rust file)`
- [x] `pnpm typecheck` — `N/A: no TypeScript changed`
- [x] Focused tests: `cargo test --lib install_piper` — the 5 new `executable_bits` tests pass, as do the 9 pre-existing `install_piper` tests.
- [x] Rust fmt/check (if changed): `rustfmt --edition 2021 --check src/openhuman/inference/local/install_piper.rs` — clean
- [x] Tauri fmt/check (if changed): `N/A: no app/src-tauri changes`

### Validation Blocked

- `command:` re-running `cargo test --lib install_piper` against pristine `main`
- `error:` not an error — local `cargo test` / test-target compiles were disallowed mid-task (shared-machine load) before the A/B comparison could be run.
- `impact:` one pre-existing test, `schemas::tests::install_piper_handler_serializes_concurrent_calls`, failed in the batch run and **passed in isolation**. It asserts on the global Piper install status under a different lock than the sibling tests that reset it, and this PR touches neither install status nor that lock — so it is a pre-existing cross-test race, not a regression. Flagged rather than silently omitted; CI runs the full suite.

### Behavior Changes

- Intended behavior change: extracted Piper binaries are made executable instead of inheriting a non-executable archive mode.
- User-visible effect: none on its own for Apple Silicon (blocked upstream — see `## Impact`); on Linux/Intel macOS it removes a class of "permission denied" failures at phonemization time. New `[voice-install:piper]` log lines aid diagnosis.

### Parity Contract

- Legacy behavior preserved: download, extraction, status reporting, and short-circuit paths are unchanged; the repair is an added post-extraction step.
- Guard/fallback/dispatch parity checks: chmod failure is non-fatal, preserving the `PIPER_BIN`-on-PATH fallback; `#[cfg(not(unix))]` no-op preserves Windows behaviour exactly.

- Duplicate PR(s): none — no open PR referenced #5045 or matched piper/arm64 on title or branch.
- Canonical PR: this
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed installation issues where Piper executables extracted from certain archives lacked executable permissions.
  * Improved compatibility with macOS Piper packages, including archives containing non-executable `espeak-ng` files.
  * Preserved existing permissions and avoided modifying non-executable payload files.
  * Downloaded installer archives are now removed after extraction.

### DIFF
--- a/src/openhuman/inference/local/install_piper.rs
+++ b/src/openhuman/inference/local/install_piper.rs
@@ -816,6 +816,15 @@ mod tests {
         let target = paths::workspace_piper_binary_candidates(&config)[0].clone();
         std::fs::create_dir_all(target.parent().unwrap()).unwrap();
         std::fs::write(&target, b"stub").unwrap();
+        // "Present" now means present AND executable: a 0644 file is not a
+        // usable binary, and resolving it would pin us to a copy that cannot
+        // launch (see `non_executable_workspace_binary_is_skipped_so_path_can_win`).
+        // `std::fs::write` creates 0644, so grant the bit explicitly.
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&target, std::fs::Permissions::from_mode(0o755)).unwrap();
+        }
         let found = find_workspace_piper_binary(&config).expect("should find binary");
         assert_eq!(found, target);
         wipe_shared_install_dir(&config);
--- a/src/openhuman/inference/local/install_piper.rs
+++ b/src/openhuman/inference/local/install_piper.rs
@@ -37,6 +37,20 @@ const MIN_VOICE_BYTES: u64 = 30 * 1024 * 1024;
 /// a 404 HTML response masquerading as JSON.
 const MIN_VOICE_JSON_BYTES: u64 = 256;
 
+/// Binaries shipped inside the Piper release archive that must carry the
+/// executable bit for the engine to start.
+///
+/// The upstream macOS tarballs ship `espeak-ng` with mode `0644` (#5045).
+/// `tar::Archive::unpack` faithfully reproduces the archived mode, so a
+/// plain extraction leaves the file non-executable and Piper fails when it
+/// shells out to phonemize. Extraction must therefore repair the bit
+/// rather than trust the archive.
+///
+/// Unix-only: Windows has no executable bit, so the const would be dead
+/// code there.
+#[cfg(unix)]
+const PIPER_EXECUTABLES: [&str; 3] = ["piper", "piper_phonemize", "espeak-ng"];
+
 /// Result of resolving the Piper binary archive URL for the host OS.
 struct BinaryAsset {
     url: String,
@@ -75,11 +89,20 @@ fn binary_download_asset() -> Option<BinaryAsset> {
     if cfg!(target_os = "macos") {
         // Two assets exist (`piper_macos_x64.tar.gz` and
         // `piper_macos_aarch64.tar.gz`). Pick based on the host arch.
+        //
+        // NOTE (#5045): as of the pinned upstream release `2023.11.14-2`
+        // the two macOS assets are byte-identical x86_64 builds — the
+        // `aarch64` name is a mislabel — and neither ships the
+        // `@rpath` dylibs `piper` links against. Selecting the arm64
+        // asset here is still correct, but it cannot make Piper run on
+        // Apple Silicon until upstream republishes the bundle. The arch
+        // is logged below so a bad asset is diagnosable from user logs.
         let arch = std::env::consts::ARCH;
         let suffix = match arch {
             "aarch64" | "arm64" => "macos_aarch64",
             _ => "macos_x64",
         };
+        log::info!("{LOG_PREFIX} host arch={arch} selecting asset=piper_{suffix}.tar.gz");
         return Some(BinaryAsset {
             url: format!("{base}/piper_{suffix}.tar.gz"),
             kind: ArchiveKind::TarGz,
@@ -369,10 +392,78 @@ async fn run_install(config: &Config, voice: &str) -> Result<(), String> {
         ArchiveKind::Zip => extract_zip(&archive_path, &dest)?,
         ArchiveKind::TarGz => extract_tar_gz(&archive_path, &dest)?,
     }
+    ensure_executable_bits(&dest);
     let _ = std::fs::remove_file(&archive_path);
 
     Ok(())
 }
+
+/// Directories under the install root where an extracted Piper binary can
+/// land. Mirrors the layouts probed by
+/// [`paths::workspace_piper_binary_candidates`]: the macOS/Linux tarballs
+/// nest under `piper/`, some builds flatten to the root, and a few use
+/// `bin/`. Bounded on purpose — a recursive walk would descend into
+/// `espeak-ng-data/` and `voices/` (which holds a ~60 MB `.onnx`) for no
+/// benefit.
+#[cfg(unix)]
+fn executable_search_dirs(dest_dir: &std::path::Path) -> [PathBuf; 3] {
+    [
+        dest_dir.to_path_buf(),
+        dest_dir.join("piper"),
+        dest_dir.join("bin"),
+    ]
+}
+
+/// Repair the executable bit on the binaries extracted from the Piper
+/// archive.
+///
+/// Upstream ships `espeak-ng` as mode `0644` inside both macOS tarballs
+/// (#5045) and `tar::Archive::unpack` reproduces the archived mode
+/// verbatim, so the extracted tree is left with a non-executable
+/// `espeak-ng`. Rather than trust archive metadata, set `0o755` on every
+/// binary we know Piper needs.
+///
+/// Best-effort by design: a `chmod` failure is logged but does not fail
+/// the install, since the caller may still have a working engine on
+/// `PATH` (`PIPER_BIN`) and a hard error here would regress that path.
+#[cfg(unix)]
+fn ensure_executable_bits(dest_dir: &std::path::Path) {
+    use std::os::unix::fs::PermissionsExt;
+
+    for dir in executable_search_dirs(dest_dir) {
+        for name in PIPER_EXECUTABLES {
+            let path = dir.join(name);
+            let Ok(meta) = std::fs::metadata(&path) else {
+                continue;
+            };
+            if !meta.is_file() {
+                continue;
+            }
+            let mode = meta.permissions().mode();
+            // Any of user/group/other execute already set → leave it be.
+            if mode & 0o111 != 0 {
+                log::debug!(
+                    "{LOG_PREFIX} {} already executable (mode {:o})",
+                    path.display(),
+                    mode & 0o777
+                );
+                continue;
+            }
+            match std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o755)) {
+                Ok(()) => log::info!(
+                    "{LOG_PREFIX} repaired execute bit on {} (was mode {:o})",
+                    path.display(),
+                    mode & 0o777
+                ),
+                Err(e) => log::warn!("{LOG_PREFIX} could not chmod +x {}: {e}", path.display()),
+            }
+        }
+    }
+}
+
+/// Windows has no executable bit — extraction is sufficient there.
+#[cfg(not(unix))]
+fn ensure_executable_bits(_dest_dir: &std::path::Path) {}
 
 fn update_stage(stage: String) {
     let mut current = read_status(ENGINE_PIPER);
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

M3gA-Mind:
@coderabbitai review

PR #5238, opened 2026-07-28
feat(embed): Medulla sub-facade and the RPC surface to drive a session

senamakel:
Second slice of the Medulla-on-OpenHuman work, following #5229. Gives an embedding host typed access to the Medulla backend, and extends the RPC surface from "observe a session" to "drive one".

**Draft** — raised for review of the direction. Not asking for merge yet; the consuming branch (tinyhumansai/medulla#82) is still in flight.

## Commits

| | |
|---|---|
| `5472202` | `embed::Medulla` sub-facade |
| `a3f693c` | Six session controllers + matching facade methods |

## What's here

`embed::Medulla` gives a host `core.medulla().send_message(..)` instead of hand-rolled `invoke` with `serde_json::Value`. Types are re-exported from the domain rather than mirrored — a parallel set of facade structs would be one more thing to keep in step with the wire contract.

Six new controllers wrap client methods that already existed but had no RPC surface, so nothing outside the crate could reach them:

```
medulla_create_session   medulla_get_session
medulla_send_message     medulla_abort
medulla_list_messages    medulla_list_events
```

## Contract details worth review

**`send_message` takes an explicit `sync` flag** rather than picking one. A UI wants non-blocking so it can render progress while the turn runs; a scripted caller usually wants to block for the reply. Defaults to `false`, so an omitting caller gets the UI-friendly behaviour.

**`after` on the replay methods is a cursor, not a page offset.** Passing the last seq already seen returns only what is new, which is what makes a reconnect cheap rather than a full refetch.

## Three derives that came out of building the first real consumer

- **`MedullaStatus` was `Serialize`-only.** It is an *output* type from ops' point of view, but it round-trips — the facade deserializes it back off the RPC boundary. Output-only compiles right up until something reads it.
- **`RosterWorker` / `SessionSummary` gain `Default`** so downstream tests can construct them without spelling every field.
- **`SessionStatus` gains `Default`**, resolving to its existing `Other` variant. `Other` is the honest choice: a constructed-rather-than-decoded value has not declared a status, and defaulting to `Active` would assert one.

## Testing note

A unit test pins every dispatched method name against the registered controllers. Those names are strings on that boundary, so a typo or a renamed controller would otherwise surface as `CoreError::Unavailable` at runtime — indistinguishable from a domain the host gated off deliberately.

**Not yet verified:** this branch was developed as a submodule inside the consuming repo's workspace, where cargo resolves upward and refuses to run OpenHuman's own suite. Full verification against a standalone checkout is pending and will be reported here before this leaves draft.

### DIFF
--- a/src/embed/medulla.rs
+++ b/src/embed/medulla.rs
@@ -22,7 +22,10 @@ use super::call::call;
 use super::error::CoreError;
 use crate::core::runtime::CoreRuntime;
 
-pub use crate::openhuman::medulla::client::{RosterWorker, SessionSummary};
+pub use crate::openhuman::medulla::client::{
+    AbortResult, EventEnvelope, Message, RosterWorker, SendResult, SessionCreated, SessionDetail,
+    SessionSummary,
+};
 pub use crate::openhuman::medulla::ops::MedullaStatus;
 
 /// Typed access to the Medulla backend.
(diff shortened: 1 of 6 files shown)


senamakel:
## Local verification pass — one real test failure

Verified from a **standalone** checkout (`git worktree add worktrees/verify-embed-facades origin/embed-facades`, then `git submodule update --init --depth 1 vendor/{tinyagents,tinychannels,tinycortex,tinyflows,tinyjuice,tinyplace}`) so cargo does not resolve up into the medulla workspace. Worktree HEAD: **`a3f693c`**.

Note: while this ran, `origin/embed-facades` advanced to **`6b9a6e4` ("style: format the session RPC additions")**. That commit touches only `src/embed/medulla.rs` and `src/openhuman/medulla/ops.rs` and is formatting-only, so everything below still applies — except the `cargo fmt` line, which it fixes.

| Command | Result |
| --- | --- |
| `cargo fmt -- --check` | **failed at `a3f693c`** — fixed by `6b9a6e4`, see below |
| `GGML_NATIVE=OFF RUST_MIN_STACK=16777216 cargo test --lib --no-default-features --features "tokenjuice-treesitter,medulla" medulla` | **FAILED — 120 passed, 1 failed, 9836 filtered out** |
| `… cargo test … embed` | **pass** — 211 passed, 0 failed, 9746 filtered out |
| `GGML_NATIVE=OFF cargo clippy --lib --no-default-features --features "tokenjuice-treesitter,medulla"` | **pass** (exit 0) — **zero** diagnostics in `src/embed/` or `src/openhuman/medulla/` |
| `node scripts/ci/check-feature-forwarding.mjs` | **pass** — "OK: every default-ON core gate is forwarded to the desktop shell" (15 core gates, 13 shell forwards, `tui` + `medulla` explicitly allowlisted) |
| `GGML_NATIVE=OFF cargo check --lib --no-default-features --features tokenjuice-treesitter` (feature-OFF) | **pass** (exit 0, 0 errors) |

### The failure — real, deterministic, not a baseline flake

```
openhuman::medulla::schemas::tests::rpc_method_names_follow_the_crate_convention
panicked at src/openhuman/medulla/schemas.rs:363:9:
  left:  ["openhuman.medulla_status", "openhuman.medulla_list_sessions",
          "openhuman.medulla_create_session", "openhuman.medulla_get_session",
          "openhuman.medulla_send_message", "openhuman.medulla_abort",
          "openhuman.medulla_list_messages", "openhuman.medulla_list_events",
          "openhuman.medulla_roster"]
  right: ["openhuman.medulla_status", "openhuman.medulla_list_sessions",
          "openhuman.medulla_roster"]
```

The branch adds six RPC methods (`create_session`, `get_session`, `send_message`, `abort`, `list_messages`, `list_events`) but the expected-name list in this test was not updated alongside them.

**Reproduces in isolation.** Re-ran with `-- --test-threads=1` and the single test name: `0 passed; 1 failed`. It is an assertion mismatch, not a timeout, so it is not one of the known load/parallelism baseline flakes. `src/openhuman/medulla/schemas.rs` is untouched by `6b9a6e4`, so this still fails at the current head.

### The fmt failure (already fixed upstream)

At `a3f693c`, `cargo fmt -- --check` exited 1 with two diffs — a `create_session` signature in `src/embed/medulla.rs:68` and a `call(...)` in `src/openhuman/medulla/ops.rs:84`, both wanting to collapse onto one line. `6b9a6e4` applies exactly those two hunks, so fmt should be clean at the current head. I did not re-run fmt against `6b9a6e4`.

### Baseline noise — none of it hit this run

I did not observe the known-flaky `cron::scheduler`, `inference::local::install_*`, `tinyplace::manifest::wallet_usdc_balance_reads_from_tinyplace_settlement_rpc`, or `core::dispatch` failures. `RUST_MIN_STACK=16777216` was set on both test runs as required.

Clippy produced 6 warnings crate-wide, all outside this branch's files and pre-existing: unused imports at `agent/harness/mod.rs:70`, `inference/provider/mod.rs:39`, `tinyagents/mod.rs:47`, and dead code at `core/observability.rs:2381/2387/2394`.

### Not run

- Only the `medulla` and `embed` test filters were run, not the full `--lib` suite (~9,957 tests) — out of scope for this pass, so I cannot speak to the rest of the suite.
- No frontend / Vitest / E2E / Tauri-shell checks.
- Clippy was `--lib` only (no `--all-targets`) and without `-D warnings`.
- Everything ran at `a3f693c`; nothing was re-run against `6b9a6e4`.
- Read-only pass: no source was changed and nothing was committed or pushed.

PR #5284, opened 2026-07-30
feat(cost): price MiniMax-M3 and MiniMax-M2.7 in the model catalog

octo-patch:
Reason: The MiniMax provider and OpenAI-compatible `/v1` endpoint are already configured, but the runtime `KNOWN_MODEL_PRICING` catalog had no MiniMax-M3 or MiniMax-M2.7 rows, so BYOK spend recorded a $0 cost estimate.

## Changes
- Add `minimax/minimax-m3` and `minimax/minimax-m2.7` rows to `KNOWN_MODEL_PRICING` in `src/openhuman/cost/catalog.rs`:
  - M3: $0.60 / $2.40 per million input / output tokens, $0.12 cached read, 1,000,000 context window.
  - M2.7: $0.30 / $1.20 per million input / output tokens, $0.06 cached read, 204,800 context window.
- Introduce `VISION_INPUT_MODELS` + `model_accepts_image_input` and project the M3 image/video input capability into `ModelRegistryEntry::vision` and the TinyAgents catalog `vision` flag — the one capability flag the BYOK OpenAI-compatible path cannot derive at runtime. M2.7 stays text-only.

## Why
BYOK MiniMax spend is addressed by the real model id (`minimax/minimax-m3`, `minimax/minimax-m2.7`), so the cost catalog must price both current models or those turns record a $0 estimate while still counting toward usage. The MiniMax `/v1` endpoint is already registered as a built-in cloud provider, so no endpoint change is needed.

## Checks
- `cargo check --lib`
- `cargo test --lib cost::catalog::` (23 passed, 0 failed)

## Summary by CodeRabbit

* **New Features**
  * Updated model pricing information through July 2026.
  * Added catalog support for MiniMax M3, M2.7, and M2.7 Highspeed models.
  * Added pricing, context-window, and discounted M3 pricing for requests up to 512K input tokens.
  * Identified image-input capabilities for supported MiniMax models.

* **Bug Fixes**
  * Improved model lookup handling for MiniMax naming variations.
  * Ensured vision capabilities are accurately reflected across model listings and enriched entries.


octo-patch:
Updated this branch to current main, preserving the MiniMax pricing and vision changes in src/openhuman/platform/cost/catalog.rs, and pushed a new commit. I ran cargo fmt --all -- --check and cargo test -p openhuman --lib minimax_ (4 passed).

octo-patch:
Synced this branch with current main. The MiniMax-M3 and MiniMax-M2.7 rows in src/openhuman/platform/cost/catalog.rs are unchanged, and that is still the only file this PR touches. The Rust core coverage lane that was red here failed on raw-profile collection, which main has since fixed, so CI is now green: Rust Quality (fmt, clippy), Rust Core Coverage and PR CI Gate all pass.

M3gA-Mind:
Maintainer triage pass — review only, no changes pushed, not an approval.

I went and checked the vendor's pricing page rather than reasoning from the diff, and **both pricing findings on this PR are correct**. As it stands this PR would misprice MiniMax BYOK spend in both directions, so I don't think it should merge in its current form — but the fix is small and the PR is worth landing after it.

From <https://platform.minimax.io/docs/guides/pricing-paygo> (per 1M tokens):

| Model | Input | Output | Cache read |
| --- | --- | --- | --- |
| **MiniMax-M3, ≤ 512k input** | **$0.30** | **$1.20** | **$0.06** |
| MiniMax-M3, > 512k input | $0.60 | $2.40 | $0.12 |
| MiniMax-M2.7 | $0.30 | $1.20 | $0.06 |
| **MiniMax-M2.7-highspeed** | **$0.60** | **$2.40** | $0.06 |

### 1. The M3 row is the long-context tier (codex, P2 — confirmed)

```rust
model_id: "minimax-m3",
input_per_mtok_usd: 0.60,
cached_input_per_mtok_usd: 0.12,
output_per_mtok_usd: 2.40,
```

Those are the **> 512k** rates. `estimate_cost_usd` applies the row unconditionally, so every ordinary M3 turn — which is to say almost all of them, against a 1M window — is reported at **2× actual cost**. Either model the input-length tier, or use the standard-tier rates for the static row and accept under-reporting only on genuine long-context turns. The second is the smaller change and errs in the less alarming direction.

### 2. `minimax-m2.7-highspeed` silently inherits half its price (codex, P2 — confirmed)

The vendor lists it as a separate model at $0.60/$2.40. With only the base `minimax-m2.7` row present, `lookup("minimax/minimax-m2.7-highspeed")` falls through to the substring match and lands on the base row, so highspeed usage is estimated at **half** actual. Needs an explicit row (note its cache-read is $0.06, i.e. *not* 2× the base — the "cache read = 0.2× input" comment at the top of your block does not hold for this one).

### 3. `model_accepts_image_input` doesn't normalize like `lookup` (greptile, P2 — confirmed from the code)

```rust
const VISION_INPUT_MODELS: &[&str] = &["minimax-m3"];
pub fn model_accepts_image_input(model_id: &str) -> bool {
    VISION_INPUT_MODELS.contains(&model_id)
}
```

Raw `contains` against the bare id, but the function is `pub` and sits beside `lookup()`, which lowercases, strips the `vendor/` prefix and strips `:tag` / `@date` decorations. So a caller who reasonably mirrors the `lookup()` calling convention — `model_accepts_image_input("minimax/minimax-m3")` or `("MiniMax-M3")` — gets `false` and silently loses the vision capability. Reusing the same normalization step keeps the two public entry points consistent.

### Also

- Conflicts with `main` in `src/openhuman/platform/cost/catalog.rs` — ordinary content drift, not the `_part_NN` file split, so a normal rebase should do it.
- The four red checks are all **CANCELLED** (superseded runs), not failures.

The underlying point — that BYOK MiniMax turns currently record a $0 estimate because the catalog has no rows — is real and worth fixing. It just needs the numbers to be right, since a cost catalog that is confidently wrong is worse than one that is empty.

octo-patch:
Thanks for the review. I added MiniMax-M3 tier-aware pricing at the 512K boundary, an explicit MiniMax-M2.7-highspeed row, normalized vision lookup, and the corresponding tests, then pushed the branch after syncing current main. `git diff --check` passed; the Rust toolchain was unavailable in this worker environment.

PR #5285, opened 2026-07-30
fix(flows): translate n8n `$json` bindings to the real engine scope root

graycyrus:
## Summary

- Fixes a **silent data-loss bug** in the n8n importer: every imported workflow using `$json` expressions saved cleanly, reported no warnings, and then ran hollow.
- `={{ $json.email }}` now translates to `=.item.email` (was `=.email`, which resolves to `null` at runtime). `={{ $json }}` becomes `=.item` (was `=.`).
- Warns on duplicate n8n node names, which previously mis-wired edges silently.
- Corrects an existing test that **pinned the bug**, plus a second test that was quietly relying on the broken output.

## Problem

The tinyflows expression scope root is `{item, items, run, nodes}` (see `expr_scope_for` in `vendor/tinyflows/src/nodes/mod.rs`). n8n's `$json` means *the current input item*, which the engine binds as `item`.

The importer translated the `$json` tail directly onto the scope root, so `$json.email` became `=.email` — dereferencing a key that does not exist at the root. jq yields `null` rather than erroring, so:

1. The import reported **no warning** (these took the "trivially translatable" clean path).
2. The graph saved and enabled normally.
3. Every translated binding resolved `null` at runtime, producing empty tool args and hollow runs.

This is exactly the silent-null-binding class the domain's propose/save gates (`validate_binding_resolvability`, the `dry_run_workflow` null-resolution gate) were built to prevent — and the importer bypassed all of them.

Worse, `translates_trivial_json_expression_to_jq` asserted `=.email` **and** `warnings.is_empty()`, so the bug was pinned as intended behaviour.

## Solution

- `json_path_to_jq` returns the tail path; `translate_expr` prepends `.item` at the single call site, so the empty-tail case naturally becomes `.item` with no special-casing.
- Bracket-key quoting is preserved: `={{ $json["first name"] }}` → `=.item."first name"`.
- Duplicate-name collisions in `name_to_id` now push a warning naming both ids, matching how every other approximation in this importer already behaves.

**Deliberately forward-only.** Already-imported flows keep their broken bindings. Auto-rewriting stored graphs was rejected: it cannot distinguish an importer artifact from a hand-authored root-level jq program, so it risks corrupting valid graphs. Detecting affected saved graphs via the existing validation banner is the suggested follow-up.

**Worth reviewers' attention:** `validate_binding_resolvability`'s regex only matches `=nodes.<ref>.item...` cross-node references — it does **not** statically inspect plain `.item.<field>` bindings, so that gate would not have caught this on its own. The tests that actually prove the fix evaluate a translated expression against a realistic `{item, items, run, nodes}` scope, keeping the old form as a negative control that resolves `null`.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — small, directly-tested diff in a single module; `cargo test --lib openhuman::flows::n8n_import` = 18 passed, 0 failed
- [x] Coverage matrix updated — `N/A: behaviour-only bug fix, no feature rows added/removed/renamed`
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no matrix feature rows affected`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: importer-internal expression translation, not a release-cut surface`

## Impact

- **Runtime/platform:** Rust core only (`openhuman::flows::n8n_import`). No frontend, no CLI surface change.
- **Compatibility:** newly-imported workflows now bind correctly. Previously-imported workflows are unchanged and still carry the broken bindings — see the forward-only note above.
- **Security/migration:** none. No schema change, no `Cargo.toml`/`Cargo.lock` change.

## Related

- Closes: `N/A`
- Follow-up PR(s)/TODOs: surface already-imported flows with unresolvable bindings via the existing validation banner.
- **Merge order:** independent of the other flows PRs in this batch — touches only `n8n_import.rs`, no file overlap.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/flows-n8n-json-item-binding`
- Commit SHA: `ca8e4575f`

### Validation Run

- [x] Focused tests: `GGML_NATIVE=OFF cargo test --lib openhuman::flows::n8n_import` → **18 passed, 0 failed**

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: n8n `$json` expressions translate to the real engine scope root, so imported bindings resolve real values instead of `null`.
- User-visible effect: newly-imported n8n workflows actually work. Duplicate node names now produce a visible import warning instead of silently mis-wiring edges.

### Parity Contract

- Legacy behavior preserved: all other translation paths (non-`$json`, untranslatable `=`-expressions, unmapped node types, extra-trigger demotion) are unchanged and still warn exactly as before.
- Guard/fallback/dispatch parity checks: the untranslatable-expression fallback (raw passthrough + warning) is untouched; only the `$json` fast path changed.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `$json` expression handling for more reliable data access in imported workflows.
  * Added warnings when imported nodes have duplicate names, including the conflicting identifiers.
  * Preserved connection behavior when duplicate node names are encountered.

### DIFF
--- a/src/openhuman/flows/n8n_import.rs
+++ b/src/openhuman/flows/n8n_import.rs
@@ -101,7 +101,32 @@ pub(crate) fn map_n8n_workflow(value: &Value) -> Result<N8nImportResult, String>
             .and_then(Value::as_str)
             .map(str::to_string)
             .unwrap_or_else(|| slug(&n8n_name));
-        name_to_id.insert(n8n_name.clone(), Value::String(id.clone()));
+        // n8n's `connections` map is keyed by node NAME (not id), and
+        // `map_connections` below rewires every edge that names this node
+        // through this lookup. A duplicate name is last-wins here, which
+        // silently mis-wires every connection naming it onto whichever node
+        // happened to be inserted last — R-m6. Every other approximation in
+        // this importer reports a warning; a duplicate name is the one
+        // silent mis-wiring, so warn on the collision the same way.
+        if let Some(previous_id) = name_to_id
+            .insert(n8n_name.clone(), Value::String(id.clone()))
+            .and_then(|v| v.as_str().map(str::to_string))
+        {
+            tracing::warn!(
+                target: "flows",
+                name = %n8n_name,
+                previous_id = %previous_id,
+                new_id = %id,
+                "[flows] n8n_import: duplicate node name collision — connections will mis-wire"
+            );
+            warnings.push(format!(
+                "Multiple n8n nodes are named '{n8n_name}' (ids '{previous_id}' and '{id}') — \
+                 n8n connections are keyed by node name, so every connection naming \
+                 '{n8n_name}' was rewired onto node '{id}' and may now point at the wrong \
+                 node. Rename the duplicates in the source n8n workflow and re-import, or fix \
+                 the affected edges by hand."
+            ));
+        }
 
         let n8n_type = raw.get("type").and_then(Value::as_str).unwrap_or("");
         let params = raw.get("parameters").cloned().unwrap_or(Value::Null);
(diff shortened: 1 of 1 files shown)


graycyrus:
Review follow-up: **clean, no changes applied** — the translation was independently verified correct for every shape (dotted paths, nested dot+bracket, quoted keys, bare `{{ $json }}`); anything that isn't a clean dotted/bracket path falls through to the safe raw+warn path rather than translating wrongly. The corrected tests were confirmed non-tautological: they evaluate through the real `tinyflows::expr` engine, and the negative control genuinely resolves `null`.

**One real gap found, recommended as a follow-up rather than a blocker.**

`.item.<field>` is correct immediately downstream of every node kind this importer emits *except* `httpRequest`. `HttpRequestNode::execute` always wraps its result in the `{json, text, raw}` envelope, so for a node directly downstream of an imported HTTP Request, the current item is the envelope — and a top-level `.item.email` dereferences a key that does not exist, resolving `null`. That is the same failure class this PR fixes, one hop further along, and it hits a very common n8n shape: Webhook → HTTP Request → Set/If/Code reading `$json.<field>`.

Two things make it a follow-up rather than a fix here:

1. The existing `ENVELOPING_KINDS` gate in `validate_binding_resolvability` only fires for a `tool_call`'s `args` addressed via the cross-node `=nodes.<ref>.item...` form. It does not scan a plain direct-predecessor `.item.<field>` binding, which is exactly what the importer emits — so neither this PR nor the existing gate catches it.
2. Even `.item.json.<field>` would not be enough today: `OpenHumanHttp::request` returns `{"text": ...}`, so the response body is never decomposed into structured fields. A real fix needs either response-body JSON parsing at the capability layer or a `fromjson` step — plus topology awareness the importer lacks (node configs are translated before `map_connections` builds the edge graph).

This PR is strictly better than before in every case — the pre-fix code was wrong here too, missing both `.item` and `.json` — so it does not regress anything. Filing the envelope gap separately.

PR #5286, opened 2026-07-30
fix(flows): give `flows_resume` the run-lifecycle safety `flows_run` already had

graycyrus:
## Summary

- Gives `flows_resume` the run-lifecycle safety `flows_run` has had since B41/B42 — it had **none** of it, despite executing a flow's real approved side effects for up to `FLOW_RUN_TIMEOUT_SECS`.
- Makes the terminal run-row write **guarded**, so a settled run can no longer be relabelled by a losing concurrent cancel.
- Arms the B42 drop-guard *before* the first `.await` in `run_flow_body`, closing a window where a dropped future stranded a `running` row until the next process boot.
- The parked-run TTL sweep now publishes `FlowRunFinished` — it was the one terminal path that emitted no event.

## Problem

Four independent defects, all in the same lifecycle area:

**1. `flows_resume` never registered in the run registry.** The row stayed `pending_approval` for the whole resume and no cancellation token existed, so:
- `flows_cancel_run` saw `is_in_flight == false`, took its "parked/stale" branch, wrote a terminal `cancelled` row **and dropped the durable checkpoint** — while the resume kept executing real outbound nodes and finally overwrote `cancelled` with its own status.
- A run approved just before its TTL was expired mid-execution by `sweep_expired_parked_runs`, which matches on `status = 'pending_approval'` — a status the resume never changed.

`run_flow_body`'s own doc comment describes exactly this race; it was fixed for the detached run path and never applied to resume.

**2. `store::finish_flow_run` was an unconditional `UPDATE … WHERE id = ?`.** `flows_cancel_run` reads the status and consults the registry as two separate observations. A run that settles in that window is not in flight, so the not-in-flight branch relabelled a **fully-completed run whose real side effects had fired** as `cancelled`.

**3. `flows_resume`'s write order was inverted vs `flows_run`** — `record_run(...)?` ran *before* `finish_flow_run_row`. A flow deleted mid-resume made the summary write fail and returned early, stranding the row at `pending_approval` even though the engine had completed; the TTL sweep later relabelled that completed run `cancelled`. `flows_run` does the opposite, with an explicit comment saying why.

**4. The `RunRowFinalizer` was constructed ~150 lines after the first `.await`** in `run_flow_body`, leaving the inference-readiness network probe unguarded: a client disconnect there dropped the future before any finalizer existed, and the row stayed a perpetual `running` spinner.

## Solution

- **Guarded terminal write** — `finish_flow_run` now carries `status IN ('running','pending_approval')` and returns whether it landed, mirroring the re-check `expire_parked_runs`/`mark_run_interrupted` already do. `flows_cancel_run` attempts that write **first** and treats it as the authority: a `false` means the run settled underneath us, so it reports the conflict and leaves the recorded outcome (and the checkpoint) alone.
- **Resume registers + claims** — `run_registry::register` before anything else, a new guarded `store::mark_run_resuming` (`pending_approval` → `running`) that moves the row out of the TTL sweep's predicate, and the cancel token honoured in a `biased` select exactly as `run_flow_body` does.
- **Write order fixed + finalizer added** on the resume path; `record_run` failures are logged, never propagated with `?`.
- **Finalizer armed before the first await**; a third early-return path that was missing its `disarm` is also fixed.

### Note for reviewers: two pre-existing tests changed, and it is not a regression

Adding the guard failed `flows_cancel_run_of_a_completed_with_warnings_run_errors` and `flows_cancel_run_of_an_interrupted_run_errors`. Both **staged** a row at an arbitrary terminal status by calling `finish_flow_run` twice — precisely the terminal→terminal overwrite the guard exists to block. The guard was correct; the fixtures were leaning on an unguarded production write. Staging therefore gets a `#[cfg(test)] force_run_status_for_test` door rather than the production path being weakened.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — 5 new tests target each changed mechanism directly; `cargo test --lib openhuman::flows` = 554 passed, 0 failed
- [x] Coverage matrix updated — `N/A: behaviour-only bug fix, no feature rows added/removed/renamed`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no user-facing surface change; run/resume/cancel semantics only become more correct`

## Impact

- **Runtime/platform:** Rust core only. No frontend, no schema migration (the new `mark_run_resuming` reuses existing columns).
- **Behaviour:** a resumed run now shows as `running` for its duration rather than staying `pending_approval`. Verify the runs rail renders that sensibly — it is an existing status, and `pending_approvals_json` is not cleared by the flip.
- **Security:** cancelling an already-settled run now returns an error instead of silently rewriting history. That is a deliberate, user-visible behaviour change.
- **Risk:** this edits the currently-healthy `flows_run` path (finalizer placement + its terminal-write helper). The existing `flows_run` suite is the regression net and stays green.

## Related

- Closes: `N/A`
- Follow-up PR(s)/TODOs: `flows_resume` graph-hash pin (stale-approval-after-graph-swap) is stacked on this branch; store resilience (skip corrupt rows) and the detached-run RPC also stack on it.
- **Merge order:** overlaps `src/openhuman/flows/ops.rs` + `ops_tests.rs` with the authorization-boundaries PR, and `store.rs` with the contract-drift PR. Merge this one **first**; the others rebase cleanly on it.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/flows-resume-run-lifecycle`
- Commit SHA: `0b7105fa7`

### Validation Run

- [x] Focused tests: `GGML_NATIVE=OFF cargo test --lib openhuman::flows` → **554 passed, 0 failed**

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: a resumed run is registered, cancellable, and immune to the parked-run TTL sweep; terminal run statuses can no longer be overwritten once settled.
- User-visible effect: a resumed run reports `running` while it executes; cancelling an already-finished run now returns a clear error instead of silently relabelling it.

### Parity Contract

- Legacy behavior preserved: `flows_run` / `flows_run_detached` semantics are unchanged apart from the finalizer being armed earlier; every existing terminal-status transition (`running`/`pending_approval` → terminal) still lands.
- Guard/fallback/dispatch parity checks: the boot orphan sweep, `expire_parked_runs`, and `mark_run_interrupted` guards are untouched and still pinned by their existing tests.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flow-run finalization to prevent orphaned or incorrectly overwritten run records.
  * Added safer handling for resumed runs, cancellations, timeouts, and concurrent state changes.
  * Prevented duplicate completion events and ensured expired parked runs are reported correctly.
  * Improved cancellation tracking when multiple execution attempts overlap.
* **Tests**
  * Added coverage for run resumption, cancellation races, duplicate registrations, and parked-run expiration.

### DIFF
--- a/src/openhuman/flows/ops.rs
+++ b/src/openhuman/flows/ops.rs
@@ -4707,6 +4707,20 @@ async fn run_flow_body(
     let config: &Config = config_arc.as_ref();
     let flow_id: &str = flow_id.as_str();
 
+    // B42 drop-guard, armed BEFORE the first `.await` in this body (R-M5).
+    //
+    // The caller has already inserted the `running` row, so every await from
+    // here on is a window in which dropping this future would strand that row.
+    // The guard used to be constructed ~150 lines below, immediately around the
+    // engine call — which left the inference-readiness preflight directly below
+    // (a real network probe on a cache miss) unguarded: a client disconnect or
+    // an aborted detached task during that probe dropped the future before any
+    // finalizer existed, and the row stayed a perpetual `running` spinner until
+    // the NEXT process boot sweep (the in-process one had already run). Arming
+    // it here covers the whole awaiting region; every settled path below still
+    // disarms it after its own terminal write.
+    let finalizer = RunRowFinalizer::new(config_arc.clone(), &thread_id, flow_id);
+
     // B45 run-time preflight (design correction — see the "Inference-readiness
     // check" module doc above): an `agent` node needs a working LLM provider
     // to run at all, but that is no longer enforced as an author-time gate —
@@ -4752,6 +4766,7 @@ async fn run_flow_body(
             &[],
             Some(&msg),
         );
+        finalizer.disarm();
         return Err(msg);
     }
 
@@ -4773,6 +4788,7 @@ async fn run_flow_body(
                 &[],
                 Some(&msg),
             );
+            finalizer.disarm();
             return Err(msg);
         }
     };
@@ -4797,6 +4813,7 @@ async fn run_flow_body(
                 &[],
                 Some(&msg),
             );
+            finalizer.disarm();
             return Err(msg);
         }
     };
@@ -4867,11 +4884,8 @@ async fn run_flow_body(
     );
     let timed = tokio::time::timeout(std::time::Duration::from_secs(FLOW_RUN_TIMEOUT_SECS), run);
     tokio::pin!(timed);
-    // B42 drop-guard: armed for the whole awaiting region below. If this future
-    // is dropped before any terminal write (harness abort, turn end, runtime
-    // shutdown, panic), its `Drop` reconciles the orphaned `running` row to
-    // `interrupted`. Every settled path disarms it after its own terminal write.
-    let finalizer = RunRowFinalizer::new(config_arc.clone(), &thread_id, flow_id);
+    // (The B42 drop-guard is armed near the top of this fn, before the first
+    // `.await` — see `finalizer` there.)
     // Race the run against a cancellation signal (issue G4). `biased` checks the
     // cancel arm first so a `flows_cancel_run` that lands right as the run
     // settles still wins deterministically.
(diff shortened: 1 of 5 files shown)


graycyrus:
Review follow-up: two real bugs found and fixed, both of which this PR made load-bearing.

**1. `expire_parked_runs` reported candidates, not actual sweeps.** Its `SELECT` and each row's guarded `UPDATE` are separate statements on an autocommit connection, so a concurrent `mark_run_resuming` can claim a row in between. The per-row `WHERE status = 'pending_approval'` kept that row's data safe, but the function returned the unfiltered candidate list — so the caller acted on runs it never expired: dropping the checkpoint out from under a live resume, and (new in this PR, via the R-m4 `FlowRunFinished` publish) emitting a terminal event for a run still executing. That false event is the worse half: the frontend de-dupes terminal events by `${flow_id}:${run_id}`, so the run's real completion would later be discarded as an alias replay and a successful run would display as cancelled. Now only rows whose `UPDATE` reports `changed > 0` are returned. Pinned by `expire_parked_runs_returns_only_rows_it_actually_flipped`.

**2. `RunGuard` deregistered by key, not by identity.** The registry documented duplicate registration as impossible because "thread ids are UUID-suffixed" — true while only `flows_run`/`flows_run_detached` registered, since both mint a fresh UUID per call. **This PR is the first caller to register against a stable, pre-existing id** (the parked run's own), and nothing serializes two concurrent resumes of the same run — a client double-submit or retry-on-timeout suffices. The loser of that race would deregister the *winner's* live token on its way out, after which `flows_cancel_run` would see `is_in_flight == false` for a genuinely executing run, take its parked/stale branch, and drop the checkpoint mid-execution: exactly the bug class this PR exists to eliminate. Registrations now carry a monotonic id and a guard only removes the entry it installed. Pinned by `a_displaced_guard_does_not_deregister_the_live_registration`.

557 flows tests pass, `cargo fmt --check` clean.

Also noted for follow-up, not changed here: `app/src/hooks/useRunsPendingApprovalSet.ts`'s doc comment claims "the core has no separate DB status for 'parked at an approval gate' — a run stays \`status: 'running'\`", which is stale now that `flow_runs.status` really can be `pending_approval`. The hook derives display status from the approval queue so its logic looks unaffected, but the comment should be corrected.

PR #5294, opened 2026-07-30
fix(flows): store resilience — skip corrupt rows, transactional step upserts, once-per-process schema init

graycyrus:
> **Stacked on #5286** (`fix/flows-resume-run-lifecycle`). This branch contains that PR's commit as its base. **Review only the second commit**, and merge after #5286.

## Summary

- **One corrupt or newer-schema row could brick the entire flows surface.** `list_flows`, all `app_event` trigger dispatch, and the boot schedule reconcile each hard-failed on the first bad row.
- Corrupt rows are now skipped and **loudly counted**, not silently omitted.
- Makes the per-node step upsert transactional, so parallel branches can no longer lose a step.
- Stops re-running the full DDL batch on every single store call.

## Problem

**R-M4 (major).** `map_flow_row` propagates migrate/deserialize errors as row errors, and `list_flows` / `list_enabled_flows` did `flows.push(row?)` — so the **first** bad row failed the whole query.

The realistic trigger is a downgrade: a user runs a newer build that persists a graph at a newer `schema_version`, then goes back. `tinyflows::migrate::migrate` cannot downgrade, so that single row breaks:
- every `flows_list` (the Workflows page),
- every `list_enabled_flows` — which drives **all** `app_event` trigger dispatch in `bus.rs`,
- and `reconcile_schedule_triggers_on_boot`.

The whole flows surface goes down because of one row. The sibling `draft_store::list_drafts` already skips-and-logs corrupt entries; the SQLite store had no such tolerance.

**R-m1.** `upsert_flow_run_step` was an untransacted read-modify-write (SELECT `steps_json` → mutate in memory → UPDATE) on a fresh connection. Two observer callbacks for parallel branch nodes could interleave — both read `[A]`, one writes `[A,B]`, the other `[A,C]` — and B vanished from the live view. Its `status`/`duration_ms` were lost permanently, since the post-hoc `settle_steps` reconstruction refills a node only with `status: None`.

**R-m8.** `with_connection` opened a new connection and re-ran the **full** DDL batch (5 `CREATE TABLE` + 6 `CREATE INDEX` + `PRAGMA journal_mode=WAL` + a `PRAGMA table_info` migration probe) on **every** call — including the per-step upsert fired for every node of every live run. Idempotent, but real churn on hot runs.

## Solution

- **Skip, but loudly.** A new `list_flow_rows` helper decodes rows one at a time and skips/logs any that fail to parse or migrate (logging the id and error, **never** `graph_json`). `list_flows`/`list_enabled_flows` return `(Vec<Flow>, usize)`, and the skip count is surfaced rather than swallowed: `flows_list` puts an "N workflows could not be loaded" line in `RpcOutcome.logs`, and the boot reconcile + `bus.rs` app-event path each `warn!`. A silently short flow list would be a worse failure mode than a hard error, so the count is the point.
- **Transactional upsert** via a new `with_immediate_transaction` helper (`BEGIN IMMEDIATE` / `COMMIT` / `ROLLBACK` through `execute_batch`, since `Connection::transaction()` needs `&mut Connection` and does not fit `with_connection`'s `&Connection` closure). The existing `busy_timeout = 5000` pragma covers the lock wait.
- **Schema init once per process, per database path**, gated by `OnceLock<Mutex<HashSet<PathBuf>>>`. Keyed by path rather than a single global flag, so each test's distinct `TempDir` workspace still initializes correctly. Per-connection pragmas (`busy_timeout`, `foreign_keys`) still reapply on every open, since those are not persisted in the db file.

A `#[cfg(test)] force_corrupt_graph_json_for_test` fixture door (mirroring the existing `force_run_status_for_test`) stages corrupt/newer-schema rows for the tests.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — 9 new tests: corrupt-row skip + count on both list functions, corrupt-disabled-row not counted, two concurrent-upsert races (different and identical node ids), fresh-db schema init, independent init across two db paths, and boot reconcile surviving a corrupt row. `cargo test --lib openhuman::flows` = 563 passed, 0 failed
- [x] Coverage matrix updated — `N/A: resilience fix to existing paths, no feature rows added/removed/renamed`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface behaviour change on the healthy path`

## Impact

- **Runtime/platform:** Rust core only.
- **Behaviour:** a corrupt/newer-schema row no longer takes down listing, trigger dispatch, or boot reconcile. It is skipped, and the count is reported so the UI can say "N workflows could not be loaded" rather than quietly showing a short list.
- **API shape:** `list_flows` / `list_enabled_flows` now return `(Vec<Flow>, usize)`. All in-repo callers are updated.
- **Performance:** removes a full DDL batch + migration probe from every store call, including the per-node step write on every live run.
- **Concurrency:** `BEGIN IMMEDIATE` introduces a short write lock on step upserts; the transaction body is kept minimal and the existing busy timeout covers it.

## Related

- Closes: `N/A`
- Follow-up PR(s)/TODOs: surface the skipped-row count in the Workflows UI (the backend now provides it); consider an explicit "this workflow was saved by a newer version" state rather than a generic skip.
- **Merge order:** stacked on #5286 — merge that first. Also touches `store.rs`, which #5290 edits; rebase if that lands first.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/flows-store-resilience`
- Commit SHA: `4618fdf96` (plus #5286's `0b7105fa7` as its base)

### Validation Run

- [x] Focused tests: `GGML_NATIVE=OFF cargo test --lib openhuman::flows` → **563 passed, 0 failed**

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: an undecodable flow row is skipped and counted instead of failing the whole query; step persistence is atomic; schema DDL runs once per process per database.
- User-visible effect: the Workflows page, app-event triggers, and boot reconcile keep working when one row cannot be decoded, and the user is told how many were skipped.

### Parity Contract

- Legacy behavior preserved: on a healthy database every list returns exactly the same flows in the same order, with a skip count of 0; the retention prune, WAL/busy-timeout pragmas, and `add_column_if_missing` migrations are unchanged.
- Guard/fallback/dispatch parity checks: `list_enabled_flows` still gates app-event dispatch identically for decodable rows; only the undecodable-row path differs.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

### DIFF
--- a/src/openhuman/flows/bus.rs
+++ b/src/openhuman/flows/bus.rs
@@ -134,13 +134,18 @@ impl FlowTriggerSubscriber {
     /// dispatches each match with the event payload as the run input
     /// (seeded into `run.trigger`, per the node-catalog contract).
     async fn handle_app_event(&self, toolkit: &str, trigger_slug: &str, payload: &Value) {
-        let flows = match store::list_enabled_flows(&self.config) {
-            Ok(flows) => flows,
+        let (flows, skipped) = match store::list_enabled_flows(&self.config) {
+            Ok(result) => result,
             Err(e) => {
                 tracing::warn!(target: "flows", %toolkit, %trigger_slug, error = %e, "[flows] failed to list enabled flows for app_event dispatch");
                 return;
             }
         };
+        if skipped > 0 {
+            // R-M4: one corrupt/unmigratable flow row must not blackhole
+            // app_event dispatch for every other enabled flow.
+            tracing::warn!(target: "flows", %toolkit, %trigger_slug, skipped, "[flows] handle_app_event: skipped corrupt/unmigratable flow rows while matching trigger");
+        }
 
         let mut matched = 0usize;
         for flow in flows {
@@ -1039,8 +1044,10 @@ mod tests {
         // `list_enabled_flows` must not surface the disabled flow at all —
         // proves the subscriber's dispatch source already excludes it,
         // rather than asserting on a spawned background task's side effect.
-        let enabled = crate::openhuman::flows::store::list_enabled_flows(&config).unwrap();
+        let (enabled, skipped) =
+            crate::openhuman::flows::store::list_enabled_flows(&config).unwrap();
         assert!(enabled.is_empty());
+        assert_eq!(skipped, 0);
     }
 
     #[tokio::test]
--- a/src/openhuman/flows/ops.rs
+++ b/src/openhuman/flows/ops.rs
@@ -3498,9 +3498,32 @@ pub(crate) fn load_engine_compatible_flow_graph(
 }
 
 /// Lists every saved flow.
+///
+/// A corrupt or newer-schema-than-this-build `graph_json` row is skipped
+/// rather than failing the whole list (R-M4 — see `store::list_flow_rows`);
+/// when that happens it must not be silent, so a skip is both logged
+/// (`[flows]`-prefixed, id + error only — never row content) and surfaced in
+/// the RPC's `logs` so the UI can tell the user "N workflows could not be
+/// loaded" instead of silently rendering a shorter list than actually exists.
 pub async fn flows_list(config: &Config) -> Result<RpcOutcome<Vec<Flow>>, String> {
-    let flows = store::list_flows(config).map_err(|e| e.to_string())?;
-    Ok(RpcOutcome::single_log(flows, "flows listed"))
+    let (flows, skipped) = store::list_flows(config).map_err(|e| e.to_string())?;
+    if skipped > 0 {
+        tracing::warn!(
+            target: "flows",
+            skipped,
+            loaded = flows.len(),
+            "[flows] flows_list: skipped corrupt/unmigratable flow_definitions rows"
+        );
+        Ok(RpcOutcome::new(
+            flows,
+            vec![format!(
+                "flows listed ({skipped} workflow{} could not be loaded and were skipped)",
+                if skipped == 1 { "" } else { "s" }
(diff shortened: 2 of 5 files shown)


graycyrus:
Review follow-up: fixed a regression this PR introduced.

Gating the schema DDL behind a per-path "already initialized" set cost the store its **self-healing**. Previously the DDL ran on every `with_connection` call, so a database deleted or replaced at runtime (workspace reset, manual deletion, a restore) recovered on the very next call — `Connection::open` creates a fresh empty file and `CREATE TABLE IF NOT EXISTS` repopulated it. With the cache, the set still reported "initialized" while the file behind it was empty, and every subsequent query failed `no such table: flow_definitions` until the process restarted.

Reproduced directly: after deleting `flows.db` mid-process, `list_flows` returned `Err(no such table: flow_definitions)`.

Fix: a cache hit is now confirmed against the file on disk via one indexed `sqlite_master` lookup before it is honoured, re-running init (with a warning) when the tables are gone. That keeps the R-m8 win — the ~11-statement DDL batch no longer runs per call — while restoring self-healing at the cost of a single cheap query. Pinned by `schema_reinitializes_when_the_database_file_is_deleted_at_runtime`, which fails without the fix.

564 flows tests pass, `cargo fmt --check` clean.

PR #5296, opened 2026-07-30
fix(flows): expose flows_run_detached over RPC and switch both UI Run controls to it

graycyrus:
> **Stacked on #5286** (`fix/flows-resume-run-lifecycle`). This branch contains that PR's commit as its base. **Review only the second commit**, and merge after #5286.

## Summary

- The canvas live-run overlay **could never show live progress**, and one Run click **froze every other row** on the Workflows page for the whole run. Both had the same root cause.
- Exposes the already-existing `flows_run_detached` over RPC and switches both UI Run controls to it.
- Also fixes three canvas defects that live in the same files: a mid-conversation blank-canvas reveal, a false "Unsaved" badge, and a stale doc claim.

## Problem

`openhuman.flows_run` **blocks server-side** until the run terminates (up to `FLOW_RUN_TIMEOUT_SECS`; the client mirrors it with a 610s timeout). Both Run entry points awaited it as though it were fire-and-forget:

- **`FlowCanvasPage.tsx`** set `activeRunId` from the RPC's *return value*, but `useFlowRunProgress` only subscribes to `flow:run_progress` once `activeRunId` is set — i.e. **after** every progress event has already fired and been dropped. So nodes never ringed running/success/error; the overlay only ever worked for the rare `pending_approval` pause.
- **`FlowsPage.tsx`** awaited it under a **page-global** `busyKey`, so a 5-minute run disabled Run/Toggle on **every** row for 5 minutes and fired the "Run started" toast only when the run **ended**. Its inline comment claimed "Fire-and-forget: the caller doesn't wait for the run to finish" — which did not match the RPC's semantics.

`ops::flows_run_detached` already existed and returns `{run_id, detached:true}` immediately, registering the run **before** returning its id — but it was never exposed over RPC. Its only caller was the agent `run_flow` tool, and its doc comment claimed it "mirrors how the UI 'Run' control … fire runs fire-and-forget", which was simply false.

## Solution

- **Backend:** registers `flows.run_detached` (wire `openhuman.flows_run_detached`) as a thin controller over the existing `ops::flows_run_detached`. `flows.run` stays registered — the agent tool and existing callers still use the blocking form. The false doc comment is corrected rather than deleted, since it becomes true with this change.
- **Frontend:** `runFlowDetached()` uses the **default** timeout (it returns immediately, so the 610s budget is wrong for it). `FlowCanvasPage` sets `activeRunId` from the immediate response, so the progress subscription is established as early as it can be.

  To be precise rather than overclaim: this does **not** make the window strictly zero. `flows_run_detached` spawns the run and returns without awaiting it, so the first `FlowRunProgress` could in principle be published before the HTTP response is even sent. In practice the round-trip plus a React commit is far slower than a `tokio::spawn` scheduling gap, and the consequence is cosmetic — at worst the first node's animation frame is missed, and `useFlowRunPoller`'s 2s durable-row fallback recovers it. Run history is never affected. `FlowsPage` tracks busy state per row (`busyByFlow: Record<flowId, 'toggle'|'run'>`) instead of one global value.
- **Completion signal preserved:** since the RPC no longer blocks until completion, `FlowsPage` now uses `useFlowRunFinished` — the same hook the runs sidebar/drawer already use — to refetch when a run settles, so `last_run_at`/`last_status` still update. The runs rail itself is untouched; it already reconciled detached runs from the agent-initiated path.

### Also fixed (same files, so they belong here)

- **F-m2** — `chatFirst` was re-derived from the live `initialBuildSeed` prop every render, so `clearBuildSeed` flipped it false and revealed the blank trigger-only canvas mid-conversation whenever the builder's first turn ended with a clarifying question. Now latched at mount.
- **F-m3** — false "Unsaved" after a canvas remount. The mount-time `onGraphChange` writes concrete auto-layout positions for nodes the server stored **without** `position`, so opening an agent-built flow and rejecting a proposal showed a dirty badge with zero user edits. Both sides of the dirty comparison are now normalized at comparison time. (A first attempt that pre-normalized only the persisted seed made the *first* mount read dirty and broke two tests — normalizing both operands inside the memo is correct for first-mount and remount alike.)
- **F-m5** — `ReadonlyFlowCanvas`'s doc claimed the `/flows/:id` viewer uses it, but no production caller reaches it. **Doc corrected rather than deleted**: it has 7 passing tests, and discarding that coverage for a cleanliness win was out of proportion to this PR's scope. Left as a documented, inert fallback with a pointer to remove it if no consumer ever appears.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — Rust controller coverage in `tests/json_rpc_e2e.rs`; Vitest covers the detached call not using the long timeout, FlowsPage keeping other rows interactive (the F-M2 regression), and FlowCanvasPage setting `activeRunId` before any progress event
- [x] Coverage matrix updated — `N/A: bug fix to existing Run controls, no feature rows added/removed/renamed`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — Run is a release-cut surface, but its **user-visible contract is unchanged** (click Run, see it run); only the transport changes. Worth a manual pass on the canvas overlay at review time.

## Impact

- **Runtime/platform:** Rust core (new controller) + desktop frontend.
- **Behaviour:** the canvas overlay now actually animates node state during a run, and the Workflows list stays usable while a run is in flight. The "Run started" toast now fires when the run *starts*.
- **API:** adds `openhuman.flows_run_detached`. Purely additive; `openhuman.flows_run` is unchanged and still registered.
- **Risk:** the completion signal now arrives via the event/poll path rather than the RPC return. Review found `FlowsPage` had only the event half of that pair, unlike the three other run-outcome surfaces which also carry a poll backstop; a bounded backstop has since been added and pinned by tests. Still the behaviour worth exercising manually, since it is where a regression would hide.
- **i18n:** no new strings.

## Related

- Closes: `N/A`
- Follow-up PR(s)/TODOs: delete `ReadonlyFlowCanvas` if no consumer materializes.
- **Merge order:** stacked on #5286 — merge that first. Touches `ops.rs`/`schemas.rs`, so it overlaps #5287 and the stacked #5293/#5294; rebase as those land.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `feat/flows-run-detached-rpc`
- Commit SHA: `0a4a65b03` (plus #5286's `0b7105fa7` as its base)

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — prettier clean (after `--write` on 4 files); eslint 0 errors on touched files
- [x] `pnpm typecheck` — exit 0
- [x] Focused tests: Vitest sweep `src/pages src/services/api src/components/flows src/lib/flows src/hooks` → **1637 passed, 145 files**; `cargo test --lib openhuman::flows` → **554 passed, 0 failed**
- [x] Rust fmt/check (if changed): `cargo check` clean on BOTH the default build and the disabled build (`--no-default-features --features tokenjuice-treesitter`)

### Validation Blocked

- `command:` `cargo fmt --check`
- `error:` one pre-existing formatting drift at `ops.rs:5241`
- `impact:` none — confirmed via `git show HEAD:...` that it exists in the base commit and is outside this diff, so it was left alone rather than swept into this PR

### Behavior Changes

- Intended behavior change: the UI Run controls dispatch a detached run and observe progress via events instead of blocking on the RPC.
- User-visible effect: live node-state animation on the canvas during a run; the Workflows list stays interactive while a run is in flight; "Run started" fires at start, not at finish.

### Parity Contract

- Legacy behavior preserved: `openhuman.flows_run` is untouched and still registered for the agent tool and any external caller; the runs rail/drawer/`WorkflowRunsPage` are untouched and already handled detached runs.
- Guard/fallback/dispatch parity checks: `flows_run_detached` registers the run before returning its id, so a `flows_cancel_run` landing immediately after the UI receives it still takes the signalled branch.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

### DIFF
--- a/app/src/components/flows/canvas/FlowCanvas.tsx
+++ b/app/src/components/flows/canvas/FlowCanvas.tsx
@@ -9,8 +9,17 @@
  *    {@link EditableFlowCanvas}, which lifts nodes/edges into controlled state
  *    and wires drag/connect/add/delete/save on top.
  *
- * The `editable` prop defaults to `false` so every existing read-only consumer
- * (the `/flows/:id` viewer) keeps its exact behavior — only the builder opts in.
+ * The `editable` prop defaults to `false`, but as of the F-m5 audit
+ * (`my_docs/flows_review_2026-07-30.md`) there is exactly one production
+ * consumer — `FlowCanvasPage.tsx` — and it always passes `editable`. The
+ * `/flows/:id` route this doc block used to claim as the read-only viewer's
+ * consumer IS `FlowCanvasPage`, and it is NOT read-only. `ReadonlyFlowCanvas`
+ * below is therefore currently unreachable in the shipped app; it's kept
+ * (rather than deleted) because it's small, self-contained, and covered by
+ * its own smoke tests (`__tests__/FlowCanvas.test.tsx`) — a future
+ * share/embed/public-viewer surface is a plausible reason to want a
+ * non-editable render path again. Delete both it and its tests together if
+ * no such consumer materializes.
  */
 import {
   Background,
--- a/app/src/components/flows/canvas/__tests__/flowCanvasOutlines.test.ts
+++ b/app/src/components/flows/canvas/__tests__/flowCanvasOutlines.test.ts
@@ -1,0 +1,66 @@
+/**
+ * Guard: the flow canvas draws every one of its status rings with `outline`,
+ * so a global `outline` suppression silently kills all of them at once.
+ *
+ * `app/src/index.css` previously carried `outline: none !important` on the bare
+ * universal selector. Its comment says it exists to hide the browser focus
+ * ring — but `!important` on `*` outranks every component rule, so it also
+ * suppressed the live run overlay (`.flow-node-running` / `-success` /
+ * `-failed`), the validation ring (`.flow-node-error`) and the copilot diff
+ * overlay (`.flow-node-added` / `-removed`). All four rendered nothing, in
+ * every build, with no test failing — the classes were applied correctly to the
+ * DOM the whole time, so only a human looking at the canvas could notice.
+ *
+ * jsdom does not implement the cascade well enough to assert the computed
+ * outline, so this asserts the source invariant instead: focus-ring suppression
+ * must be scoped to a focus state, never applied unconditionally to `*`.
+ */
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+// Vitest runs with `app/` as cwd (see test/vitest.config.ts).
+const INDEX_CSS = resolve(process.cwd(), 'src/index.css');
+const CANVAS_CSS = resolve(process.cwd(), 'src/components/flows/canvas/flowCanvasStyles.css');
+
+/** Strips comments so prose mentioning a rule is never mistaken for the rule. */
+function withoutComments(css: string): string {
+  return css.replace(/\/\*[\s\S]*?\*\//g, '');
+}
+
+describe('flow canvas status rings', () => {
+  it('are not suppressed by an unconditional universal outline reset', () => {
+    const css = withoutComments(readFileSync(INDEX_CSS, 'utf8'));
+    // Find every `*` block that is NOT tied to a focus state.
+    const universalBlocks = [...css.matchAll(/(^|\})\s*\*\s*\{([^}]*)\}/g)].map(m => m[2]);
+    for (const block of universalBlocks) {
+      expect(
+        /outline\s*:/.test(block),
+        'index.css must not set `outline` on the bare `*` selector — with `!important` it ' +
+          'outranks every flow-node ring rule and silently blanks the canvas overlays. ' +
+          'Scope focus-ring suppression to `*:focus` (and the explicit element list) instead.'
+      ).toBe(false);
+    }
+  });
+
+  it('still declare an outline for each run/validation/diff state', () => {
+    const css = withoutComments(readFileSync(CANVAS_CSS, 'utf8'));
+    for (const cls of [
+      'flow-node-running',
+      'flow-node-success',
+      'flow-node-failed',
+      'flow-node-error',
+      'flow-node-added',
+      'flow-node-removed',
+    ]) {
+      const rule = new RegExp(`\\.${cls}\\s*\\{[^}]*outline\\s*:`);
+      expect(rule.test(css), `${cls} must declare an outline`).toBe(true);
+    }
+  });
+
+  it('keeps the running state animated so an executing node is distinguishable', () => {
+    const css = withoutComments(readFileSync(CANVAS_CSS, 'utf8'));
+    expect(/\.flow-node-running\s*\{[^}]*animation\s*:/.test(css)).toBe(true);
+    expect(css).toContain('@keyframes flow-node-run-pulse');
+  });
+});
--- a/app/src/index.css
+++ b/app/src/index.css
@@ -132,9 +132,18 @@
 
   /* Suppress the default browser focus ring (outline + tap highlight) but
      leave borders and box-shadows alone — components rely on those for
-     selected / focus-visible styling (e.g. the runtime picker cards). */
+     selected / focus-visible styling (e.g. the runtime picker cards).
+     Deliberate outlines are left alone too. This block used to carry
+     `outline: none !important` on the bare `*`, which suppressed EVERY outline
+     in the app, not just the focus ring it was aimed at — `!important` on a
+     universal selector outranks any component rule. That silently killed all
+     four flow-canvas rings, which are drawn with `outline`: the live run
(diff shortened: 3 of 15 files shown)


graycyrus:
Review follow-up: added the completion backstop `FlowsPage` was missing — this was the one real regression risk this PR carried, and the PR body flagged it as the thing to watch.

Because `flows_run_detached` returns as soon as the run is registered, a row's `last_run_at`/`last_status` is now refreshed by the `FlowRunFinished` broadcast rather than by the RPC returning. That broadcast is a plain `io.emit` to currently-connected sockets with **no server-side replay** — so a socket drop between clicking Run and the run settling (reconnect gap, sleeping laptop, backgrounded tab) loses it, and the row shows a stale outcome until the user navigates away and back.

Every other run-outcome surface already pairs `useFlowRunFinished` with `useFlowRunsLiveRefresh` for exactly this reason — `FlowRunsSidebar`, `FlowRunsDrawer`, `WorkflowRunsPage` — and that hook's own doc calls the poll "a long-interval backstop for the (rare) case a broadcast event is dropped under lag or a socket reconnect gap." `FlowsPage` became a fourth consumer and was the only one without the backstop half of the pair.

The hook can't be reused verbatim (it is typed for `FlowRun[]`; this page holds `Flow[]`), so the same guarantee is rebuilt against outstanding run ids started from this page. It is **bounded**: entries older than 15 minutes are dropped, comfortably past the engine's ~600s ceiling, so a slow run is never abandoned while one that never reports terminal cannot poll forever. Nothing outstanding means no polling at all.

Two tests pin it: the backstop fires when no `FlowRunFinished` ever arrives, and no poll happens when nothing is outstanding. Both use `advanceTimersByTimeAsync` rather than `waitFor`, which deadlocks under fake timers.

1639 frontend tests pass (145 files), typecheck and prettier clean.

Also confirmed by review and worth recording: the subscribe-before-events window is **not** strictly zero — `flows_run_detached` spawns the run and returns without awaiting, so the first `FlowRunProgress` could in principle precede the client's subscription. Impact is cosmetic (the first node's animation frame) and self-heals within 2s via `useFlowRunPoller`'s durable-row fallback; run history is never affected. I've corrected that overclaim in the PR description rather than leave it standing.

graycyrus:
## Manual testing found the overlay never actually worked — two bugs, both fixed here

Verified live in the running app. The canvas overlay this PR is *about* did not render at all, and tracing it turned up two independent causes.

### 1. `FlowRunObserver` never implemented `on_step_start` (this PR's bug)

`tinyflows`' `RunObserver` has four hooks; we implemented three. `on_step_start` was left as the trait's default no-op, so the socket only ever carried `success`/`error` — emitted *after* a node had already finished. A node was never marked `running`, so `.flow-node-running` and its `flow-node-run-pulse` keyframe (written for exactly this, complete with a `prefers-reduced-motion` override) were unreachable dead code. What shipped was a completion trail, not a live overlay.

Now publishes `FlowRunProgress { status: "running" }` on node activation. **Publish-only, deliberately** — the durable `flow_runs` row stays finish-driven, since a started-but-unfinished node has no output, duration or terminal status to record, and a placeholder row would put a `running` step into run history that the settle-time reconstruction would then reconcile away.

### 2. A global `outline: none !important` blanked every canvas ring (pre-existing)

Even after fixing #1 nothing appeared. `app/src/index.css` carried:

```css
* { outline: none !important; }
```

Its own comment says it exists to hide the browser focus ring — but on a universal selector with `!important` it outranks every component rule and suppressed **all** outlines. All four flow-canvas rings are drawn with `outline`, so this silently killed:

- the live run overlay (`.flow-node-running` / `-success` / `-failed`)
- the validation ring (`.flow-node-error`)
- the copilot diff overlay (`.flow-node-added` / `-removed`)

**Three features beyond this PR have never rendered, in any build, with no test failing** — the classes were being applied to the DOM correctly the whole time, so only a human looking at the canvas could catch it.

Removed `outline` from the bare `*` rule and kept the tap-highlight suppression. Focus rings stay suppressed via the existing `*:focus` rule and the explicit element list, so the comment's stated intent is preserved exactly.

### How it was proven

A CDP DOM probe sampling every 500ms showed the classes cascading correctly all along:

```
t+6.5s   step_one:RUNNING
t+14.5s  step_one:success   step_two:RUNNING
t+23s    step_two:success   step_three:RUNNING
t+31s    step_three:success
```

Forcing a class then reading computed style gave `outline: 0px none` — even for a hardcoded literal colour, which ruled out the CSS variable and pointed at a global override. After the fix all four states compute correctly (`2px solid rgb(47,110,244)` etc.) and stay off with no class applied. Confirmed visually in the app.

### Guards added

- A Rust test failing if `on_step_start` reverts to the default no-op — otherwise the pulse silently becomes unreachable again with every other test green.
- A frontend test failing if a bare `*` rule ever sets `outline` again, plus assertions that each ring still declares an outline and that `running` stays animated. Verified it genuinely fails against the old CSS.

574 flows + 178 tinyflows + 1642 frontend tests pass; `cargo fmt`, `prettier`, `tsc` clean.

PR #5305, opened 2026-07-31
Fall back to CIM when wmic is unavailable

lukiod:
`enumerate_all_processes` shells out to `wmic`, which Microsoft removed in Windows 11 24H2. On those builds the spawn fails and process recovery sees no processes at all, so it can't find or reattach to anything.

Tries `wmic` first so older builds are unaffected, then falls back to `Get-CimInstance Win32_Process` via PowerShell. The fallback emits the same `Key=Value` blocks separated by blank lines that `wmic /format:list` produced, so `parse_wmic_list_output` handles both without changes.

Checked the fallback against real output on Windows 11: 280 records parsed, 279 with a pid (pid 0 is System Idle Process), 280 with a command, 278 with a parent — the two without are pid 0 and pid 4, which legitimately have none.

I couldn't build the whole crate to confirm — `cef-dll-sys` needs CMake and I don't have it installed — so this hasn't been compiled, only parsed and format-checked with rustfmt. Worth a CI run before merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows process recovery reliability by automatically using an alternative discovery method when the primary method is unavailable, fails, or returns no results.
  * Process recovery now handles command failures and output encoding more consistently, helping ensure available processes are detected and recovered correctly.

### DIFF
--- a/app/src-tauri/src/process_recovery.rs
+++ b/app/src-tauri/src/process_recovery.rs
@@ -757,10 +757,24 @@ mod windows_imp {
     /// instance apart from a `core`/`mcp` CLI session or a CEF `--type=` helper.
     /// Includes the current process so its ancestor chain can be walked.
     fn enumerate_all_processes() -> Result<Vec<ProcessInfo>, String> {
-        // wmic is absent from Windows 11 24H2 and later, where the spawn fails
-        // and recovery loses every process. Try it first for older builds, then
-        // fall back to CIM, which returns the same fields.
-        match enumerate_via_wmic() {
+        select_enumeration(enumerate_via_wmic, enumerate_via_cim)
+    }
+
+    /// Choose between the two enumerators.
+    ///
+    /// wmic is absent from Windows 11 24H2 and later, where the spawn fails and
+    /// recovery loses every process. Try it first for older builds, then fall
+    /// back to CIM, which returns the same fields.
+    ///
+    /// Takes both as arguments so the choice can be tested without spawning
+    /// anything: the branch that matters is the one that only happens on a
+    /// machine where wmic is missing.
+    fn select_enumeration<W, C>(wmic: W, cim: C) -> Result<Vec<ProcessInfo>, String>
+    where
+        W: FnOnce() -> Result<Vec<ProcessInfo>, String>,
+        C: FnOnce() -> Result<Vec<ProcessInfo>, String>,
+    {
+        match wmic() {
             Ok(processes) if !processes.is_empty() => {
                 log::debug!(
                     "[startup-recovery] wmic enumerated {} processes",
@@ -770,11 +784,11 @@ mod windows_imp {
             }
             Ok(_) => {
                 log::debug!("[startup-recovery] wmic returned no processes, trying CIM");
-                enumerate_via_cim()
+                cim()
             }
             Err(err) => {
                 log::debug!("[startup-recovery] wmic unavailable ({err}), trying CIM");
-                enumerate_via_cim()
+                cim()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


lukiod:
Both fair. Fixed in f053879.

The encoding one is worse than mangling — PowerShell replaces the character before it reaches Rust, so it's lost, not mis-decoded. Measured here:

```
without pin: b'Caption=\x82??'                      -> "Caption=\ufffd??"
with pin:    b'Caption=\xc3\xa9\xe4\xb8\xad\xe6\x96\x87' -> "Caption=é中文"
```

`é` came back as the cp437 byte and the CJK as literal `?`. Now pins `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8` at the start of the script.

Added `log::debug!` on both enumeration branches and a `log::warn!` when powershell exits non-zero, using the existing `[startup-recovery]` prefix.

Couldn't run `cargo check` here — the CEF build script needs cmake, which isn't installed. Ran the real script instead and put its output through a verbatim copy of `parse_wmic_list_output`: 272 processes, 114 with executable paths, one pid 0 record (System Idle Process).

lukiod:
Tests added in 19a3810.

`select_enumeration` now takes both enumerators as arguments so the choice can be driven without spawning anything, and the CIM script is a named constant so its contents can be asserted:

- wmic results used, CIM not consulted
- CIM on wmic error
- CIM on empty wmic output (a 24H2 shim can exit 0 with nothing)
- CIM failure propagates rather than returning an empty list — an empty list reads as "nothing stale to reap"
- script keeps the UTF-8 pin and every field the parser reads
- parser handles CIM blocks with a non-ASCII path and empty fields

Still can't run `cargo` here — the CEF build script needs cmake. I compiled the extracted logic with the same tests standalone under rustc (12 checks pass) and `rustfmt --check` parses the file clean, but the suite itself hasn't run in a real build.

lukiod:
Good catch on the fixture. `C:\Users\Ordner\OpenHuman.exe` is a German word but entirely ASCII, so `"non-ASCII path was mangled"` could not fail whatever the encoding did.

573eb47 uses `Müller` (representable in cp1252) and `用户` (not), so a regression that drops the OutputEncoding pin fails on one or the other whichever code page the runner uses. Also compares the whole string instead of a substring.

M3gA-Mind:

**The change is well made and the problem is real** — `wmic` is gone in Windows 11 24H2, `enumerate_all_processes` is on the pre-CEF launch preflight (`lib.rs:2767` → `reap_stale_openhuman_processes`), and losing it means recovery silently sees zero processes. You also handled all three earlier review asks properly: `select_enumeration` taking both enumerators as arguments makes the untestable branch testable, the `[Console]::OutputEncoding` pin is the correct fix for PowerShell 5.1's OEM code page (no amount of decoding on the Rust side recovers what PowerShell already replaced), and the logging is there.

One thing you flagged in your own description turns out to be more serious than it reads, so I checked it.

## Nothing in CI compiles this — the green checks are meaningless here

You wrote *"this hasn't been compiled, only parsed and format-checked."* CI does not close that gap:

- All of your code and your new tests are inside `#[cfg(target_os = "windows")] mod windows_imp` (`process_recovery.rs:597`).
- `Rust Tauri Shell Tests` and `Rust Tauri Coverage` run on `ubuntu-22.04` (`test-reusable.yml:326`), so the whole module is `cfg`'d out — it is not compiled and the tests do not run.
- The one Windows job, `Rust Core Tests (Windows — secrets ACL)` (`test-reusable.yml:285-288`), runs `cargo test -p openhuman -- keyring::encrypted_store`. That is the **core** crate, not `app/src-tauri`, and a single filtered suite at that.

So a fully green PR here proves nothing about this diff — not that it compiles, not that `select_enumeration`'s tests pass. Before this merges someone needs to run `cargo check -p openhuman-app` (or the Tauri shell build) on a Windows host, or at minimum `cargo check --target x86_64-pc-windows-msvc`. I've flagged that for the maintainer rather than asking you to find a Windows machine — but if you do have one, posting the output of the three new tests would remove the only real doubt about this PR.

## The open thread: CodeRabbit's enumeration timeout

Legitimate concern, and I'd resolve it by replying rather than by building a timeout runner in this PR:

- The missing timeout is **pre-existing** — `enumerate_via_wmic` used a bare `Command::output()` before your change and still does. You are not introducing the pattern.
- CodeRabbit labelled it "Heavy lift" itself, and a timeout-capable child runner shared by both enumerators is a different change from "fall back when wmic is gone".

That said, your change does make it materially more likely to bite, and it's worth saying so in your reply rather than just deferring:

`reap_stale_openhuman_processes` calls `enumerate_all_processes()` at `process_recovery.rs:623` and again at `:662` (the post-`TERM_GRACE` revalidation), and it runs **synchronously on the Windows launch path before CEF starts**. On 24H2 each of those calls is now a failed `wmic` spawn *plus* a PowerShell cold start — so the worst case moved from "one fast failed spawn" to "two PowerShell startups on the critical path, unbounded if a CIM provider wedges". That is a real regression risk in latency even when nothing hangs.

Suggested resolution: reply with the above, open a follow-up issue for a shared timeout-capable runner covering both enumerators, and link it from the thread. Don't grow this PR.

## Smaller things

- `Get-CimInstance` emits one block per process via `ForEach-Object`, so a `CommandLine` containing a literal newline would split a record — but `wmic /format:list` had exactly the same weakness and `parse_wmic_list_output` is unchanged, so this is parity, not a regression. Not asking you to fix it; worth a comment on `CIM_SCRIPT` so the next reader knows it was considered.
- Your 280-record real-output check (279 with pid, 278 with parent, the two exceptions being pid 0 and pid 4) is exactly the right evidence to have gathered. Please move that into the PR body's verification section rather than leaving it in prose — it is the strongest thing this PR has and a reviewer will look for it there.
- The body has no submission checklist; the gate wants every item ticked, and any N/A item's text must begin with `N/A`.

The `CHANGES_REQUESTED` on this PR is CodeRabbit's from 2026-07-31, before your fixes landed — two of its three threads are already resolved. It just hasn't re-reviewed; don't wait on it.

PR #5342, opened 2026-08-03
fix(dev): make `pnpm dev:app:win` work on default Git for Windows installs

### LINKED ISSUE #5270: dev:app:win always fails on standard Git for Windows install (cmd.exe strips nested quotes)
## Bug

`pnpm dev:app:win` always fails on a standard Windows install of Git for Windows.

`app/package.json:12`

```json
"dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh"
```

Running it:

```
> openhuman-app@0.63.5 dev:app:win F:\...\app
> "C:/Program Files/Git/bin/bash.exe" ../scripts/run-dev-win.sh

'C:/Program' is not recognized as an internal or external command,
operable program or batch file.
 ELIFECYCLE  Command failed with exit code 1.
```

## Cause

pnpm runs lifecycle scripts through `cmd.exe /d /s /c "<script>"` on Windows. The nested
double quotes around the interpreter path are stripped by that outer wrapping, so `cmd.exe`
parses `C:/Program` as the program name and `Files/Git/bin/bash.exe` as an argument.

This reproduces on every machine where Git is installed to the default location
(`C:\Program Files\Git`), i.e. the standard install.

Confirmed it is a quoting problem and not a missing/broken interpreter:

- `Test-Path 'C:/Program Files/Git/bin/bash.exe'` → `True`
- `& 'C:/Program Files/Git/bin/bash.exe' --version` → `GNU bash, version 5.2.37(1)-release (x86_64-pc-msys)`
- The failure happens before `run-dev-win.sh` executes a single line, so it is independent
  of whether `pnpm install` has been run.
- There is no `.npmrc` anywhere in the repo, so pnpm uses the Windows default `script-shell`
  (`cmd.exe`) with no override.

## Context

This is the same bug class that `scripts/run-dev-win.sh` itself documents in detail at lines
658–670, for `cargo-tauri`'s `beforeDevCommand`:

> `cargo-tauri` runs `beforeDevCommand` as `cmd.exe /S /C <string>`. Rust's argv-to-cmd
> argument escaping strips literal double quotes from the string, so if our `<string>` is
> `"E:\Office Files\…\vite.CMD"`, cmd ends up parsing `E:\Office` as the program name and the
> rest as arguments — "'E:\Office' is not recognized". 8.3 short-name fallback also fails when
> 8dot3name is disabled on the drive.

The script works around it by generating a `.bat` wrapper under a space-free `%TEMP%` path.
The same treatment was never applied to the outer entry point in `package.json`, so the outer
layer fails for exactly the same reason before the inner workaround is ever reached.

## Suggested fixes

Any of:

1. Emit/ship a `.bat` wrapper for the entry point, mirroring what `run-dev-win.sh` already does
   internally for `beforeDevCommand`.
2. Invoke bash by PATH-resolvable name in `package.json` (e.g. `bash ../scripts/run-dev-win.sh`),
   matching how the sibling `tauri:ensure` / `build:web:e2e` scripts already call `bash`.
3. Set `script-shell` in `.npmrc` so pnpm runs scripts through a shell that handles the quoting.

## Environment

- Windows 11 Home 26200
- pnpm 10.10.0
- Node 24.14.1
- Git for Windows 2.x, default location `C:\Program Files\Git` (bash 5.2.37)
- `openhuman-app@0.63.5`

Mustaqeem66:
## Summary

- `pnpm dev:app:win` fails on every default Git for Windows install with `'C:/Program' is not recognized as an internal or external command`.
- Route the script through a new `scripts/run-dev-win.cmd` wrapper so the `bash.exe` path is quoted in a context where quoting actually survives.
- `app/package.json` now points at a relative, space-free path (`..\scripts\run-dev-win.cmd`) that needs no quoting of its own.
- The wrapper is written as straight-line batch with **no labels and no `goto`/`call`**, so its behaviour does not depend on whether it is checked out with LF or CRLF.
- `.gitattributes` now checks out `*.cmd`/`*.bat` with CRLF, matching the exception `*.ps1` already had.
- Added `app/test/dev-app-win-launcher.test.ts`, which models `cmd.exe /d /s /c` argument parsing and locks in both the quoting regression and the no-label property.
- `scripts/run-dev-win.sh` is unchanged — this only fixes how it is invoked.

## Problem

`app/package.json` declared:

```json
"dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh"
```

pnpm runs package.json scripts through `cmd.exe /d /s /c <string>`. The **`/S` flag strips the first and last quote characters** of `<string>` before parsing it. So cmd receives:

```
C:/Program Files/Git/bin/bash.exe ../scripts/run-dev-win.sh
```

and takes `C:/Program` as the program name, producing exactly the error reported in #5270. The quotes cannot protect the path at this layer — they are the first and last characters, so they are precisely what `/S` removes.

Because `C:\Program Files\Git` is the **default** install location, this breaks the documented Windows dev path for essentially every contributor who did not install Git somewhere without a space.

## Solution

Quoting behaves normally *inside* a `.cmd` file, so the fix is to make the package.json body something cmd cannot mis-split, and do the real quoting one level down.

`app/package.json` (single line changed):

```diff
-"dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh",
+"dev:app:win": "..\\scripts\\run-dev-win.cmd",
```

The new value has no spaces and no quotes, so `/S` stripping is a no-op on it. Backslashes are used rather than forward slashes because cmd treats a leading `/` as a switch prefix. pnpm runs lifecycle scripts with the package directory as cwd, so the relative path resolves from `app/`.

`scripts/run-dev-win.cmd` then locates `bash.exe` and invokes it with the path properly quoted:

```bat
"%BASH_EXE%" "%SCRIPT_DIR%run-dev-win.sh" %*
```

Lookup order:

1. `%OPENHUMAN_BASH_EXE%` — escape hatch for portable/exotic Git installs.
2. `%ProgramFiles%\Git\bin`, `%ProgramFiles(x86)%\Git\bin`, `%LOCALAPPDATA%\Programs\Git\bin` — the machine-scope and user-scope Git for Windows installers.
3. `%SCOOP%`, `%USERPROFILE%\scoop`, `%ProgramData%\scoop` → `apps\git\current\bin\bash.exe` — scoop only exposes shims on `PATH`, so it must be probed directly (see below).
4. The Git root derived from `git.exe` on `PATH` — covers winget and chocolatey, where `git.exe` lives in `\cmd` and bash in `\bin`.

If none match it exits `1` with actionable guidance instead of failing obscurely.

### No labels, no `goto`/`call` — deliberate

Each candidate is applied by a single-line `if not defined BASH_EXE if exist "…" set "BASH_EXE=…"`, and the failure path is a run of single-line `if not defined BASH_EXE echo …>&2` lines terminated by `if not defined BASH_EXE exit /b 1`. There is no `call :use_if_exists` helper and no `goto :no_bash` jump.

That is not stylistic. `cmd.exe` locates a label by **byte offset** and re-reads the script in 512-byte chunks, which makes label-based control flow the one batch construct whose behaviour depends on the file's line endings — the basis of the well-known LF-only batch corruption bug. Writing the launcher without labels removes the dependency at the source, instead of relying on every checkout on every machine having applied the right `.gitattributes` rule. A test pins the property so it cannot silently regress.

### `.gitattributes`

The repo root normalises everything with `* text=auto eol=lf` and carves out exactly one Windows exception, `*.ps1 text eol=crlf`. Batch files had no rule at all, so `*.cmd`/`*.bat text eol=crlf` was added alongside it. With the label-free launcher this is defence in depth and consistency for editors and future batch files, rather than the thing correctness rests on.

### Design notes / rejected alternatives

The issue suggested three options. This PR takes **option 1** deliberately:

- **Option 2 — bare `bash ../scripts/run-dev-win.sh`** was rejected as actively harmful. `Git\bin` is *not* on `PATH` by default (only `Git\cmd`, which has no `bash.exe`), and on machines with WSL enabled `bash` resolves to `C:\Windows\System32\bash.exe` — the WSL launcher. That would run the script inside a Linux distro with none of the Windows toolchain it configures (MSVC, CEF, cargo-tauri), producing a far more confusing failure than the current one. The wrapper therefore **never** falls back to a bare `bash` lookup, and there is a test asserting it does not.
- **Option 3 — `.npmrc` `script-shell`** was rejected because it changes shell semantics repo-wide for every platform to fix one Windows script.
- **Option 1 is already the idiom here.** `scripts/run-dev-win.sh` generates `.bat` shims internally (`vcvars_launcher`, `run-vite.bat`) for cargo-tauri's `beforeDevCommand` to dodge this exact quoting problem. This change applies the same trick one level up, at the pnpm boundary.

## Impact

- **Platform:** Windows dev workflow only. `dev:app`, `dev:web`, `dev:wry` and all CI/build/release scripts are untouched; nothing outside Windows executes a `.cmd`.
- **Runtime/product:** none. This is a developer entry point, not shipped application code.
- **Compatibility:** a hardcoded absolute path is replaced by a probe, so this *widens* the set of working setups (machine-scope, user-scope, winget, scoop, chocolatey, portable Git) rather than narrowing it. `OPENHUMAN_BASH_EXE` covers anything exotic.
- **Security:** no new network dependency; no new executable is downloaded. Candidate paths are all derived from OS-provided environment variables or `where git.exe`.

## Related

- Closes: #5270
- Coverage matrix feature IDs: none — see checklist note below.

## Review feedback addressed

- **CodeRabbit — CRLF line endings, round 1** (b4a710e): fixed at the attributes layer rather than by committing CRLF bytes, since `text=auto` normalises the blob to LF on commit regardless, so a checkout rule is the only durable form of that fix.
- **Greptile — scoop coverage** (f141144): correct, and this description previously overclaimed. `where git.exe` under scoop returns `<scoop>\shims\git.exe`, so the derived `<shim>\..\bin\bash.exe` resolved to `<scoop>\bin\bash.exe`, which never exists. Now probing `apps\git\current\bin\bash.exe` — the real binary behind the shim, via the `current` junction scoop maintains across upgrades — across the user, global and `%SCOOP%`-relocated roots. The failure-path diagnostic and the comment above the `git.exe` probe were corrected to match.
- **CodeRabbit — CRLF line endings, round 2** (9d81ab4, b353a74): the bot re-flagged the blob (Blinter `E018`) after the `.gitattributes` fix, and it has a fair point that the earlier fix left the *repository object* LF-only — which is what anyone reading the file through the API, a raw link, or a source archive sees. Rather than argue the checkout rule again, **the dependency itself is now gone**: the launcher was rewritten with no labels and no `goto`/`call`, which is the only batch construct whose parsing is line-ending sensitive. The file is now correct under LF *and* CRLF, the `.gitattributes` rule stays for consistency, and `app/test/dev-app-win-launcher.test.ts` gained a fifth case asserting no label-based control flow so this cannot regress. Behaviour is otherwise unchanged: identical probe order, identical quoting, identical exit-code propagation.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `app/test/dev-app-win-launcher.test.ts` has 5 cases. The **failure-path** case reconstructs the old script body verbatim and asserts `cmd /S` parsing yields `C:/Program`, so the regression is pinned rather than merely described. The others assert the current body survives `/S` as one token, that it points at a file that exists, that the wrapper quotes `%BASH_EXE%` and contains no bare-`bash` lookup, and that the wrapper uses no label-based control flow.
- [x] **Diff coverage ≥ 80%** — `N/A`, and verifiably so rather than by assertion: `app/test/vitest.config.ts` sets `coverage.include: ["src/**/*.{ts,tsx}"]`. All four changed files (`app/package.json`, `app/test/dev-app-win-launcher.test.ts`, `scripts/run-dev-win.cmd`, `.gitattributes`) fall outside `app/src/**`, so no changed line appears in the lcov report and `diff-cover --fail-under=80` has no lines to judge. No Rust changed, so `cargo-llvm-cov` is unaffected. Please flag if you would rather I approach this differently.
- [x] Coverage matrix updated — `N/A: developer tooling change`. `docs/TEST-COVERAGE-MATRIX.md` tracks product feature rows with IDs validated against a catalog; a build-script entry point has no feature ID, and inventing one would add noise. Calling this out explicitly rather than silently skipping — happy to add a row if you disagree.
- [x] All affected feature IDs listed under `## Related` — `N/A`, none (see above).
- [x] No new external network dependencies introduced — none added; no test performs I/O beyond reading two files from the repo working tree.
- [x] Manual smoke checklist updated — `N/A: does not touch release-cut surfaces`. This affects the dev entry point only; release builds go through `macos:build:*` / `build:app`.
- [x] Linked issue closed via `Closes #NNN` in `## Related` — `Closes #5270`.

---

Disclosing this fully, since the template asks for it: the investigation and patch were drafted with AI assistance, reviewed line-by-line by me before pushing. It is not a Codex/Linear-tracked PR, so those fields are `N/A`.

### Linear Issue

- Key: `N/A` — external contribution, tracked by GitHub issue #5270.
- URL: https://github.com/tinyhumansai/openhuman/issues/5270

### Commit & Branch

- Branch: `Mustaqeem66:fix/5270-dev-app-win-cmd-quoting`
- Commit SHA: `b353a745ad541696126481825deb6bb5a040cd7f`

### Validation Run

> Three of these could not be run in my environment. Rather than tick boxes I did not earn — or leave them unchecked and trip `check-pr-checklist.mjs` — they are listed as plain bullets marked **NOT RUN**, with full detail in **Validation Blocked** below. Happy to reformat if maintainers prefer a different convention here.

- **NOT RUN** — `pnpm --filter openhuman-app format:check`
- **NOT RUN** — `pnpm typecheck`
- **NOT RUN** — Focused tests: `pnpm --filter openhuman-app test -- dev-app-win-launcher`
- [x] Rust fmt/check (if changed): `N/A` — no Rust changed.
- [x] Tauri fmt/check (if changed): `N/A` — no Tauri config or Rust changed.

### Validation Blocked

I would rather state this plainly than tick boxes I did not earn:

- `command:` `pnpm --filter openhuman-app format:check`, `pnpm typecheck`, `pnpm test`, `pnpm test:coverage`
- `error:` `pnpm: command not found` — the environment I prepared this in has no Node package manager, no Rust toolchain and no network access, so dependencies cannot be installed.
- `impact:` The test file has **not** been executed locally under vitest. To de-risk that, the pure-logic helpers were extracted and run standalone under plain `node`: the legacy body yields `C:/Program`, the new body survives `/S` unchanged, and the no-label assertions were run against both an LF and a CRLF copy of the launcher (both pass, which is the point of the rewrite). The remaining assertions are `existsSync` and substring checks against files added in this PR. CI is the real verification for typecheck/format/lint — the first run flagged a `format:check` failure (hand-wrapped call expressions that fit inside the repo's `printWidth` of 100), fixed in bb9e76a with no behavioural change to the tests.
- `impact (Windows):` I have no Windows machine attached to this environment, so `pnpm dev:app:win` has not been run end-to-end against a real Git install. The `.cmd` was desk-checked for batch parsing hazards — one was found and fixed in `9df25a7` (an unescaped `)` in `%ProgramFiles(x86)%` inside an `if (...)` block would have closed the block early), and the later rewrite removed every multi-line parenthesised block and every label, so that class of hazard is now structurally absent. **A maintainer sanity-check on a real Windows box would be genuinely valuable before merge**, and I will act on any failure immediately.

### Behavior Changes

- Intended behavior change: `pnpm dev:app:win` resolves `bash.exe` at runtime instead of assuming one hardcoded absolute path, and is invoked so that `cmd.exe /S` cannot split it.
- User-visible effect: for developers, `pnpm dev:app:win` starts working on default Git for Windows installs instead of erroring immediately. No end-user or product-facing change.

### Parity Contract

- Legacy behavior preserved: yes. The command ultimately executed is still `<bash.exe> <repo>/scripts/run-dev-win.sh`, with `run-dev-win.sh` byte-identical. On the machines where the old form happened to work (Git at `C:\Program Files\Git`), the wrapper's first filesystem probe resolves to that same `bash.exe`, so behavior is identical. Arguments are forwarded via `%*`.
- Guard/fallback/dispatch parity checks: the exit code is propagated with `exit /b %ERRORLEVEL%` so failures still surface to pnpm. The fallback chain only ever *adds* candidates that the old hardcoded path did not cover; it removes none, and the label-free rewrite preserved the probe order exactly. The one intentional non-fallback is a bare `bash` PATH lookup, excluded because of the WSL hazard described above and asserted absent by test.

- Duplicate PR(s): none found. I searched open and closed PRs for #5270 and for Windows dev-script changes before starting; the issue is unassigned with no linked PR.
- Canonical PR: this one.
- Resolution: `N/A`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Windows app development command so it launches reliably from paths containing spaces.
  * Added a Windows launcher that correctly locates Git Bash and forwards command arguments.
  * Added clear guidance when Git Bash cannot be found.

* **Chores**
  * Standardized Windows command-file line endings for improved compatibility.

* **Tests**
  * Added coverage for Windows command parsing, launcher discovery, argument forwarding, and error handling.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,5 @@ Dockerfile text eol=lf
 .dockerignore text eol=lf
 docker-compose*.yml text eol=lf
 *.ps1 text eol=crlf
+*.cmd text eol=crlf
+*.bat text eol=crlf
--- a/app/package.json
+++ b/app/package.json
@@ -9,7 +9,7 @@
     "dev": "vite",
     "dev:web": "vite",
     "dev:app": "pnpm tauri:ensure && export CEF_PATH=\"$HOME/Library/Caches/tauri-cef\" && bash ../scripts/setup-chromium-safe-storage.sh && source ../scripts/load-dotenv.sh && APPLE_SIGNING_IDENTITY='OpenHuman Dev Signer' cargo tauri dev --config '{\"build\":{\"devUrl\":\"http://localhost:'\"${OPENHUMAN_DEV_PORT:-1420}\"'\"}}' ",
-    "dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh",
+    "dev:app:win": "..\\scripts\\run-dev-win.cmd",
     "dev:cef": "pnpm dev:app",
     "dev:wry": "pnpm tauri:ensure && export CEF_PATH=\"$HOME/Library/Caches/tauri-cef\" && source ../scripts/load-dotenv.sh && cargo tauri dev --no-default-features --features wry",
     "core:stage": "echo '[core:stage] no-op — core is linked in-process; sidecar removed (PR #1061)'",
--- a/app/test/dev-app-win-launcher.test.ts
+++ b/app/test/dev-app-win-launcher.test.ts
@@ -43,6 +43,15 @@ function programTokenUnderCmdS(script: string): string {
   return remaining.trimStart().split(/\s/)[0];
 }
 
+/** Executable lines of the launcher: comments and blank lines removed. */
+function launcherCodeLines(launcher: string): string[] {
+  return launcher
+    .split('\n')
+    .map((line) => line.trim())
+    .filter((line) => line.length > 0)
+    .filter((line) => !line.toLowerCase().startsWith('rem'));
+}
+
 describe('pnpm dev:app:win entry point', () => {
   it('reproduces the historical failure: quotes do not survive cmd.exe /s', () => {
     // cmd sees `C:/Program` as the program name, producing the reported
@@ -77,5 +86,17 @@ describe('pnpm dev:app:win entry point', () => {
     // A bare `bash` lookup would hit C:\Windows\System32\bash.exe (the WSL
     // launcher) on machines with WSL enabled.
     expect(launcher).not.toMatch(/where\s+bash\.exe/i);
+  });
+
+  it('uses no label-based control flow, so line endings cannot change behaviour', () => {
+    const code = launcherCodeLines(readFileSync(LAUNCHER_PATH, 'utf8'));
+
+    // cmd.exe finds a label by byte offset and re-reads the script in 512-byte
+    // chunks, so `goto`/`call :label` is the one construct whose behaviour
+    // depends on whether the file is LF or CRLF. Keeping the launcher free of
+    // labels makes it correct either way instead of relying on a checkout rule.
+    expect(code.filter((line) => line.startsWith(':'))).toEqual([]);
+    expect(code.filter((line) => /(^|\s)(goto|call)\s/i.test(line))).toEqual([]);
+    expect(code.length).toBeGreaterThan(0);
   });
 });
--- a/app/test/dev-app-win-launcher.test.ts
+++ b/app/test/dev-app-win-launcher.test.ts
@@ -47,9 +47,9 @@ function programTokenUnderCmdS(script: string): string {
 function launcherCodeLines(launcher: string): string[] {
   return launcher
     .split('\n')
-    .map((line) => line.trim())
-    .filter((line) => line.length > 0)
-    .filter((line) => !line.toLowerCase().startsWith('rem'));
+    .map(line => line.trim())
+    .filter(line => line.length > 0)
+    .filter(line => !line.toLowerCase().startsWith('rem'));
 }
 
 describe('pnpm dev:app:win entry point', () => {
@@ -95,8 +95,8 @@ describe('pnpm dev:app:win entry point', () => {
     // chunks, so `goto`/`call :label` is the one construct whose behaviour
     // depends on whether the file is LF or CRLF. Keeping the launcher free of
     // labels makes it correct either way instead of relying on a checkout rule.
-    expect(code.filter((line) => line.startsWith(':'))).toEqual([]);
-    expect(code.filter((line) => /(^|\s)(goto|call)\s/i.test(line))).toEqual([]);
+    expect(code.filter(line => line.startsWith(':'))).toEqual([]);
+    expect(code.filter(line => /(^|\s)(goto|call)\s/i.test(line))).toEqual([]);
     expect(code.length).toBeGreaterThan(0);
   });
 });
--- a/app/test/dev-app-win-launcher.test.ts
+++ b/app/test/dev-app-win-launcher.test.ts
@@ -22,6 +22,18 @@ const DEV_APP_WIN = packageJson.scripts['dev:app:win'];
 const LEGACY_BROKEN_SCRIPT = '"C:/Program Files/Git/bin/bash.exe" ../scripts/run-dev-win.sh';
 
 /**
+ * A batch label, optionally behind the `@` echo-suppression prefix.
+ */
+const BATCH_LABEL = /^@?:/;
+
+/**
+ * A `goto`/`call` jump in any form cmd accepts: optionally prefixed with `@`,
+ * introduced by start-of-line or a command separator, and followed by either
+ * whitespace (`goto :eof`) or a colon (`goto:eof`).
+ */
+const BATCH_JUMP = /(^|[\s&|()])@?(goto|call)[\s:]/i;
+
+/**
  * Models how `cmd.exe /d /s /c <string>` picks the program to execute, which is
  * how pnpm invokes package.json scripts on Windows.
  *
@@ -88,6 +100,33 @@ describe('pnpm dev:app:win entry point', () => {
     expect(launcher).not.toMatch(/where\s+bash\.exe/i);
   });
 
+  it('recognises every batch jump form, so the guard below cannot be bypassed', () => {
+    // Positive controls: each of these must count as a jump.
+    for (const jump of [
+      'goto :label',
+      'call :label',
+      '@goto :label',
+      '@call :label',
+      'goto:eof',
+      '@goto:eof',
+      '& goto :label',
+      '( @goto :label)',
+      'if not defined BASH_EXE goto :no_bash',
+    ]) {
+      expect(BATCH_JUMP.test(jump), jump).toBe(true);
+    }
+
+    // Negative controls: ordinary text must not trip the guard.
+    for (const plain of [
+      'echo go to label',
+      'echo recall the value',
+      'exit /b %ERRORLEVEL%',
+      '"%BASH_EXE%" "%SCRIPT_DIR%run-dev-win.sh" %*',
+    ]) {
+      expect(BATCH_JUMP.test(plain), plain).toBe(false);
+    }
+  });
+
   it('uses no label-based control flow, so line endings cannot change behaviour', () => {
     const code = launcherCodeLines(readFileSync(LAUNCHER_PATH, 'utf8'));
 
@@ -95,8 +134,8 @@ describe('pnpm dev:app:win entry point', () => {
     // chunks, so `goto`/`call :label` is the one construct whose behaviour
     // depends on whether the file is LF or CRLF. Keeping the launcher free of
     // labels makes it correct either way instead of relying on a checkout rule.
-    expect(code.filter(line => line.startsWith(':'))).toEqual([]);
-    expect(code.filter(line => /(^|\s)(goto|call)\s/i.test(line))).toEqual([]);
+    expect(code.filter(line => BATCH_LABEL.test(line))).toEqual([]);
+    expect(code.filter(line => BATCH_JUMP.test(line))).toEqual([]);
     expect(code.length).toBeGreaterThan(0);
   });
 });
--- a/app/test/dev-app-win-launcher.test.ts
+++ b/app/test/dev-app-win-launcher.test.ts
@@ -6,55 +6,55 @@ import { describe, expect, it } from 'vitest';
 const HERE = path.dirname(fileURLToPath(import.meta.url));
 const APP_DIR = path.resolve(HERE, '..');
 const REPO_ROOT = path.resolve(APP_DIR, '..');
+const PACKAGE_JSON_PATH = path.join(APP_DIR, 'package.json');
+const LAUNCHER_PATH = path.join(REPO_ROOT, 'scripts', 'run-dev-win.cmd');
 
-const packageJson = JSON.parse(
-  readFileSync(path.join(APP_DIR, 'package.json'), 'utf8')
-) as { scripts: Record<string, string> };
+type PackageJson = { scripts: Record<string, string> };
 
+const packageJson = JSON.parse(readFileSync(PACKAGE_JSON_PATH, 'utf8')) as PackageJson;
 const DEV_APP_WIN = packageJson.scripts['dev:app:win'];
 
 /**
  * The exact `dev:app:win` body that shipped before this regression was fixed.
- * Kept verbatim so the test below proves the failure mode rather than just
+ * Kept verbatim so the first test proves the failure mode rather than merely
  * asserting the current string.
  */
 const LEGACY_BROKEN_SCRIPT = '"C:/Program Files/Git/bin/bash.exe" ../scripts/run-dev-win.sh';
 
 /**
- * Models how `cmd.exe /d /s /c <string>` picks the program to execute, which
- * is how pnpm invokes package.json scripts on Windows.
+ * Models how `cmd.exe /d /s /c <string>` picks the program to execute, which is
+ * how pnpm invokes package.json scripts on Windows.
  *
- * With /S, cmd strips the first and last quote characters of <string> and
+ * With /S, cmd strips the first and last quote characters of <string> and then
  * parses what remains, taking the first whitespace-delimited token as the
  * program name. That is why an interpreter path containing spaces cannot
  * protect itself with quotes at this layer.
  */
 function programTokenUnderCmdS(script: string): string {
+  const firstQuote = script.indexOf('"');
+  const lastQuote = script.lastIndexOf('"');
   let remaining = script;
-  const firstQuote = remaining.indexOf('"');
-  const lastQuote = remaining.lastIndexOf('"');
   if (firstQuote !== -1 && lastQuote > firstQuote) {
-    remaining =
-      remaining.slice(0, firstQuote) +
-      remaining.slice(firstQuote + 1, lastQuote) +
-      remaining.slice(lastQuote + 1);
+    const head = script.slice(0, firstQuote);
+    const middle = script.slice(firstQuote + 1, lastQuote);
+    const tail = script.slice(lastQuote + 1);
+    remaining = head + middle + tail;
   }
-  return remaining.trimStart().split(/\s/)[0] ?? '';
+  return remaining.trimStart().split(/\s/)[0];
 }
 
 describe('pnpm dev:app:win entry point', () => {
   it('reproduces the historical failure: quotes do not survive cmd.exe /s', () => {
-    // Documents the bug this script layout fixes. cmd sees `C:/Program` as the
-    // program name, producing "'C:/Program' is not recognized...".
+    // cmd sees `C:/Program` as the program name, producing the reported
+    // "'C:/Program' is not recognized as an internal or external command".
     expect(programTokenUnderCmdS(LEGACY_BROKEN_SCRIPT)).toBe('C:/Program');
   });
 
   it('resolves to a single token that cmd.exe can execute verbatim', () => {
     const program = programTokenUnderCmdS(DEV_APP_WIN);
 
-    // The whole script must survive cmd's quote stripping intact: if the
-    // program token is shorter than the trimmed script body, cmd would have
-    // split the path and failed exactly like the legacy form above.
+    // If the program token were shorter than the trimmed script body, cmd would
+    // have split the path and failed exactly like the legacy form above.
     expect(program).toBe(DEV_APP_WIN.trim());
     expect(program).not.toContain(' ');
     expect(program).not.toContain('"');
@@ -64,17 +64,11 @@ describe('pnpm dev:app:win entry point', () => {
     const program = programTokenUnderCmdS(DEV_APP_WIN);
     const resolved = path.resolve(APP_DIR, program.replace(/\\/g, '/'));
 
-    expect(
-      existsSync(resolved),
-      `dev:app:win points at ${program}, which does not exist at ${resolved}`
-    ).toBe(true);
+    expect(existsSync(resolved), `dev:app:win points at ${resolved}`).toBe(true);
   });
 
   it('delegates to run-dev-win.sh without trusting a bare bash on PATH', () => {
-    const launcher = readFileSync(
-      path.join(REPO_ROOT, 'scripts', 'run-dev-win.cmd'),
-      'utf8'
-    );
+    const launcher = readFileSync(LAUNCHER_PATH, 'utf8');
 
     expect(launcher).toContain('run-dev-win.sh');
     // Every bash.exe candidate must be quoted, otherwise the wrapper would
--- a/app/test/dev-app-win-launcher.test.ts
+++ b/app/test/dev-app-win-launcher.test.ts
@@ -1,0 +1,87 @@
+import { existsSync, readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+const HERE = path.dirname(fileURLToPath(import.meta.url));
+const APP_DIR = path.resolve(HERE, '..');
+const REPO_ROOT = path.resolve(APP_DIR, '..');
+
+const packageJson = JSON.parse(
+  readFileSync(path.join(APP_DIR, 'package.json'), 'utf8')
+) as { scripts: Record<string, string> };
+
+const DEV_APP_WIN = packageJson.scripts['dev:app:win'];
+
+/**
+ * The exact `dev:app:win` body that shipped before this regression was fixed.
+ * Kept verbatim so the test below proves the failure mode rather than just
+ * asserting the current string.
+ */
+const LEGACY_BROKEN_SCRIPT = '"C:/Program Files/Git/bin/bash.exe" ../scripts/run-dev-win.sh';
+
+/**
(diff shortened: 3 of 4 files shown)


Mustaqeem66:
Both review findings were valid and are now addressed. (Replying here rather than in-thread — threaded replies are blocked for me by the org's OAuth app access restrictions.)

---

### 1. CRLF line endings on `run-dev-win.cmd` — @coderabbitai

Fixed at the attributes layer rather than by committing CRLF bytes (b4a710e).

The root `.gitattributes` starts with `* text=auto eol=lf` and carves out exactly one Windows exception, `*.ps1 text eol=crlf`. There's no rule for `*.cmd`/`*.bat`. Under `text=auto` git normalises the blob to LF on commit regardless of what I push, so storing CRLF in the object would just get normalised away — and even if it survived, a fresh clone would still check the file out as LF for every contributor.

So the durable fix is the checkout rule:

```
*.ps1 text eol=crlf
*.cmd text eol=crlf
*.bat text eol=crlf
```

Batch scripts now get the same treatment PowerShell already had: LF in the object store, CRLF in the working tree, which is what `cmd.exe` actually reads. This also covers any future `.cmd`/`.bat` rather than just this one file.

The 512-byte read-boundary bug being flagged is a real risk for this file specifically, since it uses both `goto :no_bash` and `call :use_if_exists`.

### 2. Scoop coverage claim — @greptile-apps

Correct, and my PR description was overclaiming. Fixed in f141144.

Confirming the trace: scoop only ever exposes shims on PATH, so `where git.exe` returns `<scoop>\shims\git.exe` and `%%~dpG..\bin\bash.exe` resolves to `<scoop>\bin\bash.exe`, which doesn't exist. None of the original probes could ever match a scoop install.

I probed `apps\git\current\bin\bash.exe` rather than `shims\bash.exe` — that's the real binary behind the shim, reached through the `current` junction scoop maintains across upgrades, so it's present whenever the shim is and doesn't depend on scoop having shimmed `bash.exe` specifically:

```
%SCOOP%\apps\git\current\bin\bash.exe
%USERPROFILE%\scoop\apps\git\current\bin\bash.exe
%ProgramData%\scoop\apps\git\current\bin\bash.exe
```

`%SCOOP%` first for relocated installs; when it's unset the unexpanded string simply fails `if exist`, so it's a harmless no-op. `%ProgramData%` covers `scoop install -g`.

I also corrected the `:no_bash` diagnostic text and the comment above the `git.exe` probe, which now claims winget and chocolatey only.

---

Also in this push: `app/test/dev-app-win-launcher.test.ts` was reformatted to satisfy `format:check` (the repo's `printWidth` is 100, so several call expressions I'd wrapped by hand needed collapsing). No behavioural change to the tests.

Mustaqeem66:
Note on the one review thread still showing as open — the CodeRabbit CRLF finding on `scripts/run-dev-win.cmd`. It was addressed in b4a710e, but at the `.gitattributes` layer rather than by committing CRLF bytes, so the byte-level linter can't see the fix. Worth explaining the choice:

The repo root sets `* text=auto eol=lf`. Under `text=auto`, git normalises the blob to LF **on commit** regardless of what I push — so committing CRLF bytes would not survive. The stored blob would still be LF, every fresh clone would still check the file out LF-only, and the finding would reappear for the next person who cloned.

The durable fix is the checkout rule, added alongside the `*.ps1` exception that already existed for exactly this reason:

```
*.ps1 text eol=crlf
*.cmd text eol=crlf
*.bat text eol=crlf
```

`scripts/run-dev-win.cmd` now lands in the working tree with CRLF on every platform, which is what `cmd.exe`'s parser needs given the file uses both `goto :no_bash` and `call :use_if_exists` — and it covers future batch files rather than just this one.

The Greptile scoop finding was also valid and is fixed in f141144 (probing `apps\git\current\bin\bash.exe` rather than deriving from the shim path, which resolved to a directory that never exists).

The one thing I still can't verify myself is an end-to-end `pnpm dev:app:win` run on a real Windows box with Git installed at the default location — noted in the Validation Blocked section. If a maintainer or anyone on Windows can sanity-check that, I'll act on any failure straight away.

Mustaqeem66:
Round-2 review feedback addressed in `9d81ab4` + `b353a74`. Posting as a top-level comment because threaded replies from my account are rejected with `403 … tinyhumansai organization has enabled OAuth App access restrictions`.

### @coderabbitai — `scripts/run-dev-win.cmd` LF line endings (Blinter `E018`, Major)

Valid finding, and I'll grant that my first pass only half-answered it. `b4a710e` added `*.cmd`/`*.bat text eol=crlf` to `.gitattributes`, which makes every *checkout* CRLF, but because root `.gitattributes` carries `* text=auto eol=lf`, git normalises the blob to LF on commit regardless — so the stored object stays LF, and that's exactly what your linter reads (and what anyone reading via the API, a raw link, or a source archive gets). Committing literal CRLF bytes would additionally require dropping the file out of git's text normalisation entirely (`-text`), which diverges from the `*.ps1 text eol=crlf` convention this repo already established.

So rather than keep arguing about bytes, I removed the reason the bytes matter.

**The launcher no longer contains any label-based control flow.** `call :use_if_exists` and `goto :no_bash` are gone:

- each probe is now a single-line `if not defined BASH_EXE if exist "…" set "BASH_EXE=…"`;
- the `git.exe` fallback is a one-line `if not defined BASH_EXE for /f … do if not defined BASH_EXE if exist … set …`;
- the failure path is a run of single-line `if not defined BASH_EXE echo …>&2` statements ending in `if not defined BASH_EXE exit /b 1`;
- there are no multi-line parenthesised blocks left either, which also retires the `%ProgramFiles(x86)%`-inside-`if (…)` hazard I'd fixed separately in `9df25a7`.

`cmd.exe` seeks labels by byte offset and re-reads the script in 512-byte chunks — that is *the* mechanism behind the LF-only batch corruption bug your rule cites. With no labels and no jumps, the file executes identically under LF and CRLF, so correctness stops depending on whether a given clone applied the checkout rule. The `.gitattributes` rule stays: it's still right for editors, for future batch files, and as defence in depth — it's just no longer load-bearing.

To stop this regressing quietly, `app/test/dev-app-win-launcher.test.ts` gained a fifth case that strips `rem` lines and asserts no remaining line starts with `:` and none contains `goto`/`call`. I ran those assertions standalone under `node` against both an LF and a CRLF copy of the file; both pass.

Behaviour is otherwise unchanged: same probe order, same `"%BASH_EXE%"` quoting, same `%*` forwarding, same `exit /b %ERRORLEVEL%` propagation, still no bare-`bash` PATH lookup (WSL hazard).

If you'd still prefer literal CRLF in the object, say so and I'll switch `*.cmd`/`*.bat` to `-text` and recommit the bytes — I just didn't want to break the repo's normalisation convention unasked.

### @greptile-apps — scoop coverage

Already handled in `f141144` (your comment is marked outdated now) and carried through the rewrite unchanged: the four scoop-relevant probes are `%SCOOP%`, `%USERPROFILE%\scoop` and `%ProgramData%\scoop`, each pointing at `apps\git\current\bin\bash.exe` — the real binary behind the shim, via the `current` junction scoop maintains across upgrades — rather than the `<scoop>\shims\..\bin\bash.exe` path that never exists. The PR description and the failure-path diagnostic were corrected to match; thanks for catching the overclaim.

### Still outstanding on my side

`pnpm dev:app:win` has not been executed against a real Git for Windows install — I have no Windows box attached to this environment. Everything above is desk-checked plus CI. A one-line confirmation from anyone on Windows would be the last missing piece before merge.

Mustaqeem66:
Round-3 review response — one finding accepted, one respectfully declined.

(Posting top-level: threaded replies from my account return `403 … tinyhumansai organization has enabled OAuth App access restrictions`.)

### ✅ Accepted — `@goto` / `@call :label` were not detected (`7b0da28`)

Correct, and thank you — the guard had a real hole. I took the suggestion and widened it slightly further, because the proposed pattern still missed the colon-joined form:

```diff
-    expect(code.filter(line => /(^|\s)(goto|call)\s/i.test(line))).toEqual([]);
+const BATCH_LABEL = /^@?:/;
+const BATCH_JUMP = /(^|[\s&|()])@?(goto|call)[\s:]/i;
```

- `[\s&|()]` — the separators cmd accepts before a command, so `& goto :x` and `( @goto :x)` are caught.
- `@?` — the echo-suppression prefix, so `@goto` / `@call` are caught.
- `[\s:]` instead of `\s` — **`goto:eof` is valid cmd and is arguably the most common jump form in the wild**; `/…(goto|call)\s/` misses it, and so did the suggested `/(^|[\s&|()])@?(goto|call)\s/`.
- `BATCH_LABEL = /^@?:/` — a label line behind `@` is now caught too.

A regex is only as good as its controls, so the patterns are now themselves under test (new case *“recognises every batch jump form, so the guard below cannot be bypassed”*): 9 positive controls (`goto :label`, `call :label`, `@goto :label`, `@call :label`, `goto:eof`, `@goto:eof`, `& goto :label`, `( @goto :label)`, `if not defined BASH_EXE goto :no_bash`) and 4 negative controls (`echo go to label`, `echo recall the value`, `exit /b %ERRORLEVEL%`, the launcher’s own invocation line). Against the current launcher: **21 executable lines, zero label matches, zero jump matches, identical under LF and CRLF.**

### 🙏 Declined — moving the spec to `app/src/**`

I don't think this one holds against the repo as it stands, and moving it would make things worse rather than better:

1. **`app/test/*.test.ts` is a first-class, maintainer-authored include**, not an accident. `app/test/vitest.config.ts`:
   ```ts
   include: [
     "src/**/*.test.{ts,tsx}",
     "test/*.test.{ts,tsx}",
   ],
   ```
   The suggestion is to *"narrow the include pattern so it only applies under `app/test`"* — that pattern is already exactly `test/*.test.{ts,tsx}`, and this file matches it.
2. **Eight sibling specs already live there**, several of which are the same genre as mine — a test that asserts on a non-`src` repo file: `info-plist-required-keys.test.ts`, `checklist-parser.test.ts`, `coverage-matrix-parser.test.ts`, `core-rpc-node.test.ts`, `core-rpc-playwright-helper.test.ts`, plus the three `OAuth*.test.tsx`. `app/test/` is evidently where repo-artifact tests belong; `app/src/**` is for co-located component/module tests.
3. **This spec has no module under test.** Its subjects are `app/package.json` and `scripts/run-dev-win.cmd`. There is no `app/src/**` file to co-locate *with*; putting it under `src/` would place a build-tooling test inside the shipped React source tree, and it would then be picked up by `coverage.include: ["src/**/*.{ts,tsx}"]` and pollute frontend coverage.

Happy to move it if a maintainer prefers — just say the word and I'll relocate it in a follow-up commit.

### Status

- CRLF / Blinter `E018` — resolved structurally in `9d81ab4` (no labels, no jumps ⇒ line endings cannot change behaviour), and the thread is now marked addressed. Thanks for re-checking it.
- Scoop coverage (@greptile-apps) — fixed earlier; `Greptile Review` is green on the current head.
- `format:check` — `163fac01` fixes the `arrowParens: "avoid"` violation my previous commit introduced.

Still outstanding and honestly disclosed: **`pnpm dev:app:win` has not been executed on a real Windows machine from this branch.** Everything above is static/unit verification. I'll add the manual-run confirmation as soon as I'm at a Windows box.

M3gA-Mind:
Maintainer review — no commits pushed (`maintainerCanModify` is false on this PR in any case, so only you can update the branch, @Mustaqeem66).

### First, the thing that matters most: there is now a competing PR

**#5788 changes the same `app/package.json:13` line to the identical value and adds its own `scripts/run-dev-win.cmd`.** Both PRs replace

```json
"dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh"
```

with `"..\\scripts\\run-dev-win.cmd"`. They cannot both merge; the second will conflict.

I am flagging this rather than deciding it, but for whoever does decide, the honest comparison:

- **This PR (#5342) is the earlier one (2026-08-03) and the better-argued fix.** The root-cause analysis is correct and is the part I most wanted to verify: `cmd.exe /d /s /c` really does strip the first and last quote of the command string, so quoting the interpreter path at the package.json layer cannot work — the quotes are precisely the characters `/S` removes. It also carries a genuine regression test that models that parsing, and the LF/CRLF reasoning behind avoiding labels (cmd resolves a label by byte offset, re-reading in 512-byte chunks) is real and not something a reviewer would have thought to check.
- **#5788 is broader but shallower on this specific bug.** It fixes three further runtime defects in `run-dev-win.sh` that this PR does not touch (unconditional `load-dotenv.sh` on a fresh clone, a `pnpm tauri:ensure` script that does not exist, and a required `cargo-tauri.exe` that nothing stages) plus the README wording. It is currently `MERGEABLE`; this one is not.

My suggestion: keep this PR's `.cmd` launcher, quoting analysis and test, and let #5788 keep only its `run-dev-win.sh` fixes and README hunk. That way neither contributor's real work is thrown away. But that is the maintainers' call.

### The bug is still live on `main`

`app/package.json:13` on `fa044d388` is still the broken quoted-path form, and `scripts/run-dev-win.cmd` does not exist. So whichever PR lands, this is still worth landing.

### Conflicts — both small

**`.gitattributes`** — `main` gained a trailing block since you branched (a `tests/fixtures/memory_golden/**/*.db binary` rule with a comment). Your two lines still belong immediately after `*.ps1 text eol=crlf`; keep `main`'s new block below them. No interaction.

**`app/package.json`** — `main` rewrote the *adjacent* `dev:app` line, which is now simply `"bash ../scripts/run-dev-macos.sh"` instead of the long inline command your base had. Take `main`'s `dev:app` line and your `dev:app:win` line. Purely an adjacent-line collision; nothing about the two changes interacts.

### Two CodeRabbit threads still open, both quick and both fair

1. **Test location.** `app/test/dev-app-win-launcher.test.ts` sits outside the co-location path the repo expects; move it under `app/src/**` (or narrow the Vitest include). Worth doing — it is currently the only spec under `app/test/`.
2. **The `goto`/`call` regex has a real hole.** `/(^|\s)(goto|call)\s/i` misses `@goto` and `@call`, which cmd accepts. Since that assertion is the *whole* guarantee behind "line endings cannot change behaviour", a future edit could add `@goto :retry` and the test would stay green. CodeRabbit's `/(^|[\s&|()])@?(goto|call)\s/i` closes it. This one I would not skip — it is the difference between a property test and a test that looks like one.

Neither is a design objection. With those two threads answered and the two conflicts resolved, this is ready from my side.

PR #5346, opened 2026-08-03
fix(memory_tree): diagnose embeddings from the real resolver ladder

### LINKED ISSUE #5330: memory_tree doctor reports embeddings_unconfigured while embeddings are configured and working
## What happens

On an instance whose embeddings are configured and verified working:

```
openhuman.embeddings_test_connection {}
→ {"success": true, "provider": "cloud", "model": "embedding-v1", "actual_dimensions": 1024}

openhuman.memory_tree_doctor {}
→ "first_blocking_cause": {"code": "embeddings_unconfigured"},
  "healthy": false,
  stage embeddings: ok=false, "no embeddings provider configured — semantic recall is off"
```

The doctor names the one thing that is not wrong, and `first_blocking_cause` is what the UI surfaces as the remediation to follow.

## Why

Two different config fields, and the doctor reads the one nothing sets.

`src/openhuman/memory_tree/health/doctor.rs:140`:

```rust
let embeddings_provider = config
    .memory_tree
    .embedding_endpoint
    ...
    .or_else(|| config.embeddings_provider.clone())   // top-level, Option<String>, default None
```

Everything else reads `config.memory.embedding_provider`:

- `embeddings_get_settings` → `src/openhuman/embeddings/rpc.rs:57`
- `embeddings_test_connection`, the same settings block
- the tree's own embedder factory, which resolves through **six** steps (explicit Ollama override → `none` opt-out → workload local model → OpenAI-compatible → **logged-in managed cloud** → none) in `src/openhuman/memory_tree/score/embed/factory.rs:138`

`config.embeddings_provider` is top-level workload routing and defaults to `None` (`src/openhuman/config/schema/types.rs:386`, `:803`). The factory's own tests document the distinction explicitly (`factory.rs:517` — "`embeddings_provider` (here unset ...) vs `memory.embedding_provider`").

So the doctor implements a 1-step approximation of a 6-step ladder and reports `unconfigured` for every configuration that resolves at steps 3–5, including the default managed-cloud one.

## Expected

The doctor should report what the factory actually resolves. Extracting `resolve_embedder_choice` (already private in `factory.rs`) and reporting from its result would make the stage correct by construction, and keep it correct when the ladder changes.

## Impact

`healthy: false` plus a remediation key pointing at embeddings setup, on an instance where embeddings are fine. It sends the user to change a setting that is already correct, and it masks whatever the real blocking cause is — in the instance above, `summarizer_unavailable`, reported in the same response as a later stage.

Mustaqeem66:
## Summary

- The memory-tree health doctor's `embeddings` stage re-derived provider resolution instead of asking the embedder factory, so it reported `embeddings_unconfigured` on installs where embedding demonstrably works.
- Add `resolved_embedder()` + `ResolvedEmbedder` to `score/embed/factory.rs`: a diagnostic view of the factory's existing resolution ladder that reports *which provider would be selected* without handing back an embedder or touching the network.
- Route the doctor's `embeddings` stage through it, so the diagnosis is the factory's own answer rather than a parallel approximation.
- Because the embeddings stage sits near the front of the ordered stage list, the bogus failure became `first_blocking_cause` and **masked the genuine one** — this restores the doctor's core promise of "one actionable answer".
- Adds 7 tests (5 factory, 2 doctor) and makes the doctor's `test_config()` hermetic.

> **Note on the issue's file paths:** #5330 cites `src/openhuman/memory_tree/health/doctor.rs` and `src/openhuman/memory_tree/score/embed/factory.rs`. Those paths don't exist; the real ones are under `memory/tree/`. The described behaviour is otherwise exactly as reported.

## Problem

`run_doctor` computed the embeddings stage like this:

```rust
let embeddings_provider = config
    .memory_tree
    .embedding_endpoint
    .as_deref()
    .filter(|s| !s.trim().is_empty())
    .map(|_| "ollama-override".to_string())
    .or_else(|| config.embeddings_provider.clone())
    .filter(|s| !s.trim().is_empty());
```

That is a **one-step approximation of a six-step ladder**. `resolve_embedder_choice` in `score/embed/factory.rs` — the single source of truth walked by *both* `build_embedder_from_config` (read) and `build_write_embedder` (write) — resolves in this order:

1. explicit Ollama override (`memory_tree.embedding_endpoint` **and** `embedding_model`, both non-empty)
2. `embeddings_provider == "none"` → deliberate opt-out
3. `config.workload_local_model("embeddings")` → local Ollama
4. `OpenAiCompatEmbedder::try_from_config` → user OpenAI / custom endpoint
5. `cloud_session_available` → managed cloud
6. nothing usable

Three concrete defects:

1. Step 1 requires endpoint **and** model; the doctor checked only the endpoint (so it could also report a provider the factory would reject).
2. It fell back to the top-level `config.embeddings_provider`, which is workload routing and **defaults to `None`**.
3. It never saw rungs 3–5 at all — including rung 5, the path a **default logged-in install** lands on.

So the common case (logged in, no explicit embeddings config) embeds fine but is reported `embeddings_unconfigured`. In the reporter's case that masked `summarizer_unavailable`, and sent them to "fix" a setting that was already correct.

## Solution

Rather than teach the doctor to read one more config field — which would re-break the moment the ladder changes — expose the ladder's answer:

```rust
pub enum ResolvedEmbedder {
    Provider(&'static str), // same label Embedder::name() reports
    OptOut,                 // `none` — off by choice, not a fault
    Unconfigured,
}

pub fn resolved_embedder(config: &Config) -> ResolvedEmbedder;
```

The doctor then just matches on it. Design notes:

- **New public API instead of making `resolve_embedder_choice` / `EmbedderChoice` public.** `EmbedderChoice::OpenAiCompat` owns a constructed `OpenAiCompatEmbedder`; publishing it would leak embedder internals and hand callers a built embedder they didn't ask for. `ResolvedEmbedder` is `Copy` and carries only a label.
- **No network, nothing escapes.** The ladder allocates and does one `Path::exists()`; on the OpenAI-compatible rung it constructs an `OpenAiCompatEmbedder` that is dropped immediately after its label is read (thanks @greptile-apps for catching that my first doc comment overclaimed here). It does not set the process-global degraded flags — only the factories' terminals do. The doctor already performs synchronous SQLite reads, so this is comfortably the cheapest thing it does.
- **`Err(_)` is folded into `Unconfigured`.** A ladder error means the configured OpenAI-compatible provider could not be constructed — from the user's perspective embeddings are unusable, which is what `Unconfigured` says. The write path still fails loudly with the underlying error.
- **The `none` opt-out stays `ok`.** It's a deliberate choice (like `scheduler_gate = off`), reported with an honest note that can't read as a working provider — preserving the behaviour a previous CodeRabbit review asked for.

### Tests

`factory.rs` (5): the logged-in cloud default, the `none` opt-out staying distinct from unconfigured, the genuinely-unconfigured case, the OpenAI-compatible backend, and a **parity guard** asserting the diagnostic's label equals `build_embedder_from_config(...).name()` for `ollama`, `openai` **and** `cloud` — so adding a ladder rung, or renaming a label, and updating only one match arm fails CI.

`doctor.rs` (2): `working_cloud_provider_is_not_reported_unconfigured` (the #5330 regression, with a precondition assert that the ladder really resolves to cloud, so a failure can only be the doctor's fault) and `workload_local_model_is_reported_as_configured` (a mid-ladder rung).

`doctor.rs`'s `test_config()` now plants `config_path` inside the `TempDir`. The ladder's last rung probes the filesystem for `auth-profiles.json` next to the config file, so without this the doctor suite would read the developer's real session and pass or fail depending on whether the machine happens to be logged in. This mirrors the harness `factory.rs` already uses.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 7 new tests; failure/edge cases covered by `resolved_embedder_reports_unconfigured_when_nothing_resolves`, the `Err(_) → Unconfigured` arm, and the preserved `misconfigured_workspace_reports_embeddings_as_first_blocking_cause`.
- [x] **Diff coverage ≥ 80%** — verified by CI, not locally: **Rust Core Coverage (cargo-llvm-cov)** and the **PR CI Gate** both passed on this branch. I have no Rust toolchain locally (see *Validation Blocked*), so CI is the evidence here.
- [x] Coverage matrix updated — `N/A: behaviour-only fix`; no feature rows added, removed or renamed. The **Coverage Matrix Sync** check passed.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no matrix rows are affected by this change.`
- [x] No new external network dependencies introduced — `resolved_embedder` issues no network request on any rung; the new tests use a `TempDir` and a stub file only. No new crates.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface changes.` The doctor report's *shape* is unchanged; only the value of one stage is corrected.

## Impact

- **Runtime/platform:** all platforms, diagnostic-only. Affects the `memory_tree doctor` agent tool / CLI / RPC output and the status panel's `first_blocking_cause`.
- **Behaviour:** installs previously mislabelled `embeddings_unconfigured` will now show the resolved provider, and `first_blocking_cause` will surface the *real* problem (or `healthy`). Installs with genuinely no provider are unchanged.
- **Performance:** one extra ladder walk per `run_doctor` call — allocations plus a single `Path::exists()`, alongside the several SQLite queries the doctor already runs.
- **Security / migration / compatibility:** none. No serialized shape changed, no config migration, no new deps. `resolved_embedder` is additive; no existing signature changed.

## Related

- Closes: #5330
- Follow-up PR(s)/TODOs: the `status`/`pipeline_status` surface derives its own embeddings view; worth auditing whether it shares this bug, but it's out of scope here and I didn't want to widen the diff.

---

### Linear Issue

- Key: `N/A` — external contributor, no Linear access.
- URL: `N/A`

### Commit & Branch

- Branch: `Mustaqeem66:fix/5330-doctor-embeddings-ladder`
- Commit SHA: see the branch head (latest commit addresses the Greptile review)

### Validation Run

> I have no Rust/pnpm toolchain and no network in the environment this was authored in, so **nothing below was run locally.** Each box is ticked against the equivalent **CI job on this branch**, which is stronger evidence than a local run anyway. Full detail in *Validation Blocked*.

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no frontend files changed.` CI's **Frontend Checks** correctly skipped for this diff.
- [x] `pnpm typecheck` — `N/A: no TypeScript changed.` This PR touches two `.rs` files only.
- [x] Focused tests: **NOT RUN locally** — covered by CI's **Rust Core Coverage (cargo-llvm-cov)** ✅, which builds and runs the crate's test suite including the 7 new tests. Intended local commands were `cargo test -p openhuman memory::tree::health::doctor` and `... memory::tree::score::embed::factory`.
- [x] Rust fmt/check (if changed): **NOT RUN locally** — covered by CI's **Rust Quality (fmt, clippy)** ✅ and **Rust Feature-Gate Smoke (gates off)** ✅.
- [x] Tauri fmt/check (if changed): `N/A: no Tauri code touched.`

### Validation Blocked

- `command:` `cargo test` / `cargo fmt --check` / `cargo clippy` / `pnpm typecheck`
- `error:` no Rust toolchain (`cargo`/`rustc` absent) and no network access to install one or reach the crate registry in the environment this change was authored in.
- `impact:` **I could not compile or execute anything before pushing.** I compensated by hand-verification: both files were reconstructed byte-for-byte against their upstream blobs before editing (confirmed by git blob SHA) so the diff contains only intended changes; I traced all six ladder rungs against every pre-existing doctor test to confirm none flip (`"none"` → `OptOut` → still `ok`; `"ollama:bge-m3"` → rung 3 → `Provider("ollama")` → still `ok`; the misconfigured-workspace test → rung 6 → still `Unconfigured`, which the hermetic `config_path` change guarantees). **CI has since confirmed this**: fmt, clippy, the feature-gate smoke build and the coverage gate all pass on the branch. I'll keep treating CI as the authority and fix anything it flags promptly.

Disclosing plainly per the repo's AI-authorship requirement: **this PR was written with AI assistance.** The analysis, the design decision, and every claim above are ones I stand behind and can defend in review, and the validation I could not perform is stated rather than checked off silently.

### Behavior Changes

- Intended behavior change: the doctor's `embeddings` stage reports the provider the embedder factory would actually select, instead of a locally re-derived approximation.
- User-visible effect: `openhuman doctor` / the memory-tree status panel stop claiming "no embeddings provider configured" on working installs, and `first_blocking_cause` now names the real blocker instead of being masked by the false embeddings failure.

### Parity Contract

- Legacy behavior preserved: the stage id (`embeddings`), the `FailureCode::EmbeddingsUnconfigured` code, and all three note strings are unchanged verbatim. A genuinely unconfigured install produces a byte-identical stage to before. The `none` opt-out keeps its `ok`-with-honest-note treatment.
- Guard/fallback/dispatch parity checks: `resolved_embedder` and the two factories share one `resolve_embedder_choice` call, so they cannot diverge; `resolved_embedder_label_agrees_with_the_built_embedder` asserts the diagnostic's label equals `Embedder::name()` for `ollama`, `openai` and `cloud`; and the doctor test asserts a precondition on the ladder before asserting on the stage.

- Duplicate PR(s): none — I searched open PRs for `doctor`, `embeddings`, `resolved_embedder` and #5330 references before starting.
- Canonical PR: this one.
- Resolution: `N/A`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding health checks to accurately reflect the configured provider and resolution settings.
  * Added clearer status reporting for enabled providers, intentional opt-out, and unavailable configurations.
  * Improved support for managed cloud, local-model, and OpenAI-compatible embedding setups.

### DIFF
--- a/src/openhuman/memory/tree/health/doctor.rs
+++ b/src/openhuman/memory/tree/health/doctor.rs
@@ -97,6 +97,9 @@ pub fn run_doctor(config: &Config) -> DoctorReport {
     use crate::openhuman::memory::queue::store as queue;
     use crate::openhuman::memory::queue::types::JobStatus;
     use crate::openhuman::memory::store::chunks::store as chunks;
+    use crate::openhuman::memory::tree::score::embed::factory::{
+        resolved_embedder, ResolvedEmbedder,
+    };
 
     let degraded = current_degraded_state();
     let counters = DoctorCounters {
@@ -134,29 +137,34 @@ pub fn run_doctor(config: &Config) -> DoctorReport {
         ));
     }
 
-    // 1. Routing/config sanity — is *any* embeddings provider configured?
-    //    (`build_write_embedder` skips embedding when none is, so this is the
-    //    most common "empty wiki" root cause.)
-    let embeddings_provider = config
-        .memory_tree
-        .embedding_endpoint
-        .as_deref()
-        .filter(|s| !s.trim().is_empty())
-        .map(|_| "ollama-override".to_string())
-        .or_else(|| config.embeddings_provider.clone())
-        .filter(|s| !s.trim().is_empty());
-    stages.push(match embeddings_provider.as_deref() {
+    // 1. Routing/config sanity — will the embedder factory actually resolve a
+    //    provider? (`build_write_embedder` skips embedding when it can't, so
+    //    this is the most common "empty wiki" root cause.)
+    //
+    //    Ask the factory's own resolution ladder rather than re-deriving it
+    //    here. This stage used to check `memory_tree.embedding_endpoint` and
+    //    then fall back to the top-level `embeddings_provider` workload field —
+    //    a one-step approximation of a six-step ladder, and `embeddings_provider`
+    //    is `None` by default. Every install resolving further down the ladder
+    //    (unified local model, user OpenAI-compatible endpoint, or the
+    //    logged-in managed cloud that most installs land on) was therefore
+    //    reported as `embeddings_unconfigured` while embedding demonstrably
+    //    worked — masking the real blocking cause and sending users off to
+    //    "fix" a setting that was already correct.
+    stages.push(match resolved_embedder(config) {
         // Explicit `none` opt-out: semantic recall is off by the user's choice,
         // not a fault. Reported `ok` (consistent with a `scheduler_gate=off`
         // pause and the write-path opt-out treatment) but with an honest note,
         // so the prior "provider configured: none" can't read as a working
         // embeddings provider. (CodeRabbit on doctor.rs)
-        Some("none") => StageHealth::ok(
+        ResolvedEmbedder::OptOut => StageHealth::ok(
             "embeddings",
             "embeddings disabled by you (provider = none) — semantic recall is intentionally off",
         ),
-        Some(p) => StageHealth::ok("embeddings", format!("provider configured: {p}")),
-        None => StageHealth::bad(
+        ResolvedEmbedder::Provider(p) => {
+            StageHealth::ok("embeddings", format!("provider configured: {p}"))
+        }
+        ResolvedEmbedder::Unconfigured => StageHealth::bad(
             "embeddings",
             PipelineFailure::new(FailureCode::EmbeddingsUnconfigured),
             "no embeddings provider configured — semantic recall is off",
@@ -286,7 +294,24 @@ mod tests {
         cfg.workspace_dir = tmp.path().to_path_buf();
         cfg.memory_tree.embedding_endpoint = None;
         cfg.memory_tree.embedding_model = None;
+        // Plant config_path in the tempdir. The embeddings stage now walks the
+        // factory's ladder, whose last rung probes for `auth-profiles.json`
+        // next to the config file — without this the tests would read the
+        // developer's real session and pass or fail depending on whether the
+        // machine happens to be logged in. (Mirrors the factory's own harness.)
+        cfg.config_path = tmp.path().join("config.toml");
         (tmp, cfg)
+    }
+
+    /// Simulate a logged-in install: the ladder's cloud rung only checks that
+    /// `auth-profiles.json` exists next to the config file.
+    fn touch_auth_profile(cfg: &Config) {
+        let path = cfg
+            .config_path
+            .parent()
+            .map(|p| p.join("auth-profiles.json"))
+            .expect("config_path has a parent");
+        std::fs::write(&path, "{}").expect("write stub auth-profiles.json");
     }
 
     #[test]
@@ -452,6 +477,84 @@ mod tests {
             .expect("storage stage present");
         assert!(!storage.ok);
         assert!(report.degraded.storage);
+    }
+
+    /// The embeddings stage must be diagnosed from the factory's resolution
+    /// ladder, not from a local approximation of it.
+    ///
+    /// Regression: a logged-in install with no explicit embeddings config
+    /// resolves to the managed cloud and embeds fine, but this stage used to
+    /// read the top-level `embeddings_provider` workload field — which is
+    /// `None` by default — and reported `embeddings_unconfigured`. Because
+    /// embeddings sit near the front of the stage list that became
+    /// `first_blocking_cause`, hiding the genuine failure further down and
+    /// pointing the user at a setting that was already correct.
+    #[test]
+    fn working_cloud_provider_is_not_reported_unconfigured() {
+        use crate::openhuman::memory::tree::score::embed::factory::{
+            resolved_embedder, ResolvedEmbedder,
+        };
+        let _g = super::super::test_guard();
+        let (_tmp, mut cfg) = test_config();
+        // A default install: no override, no workload model, no OpenAI-compatible
+        // backend — just a session, which is all the cloud rung needs.
+        cfg.embeddings_provider = None;
+        touch_auth_profile(&cfg);
+
+        // Precondition: the ladder really does resolve a usable provider here,
+        // so a non-ok embeddings stage below can only be the doctor's own fault.
+        assert_eq!(
+            resolved_embedder(&cfg),
+            ResolvedEmbedder::Provider("cloud"),
+            "test setup: the factory must resolve to the managed cloud"
+        );
+
+        let report = run_doctor(&cfg);
+        let embed = report
+            .stages
+            .iter()
+            .find(|s| s.stage == "embeddings")
+            .expect("embeddings stage present");
+        assert!(
+            embed.ok,
+            "a provider the factory would really use must not be a fault, got: {}",
+            embed.note
+        );
+        assert!(
+            embed.note.contains("cloud"),
+            "the note should name the resolved provider, got: {}",
+            embed.note
+        );
+        assert_ne!(
+            report.first_blocking_cause.clone().map(|c| c.code),
+            Some(FailureCode::EmbeddingsUnconfigured),
+            "embeddings must not mask the real blocking cause"
+        );
+    }
+
+    /// The stage must also see a provider that resolves at the *middle* of the
+    /// ladder (unified workload model), not just the terminal cloud rung.
+    #[test]
+    fn workload_local_model_is_reported_as_configured() {
+        let _g = super::super::test_guard();
+        let (_tmp, mut cfg) = test_config();
+        cfg.embeddings_provider = Some("ollama:all-minilm:latest".into());
+        cfg.local_ai.runtime_enabled = true;
+        cfg.local_ai.embedding_model_id = "all-minilm:latest".to_string();
+        // Deliberately NOT logged in: the local model alone must be enough.
+
+        let report = run_doctor(&cfg);
+        let embed = report
+            .stages
+            .iter()
+            .find(|s| s.stage == "embeddings")
+            .expect("embeddings stage present");
+        assert!(embed.ok, "local Ollama embeddings are a real provider");
+        assert!(
+            embed.note.contains("ollama"),
+            "the note should name the resolved provider, got: {}",
+            embed.note
+        );
     }
 
     #[test]
--- a/src/openhuman/memory/tree/score/embed/factory.rs
+++ b/src/openhuman/memory/tree/score/embed/factory.rs
@@ -649,7 +649,7 @@ mod tests {
         assert_eq!(e.name(), "ollama");
     }
 
-    // ── resolved_embedder (health-doctor diagnostic view) ────────────────────
+    // resolved_embedder (health-doctor diagnostic view)
     //
     // `resolved_embedder` is the ladder's answer *without* building anything.
     // Its whole value is that it cannot drift from the factories, so the tests
--- a/src/openhuman/memory/tree/score/embed/factory.rs
+++ b/src/openhuman/memory/tree/score/embed/factory.rs
@@ -185,6 +185,48 @@ fn resolve_embedder_choice(config: &Config) -> Result<EmbedderChoice> {
     Ok(EmbedderChoice::NoProvider)
 }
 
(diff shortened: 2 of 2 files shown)


Mustaqeem66:
Thanks @greptile-apps — both points were fair and are addressed in e7333d7.

**1. Parity guard omitted the `cloud` case.** You're right, and it was the worst one to omit: `cloud` is the rung most installs actually land on, and it's the rung this whole PR exists because of. The guard's docstring promised "every provider the ladder can select" while covering two of three, so a rename of the `"cloud"` literal in either `build_cloud_embedder` or the `Ok(EmbedderChoice::Cloud)` arm would have sailed through — exactly the drift the test is supposed to prevent. Added, reusing `touch_auth_profile`:

```rust
(
    // The terminal rung, and the one most installs actually land
    // on — so a rename of the `cloud` label must break this guard
    // too. (greptile on the parity guard)
    "cloud",
    Box::new(|c: &mut Config| {
        touch_auth_profile(c);
    }),
),
```

The loop body already clears `embedding_endpoint`/`embedding_model` before running each setup, and `test_config()` plants `config_path` in a `TempDir`, so the cloud case resolves through rungs 1–4 without matching and terminates at rung 5 — no cross-case leakage. (Block body rather than `Box::new(touch_auth_profile)` to keep the `&Config`/`&mut Config` coercion explicit and avoid a `redundant_closure` argument.)

**2. "no embedder is constructed" was inaccurate on the OpenAI-compatible rung.** Correct, and I'd rather the doc under-promise than mislead someone reasoning about cost. `try_from_config` does construct an `OpenAiCompatEmbedder`; it's dropped right after `.name()` is read. Reworded to state what's actually guaranteed:

> Cheap and side-effect-free: it allocates, probes one path, and (only on the OpenAI-compatible rung) constructs an `OpenAiCompatEmbedder` that is dropped immediately after its label is read. No network request is issued on any rung, and no embedder escapes.

The two properties that matter for calling this from the doctor — **no network I/O** and **no degraded-flag mutation** (only the factories' terminals touch those) — both still hold on every rung, so the change is to the wording, not the contract.

---

For the record on process: I have no Rust toolchain or network locally, so CI is doing the verifying here — fmt, clippy, the feature-gate build and the coverage gate were all green on the previous commit, and I'll watch this one. Flagged plainly in the PR's *Validation Blocked* section rather than quietly ticking boxes.

Mustaqeem66:
## Conflicts resolved (`d3abfc8`)

Rebasing wasn't available in the environment I authored this in, so I resolved the conflict by relocating the colliding block within the file. Details below so a reviewer can verify it cheaply.

### What was actually conflicting

Only **one** region, and only in `src/openhuman/memory/tree/score/embed/factory.rs`.

Since this branch was cut (merge-base `210cd6c`), `main` is 22 commits ahead and added to that file:

- `redact_ladder_error()` + `effective_embedder_slug()`, inserted **after** `build_write_embedder`
- **nine new tests appended at the end of `mod tests`**

This branch adds:

- `ResolvedEmbedder` + `resolved_embedder()`, inserted **before** `build_write_embedder`'s doc comment
- **five new tests appended at the end of `mod tests`**

The two production-code insertions are ~60 lines apart and never collided. The conflict was entirely the two independent "append at the end of `mod tests`" edits landing on the same anchor — the closing `}` of the test module.

`src/openhuman/memory/tree/health/doctor.rs` did **not** conflict: its blob is identical at the merge-base and at `main` (`fd29ed6`), so it merges clean and I did not touch it.

### How it was resolved

Moved the five `resolved_embedder` tests from the end of `mod tests` to just after `none_provider_returns_inert()` — roughly 100 unchanged lines clear of main's insertion point.

- **No test body, name, assertion or production line was changed.** The file's multiset of lines is byte-identical to the previous branch head; only the position of one block moved.
- Test order is irrelevant in Rust, and the helpers those tests use (`test_config()`, `touch_auth_profile()`, `build_embedder_from_config()`) are module-scoped and order-independent.
- The diff stat is unchanged at **+270 −15 across 2 files**, which is itself evidence nothing was gained or dropped.

Verified by reconstructing all three merge sides byte-for-byte (confirmed via git blob SHA against the upstream blobs) and running a 3-way merge against merge-base `210cd6c`: **0 conflict regions, 0 lines dropped from `main`, 0 lines dropped from this branch**, braces balanced. GitHub now reports the PR as no longer conflicting.

### One thing worth a maintainer's call

`main`'s new `effective_embedder_slug()` walks the same `resolve_embedder_choice` ladder as this PR's `resolved_embedder()`. They coexist and compile fine, but they aren't equivalent:

- `resolved_embedder()` distinguishes `OptOut` from `Unconfigured`, and returns `openai.name()` on the OpenAI-compatible rung
- `effective_embedder_slug()` flattens OpenAI-compatible down to `"custom"`

I deliberately did **not** unify them here — that's a behaviour decision, not a conflict resolution, and folding it in would widen a diff that is currently a pure move. Happy to follow up in a separate PR if you'd like them consolidated, and equally happy to do it here if you'd rather it not ship as two overlapping accessors.

As before: I still have no Rust toolchain locally, so nothing was compiled or executed — the validation above is structural/textual only. **CI on this branch remains the authority**, and I'll fix anything it flags.

M3gA-Mind:
Thanks for this, @Mustaqeem66 — closing because both target files have been deleted.

`memory/tree/health/doctor.rs` was removed by `ba088ec20` (*chore(memory): remove entire legacy memory subsystem*), and `memory/tree/score/embed/factory.rs` went with the entire `memory/tree/score/` tree — there is no `resolved_embedder` / `ResolvedEmbedder` anywhere on `main`. `run_doctor` now lives in `memory/tree/health/report.rs` and delegates the whole stage set to the driver via `MemoryMaintenance::diagnose`, so the host no longer derives the embeddings stage at all. The parallel approximation you were replacing doesn't exist to be replaced.

Worth recording: you had already flagged that #5330's cited paths were wrong. You were right at the time — the paths have since gone away entirely. Thanks for the care on that.

PR #5433, opened 2026-08-07
Promote main -> release (conflict-resolved)

M3gA-Mind:
> **Base is `release`.** This is the promote direction — it advances what staging and production build from. A separate PR carries the mirror (release → main, base `main`); do not confuse their resolutions, they hold the version files in opposite directions.

## What this does

Brings `release` to `main`'s tree at `8deb5f276`.

**Only the six version-bearing files differ from main**, and they are held at release's higher **v0.63.10** rather than main's v0.63.6 — a release must never go backwards.

```
$ git diff --stat upstream/main
 Cargo.lock                    | 2 +-
 Cargo.toml                    | 2 +-
 app/package.json              | 2 +-
 app/src-tauri/Cargo.lock      | 4 ++--
 app/src-tauri/Cargo.toml      | 2 +-
 app/src-tauri/tauri.conf.json | 2 +-
 6 files changed, 7 insertions(+), 7 deletions(-)
```

Every hunk in those six is the same substitution, `0.63.6` → `0.63.10`. Nothing else in the tree differs from main in either direction.

It is a real two-parent merge, which is the point of doing it this way:

```
$ git log -1 --pretty=%P
2ba803a246f8af20e3278c8acb62f34be2b421f5   ← release, v0.63.10
8deb5f2769c97ad059be2a6acae9f23c9c00dc3b   ← main, #5427
```

## Release carried no unique code

Two independent checks, because the promote commits in release's history were hand-resolved and a wrong resolution is exactly the thing that would hide release-only content.

**By commit.** Release's nine unique commits are five version bumps (`v0.63.6` … `v0.63.10`), three squashed promotes of main's own content (#5242, #5256, #5414), and one real change — `42cdf77e7`, the appimage userns fix (#5251). That fix's two files, `scripts/release/test-strip-appimage-rpaths.sh` and `scripts/release/validate-appimage-runtime.sh`, are **byte-identical between main and release**: it reached main separately as #5252 (`9750b8077`).

**By tree**, which is the stronger check. `main` and `release` differ in **37 paths, every one a modification** — nothing added and nothing deleted in either direction:

```
$ git diff --name-status upstream/main upstream/release | awk '{print $1}' | sort | uniq -c
  37 M
```

For each of the 31 non-version paths, release's blob is **reachable in main's history**, appearing as the pre-image of a later main commit — i.e. release holds an older state of a file main has since moved past. There is no path where release's content is novel.

So taking main's tree wholesale discards nothing.

## The 14 conflicts are squash-trap artifacts

13 unmerged entries (the 14th line in the conflict output is the rename/delete half of a pair whose unmerged entry sits on the new path):

```
Cargo.lock, Cargo.toml, app/package.json,
app/src-tauri/{Cargo.lock, Cargo.toml, tauri.conf.json}        ← the six version files
app/src/components/intelligence/MemoryTreeStatusPanel.test.tsx (content)
app/src/features/human/voice/useRealtimeVoiceSession.ts        (add/add)
app/src/features/human/voice/useRealtimeVoiceSession.test.ts   (add/add)
src/openhuman/config/schema/load/mod.rs                        (content)
src/openhuman/inference/embeddings/cloud_adapter.rs            (add/add, paired with a
    rename/delete from src/openhuman/embeddings/cloud_adapter.rs)
src/openhuman/memory/tree/tree/rpc.rs                          (content)
vendor/tinycortex                                              (submodule)
```

None of them is divergence. The promotes were **squash-merged**, so no merge commit recorded main as an ancestor and the merge-base is frozen at `2219d4207`. Git therefore cannot see that release already holds main's tree, and re-derives moves **main made itself** during the kernelization as add/add and rename/delete pairs — `embeddings/cloud_adapter.rs` → `inference/embeddings/cloud_adapter.rs` is precisely that.

**Resolution:** `git read-tree -u --reset upstream/main` to set index and worktree to main's tree exactly — which is also what clears the rename/delete leftover a per-file checkout would have stranded — then the six version files restored from `upstream/release`. `MERGE_HEAD` survives that, so the result is a genuine merge commit rather than a squashed snapshot.

## Maintainer note — please read before merging

**Squashing this PR will recreate the problem it is fixing.**

The repository allows squash merges only (`allow_merge_commit: false`). Squashed, this lands as a single commit with one parent, main is **again** not recorded as an ancestor, the merge-base stays at `2219d4207`, and the next promote and the next back-merge re-fight this same conflict set — with everything merged in between added to it.

That has now cost **three hand-resolutions in one day** (#5414, this, and the mirror back-merge). It is not a git accident; it is a settings consequence, and it compounds.

Two durable fixes, either of which ends it:

1. **Enable merge commits for this branch** so a promote can land as a real merge and the merge-base advances.
2. **Unblock the `Promote main to release` workflow** so the sanctioned path works. It is currently unusable for two separate reasons: a dispatch from 17:40 is stuck `waiting` on the `Release-PR-Automation` environment approval, which `M3gA-Mind` cannot grant; and even once it runs it cannot resolve conflicts by hand, so it would fail on these 13 regardless.

Until one of those happens, every promote needs a human to re-derive this resolution from scratch, and the conflict set only grows.

## Verification

- `git diff --stat upstream/main` → the six version files, nothing else (pasted above)
- `git log -1 --pretty=%P` → two parents, release then main (pasted above)
- `git merge-base --is-ancestor upstream/main HEAD` → yes; same for `upstream/release`
- `git status --porcelain` → clean; no path present here and absent on main
- No build was run locally. CI is the gate.

## Summary by CodeRabbit

* **New Features**
  * Retry failed memory tasks directly from the status panel, with progress, success/error feedback, and refreshed results.
  * Upload custom mascot avatars using supported image formats, with size and type validation, previews, and reset controls.
  * Improved realtime voice session identity handling for more reliable connections.

* **Bug Fixes**
  * Resolved outdated failure notices after newer memory tasks complete successfully.
  * Improved credential resolution across account and workspace contexts.

* **Localization**
  * Added and updated translations for retry actions and custom image avatar settings across supported languages.


M3gA-Mind:
@coderabbitai review

Issue #5466, opened 2026-08-09
bug(memory): an explicit SourceScope argument is widened back to the ambient scope


senamakel:
On current `main`, `memory::guard::families::tests::guard_explicit_scope_argument_wins_over_the_ambient_one` fails **deterministically, in isolation** (not an ordering or parallelism artifact):

```
assertion `left == right` failed: a caller that computed a narrower scope must not be widened back
  left: Some("")
 right: Some("gmail:me")
   at src/openhuman/memory/guard/families_tests.rs:179
```

Repro:

```
cargo test --lib memory::guard::families::tests::guard_explicit_scope_argument_wins_over_the_ambient_one
```

The test passes an explicit `SourceScope::new(["gmail:me"])` to `query_source` while an ambient scope of `slack:#eng` is installed via `with_source_scope`. The explicit argument is supposed to win; what reaches the driver is `Some("")` — neither the explicit scope nor the ambient one, which suggests the explicit `Option<&SourceScope>` is being dropped somewhere and an empty scope substituted, rather than the ambient one simply overriding it.

This matters beyond the test: the assertion message is a security-shaped invariant — a caller that deliberately narrowed a scope must not have it widened underneath them.

Surfaced while repairing the #5459 / #5446 collision. Deliberately **not** fixed in that PR (it is a behavioural question for the memory-subsystem author, not a mechanical path rename), and the file is untouched there.

M3gA-Mind:
Thanks for catching this — a deterministically failing test is exactly the kind of report worth filing.

The failure is resolved as of #5469, though it is worth being explicit about how: the semantics were settled as **intersection with fail-closed**, not as "the explicit argument wins". The test is now `guard_explicit_scope_is_intersected_with_the_ambient_one` (`src/openhuman/memory/guard/families_tests.rs:167`) and asserts that a request outside the ambient allowlist returns nothing. So the widening you reported no longer happens.

If you specifically needed a caller-computed *narrower* scope to be honoured rather than intersected away, that is a live design question rather than a fixed one — please reopen or file separately and it can be picked up.

PR #5467, opened 2026-08-09
fix(core): repair main after the tinybus and memory-subsystem PRs collided

senamakel:
## What

`main` does not compile — neither Cargo world — and has not since #5459 and #5446 both merged. #5459 deleted `src/core/event_bus/` and moved the bus onto the vendored `tinybus` crate; #5446 added new references to `crate::core::event_bus`. Neither PR was wrong on its own; they were verified against different bases.

```
error[E0432]: unresolved import `crate::core::event_bus`
  --> src/openhuman/memory/guard/audit.rs:33:18
error[E0433]: cannot find `event_bus` in `core`
   --> src/openhuman/memory/binding.rs:398:26
error[E0432]: unresolved import `openhuman_core::core::event_bus`
  --> app/src-tauri/src/whatsapp_data/mod.rs:63:31
```

Every failing site is ungated code, so no feature selection avoids it.

## The fix

| File | Change |
| --- | --- |
| `memory/guard/audit.rs` | `publish_global(..)` → `BUS.publish(..)`; `event_bus::DomainEvent` → `core::events::DomainEvent` |
| `memory/binding.rs` | same, at the `MemoryDriverBindFailed` publish |
| `app/src-tauri/src/whatsapp_data/mod.rs` | free fn `register_native_global` → `BUS.native().register`, one-for-one |
| `memory/guard/provider_tests.rs` | the two step-7 tests migrated off `init_global(..).raw_receiver()` |

The shell one is worth calling out: `app/src-tauri` is a **separate Cargo world**, so it is invisible to `cargo check` at the root and surfaces only in the Tauri clippy lane or the pre-push hook.

The test migration is the only part that is not a path rename. The old API was a synchronous `tokio::sync::broadcast` receiver drained with `try_recv`; tinybus has no equivalent, because delivery routes through a broker and is asynchronous. The tests now subscribe a recording `EventHandler` to the global bus — the idiom `web_chat::event_bus`'s own tests already use — and poll:

- the refusal test polls up to 2s, **bounded** so a regression fails instead of hanging CI;
- the success-path test sleeps briefly before asserting absence, because asserting "nothing arrived" the instant after the call would pass even if something had. It still filters by driver id rather than asserting the recorder is empty, since the sibling test legitimately publishes onto the same process-global bus in parallel.

## The rest of the diff

Two more artefacts of the same collision, both mechanical:

- `app/src-tauri/Cargo.lock` never learned about `tinybus` / `tinybus-macros`.
- `cargo fmt --all` and prettier both had work to do — #5459 landed ~72 Rust files unformatted plus one `.tsx` reflow, so **the fmt lane is red on main too**. This is the bulk of the diff. Review the four files in the table above and skim the rest.

## Not fixed here

`memory::guard::families::tests::guard_explicit_scope_argument_wins_over_the_ambient_one` fails **deterministically on main, in isolation**, in a file this PR does not touch: an explicit `SourceScope` argument is widened back to the ambient scope (`Some("")` where `Some("gmail:me")` is expected). That is a behavioural question for the memory-subsystem author, not a mechanical collision, so it is #5466 rather than a guess here.

## Verification

```
cargo fmt --all -- --check                                  # clean
cargo check --lib --tests                                   # clean
cargo clippy --manifest-path app/src-tauri/Cargo.toml -- -D warnings   # clean
cargo test --lib memory::guard::provider                    # 12 passed, incl. both migrated tests
```

### DIFF
--- a/app/src-tauri/Cargo.lock
+++ b/app/src-tauri/Cargo.lock
@@ -5568,6 +5568,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tinyagents",
+ "tinybus",
  "tinychannels",
  "tinycortex",
  "tinycortex-api",
@@ -8902,6 +8903,28 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+]
+
+[[package]]
+name = "tinybus"
+version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tinybus-macros",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
+name = "tinybus-macros"
+version = "0.1.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
 ]
 
 [[package]]
--- a/app/src-tauri/src/whatsapp_data/mod.rs
+++ b/app/src-tauri/src/whatsapp_data/mod.rs
@@ -87,15 +87,12 @@ pub fn register_native_handlers() {
             ops::search_messages(&store, req).map_err(|e| format!("{e:#}"))
         },
     );
-    native.register::<IngestRequest, IngestResult, _, _>(
-        methods::INGEST,
-        |req| async move {
-            let store = ensure_store().await?;
-            // `{e:#}` renders the full anyhow chain so the underlying SQLite
-            // cause (locked / malformed / FK) survives to the scanner's log.
-            ops::ingest(&store, req).map_err(|e| format!("[whatsapp_data] ingest failed: {e:#}"))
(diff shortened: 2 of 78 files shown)


senamakel:
Superseded by #5469, which landed the same tinybus migration fixes on `main` (the `memory/guard/audit.rs` + `memory/binding.rs` retargeting, the `whatsapp_data` native-registration migration, the shell lockfile, and the formatting sweep). Verified against current `main` — all four sites are repaired there.

Closing rather than rebasing: there is nothing left in this branch that #5469 does not already cover.

The one thing #5469 did **not** pick up is #5466 — `guard_explicit_scope_argument_wins_over_the_ambient_one` still fails deterministically on `main`, where an explicit `SourceScope` is widened back to the ambient one. That issue stays open.

PR #5470, opened 2026-08-09
refactor(memory): confine raw profile SQLite, guard the tool_memory tools, collapse three re-export shims

senamakel:
## What

The follow-up pass to #5446, closing the three items that PR explicitly deferred.

### 1. `profile_conn` — raw SQLite confined behind a typed `ProfileStore`

`MemoryClient::profile_conn()` handed out an `Arc<Mutex<rusqlite::Connection>>` to three
domains outside the memory family, two of which wrote SQL inline at the call site. New
`memory/store/profile_store.rs` is now the only typed door onto `user_profile`; every one of
the 11 non-test call sites (`agent/learning/{startup,tools,schemas}.rs`,
`memory/sync/composio/providers/profile.rs`) is re-pointed, and `profile_conn()` is narrowed
from `pub(crate)` to `pub(in crate::openhuman::memory)` with exactly one caller left —
`profile_store()`'s own body. `FacetCache` survives as a learning-side newtype because
`FacetClass` is agent vocabulary that must not migrate into the memory family.

**Read this before believing the title.** This is a *confinement* win, not a *guard* win.
The profile/facet tables have no capability family in the `tinycortex_api` contract, so reads
and writes through `ProfileStore` still run beneath all seven of `MemoryGuard`'s policy steps:
no tier check, no source-scope predicate, no taint stamping, no redaction, no budget, no audit
event. What changed is the door's shape — raw SQLite reachable from three domains became one
typed store whose confinement the compiler enforces. Actually putting profile data under
policy needs a fourteenth capability family in `vendor/tinycortex` (out of scope here), or a
`GuardPolicy` half-measure that would make the `readonly` tier start rejecting the boot-time
learning rebuild loop — a behaviour change that deserves its own PR. The module docs and
`docs/specs/memory-guard-allowlist.md` say this in the same words; please don't let the diff
stat talk you out of it.

Related, also not fixed here: `user_profile` has no taint column, so provider-sourced
identities still land untainted.

### 2. `tool_memory` agent tools routed through the guard

`memory/tool_memory/tools/{list,put}.rs` went via `active_memory_client()` +
`tool_memory_store(client.memory_handle())`. They now resolve `active_memory_guard()` and go
through `as_tool_memory()`. `list` is 1:1 — `memory::tool_memory::ToolMemoryRule` *is*
`tinycortex_api::tool_memory::ToolMemoryRule`, so same rules, same order, same serialization.
`put` needs a read-back after the write because the contract method returns unit while the
tool answers with the *stored* rule; that read-back is exact, not lossy, and a concurrent
delete in the window errors rather than fabricating a rule.

**Two intended behaviour changes**, both consequences of actually being guarded:

- `memory_tools_put` now passes through `admit_write` → `SecurityPolicy::enforce_write_tier`,
  so it is **refused under the `readonly` autonomy tier**. It previously succeeded.
- Both tools now depend on the bound driver advertising `Capability::ToolMemory`.
  `tool_memory_store(memory_handle())` worked over any `Arc<dyn Memory>`; against a
  non-advertising driver these two tools now error where they used to function.

The four `ops/tool_memory.rs` handlers (`tool_rule_put/get/*_json/*_for_prompt`) are **not**
re-pointed — they have no contract twin — and their allowlist entries stay with their existing
reasons.

### 3. M8c — three re-export shims collapsed

Host-only import rewrites, no crate change and no gitlink bump: the `memory_diff` types shim,
the goals `GoalsDoc`/`GoalItem` shim, `goals/store.rs` onto the crate engine, and the
conversations type/function shim. Pure relocation — the dropped `async` on the goals wrappers
offloads nothing that was offloaded before (they were bare `engine::f(..).map_err(to_string)`,
no `spawn_blocking`), and error text is preserved at every call site.

Expect `grep -rn "tinycortex::" src/openhuman/` to go **up**. Deleting a `pub use` shim
re-points importers at `tinycortex::` directly. That is the shim leaving, not coupling
arriving.

## Bypass ratchet

`bypass_allowlist_tests.rs`: **52 → 49** entries. −4 `.profile_conn(`, −4 tool_memory,
+4 `.profile_store(`, +1 `client.rs .profile_conn(` (kept so the needle stays live and a
future re-widening is still watched). The profile slice is net **+1** on its own — that is the
honest accounting of a confinement-not-policy change, and the module docs say so. All six lint
tests green; `grep "out of scope for M4\|deferred to M5"` over `src/` and `docs/` is now empty.

## Two things about the merge base you should know

- **`origin/main` does not compile its own lib tests.** `394fabb92` left the
  `core::event_bus` → `core::bus` migration half-done: `memory/guard/audit.rs`,
  `memory/binding.rs` and `memory/guard/provider_tests.rs` still named the deleted module.
  Nothing on this branch could be verified until that was fixed, so `f14819bc6` does it. The
  two provider tests observed denials over a `raw_receiver()` that no longer exists; since the
  global `BUS` is a no-op under `cargo test`, `audit.rs` gains a `#[cfg(test)]` recorder seam
  with a watermark, preserving both assertions.
- **`9836f679a` is a 69-file `cargo fmt` of the merge base**, which was unformatted. It is
  isolated in its own commit so the substantive diff stays readable — skip it in review.
  `e13f7b1a9` likewise only regenerates the stale `app/src-tauri/Cargo.lock`.

## Verification

Two independent adversarial passes ran over this branch. Both found the same single blocker —
`tests/learning_phase4_integration_test.rs` stopped compiling when `FacetCache::new` changed
shape, invisible to `cargo test --lib` and to CI Lite's changed-file lane — which is fixed
here (`ProfileStore::for_tests` widened from `#[cfg(test)]` to `#[doc(hidden)] pub`, because
integration tests link the lib without `cfg(test)`).

| check | result |
| --- | --- |
| `cargo check` (lib) | 0 errors, 0 warnings |
| `cargo check --tests` | 0 errors |
| `cargo test --test learning_phase4_integration_test` | 2 passed |
| `cargo test --lib openhuman::memory` | 1489 passed, 1 failed — pre-existing |
| `cargo test --lib core::` | 762 passed, 5 failed — pre-existing |
| `cargo check --no-default-features` (slim) | pass — no feature-gate drift |
| `scripts/check-kernel-floor.sh` | pass — dep ratchet did not move |
| `cargo fmt --check` (root + `vendor/tinycortex`) | clean |

Every failure above was reproduced byte-identically on an `origin/main` scratch worktree
(with `f14819bc6` cherry-picked, since main otherwise cannot build its lib tests at all — the
most favourable possible reading for main). `guard_explicit_scope_argument_wins_over_the_ambient_one`
fails on main with the same panic, and `families.rs` / `families_tests.rs` / `policy.rs` and
the `vendor/tinycortex` gitlink are byte-identical to main on this branch.

`cargo check --manifest-path app/src-tauri/Cargo.toml` fails on **both** this branch and main
(`E0432: unresolved import openhuman_core::core::event_bus` in
`app/src-tauri/src/whatsapp_data/mod.rs`) — the shell has the same half-done migration. This
branch does not fix it; flagging it because it means the desktop shell is currently broken on
`main`.

Non-vacuity was proved for the new tests by breaking the production code and observing the
failure: reverting `put.rs::execute` to its pre-branch form made the readonly-tier test fail
*and* fired the bypass ratchet naming both needles; flipping `key LIKE ?2` to `key = ?2` made
the legacy-SQL oracle test fail on `pattern="skill:gmail:%:email"`.

## Not in this PR

- `store/namespace_store/` (~11.3k LOC) — 55% of all remaining movable mass and the only item
  needing a real schema/migration port into the crate. Its golden-parity harness pins table
  names via hardcoded `&'static [&str]` constants, so the constants and the schema must not
  change in the same commit or the test passes vacuously. Own sandwich.
- `agent/harness/archivist/lifecycle.rs` writes `user_profile` via `profile::profile_upsert`
  on an injected connection and is covered by neither bypass needle. Inert today (no
  production `ArchivistHook::new`). A `profile::profile_upsert(` needle would surface it for
  one allowlist entry — worth a follow-up.
- M6/M7 (HTTP adapter, portability RPC).

## Note on the history

The auto-commit hook fires in this worktree and checkpointed several deliberate
break-then-restore edits made while proving test non-vacuity (`50ee06732`…`f65e8aeb2`, and
`62067fae9`/`17b775513`). Net source effect is zero — `git diff` between the pairs is empty —
but they are churn. Squash-merge.

## Summary by CodeRabbit

- **New Features**
  - Memory rule updates now respect readonly access levels.
  - Memory writes provide clearer validation errors and confirm saved data.

- **Bug Fixes**
  - Prevented unauthorized memory-tool writes.
  - Improved consistency for profile and facet storage operations.

- **Refactor**
  - Standardized conversation, goals, and memory-diff storage access.
  - Simplified internal storage pathways while preserving existing behavior.

- **Tests**
  - Added coverage for readonly protection, profile storage, facet operations, and memory-tool workflows.
  - Expanded validation of guarded access and storage behavior.


senamakel:
Merged `upstream/main` (`c7e15ba04`) in — the earlier all-red CI was **not** this PR. Every failing job died at `actions/checkout` with `fatal: No url found for submodule path 'app/src-tauri/vendor/tauri-cef' in .gitmodules`: main at the time (`8774fe4a1`) carried two orphan gitlinks with no matching `.gitmodules` stanzas, reintroduced by the conflict resolution in `4d7a817f7`. #5469 has since removed them, so this re-run should be clean.

Four files conflicted, all in the `core::event_bus` → `core::bus` migration: `memory/guard/{audit,provider_tests}.rs`, `inference/provider/{factory,ops}_tests.rs`, plus `agent/triage/{escalation,events}.rs`. **Resolved in upstream's favour throughout**, and this PR's `f14819bc6` stopgap is dropped as a result — upstream's version asserts strictly more. Ours recorded denials through a `#[cfg(test)]` seam at the call site, which would have passed even if `BUS.publish` were broken; upstream subscribes to the real bus, so it proves the event actually arrives. That was the right trade even though it costs determinism (upstream polls with a sleep).

Post-merge, everything that was previously red is green:

| check | before | after |
| --- | --- | --- |
| `cargo test --lib openhuman::memory` | 1489 passed, 1 failed | **1490 passed, 0 failed** |
| `cargo test --lib core::` | 762 passed, 5 failed | **768 passed, 0 failed** |
| `cargo check --tests` | 0 errors | 0 errors |
| `cargo test --test learning_phase4_integration_test` | 2 passed | 2 passed |
| bypass ratchet (6 lint tests) | pass | pass |
| `cargo check --no-default-features` | pass | pass |
| `scripts/check-kernel-floor.sh` | pass | pass |
| `cargo fmt --check` | clean | clean |

The six failures I had classified as pre-existing were indeed upstream's, and upstream's own repairs cleared them — so the "pre-existing" column in the PR description is now stale in this PR's favour.

PR #5504, opened 2026-08-12
feat: add Cursor as a cloud LLM provider via local cursor-bridge sidecar

AnmolKamboj:
## Summary

- Adds **Cursor** as a cloud LLM provider, so users with a Cursor subscription (including free EDU accounts) can use Cursor models (Composer, GPT-5.x, Claude, Kimi, etc.) for chat and routing workloads.
- New `scripts/cursor-bridge`: a small local Node/TypeScript sidecar that adapts the agent-based [Cursor SDK](https://www.npmjs.com/package/@cursor/sdk) (`Agent.prompt`) to the OpenAI-compatible HTTP surface (`/v1/models`, `/v1/chat/completions`) OpenHuman already speaks. No core changes required.
- Cursor chip added to the built-in cloud providers panel (default endpoint `http://127.0.0.1:8790/v1`, bearer = Cursor API key).
- Cursor model parameters exposed by the SDK (reasoning effort, context window, fast mode) are encoded as model-ID variants (`~p=param:value`) and surfaced in Settings as a **grouped selector**: one sorted base-model dropdown plus per-parameter dropdowns below it, instead of a long flat variant list.

## Problem

- Cursor exposes models only through its agent SDK / CLI, not an OpenAI-compatible HTTP API, so it could not be configured as an OpenHuman provider at all.
- Naively listing every parameter combination as a separate model produces an unusably long, unsorted dropdown.

## Solution

- **Bridge sidecar pattern**: `scripts/cursor-bridge` runs locally, translates OpenAI-style requests into Cursor SDK calls (dedicated workspace cwd, per-key model cache, optional SSE streaming), and advertises the base model plus each supported parameter value as distinct model IDs.
- **UI grouping**: `ModelEntryField` parses the `~p=` encoding for the `cursor` provider and renders a base-model select (sorted, showing display name + model ID) with separate selects for each available parameter; selections are serialized back into the encoded ID, so routing/storage are unchanged.
- Tradeoff: the bridge must be running for the provider to respond (endpoint is configurable). A Rust-side builtin catalog entry was intentionally dropped because the catalog moved to the `tinymemory` repo — the UI preset alone is sufficient; happy to open a companion PR there if you want it first-class.

## Submission Checklist

- [x] Tests added or updated — `ModelEntryField.test.tsx` covers parse/serialize round-trip, malformed encoded IDs, and existing picker behavior (10 tests). Bridge HTTP paths stay outside the app harness.
- [x] N/A: Diff coverage ≥ 80% for `scripts/cursor-bridge` — that sidecar is not in Vitest/llvm-cov; UI changed lines are covered by the ModelEntryField tests above.
- [x] N/A: Coverage matrix — new opt-in provider; no existing feature rows added, removed, or renamed.
- [x] N/A: Feature IDs in Related — no matrix rows affected.
- [x] No new external network dependencies introduced — the bridge only talks to Cursor's API when the user configures the provider with their own key (BYOK); app tests unaffected.
- [x] N/A: Manual smoke checklist — not a release-cut surface.
- [x] N/A: Linked issue — no existing issue; Discord-first contribution.

## Impact

- Desktop settings UI + new opt-in local sidecar script. No impact on existing providers, routing, or users who don't configure Cursor.
- Security: the bridge binds to `127.0.0.1` only and requires the user's own Cursor API key as bearer; no keys are logged.

## Related

- Closes: (none — no existing issue)
- Follow-up PR(s)/TODOs: optional `tinymemory` catalog entry for a Rust-side builtin preset.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `AnmolKamboj:feat/cursor-provider`
- Commit SHA: `e13cee83d`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — passes
- [x] Focused tests: `pnpm exec vitest run --config test/vitest.config.ts src/components/settings/panels/__tests__/ModelEntryField.test.tsx` — 10 passed
- [x] N/A: Rust fmt/check — no Rust changes in this PR
- [x] N/A: Tauri fmt/check — no Tauri changes in this PR
- ESLint on changed files: 0 errors (2 pre-existing warnings in `AIPanel.tsx`, untouched by this PR)

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: users can select "Cursor" in Settings → AI → Cloud Providers, paste a Cursor API key, and pick Cursor models (with parameter controls) for any workload.
- User-visible effect: new provider chip + grouped Cursor model selector. Nothing changes for users who don't configure it.

### Parity Contract

- Legacy behavior preserved: yes — additive only; no existing provider, routing, or storage code paths modified beyond the `ModelEntryField`/`AIPanel` prop additions.

- Duplicate PR(s): none known
- Canonical PR: this one


M3gA-Mind:
Maintainer review (review only — nothing pushed to your branch).

Nice piece of work, and the sidecar framing is the right instinct: Cursor only exposes models through its agent SDK, so adapting it to the OpenAI-compatible surface OpenHuman already speaks keeps the core untouched. The `~p=param:value` model-ID encoding with a grouped selector is also a good answer to the combinatorial-dropdown problem. Four things stand between this and mergeable.

**1. The checklist gate — this is the actual `PR Submission Checklist` failure, and it is a two-minute fix.**

```
Checklist: 3/12 items satisfied (9 unchecked, 0 N/A)
```

`0 N/A` is the tell. The gate requires an N/A item to be **checked** *and* for its text to **begin** with `N/A`. You have them unchecked with the marker mid-sentence:

```diff
- - [ ] Coverage matrix updated — `N/A`: no existing feature rows changed; …
+ - [x] N/A: Coverage matrix updated — no existing feature rows changed; …
```

Same for the other eight. Tick every box and lead each non-applicable one with `N/A`. Keep genuinely-ran items distinct from N/A ones — annotate what you actually ran rather than blanket-ticking.

**2. `CONFLICTING` against `main`.** Needs a rebase. Note `builtinCloudProviders.ts` and the 14 i18n files are high-traffic and have moved under you since 2026-08-12.

**3. "No new external network dependencies introduced" is ticked, but this adds a dependency tree.** `scripts/cursor-bridge/package.json` + a 228-line `package-lock.json` pulls in `@cursor/sdk` and its transitive graph. I do not think that is disqualifying for an opt-in sidecar that is not part of the shipped binary, but it is exactly the kind of thing the checkbox exists to surface, so please describe it honestly rather than ticking it clean — a reviewer deciding on supply-chain risk needs to see it.

**4. The bridge is untested by any lane.** You are upfront about this ("Bridge HTTP paths stay outside the app harness", and CI reports `error: no test harness exists under scripts/`), and 353 lines of `src/index.ts` doing request translation, per-key model caching and SSE streaming is where the bugs will be. `scripts/mock-api/routes/__tests__/` shows the repo does run node tests under `scripts/` — worth checking whether you can hang a couple of translation-layer tests off that harness rather than shipping the sidecar uncovered. The `ModelEntryField` parse/serialize round-trip tests are good; it is the bridge that is bare.

One design question for the maintainers rather than for you: the body notes a Rust-side builtin catalog entry was dropped because the catalog moved to `tinymemory`, leaving the UI preset as the only registration. That is a reasonable call for a first cut, but it does mean the provider exists only in the frontend — worth an explicit decision before merge on whether that is the intended end state or a follow-up.

Not approving — a maintainer reviews and merges.

senamakel:
sidecars are too much to handle rn. will issue to work on

Issue #5508, opened 2026-08-12
Bug: agent cron create/update fails inconsistently across cores — permissions error, UI path works fine


Al629176:
## Summary

The agent's ability to create or modify crons is inconsistent across cores and unreliable on any given core. One core always fails with a permissions error; the other works intermittently. Cron management through the Settings UI works correctly on both cores, which rules out filesystem permissions and confirms this is an agent/policy authorization issue — likely related to how the newer workflows function interacts with cron permissions.

## Problem

| Path | Core A | Core B |
|---|---|---|
| Agent `cron_add` / `cron_update` | Always fails — permissions error | Works sometimes, fails sometimes |
| Settings → Cron UI | Works | Works |

The UI path and the agent path share the same underlying cron store, yet only the agent path is affected. This means:

- The permission check is happening at the agent tool-call layer, not at the data layer
- The failure is not a filesystem or OS permission issue
- Something in how the agent turn is authorized to call cron tools differs between the two cores, and is also non-deterministic on the working core

The intermittency on Core B is a strong signal that the authorization state is being set at session start rather than being structural — one core's session sometimes has the right context, sometimes doesn't.

## Suspected cause

The `cron_add`/`cron_update` tools require a security policy check that gates on some property of the agent turn context (e.g. an origin label, a session flag, or a capability granted at turn start). Two things point to the workflows function as the likely source of the regression:

1. The inconsistency across cores suggests the authorization context is set differently per-core, possibly based on how the core was initialized or what workflow context it carries.
2. The intermittency on Core B suggests the authorization is conditionally granted — which is consistent with a turn that sometimes carries a workflow origin and sometimes doesn't, depending on how the session was started.

Related: openhuman #5499 documents the same `cron_add` failure with the explicit error `agent turn has no origin label` — this issue is the multi-core, intermittent variant of the same root problem.

## Reproduction

1. On Core A: ask the agent to create or modify a cron job → always receives a permissions error.
2. On Core B: ask the agent to create or modify a cron job → sometimes succeeds, sometimes receives a permissions error.
3. On both cores: go to Settings → Cron UI → create or modify a cron → succeeds every time.

## Acceptance criteria

- [ ] The agent can reliably create and modify crons from any core without a permissions error.
- [ ] Cron tool authorization is consistent across cores and across sessions — no intermittent failures.
- [ ] The Settings UI cron path continues to work (no regression).
[...]

Issue #4774, opened 2026-07-10
Approval waiter/pending row leaks when a parked tool call is torn down by the turn wall-clock backstop


M3gA-Mind:
### Context
Follow-up carved out of #4751 (wall-clock backstop for wedged web turns). Flagged by the Codex reviewer (P2) and confirmed by @YellowSnnowmann in review; deferred out of #4751's scope (that PR only guarantees a *terminal event* fires — it deliberately does not re-architect approval lifetimes).

### Problem
`ApprovalGate::wait_for_decision` (`src/openhuman/approval/gate.rs`, ~737-832) only runs its cleanup — `evict_waiter` / `store::decide(Deny)` — inside the `tokio::time::timeout(...).await` match arms. There is **no `Drop` guard** on the waiter.

Now that a turn future can be torn down *externally* — the harness `max_wall_clock_ms` (600s, #4746) or the outer web backstop (900s) firing while a tool call is parked in the gate — that cleanup is skipped when the future is dropped. The waiter / pending row is left dangling until its own store TTL sweeps it. A later "yes"/"no" arriving before the approval's original expiry is then routed to a dead approval request and returns without starting a new chat turn.

### Blast radius (why it was deferred, not blocking)
- Needs a ≥600s user approval wait.
- The turn already ends `turn_timeout` (retryable), so the user just re-asks and a fresh approval is created.
- The stale row is reaped by the store TTL.

### Proposed fix
- Add a `Drop`-based waiter guard in `gate.rs` so eviction/deny runs on external teardown too, **and/or**
- exclude approval-parked time from the wall-clock budget so a legitimately parked turn isn't counted against the ceiling.
- Ships with a regression test that tears the waiter future down externally and asserts the waiter/pending row is cleaned up.

### Refs
- PR #4751
- Codex thread: https://github.com/tinyhumansai/openhuman/pull/4751#discussion_r3553224473

PR #5514, opened 2026-08-12
fix(approval): route WebChat approval parks via origin fallback

### LINKED ISSUE #5499: Bug: cron_add blocked by security policy — agent turn has no origin label
## Summary

The `cron_add` tool is blocked by a security policy gate with the error "agent turn has no origin label", preventing users from scheduling tasks from within a chat/agent turn. This is an app-level authorization bug, not a filesystem or OS permission issue.

## Problem

When an agent attempts to call `cron_add` to create a scheduled job, the security policy rejects the call because the agent turn carries no origin label. The policy gate requires an origin label to authorize scheduling operations from agent turns, but the label is either not being set or not being propagated when the turn is initiated from the chat interface.

Error: `agent turn has no origin label`

Observed behaviour:
- Cron scheduling via the agent fails immediately on tool call.
- Restarting the app/daemon may transiently clear the issue (stale session/policy state), but the root cause is a missing origin label on the turn context — so it recurs.
- macOS disk access permissions are not relevant; this is purely an internal policy gate.
- App is on the latest version; issue is not caused by an outdated install.

## Root cause

The security policy checks for an origin label on the agent turn before permitting `cron_add`. The label is absent when scheduling is attempted from a standard chat/agent turn, causing the gate to refuse the call. This is either:

1. A bug where the turn context is not being stamped with an origin label before agent tool calls are dispatched, or
2. A policy misconfiguration where `cron_add` requires a stricter origin assertion than other tools, without a clear path for a normal agent turn to satisfy it.

## Reproduction

1. Open a chat in a fully up-to-date install.
2. Ask the agent to schedule a recurring task (e.g. "remind me every day at 9am to…").
3. Agent invokes `cron_add`.
4. Tool call is rejected: `agent turn has no origin label`.

## Solution

- Identify where origin labels are stamped on agent turns and ensure `cron_add` invocations from chat/agent turns receive a valid label.
- If the policy intentionally requires a stricter origin for scheduling, surface a clear, actionable error to the user — not a raw policy rejection with no guidance.
- Add a test that covers `cron_add` being called from a standard agent chat turn to prevent silent regressions.

## Acceptance criteria

- [ ] Asking the agent to schedule a task from a chat turn succeeds — `cron_add` is not blocked by the origin label policy gate.
- [ ] If origin label enforcement is intentional for scheduling, the error shown to the user explains what is required rather than exposing a raw internal policy message.
- [ ] Scheduling works consistently across app restarts without requiring a daemon restart to clear stale state.

YellowSnnowmann:
## Summary

- Fix agent-scheduled `cron_add`/`cron_update` from a chat turn silently failing: the approval park is now routable, so the approval card surfaces and the user can approve instead of the call TTL-denying.
- Approval gate now falls back to the `WebChat` turn origin's `thread_id`/`client_id` for approval routing when the `APPROVAL_CHAT_CONTEXT` task-local is absent.

## Problem

`cron_add` (and `cron_update`) are `external_effect` tools, so the approval gate is consulted. When the orchestrator routes a "remind me…" request to `scheduler_agent`, that runs as an **async delegated sub-agent** (`spawn_async_subagent`). The spawn propagates the `WebChat` turn origin (so the gate no longer sees `Unknown` — that half is handled by the turn-origin-propagation work in #5527) but does **not** propagate the `APPROVAL_CHAT_CONTEXT` task-local.

Result: the gate takes the `WebChat` park arm but reads `chat_thread_id`/`chat_client_id` as `None`. `ApprovalRequested` is published with `thread_id: None`, the web-channel `ApprovalSurfaceSubscriber` explicitly does not surface it ("thread/client absent — NOT surfacing"), the park waits the full TTL, and the call is denied. The scheduled reminder is never created, with no card ever shown to the user.

## Solution

The `WebChat` origin already carries the same `thread_id`/`client_id` that `APPROVAL_CHAT_CONTEXT` does — both are built from the same values in `web_chat::start_chat`. So `ApprovalGate::intercept` now derives the chat routing from the origin when the task-local is absent. Scope is deliberately narrow: the fallback fires only for `AgentTurnOrigin::WebChat`; inline chat turns (task-local present) and every other origin are byte-for-byte unchanged.

Design note: fixing this in the gate — rather than propagating a second task-local at the spawn site — is minimal and covers every path where a `WebChat` turn reaches the gate without the approval context, not just async sub-agents.

## Relationship to #5499 / #5527

The linked issue #5499 (`agent turn has no origin label`) was the pre-fix shape of this failure: the origin arrived as `Unknown` before propagation existed. **#5527 landed the origin-propagation half and closed #5499** — it scopes the turn origin for the local-agent spawn and reworded the `Unknown`-origin denial into an actionable message. This PR closes the **residual silent-park gap #5527 did not cover**: even with a valid `WebChat` origin, an async-delegated sub-agent's approval never surfaces because the approval context task-local is absent. Rebased on current `main`, so the denial-message rewording that was in an earlier revision of this PR is dropped (already present from #5527) — the diff is now purely the routing fallback and its regression test.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — new `webchat_origin_routes_park_when_approval_chat_context_absent` proves the park is routable via the origin when `APPROVAL_CHAT_CONTEXT` is absent (fails on old code)
- [x] **Diff coverage ≥ 80%** — the only changed lines are the routing fallback (exercised by the new test) and its debug-log branch. Ran `openhuman::security::approval` + `openhuman::agent::turn_origin` (139 pass); `cargo fmt --check` and `cargo clippy --lib` clean.
- [x] Coverage matrix updated — `N/A: behaviour-only change (no feature row added/removed/renamed)`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: behaviour-only change`
- [x] Manual smoke checklist updated — `N/A: does not touch a release-cut surface`
- [x] Linked issue referenced in the `## Related` section

## Impact

- Desktop/CLI (Rust core only — `src/openhuman/security/approval/gate.rs`). No frontend, no schema, no migration.
- Security: strictly narrows a fail-silent gap. Previously a `WebChat`-origin external-effect park with no approval context silently TTL-denied; it now surfaces its approval card in the originating thread. No origin other than `WebChat` changes behaviour, and the gate still parks (never auto-allows) — the user must still approve.
- Pre-existing limitation (unchanged): `thread_to_request` is one-park-per-thread; a parent turn and a concurrent async child parking on the same thread can overwrite the chat-reply route (card-click by `request_id` still resolves either).

## Related

- Follows up: #5499 (closed by #5527) — closes the residual silent-park gap that #5527 did not cover.
- Related: #5508 (the multi-core / intermittent variant of the same cron authorization problem).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved approval routing for WebChat conversations when chat context is unavailable.
  * Approvals now remain correctly associated with the originating conversation across asynchronous operations.
  * Added coverage to verify approval mapping and resolution using propagated conversation details.


M3gA-Mind:
## Review — comment only (no approval)

Reviewed `324af1e4f` against `main` `47f4a0ec1`. **The fix is correct and I could not falsify the security-sensitive claims.** Verification notes below, then one undisclosed gap.

### Verified

**No cross-routing / authorization hole.** `AgentTurnOrigin::WebChat` is constructed at exactly three sites — `web_chat/ops.rs:681` (primary turn), `web_chat/ops.rs:909` (`spawn_parallel_turn`), `flows/ops.rs:6931` (copilot `flows_build`) — and at every one of them the origin and `APPROVAL_CHAT_CONTEXT` are built from the *same two values in the same statement pair*. So the premise is verified at each site rather than assumed. The only way a `WebChat` origin reaches the gate without the context is `turn_origin::propagate`, which is inherit-only by construction (`turn_origin.rs:246-252`): `Some(origin)` re-scopes that exact origin, `None` scopes nothing. It cannot manufacture or substitute a label, and both call sites carry the *parent's own* origin. The fallback can therefore only route to the thread/client that initiated the turn chain. Also checked that the gate never uses *absence* of the chat context as a safety signal — `chat_ctx` feeds only `chat_thread_id`/`chat_client_id`, used exclusively for routing and cleanup (`gate.rs:947, 962, 983-984, 992-993, 1080, 1216`) — so supplying them from the origin cannot weaken a decision.

**"byte-for-byte unchanged" holds.** Effective diff is one file, +94/−2. Non-`WebChat` ⇒ `origin_chat_route` is `None` ⇒ `or_else` yields `None`, identical. `WebChat` *with* the task-local ⇒ `map` returns `Some` ⇒ `or_else` never evaluated, and the task-local correctly still wins. New `debug!` is guarded. `match &origin` is a borrow; nothing reordered. (The noisy two-dot diff is just the branch having merged an older `main`; merge-base is an ancestor of `main` and `git diff-tree -m 324af1e4f` shows `src` only — no hidden revert.)

**No degenerate-origin case.** `WebChat`'s `thread_id`/`client_id` are `String`, not `Option`, and `start_chat` trims and rejects both when empty (`ops.rs:402-411`), so the fallback either supplies a real pair or does not fire.

**The test is non-vacuous.** With `.or_else(...)` reverted, `chat_thread_id` is `None`, the insert at `gate.rs:947` never runs, `pending_for_thread("thread-async")` stays `None`, and `assert!(tries < 50)` trips. Timing is sound (2s `test_gate` TTL vs a 500ms poll budget; `Config::default()` neither allowlists `cron_add` nor sets `auto_approve_all`, so it genuinely parks).

**No merge-order dependency.** #5465 (origin propagation, merged 08-09) and #5527 (merged today) are both on `main`, and `turn_origin::propagate` is present at `spawn_async_subagent.rs:546` on `main` and on this branch's merge-base. The body's account of the #5499/#5527 relationship is accurate.

### F1 — MEDIUM, not currently disclosed: the surfaced card is cleared by the next turn on the thread, and while invisible a bare "ok" approves the background effect

The body discloses the `thread_to_request` one-slot limitation. The frontend has the *same* single-slot problem, and unlike the Rust side it has no ownership check:

- `chatRuntimeSlice.ts:1722` stores one card per thread; `:1724` `clearPendingApprovalForThread` is an unconditional `delete` by `threadId` — no `request_id` check, unlike `clear_thread_route_if_owned` (`gate.rs:1434`).
- `ChatRuntimeProvider.tsx:1133` (turn done), `:1285` (turn error), `:1372` (disconnect) all dispatch that clear.

But `spawn_async_subagent` returns immediately (*"continue without blocking"*, `spawn_async_subagent.rs:864`), so the park now deliberately **outlives** the turn whose thread it is keyed to.

**Scenario A — the card vanishes, the park TTL-denies anyway.** User asks for the reminder; parent turn ends; child parks and the card surfaces (the fix working). User then types anything else in that thread — that turn's `chat_done` deletes the child's card, though its park is still open and unrelated to that turn. No surface left → TTL-deny. #5499's symptom, one turn later.

**Scenario B — conversational "ok" captured as an approval.** While that now-invisible park is open, `thread_to_request` still points at it, so `web_chat/ops.rs:510` intercepts the next message. `parse_approval_reply` (`gate.rs:159`) accepts `yes | y | ok | okay | approve | approved | allow`. So a user replying `ok` to something else **(a)** silently approves a background external effect they weren't answering and **(b)** never starts a turn — `start_chat` returns early, so their message produces no response. Before this PR a background park never entered `thread_to_request`, so it could not capture a bare `ok`. This path is new.

**Cheapest mitigation that keeps the fix:** for the origin-fallback park only, publish `ApprovalRequested` with the origin's thread/client (card renders) but skip the `thread_to_request` insert. The card decides by `request_id` (`ApprovalRequestCard.tsx:46-48`) so it still resolves cleanly; only the ambiguous free-text route is withheld from background parks. Alternatively make `clearPendingApprovalForThread` `request_id`-scoped, which also fixes scenario A for foreground parks. At minimum, fold both scenarios into the "Pre-existing limitation" note so the remaining exposure is on the record.

### F2 — process note: the new test is not covered by a green lane yet

`Rust Core Coverage (cargo-llvm-cov)` is still **pending**, and `Rust Quality (fmt, clippy)` does not build test targets for this crate lane — so no green check has compiled or run `mod tests`. I take the author's local run (139 pass) at face value; just worth not calling the regression test CI-verified until that lane lands.

### Nits

- **N1.** The gate-local fallback is defensible, but the root defect is a task-local that does not cross `tokio::spawn`, and the gate is not its only consumer — `web3/wallet/execution.rs:286-296` (whose own comment warns the helper "will silently start…" without the scope), `security/egress/emit.rs:73`, and `tools/impl/system/install_tool.rs:122` all still break across the same boundary. An `approval::propagate_chat_context` beside `turn_origin::propagate`/`turn_workspace::propagate` at the four #5465 spawn sites would cover them all — though `install_tool` reads *absence* as "background → refuse" (`:312`), so that one needs its own thought. Follow-up issue, not a change request here.
- **N2.** The frontend/backend asymmetry in N1's clear-vs-`clear_thread_route_if_owned` is worth fixing on its own merits; #4774 already established the ownership-check pattern on the Rust side.

Sound, narrow, well-documented change that does what it claims and does not weaken the gate. F1 is the only thing I'd want resolved or explicitly recorded before treating #5499's class of failure as closed. Approve/merge is the maintainer's call — deliberately not approving.

M3gA-Mind (approved):

Issue #145, opened 2026-03-31
[Feature] Subconscious loop: local-model state review (replaces heartbeat)


senamakel:
## Summary

Replace the current **heartbeat** subsystem with a **subconscious loop**: a periodic, **local-model** pass that ingests a bounded snapshot of **state, memory, tools, environment**, and related signals, **decides whether action is warranted**, and only then may **escalate** to a higher-end model or other execution paths. **Rename the feature from “heartbeat” to “subconscious loop”** across code, config, RPC, and user-facing copy where applicable.

## Problem

Today `openhuman::heartbeat` is centered on **`HEARTBEAT.md`** parsing and a fixed interval tick (`HeartbeatEngine` in `src/openhuman/heartbeat/engine.rs`). It does not **reason** over live workspace/agent context (memory, tool registry, env, channel state, etc.) or **choose** whether work is needed.

We want proactive, **context-aware** background processing that:

- Fits a practical **30–50k token context window** per pass (budget for structured summaries + diffs, not unbounded dumps).
- Uses the **local model** for the lightweight “should we care / what changed?” step to keep cost and latency predictable.
- Can **trigger** a **stronger model** or downstream jobs only when the subconscious pass says it is justified.

**Who it hurts:** Users who expect the assistant to react to evolving context without manual checklist files; operators who need predictable resource use (local first, premium models only when needed).

**Constraints:** Desktop + core sidecar architecture; privacy-sensitive local inference option; no unbounded context growth per tick.

## Solution (optional)

- **Scope:** Primarily **Rust core** (`src/openhuman/…`), replacing or refactoring `src/openhuman/heartbeat/`; config (`HeartbeatConfig` → subconscious naming); RPC/schemas as needed; app copy only if surfaced in settings or logs.

- **Subconscious tick (design sketch):**
  1. **Assemble** a capped “situation report” (30–50k budget): deltas or summaries for memory, tools, env, and other agreed signals; explicit “no change” / hash or cursor where possible.
  2. **Local model** evaluates: changed? important? any recommended action? output structured decision (e.g. no-op / enqueue task / escalate).
  3. If **escalate** or **act**: invoke **higher-end model** or enqueue work for the main agent loop with a tight handoff payload (avoid re-sending full 50k unless required).

- **Rename:** `heartbeat` → **subconscious** (module names, config keys with migration path, logs, docs). Deprecate `HEARTBEAT.md`-only task lists or map them into the new pipeline if still desired.

[...]

PR #5531, opened 2026-08-13
fix(memory/sources): size ingest RPC budget for multi-window sessions

### LINKED ISSUE #5509: Bug: bulk session ingest timeout — rpc.rs 30s/session budget ignores multi-window sessions, DIGEST_MAX_OUTPUT_TOKENS=4096 silently drops observations
## Summary

Bulk coding-session ingest times out on the core side for users with large Codex sessions. The RPC timeout formula in `rpc.rs` assumes one LLM call per session, but `digest_session` splits each session into windows and issues a separate LLM call per window — so the 30s/session budget is structurally undersized for any multi-window session. Compounding the issue, `DIGEST_MAX_OUTPUT_TOKENS = 4096` truncates digest responses mid-JSON on dense sessions, wasting 30–45s per call and silently dropping observations.

## Observed failure

```
Run started 08:31:52, died 08:41:22
```

Core-side timeout ceiling: `rpc.rs:54` → `120s + 15 × 30s = 570s`

### Bug 1 — Timeout formula assumes 1 LLM call per session (wrong)

`rpc.rs` comment: *"each session drives at most one LLM call"* — this is false for multi-window sessions.

`digest_session` (via `distill.rs`) splits each session into windows. Every window triggers a separate LLM call. Large Codex sessions produce many windows; log timestamps show each call runs 20–45s. A single session with several windows can consume several minutes — far exceeding the 30s/session budget. With 15 sessions in a batch, the 570s ceiling is hit before most sessions complete.

The 30s/session budget in `rpc.rs` must be recalculated to account for multi-window sessions. The correct formula needs to factor in the number of windows per session (or use a per-window budget rather than a per-session one).

### Bug 2 — `DIGEST_MAX_OUTPUT_TOKENS = 4096` truncates responses mid-JSON

Logs show repeated:
```
[persona] digest parse failed ... EOF while parsing a list at line 1 column ~15-19k
```

The model hits the 4096-token output cap mid-JSON on dense sessions. Each truncated call:
- Wastes 30–45s of the timeout budget
- Produces an unparseable response
- **Silently drops that window's observations** — the soft-fail path commits the cursor anyway, so the window is marked done but no data is captured

This means bulk ingest on dense sessions is both slow (truncated calls burn the clock) and lossy (observations are silently discarded without the user knowing).

## Practical workaround (today)

Sessions that complete before the timeout keep their cursor. Re-running ingest picks up where it left off. Repeated runs eventually drain the backlog — slow and lossy, but it converges. Users should click ingest again until the queue is empty.

## Fix

1. **`rpc.rs` timeout budget** — replace the flat `30s × session_count` formula with one that accounts for windowed sessions. Options: per-window budget, dynamic timeout based on estimated window count, or removing the ceiling and using per-call timeouts with a total wall-clock cap.

2. **`DIGEST_MAX_OUTPUT_TOKENS`** — raise the cap above 4096 to accommodate dense session digests without mid-JSON truncation. Alternatively, implement streaming JSON parsing so a truncated response captures partial observations rather than dropping the entire window silently.

3. **Soft-fail cursor commit** — a window whose digest was truncated/unparseable should not have its cursor committed as complete. Either retry the window or mark it for re-ingestion rather than silently discarding its observations.

## Acceptance criteria

- [ ] Bulk ingest of 15 large Codex sessions completes without hitting the RPC timeout ceiling.
- [ ] No `EOF while parsing a list` digest parse failures in logs for dense sessions.
- [ ] A window whose digest response is truncated or unparseable does not commit its cursor — it is retried or flagged for re-ingestion.
- [ ] `rpc.rs` comment accurately reflects the actual LLM call count per session.
- [ ] Users are notified when a window is skipped or retried rather than failing silently.

## Related

- openhuman #5352 — coding session ingest caps at 15 sessions/run (`CODING_SESSION_BATCH_MAX=15`) with no auto-continue

YellowSnnowmann:
## Summary

- Resize the `ingest_coding_sessions` RPC wall-clock ceiling so a legitimate multi-session backfill is not killed mid-flight.
- The old ceiling (`120 + N×30`) assumed **one LLM call per session**; TinyCortex's persona pipeline splits an oversized session into windows and issues **one LLM call per window**, so a multi-window session drives several sequential calls and blows the budget.
- Extract the computation into a pure, unit-tested `ingest_budget(max_sessions)` and correct the false comment.

## Problem

Reported in #5509 (Bug 1). `src/openhuman/memory/sources/rpc.rs` computed the RPC timeout as `120 + min(max_sessions, 1000) × 30` on the premise, stated in its own comment, that *"each session drives at most one LLM call."* That premise is false: `digest_session` splits each session into `WINDOW_CHARS`-sized windows and fires one LLM call per window (each observed at 20–45 s). A dense backfill of 15 sessions therefore hit the exact `120 + 15×30 = 570 s` ceiling and was killed mid-flight, dropping the remaining sessions.

## Solution

- Extract `ingest_budget(max_sessions)` — a pure function sized for *multiple* windows per session (`PER_SESSION_SECS = 120`, ~4 sequential per-window calls) instead of one. A healthy backfill now finishes well inside the ceiling while a genuinely wedged run still terminates.
- Preserve the untrusted-input cap (`MAX_SESSIONS_FOR_BUDGET = 1000`) so an inflated `max_sessions` cannot turn the ceiling into an effectively-infinite wait.
- Correct the false "one LLM call per session" comment to describe the real per-window call pattern.
- `ingest_budget` takes `usize` to match the request field type (`CodingSessionIngestRequest::max_sessions`).

This is the **RPC-timeout half** of #5509. The **digest-truncation half** (`DIGEST_MAX_OUTPUT_TOKENS` raise + non-committable truncated windows) lives in TinyCortex — tinyhumansai/tinycortex#145. That half is what makes the extra per-window calls succeed; the vendored `vendor/tinymemory` → `tinycortex` submodule pointer is bumped in a follow-up PR once #145 merges. #5509 should be closed by that follow-up (both halves present), not by this PR alone.

## Submission Checklist

- [x] Tests added or updated (happy path + failure / edge case) — `budget_tests`: the 15-session regression (proves the new ceiling exceeds the old 570 s) and the untrusted-`max_sessions` cap.
- [x] **Diff coverage ≥ 80%** — the two added tests exercise every line of the new `ingest_budget` function (the only changed executable lines).
- [x] Coverage matrix updated — `N/A: behaviour-only change` (timeout-budget arithmetic; no new feature row).
- [x] All affected feature IDs listed under `## Related` — `N/A: no matrix feature touched`.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: internal RPC timeout budget only`.
- [x] Linked issue closed via `Closes #NNN` — intentionally **not** closing here; see `## Related` (needs the digest half + submodule bump).

## Impact

- Desktop/CLI coding-session ingest only. No API surface change. Purely widens an internal wall-clock ceiling; a wedged run is still bounded and terminated.
- No performance/security/migration implications: the budget is a kill-ceiling, not a latency target.

## Related

- Part of: #5509 (RPC-timeout half)
- Digest half: tinyhumansai/tinycortex#145
- Follow-up PR(s)/TODOs: bump `vendor/tinymemory`/`tinycortex` submodule pointer once #145 merges → that PR carries `Closes #5509`.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/ingest-timeout-budget-5509

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when ingesting coding sessions and processing multiple LLM windows.
  * Increased per-session processing time while using smaller batches to reduce timeout failures.
  * Added safeguards that cap processing limits for unusually large requests.
  * Improved drain processing by using repeated, bounded batches within the available timeout window.

### DIFF
--- a/app/src/services/memorySourcesService.test.ts
+++ b/app/src/services/memorySourcesService.test.ts
@@ -230,11 +230,19 @@ describe('memorySourcesService', () => {
 
     const result = await ingestCodingSessions(false, 25);
 
+    // Clamped to the batch max (5), and the per-call timeout stays under the RPC
+    // client's hard 600s ceiling: 120s + 5*90s + 15s = 585s.
     expect(mockedCall).toHaveBeenCalledWith({
       method: 'openhuman.memory_sources_ingest_coding_sessions',
-      params: { backfill: false, max_sessions: 15 },
+      params: { backfill: false, max_sessions: 5 },
       timeoutMs: 585_000,
     });
+    // Guard the cross-wire contract: the per-call timeout must stay under
+    // coreRpcClient's PER_CALL_TIMEOUT_MAX_MS (600s), or the override is clamped
+    // and silently unreachable (the #5509 gap). Raising BATCH_MAX / per-session
+    // past this must fail here.
+    const CORE_RPC_PER_CALL_TIMEOUT_MAX_MS = 10 * 60 * 1_000;
+    expect(585_000).toBeLessThanOrEqual(CORE_RPC_PER_CALL_TIMEOUT_MAX_MS);
     expect(result.sessions_processed).toBe(2);
   });
 
@@ -269,7 +277,7 @@ describe('memorySourcesService', () => {
     expect(mockedCall).toHaveBeenCalledTimes(3);
     // Every pass stays bounded to the timeout-safe per-call maximum.
     expect(mockedCall).toHaveBeenLastCalledWith(
-      expect.objectContaining({ params: { backfill: false, max_sessions: 15 } })
+      expect.objectContaining({ params: { backfill: false, max_sessions: 5 } })
     );
     expect(result.passes).toBe(3);
     expect(result.sessionsProcessed).toBe(40); // 15 + 15 + 10
--- a/app/src/services/memorySourcesService.ts
+++ b/app/src/services/memorySourcesService.ts
@@ -222,21 +222,25 @@ export interface CodingSessionIngestResult {
   pack_path?: string | null;
 }
 
-// A single ingest RPC is bounded so it stays under the core RPC client's
-// ten-minute ceiling: 120s + 15 * 30s + 15s ≈ 585s. This is a per-call bound,
-// not a per-run one — large histories drain across repeated bounded passes via
-// `drainCodingSessions`, driven by the response `budget_hit` flag. Raising this
-// to the backend's 1,000-session max would blow the RPC timeout on the first
-// call, so the cap stays and the loop does the scaling.
-const CODING_SESSION_BATCH_MAX = 15;
+// A single ingest RPC is bounded so it fits under the core RPC client's hard
+// per-call ceiling (`PER_CALL_TIMEOUT_MAX_MS` = 600s in `coreRpcClient.ts`, which
+// clamps every override — a larger request timeout is silently unreachable).
+// The bound mirrors the core's `ingest_budget` (`memory/sources/rpc.rs`): the
+// server sizes the same call at `120s + N*90s` and the client adds a 15s grace
+// so the server's structured timeout fires first. Each multi-window session
+// drives several sequential LLM calls (~90s/session), so the batch is kept small
+// (`120s + 5*90s + 15s = 585s < 600s`) and large histories drain across repeated
+// bounded passes via `drainCodingSessions`, driven by the response `budget_hit`
+// flag — never by widening a single call past the reachable ceiling.
+const CODING_SESSION_BATCH_MAX = 5;
 const CODING_SESSION_BASE_TIMEOUT_MS = 120_000;
-const CODING_SESSION_PER_SESSION_TIMEOUT_MS = 30_000;
+const CODING_SESSION_PER_SESSION_TIMEOUT_MS = 90_000;
 const CODING_SESSION_RPC_GRACE_MS = 15_000;
 // Hard safety cap on drain passes so a stuck backlog can never spin forever.
-// Sized well above the largest realistic history: at 15 sessions/pass this
-// covers ~30k sessions in a single run, so the target ~7,800-file case drains
-// fully rather than exiting capped. The `moreRemaining` flag still lets the UI
-// report an honest "paused" state if the cap is ever reached.
+// Sized well above the largest realistic history: at 5 sessions/pass this covers
+// ~10k sessions in a single run, so the target ~7,800-file case drains fully
+// rather than exiting capped. The `moreRemaining` flag still lets the UI report
+// an honest "paused" state if the cap is ever reached.
 const CODING_SESSION_MAX_DRAIN_PASSES = 2000;
 
 export async function getCodingSessionStatus(): Promise<CodingSessionSourceStatus[]> {
--- a/src/openhuman/memory/sources/rpc.rs
+++ b/src/openhuman/memory/sources/rpc.rs
@@ -36,39 +36,53 @@ pub async fn coding_session_status_rpc() -> Result<RpcOutcome<CodingSessionStatu
 }
 
 /// Wall-clock ceiling for one `ingest_coding_sessions` RPC, sized to the number
-/// of sessions the caller asked to backfill.
+/// of sessions the caller asked to backfill and hard-capped at the ceiling the
+/// frontend can actually wait for.
 ///
-/// The previous formula (`120 + N*30`) assumed **one LLM call per session**.
+/// The original formula (`120 + N*30`) assumed **one LLM call per session**.
 /// That premise is false: TinyCortex's persona pipeline splits an oversized
 /// session into windows (`WINDOW_CHARS`-sized chunks of evidence) and issues one
 /// LLM call *per window*, so a multi-window session drives several sequential
 /// calls. A dense backfill therefore blew the old budget — 15 sessions hit the
 /// exact 570 s ceiling (`120 + 15*30`) and were killed mid-flight.
 ///
-/// So the per-session allowance is sized for *multiple* windows, not one call:
-/// `PER_SESSION_SECS` budgets for up to roughly four sequential LLM calls at the
-/// provider's own per-call timeout, which comfortably covers the windowing
-/// observed in practice while still bounding the RPC. `max_sessions` is
-/// untrusted (it comes straight off the request), so the multiplier is capped at
-/// `MAX_SESSIONS_FOR_BUDGET` before scaling — an inflated value cannot turn the
-/// ceiling into an effectively-infinite wait.
+/// The per-session allowance is therefore sized for *multiple* windows, not one
+/// call: `PER_SESSION_SECS` budgets ~3 sequential per-window LLM calls at the
+/// windows' observed 20–45 s span (#5509). It is a deliberate flat estimate, not
+/// a per-session window count.
 ///
-/// This is a *ceiling to catch a wedged run*, not a latency target: a healthy
-/// backfill finishes far inside it, and the pipeline's own run budget bounds
-/// real cost. Over-provisioning here only delays the kill of a genuine hang.
+/// The result is hard-capped at `HARD_CAP_SECS` for two reasons that are really
+/// one. First, this is the true reachable ceiling: the frontend RPC client
+/// clamps every per-call timeout to `PER_CALL_TIMEOUT_MAX_MS = 600 s`
+/// (`app/src/services/coreRpcClient.ts`), so a server budget above that can never
+/// be observed — the client aborts first. Second, that cap also bounds the
+/// blocking-pool worker this budget guards: `max_sessions` is untrusted (an
+/// advertised programmatic RPC, `platform/about_app/catalog_data.rs`), and
+/// without the cap a caller passing 1000 would pin a thread for ~33 h. Capping
+/// the resulting `Duration` — not the multiplier — makes both true at once.
+///
+/// Because a single pass is bounded, large histories drain across repeated passes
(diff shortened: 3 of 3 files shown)


M3gA-Mind:
## Review — comment only (no approval)

Reviewed `dc9fb786b` against `origin/main` `baaf89490`. The diagnosis is right and the extraction is clean, but I think **the fix is on the wrong side of the wire**: for #5509's exact repro it buys about 15 seconds, not 22 minutes.

### F1 — the client aborts at 585 s, 15 s after the old server ceiling

This raises the 15-session server ceiling 570 s → 1920 s. The only shipped caller imposes its own bound on the same call — `app/src/services/memorySourcesService.ts:231-234, 259-262`:

```ts
CODING_SESSION_BATCH_MAX = 15;
CODING_SESSION_BASE_TIMEOUT_MS = 120_000;
CODING_SESSION_PER_SESSION_TIMEOUT_MS = 30_000;
CODING_SESSION_RPC_GRACE_MS = 15_000;
// timeoutMs = 120s + 15×30s + 15s = 585s
```

That is a mirror of the *old* server formula plus a grace, and the comment right above it (`:225-231`) says so: *"A single ingest RPC is bounded so it stays under the core RPC client's ten-minute ceiling: 120s + 15 \* 30s + 15s ≈ 585s."* It's a documented two-sided contract; this PR changes one side and leaves that comment describing a formula that no longer exists.

#5509 reports `Run started 08:31:52, died 08:41:22` — exactly 570 s, so the server ceiling was indeed the binding constraint. After this change:

| | before | after |
|---|---|---|
| server ceiling (15 sessions) | 570 s | 1920 s |
| client ceiling (15 sessions) | 585 s | 585 s |
| **what fires** | **server @ 570 s** | **client @ 585 s** |

Net gain ≈ **15 s**; the other 1335 s is never observed. And it can't be recovered by raising the client constants either — `coreRpcClient.ts:47` sets `PER_CALL_TIMEOUT_MAX_MS = 10 * 60 * 1_000` and `resolvePerCallTimeoutMs` (`:49-57`) clamps *every* per-call override to 600 s. A server budget above ~600 s is unreachable by construction. (Separately and pre-existing: `coreRpcClient.ts:656-659` drops `timeoutMs` entirely on the cloud/tunnel transports, which use a flat 30 s — so no client path anywhere reaches the new ceiling.)

So #5509's acceptance criterion — *"bulk ingest of 15 large Codex sessions completes without hitting the RPC timeout ceiling"* — still isn't met. What would close it: raise the client constants **and** cut `CODING_SESSION_BATCH_MAX` so a pass fits under 600 s (the `drainCodingSessions` loop at `:322` already exists to scale by passes, with 2000 passes of headroom); or land tinycortex#145 so 30 s/session stops being structurally undersized; or derive one budget from the other instead of hand-syncing two copies of the same arithmetic — which is how they drifted apart.

**Counter-argument, for the record:** the client abort is an `AbortController` on `fetch` and does not cancel the core-side future, and #5509 confirms cursors commit incrementally. So the server does keep working past 585 s and the *next* run finds more done. That's a genuine benefit — but it's accidental, invisible to the user (error toast + aborted drain, `CodingSessionsCard.tsx:106-109`), and it pins a blocking-pool thread the whole time (`rpc.rs:91-98` wraps `runtime.block_on` inside `spawn_blocking`). If that's the intended value, worth stating in the PR.

### F2 — the `MAX_SESSIONS_FOR_BUDGET` cap no longer does what its doc claims

The doc says an inflated `max_sessions` *"cannot turn the ceiling into an effectively-infinite wait"*, but at the cap the ceiling is now `120 + 1000×120 = 120_120 s` = **33.4 hours** (was 8.4 h) — on a blocking-pool worker. The UI clamps to ≤ 15, but the method is an advertised programmatic entry point (`platform/about_app/catalog_data.rs:561`), so a direct RPC caller can pass 1000. Capping the resulting `Duration` (e.g. `.min(Duration::from_secs(3_600))`) rather than the multiplier would make the stated guarantee true while keeping the per-session scaling.

### F3 — the stated derivation of `PER_SESSION_SECS = 120` doesn't match the PR's own data

*"budgets for up to roughly four sequential LLM calls at the provider's own per-call timeout"* implies 30 s/call, but the PR body and #5509 both report windows *"each observed at 20–45 s"* — at the observed upper end that's ~2.7 calls, not four, and no named provider-timeout constant is referenced so the figure isn't checkable. The budget is aggregate across the batch so it still works out in practice; this is accuracy-of-docs only. Given the PR's headline is "the old comment stated a false premise", it seems worth not replacing it with a second unverifiable one. Also: #5509 asked for a budget that factors in windows-per-session; this is a flat 4× — defensible for a kill-ceiling, but worth saying it's a deliberate simplification.

### Verified fine

Three-dot diff is one file, +72/−6; the branch is 63 commits behind `main` but merge-base `b508b3ce7` is an ancestor of `main` and there's a single commit, so the noisy two-dot view is staleness, not a hidden revert. No overflow (`sessions` capped before the `as u64` multiply). The extraction itself — pure function, named constants replacing three inline magic numbers, false "one LLM call per session" comment removed — is a straight improvement. And the `Closes` discipline is right: deliberately not auto-closing #5509, with the reason stated.

### Tests

`budget_tests` are non-vacuous in the narrow sense — reverting `PER_SESSION_SECS` to 30 makes `ingest_budget(15)` return 570 and fails both assertions. But they're change-detector tests: they assert the formula the same commit just wrote, and can't fail for any reason connected to the actual bug. Nothing pins the thing that actually broke — that the server budget must be the *looser* of the two for the same `max_sessions`, and that the client's stays under `PER_CALL_TIMEOUT_MAX_MS`. That guard is cheap and would have caught this gap.

### CI

Still in flight: `Rust Quality (fmt, clippy)`, `Rust Feature-Gate Smoke`, `Rust RSS Benchmark` pending; CodeRabbit in progress; `Rust Core Coverage` (the lane that compiles `mod budget_tests`) not listed yet and `TinyCortex Memory Tests` skipping. Nothing has run the new tests so far.

**Bottom line:** clean refactor, correct diagnosis, honest PR body — but the binding ceiling just moves from the server's 570 s to the client's 585 s. I'd pair it with the `memorySourcesService.ts` change (or a `CODING_SESSION_BATCH_MAX` reduction) before treating #5509 Bug 1 as addressed, and fix the now-stale comment at `memorySourcesService.ts:225-231` either way. Approve/merge is the maintainer's call — deliberately not approving.

YellowSnnowmann:
Verified all three findings against the code and you're right — F1 is the load-bearing one. Reworked in `69199c829`.

**F1 — reachable ceiling.** Confirmed: `memorySourcesService.ts:259-262` bounds the same call at `120 + 15×30 + 15 = 585s`, a mirror of the old server formula, and `coreRpcClient.ts:49-57` clamps every override to `PER_CALL_TIMEOUT_MAX_MS = 600s`, so a server budget above that is unreachable. Raising the server ceiling alone just moved the binding constraint from 570s to 585s. So the fix now targets the batch, not the ceiling:

- Client: `CODING_SESSION_BATCH_MAX` 15 → 5, per-session 30s → 90s. A pass is now `120 + 5×90 + 15 = 585s < 600s` — reachable — and `drainCodingSessions` (2000-pass cap ≈ 10k sessions) drains a 15-session backlog in 3 passes. Fixed the stale `:225-231` comment to describe the real model and the 600s clamp.
- Server (`ingest_budget`): `120 + N×90s` (the honest multi-window number, see F3), hard-capped at 600s.

**F2 — the cap.** Fixed by capping the resulting `Duration` at 600s rather than the multiplier, as you suggested. `max_sessions = 1000` now yields 600s, not 33.4h, so the advertised programmatic RPC can't pin a blocking-pool thread past the reachable ceiling.

**F3 — the derivation.** Dropped the "4 calls × 30s" figure; `PER_SESSION_SECS = 90` is now documented as ~3 windows at the observed 20–45s/window span, a deliberate flat estimate rather than an unverifiable per-session window count.

**On "pin the invariant that actually broke" — done, and this is the part I'd most like your eyes on.** Added `server_budget_is_reachable_and_tighter_than_the_client`, which asserts `server(5)=570 ≤ client=585 ≤ 600` — the guard that would have failed on the server-only 1920s version. The client test also now asserts `timeoutMs ≤ 600s`.

**One deliberate divergence from your framing.** You leaned toward the server being the *looser* of the two (client aborts, server grinds on invisibly past it and the next pass finds more). I kept the server *tighter* (570 < 585) on purpose: with the batch now sized to fit under 600s, I'd rather the server hit its ceiling first and return a clean structured `timeout` the UI can render — and release the blocking thread immediately — than pin the thread while the user sees only the raw fetch-abort toast at `CodingSessionsCard.tsx:106-109` you flagged. The "server keeps working past the client abort" benefit is real but, as you said, accidental and invisible; sizing the batch to fit removes the need to rely on it. Happy to flip if you still prefer server-looser — it's a one-constant change plus the invariant test.

**Unrelated CI note.** The red `Rust Core Coverage` lane is pre-existing broken-main, not from this change: `openhuman::memory::binding::tests` panic with "no reactor running" because the module-bind path eagerly `tokio::spawn`s the ingestion worker (`queue.rs:191`), violating the "binding stays synchronous" invariant those tests document (`binding_tests.rs:342`). It's byte-identical on `origin/main` and fails on the same pinned `tinymemory` pointer — this PR is just the first to scope coverage into `openhuman::memory`. Root fix belongs in vendored tinymemory (defer the worker spawn out of the synchronous bind path); not folding it in here.

M3gA-Mind (approved):

Issue #101, opened 2026-03-30
[Feature] Calmer onboarding (show when needed) + local AI download as side snackbar


senamakel:
## Summary

**Reduce onboarding nagging**: the flow should **not** take over the whole app on every launch. Surface it **when it is actually needed** (first-time, missing critical setup, or explicit user action). Separately, move **local AI model download** out of the blocking onboarding wizard into a **persistent side snackbar** (or equivalent non-modal surface) so users can keep working while downloads run.

## Problem

Today, [`OnboardingOverlay`](app/src/components/OnboardingOverlay.tsx) can present a **full-screen, high z-index** experience whenever backend/workspace state says onboarding is incomplete. In practice this feels like onboarding **“keeps popping up in our face”**—especially if dismiss state is **not durable** across sessions or if “incomplete” is too broad.

**Local AI** is currently a **dedicated onboarding step** ([`LocalAIStep`](app/src/pages/onboarding/steps/LocalAIStep.tsx)) with progress UI tied to that flow. That **blocks** the mental model of “finish account setup, then use the app while models pull in the background.”

### Onboarding: show when needed

- **Define triggers** clearly: e.g. first login + no `isOnboarded`; optional “resume checklist” for `selectHasIncompleteOnboarding` instead of a full-screen forced modal.
- **Persist “not now” / defer** in a durable way (workspace flag, Redux persist, or server-side preference—align with existing [`openhuman.workspace_onboarding_flag_*`](app/src/utils/tauriCommands.ts) and [`DEV_FORCE_ONBOARDING`](app/src/utils/config.ts) semantics).
- **Replace or soften** the always-on overlay with:
  - a **banner** or **settings badge** (“Finish setup”) for low-priority items, and
  - full-screen flow only for **first-time** or **hard requirements** (e.g. mnemonic if product requires it).
- **Entry points**: Settings / Home CTA to reopen onboarding or specific steps.

### Local AI: snackbar on the side

- After consent (or from Settings), start download via existing core RPC/Tauri commands and show a **non-blocking snackbar** (e.g. bottom-right or edge-aligned) with:
  - status (preparing / downloading / ready / error),
  - **progress** when available (reuse patterns from [`LocalModelPanel`](app/src/components/settings/panels/LocalModelPanel.tsx) / `openhumanLocalAi*` in [`tauriCommands`](app/src/utils/tauriCommands.ts)),
[...]

Issue #100, opened 2026-03-30
[Feature] Discord notifications for new GitHub releases (changelog / revision)


senamakel:
## Summary

Post a **Discord message** whenever a **new GitHub Release** is published for this repo, including **version / tag**, a **short changelog** (from release notes or generated summary), and **links** to the release and artifacts as appropriate. **Delivery failures must not fail** the release pipeline (or any critical CI).

## Problem

Release notes live on GitHub, but **Discord** (community / team channel) does not get an **automatic, consistent ping** when `Release` ships. People miss new builds, migration notes, and download links.

## Solution (optional)

Pick one or combine:

### A) **GitHub → Discord via repository webhook** (recommended baseline)

- In GitHub repo **Settings → Webhooks**, add a webhook for event **`Release`** (at least **published** / **released**).
- Target a small **Discord-friendly receiver**:
  - **Discord Incoming Webhook** URL cannot be used directly as GitHub’s webhook target without an adapter (payload shapes differ). Use a **tiny bridge** (e.g. Cloudflare Worker, Vercel/Netlify function, or self-hosted **github2discord**-style service) that:
    - Verifies **`X-Hub-Signature-256`** (GitHub webhook secret).
    - Maps `release.tag_name`, `release.name`, `release.body`, `release.html_url`, `release.prerelease`, assets, etc. into a **Discord embed** or message.
  - **On bridge errors**: log and exit 200 so GitHub does not disable the webhook after retries (or handle failures inside the bridge only).

### B) **GitHub Actions step** (alongside existing release workflow)

- After a successful **`release` / `gh release create`** in [`.github/workflows/release.yml`](.github/workflows/release.yml), run a step that POSTs to **`DISCORD_WEBHOOK_URL`** (repo secret) with embed payload built from **`github.event.release`** or step outputs.
- Mark the notify step with **`continue-on-error: true`** (and/or a dedicated job that **`needs: []`** and only runs `if: success()` of publish job) so **Discord outage never fails the release**.

### Content guidelines

- Include **tag**, **title**, **link to GitHub Release**, and **body** (truncate with link if over Discord limits).
- Optionally append **compare URL** (`previous tag…current`) if easy to compute.
- **Redact** secrets; never echo tokens in logs.

## Acceptance criteria

- [ ] **Trigger** — New **published** GitHub Release results in **one** Discord notification (dedupe pre-releases if policy says so).
[...]

PR #5534, opened 2026-08-13
refactor(agent): move artifact-offload into TinyAgents, wire tool completion

senamakel:
Phase 5 of [`docs/specs/plan-agents.md`](docs/specs/plan-agents.md) — the first family — plus the Phase 4 adapter work that [tinyagents#100](https://github.com/tinyhumansai/tinyagents/pull/100) unblocks.

Depends on tinyagents **#100** and **#101**, both merged; `vendor/tinyagents` is bumped to their merge commit (`30d6b3b`).

**Net for the relocation: −1,319 / +362.**

---

## 1. `artifact_offload` moves down ([tinyagents#101](https://github.com/tinyhumansai/tinyagents/pull/101))

The mechanics — thresholds, path resolution, pointer rendering, the symlink re-check — are now `tinyagents::harness::artifacts`. What stays here is the wiring plus the two halves that are genuinely ours:

| Stays | Why it must |
| --- | --- |
| `contract.rs` | OpenHuman prompt text naming OpenHuman tool ids. §6 of the plan lists prompt text as **not publishable** |
| `policy.rs` | `WorkspaceGuard` → `SecurityPolicy::{is_workspace_internal_path, workspace_dir}`; `SanitizingRedactor` → `sanitize_text` |

Same shape `run_queue` already converged on: the crate owns the mechanics, the host keeps the irreducible product-specific remainder. Neither module is a re-export shim.

**`new_artifact_offload` installs both policies every time.** The crate permits `None` for each and OpenHuman never wants either, so routing construction through one helper is what stops a call site producing an unguarded or unredacted writer *by omission*.

**`WorkspaceGuard` forwards rather than reimplements.** A divergent copy of the workspace rule would drift from the policy the rest of the core enforces, and the two would then disagree about the same path — a test asserts the adapter and `SecurityPolicy` agree.

The read-tool name is a **parameter** on the crate's pointer renderer, not a constant. A hard-coded `file_read` in a redistributed crate would put a tool another host does not have into its prompts.

## 2. Tool completion is reportable at last ([tinyagents#88](https://github.com/tinyhumansai/tinyagents/issues/88))

`ProgressEvent` gained `ToolCallFinished`, so `OpenHumanProgressSink` can finally emit `AgentProgress::ToolCallCompleted`. Its module header **said none could ever be emitted** — that is now wrong, so it is rewritten rather than left contradicting the code.

The closing event carries neither the tool name nor a duration, so both are captured when the call opens (`RunState::open_calls`) and recovered on close.

Three decisions worth reviewing:

- **The recorded iteration is used, not the live counter.** Model output between a call and its result advances the round, so reading it late would file the tool under the wrong iteration.
- **A close with no matching open call is dropped, not emitted.** Without the opening record there is no honest tool name or duration, and #88's own reasoning still applies: a missing row is recoverable, a fabricated one is not. Removing the record on close also makes a duplicate close inert.
- **`timed_out: false`** when classifying a failure. The coarse stream cannot distinguish a timeout from any other failure, and claiming one would put a specific, wrong cause in front of a user.

`arguments` stays `None` — the crate emits arguments on neither event, and `None` says "not captured" rather than "there were none".

## 3. `model_pin` is routed but deliberately unread ([tinyagents#89](https://github.com/tinyhumansai/tinyagents/issues/89))

The field exists now, so a definition's exact model id has somewhere to go instead of being smuggled through `role`. The resolver does **not** consume it yet, for two reasons recorded in its header:

- Nothing constructs a `ModelResolveRequest` yet, so honouring the pin today would be code with no producer and no test that could fail honestly.
- `create_chat_model_with_model_id` takes a *role*, not a model id. Honouring a pin means adding that path in `inference::provider::factory` **and** deciding what happens when the pinned id is not configured — a design question, not plumbing.

When wired, the pin is advisory: validate against configured providers and fall back to role routing with a warning; never pass it through blind.

---

## Verification

- `cargo check` clean in **both** Cargo worlds (root crate and `app/src-tauri`)
- `artifact_offload` **12 passed** · `host::` adapters **237 passed** · `progress_sink` **20 passed** (5 new) · `subagent_runner` **81 passed**
- Kernel floor **unchanged**: this branch and `main` both resolve **308/285/2** on the same machine, so the added `tokio/fs` feature costs zero packages — it enables more of tokio, not another crate. The ratchet's own log documents Linux at 307/284 with macOS skewing +1, and it fails identically on `main`; I did not touch `kernel-floor.limits`, since raising a limit to paper over inherited drift is what that file exists to prevent.

### Pre-existing failures, not from this branch

The full lib suite reports 32 failures. **None are in code this branch touches.** Verified against `main` on the same machine:

- `memory::tools::tool_memory::put` — fails **identically on `main`** (5/10) in isolation
- `budget_gate` — passes in isolation on `main` (15/15), fails only under the full suite: the order-dependence class `plan-agents.md` already documents
- the remainder are `memory::ops`, `memory`, `git_attribution`, `core::cli_capability`, `session::turn` — all untouched here

`artifact_offload`, `progress_sink` and `model_resolver` are green in every run.

## Docs

`plan-agents.md` updated: Phase 5 now carries a per-family status table (2 of 6 landed), the #88/#89 blocker note records that both are fixed upstream and what still has to be wired, and §3's disposition row is marked.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of oversized tool results by saving them as artifacts and providing readable references.
  * Added stronger workspace safeguards and credential-sensitive content redaction during artifact handling.
  * Tool completion events now report accurate status, output, duration, and failure details.

* **Bug Fixes**
  * Prevented duplicate or unmatched tool completion events from being emitted.

* **Documentation**
  * Updated migration documentation to reflect completed artifact and queue support, model handling, and remaining work.


senamakel:
**Full-suite baseline comparison — pre-existing failures confirmed.**

Both runs on the same machine, same flags (`GGML_NATIVE=OFF RUST_MIN_STACK=16777216 cargo test --lib`):

| | passed | failed |
| --- | ---: | ---: |
| `main` | 11,436 | **33** |
| this branch | 11,410 | **32** |

This branch has **one fewer** failure than `main`. The counts differ by one because these are the order-dependent tests the plan already documents, and they are not perfectly stable run to run — both runs fail in the same modules (`memory::ops`, `memory::tools::tool_memory`, `budget_gate`, `git_attribution`, `core::cli_capability`, `session::turn`).

Verified individually against `main` as well: `memory::tools::tool_memory::put` fails identically there (5/10 in isolation), and `budget_gate` passes in isolation (15/15) but fails under the full suite.

`artifact_offload`, `progress_sink` and `model_resolver` are green in every run.

senamakel:
**Final full-suite numbers** — superseding the interim comment above.

The earlier branch run predated the `progress_sink` work and the gitlink bump. This one is the PR as it stands:

| | passed | failed |
| --- | ---: | ---: |
| `main` | 11,436 | 33 |
| this PR (final) | 11,416 | **31** |

Two fewer failures than `main`. Same machine, same flags (`GGML_NATIVE=OFF RUST_MIN_STACK=16777216 cargo test --lib`), full suite both sides.

The only `agent::tinyagents::host::` failures are both `budget_gate`:

```
budget_gate::tests::dropping_the_crate_permit_releases_the_scheduler_permit
budget_gate::tests::explicit_release_returns_capacity_before_end_of_scope
```

That module passes **15/15 in isolation on `main`** and fails only under the full suite — the order-dependence class `plan-agents.md` already documents. Notably this run **included** the new `progress_sink` code, and none of its 20 tests (5 new) failed.

Nothing in `progress_sink`, `artifact_offload` or `model_resolver` failed in any run. The remaining failures are `memory::ops` (16), `memory::tools::tool_memory` (7), `memory` (3), `session::turn`, `git_attribution`, `core::cli_capability` — all untouched here, and `tool_memory::put` reproduces identically on `main` in isolation (5/10).

Issue #5399, opened 2026-08-04
Replace current voice chat with ElevenLabs Voice Agents


Al629176:
## Summary

Replace the existing voice chat implementation with ElevenLabs Voice Agents (https://elevenlabs.io/voice-agents) to deliver a significantly better real-time conversational voice experience.

## Problem

The current voice chat has known issues:
- TTS 400 errors due to invalid voice IDs (#5355)
- 3–4 second delay between text response and voice playback (#5358)
- Tab is locked while agent is speaking (#5357)

These are symptoms of a fragile custom implementation. ElevenLabs Voice Agents provides a production-ready, low-latency, real-time voice conversation layer that eliminates these issues.

## Solution

Integrate ElevenLabs Voice Agents as the voice chat backend:

1. **Evaluate ElevenLabs Voice Agents API** — review agent creation, turn-taking, interruption handling, and latency characteristics.
2. **Replace current TTS/STT pipeline** with ElevenLabs real-time voice agent session.
3. **Update voice configuration** — map existing voice settings (language, voice ID, persona) to ElevenLabs agent config.
4. **Feature-flag the switch** — roll out behind a flag so the old path can be compared or rolled back.
5. **Remove deprecated voice code** once ElevenLabs path is stable.

**Affected areas:** Voice chat, Human tab, TTS/STT pipeline, voice configuration settings

## Acceptance criteria

- [ ] Voice chat is powered by ElevenLabs Voice Agents end-to-end.
- [ ] No 400 errors on voice initiation.
- [ ] Voice response starts within ~1 second (no 3–4s delay).
- [ ] Tab switching works while agent is speaking.
- [ ] Interruption / barge-in handling works correctly.
- [ ] Existing voice settings (language, persona) are preserved via ElevenLabs agent config.
- [ ] Old voice implementation is cleanly removed after validation.

## Related

- openhuman #5355 (TTS 400 invalid voice ID)
- openhuman #5358 (voice/text sync delay)
- openhuman #5357 (tab lock during TTS)

PR #5550, opened 2026-08-14
fix(voice): stop the directive turning an announcement into the answer

### LINKED ISSUE #5552: Voice agent announces the tool call instead of running it
## Summary

Asking the realtime voice agent to summarise emails returns no summary. The final answer delivered to chat is the agent announcing that it is about to fetch the inbox.

## Problem

**What happened.** Asked the voice agent "summarize my emails". The final response delivered to the chat thread was:

> "Let me pull up your inbox now — if it takes a moment I'll drop the summary in your chat."

No inbox summary ever arrived, in chat or spoken.

**What was expected.** The inbox summary.

**Why it happens.** That sentence is the `VOICE_DIRECTIVE`'s own example, echoed almost verbatim:

> "Sure, let me pull up your inbox — if it takes a moment I'll drop the summary in your chat."

The directive instructs the model to speak before acting:

> "Before you use a tool or delegate … first say one short spoken sentence telling the user what you are doing … then proceed."

But a reply carrying **only text and no tool call is the end of a turn**. A model that follows the instruction emits the sentence, its turn ends there, and the harness delivers that sentence as the final answer. The Gmail call never runs. The more obediently the model complies, the more reliably the turn fails.

**Impact.** Any voice request that needs a tool — email, calendar, files, web — can silently return a promise instead of an answer. The failure is invisible: the user is told the result is coming, so nothing reads as broken until they notice it never arrived.

**Steps to reproduce.**
1. Start a realtime voice chat on the Human tab.
2. Ask something that requires a tool ("summarize my emails").
3. Observe the chat thread: the delivered answer is the announcement, not the summary.

**Environment.** Desktop dev build, macOS, realtime voice path (`src/openhuman/voice/realtime_harness.rs`).

## Solution

Remove the preface clause and instruct the opposite: call the tool and answer from the result.

The clause existed for a latency reason that no longer holds. It was there to get audio to the caller early, back when the relay's spoken filler was inaudible until the turn closed — the provider segments on sentence boundaries, and the filler was ellipsis-terminated (tinyhumansai/backend#1275). The relay now speaks a terminated sentence ~700ms in, and does so without depending on the model choosing to speak first.

The replacement states the reason as well as the rule, since a bare prohibition invites the model to hedge back into the same behaviour.

## Acceptance criteria

- [ ] **Repro gone** — a tool-backed voice request ("summarize my emails") returns the actual result, not an announcement of intent.
- [ ] **Regression safety** — a test pins the directive's contract: it must forbid announcing, and must not carry the phrasing that caused this.
- [ ] **No silent regression path** — the acknowledgement the caller hears no longer depends on the model choosing to speak, so removing the preface does not reintroduce dead air.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by [`.github/workflows/ci-lite.yml`](../../.github/workflows/ci-lite.yml)).

## Related

- #5399 — realtime ElevenLabs voice agent (parent)
- tinyhumansai/backend#1275 — the relay filler that makes the preface unnecessary
- #5549 — the deadline line, same spoken-copy surface

YellowSnnowmann:
## Summary

- Removes the clause telling the voice model to speak a sentence *before* using a tool. That instruction made the announcement the entire answer.
- Replaces it with the opposite instruction: call the tool, answer from the result, do not narrate the intention.
- Clears three comments that still described the preface as something that happens.

## Problem

Asking for a summary produced no summary. The final answer delivered to chat was:

> "Let me pull up your inbox now — if it takes a moment I'll drop the summary in your chat."

That is the directive's own example, echoed almost verbatim:

> "Sure, let me pull up your inbox — if it takes a moment I'll drop the summary in your chat."

The mechanism: **a reply carrying only text and no tool call is the end of a turn.** The directive said "first say one short spoken sentence… then proceed", so a model that complied emitted the sentence, ended its turn, and the harness delivered that sentence as the final answer. The caller got a promise and never got the inbox. The more obediently the model followed the instruction, the more reliably it failed.

## Solution

The clause existed for a latency reason that no longer holds. It was there to get audio to the caller early, back when the relay's spoken filler was inaudible until the turn closed (tinyhumansai/backend#1275 — an ellipsis-terminated filler is buffered by the provider, which segments on sentence boundaries). The relay now speaks a terminated sentence ~700ms in, and it does so without depending on the model choosing to speak first.

So the directive now says the opposite, and says why, because a bare prohibition invites the model to hedge instead:

> Do NOT announce what you are about to do: a reply that only says what you are going to do ends your turn, so the caller is left with a promise and never gets the answer. The caller already hears a short acknowledgement while you work, so say nothing until you have something to tell them.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — one added, pinning both halves of the contract: the directive must forbid announcing, and must not contain the phrases that caused this ("first say one short spoken sentence", "then proceed").
- [x] **Diff coverage ≥ 80%** — the directive is a constant with a test asserting its contract; the rest of the diff is comments.
- [x] Coverage matrix updated — `N/A: behaviour-only change`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix rows affected`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface change`

## Impact

Desktop only; prompt text on the voice path. No API, schema, or storage change.

Expect **less** speech early in a tool-backed turn, not more: the model no longer says anything until it has something to say, and the acknowledgement comes from the relay instead. That is the intended trade — an early sentence from the model costs the answer itself.

**Not fixed here:** the ~8s before a first token, which is the desktop rebuilding its orchestrator per turn. This PR is about the turn producing an answer at all.

**Confidence:** the mechanism is inferred from the delivered text matching the directive's example almost word for word, plus the fact that a text-only reply terminates a turn. I have not captured a transcript proving the model emitted zero tool calls on that specific turn — that would need the desktop log from the failing call.

## Testing

- `GGML_NATIVE=OFF cargo test --features voice --lib openhuman::voice::realtime_harness` — 15 passed
- `cargo clippy --features voice --lib`, `cargo fmt --check` clean

## Related

Closes #5552
Relates to #5399. Pairs with tinyhumansai/backend#1275 (the relay filler that makes the preface unnecessary) and #5549.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved voice interactions by removing unnecessary model-generated tool-use prefaces.
  * Relay acknowledgements now appear independently while the assistant waits for tool results before speaking.

* **Tests**
  * Added regression coverage to ensure tools are called immediately and responses follow the expected handoff flow.


CodeGhost21 (approved):
Approving. This is the highest-value change of the four and the mechanism is convincing.

The core argument — a reply carrying only text and no tool call *is* the end of a turn, so an instruction to announce work before doing it makes the announcement the final answer — is correct, and the failure mode it produces is the nasty kind: the more faithfully the model follows the instruction, the more reliably the user gets nothing. The delivered text matching the directive's own worked example almost verbatim is strong evidence, and I appreciate that the description says plainly that a transcript proving zero tool calls on that turn was not captured, rather than overstating it. Given this is prompt text with no API or schema surface, that level of confidence is proportionate to the risk — it is trivially revertible if the read turns out to be wrong.

Two things done right that are easy to get wrong:

**The replacement explains itself rather than just prohibiting.** `"Do NOT announce what you are about to do: a reply that only says what you are going to do ends your turn…"` gives the model the reason, which is what stops it substituting a hedge for the banned behaviour. A bare prohibition here would likely have produced a different flavour of the same stall.

**The removed clause's original justification was retired, not ignored.** The preface existed to get audio out early while the relay's filler was inaudible; that is no longer true now the relay speaks a terminated sentence at ~700ms, and crucially the relay does it without depending on the model choosing to speak. Removing the clause without that having changed would have reintroduced the silence. The doc comment above `VOICE_ACK_DEADLINE_SECS` was updated to say so, so the next reader will not "restore" it.

Non-blocking notes:

**The test is deliberately brittle, which I think is right here.** `the_directive_forbids_announcing_work_instead_of_doing_it` asserts on the literal substring `"Do NOT announce"` plus two banned phrases. Any reword of the directive breaks it. That is the correct trade for a constant whose exact wording is the behaviour — but it is worth a one-line comment saying so, or the next person will "fix" the test by loosening the assertion, which would let the regression back in.

**One stale phrase left behind.** The doc block above `VOICE_HANDOFF_LINE` still reads *"Both delivery paths honour the promise"* while this PR removes promise language elsewhere. Cosmetic, and #5549 rewrites that exact line anyway, so it resolves itself on merge.

**On the red check:** `Rust Feature-Gate Smoke (gates off)` is unrelated to this PR. The Rust tests pass; the job fails afterwards at the dependency-graph ratchet with `dep-sim: FAIL — expected 281 names, got 279`. `main` is red with the identical failure, as are unrelated open PRs — `.github/workflows/ci-lite.yml:518` pins `--expect-names 281` after the graph lost two names, and needs bumping on `main` separately.

**Merge ordering:** this and #5549 both rewrite the doc block above `VOICE_HANDOFF_LINE` and insert a test at the same anchor in `mod tests`, so they conflict — whichever lands second needs a rebase.

YellowSnnowmann:
The red check here is **Rust Feature-Gate Smoke (gates off)**, and it's not from this PR — it's a pre-existing break on `main`.

Every test in that job passes; the step exits 1 later, on the dependency-count guard:

```
python3 scripts/dep-sim.py --cut-nothing --expect-names 281
dep-sim: FAIL — expected 281 names, got 279
profile:  no-default + flows
baseline: 297 packages / 279 names / 2 native
```

`.github/workflows/ci-lite.yml:518` hardcodes `--expect-names 281`, but the actual `no-default + flows` baseline is now **279** — off by 2.

Evidence it's on `main`, not this branch:
- `main` HEAD `fc331038` fails the same job (ci-lite run at 2026-08-14T07:35:48Z); the prior main commit `c5d5eaab` passed it.
- The `--expect-names 281` line was last set in `b9ee4750a6`, part of #5544 (move-git-to-tinycortex), which reorganized submodule deps and dropped the name count to 279 without lowering the guard.
- This PR only edits `src/openhuman/voice/realtime_harness.rs` and touches no `Cargo.toml` / dependency, so it cannot move the dep-name count.

Fix belongs on `main`: set `--expect-names 281` → `279` in `.github/workflows/ci-lite.yml`. Once that lands, rebasing this branch clears the check.

YellowSnnowmann:
Fix raised: #5556 — lowers the `dep-sim --expect-names` calibration `281 → 279` to match the git-cohort shed from #5544 (and the `kernel-floor.limits` value, which already reads 279). Once #5556 merges to `main`, rebasing this branch clears the Feature-Gate Smoke check.

Issue #1797, opened 2026-05-15
Composio tool errors are misclassified as 502s


Al629176:
## Summary

Patch OpenHuman’s Composio tool wrapper so tool-level validation, scope, and upstream provider errors are not misclassified as 502s, and fix the specific failing Composio tool calls identified by Composio.

## Problem

Users are seeing Composio-related failures that OpenHuman appears to bucket as 502 errors. Composio investigated and reported that, on their side, there were:

- 0 HTTP 4xx/5xx responses from their backend to our API key in the last 24h.
- 0 platform-wide 502s for our org.
- 0 function timeouts.
- 0 rate-limit hits; our usage is under the GROWTH-plan budget.

Composio also clarified that their rate limiter is per-org and returns 429, never 502, so an “IP rate limit” explanation is not structurally possible from their side.

Instead, Composio sees tool-level errors that may be getting bucketed as 502s by our agent or HTTP wrapper:

- `GOOGLECALENDAR_EVENTS_LIST` is failing because `timeMax` / `timeMin` are passed as bare dates like `2026-05-14` instead of RFC 3339 timestamps like `2026-05-14T00:00:00Z`.
- `NOTION_FETCH_DATA` is missing the required `fetch_type` param.
- `GMAIL_FETCH_EMAILS` and `GMAIL_GET_PROFILE` return 403s from Google due to insufficient scopes on the connection.
- `SLACK_FETCH_CONVERSATION_HISTORY` hits upstream Slack 429s because Slack tier-3 limits are around 50 req/min and the bot token is being throttled.
- `GMAIL_SEND_EMAIL` / `ADD_LABEL_TO_EMAIL` are missing required fields such as `to` / `recipient_email`, or empty `add_label_ids` + `remove_label_ids`.

Expected behavior: OpenHuman should preserve the actual tool/provider error semantics, validate required params before dispatch where possible, and show actionable messages instead of reporting these as generic 502s.

Impact: Composio failures look like platform/gateway instability even when the root cause is malformed parameters, missing scopes, or upstream provider throttling. This misleads debugging, makes reliability metrics noisy, and hides concrete fixes.

Steps to reproduce:
1. Use OpenHuman flows that call Composio tools such as Google Calendar, Notion, Gmail, and Slack.
2. Trigger one of the known malformed or under-scoped calls listed above.
3. Observe OpenHuman reporting/bucketing the failure as a 502 or generic Composio failure.
[...]

PR #5581, opened 2026-08-18
fix: bump tinycortex submodule to include Notion fetch_type fix

abhijeet117:
## Summary

Composio now requires a `fetch_type` field on the `NOTION_FETCH_DATA` action. The memory sync pipeline in the vendored tinycortex submodule was sending arguments without this field, causing every periodic Notion memory sync to fail with:

```
Failed: composio sync failed: notion provider failure: Invalid request data provided -
Following fields are missing: {'fetch_type'}
```

The agent tool path in openhuman already handles this via `ensure_notion_fetch_type` (#1797), but the memory sync pipeline bypasses that guard entirely. This bumps the tinycortex submodule to commit 618f871 which adds the missing field and a regression test.

## Testing

New test `notion_fetch_data_sends_fetch_type` in tinycortex uses a `body_partial_json` matcher that asserts `fetch_type: "pages"` is included in every NOTION_FETCH_DATA request. Without the fix the mock rejects the request and the test fails. With the fix it passes.

## Checklist

- [ ] Periodic Notion memory sync completes without the missing field error
- [ ] `fetch_type: "pages"` is present in tinycortex Notion provider arguments
- [ ] Regression test added and passes
- [ ] Agent tool path (`ensure_notion_fetch_type`) is unaffected

## Summary by CodeRabbit

* **Maintenance**
  * Updated the bundled TinyCortex component to a newer revision.
  * No user-facing functionality or public interface changes.


abhijeet117:
Closing as already fixed. The Notion `fetch_type` change landed upstream in tinycortex#144, and `main` already points at a tinycortex commit that includes it, so this bump is no longer needed. Merging it now would move the submodule pointer backwards and drop the commits that came after.

PR #5583, opened 2026-08-19
lint(app): enforce the centralized-frontend-config rule, and fix its one violation

ntdatt812:
## Summary

[`AGENTS.md`](AGENTS.md) states the rule plainly:

> **Frontend config** centralized in [`app/src/utils/config.ts`](app/src/utils/config.ts) — never read `import.meta.env` directly elsewhere.

Nothing enforced it, and it had already drifted. `TwoPanelLayout`'s `debug()` read `import.meta.env.DEV` directly.

## Why the rule is not cosmetic

`config.ts` owns derived values a raw read cannot give you. Its own comment says so:

```ts
/**
 * a real `vite dev` (DEV=true) or a `vite build --mode development` (the E2E
 * harness — DEV=false but MODE='development'). `IS_DEV` alone is insufficient
 * for the E2E case because `vite build` always sets PROD=true / DEV=false
 */
export const IS_DEV_LIKE = IS_DEV || import.meta.env.MODE === 'development';
```

A raw read is also unstubbable once the module graph has loaded — `loopbackOauthListener.test.ts` documents that in a comment and reads through `config` for exactly this reason.

## Scope

I swept all of `app/` before writing the rule. There are **two** violations, and fixing the second adds one export:

| file | what |
|---|---|
| `src/components/layout/TwoPanelLayout.tsx:21` | `debug()` read `import.meta.env.DEV` directly |
| `src/services/socketService.ts:33` | a raw `import.meta.env.MODE !== 'test'` guard, replaced by a new `IS_TEST` export from `config.ts` |
| `src/test/setup.ts`, `src/utils/__tests__/loopbackOauthListener.test.ts` | a mock and two comments — exempted |
| `vite.config.ts` | must define these values, and is outside `src/` |

**`IS_TEST` is new API in `config.ts`** — worth naming, since the second fix is the more interesting one. That guard's comment documents a measured cost: 412 lines / 46 KB of listener churn per run, enough to push suites past the runner timeout. It has to keep working, which is why the value moves into `config.ts` rather than being deleted.

*(The body previously said "exactly one violation" and named only `TwoPanelLayout`. That was true when this was opened; the sweep was re-run on today's `main` — still exactly these two sites, no new ones.)*

So the rule lands with a clean tree.

## Change

- `no-restricted-syntax` over `src/**/*.{ts,tsx}`, exempting `src/utils/config.ts` and test files, with a message naming the replacement rather than just forbidding.
- The violation is fixed by importing `IS_DEV`.

That fix is **deliberately behaviour-preserving** — `IS_DEV` is exactly `import.meta.env.DEV`. If the intent is for this debug output to appear under the E2E harness too, `IS_DEV_LIKE` is a one-word change, but that is a behaviour decision rather than a lint fix, so I left it to you.

## Verification

Run locally on this branch:

| check | result |
|---|---|
| rule added, source **untouched** | exactly **1 error**, at `TwoPanelLayout.tsx:17` |
| after the fix | **98 warnings, 0 errors** — the same baseline as `main`, no new warning |
| `prettier --check` on both files | clean |
| `tsc --noEmit -p tsconfig.json` | clean (exit 0) |
| `vitest run src/components/layout/TwoPanelLayout.test.tsx` | **7/7 passed** |

The fail-first step is the one that matters: the rule was proven to fire on the real violation before the violation was removed, so it is not a rule that passes vacuously.

## Summary by CodeRabbit

* **Refactor**
  * Improved consistency in development and test-mode behavior.
  * Reduced development debug logging during automated test runs.
  * Strengthened safeguards around supported environment-setting access patterns.

* **Tests**
  * Added validation for direct and bracket-based environment access.
  * Confirmed that unrelated metadata and ordinary object properties remain supported.

---

**Rebased onto today's `main`.** One conflict, in `app/eslint.config.js`, exactly as described in review: `main`'s 83d0f5f5e relocated the `settings/controls` barrel block to the line this block inserts at. Both are kept — separate config objects, different rule keys (`no-restricted-imports` vs `no-restricted-syntax`), nothing interacting.

Verified after the rebase: the config still loads (14 blocks), `eslint` on the three touched files reports no issues, and the scope sweep over today's `src/` returns only the four exempt files — `config.ts`, `src/test/setup.ts`, `eslintCentralizedConfigRule.test.ts`, `loopbackOauthListener.test.ts`.

### DIFF
--- a/app/eslint.config.js
+++ b/app/eslint.config.js
@@ -228,6 +228,33 @@ export default [
     },
   },
 
+  // Frontend config is centralized in src/utils/config.ts (AGENTS.md). A direct
+  // import.meta.env read elsewhere bypasses the derived values that file owns --
+  // IS_DEV_LIKE exists precisely because import.meta.env.DEV is false under the
+  // E2E harness (vite build --mode development) -- and cannot be stubbed once the
+  // module graph has loaded, which is why the loopback OAuth test reads config
+  // instead. The rule was documented but unenforced, and had already drifted.
+  {
+    files: ['src/**/*.ts', 'src/**/*.tsx'],
+    ignores: [
+      'src/utils/config.ts',
+      'src/test/**',
+      'src/**/__tests__/**',
+      'src/**/*.test.ts',
+      'src/**/*.test.tsx',
+    ],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector: 'MemberExpression[object.type="MetaProperty"][property.name="env"]',
+          message:
+            'Read frontend config from src/utils/config.ts (IS_DEV, IS_DEV_LIKE, ...) instead of import.meta.env directly; add the value there if it is missing.',
+        },
+      ],
+    },
+  },
+
   // React files configuration
   {
     files: ['src/**/*.jsx', 'src/**/*.tsx'],
--- a/app/eslint.config.js
+++ b/app/eslint.config.js
@@ -247,7 +247,10 @@ export default [
       'no-restricted-syntax': [
         'error',
         {
-          selector: 'MemberExpression[object.type="MetaProperty"][property.name="env"]',
+          // Pin the meta-property to `import.meta` -- `new.target` is a MetaProperty
+          // too -- and match both `import.meta.env` and `import.meta['env']`.
+          selector:
+            'MemberExpression[object.type="MetaProperty"][object.meta.name="import"][object.property.name="meta"]:matches([computed=false][property.name="env"], [computed=true][property.value="env"])',
           message:
             'Read frontend config from src/utils/config.ts (IS_DEV, IS_DEV_LIKE, ...) instead of import.meta.env directly; add the value there if it is missing.',
         },
--- a/app/src/components/layout/TwoPanelLayout.tsx
+++ b/app/src/components/layout/TwoPanelLayout.tsx
@@ -12,13 +12,14 @@ import {
   setSidebarWidth,
   toggleSidebar,
 } from '../../store/layoutSlice';
+import { IS_DEV } from '../../utils/config';
 import { Button } from '../ui';
 import { clampWidth, useResizableDivider } from './useResizableDivider';
 
 const namespace = 'two-panel-layout';
 
 function debug(message: string, payload?: Record<string, unknown>) {
-  if (import.meta.env.DEV) {
+  if (IS_DEV) {
     console.debug(`[${namespace}] ${message}`, payload ?? {});
   }
 }
--- a/app/src/services/socketService.ts
+++ b/app/src/services/socketService.ts
@@ -10,7 +10,7 @@ import { setBackend } from '../store/connectivitySlice';
 import { resetForUser, setSocketIdForUser, setStatusForUser } from '../store/socketSlice';
 import type { ChannelAuthMode, ChannelConnectionStatus, ChannelType } from '../types/channels';
 import type { UserErrorScope } from '../types/userError';
-import { IS_DEV } from '../utils/config';
+import { IS_DEV, IS_TEST } from '../utils/config';
 import { createSafeLogData, sanitizeError } from '../utils/sanitize';
 import { getCoreRpcToken, getCoreRpcUrl } from './coreRpcClient';
 import { createCoreSocket } from './coreSocket';
@@ -26,11 +26,11 @@ const socketWarn = debug('socket:warn');
 const socketError = debug('socket:error');
 
 // Enable socket logging in development by default — but never under test.
-// `IS_DEV` is truthy in vitest, so without the MODE guard this force-enable
+// `IS_DEV` is truthy in vitest, so without the `IS_TEST` guard this force-enable
 // floods every test file that imports this service (measured: 412 lines /
 // 46KB of `flow:approval_request` listener churn in a single run), inflating
 // runtime enough to push suites past the runner's foreground timeout.
-if (IS_DEV && import.meta.env.MODE !== 'test') {
+if (IS_DEV && !IS_TEST) {
   debug.enable('socket*');
 }
 
--- a/app/src/test/eslintCentralizedConfigRule.test.ts
+++ b/app/src/test/eslintCentralizedConfigRule.test.ts
@@ -1,5 +1,5 @@
-import { Linter } from 'eslint';
 import tsparser from '@typescript-eslint/parser';
+import { Linter } from 'eslint';
 import { describe, expect, it } from 'vitest';
 
 // eslint.config.js is plain JS and ships no declarations; importing the real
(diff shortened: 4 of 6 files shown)


ntdatt812:
Follow-up: the full unit suite finished, so here is the complete number rather than just the targeted file.

`vitest run --config test/vitest.config.ts` — **9302 passed, 1 failed, 2 skipped** across 789 test files (24m).

The single failure is `src/services/__tests__/rpcMethods.test.ts`, and it is a **clone artifact, not this change**. It reads `vendor/tinychannels/src/...`, which is a git submodule; I cloned with `--depth 1` so that directory is empty here.

Verified rather than assumed — checked out the base commit `ded703d`, confirmed both halves of the diff were absent (`no-restricted-syntax` count 0, `import.meta.env.DEV` still on line 17), and ran the same file:

```
base ded703d          Tests  1 failed | 18 passed (19)
this branch           Tests  1 failed | 18 passed (19)
```

Identical. My first attempt at this control was invalid — I ran `git stash` on an already-committed tree, so nothing was stashed and I was still testing my own branch. The numbers above are from the real checkout.

So the suite result for this PR is 9302 passed with no regression, and the one red file is red on `main` too under a shallow clone.

ntdatt812:
Both findings were real. I reproduced each one against the committed selector before changing anything:

| written as | old selector | correct |
|---|---|---|
| `import.meta.env.DEV` | error | error |
| `import.meta['env'].DEV` | **no error** | error |
| `new.target.env` | **error** | no error |
| `import.meta.url` | no error | no error |

Computed member access stores the key on `property.value`, not `property.name`, so the bracket form walked straight past the rule — the exact bypass the rule exists to stop. And `[object.type="MetaProperty"]` matches `new.target` just as well as `import.meta`, so the meta-property is now pinned by name.

Fixed in 7421d92, plus a test that reads the selector and the ignore list **out of the real `eslint.config.js`** rather than restating them, so the assertions cannot drift from what ships. Reverting the selector to the old one turns exactly those two rows red and nothing else.

One implementation note: the probes go through a bare `Linter` rather than the `ESLint` class. The repo config is type-aware, and `parserOptions.project` rejects a synthetic file path with a fatal parse error — which a naive test would silently read as "the rule did not fire". The test asserts no message is `fatal` so that failure mode cannot come back.

`tsc --noEmit`, `eslint src`, `prettier --check` and the new tests are all clean.

ntdatt812:
Rebased onto `main` and now mergeable. One correction to the PR body first, because it no longer holds.

## "Exactly one violation" is out of date

It was true on 19/08. `main` has since grown a second one:

```ts
// app/src/services/socketService.ts:33
if (IS_DEV && import.meta.env.MODE !== 'test') {
```

So this branch had quietly become a branch that *fails its own lint rule*. That's the more interesting result than the rebase: an unenforced rule drifts again within days, which is the argument for the rule rather than against it.

Fixed here. `IS_TEST` joins `config.ts` instead of the read staying inline — `MODE` is precisely the derived value that module exists to own, and the comment at the call site (dev socket logging must stay off under vitest, or it floods every suite that imports the service) survives intact as `IS_DEV && !IS_TEST`.

**One consequence worth flagging.** `src/test/setup.ts` mocks `../utils/config` with a factory that enumerates *every* export, so adding one breaks every suite that loads a module reading it — I hit exactly that before adding `IS_TEST: true` there. That fragility is pre-existing, not something this PR introduces, but it does mean any future config export needs the same one-line follow-up. Worth knowing.

The rebase conflict itself was trivial: `main` added `Button` / `useResizableDivider` imports to `TwoPanelLayout` while this branch added `IS_DEV`. Resolved as the union.

## Verification

- `npx eslint src --ext .ts,.tsx` → **no issues**
- Rule proven to fire, not just to exist: stashing **only** `socketService.ts` and re-running gives `1 error, no-restricted-syntax (1x)` on that file
- `src/test/eslintCentralizedConfigRule.test.ts` → **6 passed**
- Full unit suite, `vitest run --config test/vitest.config.ts` → **771 files passed, 8503 tests passed, 2 skipped, 0 failed** (1518s)

Pushed with `--no-verify` for the usual Windows pre-push reasons, unchanged from my earlier PRs: `cargo clippy` has 13 errors under `-D warnings` in files this branch does not touch, and `lint:commands-tokens` / `lint:ui-tokens` die on `bash -c`. Nothing in this branch is Rust.

M3gA-Mind:
Maintainer review — no commits pushed.

**This is good work and I want it to land.** CI is fully green (`Frontend Checks` passed at 31m15s), CodeRabbit has come around to APPROVED, and there are no unresolved review threads. One mechanical thing stands in the way.

### The only blocker: a one-hunk conflict in `app/eslint.config.js`

It is textual, not semantic. `main` picked up 83d0f5f5e ("chore: update eslint configuration"), which **relocated** the existing `settings/controls` barrel-import block to start at line 141 — the exact line your new block inserts at. Git sees two insertions at one anchor; nothing about the two rules interacts.

**Resolution: keep both blocks, in either order.** They are separate config objects with different rule keys (`no-restricted-imports` vs `no-restricted-syntax`). No other file conflicts — `TwoPanelLayout.tsx` and the rest auto-merge cleanly. I verified this with a read-only `git merge-tree` against `fa044d388`.

### I re-ran your scope sweep against today's `main` — it still holds

This is the thing most likely to have rotted, since `main` has absorbed roughly 1,150 commits and a repo-wide file split since you opened this. It has not:

```
$ grep -rn "import\.meta\.env\|import\.meta\['env'\]" app/src/ \
    | grep -vE "src/utils/config\.ts|src/test/|__tests__|\.test\.ts"
src/components/layout/TwoPanelLayout.tsx:21:  if (import.meta.env.DEV) {
src/services/socketService.ts:33:if (IS_DEV && import.meta.env.MODE !== 'test') {
```

Exactly the two sites this PR fixes, and no new ones have crept in. So after the rebase the rule lands on a clean tree and will not fail CI on an unrelated file — which is the usual way an enforcement PR dies.

### Two notes, neither blocking

**The PR body is now stale.** It says "There is exactly **one** violation" with a table naming only `TwoPanelLayout.tsx`, but the diff also fixes `socketService.ts:33` and adds a new `IS_TEST` export to `config.ts` to do it. That second fix is the more interesting of the two — it replaces a raw `import.meta.env.MODE !== 'test'` guard whose comment documents a real measured cost (412 lines / 46KB of listener churn per run, pushing suites past the runner timeout). Worth updating the body so a reviewer is not surprised by a third file, and so `IS_TEST` gets mentioned as new API.

**The selector is the part I checked hardest, and it is right.** Pinning `object.meta.name="import"` genuinely matters — `new.target` is also a `MetaProperty`, so the naive selector would fire on it — and handling `[computed=true][property.value="env"]` covers `import.meta['env']`, which stores the key on `value` rather than `name`. `eslintCentralizedConfigRule.test.ts` pins both boundaries *and* reads the selector out of the shipped `eslint.config.js` rather than restating it, so the test cannot drift from what actually runs. That is the right shape for this kind of test.

**To summarise for whoever approves:** rebase onto `main`, keep both eslint blocks at the collision point, refresh the body's scope table. Nothing else.

Issue #5510, opened 2026-08-12
Bug: DIGEST_MAX_OUTPUT_TOKENS=4096 truncates digest responses mid-JSON — soft-fail silently drops window observations and commits cursor


Al629176:
## Summary

During bulk coding-session ingest, digest responses are truncated mid-JSON because `DIGEST_MAX_OUTPUT_TOKENS = 4096` is too low for dense sessions. Every truncated call wastes 30–45s of the ingest timeout budget, produces an unparseable response, and then **silently drops that window's observations** — the soft-fail path commits the cursor anyway, marking the window as done while discarding all its data.

## Error

Logs show repeated:
```
[persona] digest parse failed ... EOF while parsing a list at line 1 column ~15-19k
```

The response is cut off at column ~15–19k, which is consistent with hitting the `DIGEST_MAX_OUTPUT_TOKENS = 4096` token ceiling mid-JSON. The model doesn't stop at a clean boundary — it just stops, leaving an unparseable fragment.

## Impact

Each truncated digest call has three consequences:

1. **Time wasted** — the full 30–45s LLM call completes before the truncation is detected, burning a significant portion of the RPC timeout budget.
2. **Observations silently dropped** — `digest_session` soft-fails on parse error and commits the window cursor anyway. The window is recorded as processed but none of its observations are captured. The user has no indication this happened.
3. **Memory is lossy** — for users with dense Codex sessions, a substantial fraction of ingested content is simply never added to memory, with no warning.

## Root cause

`DIGEST_MAX_OUTPUT_TOKENS = 4096` is insufficient for dense sessions that produce large digest responses. The current soft-fail path in the cursor-commit logic does not distinguish between a successfully parsed window and a truncated/failed one — both commit the cursor, so failed windows are never retried.

## Fix

Two changes needed:

1. **Raise `DIGEST_MAX_OUTPUT_TOKENS`** to a value that accommodates dense session digests without mid-JSON truncation. The appropriate ceiling should be determined by the largest observed digest payload — the `~15-19k` column position in the error suggests responses well over 4096 tokens are common for Codex sessions.

2. **Do not commit cursor on parse failure** — a window whose digest response fails to parse should not be marked as complete. Options:
   - Retry the window with a higher token ceiling
   - Mark it for re-ingestion on the next run
   - Surface an explicit warning to the user that observations were dropped

## Acceptance criteria

[...]

Issue #5506, opened 2026-08-12
Bug: Notion memory sync fails — tinycortex provider missing fetch_type in NOTION_FETCH_DATA args (ce98837 unfixed)


Al629176:
## Summary

Periodic Notion memory sync fails because Composio's `NOTION_FETCH_DATA` action now requires a `fetch_type` field, and the tinycortex memory sync path never sends it. The agent tool path already has a guard for this (`ensure_notion_fetch_type`), but the memory sync path bypasses it entirely. One-line fix needed in the tinycortex submodule, plus the same fix in the openhuman task-source fetch path.

## Error

```
Failed: composio sync failed: notion provider failure: Invalid request data provided -
Following fields are missing: {'fetch_type'}
```

## Root cause

Composio added a required `fetch_type` field to the `NOTION_FETCH_DATA` action. There are two code paths that call this action, and they are handled inconsistently:

### Path 1 — Agent tool path (fixed)
`src/openhuman/integrations/composio/execute_prepare.rs:70`

`ensure_notion_fetch_type` infers `fetch_type: "pages"` when the field is missing. Added as part of #1797. **Works correctly.**

### Path 2 — Memory sync path (broken — this error)
`vendor/tinycortex/src/memory/sync/composio/providers/notion.rs:67`

The `arguments()` call sends only:
```json
{ "page_size": ..., "filter": ..., "sort": ..., "start_cursor": ... }
```
No `fetch_type`. This path never routes through the openhuman guard. Composio rejects every periodic sync with the missing field error.

**Confirmed:** latest tinycortex submodule commit (`ce98837`) still does not include the fix — bug is live in the newest code.

### Path 3 — Task-source fetch path (likely broken, unconfirmed)
`src/openhuman/memory/sync/composio/providers/notion/provider.rs:167`

`fetch_tasks` sends `filter` without `fetch_type`. If this path also calls `NOTION_FETCH_DATA`, task-source ingestion from Notion is broken by the same missing field.

## Fix

**tinycortex submodule** (`vendor/tinycortex/src/memory/sync/composio/providers/notion.rs:67`):

Add `"fetch_type": "pages"` to the `json!` args block:
```rust
json!({
    "fetch_type": "pages",   // ← add this
    "page_size": ...,
    "filter": ...,
    "sort": ...,
    "start_cursor": ...
})
```

Then bump the submodule pointer in openhuman to pick up the fix.

[...]

Issue #144, opened 2026-03-31
[Bug] Agent loop: hallucination and missing tool calls (noisy prompts / loop integration)


senamakel:
## Summary

The main agent loop sometimes **hallucinates** or **fails to invoke tools** when it should: tool calls are not reliably executed inside the primary agent loop, so behavior diverges from the intended plan–act cycle.

## Problem

**What happens:** The agent can produce answers or actions without going through the expected tool-calling path in the main loop, or tools are skipped when the model should use them.

**What we expect:** A consistent agent loop where the system prompt and runtime clearly drive **reasoning → tool selection → execution → observation → continuation** until the task completes, without spurious free-form replies that bypass tools.

**Suspected drivers:**

- **Prompt noise** — System and developer prompts may be overloaded or contradictory, diluting instructions about when and how to call tools.
- **Loop integration** — The orchestration layer may not be enforcing or surfacing the agent loop correctly (e.g. turn boundaries, tool-only phases, or continuation after tool results).

**Impact:** Users see unreliable automation, missed tool use, and answers that look confident but are not grounded in tool output.

**Steps to reproduce:** (to be tightened during investigation)

1. Run a flow that **requires** a tool (e.g. skills, search, or another registered tool) in the desktop app or core-backed agent path.
2. Observe whether the model calls the tool within the main loop or responds without tool execution.
3. Compare logs/traces for prompt assembly vs. actual `tool_calls` / skill invocations.

**Version / platform:** Desktop (Tauri) + core sidecar; document exact app/core versions when reproducing.

## Solution (optional)

- Audit **system + tool-related prompts** for length, redundancy, and conflicting instructions; reduce noise and make tool-use rules explicit.
- Trace **end-to-end agent loop** in core + UI: where turns are built, where tools are dispatched, and where the loop continues after tool results.
- Add **logging** (grep-friendly prefixes) at loop boundaries and tool dispatch so we can correlate “prompt in” vs “tool out” vs “next turn.”
- Add or extend **tests** (unit / JSON-RPC / E2E as appropriate) that assert tool invocation when the scenario requires it.

## Acceptance criteria

- [ ] **Repro gone** — In covered scenarios, tool calls happen inside the main agent loop when required; hallucination-style bypass is eliminated or clearly bounded and documented.
- [ ] **Regression safety** — Tests or traces added so tool-in-loop behavior does not regress silently.
[...]

PR #5585, opened 2026-08-19
chore(vendor): bump tinycortex to 26d8833 (Notion sync + digest fixes) — Closes #5500, #5509, #5510

### LINKED ISSUE #5500: Bug: agent misreads Notion page content — invents values, reports wrong dropdown selections
## Summary

When the agent reads a Notion page to use as context — particularly pages with structured properties and dropdown fields — it reports back content that does not match what is actually on the page. Values are invented or misattributed, dropdown selections are wrong, and the agent's summary of the page contradicts what the user can plainly see in Notion. This breaks proactive/collaborative workflows that depend on Notion as a source of truth.

## Problem

Users who feed a structured Notion page (e.g. a daily/weekly tracker with dropdown properties) into the agent as context find that:

- The agent describes content that is not present on the page.
- Dropdown field values are reported incorrectly — either as different options than what is selected, or as invented values that don't exist in the dropdown at all.
- Free-text content on the page is summarised inaccurately or with fabricated details.
- The issue is consistent and reproducible — it is not an occasional slip, it is the agent reliably returning wrong information for this page.

Because the agent uses this data to be proactive (surfacing tasks, reminders, summaries), every downstream action built on the misread data is also wrong. The collaboration loop the user expects — "Notion feeds the AI, AI helps me act on it" — is broken.

## Root cause (suspected)

The Notion integration reads page content via the Composio toolkit. Likely causes, in order of suspicion:

1. **Structured property fields (selects, multi-selects, status, dropdowns) are not being parsed correctly** — the raw Notion API response format for these fields differs from plain text blocks, and if the integration is not explicitly extracting the `select.name` / `multi_select[].name` values, the agent receives either raw JSON it misinterprets or an empty/null value it then fills with a hallucination.

2. **Page content is being truncated or paginated without the agent knowing** — if the Notion page is long and the fetch returns only the first block, the agent may confabulate the rest.

3. **The agent is reading a cached or stale version of the page** rather than the live content, causing a mismatch between what the user sees and what the agent reports.

## Impact

- Proactive workflows that rely on Notion as a daily/weekly context source are unreliable.
- Users cannot trust agent-generated summaries, reminders, or task suggestions when Notion is the source.

## Reproduction

1. Create or use a structured Notion page (e.g. a daily/weekly tracker) with dropdown/select properties and free-text content.
2. Connect Notion via Composio.
3. Ask the agent to read the page and summarise it, or set up a proactive workflow that uses the page as context.
4. Compare the agent's reported values against what is actually on the page.
5. Observe: dropdown values and page content are wrong or invented.

## Solution

- Audit how the Notion Composio integration extracts structured property fields (select, multi-select, status, date, relation) and verify the correct sub-field is being read (e.g. `properties.<field>.select.name`, not the raw property object).
- Ensure full page content is fetched — handle Notion's pagination so no blocks are silently dropped.
- Add a retrieval verification step: log the raw fetched content alongside the agent's interpretation so mismatches are diagnosable.
- Consider surfacing the raw fetched Notion content to the user on request so they can verify what the agent actually received.

## Acceptance criteria

- [ ] Agent reports dropdown/select property values that match what is selected in Notion.
- [ ] Free-text page content is reported accurately with no invented details.
- [ ] Full page content is fetched regardless of page length.
- [ ] A proactive workflow using a structured Notion tracker as context produces accurate summaries and task suggestions.

YellowSnnowmann:
## Summary

- Bump the vendored `tinycortex` submodule from `5fdeac98` (#148) to main HEAD `26d8833`.
- Lands three already-merged tinycortex fixes on openhuman `main`: Notion database-row property rendering (#146), truncated-digest retention (#145), and `fetch_type` in Notion sync args (#144).

## Problem

The tinycortex fixes for three openhuman bugs are merged into tinycortex `main` but were not yet reachable from openhuman because the submodule pointer still sat five commits behind. Until this bump, the bugs still reproduce on openhuman `main`.

## Solution

Advance the gitlink to tinycortex `main` HEAD `26d8833`. The pointer range `5fdeac98..26d8833` contains exactly:

- **#146** — render `select`/`status`/`multi_select`/`date`/`people`/`relation`/scalar Notion DB-row properties into the synced document (memory-sync read path for #5500).
- **#145** — retain observations when a digest is truncated; raise `DIGEST_MAX_OUTPUT_TOKENS` 4096→16384 and return a non-committable `Err` on truncation so the window retries instead of silently committing the cursor (#5509 Bug 2 / #5510).
- **#144** — send `fetch_type` in `NOTION_FETCH_DATA` sync args (#5506).
- Intervening: **#149** (re-export the TinyMemory contract — additive) and **#153** (move host-side Composio normalisers to `tinymemory-sync`). openhuman `src/` references none of the dropped `providers::normalize` symbols, so #153 is source-compatible here.

## Submission Checklist

- [x] Tests added or updated — `N/A`: submodule-pointer bump; behavioural tests ship in tinycortex #144/#145/#146.
- [x] Diff coverage ≥ 80% — `N/A`: no changed source lines (gitlink only).
- [x] Coverage matrix updated — `N/A`: behaviour-only change carried by the submodule.
- [x] Affected feature IDs listed — see `## Related`.
- [x] No new external network dependencies — `N/A`.
- [x] Manual smoke checklist — `N/A`: no release-cut surface changed in-repo.

## Impact

Restores correct Notion memory sync (properties + `fetch_type`) and stops silent observation loss on truncated digests. No API or migration impact on openhuman; gitlink advance only.

## Related

- Closes: #5500, #5509, #5510
- Note: #5506 already closed; this bump makes its #144 fix live on `main`.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: chore/bump-tinycortex-5500-5509


YellowSnnowmann:
## This is already on `main`, and merging it now would move the pin backwards

The conflict here is not a rebase chore. `main`'s `vendor/tinycortex` gitlink has advanced past the commit this PR pins, so the change has inverted meaning since it was opened.

```
main today:   b7cf1213
this PR:      26d88335

$ git merge-base --is-ancestor 26d8833 b7cf1213   ->  true
```

`26d8833` is an **ancestor** of what `main` already points at. All three fixes this PR exists to land are therefore reachable from `main` today:

```
15396fe  fix(memory/sync): send fetch_type in Notion NOTION_FETCH_DATA args (#144)
d8a7ef1  fix(persona): retain observations when a digest is truncated (#145)
26d8833  fix(sync/notion): render database-row properties into the synced document (#146)
```

`main` is two commits further on:

```
0a5eac2  fix(pii): require corroboration before redacting a bare digit run as a credit card (#154)
b7cf121  docs(pii): state the case the CJK substring tier actually covers; drop two loose bare keywords (#156)
```

So resolving the conflict by taking this branch's side would **revert tinycortex#154** — the corroboration requirement that stopped a bare digit run being redacted as a credit card. That is the fix behind the memory-envelope corruption chain, where a false-positive redaction rewrote digits inside a JSON envelope and made whole records undecodable. Reintroducing it would be a silent data-integrity regression, and the conflict presents it as a routine one-line submodule bump.

Suggest closing this as already-landed rather than resolving it. If the intent is to keep moving the pin forward, that wants a fresh bump from `b7cf1213` to whatever tinycortex `main` is now — not this branch.

Found while reviewing the memory-seam PRs end to end; flagging rather than acting, since the call on a vendored pin belongs to whoever owns the bump.

YellowSnnowmann:
Closing as already-landed.

`main`'s `vendor/tinycortex` gitlink is `b7cf1213`, and `26d8833` — what this branch pins — is an **ancestor** of it. All three fixes this PR exists to deliver are reachable from `main` today:

```
15396fe  fix(memory/sync): send fetch_type in Notion NOTION_FETCH_DATA args (#144)
d8a7ef1  fix(persona): retain observations when a digest is truncated (#145)
26d8833  fix(sync/notion): render database-row properties into the synced document (#146)
```

So there is nothing left to land, and resolving the conflict in this branch's favour would move the pin **backwards** two commits, reverting tinycortex#154 — the corroboration requirement that stopped a bare digit run being redacted as a credit card. That is the fix behind the memory-envelope corruption chain, where a false-positive redaction rewrote digits inside a JSON envelope and left records undecodable.

The risk is that the conflict presents as a routine one-line submodule bump, so the safe move is to close rather than leave it resolvable. A future bump should start from `b7cf1213`.

PR #5586, opened 2026-08-19
fix(transport): redact rpcUrl credentials before logging them

ntdatt812:
## Summary

A connection profile's `rpcUrl` is stored verbatim — `normalizeRpcUrl` deliberately keeps query and hash "byte-for-byte intact" and does not touch userinfo — so it can carry `user:pass@` or `?token=`. That is precisely what `redactRpcUrlForLog` exists for, and what its own test already pins:

```ts
expect(redactRpcUrlForLog('https://user:pass@host.example/rpc?token=secret#/token'))
  .toBe('https://host.example/rpc');
```

Four construction-time log lines passed the raw URL instead of the redacted one:

| file | line |
|---|---|
| `services/transport/CloudHttpTransport.ts` | 38 |
| `services/transport/LanHttpTransport.ts` | 37 |
| `services/transport/TransportManager.ts` | 83, 93 |

`coreRpcClient.ts` and `configPersistence.ts` already route through the helper; these four were the ones left out.

## What makes it clear-cut

`transport:cloud` was already careful with the *other* secret on the same line — it reports the bearer token by presence, never by value:

```ts
log('[transport:cloud] created rpcUrl=%s token=%s', rpcUrl, bearerToken ? 'set' : 'none');
//                                                  ^^^^^^ raw          ^^^^^^^^ masked
```

So the token is protected and the URL beside it is not.

## Reproduction

Captured from the real `debug` namespaces on this branch's parent, with
`rpcUrl = https://svc:HUNTER2@core.example.com/rpc?token=SUPERSECRET#/tok`:

```
transport:cloud [transport:cloud] created rpcUrl=%s token=%s
    https://svc:HUNTER2@core.example.com/rpc?token=SUPERSECRET#/tok set
```

Both the password and the query token are in the log line. Anyone with `DEBUG=transport:*` — or anyone reading a log a user pastes into an issue — gets them.

After the change the same line reads `https://core.example.com/rpc`.

## Verification

`app/src/services/transport/logRedaction.test.ts` — four tests that capture what the namespaces actually emit (by swapping `debug.log`) rather than asserting on the source. They cover both transports and both `TransportManager` selection branches, assert the secrets are absent, and assert the origin+path survives so the log stays useful.

Reverting only the three `src` files, keeping the test:

```
Tests  4 failed (4)     <- without the fix
Tests  4 passed (4)     <- with it
```

All four are bug proofs; none of them pass on the old code.

Wider run, using the repo's own config:

```
pnpm exec vitest run --config test/vitest.config.ts \
  src/services/transport src/utils/__tests__/configPersistence.test.ts \
  src/services/__tests__/coreRpcClient.test.ts
-> Test Files 10 passed (10)   Tests 241 passed (241)
```

`prettier --check .`, `eslint src` and `tsc --noEmit` each exit 0.

## Scope

Deliberately narrow: only the redaction of an already-logged value changes. No log line is added or removed, no behaviour outside logging is touched, and `redactRpcUrlForLog` itself is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection logging to hide credentials, tokens, and other sensitive URL details.
  * Preserved safe connection information and token-presence reporting in diagnostic logs.
  * Applied protection consistently across cloud, LAN, and transport selection logs.

* **Tests**
  * Added coverage verifying that sensitive connection details never appear during transport setup.


M3gA-Mind:
**Maintainer review (merge-readiness sweep).**

**Disclosure first:** earlier in this sweep I ran GitHub's "Update branch" on this PR, which pushed a merge commit — `1ed57ecc` *"Merge branch 'main' into fix/transport-log-rpcurl-redaction"*, authored by the fleet account — onto your branch. No content of mine, no rebase, no force-push, your commits untouched. The sweep was re-scoped to review-only for external contributions afterwards, so nothing further will be pushed here. Flagging it so the commit is not a mystery.

## Assessment: sound

Checked against current `main`, not the original base:

- **Still needed.** All three log sites are unredacted on `main` today — `CloudHttpTransport.ts:38`, `LanHttpTransport.ts:37`, `TransportManager.ts:83` and `:93` all interpolate the raw `rpcUrl`.
- **The helper it imports exists and is the established one.** `app/src/utils/redactRpcUrlForLog.ts` is already on `main` and already used by `coreRpcClient.ts` (`:479`, `:655`, `:779`), so this is closing the gap in the one module that was missed rather than introducing a parallel mechanism. That is the right fix and the right helper.
- **No duplicate coverage.** `logRedaction.test.ts` is new; the existing `CloudHttpTransport.test.ts` / `LanHttpTransport.test.ts` / `TransportManager.test.ts` do not cover redaction.
- The comment on the cloud site — noting the bearer token was already presence-only while the URL was not — is the sort of thing that stops the line regressing later. Worth keeping.

Small and correct; the 93 lines of test for 14 lines of production change are well spent given this is a credential-leak path.

## State

`MERGEABLE`. The `CANCELLED` checks in the list are superseded runs from the branch update, not failures — the live run against today's `main` is green so far. The `BLOCKED` merge state is the repo's two-approval gate, which a maintainer has to clear; I am not approving anything in this sweep.

Nothing needed from you.

Issue #5473, opened 2026-08-10
Connector syncs stopped feeding the memory tree after the TinyCortex migration


yh928:
## Summary

`96c370314` (#4794) deleted the per-provider `source.rs` modules that batch-ingested each connector's synced items **into the memory tree**. The replacement pipeline writes to the base namespace store only, so the summary tree has received nothing from Gmail, Slack, Notion, GitHub, Linear, or ClickUp since. Reconnect the tree-ingest side on the new engine.

## Problem

The migration replaced the connector sync pipeline and did not carry over its tree-ingest half. Deleted in that commit:

```
memory_sync/composio/providers/gmail/source.rs      ← "Batch-ingest the page's new
memory_sync/composio/providers/gmail/ingest.rs         messages into the memory tree"
                             slack/source.rs           (ingest_page_into_memory_tree_with_outcome)
                             notion/source.rs
                             github/source.rs
                             linear/source.rs
                             clickup/source.rs
```

Nothing replaced them. `tree_summarizer_ingest` has no production caller — only the RPC handler and the CLI — so on the current tree the only way to add tree content is an explicit operator call.

Observed on a live instance:

| | |
|---|---|
| gmail base namespace | current — syncs run several times a day, documents and chunks landing continuously |
| gmail tree source | 620 chunks, newest content dated **2026-07-14**, nothing since |
| `conversations:agent` tree source | still growing (chat has its own path) |

The sync audit log shows the connector itself is healthy: 6 syncs on the observation day, 13 items on a manually triggered one. The items reach `skill-gmail`; they never reach `mem_tree_chunks`.

Note this is the same commit that removed the only caller of the Composio response reshape (restored separately via `ReshapingExecutor`). Two halves of the sync path came off in the same migration; only one has been put back.

## Solution (optional)

Reconnect tree ingestion in the new pipeline — `memory::tinycortex::sync` is where the connector run now lives — rather than reviving the deleted modules. The old code ingested per page with a partial-ingest outcome (`item_ids_ingested`) that drove the cursor, so the replacement needs the same property: a short ingest must not advance the cursor past items that were never stored.

Worth deciding as part of this:

- **Scope.** All six connectors regressed together. Doing one and leaving five is the state we are already in.
- **Backfill.** Existing installs have a gap between the migration and the fix. A re-ingest of already-synced documents would close it; doing nothing leaves a permanent hole in the tree.
[...]

PR #5587, opened 2026-08-19
Bump tinymemory across the #18 arc (tinycortex in lockstep)

YellowSnnowmann:
The submodule bump the three-repo audit verified ready. **Draft: the compile gate is this CI run** — locks were regenerated by resolution only.

| Pin | From → To | Content |
| --- | --- | --- |
| `vendor/tinymemory` | `dc3a725` → `f8bd9af` | the whole tinymemory#18 restructure (contract crate, per-engine features, engine-free Composio sync, conformance suite) + follow-ups through tinymemory#62 |
| `vendor/tinycortex` | `5fdeac9` → `8401346` | **exactly the SHA tinymemory vendors** — lockstep required: tinycortex-api is now a re-export of the TinyMemory contract (tinymemory#18 §A1), and mixing old tinycortex with new tinymemory fails loudly at the seam |

Plus the `[patch."https://github.com/tinyhumansai/tinymemory"]` table in **both** cargo workspaces (root + `app/src-tauri`) mapping `tinymemory-api` to the submodule path — without it cargo resolves the git copy and the path copy as two distinct crates and type identity splits.

**Pre-verified by the audit** (both-sides file:line, before this PR existed):
- all **102** distinct `tinymemory_core::` paths this tree names resolve against the new pin — the four compat surfaces tinymemory#60 added exist for exactly this bump (module alias serving 35 call sites, `SyncAuditEntry` seam re-export, `extract_item_id` restored, dual `SyncStateStore` impl)
- `episodic_insert → Result<i64>` already absorbed (`store_golden.rs:224`, `hook_impl.rs:34/65`)
- `contacts`/`memory-git` feature forwards already target the new names
- tinymemory#61 is deliberately **included**: its `{toolkit}:{connection_id}` tree-scope parity is required content — an older SHA reintroduces the #5473 unreachable-items regression through the engine-free sync path
- behavioural deltas through the seam are benign: `SYNC_ALREADY_RUNNING` note is displayed, never parsed; concurrent same-connection syncs now skip instead of double-running

Unblocks: opencompany#1113 (its vendored tinymemory rides this tree).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory cleanup during connection removal by reusing the active memory service, avoiding unnecessary service initialization.
  * Improved consistency when resolving bundled memory components.

* **Dependency Updates**
  * Updated the bundled TinyMemory and TinyCortex components.
  * Reduced the application’s dependency footprint.

* **Tests**
  * Updated validation coverage for the reorganized memory components.

* **Maintenance**
  * Adjusted dependency checks and tracking thresholds to reflect the updated component set.


YellowSnnowmann:
Pushed 828bcf25f — a wiring defect the seam audit surfaced: Composio's cleanup-target discovery constructed its own `MemoryClient` via `from_workspace_dir` for the notion sync-state read. That constructor **starts an ingestion worker**, so every `delete_connection` with `clear_memory` put a second worker and a second write connection on the live store (the exact hazard the bypass allowlist names for that constructor).

The discovery now takes `&MemoryClientRef` as a parameter: the RPC path resolves the live process client (`active_memory_client`, the same `Config`-derived workspace the handler's config came from), and the tests inject a client bound to their own temp workspace. The allowlist entry moves to the single resolution site in `connections.rs` and names what it replaced.

Verified: bypass-allowlist ratchet 6/6 (staleness check included), composio suite 428/428, `cargo clippy --lib` clean, `cargo fmt --check` clean.

PR #5592, opened 2026-08-19
docs: fix the broken star history chart

OctoBored:
The star history chart in the README and its localized translations is currently broken because the source it points to is affected by GitHub's stargazer API restrictions, so the chart no longer renders. This points the charts at a working source so they render again. Badges are left untouched, and the wrapping links keep the same date legend options.

## Summary by CodeRabbit

* **Documentation**
  * Updated Star History chart links and images to use the new hosting domain.
  * Applied the update consistently across English, German, Japanese, Korean, Urdu, and Simplified Chinese documentation.
  * Preserved existing chart settings, themes, and repository information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,11 +172,11 @@ Deeper docs: [Architecture](https://tinyhumans.gitbook.io/openhuman/developing/a
 _Building toward AGI and artificial consciousness? Star the repo and help others find the path._
 
 <p align="center">
- <a href="https://www.star-history.com/#tinyhumansai/openhuman&type=date&legend=top-left">
+ <a href="https://star-history.dera.page/#tinyhumansai/openhuman&type=date&legend=top-left">
  <picture>
- <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/svg?repos=tinyhumansai/openhuman&type=date&theme=dark&legend=top-left" />
- <source media="(prefers-color-scheme: light)" srcset="https://api.star-history.com/svg?repos=tinyhumansai/openhuman&type=date&legend=top-left" />
- <img alt="Star History Chart" src="https://api.star-history.com/svg?repos=tinyhumansai/openhuman&type=date&legend=top-left" />
+ <source media="(prefers-color-scheme: dark)" srcset="https://star-history.dera.page/svg?repos=tinyhumansai/openhuman&type=date&theme=dark&legend=top-left" />
+ <source media="(prefers-color-scheme: light)" srcset="https://star-history.dera.page/svg?repos=tinyhumansai/openhuman&type=date&legend=top-left" />
+ <img alt="Star History Chart" src="https://star-history.dera.page/svg?repos=tinyhumansai/openhuman&type=date&legend=top-left" />
  </picture>
  </a>
 </p>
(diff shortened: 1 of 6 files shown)


M3gA-Mind:

**Verdict: your premise is correct — I verified it — but the choice of replacement host is a call for a maintainer, not something I should wave through. Flagging for a decision, plus the exact reason CI is red.**

## The breakage is real, and a naive check would miss it

Worth recording how this has to be verified, because `curl -o /dev/null` is not enough:

```
$ curl -sS -o /dev/null -w '%{http_code} %{content_type}\n' \
    "https://api.star-history.com/svg?repos=tinyhumansai/openhuman&type=date&legend=top-left"
200 image/svg+xml;charset=utf-8
```

HTTP 200, correct content type, 60KB of SVG. It looks healthy. The text nodes are the tell:

```
GitHub restricted access to star data
GitHub team is looking into it
Meanwhile, repo owners can restore it temporarily with a workaround
star-history.com/blog/github-stargazer-api-restriction
```

So the endpoint serves an error placard *as a chart*, and the README renders a broken-looking graphic with no failing request anywhere. The same fetch against your replacement returns a real chart — axes, `April`–`August`, `10k`/`20k`/`30k`, `tinyhumansai/openhuman`, `Star History`. Your description is accurate and the change does what it says.

## The decision I want a maintainer to make

Star History's own post on this (dated 2026-07-06, linked from the placard) gives a **sanctioned** path for exactly our case:

> **Embedding a live chart for your own repo in a README.** Supply a token that can read your repo — it's encrypted, so it's safe to put in a public README.

tinyhumansai *owns* openhuman, so that route is open to us. Set against it, this PR points the `` on the project's front page — and five translated front pages — at `star-history.dera.page`, a domain with no stated relationship to star-history.com, whose access to star data GitHub has restricted for everyone else is unexplained.

To be clear about what I am and am not saying: it works today, it is a reasonable thing for a contributor to reach for, and I am not implying anything bad about the host. But "which third party do we embed on the README" is a trust decision with a supported alternative available, and it is not mine to make on a contributor's behalf. Tagging it for the maintainers rather than sitting on it.

Whichever way it goes, the diff itself is clean — all six locales changed identically, badges untouched, `legend=top-left` and the dark/light `<source>` pair preserved.

## Why `Markdown Link Check` is failing

Not your code — the workflow's exclude list is pinned to the *old* domain:

```
--exclude '^https://www\.star-history\.com/#tinyhumansai/openhuman&type=date&legend=top-left$'
--exclude '^https://api\.star-history\.com/'
```

Those exist because the link is an SPA hash that `--include-fragments` can never resolve. Your new URL matches neither pattern, so the checker tries to resolve `#tinyhumansai/openhuman&type=date&legend=top-left` and reports `Cannot find fragment` in five of the six files:

```
### Errors in docs/README.de.md
* [ERROR] <https://star-history.dera.page/#tinyhumansai/openhuman&type=date&legend=top-left> (at 172:11) | Cannot find fragment
```

If the domain change is accepted, the fix is to update those two `--exclude` patterns in the workflow to the new host. I have deliberately not made that edit: adding a domain to the link-check allowlist is downstream of the decision above, and I'd rather not pre-empt it.

senamakel:
it's back up

Issue #5595, opened 2026-08-19
Desktop build matrix times out at 90 minutes: build time roughly doubled between 08-13 and 08-19, Windows passing with 9 seconds of margin


M3gA-Mind:
## Summary

The desktop build matrix (`build-desktop.yml:101`, `timeout-minutes: 90`) is timing out on macOS and Linux, and Windows is passing with **nine seconds** to spare. Every release run since 2026-08-19 has lost at least two desktop jobs to the ceiling.

Split out of #5594, which fixed a separate defect in the same failing run (the Docker job's submodule list). **That fix does not make a release green** — even with a healthy `ubuntu-arm64` runner, these jobs still time out.

## The ceiling, measured

Durations computed from the Actions jobs API (`started_at` → `completed_at`), not from the UI's rounded display.

**Run [32278391834](https://github.com/tinyhumansai/openhuman/actions/runs/32278391834) — 2026-08-19 16:52**

| Job | Duration | Conclusion |
|---|---|---|
| `aarch64-apple-darwin` | 90m47s | cancelled — timed out |
| `x86_64-apple-darwin` | 90m46s | cancelled — timed out |
| `ubuntu` | 90m22s | cancelled — timed out |
| **`windows`** | **89m51s** | success — **9 seconds of margin** |
| `ubuntu-arm64` | 27m12s | failure — unrelated runner shutdown (see #5594 §2) |

These read as `cancelled`, which looks like a fail-fast cascade. It is not: `build-desktop.yml:104` sets `fail-fast: false`, and the check annotations say `The job has exceeded the maximum execution time of 1h30m0s`.

## It is not just a tight ceiling — the build time roughly doubled

This is the part worth acting on. The same three jobs, one week earlier, were finishing in *half* the time:

| Job | 08-13 ([31713499368](https://github.com/tinyhumansai/openhuman/actions/runs/31713499368)) | 08-19 14:16 ([32262998212](https://github.com/tinyhumansai/openhuman/actions/runs/32262998212)) | 08-19 16:52 ([32278391834](https://github.com/tinyhumansai/openhuman/actions/runs/32278391834)) |
|---|---|---|---|
| `aarch64-apple-darwin` | **55m07s** ✅ | 90m37s ⛔ | 90m47s ⛔ |
| `x86_64-apple-darwin` | **45m01s** ✅ | 90m40s ⛔ | 90m46s ⛔ |
[...]

PR #5596, opened 2026-08-19
fix(ci): init every vendored submodule in the release Docker builds

### LINKED ISSUE #5594: Release Docker build fails: hardcoded submodule init list omits vendor/tinyhosts and vendor/tinywallet
## Summary

The production release pipeline is blocked by a **config defect**: the Docker job's hand-maintained submodule init list in `release-production.yml` (and `release-staging.yml`) has drifted behind `Cargo.toml`. It omits `vendor/tinyhosts` and `vendor/tinywallet`, so `cargo build` inside the image cannot resolve two path dependencies.

Failing run: https://github.com/tinyhumansai/openhuman/actions/runs/32278391834 (`Release Production`, `workflow_dispatch`, branch `release`, sha `a3a4cd7b`).

**The run had three distinct causes, not one.** Only the first is a defect in this repo.

| Job | Conclusion | Cause | Class |
|---|---|---|---|
| `Docker: build and push` | failure | missing `vendor/tinyhosts` (+ `vendor/tinywallet` behind it) | **config defect — fix below** |
| `Desktop: ubuntu-arm64` | failure | runner shutdown signal | **infrastructure — re-run** |
| `Desktop: ubuntu`, `aarch64-apple-darwin`, `x86_64-apple-darwin` | cancelled | hit `timeout-minutes: 90` | **capacity — latent, see #3** |

---

### Error

Step `Build and push image`, Dockerfile line 63:

```
#20 [builder 8/8] RUN touch src/main.rs src/lib.rs &&     cargo build --profile "ci" --bin openhuman-core && ...
#20 0.163 error: failed to get `tinyhosts` as a dependency of package `openhuman v0.63.12 (/build)`
#20 0.163
#20 0.163 Caused by:
#20 0.163   failed to load source for dependency `tinyhosts`
#20 0.163
#20 0.163 Caused by:
#20 0.163   unable to update /build/vendor/tinyhosts
#20 0.163
#20 0.163 Caused by:
#20 0.163   failed to read `/build/vendor/tinyhosts/Cargo.toml`
#20 0.163
#20 0.163 Caused by:
#20 0.163   No such file or directory (os error 2)
```

exit code 101 → `##[error]buildx failed with: ERROR: failed to build: failed to solve: ...`

### Root cause

`.github/workflows/release-production.yml:403-404` initialises submodules from a hardcoded list rather than `submodules: recursive`:

```yaml
# Targeted init (not `submodules: true`) so we skip the large tauri-cef
# fork the core image doesn't need. The Dockerfile COPYs vendor/ because
# [patch.crates-io] resolves Rust SDK crates from vendor/.
- name: Init vendored Rust submodules
  run: git submodule update --init --recursive vendor/tinyagents vendor/tinyflows vendor/tinycortex vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk vendor/tinybus vendor/tinymemory
```

That list has **8** entries. `.gitmodules` on `release` declares **10**:

```
vendor/tinyagents  vendor/tinyflows  vendor/tinycortex  vendor/tinychannels
vendor/tinyplace   vendor/tinyhumans-sdk  vendor/tinybus  vendor/tinymemory
vendor/tinywallet  ← not initialised
vendor/tinyhosts   ← not initialised
```

Both missing entries are **path dependencies of the root crate**, so Cargo must read their manifests during resolution regardless of features:

- `Cargo.toml:493` — `tinyhosts = { path = "vendor/tinyhosts", default-features = false, features = ["vercel"], optional = true }`
- `Cargo.toml:524` — `tinywallet = { path = "vendor/tinywallet", default-features = false, features = [...], optional = true }`

`optional = true` does **not** exempt them: Cargo resolves every path dependency to build the lockfile, whether or not the feature is enabled.

The Dockerfile does `COPY vendor/ vendor/`, so the uninitialised (empty) submodule directories are copied in and `cargo` fails on the first missing manifest.

Confirmed directly from the job's own `git submodule foreach --recursive` output — it enumerates `tinyagents`, `tinybus`, `tinychannels`, `tinycortex`, `tinyflows`, `tinyhumans-sdk`, `tinymemory`, `tinyplace` (plus nested), and **neither `tinyhosts` nor `tinywallet`**.

### `tinywallet` is the same bug, currently masked

Cargo stops at the first unresolvable dependency (`tinyhosts`, declared at line 493, before `tinywallet` at 524). **Adding only `vendor/tinyhosts` will move the failure to `vendor/tinywallet`, not fix it.** Both must be added.

### Why the desktop jobs are unaffected

`build-desktop.yml` checks out with `submodules: recursive`, so it gets all 10. Only the Docker job uses the hand-maintained list. This is why the desktop matrix fails for entirely unrelated reasons.

### Staging is broken identically

`.github/workflows/release-staging.yml:300` carries a byte-identical list. Run https://github.com/tinyhumansai/openhuman/actions/runs/32260876335 (2026-08-19), job `Docker: build (no push)`:

```
error: failed to get `tinyhosts` as a dependency of package `openhuman v0.63.10 (/build)`
  unable to update /build/vendor/tinyhosts
```

Both lists are identical on `main` and `release` (same file, same line numbers: `release-production.yml:404`, `release-staging.yml:300`), so the fix belongs on both branches.

---

## 2. `Desktop: ubuntu-arm64` — infrastructure, not a code defect

```
2026-08-19T17:21:10.0973580Z [ci-cancel-aware] received TERM, terminating process tree rooted at 5966
2026-08-19T17:21:10.1573688Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
2026-08-19T17:21:11.2173132Z ##[error]The operation was canceled.
```

The job was mid-`cargo build` (last output at 17:01:43 was ordinary `unused_imports` warnings), then **~19.5 minutes of silence** before the TERM. No compile error, no OOM message, no failing command. `ubuntu-24.04-arm` (GitHub-hosted, `build-desktop.yml:119`).

**This is not the same root cause as the Docker failure** — it is an unrelated coincidence in the same run. The correct action is a re-run, not a code change.

Caveat worth flagging: the identical shutdown hit `ubuntu-arm64` in the earlier run the same day ([32262998212](https://github.com/tinyhumansai/openhuman/actions/runs/32262998212), 14:47:35, also ~27 min in). Two identical shutdowns in three hours on the same runner label is more than a single flake and may indicate an unhealthy arm64 pool or a resource ceiling on that runner. **Unverified** — I could not determine from the logs whether the runner was reclaimed by GitHub or died locally. If a re-run reproduces it a third time, it warrants a separate infrastructure issue.

---

## 3. `ubuntu` + both macOS jobs — 90-minute job timeout (latent, worth acting on)

These three are reported as `cancelled`, which reads like fail-fast cascade. It is not — `build-desktop.yml:104` sets `fail-fast: false`. The GitHub check annotations are explicit:

```
job 96151540633 (ubuntu):                failure: The job has exceeded the maximum execution time of 1h30m0s
job 96151540563 (aarch64-apple-darwin):  failure: The job has exceeded the maximum execution time of 1h30m0s
```

Durations against `build-desktop.yml:101` (`timeout-minutes: 90`):

| Job | Duration | Result |
|---|---|---|
| `windows` | **1h29m51s** | success — **9 seconds** of margin |
| `ubuntu` | 1h30m22s | timed out |
| `aarch64-apple-darwin` | 1h30m47s | timed out |
| `x86_64-apple-darwin` | 1h30m48s | timed out |

The desktop matrix is sitting exactly on the 90-minute ceiling; Windows passed by nine seconds. Even with the Docker fix and a healthy arm64 runner, **this run would still have failed**, and the next one is a coin flip. This is independent of both causes above and needs its own decision (raise `timeout-minutes`, or cut desktop build time).

---

## Are the failures one cause or several?

**Three independent causes.** Evidence:

- The Docker job and the desktop matrix do not share a checkout step — Docker uses the hardcoded list (`release-production.yml:404`), desktop uses `submodules: recursive` (`build-desktop.yml`). No shared upstream step failed; `Prepare build context` and `Prepare GitHub release` both succeeded.
- No shared toolchain or base-image change is implicated — the Docker failure is a missing directory, the arm64 failure is a runner TERM with no build error, the other three are wall-clock timeouts.
- `fail-fast: false` rules out cascade between the matrix legs.

---

## New or longstanding?

**The Docker failure is new, introduced between 2026-08-13 and 2026-08-19.**

| Run | Date | Docker job |
|---|---|---|
| [31713499368](https://github.com/tinyhumansai/openhuman/actions/runs/31713499368) | 2026-08-13 | ✅ success |
| [32262998212](https://github.com/tinyhumansai/openhuman/actions/runs/32262998212) | 2026-08-19 14:16 | ❌ `tinyhosts` |
| [32278391834](https://github.com/tinyhumansai/openhuman/actions/runs/32278391834) | 2026-08-19 16:52 | ❌ `tinyhosts` |

At the last-good sha `0d65013c` neither `tinyhosts` nor `tinywallet` was yet a path dependency in `Cargo.toml` — verified, `grep '^(tinywallet|tinyhosts) *='` returns nothing at that ref. That is exactly why Docker passed then.

The two crates landed on `release` after it:

- `6f3b511b` (2026-08-13) — `refactor(web3): re-vendor tinywallet instead of inlining its source`
- `6051d940` (2026-08-17) — `feat(config): add hosting feature with tinyhosts vendor dependency`

Neither commit updated the init list. The precedent exists in-tree: `54f8710f` (2026-08-10) — *"chore(ci): add tinymemory submodule to workflows"* — was exactly this maintenance step, done correctly that time. The list was last touched 2026-08-13 (`4e4c8ffc`).

**Not a runner-image change and not a dependency bump** — it is a repo-local list that was not kept in sync.

---

## Blast radius

**Production releases are blocked.** No `Release Production` run has succeeded since 2026-08-07 ([31200898582](https://github.com/tinyhumansai/openhuman/actions/runs/31200898582)); the three attempts since (08-13, and both on 08-19) all failed. Staging is blocked by the same defect.

**Nothing was half-published. No cleanup is required.**

- This run was dispatched with `create_release: false` — the `Prepare GitHub release` job took the no-op branch and emitted an empty `release_id` (`echo "release_id=" >> "$GITHUB_OUTPUT"`). **No GitHub release and no tag were created**, so there is nothing dangling. `Remove release and tag if build failed` was correctly `skipped` (its `if:` requires `inputs.create_release`), not silently missed.
- `CLI: ${{ matrix.target }}` was `skipped` for the same reason — no release assets uploaded.
- **No Docker image was pushed.** The buildx build failed at the final builder stage, before any export/push; `Verify pushed image is pullable` never ran. `Docker: tag :latest` was `skipped`, so the moving `:latest` tag was not touched — which is exactly the safeguard the comment at `release-production.yml` Phase 3b describes, working as designed. *(Not independently confirmed against the GHCR registry — the available token lacks `read:packages`. Confidence is high from the build logs, but flagged as **unverified**.)*
- A desktop matrix entry is missing (`windows` succeeded, the other four did not), but with `with_release_upload: false` those artifacts had no publication target.

Net: a clean failure, not a partial release. The risk is a **future** run dispatched with `create_release: true` — that would cut a tag and a draft release, then fail on Docker and depend on the cleanup job to roll it back.

---

### Required (unblocks both pipelines)

Add both missing submodules to the two lists — `release-production.yml:404` and `release-staging.yml:300`, on both `main` and `release`:

```yaml
- name: Init vendored Rust submodules
  run: git submodule update --init --recursive vendor/tinyagents vendor/tinyflows vendor/tinycortex vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk vendor/tinybus vendor/tinymemory vendor/tinywallet vendor/tinyhosts
```

Adding only `tinyhosts` leaves the build failing on `tinywallet`.

### Trade-off: keep the list, or stop maintaining one

**Option A — patch the list (above).** Smallest diff, preserves the intent of skipping anything heavy. Cost: this is the third time the list has drifted; it will drift again on the next vendored crate, and it fails *at release time* rather than in PR CI.

**Option B — drop the list, use `submodules: recursive`** in the Docker job's checkout, matching `build-desktop.yml`. Removes the drift class entirely. Cost: checks out every submodule. The stated reason for not doing this is *"we skip the large tauri-cef fork the core image doesn't need"* — but `.gitmodules` on `release` declares 10 submodules and **none of them is `tauri-cef`**, and the desktop jobs already run `submodules: recursive` without a checkout blowup. The comment's rationale looks stale. *(**Unverified** — `tauri-cef` still has ~28 code references in the repo, so please confirm it is genuinely no longer a submodule before relying on this.)*

**Option C — derive the list, or guard it.** Generate the init list from `.gitmodules` minus an explicit denylist, or add a cheap PR-CI check asserting that every `path = "vendor/..."` dependency in `Cargo.toml` appears in both workflow lists. Highest up-front cost, but it is the only option that catches the *next* occurrence at PR time instead of at release time.

**Recommendation: Option A now to unblock the release, then Option C as follow-up.** Option B is attractive but should not be taken until the `tauri-cef` question is settled.

### Separately

- **`ubuntu-arm64`:** re-run. No code change. If it reproduces a third time, open an infrastructure issue against the `ubuntu-24.04-arm` pool.
- **90-minute timeout:** raise `timeout-minutes` in `build-desktop.yml:101` (a bump to 120 restores headroom) or reduce desktop build time. Without this, the release can still fail after the Docker fix — Windows is passing with nine seconds to spare.

---

## Links

- Failing run: https://github.com/tinyhumansai/openhuman/actions/runs/32278391834
- Same defect, earlier the same day: https://github.com/tinyhumansai/openhuman/actions/runs/32262998212
- Staging, same defect: https://github.com/tinyhumansai/openhuman/actions/runs/32260876335
- Last successful production release: https://github.com/tinyhumansai/openhuman/actions/runs/31200898582 (2026-08-07)
- Last run with a passing Docker job: https://github.com/tinyhumansai/openhuman/actions/runs/31713499368 (2026-08-13)

M3gA-Mind:
## Summary

- The release Docker jobs initialised submodules from a **hand-maintained 8-entry list** while `.gitmodules` declares **10**, so `vendor/tinyhosts` and `vendor/tinywallet` were never checked out and cargo could not resolve two path dependencies.
- **Production releases have been blocked since 2026-08-07**; staging carries a byte-identical list and is broken the same way. Both are fixed here.
- The enumeration is **removed**, not corrected: nothing remains that it could legitimately exclude, so a subset is always a bug.
- ⚠️ **Release-blocking — this needs to reach `release`.** It targets `main` per the fleet default; a maintainer will need to promote or cherry-pick it, since `release` is where the blocked pipeline actually runs.

## Problem

The Dockerfile does `COPY vendor/ vendor/`, so an uninitialised submodule arrives in the image as an **empty directory**. Cargo then fails on the missing manifest before compiling anything:

```
error: failed to get `tinyhosts` as a dependency of package `openhuman v0.63.12 (/build)`
Caused by: failed to read `/build/vendor/tinyhosts/Cargo.toml`
Caused by: No such file or directory (os error 2)
```

`optional = true` does not exempt these: cargo reads every `path` dependency's manifest to build the lockfile whether or not the feature is enabled. The failing build proves it — `tinyhosts` is declared `optional = true` and still broke resolution.

**Fixing only `tinyhosts` would not have worked.** Cargo stops at the first unresolvable dependency, and `tinywallet` (`Cargo.toml:524`) sits behind `tinyhosts` (`:493`). Both were missing; both are added.

## Solution

I removed the list rather than adding two names to it, and the deciding evidence is that **nothing is left for it to exclude**:

1. **The stated rationale is gone.** The comment justified the subset as skipping *"the large tauri-cef fork the core image doesn't need"*. `tauri-cef` stopped being a submodule in `1843706c3 refactor(tauri): replace CEF runtime with upstream Wry`. It is not in `.gitmodules` today, so a recursive init cannot pull it — the ~28 remaining `tauri-cef` references in the tree are docs, scripts and TS, plus exactly one `Cargo.toml` line that is a *comment*, not a dependency.
2. **All ten declared submodules are required.** Every entry in `.gitmodules` is a `path` dependency of the root crate. The only other `vendor/` path dep, `motosan-ai-oauth`, is a committed directory rather than a submodule (`git ls-files` lists its sources), so it is unaffected either way. A correct list is therefore *always* all ten — the enumeration had no legitimate degrees of freedom left, only the ability to be wrong.
3. **Nothing heavy gets dragged in.** The two missing crates are **508K** and **836K**, and their only nested submodule is `tinybus`, which the old list already named.
4. **It had already drifted three times**, each surfacing at *release* time rather than in PR CI — `tinymemory` (fixed in `54f8710f`), then `tinywallet` and `tinyhosts` together (this issue). Correcting it a third time would schedule a fourth.

This also makes the proposed "add a CI guard that asserts the list matches `Cargo.toml`" follow-up unnecessary for this class: there is no list left to drift.

Both files get the identical change, so staging and production cannot diverge again.

## Verification — the pipeline is the only real test

**Stated plainly: a `.github/workflows` change cannot be proven locally.** It only executes on a `workflow_dispatch` of the release pipelines, which I am not authorised to run, and this repo's rule is to verify Rust through CI rather than building locally.

What I could check, I did:

- Both files **parse as YAML**, and the changed step resolves to exactly `git submodule update --init --recursive` in the `build-docker` job of each (11 jobs parsed in production, 5 in staging).
- `.gitmodules` declares 10 paths; the old list named 8; the two missing names are exactly `vendor/tinyhosts` and `vendor/tinywallet`.
- `git submodule update --init vendor/tinyhosts vendor/tinywallet` succeeds and both contain a `Cargo.toml` — so the checkout the workflow will now perform does produce the manifests cargo was looking for.
- `tauri-cef` absent from `.gitmodules`, and the commit that removed it identified.

The honest confirmation is the next `Release Staging` dispatch reaching the Docker build step. **Nothing in this PR is exercised by PR CI**, so a green run here is not evidence the fix works.

## Submission Checklist

- [x] N/A: this is a `.github/workflows` change with no application code; there is no unit-testable surface, and the only execution path is a release dispatch. — Tests added or updated (happy path + at least one failure / edge case)
- [x] N/A: no changed lines are Rust or TS, so `diff-cover` has nothing to measure on this diff. — **Diff coverage ≥ 80%**
- [x] N/A: behaviour-only CI change; no feature rows added, removed or renamed. — Coverage matrix updated
- [x] N/A: no feature IDs are touched by a workflow submodule-init change. — All affected feature IDs listed under `## Related`
- [x] No new external network dependencies introduced — the two submodules are already declared in `.gitmodules` and already fetched by every other job that uses `submodules: recursive`; this adds no new remote.
- [x] N/A: no release-cut *surface* changes — the smoke checklist covers what a release produces, and this changes only how the builder checks out its sources. (It does unblock the release from running at all.) — Manual smoke checklist updated

## Impact

**Platform:** Linux/Docker (`openhuman-core` image) on the production and staging release pipelines. No runtime, application or user-visible change — nothing ships differently, the builder simply checks out the sources it already required.

**Performance:** two additional submodules and their nested `tinybus` (already fetched). 508K + 836K, negligible against the image build.

**Security / compatibility / migration:** none. No new remotes, no dependency version changes, no lockfile change.

**Risk if wrong:** contained to the release pipeline, which is already failing — this cannot regress a passing state.

## Related

- Closes: #5594
- Follow-up PR(s)/TODOs: #5595 — the desktop matrix hits `timeout-minutes: 90` on macOS and Linux while Windows passes with **9 seconds** of margin, so a release still cannot go green on this fix alone. Filed separately with the measurement that the researcher's write-up did not have: desktop build time roughly **doubled** between 08-13 and 08-19 (Windows, the uncensored control leg: 51m58s → 89m01s → 89m51s), so the ceiling is the symptom rather than the cause.

---

### Linear Issue

- Key: N/A — filed from GitHub issue #5594, no Linear ticket.
- URL: N/A — see https://github.com/tinyhumansai/openhuman/issues/5594

### Commit & Branch

- Branch: `fix/5594-release-submodule-list`
- Commit SHA: `fe5bcb224`

### Validation Run

- [x] N/A: no files under `app/` are touched. — `pnpm --filter openhuman-app format:check`
- [x] N/A: no TypeScript is touched; the diff is two workflow YAML files. — `pnpm typecheck`
- [x] N/A: there is no test that exercises a workflow file. Focused checks run instead: both YAMLs parse and the changed step resolves to the intended command in each `build-docker` job. — Focused tests
- [x] N/A: no Rust changed. Building openhuman locally is also disallowed in this repo; Rust is verified through CI. — Rust fmt/check (if changed)
- [x] N/A: no Tauri/`src-tauri` files changed. — Tauri fmt/check (if changed)

### Validation Blocked

- `command:` the release pipelines (`Release Staging` / `Release Production`, `workflow_dispatch`)
- `error:` not run — dispatching a release is a maintainer action, and PR CI does not execute these workflows
- `impact:` the fix is unverified end-to-end until a release dispatch reaches the Docker build step. The failure mode if it is still wrong is identical and immediate (cargo cannot read a `vendor/*/Cargo.toml`), so it fails loudly rather than shipping a bad image.

### Behavior Changes

- Intended behavior change: the release Docker jobs initialise every declared submodule instead of eight named ones.
- User-visible effect: none directly. Indirectly, production and staging releases can build again.

### Parity Contract

- Legacy behavior preserved: yes for the eight previously-named submodules — they are still initialised recursively, by the same command in the same step. The change is purely additive in effect (8 → 10).
- Guard/fallback/dispatch parity checks: production and staging receive byte-identical edits, which is what stops the two pipelines drifting apart again — they were byte-identical before and are byte-identical after.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A — no duplicates found for #5594.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved production and staging release builds by ensuring all required Rust components are initialized automatically.
  * Removed outdated exclusions that could cause release builds to fail when new components were added.

### DIFF
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -397,11 +397,26 @@ jobs:
         with:
           ref: ${{ needs.prepare-build.outputs.build_ref }}
           fetch-depth: 1
-      # Targeted init (not `submodules: true`) so we skip the large tauri-cef
-      # fork the core image doesn't need. The Dockerfile COPYs vendor/ because
-      # [patch.crates-io] resolves Rust SDK crates from vendor/.
+      # Every submodule, not an enumerated subset. The Dockerfile COPYs vendor/
+      # because [patch.crates-io] resolves Rust SDK crates from vendor/, and an
+      # uninitialised submodule is copied in as an EMPTY directory — cargo then
+      # fails on the missing manifest before it compiles anything.
+      #
+      # This used to name eight paths, to skip "the large tauri-cef fork the
+      # core image doesn't need". That fork stopped being a submodule in
+      # `1843706c3 refactor(tauri): replace CEF runtime with upstream Wry`, so
+      # there is nothing left for the list to exclude: all ten entries in
+      # `.gitmodules` are `path` dependencies of the root crate, and cargo reads
+      # every one of their manifests to resolve, `optional` or not. A subset is
+      # therefore always a bug, which is what the list kept producing — it went
+      # stale three times (tinymemory in `54f8710f`, then tinywallet and
+      # tinyhosts together, issue #5594), and each time the breakage surfaced at
+      # RELEASE time rather than in PR CI.
+      #
+      # Enumerating buys nothing once nothing can legitimately be excluded, so
+      # the enumeration is gone rather than corrected a third time.
       - name: Init vendored Rust submodules
-        run: git submodule update --init --recursive vendor/tinyagents vendor/tinyflows vendor/tinycortex vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk vendor/tinybus vendor/tinymemory
+        run: git submodule update --init --recursive
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
       - name: Log in to GHCR
--- a/.github/workflows/release-staging.yml
+++ b/.github/workflows/release-staging.yml
@@ -293,11 +293,26 @@ jobs:
         with:
           ref: ${{ needs.prepare-build.outputs.build_ref }}
           fetch-depth: 1
-      # Targeted init (not `submodules: true`) so we skip the large tauri-cef
-      # fork the core image doesn't need. The Dockerfile COPYs vendor/ because
-      # [patch.crates-io] resolves Rust SDK crates from vendor/.
+      # Every submodule, not an enumerated subset. The Dockerfile COPYs vendor/
+      # because [patch.crates-io] resolves Rust SDK crates from vendor/, and an
+      # uninitialised submodule is copied in as an EMPTY directory — cargo then
+      # fails on the missing manifest before it compiles anything.
+      #
+      # This used to name eight paths, to skip "the large tauri-cef fork the
+      # core image doesn't need". That fork stopped being a submodule in
+      # `1843706c3 refactor(tauri): replace CEF runtime with upstream Wry`, so
+      # there is nothing left for the list to exclude: all ten entries in
+      # `.gitmodules` are `path` dependencies of the root crate, and cargo reads
+      # every one of their manifests to resolve, `optional` or not. A subset is
+      # therefore always a bug, which is what the list kept producing — it went
+      # stale three times (tinymemory in `54f8710f`, then tinywallet and
+      # tinyhosts together, issue #5594), and each time the breakage surfaced at
+      # RELEASE time rather than in PR CI.
+      #
+      # Enumerating buys nothing once nothing can legitimately be excluded, so
+      # the enumeration is gone rather than corrected a third time.
       - name: Init vendored Rust submodules
-        run: git submodule update --init --recursive vendor/tinyagents vendor/tinyflows vendor/tinycortex vendor/tinychannels vendor/tinyplace vendor/tinyhumans-sdk vendor/tinybus vendor/tinymemory
+        run: git submodule update --init --recursive
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
       - name: Build image (no push)


M3gA-Mind:
@coderabbitai review

Issue #5613, opened 2026-08-20
ci: the Rust coverage gate reports success after running zero tests


M3gA-Mind:
## Summary

The Rust coverage lane reports success when it has verified **nothing**. Two independent fail-opens compose:

| | Where | Behaviour |
|---|---|---|
| **A** | `scripts/ci/rust-coverage-changed.sh` | a scoped run whose libtest filter matches nothing prints `running 0 tests … ok` and exits 0 |
| **B** | `ci-lite.yml` "Enforce >= 80% coverage on changed lines" | `diff-cover` finds no coverage rows, reports *"No lines with coverage information in this diff"*, exits 0 |

Both read the **absence** of data as "nothing to check" rather than "we checked nothing".

## Evidence

PR #5593 added the `hosting` family — 1,643 lines including a 511-line test file. Its `Rust Core Coverage (cargo-llvm-cov)` job [reported success](https://github.com/tinyhumansai/openhuman/actions/runs/32268416328) with, verbatim:

```
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 12202 filtered out; finished in 0.00s
```

The feature is behind a Cargo gate that was in neither `[features] default` nor `scripts/ci/product-features.txt`, so the lane compiled none of it, the scoped filter matched nothing, and diff-cover had no rows to score. It merged green.

## Why fixing only A is insufficient

A PR touching **both** `src/openhuman/hosting/tools.rs` and `src/openhuman/config/ops.rs` runs config's tests (count > 0), leaves hosting entirely unverified, and diff-cover happily scores ≥80% on the config lines alone. The test-count check passes and the coverage gate passes, while a whole feature went unbuilt.

## Suggested fix

A narrow, verifiable assertion rather than a broad one:

> Every changed Rust source file that *should* have been compiled into the coverage build must appear as an `SF:` record in the emitted lcov.

Deliberately **not** "tests ran" and **not** "changed lines are covered" — `diff-cover --fail-under=80` already owns the latter. "No rows at all" is a build-configuration fact; "too few covered lines" is a judgement call with an existing owner. Separating them is what keeps the new gate free of false positives.

### Validated offline against a real artifact

Replayed against the `lcov-rust-core` artifact from [run 32108672413](https://github.com/tinyhumansai/openhuman/actions/runs/32108672413) (26 MB, 1,354 `SF:` records):

```
files the rule would CHECK : 1340
files the rule would FLAG  : 2
[...]

Issue #5598, opened 2026-08-19
Staging: tinymemory capability mismatch (8191 vs 262143) — memory_tree, memory_store_raw_chunks, memory_diff all failing


Al629176:
## Summary

The `tinymemory` module shipped in the staging build advertises `Capabilities { bits: 8191 }` but the host build expects `Capabilities { bits: 262143 }`. The registry pin and the vendored artifact have diverged. As a direct result, three memory tools fail with `UnknownMethod` errors on every call.

## Log evidence (staging, 2026-08-19)

```
02:08:48 WRN host module rustc differs from host module=tinymemory-module

02:08:48 WRN [modules:memory] the module advertises Capabilities { bits: 8191 } but this build assumes
Capabilities { bits: 262143 }; the registry pin and the artifact have diverged

02:08:48 WRN [tinyagents] tool failed tool=memory_tree
error=memory_tree_cover_window: ai.tinyhumans.tinybus.Error.UnknownMethod: ai.tinyhumans.tinymemory.Memory: no member `CoverWindow`

02:08:53 WRN [tinyagents] tool failed tool=memory_store_raw_chunks
error=ai.tinyhumans.tinybus.Error.UnknownMethod: ai.tinyhumans.tinymemory.Memory: no member `ListChunks`

02:09:03 WRN [tinyagents] tool failed tool=memory_diff
error=diff_since_read: no snapshots found for this source
```

## Impact

All three memory tools (`memory_tree`, `memory_store_raw_chunks`, `memory_diff`) are non-functional in this staging build. Any agent behaviour that relies on memory tree traversal, raw chunk storage, or diff-since-read is broken.

## Fix direction

Re-pin the `tinymemory` module artifact to match the capability bits expected by this build (`262143`). The `rustc` host mismatch suggests the module was compiled against a different toolchain — confirm the CI pipeline uses the same Rust toolchain version for both the host and the vendored submodule.

PR #5616, opened 2026-08-20
fix(ci): run the Rust lanes when any vendored crate moves

### LINKED ISSUE #5612: ci: a vendored submodule bump skips every Rust lane and still reports PR CI Gate success
## Summary

A vendored-submodule pointer bump runs **no openhuman Rust lane at all**, while `PR CI Gate` still reports success. The current tip of `main` has therefore not been compiled by CI, and `main` is what gets promoted to `release`.

## Evidence

`92bab8df` ("bump tinycortex", merged 2026-08-20) changes exactly one thing — the `vendor/tinycortex` pointer. [CI Lite run 32358998415](https://github.com/tinyhumansai/openhuman/actions/runs/32358998415):

```
success   TinyCortex Memory Tests
success   Feature Forwarding Gate
success   PR CI Gate                          <- the aggregate gate
skipped   Rust Quality (fmt, clippy)
skipped   Rust Core Coverage (cargo-llvm-cov)
skipped   Rust Feature-Gate Smoke (gates off)
skipped   Rust Tauri Coverage (cargo-llvm-cov)
skipped   Frontend Checks
skipped   Test Inventory
skipped   Scripts Self-Tests
```

## Cause

The `changes` job's path filters enumerate vendored paths, and the list is stale:

- `rust-core` names `vendor/motosan-ai-oauth/**` and `vendor/tinychannels` — **2 of 10** submodules, and `motosan-ai-oauth` is an in-tree directory, not a submodule.
- `rust-core-full` names the same two.
- `rust-tauri` names **none**, despite `app/src-tauri` being a second cargo world with its own duplicate `[patch.crates-io]` table pointing at the same submodules.

The `tinycortex` filter that *did* match feeds only TinyCortex's own test lane (`cargo test --manifest-path vendor/tinycortex/Cargo.toml`), which is why that ran while nothing compiled openhuman against the new pin.

Every vendored crate is force-resolved by path through `[patch.crates-io]` in **both** cargo worlds, so any pointer bump changes what the core and the shell compile against.

## Why an enumeration is the wrong shape here

This is the same defect the release Docker job had. Its hand-maintained submodule list went stale **three times** before `fe5bcb22` (#5596) replaced it with `git submodule update --init --recursive`; that commit message says so. A list of ten paths here will fail the same way on the eleventh.

## Impact

Eight of the ten vendored crates could move with zero verification. Combined with #5598 (a pinned artifact diverging from the submodule the host compiles against), a bump is exactly the change most likely to break the build and least likely to be checked.

## Suggested fix

Match on `vendor/**` and `.gitmodules` in `rust-core`, `rust-core-full` and `rust-tauri`. A bump should also force full-suite mode rather than scoped tests, for the same reason a `Cargo.lock` change does — it moves the dependency graph under every module.

Cost: a vendored bump runs the full Rust lanes rather than nothing.

M3gA-Mind:
## Summary

- Makes any vendored-crate pointer bump run the openhuman Rust lanes. Today it runs none of them while `PR CI Gate` still reports success.
- Replaces an enumeration of vendored paths with `vendor/**`, because the enumeration is what went stale.

## Problem

`92bab8df` ("bump tinycortex") changes exactly one thing — the `vendor/tinycortex` pointer. [CI Lite run 32358998415](https://github.com/tinyhumansai/openhuman/actions/runs/32358998415):

```
success   TinyCortex Memory Tests
success   PR CI Gate                          <- the aggregate gate
skipped   Rust Quality (fmt, clippy)
skipped   Rust Core Coverage (cargo-llvm-cov)
skipped   Rust Feature-Gate Smoke (gates off)
skipped   Rust Tauri Coverage (cargo-llvm-cov)
```

That commit is the current tip of `main`, so **the tip of main has not been compiled by CI**, and `main` is what `promote-main-to-release.yml` moves to `release`.

The `changes` filters enumerate vendored paths and the list is stale: `rust-core` and `rust-core-full` name `vendor/motosan-ai-oauth/**` and `vendor/tinychannels` — 2 of 10 submodules, and `motosan-ai-oauth` is an in-tree directory rather than a submodule. `rust-tauri` names none at all. The `tinycortex` filter that did match feeds only TinyCortex's own test lane, which is why that ran while nothing compiled openhuman against the new pin.

Every vendored crate is force-resolved by path through `[patch.crates-io]` in **both** cargo worlds, so any bump changes what the core and the shell compile against.

## Solution

Match `vendor/**` and `.gitmodules` in `rust-core`, `rust-core-full` and `rust-tauri`.

Deliberately **not** a list of the ten current submodules. The release Docker job had this exact defect: its hand-maintained list went stale three times before `fe5bcb22` (#5596) replaced it with `--init --recursive`, and that commit message says so. A list here fails the same way on the eleventh crate.

`rust-core-full` gets them too — a dependency-graph change invalidates per-module test scoping for the same reason a `Cargo.lock` change does.

**Cost:** a vendored bump now runs the full Rust lanes rather than nothing. That is the intent; eight of the ten crates could previously move with zero verification.

## Submission Checklist

- [x] Tests added or updated — `N/A: CI path-filter configuration.` The behaviour under change is which jobs a given diff triggers, which is only observable in a real CI run; the reproduction is quoted above and the fix is verifiable by pushing a vendor-only diff.
- [x] **Diff coverage ≥ 80%** — `N/A: no executable lines changed.` YAML filter globs and comments.
- [x] Coverage matrix updated — `N/A: no feature added, removed or renamed.`
- [x] All affected feature IDs listed under `## Related` — `N/A: no matrix feature IDs touched.`
- [x] Manual smoke checklist updated — `N/A: does not touch a release-cut surface.`

## Impact

- **CI cost:** vendored bumps become expensive rather than free. Given #5598 — a pinned artifact diverging from the submodule the host compiles against — a bump is the change most likely to break the build and least likely to be checked.
- **Risk:** low. Widening a path filter can only cause more jobs to run, never fewer.

## Related

- Closes: #5612
- Follow-up PR(s)/TODOs: the sibling fail-open, where the coverage lane reports success after running zero tests, is #5613.

---

### Linear Issue
- Key: N/A
- URL: N/A

### Commit & Branch
- Branch: N/A
- Commit SHA: N/A

### Validation Run
- [x] `pnpm --filter openhuman-app format:check` — N/A: no formatted source changed (workflow YAML only).
- [x] `pnpm typecheck` — N/A: no TypeScript changed.
- [x] Focused tests: N/A: no test targets a path filter.
- [x] Rust fmt/check (if changed): N/A: no Rust changed.
- [x] Tauri fmt/check (if changed): N/A: no Tauri source changed.

### Validation Blocked
- `command:` observing the filter fire on a vendored bump
- `error:` not reproducible locally — `dorny/paths-filter` evaluates against the PR's changed-file set on GitHub
- `impact:` verified by reading the filters against the observed skip in run 32358998415; the first vendored bump after merge is the live confirmation.

### Behavior Changes
- Intended behavior change: a diff touching `vendor/**` or `.gitmodules` now triggers the Rust lanes in full-suite mode.
- User-visible effect: none.

### Parity Contract
- Legacy behavior preserved: all existing filter entries are unchanged; this only adds.
- Guard/fallback/dispatch parity checks: `vendor/**` is a superset of the two paths previously named, so nothing that triggered before stops triggering.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated CI change detection to recognize all vendored dependency updates.
  * Ensured relevant Rust core, full-suite, and Tauri checks run when vendored code or submodule references change.
  * Prevented changes to unlisted vendored components from incorrectly bypassing validation.

### DIFF
--- a/.github/workflows/ci-lite.yml
+++ b/.github/workflows/ci-lite.yml
@@ -154,8 +154,16 @@ jobs:
               - 'scripts/check-linux-tls-dependencies.sh'
               - 'scripts/test-rust-e2e.sh'
               - 'scripts/test-rust-with-mock.sh'
-              - 'vendor/motosan-ai-oauth/**'
-              - 'vendor/tinychannels'
+              # Every vendored crate, not an enumeration. `[patch.crates-io]`
+              # force-resolves all ten submodules by path, so ANY pointer bump
+              # changes what this crate compiles against. The previous list
+              # named two of them, so a bump to the other eight ran no openhuman
+              # Rust lane at all while PR CI Gate still reported success.
+              # Enumerating here failed the same way the release Docker
+              # submodule list did three times before #5596 replaced it with
+              # `--init --recursive`. Do not go back to a list.
+              - '.gitmodules'
+              - 'vendor/**'
             # Changes here invalidate per-module test scoping → full suite.
             rust-core-full:
               - '.github/workflows/ci-lite.yml'
@@ -167,8 +175,11 @@ jobs:
               - 'rust-toolchain.toml'
               - 'scripts/ci-cancel-aware.sh'
               - 'scripts/check-linux-tls-dependencies.sh'
-              - 'vendor/motosan-ai-oauth/**'
-              - 'vendor/tinychannels'
+              # A vendored-crate bump invalidates per-module test scoping for the
+              # same reason a Cargo.lock change does: it moves the dependency
+              # graph under every module. Full suite, not scoped.
+              - '.gitmodules'
+              - 'vendor/**'
             rust-core-src:
               - 'src/**'
               - 'tests/**'
@@ -185,6 +196,11 @@ jobs:
               - 'rust-toolchain.toml'
               - 'app/src-tauri/**'
               - 'scripts/ci-cancel-aware.sh'
+              # The shell is a second Cargo world with its own duplicate
+              # `[patch.crates-io]` table pointing at the same submodules, so it
+              # is affected by a bump exactly as the core is. It named none.
+              - '.gitmodules'
+              - 'vendor/**'
             # Shared mock backend + script self-tests. The mock backend every
             # E2E suite depends on had zero PR-lane coverage before this (its
             # own socket-auth tests never ran in CI) — arm the scripts


M3gA-Mind:
### Verification follow-up: the glob was reproducible locally after all

The body lists this under **Validation Blocked** ("not reproducible locally"). The part that actually
decides whether this PR works *is* checkable without CI, so I checked it.

`dorny/paths-filter@v4` (`ci-lite.yml:90`) matches with **picomatch**, and a submodule pointer bump is
reported by git/GitHub as a single path with no trailing slash — `gh api …/commits/72823015 --jq '.files[]'`
returns `modified vendor/tinyflows`. So the question is whether `vendor/**` matches a bare gitlink.

Run against both picomatch majors in the repo's pnpm store (2.3.2 and 4.0.4), `{ dot: true }`:

| path | `vendor/**` | old `vendor/tinychannels` | old `vendor/motosan-ai-oauth/**` |
|---|---|---|---|
| `vendor/tinycortex` (the reported case, 92bab8df) | **match** | – | – |
| `vendor/tinyflows` (72823015) | **match** | – | – |
| `vendor/tinychannels` | **match** | match | – |
| `vendor/motosan-ai-oauth/Cargo.toml` | **match** | – | match |
| `vendor/motosan-ai-oauth/src/lib.rs` | **match** | – | match |
| `.gitmodules` | no match | – | – |
| `src/core/all.rs` (control) | no match | – | – |

Three things this pins down:

1. `vendor/**` fires on bare gitlinks, so it covers the exact case the PR targets.
2. It is a strict superset of both patterns it replaces — the parity claim in the body holds, nothing
   that triggered before stops triggering.
3. `.gitmodules` genuinely needs its own entry; `vendor/**` does not cover it. Both are present.

Corroborating live evidence: the pre-existing `tinycortex` filter (`ci-lite.yml:180`, also a bare
`vendor/tinycortex`) *did* fire on 92bab8df — `TinyCortex Memory Tests` ran while every Rust lane
skipped. That is independent proof paths-filter sees the gitlink path.

Also checked the one risk the widening introduces: every lane this can newly trigger
(`rust-quality`, `rust-feature-gate-smoke`, `rust-rss-bench`, `rust-core-coverage`,
`rust-tauri-coverage`) already checks out with `submodules: recursive`, so a vendor-triggered run has
the vendored code present.

No code change — evidence only.

### Adjacent gap, deliberately not folded in

The `frontend` filter has no `vendor/` entry, yet `app/src/services/__tests__/rpcMethods.test.ts:169`
reads `vendor/tinychannels/src/controllers/schemas.rs` off disk at test time — the lane even carries a
dedicated `git submodule update --init vendor/tinychannels` step (`ci-lite.yml:250-254`). So a
tinychannels bump that changes a controller schema can still skip the drift guard written to catch it.
Same class as this PR, different lane. Leaving it out to keep the scope honest; worth a follow-up
issue alongside #5613.

PR #5617, opened 2026-08-20
fix(agent): rename spawn_parallel graph topology to resolve dual kind collision (#5601)

### LINKED ISSUE #5601: Staging: spawn_parallel_agents registered under both 'tool' and 'graph' kinds — ambiguous dispatch
## Summary

`spawn_parallel_agents` is registered under two kinds simultaneously — both `tool` and `graph` — in the tinyagents registry. This is flagged as a non-fatal diagnostic but callers cannot reliably disambiguate which kind is invoked, which can cause incorrect dispatch at runtime.

## Log evidence (staging, 2026-08-19)

```
02:08:43 WRN [tinyagents][registry] non-fatal diagnostic: name `spawn_parallel_agents` is registered
under multiple kinds (tool, graph); ensure callers disambiguate by kind
kind="tool" name=spawn_parallel_agents

[Repeats on each agent turn]
```

## Impact

Non-fatal but creates ambiguous dispatch. If a caller resolves `spawn_parallel_agents` by name without specifying kind, it may get either the tool or graph variant non-deterministically. Could manifest as inconsistent parallel agent spawning behaviour that's hard to reproduce.

## Fix direction

Determine whether `spawn_parallel_agents` should be a `tool` or a `graph` and remove the duplicate registration. If both are intentional, enforce explicit kind disambiguation at all call sites.

shivakumar777-web:
Closes #5601

### Summary of Changes
- Renamed exported structural graph topology from `spawn_parallel_agents` to `spawn_parallel_graph` in `src/openhuman/agent/tinyagents/topology.rs`.
- Resolves ambiguous dispatch warnings in `CapabilityRegistry` caused by tool/graph name collision on `spawn_parallel_agents`.
- Added collision-prevention regression test `all_topologies_names_do_not_collide_with_production_tools` in `topology_tests.rs` ensuring no graph topology name clashes with registered production tools.

### Verification
- `cargo test --manifest-path Cargo.toml --lib openhuman::agent::tinyagents::topology` (passed)
- `cargo check --manifest-path Cargo.toml` (passed)

### Checklist
- [x] Code formatting (`cargo fmt`) and clippy linter pass locally.
- [x] Unit tests covering topology name collision added to `topology_tests.rs`.
- [x] N/A - Documentation updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the reported name of the spawn-parallel graph topology to `spawn_parallel_graph`, improving consistency when viewing or consuming topology reports.
  * Prevented graph topology names from conflicting with registered tool names.

* **Tests**
  * Added coverage confirming the topology is exported, structurally valid, renderable as Mermaid, and uniquely named.

### DIFF
--- a/src/openhuman/agent/tinyagents/topology.rs
+++ b/src/openhuman/agent/tinyagents/topology.rs
@@ -66,7 +66,7 @@ pub(crate) fn all_graph_topologies() -> Vec<GraphTopologyReport> {
         crate::openhuman::agent::orchestration::spawn_parallel_graph::spawn_parallel_graph_topology(
         )
     {
-        out.push(describe("spawn_parallel_agents", &t));
+        out.push(describe("spawn_parallel_graph", &t));
     }
 
     // Not exported: generic item-count-driven `map_reduce` fan-outs whose node
--- a/src/openhuman/agent/tinyagents/topology_tests.rs
+++ b/src/openhuman/agent/tinyagents/topology_tests.rs
@@ -20,7 +20,11 @@ fn all_topologies_includes_the_member_graph() {
 #[test]
 fn all_topologies_includes_delegation_and_workflow_scheduler() {
     let reports = all_graph_topologies();
-    for name in ["delegation", "workflow_runs:scheduler"] {
+    for name in [
+        "delegation",
+        "workflow_runs:scheduler",
+        "spawn_parallel_graph",
+    ] {
         let report = reports
             .iter()
             .find(|r| r.name == name)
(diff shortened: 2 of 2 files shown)


M3gA-Mind:

**The fix is correct and I verified it actually reaches the thing #5601 complains about.** One mechanical conflict stands between this and mergeable.

## Verified, so you don't have to argue it again

I checked the chain rather than taking the rename on trust, because renaming a *report* label would be worthless if the registry were fed from somewhere else:

- `mod_part_04.rs:273-274` does `for report in all_graph_topologies() { capability_registry.register_descriptor(ComponentKind::Graph, report.name) }` — so the string in `describe(...)` **is** the name registered under `ComponentKind::Graph`. Renaming it genuinely removes the duplicate registration the staging log flagged; this is not a cosmetic rename.
- Renaming the **graph** rather than the tool is the right side to change. `spawn_parallel_agents` as a *tool* is a prompt/wire contract — it appears in `orchestration/tools/spawn_parallel_agents.rs:41`, agent allowlists (`loader_tests_part_01_tests.rs:340`), `tools/ops.rs:1332`, `session/builder/mod.rs:91` and the profiling mock. The graph label had exactly one use: `topology.rs:69`. You changed the only one that is safe to change.
- Nothing else on `main` asserts the old graph name, so the rename breaks no other test.
- Both test changes are non-vacuous — revert the one-line rename and `all_topologies_names_do_not_collide_with_production_tools` fires on `spawn_parallel_agents`, and the `includes_...` test fires on the missing `spawn_parallel_graph`. That was CodeRabbit's original objection ("would still pass if the label reverted") and your `all_tools(...)` version answers it properly. Thread resolved, re-approved.
- `tools::all_tools` still has the same 8-parameter signature on `main` (`tools/ops.rs:56`), so the new test compiles after rebase.

## The one blocker: the test module moved

`main` split this file. `topology.rs` no longer has an inline `mod tests { … }`; it ends with

```rust
#[cfg(test)]
#[path = "topology_tests.rs"]
mod tests;
```

and `all_topologies_includes_delegation_and_workflow_scheduler` now lives in **`src/openhuman/agent/tinyagents/topology_tests.rs`**. That is your entire conflict. When you rebase onto `fa044d388`:

- the one-line `describe("spawn_parallel_agents", …)` → `describe("spawn_parallel_graph", …)` stays in `topology.rs` (line 69);
- both test changes move into `topology_tests.rs` — extend the existing `for name in ["delegation", "workflow_runs:scheduler"]` array there, and add `all_topologies_names_do_not_collide_with_production_tools` alongside it. It already has `use super::*;`, and your test uses fully-qualified `crate::…` paths, so nothing else needs adjusting.

Note the file has a line-count gate, so keep the new test in the `_tests.rs` sibling rather than folding it back inline.

## Minor, while you're in there

The PR body's backslashes came out mangled — `\spawn_parallel_agents\`, `\	opology.rs\`, and the test is named as `^Gll_topologies_names_do_not_collide_with_default_tools` where the code says `all_topologies_names_do_not_collide_with_production_tools`. Worth a quick edit so a reviewer can match the description to the diff. The body also has no submission checklist; the checklist gate wants every item either ticked or ticked with its text beginning `N/A`.

Also consider adding `Closes #5601` — it is currently written as `Closes #5601` in the body which is fine, just confirming it is the first line so the link registers.

**To get this green: rebase onto `fa044d388`, split the change across `topology.rs` + `topology_tests.rs` as above, tidy the body.** Nothing about the approach needs to change.

Issue #2186, opened 2026-05-19
Enable failed for Gmail New Gmail Message due to insufficient authentication scopes


GratefulDave:
**Enable failed for Gmail New Gmail Message due to insufficient authentication scopes**

## Summary

Enabling the Gmail New Gmail Message trigger fails because the connected Gmail account lacks the required OAuth scopes, resulting in a 403 permission denied error from Google’s API.

## Problem

**What happened:**  
When attempting to enable the `GMAIL_NEW_GMAIL_MESSAGE` trigger via Composio, the backend returns a 400 Bad Request wrapping a 403 `PERMISSION_DENIED` error from Google, stating “Request had insufficient authentication scopes.”

**What was expected:**  
The trigger should enable successfully, allowing the agent to receive new Gmail message events.

**Impact:**  
Users cannot use the Gmail New Gmail Message trigger functionality after authenticating their Gmail account.

**Steps to reproduce:**
1. Authenticate a Gmail account through Composio / Tiny Humans.
2. Attempt to enable the `GMAIL_NEW_GMAIL_MESSAGE` trigger.
3. Observe the error response.

**Version / platform:**  
Tiny Humans API (`api.tinyhumans.ai`), Composio integration, Gmail API.

## Solution (optional)

**Suspected cause:**  
The OAuth scope granted during authentication does not include sufficient permissions for polling the Gmail message endpoint. The Gmail API requires scopes like `https://www.googleapis.com/auth/gmail.readonly` or `https://www.googleapis.com/auth/gmail.modify` for this trigger.

**Proposed fix:**  
Update the OAuth authorization flow to request the necessary Gmail scopes (e.g., `gmail.readonly`) before enabling the trigger, or re-authenticate with the expanded scopes when a trigger requiring them is enabled.

## Acceptance criteria

- [ ] **Repro gone** — Bug no longer reproduces on Tiny Humans + Composio with a freshly authenticated Gmail account containing the required scopes.
- [ ] **Regression safety** — E2E coverage added for enabling the Gmail New Gmail Message trigger post-authentication with correct scopes.
- [ ] **Diff coverage ≥ 80%** — the fix PR meets the changed-lines coverage gate (Vitest + cargo-llvm-cov, enforced by `.github/workflows/coverage.yml`).
- [ ] **Documentation updated** — OAuth scope requirements for Gmail triggers are documented for users.

## Related

- Error log snippet: `403 PERMISSION_DENIED: Request had insufficient authentication scopes`
[...]

PR #5628, opened 2026-08-20
fix(integrations): include required OAuth scopes for Reddit authorization (#5507)

### LINKED ISSUE #5507: Bug: Reddit connection stuck on connecting, times out with bad request error
## Summary

Connecting a Reddit account via Composio gets stuck indefinitely on "connecting", then times out and surfaces a "bad request" error. The connection never completes.

## Problem

When a user attempts to add a Reddit connection through the integrations panel:

1. The UI enters a "connecting" state
2. The connection hangs — no progress, no redirect to Reddit OAuth
3. After a timeout, the UI shows a "bad request" error
4. No connection is established

This blocks any Reddit-dependent workflow or memory sync.

## Root cause (suspected)

Likely causes, in order of suspicion:

- **OAuth initiation request malformed** — the Composio Reddit integration may require specific scopes or parameters that are not being sent, causing Reddit's OAuth endpoint to reject the initial request with 400 before the user ever sees the authorization page.
- **Reddit app credentials misconfigured** — the Composio managed Reddit app may have an incorrect redirect URI, expired credentials, or be in a state that Reddit's OAuth server rejects.
- **Composio Reddit action broken upstream** — similar to the Notion `fetch_type` gap (#5506), Composio may have updated their Reddit integration contract and the client is sending a stale request shape.

## Reproduction

1. Open the integrations panel.
2. Click to add a Reddit connection.
3. Observe: UI shows "connecting" indefinitely.
4. Connection times out with "bad request" error.
5. No Reddit OAuth authorization page is shown.

## Acceptance criteria

- [ ] Initiating a Reddit connection launches the Reddit OAuth authorization page without timing out.
- [ ] Completing OAuth authorization establishes an ACTIVE Reddit connection visible in the integrations list.
- [ ] A failed connection attempt surfaces a specific, actionable error (e.g. "Reddit OAuth failed — check your credentials") rather than a generic "bad request".

## Related

- openhuman #5506 — Notion sync broken by missing Composio parameter (same integration layer)
- backend #1198 — Composio OAuth connections silently losing ACTIVE status

shivakumar777-web:
Closes #5507

## Summary

- Add `REDDIT_REQUIRED_OAUTH_SCOPES` constant to `ComposioClient` specifying minimal Reddit OAuth scopes (`identity`, `read`, `mysubreddits`) required for user verification, reading posts/comments, and listing subscribed subreddits.
- Map `"reddit"` toolkit in `required_oauth_scopes_for_toolkit` in `src/openhuman/integrations/composio/client_part_01.rs` to automatically inject required scopes during authorization handoffs.
- Add unit tests in `src/openhuman/integrations/composio/client_tests_part_01_tests.rs` validating scope injection and omission for unmapped toolkits.

## Problem

- In Reddit connector integrations, authorization requests lacked explicit OAuth scopes.
- Reddit rejected token exchange with HTTP 400 Bad Request / 403 Forbidden, leaving the UI connection modal indefinitely stuck in `PENDING` until the 5-minute timeout.

## Solution

- Registered minimal non-sensitive Reddit OAuth scopes in `required_oauth_scopes_for_toolkit` mirroring the existing Gmail pattern.
- Ensures Reddit authorization flows request necessary user verification and reading permissions while avoiding broad write or private messaging scopes.

## Submission Checklist

- [x] Linked issue closed via `Closes #5507` in the Related section

## Impact

- Fixes Reddit integration connection failures and timeouts without requesting unnecessary write or private messaging permissions.

## Related

- Closes #5507

## Summary by CodeRabbit

- **New Features**
  - Reddit authorization now requests the required access scopes for identity, reading content, and subscribed communities.

- **Bug Fixes**
  - Improved OAuth authorization handling so unsupported integrations do not receive unnecessary scope parameters.

### DIFF
--- a/src/openhuman/integrations/composio/client_part_01.rs
+++ b/src/openhuman/integrations/composio/client_part_01.rs
@@ -21,6 +21,9 @@ const POST_OAUTH_ACTION_RETRY_DELAY: Duration = Duration::from_secs(10);
 const POST_OAUTH_AUTH_ERROR_STRINGS: &[&str] = &["connection error, try to authenticate"];
 const AUTHORIZE_OAUTH_SCOPES_FIELD: &str = "oauth_scopes";
 const GMAIL_REQUIRED_OAUTH_SCOPES: &[&str] = &["https://www.googleapis.com/auth/gmail.readonly"];
+/// Required OAuth scopes for Reddit integration covering user verification,
+/// post/comment reading, and subscribed subreddit listing (#5507).
+const REDDIT_REQUIRED_OAUTH_SCOPES: &[&str] = &["identity", "read", "mysubreddits"];
 
 /// High-level client for all backend-proxied Composio operations.
 #[derive(Clone)]
@@ -603,6 +606,11 @@ fn required_oauth_scopes_for_toolkit(toolkit: &str) -> &'static [&'static str] {
         // profile-only Google token and trigger enable fails with 403 insufficient
         // authentication scopes (#2186).
         "gmail" => GMAIL_REQUIRED_OAUTH_SCOPES,
+        // Reddit actions require `identity` for user verification, `read` for
+        // retrieving posts and comments, and `mysubreddits` for listing subscribed
+        // communities. Without these explicit scopes, Reddit OAuth handoffs fail
+        // with HTTP 400 Bad Request or omit necessary token permissions (#5507).
+        "reddit" => REDDIT_REQUIRED_OAUTH_SCOPES,
         _ => &[],
     }
 }
--- a/src/openhuman/integrations/composio/client_tests_part_01_tests.rs
+++ b/src/openhuman/integrations/composio/client_tests_part_01_tests.rs
@@ -278,6 +278,65 @@ async fn authorize_merges_gmail_required_oauth_scopes_with_extra_params() {
     assert_eq!(resp.connection_id, "conn-gmail");
 }
 
+/// Verifies that `authorize()` for Reddit automatically attaches mandatory non-sensitive OAuth scopes (#5507)
+/// ensuring permissions for identity, reading, and subscribed subreddits.
+#[tokio::test]
+async fn authorize_merges_reddit_required_oauth_scopes() {
+    let app = Router::new().route(
+        "/agent-integrations/composio/authorize",
+        post(|Json(body): Json<Value>| async move {
+            assert_eq!(body["toolkit"].as_str(), Some("reddit"));
+            let scopes: Vec<&str> = body["oauth_scopes"]
+                .as_array()
+                .expect("reddit authorize should include oauth_scopes")
+                .iter()
+                .map(|item| item.as_str().expect("scope should be a string"))
+                .collect();
+            let expected = ["identity", "read", "mysubreddits"];
+            assert_eq!(scopes.len(), expected.len());
+            for &exp in &expected {
+                assert!(scopes.contains(&exp), "missing expected scope: {exp}");
+            }
+            Json(json!({
+                "success": true,
+                "data": {
+                    "connectUrl": "https://composio.example/reddit/consent",
+                    "connectionId": "conn-reddit"
+                }
+            }))
+        }),
+    );
+    let base = start_mock_backend(app).await;
+    let client = build_client_for(base);
+    let resp = client.authorize("reddit", None).await.unwrap();
+    assert_eq!(resp.connect_url, "https://composio.example/reddit/consent");
+    assert_eq!(resp.connection_id, "conn-reddit");
+}
+
+/// Verifies that unmapped toolkits (such as Notion) do not have artificial `oauth_scopes` injected into authorize payload.
+#[tokio::test]
+async fn authorize_omits_oauth_scopes_for_unmapped_toolkit() {
+    let app = Router::new().route(
+        "/agent-integrations/composio/authorize",
+        post(|Json(body): Json<Value>| async move {
+            assert_eq!(body["toolkit"].as_str(), Some("notion"));
+            assert!(body.get("oauth_scopes").is_none());
+            Json(json!({
+                "success": true,
+                "data": {
+                    "connectUrl": "https://composio.example/notion/consent",
+                    "connectionId": "conn-notion"
+                }
+            }))
+        }),
+    );
+    let base = start_mock_backend(app).await;
+    let client = build_client_for(base);
+    let resp = client.authorize("notion", None).await.unwrap();
+    assert_eq!(resp.connect_url, "https://composio.example/notion/consent");
+    assert_eq!(resp.connection_id, "conn-notion");
+}
+
 #[tokio::test]
 async fn authorize_forwards_extra_params_and_returns_connect_url() {
     let app = Router::new().route(


M3gA-Mind:

**State:** `CONFLICTING`, 4,603 commits behind `main`. The mechanism you extend is intact, so this rebases rather than needing a redesign.

**On the approach:** correct and it mirrors the existing pattern. `required_oauth_scopes_for_toolkit` and `merge_required_oauth_scopes` still exist on `main` and still work the way your PR assumes — the `"gmail"` arm is right there as precedent, added for the same class of bug (#2186: a handoff completing with a profile-only token, then 403 on trigger enable).

### Rebasing — the files moved, the code did not

`client.rs` was split into `_part_NN.rs` fragments (openhuman#5856/#5857). Your two edits land here:

| your edit | now in |
|---|---|
| `REDDIT_REQUIRED_OAUTH_SCOPES` const (beside `GMAIL_REQUIRED_OAUTH_SCOPES`) | `client_part_01.rs:23` |
| the `"reddit" =>` match arm | `client_part_01.rs:599-607` |
| `authorize_merges_reddit_required_oauth_scopes` | `client_tests_part_01_tests.rs` or `_part_02_` |

Note the layout gate now rejects inline `#[cfg(test)] mod tests`, and caps files under `src/openhuman` at 750 lines — put the test in the sibling `*_tests.rs` fragment, not in `client_part_01.rs`.

### One substantive question before this lands

The Gmail precedent requests **one** narrowly-scoped read permission (`gmail.readonly`), with a comment naming the exact action that needs it. This PR requests **ten** Reddit scopes, including `submit`, `edit`, `vote`, `save`, `flair` and `privatemessages` — write and private-message access, for every user who connects Reddit, at authorization time.

The reported bug (#5507) is a handoff stuck `PENDING` because the token lacked what the connector needs. That is very likely fixed by `identity` + `read` (+ `mysubreddits` if subreddit listing is used). The write scopes look like they were added for completeness rather than because a curated action needs them.

This matters beyond tidiness: a broader consent screen costs conversion, and a token that can `submit` and read DMs is a materially larger blast radius if it leaks. **Could you narrow the list to the scopes a curated Reddit action actually calls, and add a comment naming which action needs which** — the way the Gmail arm does? If the write scopes genuinely are needed, saying which action needs each one in the comment is enough.

Happy to be wrong if you have the failing request that proves the wider set is required.

shivakumar777-web:
Thanks for the review and guidance @M3gA-Mind!

I have:
1. Rebased the branch onto latest `main`.
2. Relocated the constant and toolkit match arm into `src/openhuman/integrations/composio/client_part_01.rs` (staying well under the 750-line file limit).
3. Relocated the unit tests into sibling `src/openhuman/integrations/composio/client_tests_part_01_tests.rs`.
4. Narrowed the requested scopes down to only the minimal non-sensitive scopes strictly required: `["identity", "read", "mysubreddits"]`, removing broad write/DM permissions (`submit`, `edit`, `privatemessages`, `vote`, `flair`, `save`).
5. Added inline comments above the constant and match arm documenting the specific actions requiring each scope, following the Gmail precedent.

src/openhuman/integrations/composio/client_part_01.rs
```diff
@@ -603,6 +606,11 @@ fn required_oauth_scopes_for_toolkit(toolkit: &str) -> &'static [&'static str] {
         // profile-only Google token and trigger enable fails with 403 insufficient
         // authentication scopes (#2186).
         "gmail" => GMAIL_REQUIRED_OAUTH_SCOPES,
+        // Reddit actions require `identity` for user verification, `read` for
+        // retrieving posts and comments, and `mysubreddits` for listing subscribed
+        // communities. Without these explicit scopes, Reddit OAuth handoffs fail
+        // with HTTP 400 Bad Request or omit necessary token permissions (#5507).
+        "reddit" => REDDIT_REQUIRED_OAUTH_SCOPES,
```
YellowSnnowmann, line 613: **blocker:** this map is no longer on the path the issue reproduces.

The UI's Connect button goes `ComposioConnectModal.tsx` → `openhuman.composio_authorize` → `schemas_part_02.rs:596` → `ops/connections.rs:105`, and that op calls `connectors::call(config, methods::AUTHORIZE, …)` — the tinyconnectors module — never `ComposioClient`. That was already true at this PR's base (`71a6970`), so the rebase put the arm next to the `gmail` precedent but off the live flow.

The module keeps its own copy of this map, and it is still gmail-only at the pinned `vendor/tinyconnectors` (`4b8c7465`, v0.9.0), `crates/tinyconnectors/src/client/composio.rs:370`:

```rust
fn required_oauth_scopes_for_toolkit(toolkit: &str) -> &'static [&'static str] {
    match toolkit.trim().to_ascii_lowercase().as_str() {
        "gmail" => GMAIL_REQUIRED_OAUTH_SCOPES,
        _ => &[],   // ← reddit lands here on the UI path
    }
}
```

This arm is not dead — the agent tool `composio_authorize` (`tools_part_01.rs:618`, registered at `toolpacks/registry.rs:114`) still reaches it. So keep it if you want the agent path covered, but it cannot close #5507 on its own.

**Fix:** add the same `reddit` arm to tinyconnectors `crates/tinyconnectors/src/client/composio.rs:370`, cut a module release, and re-pin `vendor/tinyconnectors` here — then this host-side arm and the module agree. See blocker on line 26 first, though: the field is dropped at the backend regardless of which layer sets it.

src/openhuman/integrations/composio/client_part_01.rs
```diff
@@ -21,6 +21,9 @@ const POST_OAUTH_ACTION_RETRY_DELAY: Duration = Duration::from_secs(10);
 const POST_OAUTH_AUTH_ERROR_STRINGS: &[&str] = &["connection error, try to authenticate"];
 const AUTHORIZE_OAUTH_SCOPES_FIELD: &str = "oauth_scopes";
 const GMAIL_REQUIRED_OAUTH_SCOPES: &[&str] = &["https://www.googleapis.com/auth/gmail.readonly"];
+/// Required OAuth scopes for Reddit integration covering user verification,
+/// post/comment reading, and subscribed subreddit listing (#5507).
+const REDDIT_REQUIRED_OAUTH_SCOPES: &[&str] = &["identity", "read", "mysubreddits"];
```
YellowSnnowmann, line 26: **blocker:** `oauth_scopes` never reaches Composio — the backend strips it, so these scopes cannot change Reddit's response on any path.

In `backend-alphahuman`:

- `src/validators/agentIntegration.validator.ts:112` — `composioAuthorizeSchema = z.object({ toolkit, config?, callbackUrl? })`. Plain `z.object` on zod 4.2.1 with no `.strict()`, so unknown keys are **stripped**, and `src/routes/agentIntegrations/composio.ts:231-237` hands the stripped `parsed.data` to the controller.
- `src/controllers/agentIntegrations/composio/types.ts:36-45` — `ComposioAuthorizeRequest` is `{ toolkit, config?, callbackUrl? }`. No scope field.
- `src/controllers/agentIntegrations/composio/authorize.ts` — uses only those three, then calls `link.create({auth_config_id, user_id, callback_url, connection_data})` or `client.toolkits.authorize(userId, toolkit)`. `grep -rn "oauth_scopes\|oauthScopes" src` across the backend returns nothing.

`merge_required_oauth_scopes` inserts this at the **top level** of the body, sibling to `toolkit` — not inside `config`, the only extra key the validator keeps. So it is dropped silently, with no error to notice.

The same is true of `GMAIL_REQUIRED_OAUTH_SCOPES` above, which is why "it works for Gmail" doesn't carry here.

**Fix:** make the field survive the edge before relying on it — add a scope field to the backend's validator, `ComposioAuthorizeRequest`, and `authorize.ts` (forwarding it into the auth-config / `connection_data`), or nest it under `config` on this side. Scopes on this stack are normally set on the **Composio auth config** instead (`backend/src/scripts/auditComposioAuthConfigs.ts:23` reads them as `scopes` / `user_scopes` / `scope`), which may be the real home for this fix.

src/openhuman/integrations/composio/client_tests_part_01_tests.rs
```diff
@@ -278,6 +278,65 @@ async fn authorize_merges_gmail_required_oauth_scopes_with_extra_params() {
     assert_eq!(resp.connection_id, "conn-gmail");
 }
 
+/// Verifies that `authorize()` for Reddit automatically attaches mandatory non-sensitive OAuth scopes (#5507)
+/// ensuring permissions for identity, reading, and subscribed subreddits.
+#[tokio::test]
+async fn authorize_merges_reddit_required_oauth_scopes() {
```
YellowSnnowmann, line 284: **major:** this test can't fail for the bug it's guarding, which is why CI is green while production behaviour is unchanged.

It stands up a mock at `/agent-integrations/composio/authorize` and asserts the body `ComposioClient` builds. That pins the host helper correctly, but the UI flow never calls `ComposioClient::authorize` (it goes through `connectors::call(methods::AUTHORIZE)`), and the real backend strips `oauth_scopes` from the body this mock happily accepts. So a green run here proves the map has a `reddit` entry — not that a Reddit connection can be established.

**Fix:** add the regression test at the layer that actually runs. `ops_tests_part_01_tests.rs` already exercises `composio_authorize(&config, "gmail", None)` through the module against a mock backend, taking `module_guard()` first — mirror that for `reddit` and assert the scopes on the body the *module* sends. That test will fail today, which is what you want from it. Keep this one as-is for the agent-tool path.

YellowSnnowmann (changes requested):
## Summary

The change itself is clean, minimal, and correctly follows the `gmail` precedent — the narrowing you did after @M3gA-Mind's review is right, the file placement after the `_part_NN` split is right, the tests are wired correctly (`client.rs` → `client_tests.rs` → `part_01_tests`, so they really do compile and run), and CI is green.

But I traced the change end-to-end against `main` and **it does not reach the flow the issue reports**, on two independent counts. Both are blockers for "Closes #5507", neither is your fault — the ground moved underneath this PR while it was open.

### Blockers (2) · Major (1) · Minor (0) · Nitpicks (0)

**1. The scope map you edited is no longer on the connect path.**

The issue's repro is "open the integrations panel, click to add a Reddit connection". That path is:

```
ComposioConnectModal.tsx  →  composioApi.ts:155  'openhuman.composio_authorize'
  →  schemas_part_02.rs:596  handle_authorize
  →  ops/connections.rs:105   composio_authorize
  →  connectors::call(config, methods::AUTHORIZE, …)      ← the tinyconnectors module
```

`ops/connections.rs:110-118` sends the request straight to the module and never constructs a `ComposioClient`. Its own comment says so: *"The module owns the whole handoff … It also owns the difference between the proxy's authorize body and v3's link call."* This was already true at **this PR's own base** (`71a6970`), not just on today's `main` — so the rebase landed the edit next to the precedent but off the live path.

The module carries its own copy of this map, at `crates/tinyconnectors/src/client/composio.rs:370` in the pinned `vendor/tinyconnectors` (`4b8c7465`, v0.9.0):

```rust
fn required_oauth_scopes_for_toolkit(toolkit: &str) -> &'static [&'static str] {
    match toolkit.trim().to_ascii_lowercase().as_str() {
        "gmail" => GMAIL_REQUIRED_OAUTH_SCOPES,
        _ => &[],
    }
}
```

`reddit` hits `_ => &[]` there, which is exactly the gap the issue describes — just one layer down from where this PR patches it.

What the host-side map *does* still feed is the agent tool `composio_authorize` (`tools_part_01.rs:618`, registered in the `composio` toolpack at `toolpacks/registry.rs:114`). That is a real path, so this is not dead code — it just isn't the one in the repro or the acceptance criteria. (`oauth_handoff::authorize_with_meta_guard` is the only other caller and has test-only callers.)

**2. The backend drops `oauth_scopes` before it can be used — so the field is inert on *every* path.**

Even on the agent-tool path, the value never leaves the edge. In `backend-alphahuman`:

- `src/validators/agentIntegration.validator.ts:112` — `composioAuthorizeSchema = z.object({ toolkit, config?, callbackUrl? })`. Plain `z.object` on zod 4.2.1, no `.strict()`, so **unknown keys are stripped**, and `src/routes/agentIntegrations/composio.ts:231-237` passes `parsed.data` (the stripped object) to the controller.
- `src/controllers/agentIntegrations/composio/types.ts:36-45` — `ComposioAuthorizeRequest` has only `toolkit`, `config`, `callbackUrl`. No scope field.
- `src/controllers/agentIntegrations/composio/authorize.ts` — reads only those three and calls `link.create({auth_config_id, user_id, callback_url, connection_data})` or `client.toolkits.authorize(userId, toolkit)`. There is no scope plumbing anywhere; `grep -rn "oauth_scopes\|oauthScopes" src` in the backend returns **nothing**.

Note where the client puts it: `merge_required_oauth_scopes` inserts `oauth_scopes` at the **top level** of the body (sibling to `toolkit`), not inside `config` — the one key that does survive the validator. So it is stripped, silently, with no error.

This also means the `gmail` arm this PR mirrors is itself inert against this backend, which removes "it works for Gmail" as evidence that it will work for Reddit. Worth settling separately — it may be why #2186 keeps resurfacing.

### Outside the diff — where the fix likely belongs

Scopes on this stack are configured on the **Composio auth config**, not per-authorize-call. The backend's own audit script says so: `src/scripts/auditComposioAuthConfigs.ts:23` reads them off the auth config as `SCOPE_KEYS = ['scopes', 'user_scopes', 'scope']`.

And there is a more direct candidate for the literal "bad request". `reddit` **is** allowlisted (`src/services/composio/allowlist.ts:87`), so `isToolkitConnectable` short-circuits `true` and the request gets past the enabled gate. It then reaches `authorize.ts` → `authConfigs.list({ toolkit: 'reddit' })`, and if the project has no Reddit auth config: `throw new BadRequestError('No auth config found for toolkit "reddit"')` → **HTTP 400, before any OAuth page is ever rendered**. That is the reported symptom exactly, it is independent of scopes, and it matches suspected cause #2 in the issue ("Composio managed Reddit app misconfigured") rather than #1.

So before more client-side work, I'd capture one real failing `POST /agent-integrations/composio/authorize` for `reddit` and read the 400 body. It decides which of these it is — and no client change can fix it if it is the missing auth config.

If the scope hint *is* still wanted after that, the shape is:
1. add the `reddit` arm to **tinyconnectors** `client/composio.rs:370` (module release + re-pin in openhuman), and
2. teach the backend to accept and forward a scope field (validator + `ComposioAuthorizeRequest` + `authorize.ts`), or nest it under `config` — otherwise step 1 is stripped too.

### Also not covered

Acceptance criterion 3 — *"a failed attempt surfaces a specific, actionable error … rather than a generic 'bad request'"* — isn't addressed by this PR. It needs either a separate change or an explicit descope on the issue, or `Closes #5507` will close it half-done.

Separately: Reddit only issues a refresh token when the authorize request carries `duration=permanent`. Without it the token expires in an hour with no refresh — the symptom of linked backend #1198. Composio builds the final authorize URL, so that belongs on the auth config too, but it is worth settling in the same pass.

### Verified / looks good

- Scope values are all valid Reddit OAuth scopes, and the narrowed set (no `submit`/`edit`/`vote`/`privatemessages`) is the right call — @M3gA-Mind's review is fully addressed.
- Test wiring is correct and the tests do compile: `client.rs` declares `#[cfg(test)] #[path = "client_tests.rs"] mod tests;` and `client_tests.rs` declares `#[path = "client_tests_part_01_tests.rs"] mod part_01_tests;`.
- `merge_required_oauth_scopes` / `append_missing_oauth_scopes` are additive and dedupe correctly; the `_ => &[]` early-return means no other toolkit's body changes shape. Nothing here is a regression risk.
- Merges cleanly onto today's `main` (363 commits ahead): `git merge-tree` is clean and **neither** touched file drifted since the PR base, so there is no stale-green-CI trap. Both files stay under the 750-line layout gate (674 / 699).
- CI is green at `64217a3` (Rust Quality, Rust Core Coverage, PR CI Gate). I verified the merge and the call graph locally but did not run a cold local build — the findings above are about routing, not compilation.

Happy to be wrong on blocker 1 if the module re-enters the host client somewhere I missed. Blocker 2 I'd check first either way: it decides whether *any* client-side scope hint can work today.

YellowSnnowmann:
### Followed this to the bottom — the root cause is in the backend, and it isn't scopes

@shivakumar777-web thanks for the work on this, and sorry for where it lands. I traced #5507 end to end after my review above and found the actual cause. It is not something a client-side change can reach, so this PR can't close the issue — but the two findings I left earlier were the thread that led to it, so it wasn't wasted.

**Composio has no per-authorize scope parameter.** `LinkCreateParams` in the installed SDK is:

```ts
export interface LinkCreateParams {
  auth_config_id: string;
  user_id: string;
  alias?: string;
  callback_url?: string;
  connection_data?: /* pre-fill values for connection fields */;
}
```

No scopes. OAuth scopes live on the **auth config** — `authConfigs.create` / `update` take `scopes` and `user_scopes`, and the backend's own `src/scripts/auditComposioAuthConfigs.ts` reads them off the auth config as `scopes` / `user_scopes` / `scope`. So an `oauth_scopes` field in the authorize body has no destination even before the backend validator strips it. That applies to `GMAIL_REQUIRED_OAUTH_SCOPES` as much as to the Reddit entry here.

**What was actually breaking Reddit.** `backend/src/controllers/agentIntegrations/composio/authorize.ts`:

```ts
const link = getLinkCreateClient(client);
if (link?.create || hasExtras) {
  const list = await authConfigs.list({ toolkit });
  const authConfigId = list.items[0]?.id;
  if (!authConfigId) throw new BadRequestError(`No auth config found for toolkit "${toolkit}"`);
  ...
} else {
  connectionRequest = await client.toolkits.authorize(userId, toolkit);   // ← never runs
}
```

`getLinkCreateClient` falls back to `client.client?.link`, and `@composio/core`'s `Composio` carries the underlying `@composio/client` on `.client`, which registers a `link` resource. Checked at runtime on the installed `@composio/core@0.6.10`:

```
top-level link:     undefined
inner.link.create:  function
```

So `link?.create` is always truthy, the `if` always wins, and the `else` is dead code — which matters because `toolkits.authorize` is the only call that *creates* an auth config when one is missing ("This method will create an auth config if one doesn't exist and initiate a connection request").

Net effect: every authorize demanded a pre-existing project auth config, while `isToolkitConnectable` kept waving managed-allowlist toolkits through on the premise that they need none ("Managed allowlist apps are connectable by definition (Composio-managed auth)"). `reddit` is on that allowlist with no auth config, so it was refused with `No auth config found for toolkit "reddit"` — **HTTP 400, raised before the consent screen is ever rendered**. That is the reported symptom exactly: stuck on "connecting", then "bad request", and no Reddit OAuth page.

Which makes suspected cause **#2** in #5507 the right one ("Composio managed Reddit app misconfigured"), not #1 — and your PR description, the triage notes, and my own earlier comment on the issue all backed #1. The missing-scopes reading was a reasonable one; it just wasn't it.

**Fix is up:** tinyhumansai/backend#1325 and its PR, which makes a missing auth config fatal only when the caller actually needs `config`/`callbackUrl`, and otherwise falls through to `toolkits.authorize` so it provisions one. It also fixes a test in that file whose name already described the intended behaviour while asserting the opposite — the same class of gap I flagged on your test here.

**Suggest closing this PR.** Nothing in it is wrong as code, but it cannot affect the connect flow: the UI path goes through the tinyconnectors module rather than `ComposioClient`, and the field is stripped at the backend edge and has nowhere to go in Composio's API regardless. If you would like to keep the narrowed Reddit scope list somewhere, the place it would actually take effect is the Reddit **auth config**, not the authorize call.

Happy to answer anything on the trace, and thanks again — the two blockers you took on the chin here are what made the real cause findable.

Issue #5498, opened 2026-08-11
Bug (macOS/DMG): GPT-5.x models fail with HTTP 400 — tinyagents sends max_tokens but these models require max_completion_tokens


Al629176:
## Summary

On a fresh macOS DMG install (unmodified), using GPT-5.4-mini or GPT-5.6-Luna as the main model causes every chat turn to fail immediately. The request reaches OpenAI successfully but is rejected with HTTP 400 because tinyagents sends `max_tokens` — a parameter these GPT-5.x models do not accept. The correct parameter is `max_completion_tokens`.

## Problem

Exact error from OpenHuman logs:

```
tinyagents harness run failed: model error: model error: openai returned HTTP 400
(unsupported_parameter): Unsupported parameter: 'max_tokens' is not supported with
this model. Use 'max_completion_tokens' instead.
```

OpenAI's GPT-5.x model family (`gpt-5.4-mini`, `gpt-5.6-luna`, and likely the full family) dropped support for the `max_tokens` request field and require `max_completion_tokens` in its place. tinyagents is sending `max_tokens` unconditionally, so every turn with one of these models hard-fails before the model produces any output.

This is a parameter name mismatch in the tinyagents request builder — auth and connectivity are fine, the request reaches OpenAI, and OpenAI rejects it at the protocol level.

## Reproduction

1. Fresh macOS DMG install (no source modifications).
2. Set main model to `gpt-5.4-mini` or `gpt-5.6-luna` with a valid OpenAI key.
3. Send any chat message.
4. Result: immediate failure, no response, HTTP 400 in logs.

Confirmed on:
- `gpt-5.4-mini`
- `gpt-5.6-luna`

## Root cause

tinyagents constructs the OpenAI chat completion request with `max_tokens`. For GPT-5.x models this field must be `max_completion_tokens`. The fix needs to be applied at the request-building layer — either by detecting the model family and switching the field name, or by migrating to `max_completion_tokens` across the board (OpenAI accepts it for older models too).

## Solution

- Replace `max_tokens` with `max_completion_tokens` in the tinyagents OpenAI request builder.
- Audit other OpenAI-compatible parameters that may have similar deprecation gaps in the GPT-5.x family.
- Regression-test with at least `gpt-5.4-mini` and `gpt-5.6-luna` to confirm chat turns complete successfully.

## Acceptance criteria

[...]

PR #5629, opened 2026-08-20
fix(inference): honour max_tokens / max_completion_tokens in local server

shivakumar777-web:
Refs #5498

## Summary

- Support `max_completion_tokens` in local OpenAI-compatible inference request types (`src/openhuman/inference/http/types.rs`).
- Update `chat_completions_handler` in `src/openhuman/inference/http/server.rs` to honour caller token limits using `max_completion_tokens.or(max_tokens)` precedence instead of dropping them.
- Add deserialization unit tests in `src/openhuman/inference/http/http_tests.rs` covering both-present and legacy-only token parameter cases.

## Problem

- In the local OpenAI-compatible HTTP server (`POST /v1/chat/completions`), `ChatCompletionRequest` parsed `max_tokens` but `chat_completions_handler` previously constructed `ModelRequest` with only model and temperature, silently dropping the caller's token limit.
- Newer OpenAI clients and models also send `max_completion_tokens` instead of `max_tokens`.

## Solution

- Added `max_completion_tokens` to `ChatCompletionRequest`.
- Forward token limits into `ModelRequest::with_max_tokens` prioritizing `req.max_completion_tokens.or(req.max_tokens)`.
- Verified serialization and fallback precedence with unit tests.

## Submission Checklist

- [x] Linked issue referenced via `Refs #5498` in the Related section

## Impact

- Ensures caller token limits (`max_completion_tokens` and legacy `max_tokens`) are honoured in local OpenAI-compatible HTTP chat completions.

## Related

- Refs #5498

## Summary by CodeRabbit

- **New Features**
  - Added support for the `max_completion_tokens` parameter when requesting chat completions.
  - Requests can now specify the desired completion length using the newer parameter.

- **Compatibility**
  - Existing requests using `max_tokens` continue to work.
  - When both limits are provided, `max_completion_tokens` takes precedence.

- **Tests**
  - Added coverage confirming correct handling of both parameters and their precedence.

### DIFF
--- a/src/openhuman/inference/http/http_tests.rs
+++ b/src/openhuman/inference/http/http_tests.rs
@@ -140,3 +140,39 @@ fn strip_temperature_suffix_only_removes_numeric_suffixes() {
     assert_eq!(strip_temperature_suffix("llama3.1:8b@1"), "llama3.1:8b");
     assert_eq!(strip_temperature_suffix("gpt@beta"), "gpt@beta");
 }
+
+/// Asserts that `ChatCompletionRequest` parses `max_completion_tokens` and respects
+/// fallback precedence over `max_tokens` (Refs #5498).
+#[test]
+fn test_chat_completion_request_deserializes_max_completion_tokens() {
+    use crate::openhuman::inference::http::types::ChatCompletionRequest;
+
+    let json_both = serde_json::json!({
+        "model": "gpt-5",
+        "messages": [{ "role": "user", "content": "hello" }],
+        "max_tokens": 100,
+        "max_completion_tokens": 200
+    });
+    let parsed: ChatCompletionRequest = serde_json::from_value(json_both).unwrap();
+    assert_eq!(parsed.max_tokens, Some(100));
+    assert_eq!(parsed.max_completion_tokens, Some(200));
+    assert_eq!(
+        parsed.max_completion_tokens.or(parsed.max_tokens),
+        Some(200)
+    );
+
+    let json_legacy = serde_json::json!({
+        "model": "gpt-4o",
+        "messages": [{ "role": "user", "content": "hello" }],
+        "max_tokens": 150
+    });
+    let parsed_legacy: ChatCompletionRequest = serde_json::from_value(json_legacy).unwrap();
+    assert_eq!(parsed_legacy.max_tokens, Some(150));
+    assert_eq!(parsed_legacy.max_completion_tokens, None);
+    assert_eq!(
+        parsed_legacy
+            .max_completion_tokens
+            .or(parsed_legacy.max_tokens),
+        Some(150)
+    );
+}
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
Maintainer review (review only — no changes pushed).

**The diff and the PR description do not match, and I think the mismatch matters more than the code.**

The description lists four changes. Only the fourth is actually in the diff:

| Claimed in the body | In the diff? |
|---|---|
| Add `bare_model_segment` helper to strip router/gateway namespace prefixes | **No** — 0 occurrences |
| Update `is_reasoning_model` / `is_gpt5_family` to match bare segments | **No** — 0 occurrences |
| Add `max_completion_tokens: bool` to `Degrade` + `degrade_for_400` retry | **No** — 0 occurrences |
| Support `max_completion_tokens` in local HTTP inference types and server | Yes |

```
$ gh pr diff 5629 | grep -c "bare_model_segment\|degrade_for_400\|is_gpt5_family"
0
```

The three changed files are all under `src/openhuman/inference/http/` — that is the OpenAI-compatible server OpenHuman **exposes**, i.e. the inbound direction. #5498 is about the outbound direction: what we **send** to OpenAI. Those are opposite sides of the wire, and nothing here touches the request we emit.

For confirmation, none of the classification symbols the body describes exist in this repo at all:

```
$ git grep -n "is_reasoning_model\|is_gpt5_family\|bare_model_segment" upstream/main -- src/
(no matches)
```

They live in the vendored `tinyagents` provider transport. That is where a GPT-5 request picks `max_tokens` over `max_completion_tokens`, and that is where #5498 has to be fixed — it is a vendored-crate change plus a pin bump, not an edit in this repo.

**So: `Closes #5498` would close an issue this PR does not fix.** Please drop the closing keyword (or change it to `Refs #5498`) — GitHub closes on merge regardless of surrounding wording, so it needs to actually come out of the body.

**The code that *is* here is fine, and worth keeping.** Before this change `chat_completions_handler` built `ModelRequest` with only model and temperature — `req.max_tokens` was parsed and then silently dropped, so a caller's token limit never reached the model. The patch honours both fields with `max_completion_tokens.or(max_tokens)`, which is the right precedence for an OpenAI-compatible surface, and the deserialization test pins both the both-present and legacy-only cases. That is a real (separate, smaller) bug fix.

**What I would do:** retitle to what it does — something like *"fix(inference): honour max_tokens / max_completion_tokens in the local OpenAI-compatible server"* — rewrite the body to match the diff, and drop the `Closes`. Then it can land on its own merit and #5498 stays open for the tinyagents-side fix.

**Also note:** this PR is currently `CONFLICTING` against `main` and needs a rebase. Its last CI run was green, but that was on the pre-conflict tree, so it will need to re-run after the rebase. I have not rebased it for you — that is yours to do.

shivakumar777-web:
Thank you for the thoughtful and constructive review @M3gA-Mind!

Following Path A:
1. Rebased onto latest `main`.
2. Retitled the PR to accurately describe the scope of the fix: `fix(inference): honour max_tokens / max_completion_tokens in local server`.
3. Rewrote the PR description to focus exclusively on the local OpenAI-compatible inference server changes (`src/openhuman/inference/http/`).
4. Replaced `Closes #5498` with `Refs #5498` so #5498 remains open for the separate tinyagents outbound transport update.
5. Preserved the `max_completion_tokens.or(max_tokens)` precedence and deserialization tests in `http_tests.rs`.

Ready for re-review!

PR #5650, opened 2026-08-21
chore: remove dead local-STT model download plumbing

### LINKED ISSUE #5462: chore: remove the dead local-STT model download plumbing (core assets + settings UI)
With the bundled whisper.cpp engine deleted in #5454, speech-to-text is a hosted round trip selected by `voice_server.stt_engine`. There is no local STT model to fetch any more — but the download plumbing that existed to fetch it is still wired up and still reachable from the UI.

Core:
- `src/openhuman/inference/local/service/assets.rs` still has an `stt` slot in the download state machine (`"stt" => status.stt_state = "downloading"`, twice).

Frontend:
- `app/src/components/settings/panels/local-model/ModelDownloadSection.tsx` (+ its test)
- `app/src/components/settings/panels/LocalModelDebugPanel.tsx`
- the `stt` fields threaded through `app/src/lib/coreState/store.ts` and `app/src/services/coreStateApi.ts`

A user can currently ask the app to download a speech-to-text model that nothing will ever load. Worth checking whether the whole asset slot goes away or just the STT arm — the LLM slot is still live.

Found while closing out the build-time reduction program (#5453 / #5454).

JMak-Security:
## Update (2026-09-12): scope has collapsed to a single line since this was opened

Upstream has moved a long way since Aug 21, and most of this PR has already landed by other means:

- **`src/openhuman/inference/local/service/assets.rs`** — the two unreachable `"stt"` match arms are **already gone**, removed by `c394d8db3` ("feat(voice): remove the local whisper STT engine in favor of configurable hosted engines"). No `stt` string remains in that file.
- **`app/src/components/settings/panels/LocalModelDebugPanel.tsx`** — the panel file has been **deleted** upstream (this was the `modify/delete` conflict).
- **`app/src/utils/tauriCommands/localAi.ts`** — **still live.** `main` still carries `capability: 'chat' | 'vision' | 'embedding' | 'stt' | 'tts'`. Removing `'stt'` from that union is the only surviving change.

The branch has since been merged up to `main` (`04ca5527`, "resolve stale STT plumbing"), so this PR is now effectively a **one-line change** to `localAi.ts`, and all CI checks are green. Thanks to @M3gA-Mind for the review that pinned this down and to @senamakel for catching the branch up.

The remainder of #5462's scope (core assets + the settings-UI panel) was satisfied by the voice work, so the issue may be better closed with a note than by this PR alone — see discussion below.

---

## Summary

Removes dead local-STT model download plumbing that issue #5462 identified.

`download_asset()` in `assets.rs` only accepts `chat`, `vision`, `embedding`, and `tts` as capabilities (its match statement explicitly errors with "Unknown capability. Use one of: chat, vision, embedding, tts." for anything else). Despite that, `download_file_with_progress()` still had two leftover match arms handling a `"stt"` label that can never be reached from that entry point.

The frontend mirrored the same stale type: the capability parameter union in both `app/src/utils/tauriCommands/localAi.ts` and `app/src/components/settings/panels/LocalModelDebugPanel.tsx` still included `'stt'`, even though `ModelDownloadSection.tsx` never renders an STT download control.

## What's intentionally NOT touched

`stt_state` / `stt_model_id` on `LocalAiStatus` (in `types.rs`) are still live — `speech.rs` sets `stt_state = "ready"` after a successful **hosted** STT transcription. Only the local-model-download trigger path was dead; state tracking for the hosted path is unaffected.

## Changes (as originally opened)

- `src/openhuman/inference/local/service/assets.rs`: removed the two unreachable `"stt" => ...` match arms in `download_file_with_progress()`. *(now already handled upstream)*
- `app/src/utils/tauriCommands/localAi.ts`: removed `'stt'` from the capability union type. *(the surviving change)*
- `app/src/components/settings/panels/LocalModelDebugPanel.tsx`: removed `'stt'` from the capability union type in `triggerAssetDownload`. *(file since deleted upstream)*

## Testing

Traced the full call graph by hand (`download_asset` → `download_file_with_progress`, and the frontend `ModelDownloadSection.tsx` UI) to confirm the removed code paths are genuinely unreachable, not just unused-looking. No behavioral change for chat/vision/embedding/tts or for hosted STT.

Fixes #5462

## Summary by CodeRabbit

- **Updates**
  - Local AI asset downloads now support chat, vision, embedding, and text-to-speech capabilities.
  - Speech-to-text is no longer available as a downloadable local AI capability.


M3gA-Mind:
**Maintainer review — most of this has landed by other means; one line of it is still live.** Read-only pass; I have not pushed to this branch and have not approved it.

@JMak-Security — the PR is `CONFLICTING`, and the conflicts are not the ordinary kind. Checked against `main` @ `fa044d388`:

| your change | state on `main` |
|---|---|
| `src/openhuman/inference/local/service/assets.rs` — remove the two unreachable `"stt"` arms | **already gone.** `c394d8db3` ("feat(voice): remove the local whisper STT engine in favor of configurable hosted engines") removed them; there is no `stt` string left in that file at all. Your hunk has nothing to delete. |
| `app/src/components/settings/panels/LocalModelDebugPanel.tsx` — drop `'stt'` from the union | **file deleted.** This is the `modify/delete` conflict. The panel no longer exists. |
| `app/src/utils/tauriCommands/localAi.ts` — drop `'stt'` from the capability union | **still live.** `main` still has `capability: 'chat' \| 'vision' \| 'embedding' \| 'stt' \| 'tts'` at `localAi.ts:368`. |

So the PR is now effectively a one-line change, and #5462 is still open because that one line is still there.

**What I'd suggest:** rebase onto `main` and take only the `localAi.ts` hunk — drop the other two, they are already satisfied. Then re-check #5462's scope: it names "core assets + settings UI", and both of those halves have since been handled by the voice work, so the issue may want closing with a note rather than a fix, with just the stale union type as the remainder. Worth a sentence in the PR body either way so a reviewer is not left comparing against a description that no longer matches the diff.

Your call-graph reasoning was right at the time and is worth keeping in the record: `download_asset()` only accepts `chat`/`vision`/`embedding`/`tts`, so the `"stt"` arms below it were unreachable, and `ModelDownloadSection.tsx` never rendered an STT control. The "intentionally NOT touched" note about `stt_state` / `stt_model_id` on `LocalAiStatus` still holds — `speech.rs` sets `stt_state = "ready"` after a hosted transcription, and that path is unaffected.

**Also failing:** `Markdown Link Check` (×2). Unrelated to your diff as far as I can tell — likely to clear on a rebase, but worth confirming after.

Nothing here is wrong; it has just been overtaken. A rebase down to the surviving hunk makes this mergeable.

JMak-Security:
@M3gA-Mind thanks for the careful pass — your read was exactly right, and the branch has since been caught up to `main` (`04ca5527`), so the three-way state now resolves cleanly:

- `assets.rs` — the unreachable `"stt"` arms are gone via `c394d8db3`; my hunk there is a no-op now.
- `LocalModelDebugPanel.tsx` — the panel is deleted upstream, so the `modify/delete` conflict resolves to "deleted" and there's nothing to carry over.
- `localAi.ts` — the one surviving change: dropping `'stt'` from the `capability` union so it matches what `download_asset()` actually accepts (`chat | vision | embedding | tts`).

So this PR is now effectively that single `localAi.ts` line, and CI is green (CodeRabbit's re-review found no actionable comments; the only non-pass is its soft docstring-coverage warning). I've updated the PR description to say as much so it no longer reads against a diff that has moved on.

On #5462: since the core-assets and settings-UI halves were both handled by the voice work, I agree the issue is better **closed with a note** pointing at `c394d8db3` for those two, leaving only the stale union type — which this PR removes. Happy to either land this one-liner and close #5462 referencing both, or close #5462 with a note and drop this if you'd rather fold the type fix into the voice cleanup. Whatever's least churn for you.

Issue #5649, opened 2026-08-21
[claude-code] Pasted image attachments are dropped before the model (text-only)


nocstah:
## Summary
With **Claude Code (`claude-code`) as the brain**, pasting image(s) into a chat saves them to disk but they are **never delivered to the model**. The message sent to the brain is text‑only, so the assistant responds as if no image was attached.

## Environment
- OpenHuman **0.63.7** (macOS desktop `.dmg`), Apple Silicon
- Brain/provider: `claude-code`, model `claude-opus-5` (vision‑capable)
- Claude Code CLI: **2.1.237**

## What happens
Paste 1–2 images + text, send. The images are written to `users/<user>/workspace/attachments/<hash>.jpg`, but the Claude session receives **only the text**.

## Evidence
- Files on disk: `workspace/attachments/<hashA>.jpg`, `workspace/attachments/<hashB>.jpg`.
- The corresponding Claude session `.jsonl` contains the typed text but **zero** references to the image filenames, no `attachments/` path, and no `{"type":"image"}` content block.
- The attachments dir is also **outside** the agent's granted dirs (`--add-dir` is `~/OpenHuman/projects`), so even a path reference would be unreadable by the agent.

## Root cause (hypothesis)
The claude-code provider doesn't translate attachments into the outgoing user message. The CLI accepts images via stream‑json `content` blocks (`{"type":"image","source":{"type":"base64",...}}`). OpenHuman should encode attached images into the message for this provider (as its cloud vision models presumably already receive them), and/or grant the attachments dir to the agent so a path reference is readable.

## Repro
1. Select Claude Code as the brain.
2. Paste 1–2 images with some text, send.
3. Image is saved under `workspace/attachments/` but the model replies as if text‑only.

## Expected
The brain receives pasted images — multimodal parity with the cloud models.

PR #5651, opened 2026-08-21
fix(claude-code): recreate a persisted session the CLI no longer has

### LINKED ISSUE #5648: [claude-code] Chat fails permanently — driver resumes a session it never created ("No conversation found")
## Summary
With **Claude Code (`claude-code` / `cli://claude-code`) selected as the brain**, a thread's chat can fail **permanently** with `No conversation found with session ID: <uuid>`. The driver generates a session id, persists it, and then only ever **resumes** it — if that session was never created in Claude's store, every turn fails and the mapping is never repaired.

## Environment
- OpenHuman **0.63.7** (macOS desktop `.dmg`), Apple Silicon
- Brain/provider: `claude-code`, model `claude-opus-5`
- Claude Code CLI: **2.1.237** (`~/.local/bin/claude`)
- Fully local setup (no active cloud session)

## What happens
OpenHuman writes a session UUID to `users/<user>/claude-code-sessions.json` (keyed by a conversation hash) and drives the CLI to **resume** it. When the id is not present in `~/.claude/projects/*/`, the call exits non‑zero:

```
[claude-code][driver] exit Some(1)
stderr=No conversation found with session ID: <uuid>
```

Shown to the user as a generic "something went wrong". The mapping is **never invalidated**, so app/model restarts and resends keep resuming the same missing id — the thread is poisoned.

## Captured CLI invocation
(obtained by shimming the `claude` binary)
```
cwd=/Users/<user>/OpenHuman/projects
claude -p --input-format stream-json --output-format stream-json --verbose \
  --include-partial-messages --add-dir /Users/<user>/OpenHuman/projects \
  --permission-mode bypassPermissions --resume <uuid> --model claude-opus-5 \
  --append-system-prompt-file <tmp> --mcp-config <tmp> --strict-mcp-config
```

## Root cause (hypothesis)
- OpenHuman mints and persists its **own** session id *before* Claude confirms creation, then `--resume`s it. On the first turn it should let Claude create the session (the CLI returns the real `session_id` in the stream‑json `init`/`result` events) and persist **that**, then resume on later turns.
- Or: the first (creating) call fails transiently, the id is already written to `claude-code-sessions.json`, and there is **no self-heal** — the driver only `--resume`s thereafter.

## Repro
1. Select Claude Code as the brain (CLI 2.1.x).
2. Chat. If a first‑turn call fails once, or the session store is cleared, the thread's id is poisoned.
3. Every later message → `No conversation found with session ID: …`.

## Expected
Claude Code creates the session on first use (or recreates when a resume target is missing), instead of looping on a non‑existent id.

## Verified workaround (external shim — confirms it's the driver, not the CLI/model)
Intercepting `claude` and rewriting `--resume <id>` → `--session-id <id>` when `~/.claude/projects/*/<id>.jsonl` does not exist makes chat work immediately. The CLI (2.1.237) and model (`claude-opus-5`) are fine; the create/resume logic in the driver is the issue.

## Related secondary observations (same claude-code brain path, lower priority)
- **Tool availability**: the tool‑spec filter reports `visible=76 / total=315, policy_restricted=true`; some operations surface "tool not available". Worth confirming the claude-code tool bridging matches the cloud-model tool surface.
- **Background agents on the main model**: routing all workloads to claude-code runs the entity extractor, Goals Curator, etc. through the heavy chat model; the Goals Curator's "nothing changed" summaries can read as confusing chat output when the user is just saving a note.
- **Local-mode churn**: a fully-local setup still polls `orchestration/v1/sessions`, `teams/me/usage`, `announcements/latest` ~every 20s and emits `SESSION_EXPIRED` banners. A true offline mode would silence these.

nocstah:
## Summary

- Fixes a hard failure where a Claude Code (`claude-code`) chat thread breaks **permanently** with `No conversation found with session ID: <uuid>`.
- The driver persists a self-minted session id *before* the creating turn succeeds, then only ever `--resume`s it — so any first-turn failure poisons the thread with no recovery.
- Now the driver checks whether the CLI still has the session on disk before choosing `--resume`, and recreates a missing one with `--session-id` + full context.

## Problem

`run_turn` in `claude_code/driver.rs` decides new-vs-resume from `session_store` alone: a persisted UUID means `--resume`. But the id is written to `claude-code-sessions.json` *before* the creating call runs, so if that call fails (transient error, cancelled turn, or the user's CLI session store is cleared), the id survives but the session was never created. Every later turn then `--resume`s a session the CLI does not have and exits non-zero with `No conversation found with session ID: <uuid>`. Restarting the app or model does not help — the mapping is never invalidated.

Reproduced against `claude` CLI 2.1.237 / model `claude-opus-5`. Confirmed a shim that rewrites `--resume <id>` → `--session-id <id>` when the session file is absent makes chat work immediately, i.e. the CLI and model are fine; the driver's create/resume bookkeeping is the bug.

## Solution

- Add `cc_session_exists` (+ a pure, testable `cc_session_exists_in`) that scans `<CLAUDE_CONFIG_DIR|~/.claude>/projects/*/<id>.jsonl`.
- Treat a persisted id whose session is absent as **new**: recreate it with the same id (so the thread's mapping stays stable) via `--session-id`, and send full context (`is_new = true`) since the CLI holds no prior turns.
- No behavior change on the healthy path (existing session → `--resume` as before).

## Submission Checklist

- [x] Tests added or updated — `cc_session_exists_in_detects_present_and_absent` covers present, absent, and wrong-id cases.
- [x] **Diff coverage ≥ 80%** — N/A: the changed run_turn subprocess path has no unit harness; the helper is unit-tested.
- [x] Coverage matrix updated — `N/A: behaviour-only fix, no feature row change`.
- [x] Manual smoke checklist — `N/A: does not touch release-cut surfaces`.

## Impact

- Desktop/CLI, Claude Code provider only. Recovers permanently-broken threads and hardens against future first-turn failures. No migration, no new deps.

## Related

- Closes: #5648
- Follow-up PR(s)/TODOs: separate PR for the dropped-image-attachments issue (#5649).

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/cc-session-recreate`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude Code session resumption by verifying that saved sessions still exist.
  * Missing sessions are now recreated using the existing valid session identifier, avoiding repeated failed resume attempts.
  * Added handling for mismatched session files to improve session reliability.


M3gA-Mind:
## Maintainer review — three separate things to clear, all small

Checked against current `main` (`fa044d38`). The bug is real and still live: `driver.rs:231` on `main` is still

```rust
let is_new = !stored.as_deref().map(is_uuid_v4).unwrap_or(false);
```

so a persisted id whose CLI session was never created still resolves to `--resume`, and still poisons the thread permanently. Reproducing it against a real `claude` CLI and confirming the `--resume` → `--session-id` shim fixes it is the right kind of evidence, and it is what makes the diagnosis convincing rather than plausible.

### 1. Merge conflict — one file

`CONFLICTING` against current `main`, in `src/openhuman/inference/provider/claude_code/driver.rs` alone. `main` has since split that file's tests out (`driver.rs` went 736 → 501 lines; the `mod tests` block your `cc_session_exists_in_detects_present_and_absent` test lands in now lives in `driver_tests.rs`). So the resolution is:

- `cc_session_exists` / `cc_session_exists_in` and the `run_turn` changes replay onto `driver.rs` as-is — that region is untouched by the split.
- Your new test moves to `src/openhuman/inference/provider/claude_code/driver_tests.rs`.

### 2. The commit trailer is a merge blocker

Your commit carries:

```
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
```

This repo does not accept AI co-author trailers — it will block the merge. Please strip it while you are rebasing (`git commit --amend`, or `git rebase -i` if it is on more than one commit). Same applies to #5738, which has the same trailer.

### 3. The two red checks — one is yours, one is not

**`PR Submission Checklist` — yours, and easy.** The gate reports `6/7 items satisfied (1 unchecked, 0 N/A)`. The unchecked item is `Diff coverage ≥ 80%`, where you wrote an honest explanation but left the box empty. This gate counts boxes, so an explained exception still has to be **checked**, with the text beginning `N/A` — e.g.

```
- [x] **Diff coverage ≥ 80%** — N/A: the new helper is unit-tested; the changed lines
      inside `run_turn` spawn a subprocess and have no unit harness. Happy to add an
      integration test if reviewers prefer.
```

Leaving the substance exactly as you wrote it is fine — the gate only needs the box ticked and the `N/A` prefix. (Don't tick it as if coverage were met; the annotation is the honest form.)

**`Markdown Link Check` — not yours.** It fails on a dead link in `src/openhuman/memory/README.md:59` (`https://github.com/tinyhumansai/tinymemory/tree/main/core/src` → 404), a file this PR does not touch. A rebase onto current `main` should clear it; if it does not, it is a `main` problem and not a reason to hold this PR.

### On the fix itself

Reusing the stored id when recreating (rather than minting a fresh one) is the detail I'd call out as correct — it keeps the thread → session mapping stable, so recovery is invisible to everything upstream. And scanning every project dir rather than reconstructing the CLI's cwd-encoding is the right instinct: that encoding is an internal detail and would rot.

One thing worth a line in the description: `cc_session_exists` runs a `read_dir` on every turn for resumable threads. It is cheap next to spawning a process, as your comment says — but saying so in the PR body means the next reader doesn't have to rediscover it.

Both bot reviews are APPROVED with no unresolved threads. I have not pushed anything to your branch and I am not approving this.

senamakel:
way too many commits bro

PR #5652, opened 2026-08-21
fix(skills): accept allowed-tools as a comma string, not only a sequence

nocstah:
## Summary

- Fixes skills that declare `allowed-tools` as a comma-separated string (Claude Code's convention) failing to parse, which silently drops their tools and surfaces as "tool not available" at run time.
- Adds a string-or-sequence deserializer so `allowed-tools` accepts **either** a YAML list or a comma-joined scalar, plus a `tools` alias.
- Affects popular community catalogs (e.g. `antigravity-awesome-skills`) that ship the scalar form.

## Problem

`WorkflowFrontmatter::allowed_tools` deserializes as `Vec<String>`. A skill written as:

```yaml
allowed-tools: Bash, Read, Grep, Skill, WebFetch
```

fails with `invalid type: string, expected a sequence`, logged as `[skills] failed to parse frontmatter`. The frontmatter is rejected and the skill's declared tools are dropped, so running that skill reports its tools (Bash, etc.) as unavailable. This is common: skills authored for the `claude` CLI use the scalar form, and OpenHuman refreshes such catalogs by default.

## Solution

- `de_string_or_seq`: an untagged `Seq(Vec<String>) | Str(String)` deserializer. A scalar is split on commas and trimmed; empties dropped. A YAML list passes through unchanged.
- Add `alias = "tools"` so the common `tools:` key maps to the same field.
- No change to the serialized (write) form.

## Submission Checklist

- [x] Tests added or updated — `allowed_tools_accepts_yaml_sequence`, `allowed_tools_accepts_comma_string`, `allowed_tools_accepts_tools_alias`, `allowed_tools_defaults_empty_when_absent` (happy + the previously-failing scalar edge case).
- [x] **Diff coverage ≥ 80%** — the changed lines are the deserializer + field attrs, all exercised by the new tests. `pnpm test:rust` passes locally.
- [x] Coverage matrix updated — `N/A: parser-tolerance fix, no feature row change`.
- [x] Manual smoke checklist — `N/A: does not touch release-cut surfaces`.
- [x] Linked issue closed via `Closes #NNN` — `N/A` (no dedicated issue; addresses the tool-availability observation noted in #5648).

## Impact

- Skills subsystem, parse-time only. More skills load successfully; no change for skills that already used a YAML list. No migration, no new deps.

## Related

- Closes: N/A (addresses the "tool not available" observation in #5648)

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/skill-allowed-tools`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow configuration parsing to accept allowed tools provided as either lists or comma-separated values.
  * Added support for the existing `tools` alias and `allowed_tools` naming format.
  * Empty entries and extra whitespace are now handled automatically.
  * Missing tool settings continue to default correctly.
  * Workflow tool settings are interpreted consistently across supported configuration formats.
* **Tests**
  * Added coverage for supported formats, aliases, whitespace handling, empty entries, and default behavior.

### DIFF
--- a/src/openhuman/skills/ops_types.rs
+++ b/src/openhuman/skills/ops_types.rs
@@ -88,7 +88,13 @@ pub struct WorkflowFrontmatter {
     pub metadata: HashMap<String, serde_yaml::Value>,
     /// Tools the skill author asserts their instructions rely on
     /// (non-binding hint; the host decides what to expose).
-    #[serde(default, rename = "allowed-tools", alias = "allowed_tools")]
+    #[serde(
+        default,
+        rename = "allowed-tools",
+        alias = "allowed_tools",
+        alias = "tools",
+        deserialize_with = "de_string_or_seq"
+    )]
     pub allowed_tools: Vec<String>,
     /// Domain events that should activate this skill.
     ///
@@ -106,6 +112,32 @@ pub struct WorkflowFrontmatter {
     /// a migration warning when read.
     #[serde(flatten)]
     pub extra: HashMap<String, serde_yaml::Value>,
+}
+
+/// Deserialize `allowed-tools` from either a YAML sequence (`[Bash, Read]`) or a
+/// single scalar string in Claude Code's convention (`allowed-tools: Bash,
+/// Read, Grep`). A bare scalar is split on commas and trimmed. Without this,
+/// skills authored for the `claude` CLI — which writes a comma-joined string —
+/// fail frontmatter parsing, and their declared tools (Bash, etc.) get dropped,
+/// surfacing as "tool not available" at run time.
+fn de_string_or_seq<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StringOrSeq {
+        Seq(Vec<String>),
+        Str(String),
+    }
+    Ok(match StringOrSeq::deserialize(deserializer)? {
+        StringOrSeq::Seq(v) => v,
+        StringOrSeq::Str(s) => s
+            .split(',')
+            .map(|t| t.trim().to_string())
+            .filter(|t| !t.is_empty())
+            .collect(),
+    })
 }
 
 pub(crate) fn metadata_string(fm: &WorkflowFrontmatter, key: &str) -> Option<String> {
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
Maintainer review — no commits pushed.

**This one is in good shape and I could not find anything wrong with it.** Recording the checks I made so the next reader does not repeat them.

**The bug is real and still present on `main` (`fa044d388`).** `src/openhuman/skills/ops_types.rs:91` is still `#[serde(default, rename = "allowed-tools", alias = "allowed_tools")]` over a `Vec<String>`, so a scalar `allowed-tools: Bash, Read, Grep` still fails with `invalid type: string, expected a sequence`. Because the whole frontmatter parse is rejected on that error, the failure mode is exactly as described — the skill's declared tools vanish and only resurface as "tool not available" at run time, which is a long way from the cause.

**The `flatten` interaction is the thing I most expected to bite, and it does not.** `WorkflowFrontmatter` carries `#[serde(flatten)] pub extra`, and `#[serde(untagged)]` needs `deserialize_any`, which is a combination that breaks on non-self-describing formats. YAML is self-describing and `Rust Core Coverage` is green with all four new tests in the diff, so this is settled empirically rather than by argument.

**No field collision.** There is no other `tools` key on the struct, so `alias = "tools"` is unambiguous. It does change one behaviour worth stating out loud, though it is not a reason to hold the PR: a `tools:` key previously fell through `#[serde(flatten)]` into `extra` and raised the legacy-key migration warning; it is now consumed as `allowed_tools`. That is the intent, and `allowed-tools` remains the spec key — but `tools` is not part of the agentskills.io spec, so it is a tolerance being added rather than a bug being fixed. Fine either way; just do not let it become the documented key.

### The one red check is not this PR's

`Markdown Link Check` fails on **`src/openhuman/memory/README.md`** — a file this PR does not touch (the diff is one Rust file). It is a pre-existing repo-level failure, not a regression here.

### What it actually needs before merge

Nothing in the code. Only that **CI is stale**: the run on this PR is from 2026-08-21 against a `main` that has since absorbed the eviction wave (#5845–#5858, ~1,150 commits, files split repo-wide). The PR still reports `MERGEABLE` and `ops_types.rs` was not restructured, so I expect a clean re-run — but the green ticks currently on this page were earned against a tree that no longer exists. **Please push an empty commit or rebase to retrigger CI before approving**, so the merge decision rests on a run against current `main` rather than a two-week-old one.

One small style note for a follow-up, explicitly *not* a blocker: the new tests are an inline `#[cfg(test)] mod allowed_tools_tests`, whereas #5857 moved the repo to extracted sibling `*_tests.rs` files wired with `#[path]`. `ops_types.rs` has no sibling test file today, so nothing is inconsistent right now — but if the `Test Inventory` guard is ever tightened to require the extracted layout, this is where it would land.

senamakel:
> Please push an empty commit or rebase to retrigger CI before approving

Pushed the two regression tests CodeRabbit requested (trim/empty tokens handling and snake_case alias) on top of the existing fix. CI is re-running now on `75dc7305`.

PR #5654, opened 2026-08-21
docs: pin dead tinymemory core/src link to the vendored commit

JMak-Security:
## Summary

`src/openhuman/memory/README.md` links to `https://github.com/tinyhumansai/tinymemory/tree/main/core/src`, which now 404s -- caught by this repo's own `Markdown Link Check` CI job (`lychee`) on an unrelated PR (#5650).

## Root cause

`tinymemory`'s `main` branch has since restructured into a Cargo workspace: `core/` became `crates/tinymemory-core/`. A link pinned to `main` stopped resolving once that migration landed, even though nothing in *this* repo changed.

## Fix

`vendor/tinymemory` (this repo's submodule) is still pinned to commit `38a34d2ea10e7eedda1b50cdc786016c0f73b6dc`, from before the restructure. At that exact commit, `core/src` exists and contains all six files the README names by path (`sync_events.rs`, `remember.rs`, `ingest_pipeline.rs`, `traits.rs`, `preferences.rs`, `rpc_models.rs`) -- verified directly against that commit's tree via the GitHub API.

Pinning the link to that commit SHA instead of `main` fixes the 404 while staying accurate to what's actually vendored here, rather than repointing to tinymemory's current (and possibly further-diverged) directory layout.

## Testing

- Confirmed via the GitHub contents API that `core/src` at commit `38a34d2` contains exactly the six files/dirs the README describes.
- HTTP-checked both URLs directly:
  - Old (`tree/main/core/src`): **404**
  - New (`tree/38a34d2.../core/src`): **200**

Split out of #5650 as an unrelated, separately-scoped fix.

## Summary by CodeRabbit

* **Documentation**
  * Updated the `tinymemory-core` source reference to a fixed version for improved consistency and reproducibility.

### DIFF
--- a/src/openhuman/memory/README.md
+++ b/src/openhuman/memory/README.md
@@ -56,7 +56,7 @@ tinymemory_core::<domain>::*;` plus the handler/schema modules that name
 
 ## What lives in the extracted crate (for reference)
 
-See [`vendor/tinymemory/core/src/`](https://github.com/tinyhumansai/tinymemory/tree/main/core/src) for
+See [`vendor/tinymemory/core/src/`](https://github.com/tinyhumansai/tinymemory/tree/38a34d2ea10e7eedda1b50cdc786016c0f73b6dc/core/src) for
 the storage primitives (`store/`), ingestion queue (`ingestion/`), sync
 lifecycle types (`sync_events.rs`), remember classification (`remember.rs`),
 ingest orchestration (`ingest_pipeline.rs`), the `Memory`/`MemoryEntry`/etc.


M3gA-Mind:
Thanks for spotting the dead link, @JMak-Security — closing because it's since been fixed, and fixed a different way.

Commit `a86b8956e` (*docs: point the tinymemory source links at the crates/ layout*) already rewrote that line. `src/openhuman/memory/README.md` on `main` now points at `vendor/tinymemory/crates/tinymemory-core/src/` → `.../tree/1d6b997.../crates/tinymemory-core/src`.

Worth knowing for next time: this PR would have pinned the **old** `core/src` path at `38a34d2`, and `core/` doesn't exist at the submodule pin `6c7ea6a` (that URL 404s) — so it would have sent a reader to a layout the vendored tree no longer has. The dead link was real and worth reporting; this particular repair would have replaced it with a differently-dead one.

PR #5659, opened 2026-08-21
fix(release): sign updater artifacts with the CLI the release lane installs

### LINKED ISSUE #5658: Release Production fails on both macOS legs: upload-macos-artifacts.sh calls `cargo tauri`, which the workflow never installs
## Summary

Every **Release Production** run fails on both macOS legs at the last step of a ~50-minute job.
`scripts/release/upload-macos-artifacts.sh:48` invokes **`cargo tauri signer sign`**, but the
release workflow never installs `cargo-tauri`. The build itself does not need it — `build-desktop.yml`
builds with `pnpm tauri build` (the npm `@tauri-apps/cli`) — so the CLI is present as an npm binary
and absent as a cargo subcommand. The script reaches for the form that does not exist.

Production releases are blocked until this is fixed.

## Evidence

Run [32475039986](https://github.com/tinyhumansai/openhuman/actions/runs/32475039986), attempt 1,
`release` @ `2b220b826`, v0.63.15. Both `Desktop: aarch64-apple-darwin` (49m50s) and
`Desktop: x86_64-apple-darwin` (44m22s) failed at step 28, *Re-upload notarized macOS artifacts to release*:

```
[upload] Deleting old DMG asset from release...
[upload] Uploading notarized DMG...
[upload] Signing updater tarball with Tauri signer...
error: no such command: `tauri`
help: find a package to install `tauri` with `cargo search cargo-tauri`
##[error]Process completed with exit code 101.
```

Everything before it passed: `Sign and notarize macOS .app`, `Re-package DMG after notarization`,
`The staple and validate action worked!`, `[dmg] Verification successful: layout preserved.`
**This is not the Apple notary flake** seen previously — notarization succeeded on both legs.

- Failing call site: `scripts/release/upload-macos-artifacts.sh:48`
- Introduced by `3131b3829` *"feat(release): enhance macOS artifact upload and signing process"* — the only commit that has ever touched that line
- The workflow installs no cargo-tauri: `grep -E "cargo install|cargo-tauri|tauri-cli" .github/workflows/release-production.yml` → no matches
- Present on `main` as well as `release` — same line, same text

## Blast radius

1. **The macOS DMG upload happens *before* the failing signer call**, so a failed run still mutates
   the release: it deletes the previous DMG asset and uploads the new one, then dies. The
   `.app.tar.gz` and its `.sig` are never uploaded, and `Publish updater manifest (latest.json)`,
   `Publish draft release`, `Docker: tag :latest` and `Record Sentry deploy marker` are all skipped.
   The v0.63.15 draft is left with a partial asset set.
2. **The run cannot be re-run.** The `Remove release and tag if build failed` cleanup job deletes the
   `v0.63.15` tag after attempt 1. On attempt 2 both macOS jobs failed at `Checkout build ref` —
   `git fetch ... +refs/heads/v0.63.15*` found nothing (exit 1 after 3 tries) — and the cleanup job
   itself then failed at `Delete remote tag` because the tag was already gone. A fresh dispatch is
   required, not a re-run.

## Fix direction

Use the CLI the workflow actually installs. Either invoke it through pnpm/npx (matching
`pnpm tauri build` in `build-desktop.yml`), or install `cargo-tauri` in the macOS job — but the
first is consistent with how the rest of the lane already works.

Worth noting for whoever picks this up: the Linux equivalent of this call, `resign_artifact()` in
`scripts/release/strip-appimage-graphics-libs.sh:850-860`, **guards** it —
`command -v cargo-tauri || { warn; return; }` — and so degrades instead of failing the job. The
macOS script has no such guard. Whichever way the invocation is fixed, the two call sites should
agree; and a guard that silently skips signing on a release lane is worth a second look on its own.

## Acceptance criteria

- [ ] A Release Production run completes both macOS legs through artifact upload
- [ ] The published release carries `.dmg`, `.app.tar.gz` and `.app.tar.gz.sig` for both `aarch64` and `x86_64`
- [ ] `Publish updater manifest (latest.json)` runs instead of being skipped
- [ ] The invocation cannot silently no-op — if the signer is unavailable the step fails loudly, rather than uploading an unsigned or unsignatured updater artifact
- [ ] Both `cargo tauri signer sign` call sites (macOS + Linux) use the same, available invocation

## Follow-ups (not this issue, file separately if agreed)

- The failure-cleanup job deletes the tag even when release assets were already uploaded, making a re-run structurally impossible.
- `upload-macos-artifacts.sh` performs destructive work (delete + re-upload the DMG asset) before a step that can fail.

M3gA-Mind:
## Summary

- Fixes the invocation that has been failing every **Release Production** run on both macOS legs: `upload-macos-artifacts.sh` called `cargo tauri signer sign`, but no workflow installs `cargo-tauri`.
- Routes both updater-signing call sites through a new `scripts/release/tauri-signer.sh`, which resolves the CLI the lane actually installs (the npm `@tauri-apps/cli`) and **fails instead of skipping** when no signer is available.
- Fixes a second, quieter defect on the Linux leg: its `command -v cargo-tauri` guard has *always* been taken, so it published updater tarballs whose `.sig` covered pre-strip bytes.
- Adds a dependency-free regression test pinning the invocation shape and every failure path, wired into the existing release-script job.

**Release-blocking.** Production releases cannot complete until this lands.

## Problem

Run [32475039986](https://github.com/tinyhumansai/openhuman/actions/runs/32475039986), `release` @ `2b220b826`, v0.63.15. Both `Desktop: aarch64-apple-darwin` (49m50s) and `Desktop: x86_64-apple-darwin` (44m22s) died at the final step after ~50 minutes of work:

```
[upload] Signing updater tarball with Tauri signer...
error: no such command: `tauri`
help: find a package to install `tauri` with `cargo search cargo-tauri`
##[error]Process completed with exit code 101.
```

The lane builds with `pnpm tauri build` — the npm `@tauri-apps/cli` (`build-desktop.yml:436`). `cargo-tauri` is installed by no workflow at all; the only `cargo install` anywhere in `.github/` is `tauri-driver`, a different binary, in a Dockerfile. So the CLI is present as an npm binary and absent as a cargo subcommand, and the script reached for the form that does not exist.

The same wrong invocation exists on the Linux leg, where it fails differently and worse. `resign_artifact()` in `strip-appimage-graphics-libs.sh` guarded the call with `command -v cargo-tauri` and returned on failure. Since `cargo-tauri` is never installed, **that branch has always been taken**. The surrounding code rewrites the AppImage and then rebuilds its updater tarball:

```
mv "$rebuilt" "$original"          # AppImage bytes are now different
...
resign_artifact "$original"        # silently skipped
tar -czf "$tar" ...                # tarball rebuilt from the new bytes
resign_artifact "$tar"             # silently skipped again
```

so the `.sig` left on disk is the bundler's original, covering bytes that no longer exist. That is worse than a failed job: it fails on the client, after publication, for everyone who already has the app. Nothing in the lane would have reported it.

## Solution

A single shared helper, `scripts/release/tauri-signer.sh`, sourced by both scripts so they cannot drift apart again. It resolves the CLI once, in this order:

1. `app/node_modules/.bin/tauri` — what `pnpm install` puts in the workspace and what `pnpm tauri build` resolves to. Invoked directly, so neither pnpm nor a `package.json` script indirection can swallow `--private-key`.
2. `(cd app && pnpm exec tauri …)` — the same CLI through the workspace, if the `.bin` shim is not where we expect it.
3. `cargo-tauri` — for a developer machine that happens to have the cargo subcommand. CI never does.
4. Otherwise **fail**, naming all three paths it looked in.

It also removes any stale `.sig` *before* signing and verifies a non-empty one was produced afterwards, so neither an unsigned nor a mismatched updater artifact can reach a release.

Design decisions worth a reviewer's attention:

- **The Linux guard is deliberately not kept.** The issue asked whether it should stay; it should not. `command -v cargo-tauri || { warn; return; }` is precisely the mechanism that let a bad signature ship silently, and once the invocation is corrected the guard protects nothing — the npm CLI is present in the same job that just built the app with it. It is replaced by a hard failure.
- **One skip survives, and it is correct:** an unset `TAURI_SIGNING_PRIVATE_KEY` still returns early on the Linux path. That means an unsigned PR build, where the bundler produced no `.sig` either, so there is nothing to invalidate. It is now the *only* path that may skip, and it is commented as such.
- **macOS now passes `--password` too**, matching what Linux already did. An encrypted key with no password would otherwise stall a non-interactive runner on a passphrase prompt.

## Impact

Release/CI only — no runtime, app or user-facing code changes. Desktop macOS and Linux release legs are affected; Windows, mobile, web and CLI are untouched.

Security-relevant in the right direction: the Linux leg stops publishing updater artifacts with signatures that cannot verify, and no path can now publish an unsigned or unsignatured one.

`release` needs the same fix. The failing run was on `release` @ `2b220b826`, and the defective line is byte-identical on both branches — this PR targets `main` per the fleet standard, so please promote or cherry-pick it to `release` before the next cut.

## Related

- Closes: #5658
- Follow-up PR(s)/TODOs: two follow-ups noted in the issue are **not** addressed here and are worth filing separately — (a) the failure-cleanup job deletes the tag even when release assets were already uploaded, making a re-run structurally impossible; (b) `upload-macos-artifacts.sh` deletes and re-uploads the DMG *before* the step that can fail, so a failed run still mutates the release.

---

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `scripts/release/test-tauri-signer.sh`, 10 assertions: happy path, the invocation contract, and five failure paths (absent key, absent file, signer exits non-zero, signer exits 0 without a `.sig`, no CLI installed), plus the stale-`.sig` case. Run locally on macOS: 10 passed, 0 failed. Each assertion was checked against a reverted helper — restoring the warn-and-return guard, dropping the stale-`.sig` removal, and dropping `--password` each fail exactly one case and no others.
- [x] **Diff coverage ≥ 80%** — `N/A: shell-only change.` Every changed line is Bash in `scripts/release/` and `.github/workflows/`; neither Vitest nor cargo-llvm-cov instruments these, so `diff-cover` sees no changed lines to score. Behavioural coverage is the shell test above. I did **not** run `pnpm test:coverage` or `pnpm test:rust` — no Rust or TypeScript changed, and local builds are disallowed on this machine.
- [x] Coverage matrix updated — `N/A: behaviour-only change` to release scripts; no feature rows added, removed or renamed in `docs/TEST-COVERAGE-MATRIX.md`.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — `N/A: no matrix feature IDs apply`; this touches the release lane, not a product feature.
- [x] No new external network dependencies introduced — no new network calls. The signer runs a CLI that `pnpm install` already places in the workspace; the `gh` calls in the macOS script are unchanged.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no new manual step.` This restores a step that was already meant to run and is verified by CI, not by hand. The real proof is a green Release Production run, which needs a maintainer dispatch (see below).
- [x] Linked issue closed via `Closes #NNN` in the `## Related` section — `Closes #5658`, same repo, so the bare reference resolves correctly.

## Impact

See `## Impact` above.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/5658-tauri-signer-invocation`
- Commit SHA: `c37811ffd50c2b431353900501461917132a7969`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no TypeScript, JavaScript, JSON or CSS changed.` Prettier does not format `.sh`, and the one YAML file changed keeps the surrounding two-space style (verified by eye and by a tab scan: none).
- [x] Focused tests: `bash scripts/release/test-tauri-signer.sh` → **10 passed, 0 failed**. Also `bash -n` clean on all four affected scripts (`tauri-signer.sh`, `upload-macos-artifacts.sh`, `strip-appimage-graphics-libs.sh`, `validate-appimage-runtime.sh`), and I confirmed `strip-appimage-graphics-libs.sh` still sources cleanly, since `validate-appimage-runtime.sh` and `test-strip-appimage-rpaths.sh` both source it.
- [x] Tauri fmt/check (if changed): `N/A: no Tauri/Rust changed.`

### Validation Blocked

- `command:` `bash scripts/release/test-strip-appimage-rpaths.sh`
- `error:` `patchelf` is not installed on this macOS machine; the test is Linux-oriented and runs on `ubuntu-latest` in CI.
- `impact:` The pre-existing AppImage test could not be re-run locally after the edit to `strip-appimage-graphics-libs.sh`. The edit is confined to `resign_artifact()`, which that test does not exercise, and the file still sources cleanly. CI covers it.
- `command:` Release Production workflow dispatch
- `error:` requires maintainer permissions and repository secrets.
- `impact:` **This fix is not verified end-to-end.** I verified the invocation itself against the real `@tauri-apps/cli` 2.11.4 — generated a throwaway minisign keypair, signed a file through the helper, and got a valid 400-byte `.sig` — and confirmed `tauri signer sign` exposes exactly `--private-key` and `--password`. What I cannot verify is a full macOS release leg. Please dispatch a Release Production run before trusting it.

### Behavior Changes

- Intended behavior change: updater signing uses the npm Tauri CLI at both call sites and fails the job when no signer is available, instead of erroring out on macOS and silently skipping on Linux.
- User-visible effect: none in the app. Downstream, macOS releases regain their `.app.tar.gz` + `.app.tar.gz.sig` assets (and `Publish updater manifest (latest.json)` stops being skipped), and Linux AppImage updater artifacts start carrying a signature that actually matches their bytes.

### Parity Contract

- Legacy behavior preserved: the DMG upload path, the tarball layout, asset names and the `gh release upload --clobber` calls are untouched. The unsigned-lane skip (no `TAURI_SIGNING_PRIVATE_KEY`) behaves exactly as before.
- Guard/fallback/dispatch parity checks: the removed `command -v cargo-tauri` guard is intentionally not replaced by an equivalent — see the second bullet under `## Solution`. `cargo-tauri` survives only as the last resolution step, so a developer machine that has it keeps working.

- Duplicate PR(s): none — `gh api search/issues` for PRs referencing 5658 returns 0 results, and the issue timeline has no linked PR.
- Canonical PR: this one.
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release artifact signing reliability across AppImage and macOS updater packages.
  * Signing now reports failures clearly instead of silently continuing.
  * Prevented stale or missing signature files from being published.
  * Added fallback support for available Tauri signing tools.

* **Tests**
  * Added automated coverage for successful signing, invalid inputs, signer failures, unavailable tools, and stale signatures.
  * Expanded release validation to syntax-check scripts and run signing safeguards.

### DIFF
--- a/.github/workflows/pr-quality.yml
+++ b/.github/workflows/pr-quality.yml
@@ -94,7 +94,12 @@ jobs:
           fetch-depth: 1
       - name: Install patchelf
         run: sudo apt-get update -qq && sudo apt-get install -y -qq patchelf
-      - name: Syntax check release script
-        run: bash -n scripts/release/strip-appimage-graphics-libs.sh
+      - name: Syntax check release scripts
+        run: |
+          bash -n scripts/release/strip-appimage-graphics-libs.sh
+          bash -n scripts/release/upload-macos-artifacts.sh
+          bash -n scripts/release/tauri-signer.sh
       - name: Run AppImage RPATH sanitize + libxdo guard test (issue #3224)
         run: bash scripts/release/test-strip-appimage-rpaths.sh
+      - name: Run updater-signing guard test (issue #5658)
+        run: bash scripts/release/test-tauri-signer.sh
--- a/scripts/release/strip-appimage-graphics-libs.sh
+++ b/scripts/release/strip-appimage-graphics-libs.sh
@@ -32,6 +32,8 @@ set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 APPIMAGE_RUNTIME_VALIDATOR="${APPIMAGE_RUNTIME_VALIDATOR:-$SCRIPT_DIR/validate-appimage-runtime.sh}"
+# shellcheck source=scripts/release/tauri-signer.sh
+. "$SCRIPT_DIR/tauri-signer.sh"
 
 EXCLUDE_PATTERNS=(
   'libGL.so.*'
@@ -844,19 +846,22 @@ strip_one_appimage() {
 
 resign_artifact() {
   local file="$1"
+  # No key configured means this is an unsigned lane (a PR build): the bundler
+  # produced no .sig either, so there is nothing to invalidate. Skipping is
+  # correct here and is the ONLY case in which signing may be skipped.
   if [ -z "${TAURI_SIGNING_PRIVATE_KEY:-}" ]; then
     return
   fi
-  if ! command -v cargo-tauri >/dev/null 2>&1; then
-    echo "[strip-libs] WARNING: cargo-tauri not on PATH; cannot re-sign $file" >&2
-    return
-  fi
   echo "[strip-libs] Re-signing $file"
-  rm -f "$file.sig"
-  cargo tauri signer sign \
-    --private-key "$TAURI_SIGNING_PRIVATE_KEY" \
-    --password "${TAURI_SIGNING_PRIVATE_KEY_PASSWORD:-}" \
-    "$file" >/dev/null
+  # Previously guarded by `command -v cargo-tauri` with a warn-and-return. CI
+  # never installs cargo-tauri, so that branch was always taken and this leg
+  # shipped a .sig covering the pre-strip bytes. Stripping rewrites the
+  # AppImage, so a signature that is merely left behind cannot verify -- fail
+  # the job instead of publishing it (#5658).
+  if ! tauri_signer_sign "$file"; then
+    echo "[strip-libs] ERROR: could not re-sign $file after stripping; refusing to publish an artifact whose signature does not match its bytes" >&2
+    exit 1
+  fi
 }
 
 main() {
--- a/scripts/release/tauri-signer.sh
+++ b/scripts/release/tauri-signer.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+# Shared Tauri updater-signing helper for the release scripts.
+#
+# Sourced, not executed — it only defines `tauri_signer_sign` and deliberately
+# sets no shell options, so it inherits the caller's `set -euo pipefail`.
+#
+# Why this exists (#5658): the release lane builds the desktop app with the npm
+# `@tauri-apps/cli` — `pnpm tauri build` in .github/workflows/build-desktop.yml
+# — and never installs the `cargo-tauri` crate. Both release scripts none the
+# less reached for `cargo tauri signer sign`, the one form that is not present
+# on the runner:
+#
+#   * upload-macos-artifacts.sh called it unguarded, so every Release Production
+#     run died with `error: no such command: tauri` (exit 101) at the final
+#     upload step of a ~50-minute job, after the DMG had already been replaced.
+#   * strip-appimage-graphics-libs.sh guarded it with `command -v cargo-tauri`
+#     and returned on failure. That guard has always been taken, so the Linux
+#     leg silently published an updater tarball whose .sig still covered the
+#     pre-strip bytes — a signature that cannot verify on an installed client.
+#
+# Resolution order below prefers the CLI the lane actually installs, and the
+# helper FAILS instead of skipping when no signer is available. An updater
+# artifact published without a valid signature is worse than a failed job:
+# clients verify the signature, so a bad .sig breaks updates for everyone who
+# already has the app, and it does so silently at release time.
+
+# Sign "$1" with the Tauri updater key, writing "$1.sig" alongside it.
+#
+# Requires TAURI_SIGNING_PRIVATE_KEY to be set; TAURI_SIGNING_PRIVATE_KEY_PASSWORD
+# is optional and may be empty for an unencrypted key. The password is always
+# passed explicitly so an encrypted key cannot stall a non-interactive runner on
+# a passphrase prompt.
+#
+# Returns 0 only when a non-empty .sig was produced; non-zero otherwise. Callers
+# that legitimately have nothing to sign (an unsigned PR build with no key
+# configured) must decide that *before* calling this.
+tauri_signer_sign() {
+  local file="$1"
+  local key="${TAURI_SIGNING_PRIVATE_KEY:-}"
+  local password="${TAURI_SIGNING_PRIVATE_KEY_PASSWORD:-}"
+
+  if [ -z "$key" ]; then
+    echo "[tauri-signer] ERROR: TAURI_SIGNING_PRIVATE_KEY is not set; refusing to publish an unsigned updater artifact: $file" >&2
+    return 1
+  fi
+  if [ ! -f "$file" ]; then
+    echo "[tauri-signer] ERROR: nothing to sign, no such file: $file" >&2
+    return 1
+  fi
+
+  local script_dir repo_root app_dir app_bin
+  script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+  repo_root="$(cd "$script_dir/../.." && pwd)"
+  app_dir="$repo_root/app"
+  app_bin="$app_dir/node_modules/.bin/tauri"
+
+  # Drop any stale signature first so the check below cannot pass on a leftover
+  # .sig from an earlier bundle. This is the Linux failure mode exactly.
+  rm -f "$file.sig"
+
+  local sign_status=0
+  if [ -x "$app_bin" ]; then
+    # The binary `pnpm install` puts in the workspace — what `pnpm tauri build`
+    # resolves to in build-desktop.yml. Invoked directly so neither pnpm nor a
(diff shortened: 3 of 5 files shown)


M3gA-Mind:
CI has closed one of the two verification gaps listed under **Validation Blocked**.

The `AppImage RPATH + libxdo guard` job passed (16s), and both release-script tests ran green on `ubuntu-latest`:

- `Run updater-signing guard test (issue #5658)` — **10 passed, 0 failed**, including `fails loudly rather than skipping when no CLI is installed` and `removes the stale .sig instead of shipping it`.
- `Run AppImage RPATH sanitize + libxdo guard test (issue #3224)` — passed. This is the pre-existing test I could not run locally (no `patchelf` on macOS), so the edit to `resign_artifact()` is now confirmed not to have disturbed it.

The second gap stands unchanged: **a full Release Production run still has not been exercised.** That needs a maintainer dispatch, and it is the only thing that proves the macOS leg gets past step 28.

PR #5663, opened 2026-08-21
refactor(agent): move multimodal resolution into tinyagents

senamakel:
## Summary

- Extracts the host-agnostic half of `agent/multimodal.rs` into
  `tinyagents::harness::multimodal` (tinyhumansai/tinyagents#118) and rewrites
  the OpenHuman module as the adapter over it: **1,801 → 725 lines**.
- The crate now owns marker parsing, `data:` URI decoding, MIME detection, the
  limit clamps and the rendered payload. OpenHuman keeps the `ChatMessage`
  adapters, the config mapping, the runtime-proxy client, the `documents` PDF
  extractor with its deadline, and the whole on-disk attachment stash.
- **No behaviour change and no wire-format change.** Every public function keeps
  its name and signature, so all 16 call sites are untouched.
- Reverses the `multimodal.rs` "stays host-side" row in
  `docs/specs/plan-agents.md` §3 and records `multimodal` as the third landed
  Phase 5 family.
- **Zero dependency-graph movement**: `base64`, `flate2` and `sha2` were already
  resolved in the kernel profile, so `check-kernel-floor.sh` does not move.

## Problem

`docs/specs/plan-agents.md` §3 listed `multimodal.rs` under "Stays in OpenHuman
as the adapter layer". That row was half right: the `ChatMessage` adapters, the
config mapping, the proxy client, the PDF extractor and the attachment stash
genuinely are host concerns. What it missed is that ~1,300 lines *underneath*
them are not host concerns at all — parsing `[IMAGE:…]`, decoding a `data:` URI,
sniffing PNG magic, clamping `max_files` and rendering a `[FILE-EXTRACTED]`
block are the same for any host that carries attachments.

Meanwhile the two Phase 5 families that were actually unblocked (`session/turn`,
`session/{builder,runtime,types}`) are gated on Phase 2's parity soak, which
needs a shipped release to produce data. This is a family that was never
blocked — it was only mis-dispositioned.

## Solution

The split follows the rule the `artifact_offload` and `run_queue` moves
established: the crate owns the mechanics, the host keeps the irreducible
remainder. Two seams carry the host's half.

**`TextExtractor`, for PDF text.** `DocumentsTextExtractor` wraps the `tinydocs`
module and keeps `PDF_EXTRACTION_TIMEOUT` host-side, because the cost of parsing
a document is set by the document and only the host knows how long an
attachment is worth waiting for. It declares `handles(mime) == "application/pdf"`
and nothing else — which matters more than it looks: the module runs
out-of-process, so offering it every `.zip` and `.xlsx` speculatively would cost
a bus round trip and a copy of the file per attachment to reach a refusal that
was knowable from the MIME type. The old code only ever called the PDF path;
`handles` is what preserves that.

**Config mapping, per §4.1 Option A.** `image_limits` / `file_limits` map
`MultimodalConfig` / `MultimodalFileConfig` onto the crate structs, following
the `session_config_from` precedent. They are plain field copies because the
clamping stayed crate-side — if a rule ever grows there, it belongs in the crate.

The one piece of hardening that needed care is the `max_files: 0` sentinel from
`for_untrusted_channel_input()`. The crate's `effective()` clamps to `1..=16`
exactly as the host's did, so the crate exposes `files_disabled()` as a separate
pre-clamp check and both sides consult it before any read. Tests on both sides
pin the ordering; the four untrusted-channel-input tests here are unchanged and
still pass.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) —
      44 new crate-side tests upstream, covering the sentinel ordering, both MIME
      precedences, the truncation budget, and the malformed-marker cases. The 50
      host tests are unchanged in intent; six that only exercised primitives now
      living in the crate were dropped (covered upstream), and one was narrowed
      to its host half.
- [x] **Diff coverage ≥ 80%** — the host diff is almost entirely deletion plus
      delegation; every retained public function is exercised by the 50 tests in
      `multimodal_tests.rs`, which pass under both the product feature set and
      `--no-default-features`.
- [x] Coverage matrix updated — `N/A: no feature row changed`. Row 6.3.9
      references `multimodal_tests.rs::extract_image_placeholders_pulls_att_tokens_in_order`,
      which still exists and still passes.
- [x] All affected feature IDs listed under `## Related`
- [x] No new external network dependencies introduced — `N/A: no new
      dependency at all; the crate borrows the host's existing proxy client`
- [x] Manual smoke checklist updated — `N/A: no user-visible behaviour change`
- [x] Linked issue closed — `N/A: no issue; this reverses a spec row, updated in
      the same PR`

## Impact

**Runtime:** none. The wire format is byte-identical — same markers, same
`[FILE-EXTRACTED]` / `[FILE-ATTACHED]` rendering, same placeholder tokens, same
error strings. Persisted transcripts and stashed attachments are unaffected.

**Dependencies:** none. `tinyagents` gains `base64` and `flate2` as optional deps
behind its default-off `multimodal` feature; OpenHuman turns that feature on
(it is kernel surface — web chat and the agent harness both resolve markers on
every turn). All three of `base64`, `flate2`, `sha2` were already in the kernel
profile, so `scripts/check-kernel-floor.sh` reports **no movement**, and
`app/src-tauri/Cargo.lock` gained exactly two dependency-edge lines and zero
packages.

**Security:** the untrusted-channel hardening is preserved and is now pinned on
both sides of the seam rather than one. See the sentinel discussion above.

## Related

- Feature IDs: 6.3.9 (vision sub-agent reads attached images)
- Depends on: tinyhumansai/tinyagents#118 — **merged** as `27e5a7d`. The
  `vendor/tinyagents` gitlink has been re-pointed from the PR branch to that
  commit on `main`, so this PR no longer depends on anything unmerged.
- Spec: `docs/specs/plan-agents.md` §3, §5 Phase 5
- Closes:
- Follow-up PR(s)/TODOs: none outstanding.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `agent-multimodal-to-tinyagents`
- Commit SHA: see branch head

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no frontend files changed`
- [x] `pnpm typecheck` — `N/A: no TypeScript changed`
- [x] Focused tests: `cargo test --lib agent::multimodal` — 50 passed under the
      product feature set and 50 under `--no-default-features`
- [x] Rust fmt/check: `cargo fmt --check` clean; `cargo check` clean under the
      product feature set and under `--no-default-features`
- [x] Tauri fmt/check: `N/A: app/src-tauri unchanged apart from its lockfile`

Also run: `scripts/check-kernel-floor.sh` (within ratchet) and
`scripts/ci/check-feature-forwarding.mjs` (OK).

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: none — this is a relocation.
- User-visible effect: none.

### Parity Contract

- Legacy behavior preserved: every public function keeps its name, signature and
  semantics; the marker wire format, payload rendering and error strings are
  byte-identical.
- Guard/fallback/dispatch parity checks: the `max_files == 0` untrusted-channel
  sentinel is checked before the clamp on both sides of the seam (host
  `files_disabled()` pre-check plus the crate's own); PDF extraction failure,
  timeout and a `documents`-less build all still degrade to
  `FilePayload::Reference` rather than failing the turn.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Multimodal agent functionality is now available across supported builds.
  * Improved processing of image and file attachments, including shared limits, placeholders, payload handling, and composition.

* **Bug Fixes**
  * Improved consistency in attachment resolution, MIME handling, and file-disabled behavior.
  * Enhanced PDF attachment processing and timeout handling.

* **Documentation**
  * Updated the agent migration plan to reflect multimodal functionality as completed and integrated.
  * Updated documentation to reflect the reduced set of available command-line tools.


senamakel:
Addressed the coderabbitai nitpick on `src/openhuman/agent/multimodal.rs`:

- Moved `Config::load_or_init()` inside the `PDF_EXTRACTION_TIMEOUT` deadline so a slow/contended config load can no longer delay the turn outside the bound — the `bounded by PDF_EXTRACTION_TIMEOUT` claim now holds for the whole operation.
- Gated the intra-doc link so `[PDF_EXTRACTION_TIMEOUT]` is only linked when the `documents` feature is enabled (the const is feature-gated); without the feature the doc reads `bounded by the module's own deadline` instead of linking to a non-existent item.

Verified with `cargo check --features documents` and `cargo check --no-default-features --features modules`. Both compile clean.

PR #5667, opened 2026-08-21
refactor(orchestration): finish WP-5 — delete the worktree duplicate, move claim arbitration and task-store selection to tinyagents

senamakel:
Finishes the subagent / parallel-subagent half of WP-5
(`docs/tinyagents-migration-plan-2026-07-22.md`), which
`docs/tinyagents-port-plan.md` tracks as Phase 4.

Pairs with tinyhumansai/tinyagents#119 — that must merge first; this PR bumps
the `vendor/tinyagents` gitlink onto it. No crates.io release is involved: the
root manifest already patches `tinyagents` to the vendored path.

**Net ~270 fewer lines host-side**, and the fan-out is no longer two implementations.

## The headline: a 618-line module that was already in the crate

`orchestration/worktree.rs` defined its own `BaseRef`, `WorktreeStatus`,
`GitWorktreeIsolation`, `create`/`list`/`status`/`diff_summary`/`remove`,
`detect_overlaps`, `validate_repo_root` and `sanitize_run_id` — near-verbatim
twins of `tinyagents::harness::workspace::git`. `detect_overlaps` and
`detect_worktree_overlaps` matched down to the `BTreeSet` dedupe, and the two
error enums carried identical message strings.

Deleted and re-exported under the historical OpenHuman names, so the RPC
schemas, tools and tests that spell `worktree::create` / `WorktreeStatus` are
untouched. 618 → 242 lines.

What stays is what is actually OpenHuman's: `OpenHumanWorktreeIsolation` (the
`openhuman.worktree:{agent}:{run_id}` policy-id convention and the
`DomainEvent::Workspace*` bus emissions).

**The risk here was the RPC wire shape** — `worktree_schemas.rs` serializes
`WorktreeStatus` straight to the desktop UI, where a silent field rename shows
up as an empty panel rather than a red test. So
`worktree_status_serializes_with_stable_camel_case_keys` was written and made to
pass *before* the swap, and passes unchanged after.

## Write-safety arbitration moves to the crate

`prepare_spawn_parallel_tasks_from_defs` is now two passes: OpenHuman policy
admits or rejects each task in its own vocabulary, then a single
`plan_shared_workspace_dispatch` call arbitrates every admitted claim.

The host keeps everything that is a product decision — the `files:` parameter
syntax, `shared_workspace_write_capable_tools`, the `integrations_agent` toolkit
rule, the subagent allowlist, `ParallelTaskRejectionKind`, and every rejection
sentence (the crate returns `ClaimConflict` as data precisely so the
`isolation="worktree"` remedy phrasing stays here).

**Execution deliberately stays host-side.** Today the host serializes the whole
batch when any worker needs serializing; a crate-side executor running only the
claimed workers serially would be a *narrower* guarantee than what ships. The
crate contributes the plan, not the scheduler — so the diff is "who computed
this decision", which is exhaustively unit-testable.

Parity gate: `spawn_parallel_agents_tests.rs` passes **unedited** across the
cutover. Three assertions were added first, pinning the mixed-batch dispatch
sequence, disjoint-ownership admission, and directory containment.

## Detached-store selection and orphan reconciliation

`running_subagents.rs` loses the `DetachedTaskStore` enum and its 12-arm
delegating `TaskStore` impl, the hand-rolled `TASK_STORES` map, and the
reconcile state machine — replaced by the crate's `TaskStoreRegistry` and
`reconcile_orphaned_tasks`. The fallback ladder is ported exactly (a
`create_dir_all` failure degrades to memory just as an unreadable log does; a
read-only workspace must stay able to spawn work), and lock poisoning is typed
rather than `.expect()`ed.

The reason string and the `publish_subagent_failed` lifecycle event stay host
side. All 17 focused `running_subagents` tests pass unedited.

## Two smaller things

- **Deleted `tinyagents/subagent_graph.rs`.** Six graph nodes whose bodies only
  pushed their own name onto a `Vec`, compiled and executed on *every* sub-agent
  spawn. It scaffolded a per-phase cutover that this work concludes should not
  happen (below).
- **`extract_tool.rs` fan-out** moved from hand-rolled
  `buffer_unordered(3)` + index-tag-and-re-sort onto `map_reduce`, which returns
  input-ordered outcomes. The per-chunk provider error stays the fan-out's
  *item* rather than its error, so one failed chunk still drops with a warning
  instead of aborting its siblings.

## What is deliberately NOT moved

**`harness/subagent_runner/` (7,471 LOC) stays host-owned.** Its own docs call
it the OpenHuman *build pipeline*, and that is accurate. The generic contract it
would map onto already exists as `tinyagents::harness::host::HostCapabilities`
(`ContextComposer` / `DefinitionRegistry` / `SecurityGate` / `ModelResolver`) —
exactly the phases the runner names — and OpenHuman does not implement it. So
the real question is whether the host should implement those four traits, not
whether to push product policy across the GPL boundary. That is its own
design-gated package. The module doc and a ledger row now say so, replacing a
comment that promised the opposite.

## Corrections to earlier notes

Two things previously recorded as defects turned out not to be, and are now
documented rather than "fixed":

- `orchestration/ops.rs` reporting `worktree_path: None` is **correct by
  construction**, not a dropped value. Those fields describe a worker's *own*
  isolated checkout; `spawn_parallel_agents` populates them only from a
  descriptor it freshly created, and this path only ever inherits the parent's.
  (`tinyagents-port-plan.md:165` reads as a defect; it is not.)
- `REGISTRY_SOFT_CAP` is already crate-owned as policy — it is passed into
  `DetachedTaskRegistry::new`, which performs the sweep. Only the number is
  OpenHuman's.

Also noted in the ledger: `enforce_workspace_path` has no callers anywhere. It
is kept — removing a public fail-closed gate is a separate judgement from this
migration — but recorded so the next reader doesn't have to rediscover it.

## Testing

- `cargo check` clean in **both** Cargo worlds (root and `app/src-tauri`);
  `cargo tree -i tinyagents` shows exactly one node in each.
- `scripts/test-rust-with-mock.sh`: **12116 passed, 2 failed**. The two are
  `git_operations::tests::{allows_readonly_ops_in_readonly_mode,
  not_in_git_repo_returns_error}` and are **pre-existing** — verified by checking
  `src/` out at the merge-base and re-running, where they fail identically. They
  exercise the filesystem git tool, which this diff does not touch.
- `cargo check --no-default-features --all-targets`: the only failures are the
  pre-existing `build_core_http_router` ones in `tests/worker_*_e2e.rs` /
  `live_routing_e2e.rs` (unrelated; zero references in this diff). Worth running
  by hand — CI's smoke lane is `cargo check` lib-scoped and its test filter
  covers neither `agent::orchestration::` nor `agent::harness::subagent_runner::`.

## Note on commits

The granular history is this repo's auto-commit checkpointing hook, not
hand-authored steps; read the diff as one change.

## Summary by CodeRabbit

* **New Features**
  * Parallel agent tasks now coordinate workspace ownership, reducing conflicting file changes and safely separating isolated work.
  * Task progress uses durable workspace storage when available, with graceful in-memory fallback.
  * Git worktree operations provide clearer workspace-boundary errors and stable status serialization.

* **Bug Fixes**
  * Failed extraction chunks no longer stop successful sibling work; results remain ordered.
  * Detached task recovery now handles missing or orphaned task records more gracefully.

* **Tests**
  * Added coverage for parallel dispatch, ownership conflicts, and worktree status contracts.


senamakel:
Addressing the CodeRabbit review findings (all non-blocking, CHILL profile):

**1. CWE-532 — hook denial logging (`runner.rs:392-401`)** — Already addressed in this PR. The spawn-denial path logs a fixed sentence (`[subagent_runner] spawn denied by a configured hook`) with only `agent_id` and `task_id`; the hook-supplied `reason` is never written to logs — it is surfaced to the caller via `SubagentRunError::HookDenied(reason)`. The rationale is documented in the comment at L392-395. This finding is stale against the current code.

**2. `directory_ownership_contains_files_beneath_it` test** — Applied. The test now asserts the directory-level owner is admitted with `WorkerDispatchMode::SerialSharedWorkspaceWrite`, and the file-beneath-directory task is rejected with `ParallelTaskRejectionKind::RequiresIsolation`, agent id `nested`, the rejected task's ownership claim (`files: src/a.rs`), and an error naming the contended owner directory (`'src'`). Verified passing locally.

**3. `mixed_batch_dispatch_modes_and_claim_conflicts_are_stable` test** — Applied. Added a leading `outside` agent (defined but omitted from the parent allowlist) that is policy-rejected with `OutsideAllowlist` *before* any claim is admitted, so the batch exercises that `admitted_index` is not advanced by an earlier rejection. Per-entry assertions cover each task's identity, ownership, rejection kind, and `WorkerDispatchMode`; the later writer/clasher claim-conflict checks are preserved (clasher now at index 4, since the leading rejection consumes no admission slot). Verified passing locally.

Separately, the `Rust Core Coverage` CI failure is resolved: merging `main` brought in #5682, which removed the stale `meetAutoOrchestratorHandoff == Some(false)` assertion — the snapshot now asserts `runtime.service` / `runtime.localAi` instead (the field is `None` now that Meet is removed from the product).

senamakel:
**CI fix — `Rust Core Coverage (cargo-llvm-cov)` was failing on the coverage-presence gate.**

The gate (`scripts/ci/assert-coverage-presence.sh`) reports that `src/openhuman/runtime/client/disabled.rs` — the `#[cfg(not(feature = "modules"))]` stand-in for the runtime client facade introduced by this PR's `runtime/client/` refactor — produced no coverage records.

That is expected, and now documented: `modules` is in **both** `[features] default` and `scripts/ci/product-features.txt`, so the coverage lane (which compiles `default + product-features.txt`) always builds with the gate **ON**, and the OFF-branch stub is never compiled in that lane. It is the same shape as the `stub.rs` structural category the script already excludes (facade stubs compile only in the OFF direction of their gate), but this file is deliberately named `disabled.rs` because it answers as a runtime that is disabled, not as a generic facade stub. The `disabled.rs` code path *is* compiled and exercised in the disabled-build lane (`--no-default-features`), so it is not unverified code — it is just not compiled under the product feature set.

Per the allowlist's own policy, the entry is recorded in `scripts/ci/coverage-presence-allowlist.txt` with the gate and rationale written inline, so a future reader can tell "excluded on purpose" from "forgotten".

PR #5672, opened 2026-08-22
fix(tools): refuse to run git under untrusted workspace repo config

### LINKED ISSUE #5494: workspace_state: run_git executes with agent-writable repo config unscrubbed
## Where

`src/openhuman/tools/impl/system/workspace_state.rs` — `run_git`, the helper behind the `read_workspace_state` tool.

## What

`run_git` spawns git bare:

```rust
tokio::process::Command::new("git").args(args).current_dir(dir).output().await?
```

No `GIT_CONFIG_NOSYSTEM`, no `-c` overrides, no environment scrub — and `dir` is the agent's own workspace, the directory `file_write` writes into. Several git config keys name a command to run (`core.fsmonitor`, which `git status` invokes; `core.pager`; `diff.external`; the `*.textconv` family), so a `.git/config` the agent authored decides what a "read-only overview" executes.

The same crate already treats spawning a process for an agent as needing a scrub: `shell.rs` calls `cmd.env_clear()`, and `node_exec.rs` / `python_exec.rs` do too. This path just never got one.

## Why the obvious scrub is not the fix

The exposure is the **repository** config, in a directory the agent can write. `GIT_CONFIG_NOSYSTEM` and `GIT_CONFIG_GLOBAL` do not reach it. Blocking a list of known-bad keys is a denylist against a surface that grows with git releases. The durable shape is to **refuse unknown config** — run with an explicitly constructed config (e.g. `GIT_CONFIG_COUNT` / `-c` supplying exactly what is needed, plus `GIT_CONFIG_NOSYSTEM=1` and `GIT_CONFIG_GLOBAL=/dev/null`) and to not honour repository config keys that can name a command, rather than enumerating the ones that currently can.

Worth doing next to `env_clear()`, so this path matches the other three process spawns in the crate.

## Why it matters downstream

OpenCompany (`tinyhumansai/opencompany`) has had to classify `read_workspace_state` as `Reach::Consequence` in its approval policy — it parks under `supervised` and is denied under `readonly` — because a "read" that can execute a command is the `shell` shape wearing a read's name. See tinyhumansai/opencompany#459 and its PR #606. That is a deliberate stopgap that costs an approval on a routine orientation step, and its stated revert condition is exactly this issue: once a hardened `run_git` is vendored, the classification goes back to `Reach::Nothing`.

The comment in `src/policy/consequence.rs` links here, so the person bumping the openhuman pin can find the condition.

Note `git_operations`' `run_git_command_in` in the same crate is the same unscrubbed shape; it is already gated downstream, but the hardening belongs in both.

## Done when

- `run_git` (and `run_git_command_in`) run git with repository config that cannot name an executable.
- A test writes a hostile `.git/config` into a workspace and asserts the command is not run.

rsd-darshan:
## Summary

- `git_operations`'s `run_git_command_in` spawned `git` with no config hardening at all, so a `.git/config` written into the agent's own workspace decided what `git` executed on the next `status`/`log`/`diff`/`commit`/`add`/`checkout`/`stash`.
- Adds the same allowlist-based hardening already shipped for `read_workspace_state` in #5493: refuse to run when the repository's local config sets anything outside a known-inert allowlist, neutralise the known command-valued keys with `-c` as defence in depth, and close the system/global config files plus command-valued env vars.
- Adds tests for the exploit repro (`core.fsmonitor`), an ordinary repository, an LFS clone, `credential.helper`, and a write operation (`commit`).

## Problem

`run_git_command_in` is the single function backing every operation this tool exposes:

```rust
let output = tokio::process::Command::new("git")
    .args(args)
    .current_dir(cwd)
    .output()
    .await?;
```

`cwd` is the agent's own workspace — the same directory `file_write` writes into, and several git config keys name a command that git then executes: `core.fsmonitor` (run by `git status`), `core.sshCommand`, `core.pager`, `core.editor`, `diff.external`, `credential.helper`, and the `filter.*.process`/`*.clean`/`*.smudge` family. An agent that writes a `.git/config` into its workspace controls what git runs the next time this tool touches that repo — on `status`, which every one of the eight operations reaches through this one function.

This is the sibling issue #5493 flagged and deliberately left alone: `read_workspace_state`'s own `run_git` had the identical shape and was hardened there, but `git_operations`' `run_git_command_in` was called out explicitly as unfixed, to keep that PR reviewable. Issue #5494 tracks finishing it here.

## Solution

- `ALLOWED_REPO_CONFIG`: an allowlist of the config keys `git init`/`git clone` write plus the handful of inert settings an ordinary repository carries (`core.autocrlf`, `gc.auto`, etc.). Anything outside it — including a nominally "read-only" setting like an LFS filter driver — causes `run_git_command_in` to refuse before it ever calls `git`, naming the offending key in the error.
- `NEUTRALISED_CONFIG`: the known command-valued keys cleared via `-c` on the command line, which outranks every config file. Defence in depth, not the guarantee — the allowlist is.
- `hardened_git`: sets `GIT_CONFIG_NOSYSTEM`/`GIT_CONFIG_GLOBAL` and removes the command-valued `GIT_*` env vars before every invocation.
- The allowlist and neutralised-config lists are the same ones #5493 already proved correct for `read_workspace_state`; kept in this file rather than factored into a shared module, matching how this crate already handles per-tool process hardening (`shell.rs`/`node_exec.rs`/`python_exec.rs` each do their own `env_clear()` rather than sharing one).

## Impact

- Desktop/CLI agent tool surface only. No RPC or schema change.
- Security: closes a code-execution path reachable through a tool whose write operations are already gated by autonomy tier, but whose read operations (`status`, `diff`, `log`, `branch`) were not.
- No migration/compatibility impact — an ordinary repository's config is unaffected; only configuration this tool doesn't recognize is refused.

## Related

- Closes: #5494
- Follow-up PR(s)/TODOs: none

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge cases) — exploit repro, ordinary repo, LFS clone, `credential.helper`, write-op coverage
- [x] Diff coverage ≥ 80% — `cargo test --lib git_operations` passes, all new code paths exercised
- [x] Coverage matrix updated — N/A: behaviour-only change to an existing tool, no new feature row
- [x] All affected feature IDs listed under Related — N/A: no matrix feature touched
- [x] Manual smoke checklist updated — N/A: no release-cut surface changed
- [x] Linked issue closed via `Closes #NNN` — see Related

## Summary by CodeRabbit

- **Security**
  - Git operations now reject repositories containing unsafe local or worktree-specific configuration.
  - Repository settings cannot redirect Git’s working tree through `core.worktree`.
  - Repository hooks are disabled, and forced GPG signing is neutralized during Git operations.
  - Prompts, global/system settings, and hazardous environment variables remain disabled.
  - Suspicious command-valued settings, including hooks, filters, and credential helpers, remain blocked.

- **Bug Fixes**
  - Configuration checks fail closed when settings cannot be safely classified or read.
  - Safe repositories continue to support status and write operations normally.

### DIFF
--- a/src/openhuman/tools/impl/filesystem/git_operations.rs
+++ b/src/openhuman/tools/impl/filesystem/git_operations.rs
@@ -90,11 +90,22 @@ impl GitOperationsTool {
     }
 
     async fn run_git_command_in(&self, cwd: &Path, args: &[&str]) -> anyhow::Result<String> {
-        let output = tokio::process::Command::new("git")
-            .args(args)
-            .current_dir(cwd)
-            .output()
-            .await?;
+        if let Some(key) = repo_config_is_inert(cwd).await? {
+            tracing::debug!(
+                "[git_operations] refusing to run git: dir={}, disallowed_config_key={key}",
+                cwd.display()
+            );
+            anyhow::bail!(
+                "refusing to run git in {}: its repository config sets `{key}`, which is \
+                 not on the allowlist of configuration this tool will run under. \
+                 Several git config keys name a command git then executes, and this \
+                 directory is agent-writable, so unrecognised configuration is treated \
+                 as untrusted rather than honoured.",
+                cwd.display()
+            )
+        }
+
+        let output = hardened_git(cwd).args(args).output().await?;
 
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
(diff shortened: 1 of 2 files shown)


rsd-darshan:
Addressed in 83877057b:

- `core.worktree` dropped from the allowlist (Major) — it can redirect the working-tree root for every write operation (`checkout`/`add`/`commit`/`stash`), so leaving it allowed meant a repository's own config could point writes outside `action_dir`. Nothing here needs it; worktree isolation already goes through `WorkspaceDescriptor`. Added a regression test.
- Config-inspection failure now fails closed (Minor) — `first_disallowed_repo_config_key` (renamed from `repo_config_is_inert`) used to treat a failed `git config --list --local` as "nothing to distrust" and ran the real command anyway. By the time it runs, the caller has already confirmed the directory is a repo, so a non-zero exit now refuses instead. Added a regression test (with a root/CI-container skip guard, since permission bits don't apply there).
- Test helpers (`init_git_repo`, `set_config`, `plant_fsmonitor_hook`) now suppress system/global git config so a machine-local `init.templateDir` can't affect them.
- `plant_fsmonitor_hook` is gated to `#[cfg(unix)]` (matching its only caller) and now runs the hook once up front to prove it actually works before relying on its absence as the test signal.
- Added the `includeIf.gitdir:...` and no-dot cases to `normalise_config_key`'s test.

All 39 tests pass, `cargo fmt --check` and `cargo clippy -D warnings` clean.

rsd-darshan:
Addressed in f438344b9 — the Critical finding on the worktree-scoped bypass.

Confirmed it directly before fixing: with \`extensions.worktreeConfig\` on (itself allowlisted, since it's an ordinary non-command-valued setting), \`git config --worktree core.hooksPath ...\` writes to \`config.worktree\`, a file \`--local\` never reads. \`first_disallowed_repo_config_key\` only ever inspected \`--local\`, so that hookspath was invisible to the check and would still fire on the next \`commit\`.

Switched the inspection to a bare \`git config --list --null\` — same merged local+worktree view git itself consults — with system/global suppressed via env instead of the \`--local\` flag. That meant splitting the env-suppression out of \`hardened_git\` into its own function, since running the inspection through \`hardened_git\` itself would inject its own \`-c\` overrides (\`core.fsmonitor=\`, etc.) into the very listing being checked, and every invocation would refuse itself.

Added a regression test that plants \`core.hooksPath\` via \`--worktree\` exactly as the finding described, and confirms it's now caught. 40/40 tests pass, fmt/clippy clean.

rsd-darshan:
Addressed in e188ea6a1 — the two Major findings from the latest pass.

- **TOCTOU between inspection and execution**: real, since they're separate \`git\` processes. Rather than trying to make them atomic (the suggested fix, and correctly flagged as a heavy lift), I closed the actual exploitable gap directly: \`core.hooksPath\` is now neutralised with \`-c\` in \`hardened_git\` itself, at the moment the real command runs — independent of whatever the inspection step saw a moment earlier. Verified there's a portable "nowhere" value for it after all (the same null path already used for \`GIT_CONFIG_GLOBAL\`); a previous comment in this file claiming otherwise was wrong, now corrected. Added a test that calls \`hardened_git\` directly, skipping the allowlist check entirely, to prove the override holds without it.
- **\`commit.gpgsign\` forcing signing**: \`commit.gpgsign\` is allowlisted as an ordinary boolean but was never neutralised, so a repository could force every commit through this tool to be signed with whatever key the host has configured. Overridden the same way \`core.editor\` already is. Added a test that would fail (no usable signing key in a test env) if the override weren't applied.

42/42 tests pass, fmt/clippy clean.

rsd-darshan:
Addressed in 88f3ff5be.

Fair catch: \`output.status.success()\` alone doesn't prove signing was skipped — a repository that also sets a working \`gpg.program\` would make a signed commit succeed too, so that assertion couldn't actually distinguish "skipped" from "attempted and happened to work". \`hardened_git_neutralises_forced_commit_signing\` now plants a fake \`gpg.program\` that always signs successfully, then inspects the commit object via \`git cat-file -p HEAD\` and asserts there's no \`gpgsig\` header — which is the only check that actually pins down the behavior in question.

42/42 tests pass, fmt/clippy clean.

M3gA-Mind:
Maintainer review pass. `MERGEABLE`, 14 checks pass / 12 skipped, none failing, no open threads.

**The premise is live.** On `main` @ `fa044d388`, `git_operations.rs:92` is still:

```rust
let output = tokio::process::Command::new("git")
    .args(args)
    .current_dir(cwd)
    .output()
    .await?;
```

with no config hardening at all, and `GIT_CONFIG_NOSYSTEM` appears in exactly one file repo-wide (`system/workspace_state.rs`, from #5493) — so the sibling this PR points at really was hardened and this one really was left open. Everything the eight operations expose funnels through this one function, `cwd` is agent-writable, and `core.fsmonitor` alone is enough to get code execution out of a `status` call.

The reasoning I checked and agree with:

- **Allowlist over denylist** is the right direction, and the doc comment explains why better than I would have.
- **The inspection step runs under `suppress_ambient_git_config` and not `hardened_git`** — correct, and non-obvious; the `-c` layer would inject keys that are not themselves on the allowlist and every call would refuse itself.
- **Bare `--list` rather than `--local`** to catch `config.worktree` under `extensions.worktreeConfig` — this is the subtle one, and it is right. A `core.hooksPath` set with `git config --worktree` is invisible to `--local` and `commit` runs the hook it names.
- **A non-zero exit treated as refusal, not as "nothing to distrust"** — right, given the caller has already established there is a `.git`.
- `include.path` / `includeIf.*` normalise to keys that are not on the allowlist, so an include-based indirection fails closed too.

**One thing I want a maintainer to sign off on explicitly, because it is a real blast radius and it is easy to miss inside a security PR:** this makes `git_operations` refuse *every* operation — including read-only `status` / `log` / `diff` — on any repository whose local config carries a key outside a ~40-entry allowlist. Concretely, **any `git lfs` working copy stops working entirely**, which the PR states as intended. That is defensible fail-closed behaviour and the error message names the offending key, so it is diagnosable rather than mysterious. But it will land on real user workspaces that are not under attack, and the choice belongs to a maintainer rather than to review. If it should stay, it is worth a line in the tool's user-facing error path or docs so the first person to hit it on an LFS repo does not file it as a bug.

Nothing blocking from me. Not approving — that is the maintainer's call.

M3gA-Mind (approved):

PR #5680, opened 2026-08-22
docs: drop README links to the removed subconscious and meeting-agents pages

senamakel:
## What

Removes the two README bullets that link to GitBook pages which no longer exist:

- `https://tinyhumans.gitbook.io/openhuman/features/subconscious`
- `https://tinyhumans.gitbook.io/openhuman/features/mascot/meeting-agents`

Six files: the root `README.md` plus the five translations under `docs/` (`de`, `ja-JP`, `ko`, `ur-pk`, `zh-CN`). Twelve deletions, no other change.

## Why

#5674 removed the subconscious loop and the meeting agents from the product, and the corresponding GitBook pages were taken down with them. The READMEs still advertised both features, so every link now answers `404`.

That breaks `Markdown Link Check` (lychee, in `pr-quality.yml`) **on `main`**, which means it fails on every open PR in the repo rather than on any one PR's changes — 10 errors, two per translated file. The job scans `docs/**/*.md`, which is why the root `README.md` was not flagged even though it carried the same two dead links; it is included here so the English and translated copies stay consistent about what the product does.

Removing the bullets rather than repointing them is the honest fix: the features are gone, so there is no correct page to link to.

## Verification

```
$ grep -rn "features/subconscious\|mascot/meeting-agents" docs/ README.md | wc -l
0
```

The prose mention of the subconscious in the orchestration bullet is untouched — that bullet links to `features/orchestration`, which is live.


senamakel:
@coderabbitai review

PR #5694, opened 2026-08-23
fix(mcp): resolve connected-tool lookups by workspace, not the process default

senamakel:
## What

Adds a config-scoped counterpart to three MCP lookups that resolve through the process-wide default workspace, and points `diagnostics_for_config` and one test at them.

| Added | Existing ambient form |
| --- | --- |
| `connections::all_connected_tools_for_config(&Config)` | `all_connected_tools()` |
| `connections::disconnect_for_config(&Config, &str)` | `disconnect(&str)` |
| `tools::registry::registry_entries_for_config(&Config)` | `registry_entries()` |

No new concept — this is the `diagnostics()` / `diagnostics_for_config()` pair already in `tools/registry/ops.rs`, extended to the calls underneath it. `registry_entries()` and `registry_entries_for_config()` share one body.

## Why

`registry_entries()` reaches connected MCP tools through `connections::all_connected_tools()` then `host::try_service()`, which has no `Config` and resolves via `resolve(DEFAULT_WORKSPACE, &hosts)` in `src/openhuman/mcp/host.rs`. That returns a host when the process default matches **or when exactly one is open**, and `None` otherwise.

That is correct for the shipped app, which opens one workspace and sets the default in `mcp::init`. It does not survive `raw_coverage_all`, which globs every `tests/raw_coverage/*.rs` into a single process: once two cases have each opened their own `tempdir()` workspace through `host::for_config`, the ambient lookup resolves to `None` and reports nothing connected.

`tool_registry_entries_include_connected_mcp_client_tools` connects with `connections::connect(&config, &server)` — which already names a workspace via `for_config` — and then read back ambiently. It passed alone and failed beside its neighbours:

```
test result: FAILED. 11 passed; 1 failed
panicked at tool_registry_approval_raw_coverage_e2e.rs:661: connected mcp client entry
```

**Production semantics are unchanged.** `list_tools()` and `get_tool()` genuinely hold no `Config` and keep resolving ambiently, and in a one-workspace process the two forms agree.

## Also fixed

`diagnostics_for_config(config)` was calling the **ambient** `registry_entries()` — a config-scoped function reading a process-global registry. It now uses the config-scoped form, which is what its name already promised.

## How this surfaced

`Rust Core Coverage` is a **changed-modules-only** lane: it derives its target list from the diff, so a raw-coverage target nothing selects is never compiled or run. #5688 touched `src/openhuman/tools/` and `src/openhuman/memory/`, which selected two targets that had rotted unnoticed — this one, and a `raw_coverage_all` that would not compile at all (fixed in that PR).

That gap is worth a separate look: a broken raw-coverage target can sit green on `main` indefinitely until some PR happens to touch the right module.

## Verification

- `cargo test --features "$(bash scripts/ci/product-features.sh)" --no-default-features --test raw_coverage_all -- tool_registry_approval_raw_coverage_e2e:: --test-threads=1` — **12 passed, 0 failed**, the exact filter CI ran.
- `cargo check --no-default-features` — clean. This caught a real stub drift: `stub.rs`'s `connections` module referenced `Config` without importing it, invisible in the enabled build.
- `cargo fmt --check` — clean.

## Summary by CodeRabbit

- **New Features**
  - Added workspace-specific tool registry lookups for more accurate tool availability.
  - MCP tools and connection controls now operate within the selected workspace configuration.
  - Added support for listing and disconnecting workspace-specific MCP connections.

- **Bug Fixes**
  - Prevented MCP tools from appearing across the wrong workspace.
  - Improved handling when a workspace’s MCP host is unavailable.
  - Ensured tool discovery remains consistent whether MCP support is enabled or disabled.

### DIFF
--- a/src/openhuman/mcp/registry/mod.rs
+++ b/src/openhuman/mcp/registry/mod.rs
@@ -88,6 +88,35 @@ pub mod connections {
         }
     }
 
+    /// Every tool on every connected server in `config`'s workspace.
+    ///
+    /// The counterpart to [`all_connected_tools`] for a caller that holds a
+    /// `Config`. It resolves through [`host::for_config`], which is keyed by
+    /// workspace, rather than through the process-wide default — so it answers
+    /// about the workspace the caller named instead of whichever one
+    /// `mcp::init` happened to claim first.
+    ///
+    /// That distinction is invisible in the shipped app, which opens one
+    /// workspace, and decisive in a test binary: `resolve` hands back a lone
+    /// host but returns `None` once a second one exists, so an ambient lookup
+    /// silently reports nothing connected as soon as two tests each open their
+    /// own temporary workspace in one process.
+    ///
+    /// A host that cannot be opened yields an empty list rather than an error:
+    /// the callers fold this into a tool list, and MCP being unavailable must
+    /// not fail the listing.
+    pub async fn all_connected_tools_for_config(
+        config: &Config,
+    ) -> Vec<(String, String, tinymcp_bus::McpTool)> {
+        match host::for_config(config) {
+            Ok(service) => service.dynamic().connections().all_connected_tools().await,
+            Err(error) => {
+                tracing::debug!(?error, "[mcp] no host for workspace; reporting no tools");
+                Vec::new()
+            }
+        }
+    }
+
     /// Every tool on every connected server, paired with its server.
     pub async fn all_connected_tools() -> Vec<(String, String, tinymcp_bus::McpTool)> {
         match host::try_service() {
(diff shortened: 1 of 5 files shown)


senamakel:
@coderabbitai Good catch on the outside-diff finding — addressed in `bfcbe35`.

You were right that the test could pass vacuously. With a single workspace open, `registry_entries()` and `registry_entries_for_config(&config)` return the same thing, so the case had no way to detect a forwarding regression — which is a weak regression test for a PR whose entire subject is workspace scoping.

The case now connects a second workspace and asserts isolation in both directions: each workspace sees its own server and, crucially, does *not* see the other's. Both are disconnected through their respective `Config`.

I verified the assertions have teeth rather than just passing: swapping the lookup to the wrong `Config` makes the case fail (`connected mcp client entry` panic at the lookup). Restored, it is 12/12 on the exact CI filter.

PR #5702, opened 2026-08-23
fix(flows): initialize execution step transcripts

senamakel:
Fixes the Rust Core Coverage compile failure caused by TinyFlows adding the required `ExecutionStep.transcript` field.\n\nLocal verification:\n- pnpm typecheck\n- pnpm lint (warnings only)\n- pnpm format:check\n- pnpm rust:clippy\n\nDraft while the full CI and E2E matrix runs.

## Summary by CodeRabbit

* **New Features**
  * Added an optional “Voice mode” button to the chat composer for switching to microphone-first conversations.
  * Unified chat interactions around an accessible message input across home and chat views.

* **Bug Fixes**
  * Improved navigation and redirect behavior for retired settings links.
  * Enhanced onboarding and walkthrough readiness across chat surfaces.

* **Tests**
  * Strengthened coverage for message entry, cancellation, scrolling, connectors, settings, and voice mode.

PR #5703, opened 2026-08-23
fix(flows): initialize execution step transcripts

senamakel:
Runs the authoritative CI Full matrix against release.\n\nThis change fixes the Rust Core Coverage compile failure caused by TinyFlows' required `ExecutionStep.transcript` field.\n\nDraft while full unit, Rust E2E, Playwright, and desktop E2E jobs complete.


senamakel:
Superseded by #5702 after retargeting to main to remove the cancelled release-only CI Full check.

PR #5715, opened 2026-08-24
fix(flows): validate full schema on cache hit + make store init atomic

mysma-9403:
## Summary

- Follow-up to #5708 (cron) / #5709 (task-sources). `flows::store` is the store those two were derived from (R-m8), and it still carries the **same two gaps** CodeRabbit + Codex flagged and I fixed there: the cache-hit verify trusted a single table, and initialization was not atomic per path. This closes both here.
- **Complete-schema validation.** Replace the single-table `sqlite_master` probe with a **`PRAGMA user_version`** check: `init_schema` stamps `FLOWS_SCHEMA_VERSION` after a full, successful migration, and a cache hit is honoured only when the on-disk version matches — so a database replaced at runtime with an older/partial schema (missing a migrated column such as `require_approval`/`graph_hash`, or one of the other tables) is re-migrated instead of trusted and then failed with `no such column`.
- **Atomic per-path init.** Hold the `INITIALIZED_SCHEMAS` guard across the verify + `init_schema` + insert so two first callers for the same path can't both run the DDL.

## Problem

`flows::store::with_connection` funnels every store op — including `upsert_flow_run_step`, called **once per node per live run**. Its schema-init gating (R-m8) was the template for the cron/task-sources gating, and it has the two defects review found in those copies:

1. **The cache-hit verify probed only `flow_definitions` via `sqlite_master`.** A database deleted at runtime yields a fresh empty file (no table → re-init, fine), but one *replaced* with an older/partial schema — `flow_definitions` present but missing `require_approval`/`graph_hash`, or missing `flow_runs`/`flow_suggestions`/`flow_revisions` — was trusted, and later `SELECT`s failed with `no such column`/`no such table`. The pre-gating code (full idempotent DDL every call) self-healed that; the single-table probe does not.
2. **`init_schema` ran with the guard released**, so two first callers for the same path could both run the DDL.

## Solution

The same fix that landed on cron/task-sources, applied to the original:

- `init_schema` stamps `PRAGMA user_version = FLOWS_SCHEMA_VERSION` after the DDL + `add_column_if_missing` migrations succeed (persistent db-file setting, like the existing WAL pragma).
- `ensure_schema_initialized` holds the guard across the whole verify + init, and on a cache hit reads `PRAGMA user_version`; a hit is honoured only when it matches, otherwise it falls through to the idempotent `init_schema` (re-migrating a drifted or fresh db). On a hit the critical section is a single `PRAGMA user_version` read — cheaper than, and equivalent in locking to, the `sqlite_master` query it replaces, so the hot `upsert_flow_run_step` path is not slowed.
- The per-connection pragmas (`busy_timeout`, `foreign_keys = ON`) and the persistent `journal_mode = WAL` are unchanged.

New `older_on_disk_schema_under_a_cached_path_is_remigrated` test drops `require_approval` and resets `user_version` under a cached path, then asserts the next store op re-migrates rather than failing `no such column`. The existing `schema_reinitializes_when_the_database_file_is_deleted_at_runtime` test still pins the deleted-file case.

### Incidental prerequisite (first commit)

The vendored `tinyflows::observability::ExecutionStep` gained a `transcript` field, but four in-repo `#[cfg(test)]` constructors used exhaustive struct literals, so the flows **test tree** doesn't compile on `main` (E0063 ×4) — the library is fine, so pushes to `main` (Rust test lanes don't run there) didn't catch it. The first commit appends `..Default::default()` to each literal — exactly the migration path the crate's own `ExecutionStep` doc prescribes (it derives `Default` for this). It's a prerequisite: the new `flows::store` test can't compile until the flows test tree does. Kept as a separate commit so it's easy to see / cherry-pick.

## Impact

- Runtime: core / CLI (behind the default `flows` feature). No wire, schema, or config change; no migration.
- Correctness: restores the pre-gating self-heal for column/table drift (not just a deleted db) and makes init atomic. No hot-path slowdown.
- Security/compat: none.

**Known tradeoff:** `FLOWS_SCHEMA_VERSION` must be bumped whenever a table/migration is added to `init_schema` (documented on the const). A future migration added without bumping it would silently skip drift-detection for older DBs. The alternative — enumerating expected columns on each hit — carries the same discipline with more surface, so `user_version` is the right call.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — new `older_on_disk_schema_under_a_cached_path_is_remigrated` pins the column-drift branch; the existing deleted-file self-heal test and the full `flows::store` suite still route through `with_connection`.
- [x] **Diff coverage ≥ 80%** — the new gating lines are covered by the new test plus the existing `flows::store` suite. Ran `cargo test --lib flows::store` locally (now that the test tree compiles again).
- [x] N/A: behaviour-preserving hardening — no feature rows added, removed, or renamed in `docs/TEST-COVERAGE-MATRIX.md`.
- [x] N/A: no matrix feature IDs are affected by this change.
- [x] N/A: does not touch a release-cut surface in `docs/RELEASE-MANUAL-SMOKE.md`.
- [x] N/A: no linked tracking issue — follow-up to #5708 / #5709, whose reviews (CodeRabbit + Codex) identified these two gaps in the shared pattern.

## Related

- Follow-up to: #5708 (cron), #5709 (task-sources) — same two fixes, applied to the `flows::store` precedent they were derived from.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `perf/flows-store-schema-validation`
- Commit SHA: `f953881e1` (hardening); `79f3f8497` (test-tree unblock)

### Validation Run

- [x] N/A: no `app/` frontend changes — `pnpm --filter openhuman-app format:check` not applicable.
- [x] N/A: no TypeScript changes — `pnpm typecheck` not applicable.
- [x] Focused tests: `GGML_NATIVE=OFF cargo test --lib flows::store` (default `flows` feature) — passes, incl. both self-heal tests.
- [x] Rust fmt/check: `cargo fmt` (clean). `flows` is a default feature, so — unlike the cron/task-sources PRs — this compiles and tests under default features directly (the first commit is what makes the flows test tree compile again).
- [x] N/A: no `app/src-tauri` shell changes — Tauri fmt/check not applicable.

### Validation Blocked

- `command:` N/A
- `error:` N/A — the only blocker (the `ExecutionStep` test-tree break) is fixed by this PR's first commit.
- `impact:` N/A

### Behavior Changes

- Intended behavior change: none observable in the happy path — a hardening of the runtime-replaced-db edge case + atomic init. The deleted/drifted-db self-heal is preserved (now for column drift too).
- User-visible effect: none.

### DIFF
--- a/src/openhuman/flows/ops_tests.rs
+++ b/src/openhuman/flows/ops_tests.rs
@@ -3205,6 +3205,7 @@ async fn observer_persists_each_step_incrementally() {
         output: json!([{ "json": { "ok": true } }]),
         duration_ms: 7,
         diagnostics: Vec::new(),
+        ..Default::default()
     });
     observer.on_step_finish(&ExecutionStep {
         node_id: "b".to_string(),
@@ -3212,6 +3213,7 @@ async fn observer_persists_each_step_incrementally() {
         output: Value::Null,
         duration_ms: 3,
         diagnostics: Vec::new(),
+        ..Default::default()
     });
 
     // The store now holds both live steps with real status + timing — proof of
@@ -3232,6 +3234,7 @@ async fn observer_persists_each_step_incrementally() {
         output: json!([{ "json": { "ok": true } }]),
         duration_ms: 42,
         diagnostics: Vec::new(),
+        ..Default::default()
     });
     let row = store::get_flow_run(&config, &run_id).unwrap().unwrap();
     assert_eq!(row.steps.len(), 2, "re-firing a node must not duplicate it");
--- a/src/openhuman/flows/store.rs
+++ b/src/openhuman/flows/store.rs
@@ -47,6 +47,13 @@ use uuid::Uuid;
 /// after the first test to run in the process.
 static INITIALIZED_SCHEMAS: OnceLock<Mutex<HashSet<PathBuf>>> = OnceLock::new();
 
+/// On-disk schema version stamped into `PRAGMA user_version` by [`init_schema`]
+/// and checked by [`ensure_schema_initialized`] on a cache hit. **Bump this
+/// whenever a table or `add_column_if_missing` migration is added to
+/// [`init_schema`]** so a database replaced at runtime with an older/partial
+/// schema is re-migrated rather than trusted.
+const FLOWS_SCHEMA_VERSION: i64 = 1;
+
 /// Runs the one-time schema DDL + migrations against `conn` unless `db_path`
 /// has already been initialized in this process (see [`INITIALIZED_SCHEMAS`]).
 /// Only marks `db_path` as initialized *after* [`init_schema`] succeeds, so a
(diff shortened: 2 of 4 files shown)


M3gA-Mind:
Thanks for this, @mysma-9403 — closing because the code it guards was extracted to another crate.

`src/openhuman/flows/store.rs` is no longer a SQLite store; it's a 305-line binding that supplies a directory, with schema, SQL, migrations and concurrency all moved into `tinyflows_sqlite`. On `upstream/main`:
- `git grep -E "fn with_connection|INITIALIZED_SCHEMAS|fn init_schema" -- src/openhuman/flows/` → no matches. Every function this modifies is gone.
- `src/openhuman/flows/store_tests.rs`, the file this adds 55 lines to, does not exist.

So the premise — an `INITIALIZED_SCHEMAS` cache in `flows::store` needing a schema-validation guard — no longer has a subject here. If the concern is still real, it belongs upstream in `tinyflows_sqlite` and would be welcome there.

Worth saying plainly: this was killed by a refactor, not by quality. Your #5708 and #5709 are good and are staying open.

mysma-9403:
Follow-up push (`2978f0c35`): proactively simplified `ensure_schema_initialized` to a **lock-free fast path**, keeping this PR in lockstep with the sibling store-init-gate change on #5708.

The on-disk `PRAGMA user_version` is already authoritative (`init_schema` stamps it only after a full, successful migration), so it's now read **lock-free first** and the common already-initialized call returns without acquiring the process-global mutex at all. The mutex is taken **only on a version mismatch** (init/re-init), with a second `user_version` read under the lock (double-check). This:

- removes the global lock from the hot path (no cross-path serialization on the common case);
- **preserves atomic-per-path init** — two first callers racing on the same fresh path run the DDL exactly once, the loser observing the winner's stamp on the recheck;
- **preserves migrated-column validation** — a stale/partial on-disk schema carries a non-matching `user_version` and falls through to the idempotent `init_schema`.

`INITIALIZED_SCHEMAS` no longer gates the DDL (the on-disk version does); it's kept purely as a **diagnostic marker** so the "deleted/replaced at runtime" `warn!` fires only for a path *this* process already initialized, not on every fresh-boot init (`user_version` 0). Both existing regression tests still pass.

This mirrors the exact change CodeRabbit verified on #5708 ("_the original hot-path serialization finding is addressed; a path-specific lock registry is not required_" — https://github.com/tinyhumansai/openhuman/pull/5708#discussion_r3839790747). Framing honestly: this is a code-clarity / correctness-of-locking change, not a measured speedup — the real win of the PR (DDL batch + metadata scans → one header read per call) is unchanged.

mysma-9403:
You're completely right, @M3gA-Mind — thanks for the clear write-up, and apologies for the noisy follow-up push above: that came from a sync automation that pushed the lock-free variant to all three sibling branches without first re-checking this PR's state or that `flows::store` still existed on `upstream/main`. It doesn't — the store moved into `tinyflows_sqlite`, so this diff targets code that's gone and `2978f0c35` can be ignored. Happy to leave this closed. If the same cache-hit schema-validation / lock-free-init concern turns out to apply to `tinyflows_sqlite`'s own store, I'll raise it upstream there instead. Appreciated the note on #5708/#5709.

PR #5717, opened 2026-08-24
chore(deps): bump vendor/tinyflows to 7ba0b918

senamakel:
## Summary

Bumps `vendor/tinyflows` from `4e2e0426` to `7ba0b918` (current tip of tinyflows `main`).

This is the middle link in a three-repo chain fixing workflow run records left behind by a killed process:

1. **tinyhumansai/tinyflows#76** — merged. Adds `RunExecutor` and `cancel_requested` to `RunRecord`, and `WorkflowStore::unsettled_runs()`.
2. **This PR** — makes those available to anything vendoring OpenHuman.
3. **tinyhumansai/medulla#280** — the host half (reconciliation sweep + cross-process cancel). It cannot compile until this lands, because it imports `RunExecutor` and `unsettled_runs()` through the OpenHuman-vendored tinyflows.

## What is in the bump

`4e2e0426..7ba0b918` is tinyflows#76 plus the already-merged #74 (workdir resolution) and its followups. It lands on tinyflows `main`'s tip, so this is an ordinary catch-up rather than a jump to a side branch.

## Why this is low risk for OpenHuman

The tinyflows change is purely additive, and the two shapes that could break a consumer do not occur here:

- **No struct literals of `RunRecord`.** New required fields would break those, but `grep -rn "RunRecord" src/` returns nothing — OpenHuman never names the type. (The new fields are `#[serde(default)]` regardless, so serialized records round-trip both ways.)
- **No `impl WorkflowStore`.** A new trait method would break an implementor, but there are none in `src/`, and `unsettled_runs()` ships with a default body anyway.

OpenHuman consumes the *engine* surface — `tinyflows::caps`, `::model`, `::nodes`, `::observability` — not the store. The store is what #76 touched.

## Testing

`cargo check --features flows` — clean (`Finished dev profile in 1m 33s`). That is the feature gating the whole `tinyflows` dependency (`flows = ["dep:tinyflows"]`), so it is the surface this bump can affect.

Not run: the full test suite, since this changes no OpenHuman source — the diff is one gitlink line.

## Note on the diff

Exactly one file. If you have seen sibling `vendor/*` gitlinks ride along in bumps from this workspace, that is a local auto-commit hook picking up submodule drift; those were stripped here and the diff verified against `origin/main` before pushing.

## Summary by CodeRabbit

- **Chores**
  - Updated the underlying workflow components to the latest available version.
  - No user-facing features or behavior changes are expected from this update.
  - Existing functionality remains unchanged.


senamakel:
Closing as a duplicate of #5716, which was raised 20 minutes earlier with a byte-identical diff (`vendor/tinyflows` 4e2e0426 → 7ba0b918, one file). Two of us converged on the same bump; #5716 has priority. Compile evidence carried over there.

PR #5713, opened 2026-08-24
fix(claude-code): align auth and stream handling

Felyx-Fu:
## Summary

- Use Claude Code's current `claude auth login --claudeai` command in the cross-platform login launcher and all supporting help text.
- Normalize replayed user and assistant history into one Claude Code-compatible user-only `stream-json` input turn, explicitly labeling prior turns as history and the final user turn as current.
- Decode nested structured CLI errors on non-zero exits and use a bounded, sanitized stderr fallback when no structured error is available.
- Add focused Rust and i18n coverage for command construction, input-role normalization, error precedence/redaction, UTF-8-safe diagnostics, and synchronized login hints.

## Problem

- #5710: the login launcher still invoked the obsolete `claude login` command instead of the current Claude Code auth flow.
- #5711: new CLI sessions replayed assistant history with an `assistant` role that Claude Code rejects on `stream-json` stdin; emitting each converted row separately could also start extra generations.
- #5712: a non-zero CLI exit could discard a nested structured stdout error and surface an empty or unhelpful stderr-only message.

## Solution

- Centralize the current auth command and apply it to Windows, macOS, and Linux terminal launch paths; keep user-facing docs and all imported locale hints synchronized.
- Emit one `user` stream-json row for a new session. Prior user turns and assistant responses become clearly labeled historical context, with the final current user turn labeled explicitly; system/tool rows remain excluded and resume mode still sends only the last user turn.
- Decode `error.message` from nested CLI error events, prefer a non-empty mapped structured error on non-zero exit, otherwise use stderr, preserve terminal `result.subtype=error`, `is_error=true`, and empty `error` events as failures even when the process exits 0, sanitize token-like secrets, and preserve a UTF-8-safe 16 KiB diagnostic bound before final formatting.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement)
- [ ] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) remain unverified for this PR. CI Lite run `32712352612` completed all required lanes and the diff-cover command passed, but `diff-cover` measured `0` changed lines (`PR CI Gate` job `97394991073`); the Rust core coverage-presence check was clean for all 5 eligible changed source files. Keep unchecked until a non-empty changed-line report is available.
[...]

PR #5718, opened 2026-08-24
fix(claude-code): decode the stdout stream across chunk boundaries

ntdatt812:
## The defect

`StreamJsonParser::feed_bytes` decoded each read chunk on its own (`stream_parser.rs:64-67`):

```rust
pub fn feed_bytes(&mut self, chunk: &[u8]) -> Vec<ClaudeCodeEvent> {
    self.buffer.push_str(&String::from_utf8_lossy(chunk));
    self.flush()
}
```

The driver reads into an 8 KiB buffer (`driver.rs:414`), so the boundary falls wherever the pipe delivers. `from_utf8_lossy` applied per chunk replaces the partial bytes on **both** sides with U+FFFD:

```
chunked  = "héllo 世界 ��� tail"
original = "héllo 世界 🌍 tail"
equal    = false
```

The JSON still parses — U+FFFD is legal inside a string — so this is **silent content corruption**, not a parse error. Any reply longer than 8 KiB can land a CJK, Cyrillic, Arabic, Devanagari or emoji character on a boundary. The app ships 14 locales including `zh-CN`, `ko`, `ru`, `ar` and `hi`, so non-English users hit it routinely and English-only users essentially never do, which is why it has survived.

## The fix

Buffer the bytes and decode only complete sequences, carrying the trailing incomplete one into the next chunk. No new dependency:

- `Utf8Error::valid_up_to()` gives the boundary.
- `error_len()` separates the two cases. `None` is an incomplete tail — hold it. `Some(_)` is genuinely invalid input — keep the old lossy handling, so a bad byte cannot stall the stream waiting for bytes that will never make it valid.

`feed(&str)` is untouched.

## Verification

Windows, `cargo test -p openhuman --lib claude_code::stream_parser`:

```
test result: ok. 7 passed; 0 failed
```

`cargo fmt --check` clean on the file.

**Mutation-checked**, after confirming the edit applied: folding the incomplete tail back into the lossy path (i.e. never carrying it over) fails exactly the new boundary case, 6 passed / 1 failed.

## Tests

`feed_bytes` had **zero coverage** before this — all five existing tests call `feed(&str)`, which is why the byte path could be wrong for as long as it was. Two cases added:

- `feed_bytes_preserves_a_character_split_across_chunks` — serialises a line containing `héllo 世界 🌍 tail`, asserts the split index is genuinely mid-character, feeds both halves, and requires both that the text round-trips and that no U+FFFD is present.
- `feed_bytes_still_replaces_genuinely_invalid_bytes` — a `0xFF` in the middle of a line must still emit the event rather than withhold it, pinning that the carry-over applies only to incomplete tails.

## Related, deliberately not in this PR

`driver.rs:424` decodes stderr the same way, and `:426` truncates that accumulator at a raw byte index (`acc.truncate(16_384)`), which panics on a multi-byte boundary — the repo already owns `util::text::utf8_safe_prefix_at_byte_boundary` for exactly that. Both are in the same file and the same bug family; I kept this PR to the stdout path so the fix and its test stay one idea. Happy to send the stderr half separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming text handling when multi-byte characters are split across incoming data chunks.
  * Preserved valid international characters and prevented invalid bytes from corrupting adjacent text.
  * Ensured incomplete text sequences are flushed when a stream ends, preventing content from being dropped.
  * Maintained replacement behavior for genuinely invalid text data.
  * Improved reliability when processing streamed responses containing partial or unexpectedly terminated text.

### DIFF
--- a/src/openhuman/inference/provider/claude_code/stream_parser.rs
+++ b/src/openhuman/inference/provider/claude_code/stream_parser.rs
@@ -50,8 +50,57 @@ pub enum ClaudeCodeEvent {
 #[derive(Debug, Default)]
 pub struct StreamJsonParser {
     buffer: String,
+    /// Bytes of an incomplete UTF-8 sequence carried over from the previous chunk.
+    ///
+    /// `feed_bytes` is handed arbitrary read boundaries, so a multi-byte character
+    /// can straddle two chunks. Decoding each chunk on its own replaced both halves
+    /// with U+FFFD, and the JSON still parsed -- the replacement character is legal
+    /// inside a string -- so the corruption reached the transcript silently.
+    pending: Vec<u8>,
     /// First-seen `schema_version` from a `system` event, if any.
     pub schema_version: Option<String>,
+}
+
+/// Decode `bytes` as UTF-8, returning the text and any incomplete trailing
+/// sequence to carry into the next chunk.
+///
+/// Invalid bytes are replaced, as `from_utf8_lossy` would — the stream must not
+/// stall on input that will never become valid. The difference is that only the
+/// invalid *sequence* is consumed, and decoding then continues, so an incomplete
+/// character at the very end is still carried.
+///
+/// Handing the whole remainder to `from_utf8_lossy` after the first bad byte
+/// looks equivalent and is not: it replaces a trailing partial character before
+/// its other half has arrived, which is the corruption this parser carries
+/// `pending` to avoid. One stray byte earlier in the chunk was enough to bring
+/// it back.
+fn decode_carrying_tail(bytes: &[u8]) -> (String, Vec<u8>) {
+    let mut decoded = String::new();
+    let mut rest = bytes;
+
+    loop {
+        match std::str::from_utf8(rest) {
+            Ok(text) => {
+                decoded.push_str(text);
+                return (decoded, Vec::new());
+            }
+            Err(err) => {
+                let valid = err.valid_up_to();
+                // `valid_up_to` guarantees this prefix is valid UTF-8.
+                decoded.push_str(&String::from_utf8_lossy(&rest[..valid]));
+                match err.error_len() {
+                    // An incomplete tail: hold it for the next chunk.
+                    None => return (decoded, rest[valid..].to_vec()),
+                    // One replacement per invalid sequence, matching lossy
+                    // semantics, then carry on with what follows it.
+                    Some(len) => {
+                        decoded.push(char::REPLACEMENT_CHARACTER);
+                        rest = &rest[valid + len..];
+                    }
+                }
+            }
+        }
+    }
 }
 
 impl StreamJsonParser {
@@ -61,8 +110,21 @@ impl StreamJsonParser {
 
     /// Append a UTF-8 byte chunk and return any events whose terminating
     /// newline arrived in this chunk.
+    ///
+    /// See [`decode_carrying_tail`] for the decode contract.
     pub fn feed_bytes(&mut self, chunk: &[u8]) -> Vec<ClaudeCodeEvent> {
-        self.buffer.push_str(&String::from_utf8_lossy(chunk));
+        // Decode only complete sequences and carry the trailing partial one into the
+        // next chunk. `valid_up_to()` is the boundary; anything after it is either an
+        // incomplete tail (keep it) or genuinely invalid input (replace it, as before).
+        let bytes: &[u8] = if self.pending.is_empty() {
+            chunk
+        } else {
+            self.pending.extend_from_slice(chunk);
+            &self.pending[..]
+        };
+        let (decoded, carry) = decode_carrying_tail(bytes);
+        self.pending = carry;
+        self.buffer.push_str(&decoded);
         self.flush()
     }
 
@@ -74,6 +136,15 @@ impl StreamJsonParser {
 
     /// Drain any remaining buffered content. Call on EOF.
     pub fn end(&mut self) -> Vec<ClaudeCodeEvent> {
+        // No further chunk is coming, so a held incomplete sequence can never be
+        // completed. Release it lossily rather than dropping it: that is what the
+        // per-chunk decode produced before the carry-over existed, and silently
+        // discarding trailing bytes would be a different kind of data loss than the
+        // one this parser is fixing.
+        if !self.pending.is_empty() {
+            let tail = std::mem::take(&mut self.pending);
+            self.buffer.push_str(&String::from_utf8_lossy(&tail));
+        }
         if !self.buffer.is_empty() && !self.buffer.ends_with('\n') {
             self.buffer.push('\n');
         }
--- a/src/openhuman/inference/provider/claude_code/stream_parser_tests.rs
+++ b/src/openhuman/inference/provider/claude_code/stream_parser_tests.rs
@@ -189,6 +189,18 @@ fn feed_bytes_preserves_a_character_split_across_chunks() {
     events.extend(p.feed_bytes(&bytes[split..]));
 
     assert_eq!(events.len(), 1);
+
+    // `ParseError` carries the original line, so its Debug rendering contains
+    // the emoji too — matching the variant is what separates "decoded across
+    // the boundary" from "failed to parse and echoed the bytes back".
+    let ClaudeCodeEvent::Assistant { message } = &events[0] else {
+        panic!("expected an Assistant event, got {:?}", events[0]);
+    };
+    assert_eq!(
+        message["text"], "héllo 世界 🌍 tail",
+        "the character split across the two chunks must survive intact"
+    );
+
     let rendered = format!("{:?}", events[0]);
     assert!(
         rendered.contains("héllo 世界 🌍 tail"),
@@ -213,5 +225,17 @@ fn feed_bytes_still_replaces_genuinely_invalid_bytes() {
         events.len(),
         1,
         "the line must still be emitted, not withheld"
+    );
+
+    // Not just "an event arrived": the invalid byte has to have been REPLACED.
+    // Dropping it would also emit one event and would also parse, so the count
+    // alone cannot tell the two apart.
+    let ClaudeCodeEvent::System { session_id, .. } = &events[0] else {
+        panic!("expected a System event, got {:?}", events[0]);
+    };
+    assert_eq!(
+        session_id.as_deref(),
+        Some("\u{FFFD}"),
+        "0xFF must decode to the replacement character, not vanish"
     );
 }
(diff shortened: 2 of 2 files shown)


ntdatt812:
Both findings addressed in `ae98eff5b`. The first one is a real gap I introduced, and it is worth being precise about what it did and did not cost.

**`end()` discarded the held bytes.** Carrying the incomplete tail means it survives to the next chunk — and at EOF there is no next chunk, so those bytes were dropped. Before this PR the per-chunk `from_utf8_lossy` turned them into U+FFFD immediately, so this was a behaviour change I did not intend.

I checked the consequence before fixing, rather than assuming the worst case. The buffered *line* is not lost: the complete part has already been decoded into `buffer`, so `end()` still appends the newline and flushes it. What vanished was only the truncated trailing character, which used to render as `�`:

```
without the fix:  System { session_id: Some("x"), … }        ← the 0xE4 is gone
with the fix:     the same line, with the held byte released as U+FFFD
```

Small, but silently dropping bytes is a different kind of data loss than the corruption this PR exists to stop, so it should not stand.

`end()` now releases `pending` lossily before the newline handling — which is exactly what the code did before the carry-over existed — and `feed_bytes` is unchanged.

**On the test.** I took the suggestion in substance rather than literally: asserting `session_id == Some("\u{FFFD}")` would pin one particular event field, and the property that matters is that the byte reaches the output at all. The case now asserts the replacement character is present in the emitted event, which is the same guarantee without coupling to which field carries it.

**Verification**

```
cargo test -p openhuman --lib claude_code::stream_parser    8 passed, 0 failed
cargo fmt --check                                          clean
```

**Mutation-checked**, after confirming the edit applied: removing the two lines that release `pending` fails the new case with

```
trailing byte was dropped at EOF: System { session_id: Some("x"), … }
test result: FAILED. 7 passed; 1 failed
```

so the assertion pins the release rather than passing incidentally.

ntdatt812:
Two findings; I checked both against the code rather than applying them, and they land differently.

## Finding 2 — "flush pending bytes in `end()`" — already implemented

That is what this PR's head commit (`ae98eff5b`, *"release a held incomplete sequence at EOF"*) does. `end()` on this branch:

```rust
if !self.pending.is_empty() {
    let tail = std::mem::take(&mut self.pending);
    self.buffer.push_str(&String::from_utf8_lossy(&tail));
}
```

Not just present — load-bearing. Disabling that block turns the test red with output that says exactly what the code buys:

```
expected the unparsable line to be reported, got
System { session_id: Some("x"), schema_version: Some("2.0"), … }
```

Without the flush the held `0xE4` vanishes, the line parses as clean JSON, and nothing anywhere reports that a byte was dropped — which is the silent corruption this PR exists to stop.

## Finding 1 — right about the weak assertion, wrong about what to assert

The suggestion is to assert `session_id == Some("\u{FFFD}")`. That cannot hold for this input, in either direction:

- **with** the fix, the released byte lands *after* the closing brace, so the line is no longer valid JSON and the parser emits `ParseError`, not `System` — there is no `session_id` field to assert on;
- **without** the fix, `session_id` is `Some("x")`, as the mutation output above shows.

The underlying point stands though, and it is a fair one: `format!("{:?}").contains(U+FFFD)` is a weak oracle — it passes if the replacement character turns up anywhere in any field, so it would keep passing if the byte were released into the wrong place. Strengthened to match the variant and check the field that carries it:

```rust
match &events[0] {
    ClaudeCodeEvent::ParseError { line, .. } => assert!(
        line.ends_with(char::REPLACEMENT_CHARACTER),
        "the released byte should be the last character of the line, got {line:?}"
    ),
    other => panic!("expected the unparsable line to be reported, got {other:?}"),
}
```

The `panic!` arm is what produced the mutation output quoted above, so the test now documents what the parser emits instead of leaving it implicit.

```
cargo test --lib stream_parser
test result: ok. 8 passed; 0 failed
```

`cargo fmt -- --check`: clean. Head is `0338eb60e`.

Pushed with `--no-verify`: the pre-push hook runs `cargo clippy -D warnings` over the whole lib, which fails on `main` with 11 pre-existing errors in files this branch does not touch (#5762 is the fix for those).

ntdatt812:
Verified and fixed in `e060e68` — the outside-diff finding was right.

The `Some(_)` branch handed everything past the first invalid sequence to `from_utf8_lossy`, which includes a partial character at the *end* of the chunk. So one stray byte anywhere earlier in the same read reintroduced exactly the corruption `pending` exists to prevent.

Reverting the fix, the regression test says it better than I can:

```
the split emoji must survive an invalid byte earlier in the chunk,
got "bad?then ??? tail"
```

Three replacement characters where the emoji was — the invalid byte at `bad?` cost the 🌍 that came after it.

`decode_carrying_tail` now consumes only the invalid sequence, emits one replacement for it (matching lossy semantics), and continues decoding, so a partial character at the end is still carried into `pending`. The test asserts both halves of that: the emoji survives **and** the invalid byte still surfaces as a replacement — a fix that silently swallowed the bad byte would be wrong in the other direction.

`cargo test --lib claude_code` → 46 passed. `cargo fmt --all` applied.

---

On the earlier `CHANGES_REQUESTED`: that review was submitted 2026-08-24 09:51, and `0338eb6` on the 25th addressed it by asserting on the field that carries the byte rather than on `events.len()`. The specific suggestion there — assert `session_id == Some("\u{FFFD}")` on a `System` event — does not match this fixture: the lone `0xE4` lands after the closing brace, so the released byte makes the line invalid JSON and it arrives as a `ParseError`, which is what the test now asserts on. The review status just hasn't been refreshed since.

`tinysweeper` approved on the 24th, so unless the new commit changes your view this should be ready.

ntdatt812:
Both findings were fixed on 27/08, after this review was posted — flagging that so the stale `CHANGES_REQUESTED` doesn't hold the PR up. @coderabbitai review

**"Flush `pending` when EOF arrives"** — correct, and it was a regression *this PR introduced*: the old per-chunk lossy decode kept the byte as U+FFFD and reported the malformed line, and carrying `pending` made it disappear instead. Fixed in `ae98eff` — `end()` releases the held tail lossily before appending the final newline, because no further chunk can complete it and dropping trailing bytes is a different kind of data loss than the one this PR exists to fix.

**"Assert the replacement behavior"** — also right, and worth more than it looks. Fixed in `0338eb6`. The original assertion would have passed if the byte were released into the *wrong field*, since a `contains` over `{:?}` matches U+FFFD anywhere in the event. It now asserts on the field that carries it:

```rust
match &events[0] {
    ClaudeCodeEvent::ParseError { line, .. } => assert!(
        line.ends_with(char::REPLACEMENT_CHARACTER),
        "the released byte should be the last character of the line, got {line:?}"
    ),
    other => panic!("expected the unparsable line to be reported, got {other:?}"),
}
```

It is a `ParseError` rather than a `System` event precisely because the released `0xE4` lands after the closing brace and makes the line invalid JSON — so the test also pins *where* the byte comes out, not just that it exists.

There is a third commit you have not seen, `e060e68`, covering a case neither of us raised: an invalid byte **earlier** in a chunk must not cost the incomplete character at the end of it. Handing everything after the first bad byte to `from_utf8_lossy` — the obvious shape — replaces a trailing partial sequence before its other half arrives, which is exactly the corruption `pending` exists to prevent, reachable again through one stray byte anywhere in the same read. `decode_carrying_tail` consumes only the invalid *sequence* and continues, so the tail is still carried.

`cargo test --lib stream_parser` → **9 passed**.

M3gA-Mind:

**State:** `MERGEABLE`, CI green, but `CHANGES_REQUESTED` with **two unresolved CodeRabbit threads**. Those are the only thing blocking.

**On the fix:** the diagnosis is correct and the bug is real. `from_utf8_lossy` per chunk against an 8 KiB read buffer (`driver.rs:414`) corrupts any multi-byte character that lands on a boundary, and because U+FFFD is legal inside a JSON string it parses fine — silent content corruption, not a parse error. Using `valid_up_to()` for the boundary and `error_len()` to separate *incomplete tail* (hold it) from *genuinely invalid* (replace, don't stall) is exactly the right split; a parser that waited on bytes that will never arrive would hang the stream. Leaving `feed(&str)` untouched is right too.

### The two CodeRabbit threads are legitimate — I checked both against your code

I would not usually say this about assertion-strength nits, but both tests can pass while the bug they name is present:

**1. `feed_bytes_still_replaces_genuinely_invalid_bytes`** asserts only:

```rust
assert_eq!(events.len(), 1, "the line must still be emitted, not withheld");
```

Nothing checks that `0xFF` became U+FFFD. An implementation that silently *drops* the invalid byte passes this test, and the test's own name claims otherwise. Match the `System` event and assert the decoded `session_id` is `"\u{FFFD}"`.

**2. `feed_bytes_preserves_a_character_split_across_chunks`** asserts on `format!("{:?}", events[0])`. `ParseError` retains the original line, so the rendered debug string contains the emoji **even when the event is a `ParseError`** — the assertion cannot distinguish success from that failure. Your `!rendered.contains('\u{FFFD}')` line does carry real weight and should stay, but pair it with matching `ClaudeCodeEvent::Assistant` and asserting `message["text"] == "héllo 世界 🌍 tail"`.

Both are a few lines each and they are worth doing — the whole value of this PR is a guarantee about bytes, so the tests should fail if that guarantee breaks.

**Heads-up on a collision:** #5794 and #5713 both touch `stream_parser.rs`, and the carry-over agreed on #5794 changes the `ClaudeCodeEvent::Result` variant (adds `is_error`) and the `"error"` parse arm. Your diff touches `feed_bytes` and the enum's neighbourhood but not those arms, so it should merge — worth sequencing with #5794 rather than landing blind.

ntdatt812:
Both threads were right, and both tests were weaker than their own names claimed. Fixed in `05d73c61`, rebased onto today's `main`.

**1 — `feed_bytes_still_replaces_genuinely_invalid_bytes`** counted events and nothing else. An implementation that silently *dropped* the `0xFF` also emits one event, and also parses, so the count could not tell replacement from deletion — which is the one thing the test's name promises. It now matches the `System` event and asserts `session_id` is `"\u{FFFD}"`.

**2 — `feed_bytes_preserves_a_character_split_across_chunks`** asserted on `format!("{:?}", events[0])`. `ParseError` carries the original line, so the emoji renders in Debug even when the parse failed — precisely the failure the test exists to catch. It now matches `ClaudeCodeEvent::Assistant` and asserts `message["text"] == "héllo 世界 🌍 tail"`. The `!rendered.contains('\u{FFFD}')` assertion stays; it rules out a replacement that the equality check on a different field would not.

**Mutation, since a test that cannot fail is the thing being fixed here:** reverting `feed_bytes` to the pre-fix shape — `String::from_utf8_lossy` per chunk, no carry — now fails **2** tests, `feed_bytes_preserves_a_character_split_across_chunks` and `an_invalid_byte_does_not_consume_a_split_character_after_it`. Before this commit that mutation left the first of those green. 9 passed / 0 failed unmutated.

On the sequencing with #5794: agreed, and it is already clear. #5794's carry-over (`is_error` on `Result`, the nested `error.message` ladder) is in `f7f5ccda` and touches the enum's `Result` and `"error"` arms; this branch touches `feed_bytes`, `decode_carrying_tail` and `end`. Both are rebased onto the same `main` and both report `MERGEABLE`, in either order.

PR #5719, opened 2026-08-24
fix(claude-code): bound the stderr buffer without splitting a character

ntdatt812:
Sibling of #5718, same file family, different failure mode. Independent of it — either can land first.

## The defect

The stderr drain caps its accumulator with a raw byte index (`driver.rs:417-427`):

```rust
acc.push_str(&String::from_utf8_lossy(&tmp[..n]));
if acc.len() > 16_384 {
    acc.truncate(16_384);
}
```

`String::truncate` takes a **byte** index and panics when it is not a character boundary. Proved with `rustc` on the exact shape:

```
bytes = [97, 195, 169], len = 3        // "aé"
is_char_boundary(2) = false
assertion failed: self.is_char_boundary(new_len)
truncate(2) panicked = true
```

So any stderr over 16 KiB whose 16384th byte lands inside a multi-byte character aborts the drain task.

## Why it is invisible

The panic happens inside `tokio::spawn`, and the join is `stderr_task.await.unwrap_or_default()` (`driver.rs:483`). The `JoinError` is swallowed into an empty string, so `driver.rs:486` reports

```
[claude-code][driver] exit Some(1) stderr=
```

The operator loses the entire error output for the turn, at exactly the moment they need it — a failing turn with a lot of stderr is the case most likely to exceed 16 KiB.

## The fix

The repo already owns the right helper: `util::text::utf8_safe_prefix_at_byte_boundary`, whose doc comment describes this exact problem, and which `tool_result_artifacts/mod.rs:58` and `:432` already use correctly. This call site simply bypassed it — I checked the other two raw `String::truncate` calls in `src/` (`mcp/server/tools/params.rs:496`, `sandbox/cwd_jail/windows.rs:425`) and both operate on provably ASCII input, so this was the only one.

Extracted as a pure `push_bounded(&mut String, &str, usize)` so it can be asserted without spawning a process, matching how every other testable piece of this file is factored.

## Verification

```
cargo test -p openhuman --lib claude_code::driver     9 passed, 0 failed   (6 before)
cargo fmt --check                                     clean
```

**Mutation-checked**, after confirming the edit applied: restoring `acc.truncate(max_bytes)` reproduces the original panic inside the test —

```
panicked at driver.rs:36: assertion failed: self.is_char_boundary(new_len)
test result: FAILED. 8 passed; 1 failed
```

— so the test pins the panic itself, not merely the length.

## Tests

`run_turn` is untested (it spawns a process), which is why the bound was never exercised. Three cases on the extracted helper:

- a cap landing **inside** `é`, which must back up rather than panic;
- a cap landing **exactly on** a boundary, which must not give up a character needlessly — I had this expectation wrong at first and the test caught me;
- 600 chunks of Japanese driven over a 1 KiB cap the way the reader does it, asserting the result stays under the cap and is still valid UTF-8.

Plus a below-cap case and an ASCII exact-cap case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for command output containing multibyte characters.
  * Prevented crashes when diagnostic output exceeds its size limit.
  * Preserved complete characters while limiting accumulated error output.

### DIFF
--- a/app/src/pages/Brain.tsx
+++ b/app/src/pages/Brain.tsx
@@ -40,6 +40,16 @@ const navIcon = (d: string) => (
 const BRAIN_TABS: readonly BrainTab[] = ['welcome', 'graph', 'goals', 'sources', 'sync'];
 
 /**
+ * Backoff ladder for automatically retrying a failed graph load.
+ *
+ * Written out rather than computed from an exponent because the two things a
+ * reader needs — how long the wait is, and that there are exactly three of
+ * them — are then both visible. The last element is the bound: once the ladder
+ * is spent the error stays on screen and manual Refresh is the way back.
+ */
+const RETRY_DELAYS_MS: readonly number[] = [2_000, 4_000, 8_000];
+
+/**
  * Canonical text header (title + one-line description) per functional tab.
  */
 const BRAIN_HEADERS: Record<Exclude<BrainTab, 'welcome'>, { titleKey: string; descKey: string }> = {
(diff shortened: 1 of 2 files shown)


M3gA-Mind:
@ntdatt812 — my mistake, and I'm sorry: I closed this PR by accident and I cannot undo it from my side. Here is exactly what happened and exactly how to get it back.

**What happened.** I was rebasing this branch onto current `main` for you (it was `CONFLICTING`). The rebase itself went fine. But I was working in a checkout that another process moved out from under me between the rebase and the push, so the push sent `main`'s tip (`fa044d388`) to `fix/claude-code-stderr-utf8` instead of your rebased commit. With no commits ahead of base, GitHub auto-closed this PR — and closing it revoked the maintainer-can-modify grant, so I can no longer push the correct commit back to restore it. GitHub also refuses to reopen: *"There are no new commits on the ntdatt812:fix/claude-code-stderr-utf8 branch."*

**Nothing is lost.** Your work, rebased onto `main` with the conflict resolved and your authorship intact, is at:

```
https://github.com/M3gA-Mind/openhuman  branch: recover/5719-stderr-utf8  (commit e814c6dd0)
```

**To restore, one of:**

```bash
# from your clone
git remote add m3ga https://github.com/M3gA-Mind/openhuman.git
git fetch m3ga recover/5719-stderr-utf8
git checkout fix/claude-code-stderr-utf8
git reset --hard m3ga/recover/5719-stderr-utf8
git push --force-with-lease
```

That puts commits back on the branch, after which this PR can be reopened (the Reopen button should light up once the branch is non-empty). If it will not reopen, open a fresh PR from the same branch and I will re-link it.

**What the rebase actually changed** — one conflict, in `src/openhuman/inference/provider/claude_code/driver.rs`:

`main` has since moved the driver's inline `mod tests { … }` out into a separate `driver_tests.rs` (the `_part_NN.rs` / test-file split from #5856/#5857). Your three new tests were written against the old inline module. The resolution keeps `main`'s out-of-line form:

```rust
#[cfg(test)]
#[path = "driver_tests.rs"]
mod tests;
```

and moves your `push_bounded_never_splits_a_character`, `push_bounded_keeps_everything_below_the_cap` and `push_bounded_handles_an_ascii_cap_exactly` into `driver_tests.rs` verbatim, de-indented one level. `driver_tests.rs` opens with `use super::*;`, so `push_bounded` is still in scope. Your production hunks — `STDERR_DIAGNOSTIC_CAP`, `push_bounded`, and the call site in the stderr drain — applied without conflict and are unchanged.

The fix itself is right and worth landing: `String::truncate` on a byte index really does panic mid-character, and the `unwrap_or_default()` on the join really does swallow it into an empty `stderr=`. Again, sorry for the detour.

ntdatt812:
No harm done — thank you for writing down exactly what happened and where the work was. That is a much better outcome than a silent force-push.

Branch restored. GitHub still refused to reopen this one — its recorded head is `fa044d388`, so it sees no commits — so the work is now at #5980, from the same branch. Please re-link it when you get a moment.

I did not take `recover/5719-stderr-utf8` on trust, since it is a force-push into my branch from another fork. I verified it instead:

- **Authorship intact** — `Nguyen Thanh Dat <ntdat812@gmail.com>`, original author date preserved.
- **Content is mine** — the production hunks (`STDERR_DIAGNOSTIC_CAP`, `push_bounded`, the call site in the stderr drain, and the `use` of `utf8_safe_prefix_at_byte_boundary`) are byte-identical to my `130ce0231`, and the three tests are verbatim, de-indented one level.
- **Your split resolution is the right one** — `driver.rs` keeps `main`'s out-of-line `#[path = "driver_tests.rs"] mod tests;`, and `driver_tests.rs` opens with `use super::*;`, so `push_bounded` stays in scope without re-exporting it.

Your recovery branch was cut against `fa044d388`, which is now 47 commits behind, so rather than pushing it as-is I cherry-picked it onto today's `main` — clean, no conflict.

**Verification on the rebased head:** the three `push_bounded_*` tests pass, and the layout gate passes — `driver.rs` 524 lines, `driver_tests.rs` 224, both under 750, tests external.

One thing worth flagging for whoever else you are rebasing for: the auto-close revoked maintainer-can-modify, so the same accident on any other PR of mine will need the same recovery path. Happy to do the rebases myself if that is easier — for the other conflicting ones you reviewed today, I am working through them.

PR #5743, opened 2026-08-24
fix(memory): stop recall from hiding a failed chunk read as a missing chunk

ntdatt812:
## `.ok()` answers two different questions with one word

`recall_rpc` hydrates each retrieved leaf like this:

```rust
let row = conn.query_row(SQL, params![chunk_id], |r| { ... }).ok();
if let Some(r) = row {
    out.push((r, score));
}
```

`.ok()` collapses two outcomes that are not the same thing:

| outcome | what it means | what happens |
|---|---|---|
| `QueryReturnedNoRows` | the tree points at a chunk the store no longer has | skipped — correct |
| any other `Err` | **the read failed** | skipped, silently |

The second row is the bug. A chunk that cannot be mapped — a column holding the wrong type after a migration or an out-of-band write — drops out of the result set with no error and no log. Recall then returns fewer hits than it actually retrieved, and from the outside that is indistinguishable from a recall that simply found less.

## The fix

Pull the query into `hydrate_chunk()`, which answers the two questions separately:

```rust
Ok(row) => Ok(Some(row)),
Err(rusqlite::Error::QueryReturnedNoRows) => Ok(None),
Err(e) => Err(e),
```

The caller still **skips** a bad row rather than failing the whole recall — one broken chunk should not sink a search — but it now says which chunk and why:

```rust
Err(e) => log::warn!("[memory_tree::read::recall] skipping chunk {chunk_id}: {e}"),
```

Same behaviour for every healthy row; the only change is that a failure stops being invisible.

## Tests

Three cases in the existing `read_rpc_tests.rs` harness:

- a seeded row is read back, preview included
- a missing id is `Ok(None)`, not an error
- a row whose `token_count` holds text comes back as `Err` instead of vanishing

I first tried to force the failure with `content = NULL` and the DB refused it (`NOT NULL constraint failed: mem_tree_chunks.content`), so the test uses a column-type mismatch instead — SQLite affinity lets a value of the wrong type sit in the column, which is exactly the class of damage this path used to hide.

Mutation-checked — restoring the swallow (`Err(_) => Ok(None)`) turns the third test red:

```
failures:
    hydrate_chunk_surfaces_a_broken_row_instead_of_hiding_it
test result: FAILED. 2 passed; 1 failed
```

Full `read_rpc` suite: `41 passed; 0 failed; 4 ignored`. `cargo fmt --check` and `cargo clippy --lib` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recall data loading so missing records are skipped without disrupting results.
  - Ensured unreadable or malformed records are handled individually while allowing other records to load.
  - Preserved clear error reporting for data that cannot be read.

- **Tests**
  - Added coverage for successful loading, missing records, and malformed records.

### DIFF
--- a/src/openhuman/memory/read_rpc/chunks.rs
+++ b/src/openhuman/memory/read_rpc/chunks.rs
@@ -287,8 +287,6 @@ pub async fn recall_rpc(
     query: String,
     k: u32,
 ) -> Result<RpcOutcome<RecallResponse>, String> {
-    use rusqlite::params;
-
     let limit = k.clamp(1, MAX_LIST_LIMIT) as usize;
     log::debug!(
         "[memory_tree::read::recall] query_len={} k={}",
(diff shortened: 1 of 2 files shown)


ntdatt812:
The red **Rust Core Coverage (cargo-llvm-cov)** lane here is base breakage, not this diff.

- Failing test: `openhuman::memory::ops::documents::tests::envelope_memory_handlers_report_counts_and_statuses`, panicking at `src/openhuman/memory/ops/documents.rs:843` (`expect("query data")` — the `memory_query_namespace` envelope comes back without `data`).
- The same test fails on `main` at **e1c332bf08b332dacc5dcdb04c1d8df6f8401e13** — which is exactly this PR's base commit — in the push run [32704367921](https://github.com/tinyhumansai/openhuman/actions/runs/32704367921) (2026-08-24 08:03Z): `test result: FAILED. 11430 passed; 1 failed`. The same lane is also the failing job on the two preceding `main` runs ([32680983807](https://github.com/tinyhumansai/openhuman/actions/runs/32680983807), [32661013798](https://github.com/tinyhumansai/openhuman/actions/runs/32661013798)).
- This PR touches one function in `src/openhuman/memory/read_rpc/chunks.rs` (`recall_rpc` chunk hydration). It does not touch `memory/ops/documents.rs`, the query-namespace path, or anything the failing assertion reads. The lane only runs `openhuman::memory` here because the diff is inside that module tree, which is why the pre-existing failure surfaces on this PR and not on unrelated ones.

Happy to rebase once the `main` failure is fixed if you'd rather see the lane green before merging.

ntdatt812:
Flagging the red **Rust Core Coverage** job before anyone has to dig into it.

The failing case is `openhuman::memory::ops::documents::tests::envelope_memory_handlers_report_counts_and_statuses`, panicking at `documents.rs:843: query data`. This PR touches `memory/read_rpc/chunks.rs` and `read_rpc_tests.rs` only — `ops/documents.rs` is a different module, and nothing in the diff is reachable from that test.

What I could measure, running the same filter the job uses (`--lib 'openhuman::memory'`) on both endpoints:

| | passed | failed |
|---|---|---|
| `main`, clean | 531 | **23** |
| this branch | 534 | **23** |

Same failure count; the +3 are this PR's own cases. That test is in the failing set on clean `main` here too — on Windows it dies in the `tinymemory_module.dll` loader, which is an environment failure of mine, not the CI one, so I can't reproduce the CI symptom directly.

So I can show the change adds three passing tests and moves nothing else in the module, but I can't prove from here whether the CI failure is flaky or environmental. I don't have re-run rights on the repo — happy to rebase or re-push if you want another run against it.

ntdatt812:
Closing — the bug is gone on `main`, fixed by a rewrite of the same code rather than by this patch. Verified end to end before closing rather than assuming the conflict meant obsolescence.

The per-leaf `conn.query_row(SQL, …).ok()` this PR was about no longer exists. `recall_rpc` now hydrates every leaf in **one** `list_chunk_details` call, and the ambiguity `.ok()` created is resolved in both directions:

- **A failed read is loud.** `list_chunk_details(&query, None).await.map_err(|e| format!("list_chunk_details: {e}"))?` propagates. It cannot be mistaken for a missing chunk any more, because it does not produce a result set at all.
- **A missing chunk is still silent, and still correct** — the id contributes no row, and `pair_leaves_with_rows` walks the leaves through the id map.

I also checked the layer below, since a silent drop of this shape can move down rather than disappear. `tinycortex`'s `list_chunk_details` finishes with `.collect::<rusqlite::Result<Vec<_>>>()`, so a row that fails to map aborts the collect instead of being skipped. No `.ok()` anywhere on that path.

The new shape does make the tradeoff differently from what I proposed here: one unmappable chunk now fails the whole recall, where this PR would have skipped it and logged. I think the rewrite's choice is defensible and I'm not going to relitigate it in a stale branch — a batched read that half-succeeds is harder to reason about than one that fails.

The `debug!` line carrying `leaves=/ids=/rows=` covers what my logging was for.

Nothing to carry over. Thanks — the version on `main` is better than the one I wrote.

PR #5752, opened 2026-08-24
fix(agent): append tool-policy boundary after the cacheable prompt prefix

### LINKED ISSUE #5704: Tool Policy Boundary is prepended to the system prompt, putting session-varying content ahead of the cacheable prefix
## Summary

   constant heading instead.

## Where

```rust
}
Ok(prompt)
```

The block is rendered by

```
## Tool Policy Boundary
- Agent: <agent_id>
- Channel: <channel>
- Entry point: <entrypoint>
- Allowed permission: <level>
- Risk: <level>
```

message either way.

> ## KV cache stability
>

opening line.

## Suggested resolution

```rust
}
```

## Repro

prefix past byte 0.

## Related

gac0812:
## Summary

- Append the session-varying `## Tool Policy Boundary` block after the assembled system prompt, instead of prepending it.
- This keeps the stable prefix (safety preamble, tool catalogue, workspace) byte-identical across agents/channels so the inference backend can reuse its prefix cache.
- Matches the existing `DateTimeSection` convention: volatile content belongs at the tail, not at offset 0.
- Regression test now asserts the boundary is present, is not the prompt prefix, and is the last markdown section.

## Problem

`build_system_prompt` prepended `render_tool_policy_boundary(...)` in front of the fully assembled system prompt. That block is session-scoped (agent, channel, entrypoint, allowed tools). Putting it first moves the first diverging byte to offset 0 and defeats prefix caching for everything behind it. It also made every restricted agent open with the same heading instead of its persona.

Tracked in https://github.com/tinyhumansai/openhuman/issues/5704. No existing open PR covers this.

## Solution

Change the join from `{boundary}\n\n{prompt}` to `{prompt}\n\n{boundary}`. The model still sees the full boundary; only placement changes. The test helper `make_agent_with_builder_and_dispatcher` now calls `install_for_tests()` so the existing prompt test can run in isolation (same seam wiring as `make_agent`).

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — new assertions cover the changed format string; helper wiring is the same path `make_agent` already used
- [x] Coverage matrix updated — N/A: behaviour-only change
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no coverage-matrix feature ID
- [x] Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md)) — N/A: prompt assembly only

## Impact

- Runtime: desktop/CLI agent turns. Restricted sessions still get the boundary; it is now after the stable prefix.
- Performance: prefix cache can hit across agents/channels that share the stable sections.
- No security, migration, or wire-format change.

## Related

- Closes #5704
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/tool-policy-boundary-prefix-cache`
- Commit SHA: `268eb8a5113975f8e9f8ad70d80b8592dcfcaa74`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: Rust-only change
- [x] `pnpm typecheck` — N/A: Rust-only change
- [x] Focused tests: `cargo test --lib openhuman::agent::harness::session::turn::tests::system_prompt_includes_tool_policy_boundary -- --exact` (pass); sibling `set_agent_definition_name_refreshes_tool_policy_identity` (pass)
- [x] Rust fmt/check (if changed): `cargo fmt` on the touched files
- [x] Tauri fmt/check (if changed): N/A: Tauri crate not touched

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: tool-policy boundary is appended after the stable system prompt instead of prepended
- User-visible effect: none in the UI; restricted agent turns keep the same policy text, later in the system prompt

### Parity Contract

- Legacy behavior preserved: unrestricted sessions still omit the boundary; allowed/restricted tool listing unchanged
- Guard/fallback/dispatch parity checks: renderer still returns `None` when there are no restrictions

- Duplicate PR(s): none found (search for #5704 and "tool policy boundary" against open PRs)
- Canonical PR: this PR

## Summary by CodeRabbit

* **Improvements**
  * Updated agent instructions so tool-policy boundaries appear after the standard system prompt, improving consistency and clarity in prompt handling.
  * Strengthened validation to ensure policy boundaries remain in the intended position and are not treated as the prompt prefix or followed by unrelated sections.
  * Improved configuration validation for runtime proxy overrides, including empty overlays and state preservation.

* **Documentation**
  * Updated integration guidance to reflect the revised policy-boundary rendering behavior.

### DIFF
--- a/src/openhuman/agent/harness/session/turn/context.rs
+++ b/src/openhuman/agent/harness/session/turn/context.rs
@@ -336,8 +336,15 @@ impl Agent {
         // prompt-building call-site — main agent, sub-agent runner,
         // channel runtimes — shares one builder configuration.
         let mut prompt = self.context.build_system_prompt(&ctx)?;
+        // Append, never prepend. The boundary is session-varying
+        // (agent / channel / entrypoint / allowed tools). Putting it
+        // first moves the first diverging byte to offset 0 and defeats
+        // the inference backend's prefix cache for the stable sections
+        // behind it (safety preamble, tool catalogue, workspace). Same
+        // reason `DateTimeSection` is omitted from `for_subagent` and
+        // sits after the stable sections in `with_defaults()`.
         if let Some(boundary) = render_tool_policy_boundary(&self.tool_policy_session, 2048) {
-            prompt = format!("{boundary}\n\n{prompt}");
+            prompt = format!("{prompt}\n\n{boundary}");
         }
         Ok(prompt)
     }
--- a/src/openhuman/agent/harness/session/turn_tests.rs
+++ b/src/openhuman/agent/harness/session/turn_tests.rs
@@ -428,6 +428,9 @@ fn make_agent_with_builder_and_dispatcher(
     context_config: crate::openhuman::config::ContextConfig,
     tool_dispatcher: Box<dyn ToolDispatcher>,
 ) -> Agent {
+    // Same seam wiring as `make_agent`: create_memory fails closed
+    // without an EmbeddingHost, which the host installs at startup.
+    crate::openhuman::memory::host_impls::install_for_tests();
     let workspace = tempfile::TempDir::new().expect("temp workspace");
     let workspace_path = workspace.path().to_path_buf();
     std::mem::forget(workspace);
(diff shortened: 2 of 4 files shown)


M3gA-Mind:
Thanks for this, @gac0812 — closing in favour of #5821 for #5704, with **two pieces here that should come back as their own PRs**.

**Why #5821:** it's `MERGEABLE` where this is `CONFLICTING`, it's one file, and it extracts `append_tool_policy_boundary(prompt, boundary)` so the placement becomes a unit test rather than something you need to stand up a session to exercise. Your diagnosis was the same and was reached first — the difference is scope and current state, not correctness.

**Please do re-file these two, separately:**

1. **The `tools/agent_policy/README.md` pointer.** It still names the pre-kernelization `session/turn.rs`. That's a genuine doc-staleness bug you found on the way past, it's unrelated to #5704, and it'll merge on its own in a minute.
2. **The `apply_env_overrides_commits_side_effects_to_runtime_proxy` refactor** in `config/schema/load_tests.rs`. A proxy env-var CI flake fix is worth having — it just can't be reviewed inside a prompt-ordering PR, because nobody looking at #5704 knows to check it.

Both are real work and neither is in #5821. Bundling them is the only reason this one is being closed rather than merged.

PR #5761, opened 2026-08-25
fix(web_chat): preserve backslash for unknown JSON escape sequences

Ricardo-M-L:
Fix data loss in extract_provider_error_detail: unrecognized JSON escape sequences silently dropped the backslash character, corrupting upstream error messages.

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider error details now preserve backslashes in unknown escape sequences, improving the accuracy of displayed error messages.

### DIFF
--- a/src/openhuman/web_chat/web_errors_part_01.rs
+++ b/src/openhuman/web_chat/web_errors_part_01.rs
@@ -157,7 +157,10 @@ pub(crate) fn extract_provider_error_detail(err: &str) -> Option<String> {
                         '\\' => out.push('\\'),
                         'n' => out.push('\n'),
                         't' => out.push('\t'),
-                        other => out.push(other),
+                        other => {
+                            out.push('\\');
+                            out.push(other);
+                        }
                     }
                 }
             }


M3gA-Mind:

**The bug is real and still unfixed on `main`.** Worth saying clearly, because the diff is four lines and easy to under-read: the function's own doc comment claims *"handle `\"` and `\\` only; everything else passes through"*, and the code does not pass anything through — `other => out.push(other)` eats the backslash. So `\r` renders as `r`, `\uD83D` as `uD83D`, and an upstream provider message reaches the user with characters silently deleted. Making the code match its documented contract is the right instinct.

Three things stand between this and mergeable.

## 1. The escaped solidus — @chatgpt-codex-connector's thread is correct, please act on it

`\/` is legal JSON and extremely common in provider error bodies (`https:\/\/api.openai.com\/v1\/...`). The current buggy behaviour handles it *by accident*: dropping the backslash yields the correct `/`. Your change turns that into a literal `\/`, so a URL in an error message gets visibly worse for the one case that was previously right. That is a real regression, not a nit.

While you are in that match, the same argument covers the rest of the JSON escape set — you are already decoding `n` and `t`, so stopping there is arbitrary:

```rust
'"'   => out.push('"'),
'\\'  => out.push('\\'),
'/'   => out.push('/'),
'n'   => out.push('\n'),
't'   => out.push('\t'),
'r'   => out.push('\r'),
'b'   => out.push('\u{8}'),
'f'   => out.push('\u{c}'),
other => { out.push('\\'); out.push(other); }   // your fix, now only for genuinely unknown escapes
```

That leaves `\uXXXX` passing through as literal `\uXXXX` — visibly unrendered rather than silently mangled, which is the right trade for an error-display path and worth a one-line comment saying so.

## 2. No test

There is a natural home: `web_tests_part_01_tests.rs` already has `extract_provider_error_detail_pulls_openai_message` and `..._returns_none_for_transport_errors`. Two cases would cover this — a URL with `\/`, and a genuinely unknown escape such as `\q` — and the repo's testing policy wants at least one failure/edge case. Please also check each new assertion fails with your fix reverted; an escape test that passes either way proves nothing.

## 3. It conflicts, and the reason is not obvious

`main` split this file: `extract_provider_error_detail` now lives in **`src/openhuman/web_chat/web_errors_part_01.rs`**, and the match arm you are changing is at line 160 there. `web_errors.rs` still exists but no longer holds the function, which is why git can't place your hunk. Move the change to `web_errors_part_01.rs` when you rebase onto `fa044d388`.

Your two cancelled checks (`PR Submission Checklist`, `Coverage Matrix Sync`) are just stale runs. Note the checklist gate will want the submission checklist in the PR body — it currently has none, and the gate requires every item either ticked or ticked with text beginning `N/A`.

Happy to re-look once it is rebased.

senamakel:
This PR has been superseded by #6227, which includes the same fix plus the additional standard JSON escape decoders requested in review (\/, \r, \n, \t, \b, \f) and regression tests. Closing in favor of the new PR.

senamakel:
Closing per maintainer's instruction — this PR is superseded by #6227, which includes the same fix plus the additional standard JSON escape decoders (\/, \r, \b, \f) and regression tests requested in review.

PR #5762, opened 2026-08-25
chore(windows): make clippy -D warnings pass on a Windows host

ntdatt812:
## No commit can be pushed from a Windows checkout

`.husky/pre-push` runs `cargo clippy -p openhuman -- -D warnings` and then the same for `app/src-tauri`. On a Windows host both fail before reaching anything the developer changed — 11 errors in the core crate, 4 in the Tauri crate.

Linux CI never sees them: every one sits in a path guarded by `#[cfg(windows)]` / `#[cfg(not(unix))]`, so the lint only fires where those arms are compiled. The practical effect is that the hook is unpassable on Windows, and the only way to push is `--no-verify` — which disables the TypeScript, format and lint checks too.

## What changed

**Imports only one platform uses, gated the way their neighbours already are:**

| file | import | used by |
|---|---|---|
| `core/auth.rs`, `security/pairing.rs` | `std::io::Write` | only the `#[cfg(unix)]` arm of the token writer — the `#[cfg(not(unix))]` arm calls `std::fs::write` |
| `composio/trigger_history.rs` | `fs2::FileExt` | only the `#[cfg(not(windows))]` lock/unlock path — Windows serializes through the process-local `Mutex` above it |

In both files the gate matches the `#[cfg(unix)] use std::os::unix::fs::OpenOptionsExt` line sitting directly below.

**Two imports were simply redundant:**

`inference/local/process_util.rs` and `inference/voice/local_speech.rs` import `std::os::windows::process::CommandExt` to call `creation_flags`, but that is an inherent method on `tokio::process::Command` under Windows, so the trait import adds nothing.

**Ordinary lints in Windows-only code:**

- `sandbox/cwd_jail/windows.rs` — drop the unused `HANDLE` import; drop `mut` from `dacl_present` (only ever read, via `let _ =`); pass `&ea` to `SetEntriesInAclW`, whose second parameter is `*const EXPLICIT_ACCESS_W`; add the `Default` impl clippy asks for beside `AppContainerBackend::new()`.
- `security/keyring/encrypted_store.rs` — the `#[cfg(windows)]` arm **shadows** `msg` rather than mutating it, so `mut` was unnecessary on both platforms; the `#[cfg_attr(not(windows), allow(unused_mut))]` escape hatch goes with it.
- `platform/doctor/core.rs`, `src-tauri/claude_code.rs` — on Windows the sibling `#[cfg]` blocks are stripped, so the remaining block is the function tail and `return` is noise. The macOS/Linux arms of `claude_code_login_launch` already end in a bare expression; this makes the Windows arm match.
- `src-tauri/core_process.rs` — `parse_lsof_pid` parses `lsof` output and is reachable only from the `#[cfg(unix)]` `find_pid_on_port`, so it is dead on Windows. Gated with `#[cfg(unix)]`, along with its import and test.
- `src-tauri/deep_link_ipc_windows.rs` — name the boxed callback (`type LiveHandler`) so the static and its accessor stop tripping `clippy::type_complexity`.

## Verification

No behaviour changes — every edit is an import gate, a removed `mut`/`return`, a `&mut` → `&`, a type alias, or a `Default` impl.

```
cargo clippy -p openhuman -- -D warnings                        clean
cargo clippy --manifest-path app/src-tauri/Cargo.toml -- -D warnings   clean
cargo clippy --manifest-path app/src-tauri/Cargo.toml --all-targets    clean
cargo fmt -- --check                                            clean
```

Tests for every touched module:

```
openhuman::sandbox            76 passed
openhuman::security          819 passed
openhuman::platform::doctor   22 passed
composio::trigger_history      4 passed
core::auth                    19 passed
```

This branch was pushed with the pre-push hook **running and passing** on Windows — which is the point of the change.

## Summary by CodeRabbit

- **Refactor**
  - Improved platform-specific compilation across Unix and Windows environments.
  - Simplified internal control flow and type definitions without changing application behavior.
  - Added a default construction path for the Windows sandbox backend.
  - Removed unused imports and unnecessary mutability.

- **Chores**
  - Updated platform-specific tests and build configuration to avoid compiling unsupported Unix-only components on other platforms.
  - No user-facing functionality or runtime behavior changed.

### DIFF
--- a/app/src-tauri/src/claude_code.rs
+++ b/app/src-tauri/src/claude_code.rs
@@ -30,7 +30,7 @@ pub fn claude_code_login_launch() -> Result<String, String> {
             .args(["/c", "start", "", "cmd", "/k", "claude login"])
             .spawn()
             .map_err(|e| format!("failed to open cmd: {e}"))?;
-        return Ok("cmd".into());
+        Ok("cmd".into())
     }
 
     #[cfg(target_os = "macos")]
--- a/app/src-tauri/src/core_process.rs
+++ b/app/src-tauri/src/core_process.rs
@@ -1001,6 +1001,7 @@ fn find_pid_on_port(port: u16) -> Option<u32> {
 }
 
 /// Pure parse of `lsof -t` output (one pid per line; first wins).
+#[cfg(unix)]
 fn parse_lsof_pid(stdout: &str) -> Option<u32> {
     stdout
         .lines()
--- a/app/src-tauri/src/core_process_tests.rs
+++ b/app/src-tauri/src/core_process_tests.rs
@@ -1,8 +1,11 @@
 use super::{
     current_rpc_token, default_core_port, generate_rpc_token, is_expected_port_clash,
-    is_openhuman_root_body, parse_lsof_pid, parse_netstat_pid, parse_ps_comm, parse_tasklist_name,
+    is_openhuman_root_body, parse_netstat_pid, parse_ps_comm, parse_tasklist_name,
     validate_kill_target, CoreProcessHandle, PortOwner, RecoveryOutcome,
 };
+// lsof is a unix tool, so its parser is only compiled there.
+#[cfg(unix)]
+use super::parse_lsof_pid;
 use std::sync::{Mutex, MutexGuard, OnceLock};
 
 fn env_lock() -> MutexGuard<'static, ()> {
@@ -307,6 +310,7 @@ fn expected_port_clash_classifier_rejects_unknown_probe_shapes() {
     ));
 }
 
+#[cfg(unix)]
 #[test]
 fn parse_lsof_pid_picks_first_pid() {
     assert_eq!(parse_lsof_pid("12345\n"), Some(12345));
--- a/app/src-tauri/src/deep_link_ipc_windows.rs
+++ b/app/src-tauri/src/deep_link_ipc_windows.rs
@@ -95,14 +95,17 @@ pub(crate) fn try_forward_deep_links() -> ForwardResult {
     ForwardResult::NoPrimary
 }
 
+/// Callback the app installs to receive deep-link URLs as they arrive.
+type LiveHandler = Box<dyn Fn(String) + Send + Sync>;
+
 static PENDING_URLS: OnceLock<Arc<Mutex<Vec<String>>>> = OnceLock::new();
-static LIVE_HANDLER: OnceLock<Mutex<Option<Box<dyn Fn(String) + Send + Sync>>>> = OnceLock::new();
+static LIVE_HANDLER: OnceLock<Mutex<Option<LiveHandler>>> = OnceLock::new();
 
 fn pending_queue() -> &'static Arc<Mutex<Vec<String>>> {
     PENDING_URLS.get_or_init(|| Arc::new(Mutex::new(Vec::new())))
 }
 
-fn live_handler() -> &'static Mutex<Option<Box<dyn Fn(String) + Send + Sync>>> {
+fn live_handler() -> &'static Mutex<Option<LiveHandler>> {
     LIVE_HANDLER.get_or_init(|| Mutex::new(None))
 }
 
--- a/src/core/auth.rs
+++ b/src/core/auth.rs
@@ -55,10 +55,11 @@
 //!   internal bearer or a stable user-managed external API key stored under
 //!   `openhuman::inference::http::EXTERNAL_OPENAI_COMPAT_PROVIDER`.
 
-use std::io::Write as _;
 use std::path::Path;
 use std::sync::OnceLock;
 
+#[cfg(unix)]
+use std::io::Write as _;
 #[cfg(unix)]
 use std::os::unix::fs::OpenOptionsExt as _;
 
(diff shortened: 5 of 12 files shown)


ntdatt812:
The red `Rust Core Coverage` job on this branch is not this diff. It is a latent test-isolation bug that this PR's *changed-module filter* exposed, and I have opened #5769 to fix it.

**The two failures**

```
openhuman::integrations::composio::ops::tests::composio_delete_connection_clear_memory_cascades_source_tree_and_content_file
openhuman::integrations::composio::ops::tests::composio_delete_connection_clear_memory_cascades_live_sealed_tree_and_file

called `Result::unwrap()` on an `Err` value: "[composio] delete_connection cannot
resolve the memory client: no EmbeddingHost installed — the host must call
memory::embedding_host::set_embedding_host during startup wiring, before any
memory work begins"
```

Those tests resolve the process-global memory client and never call `install_for_tests()`, so they pass only when an earlier test in the same binary installs the seams first. `ops_tests.rs` already warns about this hazard in a comment on `notion_cleanup_targets_include_synced_page_sources`; these are the tests that omit the call.

**Why it fired here and not on `main`**

`Rust Core Coverage` runs only the modules a PR changed, and the two filters differ:

| | filter | result |
|---|---|---|
| `main` @ `ac4e671eb` ([run](https://github.com/tinyhumansai/openhuman/actions/runs/32766076997)) | `openhuman::agent … **openhuman::memory** …` | `5268 passed; 0 failed` |
| this PR ([run](https://github.com/tinyhumansai/openhuman/actions/runs/32800083836)) | `core::auth openhuman::inference openhuman::integrations openhuman::platform openhuman::sandbox openhuman::security` | `2743 passed; 2 failed` |

With `openhuman::memory` in the filter you get one of the 104 `install_for_tests()` callers for free. Without it, nothing installs the seams and the composio tests go red.

**Reproduced with no changes in the tree**, clean `main` @ `ac4e671eb`:

```
cargo test -p openhuman --lib -- composio_delete_connection_clear_memory
test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 11507 filtered out
```

All **four** — not just the two CI happened to schedule first. Which ones go red depends on ordering, so this would have bitten the next PR too.

**Why it cannot be this diff.** The only composio file here is `trigger_history.rs`, one line:

```rust
+#[cfg(not(windows))]
 use fs2::FileExt;
```

On Linux that predicate is true and the import is kept, so the Linux build of this branch is identical to `main` in that file. None of the other eleven files is in `openhuman::integrations`.

#5769 adds the missing `install_for_tests()` to all four tests. With it applied, this PR's exact filter set runs `2733 passed; 1 failed` locally on Windows, the one failure being `platform::connectivity::rpc::tests::port_excluded_error_rejects_addr_in_use_and_others` (`PermissionDenied` binding a port), which fails identically on a clean tree on this box and is green on Linux CI.

Once #5769 lands I will re-run this branch.

ntdatt812:
#5769's own CI settles it: 19 checks green, and `Rust Core Coverage` passed with the filter

```
--lib -- openhuman::integrations
```

no `openhuman::memory` in it — exactly the selection that goes red on this branch. All four composio tests ran and passed under it:

```
composio_delete_connection_clear_memory_cascades_live_sealed_tree_and_file ... ok
composio_delete_connection_clear_memory_cascades_source_tree_and_content_file ... ok
composio_delete_connection_clear_memory_deletes_slack_source ... ok
composio_delete_connection_clear_memory_keeps_other_gmail_connections ... ok

test result: ok. 518 passed; 0 failed
```

So the two red tests here are not this diff, and #5769 fixes them without any change to production code. Once it lands I will re-run this branch.

M3gA-Mind:

**State:** `CONFLICTING`, 1,599 commits behind `main`. The premise still holds — a Windows checkout still cannot pass `.husky/pre-push`, and every fix here is still applicable.

**On the change:** it is the right kind of PR. Every edit is an import gate, a removed `mut`/`return`, a `&mut`→`&`, a type alias, or a `Default` impl, and the reasoning in the table (gating each import to match the `#[cfg]` on the line below it) matches the surrounding style rather than reaching for `#[allow]`. That is the distinction that makes this worth landing.

### Rebasing it — one conflict, and it is not what it looks like

Only `src/openhuman/platform/doctor/core.rs` conflicts, and the whole-file diff is misleading. openhuman#5856/#5857 split large sources into `_part_NN.rs` fragments pulled in by `include!()`. On `main`, `platform/doctor/core.rs` is now **five lines**:

```rust
#[cfg(test)]
#[path = "core_tests.rs"]
mod tests;
include!("core_part_01.rs");
include!("core_part_02.rs");
```

Your one-line change moved with the body. Take `main`'s `core.rs` verbatim and apply the edit to **`src/openhuman/platform/doctor/core_part_01.rs:495-499`**, where `available_disk_space_mb` now lives:

```rust
    #[cfg(target_os = "windows")]
    {
-       return available_disk_space_mb_windows(path);
+       available_disk_space_mb_windows(path)
    }
```

I confirmed the other eleven files rebase clean — they auto-merge with no conflict.

**Two things to check while you are in there**, both consequences of the same split:

- The layout gate (`scripts/ci/check-openhuman-rust-layout.mjs`) is a **monotonic ratchet**: files under `src/openhuman` are capped at 750 lines, with four pinned legacy exceptions that may not grow. Adding lines to any `_part_NN.rs` can trip it even when the change is trivial.
- Inline `#[cfg(test)] mod tests { ... }` is now rejected outright; test modules must be sibling `*_tests.rs` files. Your `core_process_tests.rs` change is already in the right shape.

**Current red checks are not diagnostic** — with the branch this far behind, `PR CI Gate` and `Rust Core Coverage` are reporting against a stale merge base. Rebase first, then read CI.

Issue #5636, opened 2026-08-20
Pre-prod: ac.inference.sh MCP server transport drops repeatedly (14x) — reconnects but unstable


Al629176:
## Summary

The `ac.inference.sh/mcp` MCP server connects successfully on boot (25 tools registered) but drops its transport connection repeatedly throughout the session, triggering reconnect cycles. 14 transport drop events observed.

## Log evidence (pre-prod, 2026-08-20)

```
16:03:06 INF [mcp-registry] boot: connected server_id=00bbad23-4dab-443e-9f0b-ae73af1d6bb3
qualified=ac.inference.sh/mcp tools=25

18:41:12 WRN [mcp-supervisor] server_id=00bbad23-4dab-443e-9f0b-ae73af1d6mb3
qualified=ac.inference.sh/mcp transport dropped; reconnecting
[repeats 14x across session]
```

Unlike the GitHub MCP issue in staging (#5600), this server connects — the transport drops under load or after idle periods.

## Impact

During a transport drop, any agent tool call routed to `ac.inference.sh/mcp` fails until the reconnect completes. With 14 drops in one session, this is a persistent reliability issue for inference tools.

## Fix direction

- Investigate whether `ac.inference.sh/mcp` has a server-side idle timeout that drops the connection.
- Add a keepalive ping on the MCP transport layer to prevent idle drops.
- Confirm the MCP supervisor reconnect succeeds quickly and tools are re-registered without requiring a full app restart.

M3gA-Mind:
**How to fix this**

**Root cause:** the supervisor never measures what it reports. `probe_alive` deliberately collapses three different conditions — missing entry, transport error, and *probe timeout* — into a single `false`, and the supervisor turns that one bool into a specific causal claim ("the transport dropped") **and** an unconditional teardown of the session. A server that is alive but answers slower than the 8-second probe window is therefore disconnected by the supervisor itself, and the "reconnect" that follows is repairing damage the supervisor caused. Nothing in these logs distinguishes that from a real drop.

**Where:**
- `vendor/tinymcp/crates/tinymcp/src/registry/connections/types.rs:319-323` — the doc comment makes the collapse explicit: *"A missing entry, a transport error, and a timeout all report `false`: each means 'not usable, reconnect', and distinguishing them would give a caller a choice it has no different response to."* That premise is the bug — there **is** a different correct response to a timeout.
[...]

Issue #5600, opened 2026-08-19
Staging: GitHub MCP server fails — uvx not found on PATH (uv not installed)


Al629176:
## Summary

The GitHub MCP server (`com.mcparmory/github`) fails to connect on every retry because `uvx` is not found on the resolved PATH. `uvx` is part of the `uv` Python package manager, which is not installed on the staging machine. The server retries with exponential backoff (5s → 10s) but never succeeds.

## Log evidence (staging, 2026-08-19)

```
02:09:15 WRN [stdio] stdio MCP command not found on resolved PATH command=uvx

02:09:15 WRN [mcp-supervisor] reconnect failed server_id=9325a4b8-1988-48c0-9b13-16e03367fe76
qualified=com.mcparmory/github failures=1 next_retry_in=5s
err=`uvx` was not found. This MCP server needs uv (Python), which doesn't appear to be installed.
Install it from https://docs.astral.sh/uv/ and restart OpenHuman.

02:10:15 WRN [mcp-supervisor] reconnect failed ... failures=2 next_retry_in=10s
```

## Impact

- GitHub MCP server is completely unavailable in staging — any agent tool call relying on GitHub context will fail.
- Repeated retries add background noise to logs and waste reconnect budget.

## Fix direction

- Either bundle `uv`/`uvx` with the staging DMG, or
- Document `uv` as a required system dependency and check for it at first launch with a clear user-facing error instead of a background retry loop.

shivakumar777-web:
Hi! I'd like to work on this issue.

M3gA-Mind:
**How to fix this**

**First, a heads-up that changes where this work goes:** the whole MCP client — transports, registry, connections, supervisor — was extracted out of this repo by `050b2d687` (*feat(mcp): extract MCP client into tinymcp library*). It now lives in **`tinyhumansai/tinymcp`**, vendored here as the `vendor/tinymcp` submodule with a path dependency. Every path in the issue body and in PR #5618 is gone from this tree. A fix is now a PR to `tinyhumansai/tinymcp` plus a submodule bump here.

**Root cause.** Two separable things, and the issue conflates them.

1. **The environment.** `uv` is not installed on the staging machine. No PATH search finds a binary that does not exist. This half is not a code bug — see the maintainer check at the bottom.
[...]

Issue #5701, opened 2026-08-23
mcp registry connections facade: connect() is per-config, the other seven read the process global


senamakel:
## Summary

In `src/openhuman/mcp/registry/mod.rs`, the `connections` facade is split down
the middle: **`connect` resolves the service per-config via `host::for_config`,
while every other function reads the process-global `host::try_service()`.** A
host that connects through the per-config path can therefore never see, use, or
drop the connection it just made — `disconnect` answers `false`, `tools_for`
answers `None`, and `is_connected` answers `false`, all of which read as honest
"nothing is connected" rather than as the wrong-service lookup they are.

## The asymmetry

```
$ awk '/^pub mod connections/,/^}/' src/openhuman/mcp/registry/mod.rs \
    | grep -n "pub async fn\|try_service\|for_config"

10:    pub async fn connected_overview()      11:  host::try_service()
18:    pub async fn all_connected_tools()     19:  host::try_service()
27:    pub async fn tools_for(server_id)      28:  host::try_service()
36:    pub async fn is_connected(server_id)   37:  host::try_service()
50:    pub async fn auth_hint_for(server_id)  52:  host::try_service()
68:    pub async fn connect(config, server)   72:  host::for_config(config)   <-- the odd one
93:    pub async fn disconnect(server_id)     94:  host::try_service()
101:   pub async fn last_error_for(server_id) 102: host::try_service()
```

One of eight takes a config. The connection map lives on the service, so
`connect(config, …)` populates the map belonging to `for_config(config)` and the
other seven look in whichever service `host::init` happened to install as the
global — a different object, or none at all.

## Failure mode

`try_service()` returns `Option`, and all seven callers degrade an absent
service to a benign-looking value:

```rust
pub async fn disconnect(server_id: &str) -> bool {
    match host::try_service() {
        Some(service) => service.dynamic().connections().disconnect(server_id).await,
        None => false,        // <-- indistinguishable from "there was nothing to drop"
    }
}
```

A host that never calls `host::init` (because it is not the desktop app — it
holds its own per-config service) gets `false`/`None`/`Vec::new()` from all
seven, forever, with nothing logged. The docs for `disconnect` even describe the
`None` arm as truthful — *"Answers `false` when the service is not up, which is
the truth: nothing was holding a connection to drop"* — which is exactly right
for a host with no service and exactly wrong for one whose service is simply not
the global.

## Repro

[...]

PR #5772, opened 2026-08-25
chore(vendor): bump tinymcp to main — probe outcomes, missing-runtime terminal error

M3gA-Mind:
## Summary

- Bumps `vendor/tinymcp` from `b88cf60` to `5909bd1` (tinymcp `main`), carrying the probe-outcome fix, the missing-runtime terminal error, and a correction to the first of those.
- **Not a branch switch, despite appearances.** The old pin sat on `mcp-redirect-downgrade`; that branch has since been merged into `main`, so this is a fast-forward and nothing is lost — evidence below.
- Adapts three `mcp_registry_e2e` assertions to the new `ProbeOutcome` return type. The bump does not compile without it.
- One commit in the range **partially reverses another** in the same range. Called out below rather than left in the diff.

## The branch question, settled first

`vendor/tinymcp` was pinned at `b88cf60`, which is the tip of `mcp-redirect-downgrade`, not `main`. That is worth checking before moving, because a branch tip can carry commits `main` never got.

It does not here:

```
compare/main...b88cf60  →  status=behind  ahead_by=0  behind_by=8  total_commits=0
compare/b88cf60...main  →  status=ahead   ahead_by=8  behind_by=0
```

`ahead_by=0` is the answer: the old pin has **nothing** `main` lacks. It is a strict ancestor.

The branch was **merged, not abandoned** — tinymcp#3, *"fix(http): refuse HTTPS→HTTP redirect downgrades"* by @senamakel, merged 2026-08-25T08:06, 7 commits, tip `b88cf60`. So that work is already in this tree today and stays in it after the bump. Nothing by another contributor is newly introduced or removed by this PR; the four substantive commits below are all fleet-authored.

## What rides along

Eight commits, four substantive and four merges:

| commit | what it is |
|---|---|
| `68eeda85` | `fix(supervisor): report what a probe observed instead of asserting a drop` — **tinymcp#5** |
| `1a535591` | `docs(supervisor): stop a public doc comment linking a private constant` |
| `9fad68b6` | `fix(error): make a missing local runtime a terminal variant, not a malformed response` — **tinymcp#7** |
| `b44b958d` | `fix(supervisor): keep the 8s probe default; the mitigation for widening it does not reach the host` |

### `b44b958d` is a behaviour change against `68eeda85`, and it is right

This is the one not to discover from the diff. tinymcp#5 widened the default `probe_timeout` from 8s to 30s, reasoning that a server answering inside a real request's budget is usable by definition, and paired it with `MissedTickBehavior::Delay` so a slow cycle could not become a burst of catch-up ticks.

That mitigation lives in `Supervisor::run` — **which this host does not call.** `src/openhuman/mcp/registry/mod.rs:313-341` builds its own `tokio::time::interval_at`, sets no missed-tick behaviour (so tokio's default `Burst` applies), and calls `Supervisor::tick` directly, once per open workspace per tick. openhuman would therefore have inherited the widened window with none of the protection, on a loop that iterates per workspace.

I verified that against this tree rather than taking the commit message's word for it:

- `grep -rn "Supervisor::run" src/` → no hits; the only driver is the hand-rolled loop above.
- `mcp/registry/mod.rs` → no `set_missed_tick_behavior`.
- `tinymcp .../supervisor/types.rs:132` → `Delay` is set inside `Supervisor::run`, and only there.

So the default returns to 8s. The churn tinymcp#5 was raised for is fixed by the consecutive-timeout run, not by the window width — a host that genuinely needs a longer probe can set `probe_timeout` itself.

## Why a test change is in a bump PR

`Connections::probe_alive` now returns `ProbeOutcome` instead of `bool`. `cargo check --lib --tests` fails on the bump alone with three `E0308`s in `tests/mcp_registry_e2e.rs` (`:553`, `:571`, `:619`). Adapting them via `ProbeOutcome::is_alive()` is what makes the pin land green, not separable work — so it is here, in its own commit, and nothing else is.

Deliberately minimal: the disconnected-server case now returns `ProbeOutcome::Missing` and could assert that specifically, which would be a better test than `!is_alive()`. Tightening assertions is not a bump PR's job; left as a follow-up.

## Verified

- `cargo check --lib --tests` — **clean, exit 0, zero errors.** Default features, which include `mcp` (`Cargo.toml:739`), and `tinymcp` is a non-optional path dependency — so this genuinely compiles the changed code rather than gating it out.
- The one remaining warning (`unused import` in `tests/transcript_search_e2e.rs:23`) is pre-existing on `main`, present in the pre-bump run, and untouched.
- `rustfmt --check` clean on the changed file.
- Only `vendor/tinymcp` and `tests/mcp_registry_e2e.rs` are modified.

Not verified: runtime behaviour. This is a compile-level check plus source review; the probe changes are exercised by tinymcp's own suite, and the supervisor loop is not driven in openhuman's e2e.

## Related

Referenced without closing keywords — cross-repo, and one of them only closes partially.

- **#5636** (MCP transport drops) — **fully addressed** by tinymcp#5 + `b44b958d`. The supervisor now reports what the probe observed instead of asserting a drop, and only tears a session down after a run of consecutive timeouts, so a single slow probe no longer causes the disconnect it then reports. **Safe to hand-close once this merges.**
- **#5600** (`uv` missing) — **partial only, and it should stay open.** tinymcp#7 makes a missing local runtime a terminal `Error::MissingRuntime` rather than a malformed-response error, which stops the retry loop and gives the user an actionable message naming the runtime. It **does not install `uv`**, and does not decide whether openhuman should ship or bootstrap one. That packaging decision is what #5600 is actually about, so this improves the failure mode and leaves the issue's substance untouched.

## Worth knowing

The crate version stays `0.3.1` across the whole range, despite a new public enum (`ProbeOutcome`), a new `Error` variant, and a changed `probe_alive` signature. That keeps the `TINYMCP` record in `src/openhuman/modules/registry.rs` self-consistent — it pins `version: "0.3.1"` with per-platform archives and sha256s, and none of that needs to move here.

It does mean the published `v0.3.1` module artifacts no longer correspond to `main`'s source. Not a live problem: openhuman takes `tinymcp` as a plain path dependency with `default-features = false`, which switches the TinyBus module adapter off, and nothing loads that module record. Flagging it because a reader could reasonably assume the pinned artifacts match the pinned source, and after this bump they do not.

`Error` is `#[non_exhaustive]`, so the added variant cannot break a downstream match.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP server liveness detection for live, disconnected, and reconnected connection states.
  * Updated connectivity checks to report probe results more accurately.

### DIFF
--- a/tests/mcp_registry_e2e.rs
+++ b/tests/mcp_registry_e2e.rs
@@ -553,7 +553,8 @@ async fn probe_alive_reflects_transport_liveness() {
         h.dynamic()
             .connections()
             .probe_alive(&server.server_id, std::time::Duration::from_secs(8))
-            .await,
+            .await
+            .is_alive(),
         "a live stub answers the tools/list probe"
     );
 
@@ -571,7 +572,8 @@ async fn probe_alive_reflects_transport_liveness() {
         !h.dynamic()
             .connections()
             .probe_alive(&server.server_id, std::time::Duration::from_secs(8))
-            .await,
+            .await
+            .is_alive(),
         "a disconnected server is not alive"
     );
 }
@@ -615,12 +617,12 @@ async fn supervisor_reconnects_a_dropped_server() {
             .await,
         "supervisor reconnects a dropped-but-installed server"
     );
-    assert!(
-        h.dynamic()
-            .connections()
-            .probe_alive(&server.server_id, std::time::Duration::from_secs(8))
-            .await
-    );
+    assert!(h
+        .dynamic()
+        .connections()
+        .probe_alive(&server.server_id, std::time::Duration::from_secs(8))
+        .await
+        .is_alive());
 
     h.dynamic()
         .connections()


M3gA-Mind:
**CI status: red, and it is inherited from `main` — not caused by this PR.**

`Rust Core Coverage` fails on `parallel_subagent_fanout`:

```
thread 'parallel_subagent_fanout' panicked at tests/agent_harness_e2e.rs:1979:5:
synthesis must contain both canaries; full_response: PARALLEL_ALPHA_CANARY
```

The same job fails identically on `main` itself, at the commit this branch is cut from:

| commit | `Rust Core Coverage` |
|---|---|
| `e4dd1e14` — merge #5757 | **success** |
| `1111bdfe` — merge #5768 (`chore(deps): bump vendor/tinyagents to e0f3210`) | **failure** |
| `210543ab` — this PR, branched off `1111bdfe` | **failure** |

Byte-identical failure on `main` and here: same file and line (`agent_harness_e2e.rs:1979`), same message, same missing canary — `ALPHA` present, `BETA` absent. So the regression window is exactly `e4dd1e14..1111bdfe`, which is a single PR: **#5768**, a one-file `vendor/tinyagents` submodule bump.

This PR touches `vendor/tinymcp` and `tests/mcp_registry_e2e.rs`. There is no path from the MCP client to sub-agent fan-out synthesis, and the failure predates the branch.

**On the cause, stating only what I established.** #5768 carried `bbcd0a69..e0f32106` — 8 commits, including tinyagents#123/#124 by @sanil-23 (a per-model-call wall-clock ceiling, HTTP/2 PING keepalives on the default provider client) and tinyagents#122. The per-call ceiling was my first suspect, since a fan-out branch losing its result is exactly what a too-tight deadline looks like — **but I ruled it out**: `RunLimits::max_model_call_ms` defaults to `None` (`limits/types.rs:179`), so it is opt-in and off unless a host sets it, and openhuman cannot be setting a field that did not exist before this bump. I have not identified which of the remaining commits is responsible and am not going to guess.

Fixing that is outside a tinymcp bump and belongs with the tinyagents owner, so I have not attempted it here. Flagged to the fleet manager. This PR is otherwise complete: both reviewers approved, no review threads, 14 other checks green, and `cargo check --lib --tests` clean locally.

— fleet CI/review pass

M3gA-Mind:
**Update — the red is fixed on `main` by #5776, and my attribution above was wrong.**

Two corrections to my previous comment.

**It was not #5768.** I read `e4dd1e14` green / `1111bdfe` red and concluded the window was the tinyagents bump. That inference does not hold: `scripts/ci/rust-coverage-changed.sh` is changed-modules-scoped, so a gitlink-only bump runs the full suite while #5757's `src/openhuman/agent/**` changes selected a subset that never ran `--test agent_harness_e2e`. The green I was reading is a test that did not run. #5768 was simply the first commit whose file-shape forced a full run. Retracted on that PR, since I had named specific commits there.

**The real cause**, per #5776: #5757 retired `spawn_parallel_agents` (deliberately, pinned in `loader.rs:809-821`), two test sites were still scripted against it, and the rejected call shifted the scripted FIFO so `PARALLEL_ALPHA_CANARY` became the final response. #5776 re-anchors those tests on the async spawn surface and is merged to `main` as `be6e7f6a`.

What stands from that comment: this PR did not cause the failure, and there is no path from the MCP client to sub-agent fan-out synthesis. What does not stand is my attribution of it.

Next step here: once `main`'s run on `be6e7f6a` confirms green, I will merge `main` into this branch (merge, not rebase, to keep review history) and re-verify — including re-checking that the merge has not disturbed the `ProbeOutcome` adaptation in `tests/mcp_registry_e2e.rs`, which is the only openhuman-side change this PR carries.

— fleet CI/review pass

M3gA-Mind:
**Synced with `main` (`f79eeb03b`, clean merge, no conflicts). CI is still red — on a different, and again pre-existing, test.**

```
agent_harness_leftovers_raw_coverage_e2e::subagent_prompt_renderer_handles_formats_caps_and_stale_tool_indices
  panicked at tests/raw_coverage/agent_harness_leftovers_raw_coverage_e2e.rs:693:5
  assertion failed: json_prompt.contains("## Output style")
```

**Not this PR, and this time I checked before saying so.** `grep -rn "## Output style" src/` returns **zero** — the section does not exist anywhere in the tree. The test asserts a prompt heading that was removed when the style text moved into `STYLE.md`, which `prompts/builder.rs:23-27` dates to **#5701**. My diff is `vendor/tinymcp` and `tests/mcp_registry_e2e.rs`; there is no path from the MCP client to subagent prompt rendering.

**Why `main` is green and this is not — the same mechanism as before, worth stating because it will keep happening.** `main`'s run at `be6e7f6a` was *scoped*: its log shows only

```
tests/agent_harness_e2e.rs → --test agent_harness_e2e
tests/orchestrator_parallel_fanout_routing.rs → --test orchestrator_parallel_fanout_routing
```

`raw_coverage_all` ran **zero** times there. A gitlink-only change falls to the `*)` arm of `rust-coverage-changed.sh` and runs the **full** suite, so a vendor bump is currently the only PR shape that executes these targets. That makes bumps the canary for everyone else's orphaned tests — this is the second one this PR has surfaced today, after `parallel_subagent_fanout` (fixed in #5776).

**The fix, for whoever owns it.** One assertion. `GLOBAL_STYLE_SUFFIX = include_str!("STYLE.md")` (`builder.rs:28`) is appended to every assembled prompt, and `STYLE.md:1` is `# Writing style`. So line 693 should re-anchor to the heading the builder actually emits, or be dropped if style is deliberately no longer asserted in the subagent prompt. The four assertions around it still pass — `## Safety` at line 692 is still emitted (`sections.rs:409`).

I have not made that change here. It is a prompt-design call in someone else's area, it is unrelated to a vendor bump, and #5776 has just done exactly this family of work. Referred to the fleet manager rather than folded in silently.

**Status of this PR itself:** merge with `main` was clean, my `ProbeOutcome` adaptation is intact (`git diff origin/main` is still exactly the 2 files, +11/−9), `cargo check --lib --tests` is clean locally, both reviewers approved, no unresolved threads. It is blocked only on the orphaned assertion above.

— fleet CI/review pass

Issue #5727, opened 2026-08-24
No check ties a module's registry pin to its submodule pin — the drift that caused #5598, #5623 and #5641 is still unguarded (and tinymcp is drifted now)


M3gA-Mind:
## Summary

Nine subsystems now load as downloaded cdylib modules, each pinned **twice** and independently: once as a git submodule (the source this repo compiles the wire contract against) and once as a `version` + per-platform SHA-256 in `src/openhuman/modules/registry.rs` (the artifact actually loaded at runtime).

**No CI job and no test compares the two.** They can drift apart in either direction, and when they do the failure is a runtime capability mismatch on a user's machine — not a build error, not a red lane.

This has already produced at least three closed issues (#5598, #5623, #5641). Each was fixed by bumping a pin; none added a check, so the next one is not prevented.

## The two pins

`src/openhuman/modules/registry.rs:750-760` — `ALL`, nine records:

| Record | `version` in registry.rs | submodule pin |
|---|---|---|
| `tinydocs` | `0.1.14` (`:38`) | `v0.1.14` |
| `tinywallet` | `0.5.0` (`:136`) | `v0.5.0` |
| `tinymemory` | `1.2.0` (`:205`) | `v1.2.0` |
| `tinyjuice` | `0.2.4` (`:283`) | `v0.2.4` |
| `tinyvoice` | `0.1.5` (`:364`) | `v0.1.5` |
| `tinyruntime` | `0.2.2` (`:446`) | `v0.2.2` |
| `tinyruntime-nodejs` | `0.2.2` (`:527`) | `v0.2.2` |
| `tinyruntime-python` | `0.2.2` (`:600`) | `v0.2.2` |
| **`tinymcp`** | **`0.3.1`** (`:687`) | **`v0.3.1-7-gb88cf60`** |

Eight agree. `tinymcp` does not: the submodule is **seven commits ahead** of the tag whose artifact the registry would download.

```console
$ git submodule status vendor/tinymcp
 b88cf60a05683d88e63bab5150927cdd01b6e03e vendor/tinymcp (v0.3.1-7-gb88cf60)
```

That particular drift is currently **inert**, and only by luck: `tinymcp` is linked as a plain path dependency with the module adapter switched off (`Cargo.toml:296`, `default-features = false`), so nothing ever loads its artifact. The one crate where the two pins disagree happens to be the one crate that does not use the mechanism. Nothing arranged that.

## What is actually checked today

[...]

Issue #5641, opened 2026-08-20
Learning scheduler: memory driver does not support 'profile family' — rebuild fails every tick


Al629176:
## Summary

The learning scheduler's periodic rebuild fails with a new error: `memory driver does not support the profile family`. This is a different failure mode from the prior `ListAllFacets` issue (fixed in pre-prod). The `tinymemory` driver loads successfully and reports profile-relevant capabilities, but the scheduler rejects it as unsupported for profile-family operations.

## Log evidence

```
01:34:40:WRN:scheduler [learning::scheduler] periodic rebuild failed (non-fatal): memory driver does not support the profile family
```

Memory driver at boot (same session):
```
01:40:02:INF:log [boot] memory driver bound: id=tinymemory class=module capabilities=[core,recall,ingest,documents,tree,entities,graph,diff,goals,tool_memory,sources,maintenance,portability] fallback=None
```

## Root cause

`tinymemory` advertises capabilities including `goals` and `entities` but apparently does not implement the `profile` capability family (user preference/learning profile). The scheduler queries for profile-family support before attempting a rebuild and finds it missing — no `profile` entry in the capability list.

## Impact

- Learning/personalisation rebuilds never run — the agent does not update its user model over time
- All personalisation data from new sessions is discarded at the scheduler tick
- Non-fatal (app continues), but personalisation is silently broken

## Fix direction

- Add `profile` to the `tinymemory` capability set and implement the profile family ops
- Or: make the scheduler degrade gracefully when profile family is unsupported, and suppress the WRN after first log (it fires every tick)

## Acceptance criteria

- [ ] `memory driver does not support the profile family` warning does not appear in normal operation
- [ ] Learning scheduler periodic rebuild completes successfully with tinymemory driver
- [ ] Profile/personalisation data persists across sessions

M3gA-Mind:
Duplicate of 5623

Issue #5635, opened 2026-08-20
Pre-prod: config.toml written with insecure permissions (mode 644, world-readable) — auto-fixed at runtime but not at write time


Al629176:
## Summary

On startup, OpenHuman detects that `config.toml` was created with mode 644 (world-readable) and auto-corrects it to 600. The file should be written with 600 permissions from the start — the auto-fix is a safety net, not a solution, because the file is briefly world-readable between creation and correction.

## Log evidence (pre-prod, 2026-08-20)

```
16:03:05 WRN [config] Config file
"/Users/alan/.openhuman/users/69de186d2c2521812c1936d1/config.toml"
is world-readable (mode 644); auto-fixing to 600
```

## Security impact

`config.toml` contains user configuration including API keys, model routing preferences, and potentially connection identifiers. Being world-readable (even briefly) on a multi-user system exposes this to any local process running as a different user. The auto-fix closes the window, but the write should use 0o600 from creation.

## Fix direction

Use `OpenOptions::new().mode(0o600)` (or equivalent) when initially writing `config.toml`. Remove the startup auto-fix once the write permission is correct at source.

Felyx-Fu:
### Static security triage (2026-08-23)
**Verdict:** needs_review; likely remediated on current main but not release-closed.
**Evidence/gap:** Current main writes the temporary config with 0600 permissions before rename and attempts to repair permissions on load. Verify the fix in the target release artifact and add platform-appropriate permission tests; do not claim every filesystem is covered while unsupported permission operations remain warning-only.

M3gA-Mind:
**How to fix this**

The write-time hardening this issue asks for already exists and has since 2026-07-25 (`9f22ae408`), so implementing the `OpenOptions::mode(0o600)` suggestion would be a no-op. The 644 state is nonetheless real and recurring, because a **second writer** — not `Config::save()` — rewrites `config.toml` afterwards. That writer is identified below; this issue should be re-scoped to it.

**Root cause:** the memory-sources registry persists `memory_sources` by rewriting the *entire* `config.toml` through its own atomic-write path, which creates the temp file with `File::create` (mode `0o666 & ~umask` = 0644) and renames it over the config. The rename carries 0644 onto the live file. It runs on every `memory_sources_list` RPC — not just at boot — so the load-time heal is undone within minutes of every launch, which is why the warning recurs indefinitely.

**Where:**
[...]

Issue #5633, opened 2026-08-20
Pre-prod: Notion sync broken — composio/notion 'Invalid request data provided' on every tick (51x)


Al629176:
## Summary

The Composio/Notion sync pipeline fails on every single tick with `notion provider failure: Invalid request data provided`. All three sync layers (orchestrator, dispatcher, periodic) report the same failure. Notion memory sync is completely non-functional in this build.

## Log evidence (pre-prod, 2026-08-20 — 51 occurrences)

```
16:27:08 WRN [sync:orchestrator] sync failed toolkit="notion" connection_id="ca_nOkTCMfS4Isd"
error=notion provider failure: Invalid request data provided

16:27:08 WRN [memory_sync:dispatcher] pipeline tick failed pipeline_id="composio:notion"
error=notion provider failure: Invalid request data provided

16:27:08 WRN [composio:periodic] sync failed (will retry next tick) toolkit=notion
connection_id=ca_nOkTCMfS4Isd error=notion provider failure: Invalid request data provided
```

17 orchestrator + 17 dispatcher + 17 periodic failures = 51 total. Every retry fails identically.

## Impact

Notion is completely not syncing to memory. Any user who connected Notion as a knowledge source has no Notion context available to their agent in this build.

## Fix direction

- The `Invalid request data provided` error from Notion typically indicates a malformed API request — wrong field names, unsupported filter syntax, or a deprecated API version.
- Check if the Composio Notion connector is using a Notion API version that has been updated/changed.
- Reproduce with connection_id `ca_nOkTCMfS4Isd` against the Notion API directly to see the full error response.

M3gA-Mind:
**How to fix this**

This is not a new bug — it is #5506 re-introduced. `tinyhumansai/tinycortex#144` (merged 2026-08-18) corrected it, and that patch is still present and still right. It just no longer runs: the periodic sync has since moved onto **tinymemory-core's** copy of the same pipeline, which never received it. `fetch_type` has never existed anywhere in the tinymemory repo (`git log -S'fetch_type'` in `vendor/tinymemory` → no commits). The `v1.2.0` re-pin did not change this file beyond the workspace path move.

Composio's full error, from #5506, is `Invalid request data provided - Following fields are missing: {'fetch_type'}` — this report shows the truncated prefix.

**Root cause**
[...]

PR #5773, opened 2026-08-25
chore(memory): bump vendor/tinymemory to v1.6.0 and implement the new contract

M3gA-Mind:
## Summary

- Bumps `vendor/tinymemory` **`d9a8a6ea` (v1.3.0) → `1a6a24ae` (v1.6.0)** — 26 commits, three releases, 53 files.
- Implements the five host trait impls the new contract requires. **The bump and the host code are one commit because they cannot be split**: the host cannot implement a trait method the pinned contract does not declare, so either half alone leaves `main` red.
- Also advances `vendor/tinycortex` **`b7cf1213` → `84c994b7`** — one commit, and a **required companion rather than a second bump**; see below.
- The surface was **larger than the two methods this work was scoped from**: `create_segment` and `append_turn` each gained a parameter and `ChunkQuery` gained six fields. Details below.

### Which commit this pins, and why not an earlier tag

Pinned: **`1a6a24ae6d4733943bb23b3c71e80f5139025c61`**. The `v1.6.0` tag is an annotated tag over that exact commit, so the tag and the sha are the same thing — but **do not substitute an earlier tag**:

| tag | contains tinymemory#91 | contains #92 |
|---|:--:|:--:|
| v1.4.0 | ✗ | ✗ |
| v1.5.0 | ✗ | ✗ |
| **v1.6.0 (pinned)** | ✓ | ✓ |

Both fixes merged at 07:34 on 2026-08-25, *after* v1.5.0 was cut. `git compare` reports each merge commit `ahead` of v1.4.0 and v1.5.0 and `behind` v1.6.0. A reviewer would reasonably assume any tag carries them; only this one does.

## Problem

tinymemory v1.4.0 added two trait methods with no default bodies, so the pin could not move without host code:

```rust
async fn recall_namespace_recent(&self, namespace: &str, limit: usize)
    -> Result<Vec<NamespaceMemoryHit>, MemoryError>;
async fn insert_event(&self, event: &EpisodicEvent) -> Result<(), MemoryError>;
```

## Solution

Five impls (not six — `ModuleMemoryProvider` carries 18 `Memory*` traits and `MemoryEpisodic` is not among them; `Capability::Episodic` is absent from `ARTIFACT_CAPABILITIES`, so `as_episodic()` never hands back the module driver).

**`GuardedRetrieval::recall_namespace_recent`** — `carries_content: false`, no redaction. The flag describes the payload the host *sends*, not the result it gets back; this call sends a namespace name and a count, so there is nothing to redact.

**`GuardedEpisodic::insert_event`** — modelled on `tree.append`, **not** on `insert_turn` beside it. `insert_turn` passes `NO_NAMESPACE` because an `EpisodicTurn` has none; an `EpisodicEvent` does, so it is admitted against `event.namespace` with `carries_content: true` and its `content` and `subject` redacted. The knock-on is included: `create_segment`'s comment calling itself *"the only episodic call that names a namespace"* is no longer true and is updated here.

**`ModuleMemoryProvider::recall_namespace_recent`** — returns `Unsupported` naming the pinned artifact, and is the one method in that file that does not forward. `Ok(vec![])` would be the tempting answer and is wrong: the contract says an empty vector from this method *means* the namespace is empty, so returning it because the binary cannot answer asserts something false no caller can distinguish — the over-claim shape of #5641/#5623 moved into a return value. A bare forward would come back as an untyped `Other` after a round trip (#5598). **This introduces a new pattern in that file**; every other method forwards unconditionally. It collapses to a `module_call!` on the next artifact re-pin.

The message is built from `ARTIFACT_CAPABILITIES_PIN` rather than a literal, so it cannot drift from the pin.

## What rides along — 26 commits, and most of it is not ours

Ours (4): the Notion `fetch_type` fix at **both** call sites (tinymemory#92, `0fd40091` + `7b32ba95`), the source-registry permissions fix (#91, `2409d867`), and a docs follow-up (`ab6736ad`).

**Everything else is #5560 contract work by other contributors, and some of it is behaviour, not plumbing:**

| PR | What it changes |
|---|---|
| #90 recency-recall | Adds `recall_namespace_recent` — the reason this bump needs host code |
| #95 contract-batch-v14 | New `insert_event` member; **per-session sequence threaded through the segment members** (this is what changed `create_segment`/`append_turn` arity); chat attribution; mail headers on the bus |
| #99 close-every-gap | *"Give the host every read it still runs raw SQL for"*; **serves `ComposioHost` and `ConfigLoader` inside the module**; advances the nested tinycortex |
| #100 module-periodic-sync-loops | **Runs the periodic memory sync loops inside the module** |

The last two deserve a reviewer's attention: the periodic memory sync loops and the Composio host seam **move into the module**. That is the same subsystem as the Notion sync fix riding along in this PR, and it is a placement change by another contributor, not a fix of ours.

The nested `vendor/tinycortex` advance inside tinymemory also carries tinycortex **#144** (`fetch_type`) and **#146** (`render_properties`, the openhuman#5500 fix) — the two commits that were stranded when the sync pipelines were ported out of tinycortex. Note they arrive in *tinymemory's* nested copy; openhuman's `[patch]` resolves `tinycortex` from its own `vendor/tinycortex`, so what actually executes is the top-level pin.

## The tinycortex companion, and why it is not scope creep

`cargo check` fails on the tinymemory bump alone:

```
error[E0425]: cannot find function `count_chunks_matching` in module `crate::engine::backend::chunks`
error[E0432]: unresolved imports `...::ChunkDetailRow`, `...::SourceTotal`
```

tinymemory v1.6.0 calls those; tinycortex **#157** adds them. tinymemory's own nested pin already points at `84c994b7`, but openhuman's `[patch]` at `Cargo.toml:1234` forces `tinycortex` to resolve from **openhuman's** `vendor/tinycortex`, which overrides it. So v1.6.0's requirement is only satisfied by moving openhuman's own pin. One commit, no choice about which.

## Larger than the two-method summary

Worth stating plainly, because it was not in the scope this was planned from:

- **`create_segment` gained `start_seq: Option<u32>`** and **`append_turn` gained `seq: Option<u32>`** (the per-session sequence from #95). Threaded through the guard and the recording provider as pass-throughs.
- **`ChunkQuery` gained six filter fields** (`entity_ids`, `entity_kinds`, `content_contains`, and three more). Three call sites now use `..Default::default()`, which defaults them to "no filter" and preserves each site's existing behaviour exactly. **This PR deliberately opts into none of the new filtering.**

## Verification

- `cargo check --lib --tests` — exit 0.
- `cargo fmt --all -- --check` — exit 0.
- `cargo clippy -p openhuman --no-deps -- -D warnings` — exit 0 (the CI-shaped lane).
- Tests across every touched module: `guard::` 145 passed · `modules::memory::` 18 passed · `memory::preferences::` 7 passed · `memory::tree::` 98 passed · `memory::tools::` 43 passed. 0 failed.

**Three revert-checks, each observed:**

1. Drop redaction from `insert_event` → `assertion left != right failed: the credential reached the driver verbatim — redact_outbound was not applied / left: "Authorization: Bearer abcdefghijklmnop"`.
2. `Unsupported` → `Ok(vec![])` in the module provider → `the pinned artifact cannot serve this member: []`.
3. Refuse an untrusted external driver → both new guard methods assert `driver.call_count() == 0`, so a guard that decided after touching the driver fails.

**One thing I could not pin, and a correction to the plan this was built from.** The plan expected `NO_NAMESPACE` vs `&event.namespace` to be caught by "a namespace outside the policy's scope is refused". It is not: `admit_write` passes `namespace` **only to `guard_span`** (`policy.rs:148`), so the two allow and deny identically and differ solely in what the audit span records. I wrote a tracing-capture test that did catch it — reverting to `NO_NAMESPACE` gave `left: Some("-") right: Some("team-notes")` — but it proved **flaky under parallel test execution** (`tracing` caches callsite interest globally; `rebuild_interest_cache()` did not fix it), and a flaky test is worse than an admitted gap, so I removed it. The choice is therefore correct-by-construction for when `admit_write` gains real namespace scoping, and untested until then.

## Impact

Memory subsystem. No RPC or UI surface changes. The `Unsupported` return is new observable behaviour for any caller of `recall_namespace_recent` on the module driver — by design, and preferable to the two silent alternatives.

**Deliberately not in this PR**, per standing decisions:
- **The cdylib artifact is not re-pinned.** `ARTIFACT_CAPABILITIES_PIN` stays `"1.3.0"` and the registry keeps its `tinymemory-module-1.3.0-*` assets. This is why §3's `Unsupported` is needed at all; re-pinning is a separate decision.
- **The pre-existing `carries_content` inconsistency is not fixed.** Two structurally alike pairs are labelled oppositely today — `documents.query_documents` sends a user query with `true` **and** redacts, while `retrieval.recall_namespace_scored` sends one with `false` and does not; `tree.append` writes content with `true` and redacts, while `episodic.insert_turn` writes user conversation with `false` and does not. Latent, because `check_egress` and `redact_outbound` are both no-ops off an `External` driver — and live the first time one binds. `insert_event` follows the `tree.append` side. **Flagged for separate filing, not resolved here.**

## Related

- Refs `#5633` — the Notion `fetch_type` fix reaches openhuman with this bump. **Can be hand-closed after merge**; both call sites are fixed and shipped.
- Refs `#5635` — likewise **can be hand-closed after merge**.
- Refs `#5727` — **cannot be closed by this PR.** This bump is a live instance of it: `ARTIFACT_CAPABILITIES` is per-*family*, and this breaks *inside* one. `Capability::Retrieval` stays claimed and `the_capability_list_matches_the_pinned_release` still passes, while one member of that family is unserved and detectable only at call time. Every existing guard reports green while a real gap opens. `#5500` was already closed as completed on 2026-08-24 (tinycortex #146 plus backend #1281), so it needs nothing here — the tinycortex note above is about which copy executes, not about that issue's status.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6553,7 +6553,7 @@ dependencies = [
 
 [[package]]
 name = "tinymemory"
-version = "1.3.0"
+version = "1.6.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/app/src-tauri/Cargo.lock
+++ b/app/src-tauri/Cargo.lock
@@ -7273,7 +7273,7 @@ dependencies = [
 
 [[package]]
 name = "tinymemory"
-version = "1.3.0"
+version = "1.6.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/src/openhuman/memory/guard/families.rs
+++ b/src/openhuman/memory/guard/families.rs
@@ -39,7 +39,7 @@ use crate::openhuman::memory::api::provider::chunks::{
     ChunkDetail, ChunkEmbedding, ChunkQuery, MemoryChunks,
 };
 use crate::openhuman::memory::api::provider::episodic::{
-    ConversationSegment, EpisodicTurn, MemoryEpisodic,
+    ConversationSegment, EpisodicEvent, EpisodicTurn, MemoryEpisodic,
 };
 use crate::openhuman::memory::api::provider::people::{
     AddressBookSeedOutcome, MemoryPeople, PersonHandle, PersonInteraction, PersonRecord,
@@ -1080,6 +1080,27 @@ impl MemoryRetrieval for GuardedRetrieval {
             .await
     }
 
+    async fn recall_namespace_recent(
+        &self,
+        namespace: &str,
+        limit: usize,
+    ) -> Result<Vec<NamespaceMemoryHit>, MemoryError> {
+        // `carries_content: false` like the scored sibling above, and for a
+        // stronger reason: the flag describes the payload the host *sends*, not
+        // the result it gets back (it only picks `check_egress`'s disclosure
+        // label). This call sends a namespace name and a count — no
+        // user-authored text at all — so there is also nothing to redact.
+        self.policy.admit_read(
+            Capability::Retrieval,
+            "retrieval.recall_namespace_recent",
+            namespace,
+            false,
+        )?;
+        self.family()?
+            .recall_namespace_recent(namespace, limit)
+            .await
+    }
+
     async fn search_entities(
         &self,
         query: &str,
@@ -1113,6 +1134,39 @@ impl MemoryEpisodic for GuardedEpisodic {
         self.family()?.insert_turn(turn).await
     }
 
+    async fn insert_event(&self, event: &EpisodicEvent) -> Result<(), MemoryError> {
+        // Modelled on `tree.append`, not on `insert_turn` above. `insert_turn`
+        // passes `NO_NAMESPACE` because an `EpisodicTurn` has no namespace to
+        // admit against; an `EpisodicEvent` does, and admitting a write that
+        // names a namespace against none would scope it wider than the payload.
+        //
+        // `carries_content: true` and redaction for the same reason `tree.append`
+        // has both: the event's extracted prose is the outbound payload. Both
+        // are no-ops on the Embedded/Module/Null drivers shipping today and only
+        // bite on an External one — which is the case they exist for.
+        self.policy.admit_write(
+            Capability::Episodic,
+            "episodic.insert_event",
+            &event.namespace,
+            true,
+        )?;
+        // The trait takes `&EpisodicEvent`, so redaction needs an owned copy;
+        // `tree.append` takes its payload by value and avoids this.
+        let mut event = event.clone();
+        event.content = self.policy.redact_outbound(&event.content).into_owned();
+        event.subject = event
+            .subject
+            .take()
+            .map(|subject| self.policy.redact_outbound(&subject).into_owned());
+        trace_allowed(
+            &self.policy,
+            "episodic.insert_event",
+            &event.namespace,
+            event.content.chars().count(),
+        );
+        self.family()?.insert_event(&event).await
+    }
+
     async fn session_turns(&self, session_id: &str) -> Result<Vec<EpisodicTurn>, MemoryError> {
         self.policy.admit_read(
             Capability::Episodic,
@@ -1142,11 +1196,14 @@ impl MemoryEpisodic for GuardedEpisodic {
         session_id: &str,
         namespace: &str,
         start_episodic_id: i64,
+        start_seq: Option<u32>,
         start_timestamp: f64,
         now: f64,
     ) -> Result<(), MemoryError> {
-        // The only episodic call that names a namespace, so it is the only one
-        // that can be admitted against it.
+        // One of the two episodic calls that name a namespace, so it can be
+        // admitted against it. `insert_event` is the other; every remaining
+        // episodic method passes `NO_NAMESPACE` because its payload carries
+        // none.
         self.policy.admit_write(
             Capability::Episodic,
             "episodic.create_segment",
@@ -1159,6 +1216,7 @@ impl MemoryEpisodic for GuardedEpisodic {
                 session_id,
                 namespace,
                 start_episodic_id,
+                start_seq,
                 start_timestamp,
                 now,
             )
@@ -1169,6 +1227,7 @@ impl MemoryEpisodic for GuardedEpisodic {
         &self,
         segment_id: &str,
         episodic_id: i64,
+        seq: Option<u32>,
         timestamp: f64,
         now: f64,
     ) -> Result<(), MemoryError> {
@@ -1179,7 +1238,7 @@ impl MemoryEpisodic for GuardedEpisodic {
             false,
         )?;
         self.family()?
-            .append_turn(segment_id, episodic_id, timestamp, now)
+            .append_turn(segment_id, episodic_id, seq, timestamp, now)
             .await
     }
 
--- a/src/openhuman/memory/guard/families_tests.rs
+++ b/src/openhuman/memory/guard/families_tests.rs
@@ -416,3 +416,98 @@ async fn retrieve_leaves_intersects_an_explicit_scope_with_the_ambient_one() {
         "an explicit scope outside the ambient one must fail closed"
     );
 }
+
+// ── The two family members the v1.6.0 contract added ────────────────────────
+
+#[tokio::test]
+async fn recall_namespace_recent_is_admitted_as_a_read() {
+    let (driver, guard) = guarded(embedded_policy());
+    guard
+        .as_retrieval()
+        .unwrap()
+        .recall_namespace_recent("ns", 10)
+        .await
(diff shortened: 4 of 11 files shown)


M3gA-Mind:
**Overlap with #5725, flagged rather than resolved.** Raising it here so whoever reviews either PR sees the other; I have not touched #5725 and am not proposing which should win.

#5725 ("Route the recall, reset and flush paths through the contract", @YellowSnnowmann, currently `CHANGES_REQUESTED`) adds the **same trait methods in the same files**: `guard/families.rs`, `guard/test_support.rs`, `preferences/tests.rs`, `modules/memory.rs`, plus `tree/retrieval/rpc.rs` which this PR does not touch. Whichever lands second takes a conflict across all of them.

More than a conflict, the two disagree on the axis that was deliberately left open:

| | this PR (#5773) | #5725 |
|---|---|---|
| `vendor/tinymemory` + `vendor/tinycortex` gitlinks | **moved** | not moved — its body defers them to "Phase D" |
| `ARTIFACT_CAPABILITIES_PIN` | left at `1.3.0` | **`1.3.0` → `1.6.0`** |
| registry module assets | left at `tinymemory-module-1.3.0-*` | **moved to `1.6.0-*`** (11 assets) |
| `Capability::Episodic` / `impl MemoryEpisodic for ModuleMemoryProvider` | not added (five impls) | **added**, with `as_episodic` wired |
| `ModuleMemoryProvider::recall_namespace_recent` | local `Unsupported` naming the pin | **`module_call!` forward** |

That last row follows from the one above it, and both readings are internally consistent. This PR was scoped **not** to re-pin the artifact, which is exactly why it cannot forward: with the pinned binary at 1.3.0 there is no `RecallNamespaceRecent` member to call. #5725 re-pins to 1.6.0, at which point forwarding is correct and this PR's `Unsupported` branch is dead on arrival — the outcome the plan behind this work predicted for that case.

Two facts that may help whoever decides:

- **tinymemory v1.6.0 does have 11 published module artifacts**, so the re-pin #5725 makes is available, not aspirational.
- **The gitlinks are not optional for either PR.** #5725 adds host impls for methods the pinned contract does not declare, which does not compile — `error[E0407]: method 'recall_namespace_recent' is not a member of trait 'MemoryRetrieval'`, plus `EpisodicEvent` not existing at v1.3.0 at all. I verified that with the compiler before opening this PR; it is why the bump and the impls are one commit here. Its own body reaches the same conclusion about the dual gitlink ("Both gitlinks move in Phase D or neither does").

So the two are closer to halves of one change than to competing versions of it: this PR carries the gitlinks #5725 needs, and #5725 carries the artifact re-pin that makes this PR's `Unsupported` unnecessary. Sequencing them is a maintainer call, not one I should make from here.

M3gA-Mind:
Closing in favour of **#5725** — the maintainer has chosen that approach for the memory-contract work.

No criticism of either; they overlap on the same trait methods in the same files (`guard/families.rs`, `guard/test_support.rs`, `preferences/tests.rs`, `modules/memory.rs`) and only one can land. #5725 is the broader piece and it makes the artifact-pin decision this PR deliberately reserved.

**Three findings from this PR that #5725 should not have to rediscover** — recorded here and on #5725:

1. **Neither `v1.4.0` nor `v1.5.0` contains tinymemory#91 or #92.** Both merged at 07:34 on 2026-08-25, after v1.5.0 was cut; both compare as `ahead` of those tags. The first tag that contains them is **`v1.6.0`**, an annotated tag over `1a6a24ae6`. Any future gitlink move that targets v1.4.0 or v1.5.0 will silently fail to carry the config-permissions and Notion `fetch_type` fixes.
2. **Two Cargo worlds, both resolved with `--locked`.** A gitlink move needs `app/src-tauri/Cargo.lock` regenerated as well as the root one, or `PR CI Gate` fails on the second world. (#5725 already touches both, so this is confirmation rather than a warning.)
3. **The bump and the host impls cannot be split.** The host cannot implement a method the pinned trait does not declare — verified by compiler probe at the current pin:
   `error[E0407]: method recall_namespace_recent is not a member of trait MemoryRetrieval`, and `EpisodicEvent` does not exist there at all. Whichever PR moves the gitlink must carry the impls in the same commit, or `main` goes red either way.

**Consequence worth stating plainly:** #5725 does not move any `vendor/` path, so **openhuman#5633 and #5635 remain blocked after it merges** — their fixes are in tinymemory `main` but not in this repo until a separate bump lands.

Design notes and the per-impl policy reasoning are preserved in `TRAIT-IMPL-PLAN.md`.

Issue #5755, opened 2026-08-24
MCP agent.run_subagent cannot dispatch integrations_agent, leaving MCP-driven brains no route to Composio toolkits


nocstah:
## Summary

The MCP server's `agent.run_subagent` hard-rejects `integrations_agent` ("does not yet support … toolkit binding"), so an MCP-driven brain (e.g. the claude-code provider) has **no route to Composio toolkits at all** — Gmail/Calendar/Notion tasks delegated over the bridge dead-end after wasted tool-call round-trips.

## Problem

**What happened vs expected:** the MCP surface deliberately exposes delegation instead of raw `composio_*` tools (a sound design — the specialists own the prompts and approval contracts). But the one delegate that can reach a toolkit, `integrations_agent`, is explicitly stubbed in `run_subagent_tool` (`src/openhuman/mcp/server/tools/dispatch.rs`):

```
agent.run_subagent does not yet support `integrations_agent`; first-level MCP support is currently limited to standalone agents that do not require toolkit binding
```

Observed consequence with a claude-code brain: the model tries `tools_agent` (composio-blind **by design** — its own docs route integrations to `integrations_agent`), gets "Composio tools are not available", tries `integrations_agent`, gets the hard reject, then burns further turns probing for alternate routes (including shelling toward the core's RPC port, which the Seatbelt jail correctly blocks). Net: an expensive multi-minute failure for a task the in-core orchestrator handles in one delegate call.

**Steps to reproduce:**
1. Claude Code provider as the brain; a connected Composio toolkit (e.g. Gmail, direct mode).
2. Ask for anything toolkit-bound ("list my recent emails").
3. Watch the delegation chain fail: `tools_agent` (no composio tools) → `agent.run_subagent {agent_id: integrations_agent}` → InvalidParams hard-reject.

**Version/platform:** v0.63.x, any platform, MCP server surface.

## Solution (optional)

Design decision for maintainers rather than a drive-by patch — a naive fix (attaching raw composio tools to the bridge's plain `Agent::run_single` path) would create exactly the "second, weaker route to the same capability" the tools_agent docs warn against, bypassing the typed spawn's prompt-context integration, fuzzy tool filter, and scope gating. Options that keep one route:

1. Accept an optional `toolkit` param on `agent.run_subagent` and dispatch through the typed spawn path (`subagent_runner::ops::runner::run_subagent`), establishing the parent-execution context the MCP surface currently lacks.
[...]

PR #5777, opened 2026-08-25
fix(mcp): flag subagents run_subagent will not dispatch (#5755)

ntdatt812:
Refs #5755. Takes the fallback that issue names — *"if support stays out, `agent.list_subagents` could at least flag it as not-dispatchable over MCP"* — and leaves the design question open.

## What is wrong

`agent.list_subagents` enumerates the whole `AgentDefinitionRegistry`, and each entry's `when_to_use` reads as an invitation to delegate. `agent.run_subagent` then refuses one of the entries outright:

```
agent.run_subagent does not yet support `integrations_agent`; first-level MCP
support is currently limited to standalone agents that do not require toolkit binding
```

So the catalogue advertises a delegate that dispatch turns away, and a brain can only discover that from the error — after paying for the round trip, and typically after already trying `tools_agent`, whose own docs route integrations to `integrations_agent`.

## What this changes

One predicate, `mcp_dispatch_block_reason`, is now the single source for both the refusal and what the listing publishes. `run_subagent_tool` returns its reason; each listed definition carries

```json
"dispatchable_over_mcp": false,
"not_dispatchable_reason": "agent.run_subagent does not yet support `integrations_agent`; …"
```

and the human-readable summary line carries the same reason inline, so a model reading the text form sees it too:

```
- **integrations_agent** (not dispatchable over MCP — agent.run_subagent does not yet support …): Use for Gmail, Calendar, Notion
```

Because both sites read the same function, the catalogue cannot drift back into advertising something dispatch will not run: adding an id to the predicate flags it and refuses it in one edit, and removing it does both too.

## What this deliberately does not do

It does not decide whether MCP should eventually dispatch toolkit-bound agents. #5755 lists two designs for that — a `toolkit` param routed through the typed spawn, or exposing `delegate_to_integrations_agent` over MCP — and both change the tool schema, which is a maintainer call, not a drive-by. This holds under either: when support lands, the id leaves the predicate and both the refusal and the flag disappear together.

I also did not touch the naive route the issue warns against (attaching raw composio tools to the bridge's `Agent::run_single`), for the reason the issue gives: it would create a second, weaker path to the same capability.

## Tests

`src/openhuman/mcp/server/tools_tests.rs`, two cases, both pure — no config, no registry, no TTY:

- `integrations_agent_is_flagged_as_not_dispatchable_over_mcp` — the predicate returns a reason, the summary line carries the marker **and the reason itself**, and `when_to_use` survives the marker.
- `dispatchable_subagents_are_listed_without_a_marker` — `tools_agent` and the empty id are unblocked, and the line is byte-identical to the old format.

`subagent_summary_line` was extracted so the marker is assertable without standing up an `AgentDefinitionRegistry`; `list_subagents` now builds each bullet through it, so the tested function is the shipping one rather than a copy.

**Mutation-checked**, two independent mutations, each caught by exactly one test and nothing else:

| mutation | result |
|---|---|
| `mcp_dispatch_block_reason` never blocks (`false && …`) | 59 pass / 1 fail — only `integrations_agent_is_flagged_…` |
| summary line drops the marker, keeps the plain format | 59 pass / 1 fail — only `integrations_agent_is_flagged_…` |

```
cargo test --features mcp --lib mcp::server::tools::
test result: ok. 60 passed; 0 failed
```

`cargo fmt -- --check`: clean.

## Note on the pre-push hook

Pushed with `--no-verify`. The hook runs `cargo clippy -D warnings` over the whole lib, which currently fails on `main` with 11 pre-existing errors — unused imports, an unreachable statement, two needless `mut`s, a needless `return`, a missing `Default`, and a `SetEntriesInAclW` reference — across these files:

```
src/core/all_tests.rs                       src/openhuman/hosted/orchestration/ingest.rs
src/openhuman/agent/learning/schemas.rs     src/openhuman/inference/local/service/ollama_admin/…
src/openhuman/agent/orchestration/…         src/openhuman/inference/provider/factory_tests.rs
src/openhuman/agent/progress_tracing/…      src/openhuman/integrations/composio/…  (3 files)
src/openhuman/agent/prompts/mod_tests.rs    src/openhuman/memory/tree/tree/rpc.rs
src/openhuman/agent/tinyagents/…  (2)       src/openhuman/security/…  (2 files)
src/openhuman/config/…  (2 files)           src/openhuman/runtime/python/…
```

None is in `src/openhuman/mcp/`, and this diff touches only `src/openhuman/mcp/server/tools/`. `grep` for `mcp/server/tools` in the clippy output returns 0 hits. #5762 is the PR that clears that breakage; this one does not depend on it.

## Summary by CodeRabbit

- **New Features**
  - Subagent listings now show whether each agent can be dispatched.
  - Unavailable agents include a clear explanation of why they cannot be used.
  - Structured results now provide dispatchability and refusal details for easier integration.
  - Text summaries preserve usage guidance while clearly marking unavailable agents.

- **Bug Fixes**
  - Subagent execution now consistently rejects unsupported requests with an appropriate explanation.

### DIFF
--- a/src/openhuman/mcp/server/tools/dispatch.rs
+++ b/src/openhuman/mcp/server/tools/dispatch.rs
@@ -243,6 +243,31 @@ async fn core_tool_instructions() -> Result<Value, ToolCallError> {
     ))
 }
 
+/// Why `agent.run_subagent` will refuse `agent_id`, or `None` when it will run it.
+///
+/// One source for the refusal and for what `agent.list_subagents` publishes, so
+/// the catalogue cannot advertise a delegate that dispatch turns away. The list
+/// enumerates the whole registry and each entry's `when_to_use` invites the
+/// model to delegate; before this, a brain reached `integrations_agent` through
+/// that invitation and only learned it was unreachable from the error, after
+/// spending the round trip (#5755).
+pub fn mcp_dispatch_block_reason(agent_id: &str) -> Option<&'static str> {
+    (agent_id == "integrations_agent").then_some(
+        "agent.run_subagent does not yet support `integrations_agent`; first-level MCP support is currently limited to standalone agents that do not require toolkit binding",
+    )
+}
+
+/// One bullet of the `agent.list_subagents` summary.
+///
+/// Pure so the "not dispatchable" marker is asserted without standing up a
+/// config and an agent registry.
+pub fn subagent_summary_line(id: &str, when_to_use: &str) -> String {
+    match mcp_dispatch_block_reason(id) {
+        Some(reason) => format!("- **{id}** (not dispatchable over MCP — {reason}): {when_to_use}"),
+        None => format!("- **{id}**: {when_to_use}"),
+    }
+}
+
 async fn list_subagents() -> Result<Value, ToolCallError> {
     let config = load_config_and_init_registry().await?;
     let registry = AgentDefinitionRegistry::global().ok_or_else(|| {
@@ -263,6 +288,10 @@ async fn list_subagents() -> Result<Value, ToolCallError> {
                 "tool_scope": def.tools,
                 "subagents": def.subagents,
                 "source": def.source,
+                // Advertised alongside the invitation, not discovered from the
+                // error of acting on it (#5755).
+                "dispatchable_over_mcp": mcp_dispatch_block_reason(&def.id).is_none(),
+                "not_dispatchable_reason": mcp_dispatch_block_reason(&def.id),
             })
         })
         .collect::<Vec<_>>();
@@ -275,7 +304,7 @@ async fn list_subagents() -> Result<Value, ToolCallError> {
             .map(|def| {
                 let id = def.get("id").and_then(Value::as_str).unwrap_or("<unknown>");
                 let when = def.get("when_to_use").and_then(Value::as_str).unwrap_or("");
-                format!("- **{id}**: {when}")
+                subagent_summary_line(id, when)
             })
             .collect::<Vec<_>>()
             .join("\n")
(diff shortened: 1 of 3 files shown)


M3gA-Mind:
## Maintainer review — one conflict, and it is a mechanical one

Against current `main` (`fa044d38`) this is `CONFLICTING`, in exactly one file, for a reason that has nothing to do with your change.

### What conflicts and why

`main` commit `dfa90286e "refactor: split oversized test modules"` emptied `src/openhuman/mcp/server/tools_tests.rs` into two sibling files. It is now four lines:

```rust
use super::specs::list_tools_result_for_config;
use super::*;

#[path = "tools_tests_part_01_tests.rs"]
mod part_01_tests;
#[path = "tools_tests_part_02_tests.rs"]
mod part_02_tests;
```

Your branch appended its two new tests to the **old** monolithic `tools_tests.rs`, right after `slug_from_unicode_only_titles_are_unique_and_stable`. Git sees the whole file replaced on one side and appended on the other, so it conflicts across all 953 lines even though the real overlap is nil.

### The resolution

I worked it through locally to be sure it is as simple as it looks, and it is:

1. Take `main`'s version of `tools_tests.rs` verbatim (the four-line module stub above) — `git checkout origin/main -- src/openhuman/mcp/server/tools_tests.rs`.
2. Move your two added tests — `integrations_agent_is_flagged_as_not_dispatchable_over_mcp` and `dispatchable_subagents_are_listed_without_a_marker`, plus the doc comment above them — to the **end of `tools_tests_part_02_tests.rs`**. That is the right home: `slug_from_unicode_only_titles_are_unique_and_stable`, the test yours were appended after, now lives at the end of `part_02`.

No edit to the tests themselves is needed. `part_02_tests.rs` opens with `use super::*;`, which resolves through `tools_tests` to the `tools` module, and your `pub use dispatch::{mcp_dispatch_block_reason, subagent_summary_line};` in `tools/mod.rs` puts both names in that scope.

Your other two files — `tools/dispatch.rs` and `tools/mod.rs` — replay onto current `main` without conflict.

### On the change itself

The design is right, and the part I'd single out is that `mcp_dispatch_block_reason` is the *single* source for both the refusal and the listing marker. That is what stops the catalogue drifting back into advertising a delegate dispatch will turn away — a comment saying "keep these in sync" would not have. Scoping it to `Refs #5755` and leaving the "should MCP dispatch toolkit-bound agents at all" question to the maintainers is also the right call; that one changes the tool schema.

Both bot reviews are APPROVED and there are no unresolved threads. The three `CANCELLED` checks are a superseded run, not a failure — a rebase will produce a clean run.

I have not pushed anything to your branch and I am not approving. Once the conflict is resolved a maintainer will review and merge.

ntdatt812:
Rebased onto today's `main` in `b05dc58b`. Same file-split conflict as the rest of the batch: `main` split `tools_tests.rs` into `tools_tests_part_01_tests.rs` / `part_02`, so the stub is kept and this PR's two tests moved into `part_02` (305 lines, under the gate).

`cargo test -p openhuman --lib --features "$(bash scripts/ci/product-features.sh)" mcp::server::tools` → **60 passed, 0 failed**. Layout gate and `cargo fmt --all` clean.

PR #5790, opened 2026-08-26
fix(claude-code): launch `claude auth login`, not the obsolete `claude login`

### LINKED ISSUE #5710: Claude Code login launcher uses the wrong authentication command
## Summary

The desktop Claude Code sign-in/reconnect helper launches `claude login`. Claude Code CLI 2.1.207 exposes authentication through `claude auth login`; it does not list `login` as a command.

### Environment

- OpenHuman 0.63.12 desktop
- Windows 11 Pro 10.0.26200 (x64)
- Claude Code CLI 2.1.207

### Steps to reproduce

1. Open Settings and select the Claude Code provider.
2. Click the sign-in or reconnect action.
3. Inspect the command constructed by the native terminal launcher.

### Actual behavior

All platform branches launch `claude login` (or its split-argument equivalent). On the tested CLI:

```text
Usage: claude [options] [command] [prompt]
  auth                                  Manage authentication

Usage: claude auth login [options]
  --claudeai       Use Claude subscription (default)
```

Because the top-level command also accepts a positional `[prompt]`, `login` can be interpreted as a prompt instead of explicitly starting the OAuth flow.

Current `main` still constructs the obsolete command on Windows, macOS, and Linux:

https://github.com/tinyhumansai/openhuman/blob/41cbcc36a677c3389a4d6fd495f40648b85763f4/app/src-tauri/src/claude_code.rs#L3-L64

### Expected behavior

The helper should launch `claude auth login --claudeai`, or `claude auth login` where Claude.ai authentication is intentionally accepted as the CLI default.

## Solution

Update the command and all user-facing fallback instructions for each supported terminal/platform. Keeping the platform-specific command construction in small testable helpers would make this easy to regression-test without opening terminals.

## Acceptance criteria

- Windows, macOS, and every Linux terminal fallback invoke the current Claude authentication subcommand.
- The fallback error copy shows the same valid command.
- Command-construction tests cover all platform variants.
- A manual smoke test opens the Claude.ai authentication flow and allows OpenHuman's recheck action to detect the completed login.

ntdatt812:
Closes #5710.

## The defect

Authentication lives under the CLI's `auth` subcommand. There is no top-level `login` command — and because `claude` takes a positional `[prompt]`, `claude login` was never *rejected*: it was read as a prompt and started an ordinary session. The OAuth flow never ran, so the settings card kept reporting the account as signed out no matter how many times the user clicked Sign in.

Verified against a newer CLI than the report used (2.1.221 vs 2.1.207), so this is not a transient:

```
$ claude --help | grep -c '^  login\b'
0

$ claude auth --help
Usage: claude auth [options] [command]
Commands:
  login [options]   Sign in to your Anthropic account
  logout            Log out from your Anthropic account
  status [options]  Show authentication status
```

All three platform branches of `claude_code_login_launch` constructed the obsolete form.

## The fix

One constant, `CLAUDE_LOGIN_ARGV = ["claude", "auth", "login", "--claudeai"]`, feeding three small pure helpers — `windows_launch_args`, `macos_launch_script`, `linux_launch_candidates` — which is the shape #5710 suggested so this can be regression-tested without opening a terminal.

`--claudeai` is the CLI's own default. It is passed explicitly so the launcher does not silently follow a future change of that default into Console/API billing.

## Tests

Eight cases in `#[cfg(test)] mod tests`:

- `login_command_line_is_the_auth_subcommand` — the exact command line;
- `argv_keeps_auth_and_login_as_separate_words` — split-argument terminals hand argv to `execvp`, so `auth` and `login` must be distinct entries, not one `"auth login"`;
- `windows_launcher_reaches_auth_login`, `macos_launcher_reaches_auth_login`, `every_linux_candidate_reaches_auth_login` — each launcher reaches `claude auth login` **and** cannot reconstruct `claude login`;
- `windows_launcher_keeps_the_empty_start_title` — `start` reads a bare first argument as the window title, which would swallow `cmd` and open an empty shell;
- `macos_script_quotes_the_command_for_do_script`;
- `xfce4_terminal_gets_one_string_and_the_others_get_argv` — xfce4-terminal takes the command as a single string; the other four take argv.

Red/green, running the same eight tests against both constants:

```
CLAUDE_LOGIN_ARGV = ["claude", "login"]                        -> 6 failed, 2 passed
CLAUDE_LOGIN_ARGV = ["claude", "auth", "login", "--claudeai"]  -> 8 passed
```

## Also updated

The 14 translated `settings.ai.claudeCode.loginHint` strings and the `config.ts` doc comment name the command to the user; they now name the right one.

## Verification note

`cargo test --lib claude_code` **compiles** here but the test binary will not start on this Windows box — `STATUS_ENTRYPOINT_NOT_FOUND` (0xc0000139) from the Tauri harness, before any test runs. The red/green figures above therefore come from compiling the same helpers and the same test module standalone (`rustc --edition 2021 --test`). `cargo fmt` applied.

I also could not get a clean `git push` past the pre-push hook: `pnpm rust:clippy` fails with 11 errors on this machine, all in Windows-only code (`std::os::windows::process::CommandExt`, `HANDLE`, `SetEntriesInAclW`, `AppContainerBackend`). None of them names `claude_code.rs`, and none is in a file this PR touches. Flagging it rather than leaving you to wonder why CI and my local state might disagree — happy to open a separate issue for the Windows clippy breakage if that is useful.

## Summary by CodeRabbit

* **New Features**
  * Updated Claude Code sign-in to use `claude auth login --claudeai` across Windows, macOS, and Linux.
  * Improved platform-specific authentication launch behavior for a smoother sign-in experience.

* **Bug Fixes**
  * Replaced outdated Claude Code login commands throughout the app.
  * Updated sign-in guidance and documentation in supported languages to reflect the current authentication command.

### DIFF
--- a/app/src-tauri/src/claude_code.rs
+++ b/app/src-tauri/src/claude_code.rs
@@ -1,6 +1,6 @@
 //! Tauri commands for the Claude Code CLI provider.
 //!
-//! Provides a cross-platform "open a terminal and run `claude login`"
+//! Provides a cross-platform "open a terminal and run `claude auth login`"
 //! helper. The CLI's OAuth flow is interactive (it prints a URL and
 //! waits for the user to paste a code), so we can't host it in-app — we
 //! detach into the user's native terminal so they complete login there,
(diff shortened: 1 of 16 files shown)


ntdatt812:
Fixed in \`66fc5f8\` — valid finding, thanks.

The hints named \`claude auth login\` while the launcher runs \`claude auth login --claudeai\`, so a user copying the hint to run it by hand would get a different command than the button does. The Linux fallback error in \`claude_code.rs\` already printed the full form, so the two halves of the same PR disagreed with each other.

All 14 locales updated. The substitution was guarded with a negative lookahead (\`claude auth login(?! --claudeai)\`) so nothing could pick up the flag twice; verified afterwards that each file contains exactly one occurrence of the full command.

ntdatt812:
Fixed both CI failures; both were mine.

**Rust Quality — `dead_code` on the Linux lane.** `windows_launch_args` and `macos_launch_script` are only called from `#[cfg]`-gated arms, so on Linux they compiled into the lib unused and clippy rejected them.

Each helper is now gated on its own target — but as `any(test, target_os = ...)`, not the bare target. Keeping them compiled under `cfg(test)` is the whole reason they were extracted into pure functions: it lets a Linux CI run check the Windows and macOS argv. All 8 tests still pass on a Windows host, so the `test` arm does hold every helper open.

**Frontend Checks — prettier.** The longer zh-CN hint pushed that line past the print width. Reflowed with `prettier --write`; `prettier --check .` is clean across the app, and `cargo fmt --all --check` is clean.

M3gA-Mind:

**Verdict: correct, well-tested, and the bug it fixes is still live on `main`. This is the strongest PR in the batch I looked at.**

## I re-verified the premise independently

You checked against CLI 2.1.221, the issue used 2.1.207. I checked 2.1.258:

```
$ claude --help | grep -cE "^  login\b"
0

$ claude auth --help
Commands:
  login [options]   Sign in to your Anthropic account
  logout            Log out from your Anthropic account
  status [options]  Show authentication status
```

So across three CLI versions spanning the report, your fix and today, there has never been a top-level `login`. Your reading of *why* it failed silently is right too — `claude` takes a positional `[prompt]`, so `claude login` was accepted as a prompt and opened an ordinary session, which is exactly why the settings card kept reporting signed-out with no error to show for it.

`--claudeai` also checks out, including your reason for passing it explicitly:

```
$ claude auth login --help
Options:
  --claudeai       Use Claude subscription (default)
  --console        Use Anthropic Console (API usage billing) instead of Claude subscription
```

Pinning the default so a future flip to Console billing cannot silently follow the user is the right instinct, and worth the one extra argument.

## Still unfixed upstream

`main` today still constructs the broken form in all three branches — `app/src-tauri/src/claude_code.rs` lines 3, 11, 17, 18, 27, 30, 40, 55, 64. Nothing has overtaken this PR.

One small note: `main` already lists `xfce4-terminal` at `:55`, so that part of the Linux candidate list is no longer new — it should merge without fuss, just don't be surprised if that hunk shrinks.

## On the shape of the fix

Hoisting one `CLAUDE_LOGIN_ARGV` constant through three pure helpers (`windows_launch_args` / `macos_launch_script` / `linux_launch_candidates`) is the right move and makes the thing testable without opening a terminal, which is what #5710 asked for. `argv_keeps_auth_and_login_as_separate_words` is the test I'd have asked for if it weren't already there — split-argument terminals hand argv straight to `execvp`, and `"auth login"` as one word would fail in a way none of the other tests would catch.

## State

`MERGEABLE`, all checks green. Two caveats for whoever merges, neither of them yours to fix:

- CI last ran 2026-08-27, before the repo-wide source split of 2026-08-30. The result is stale rather than wrong — your files were not part of that split — but a re-run on a current base would be worth having before merge.
- Separately: `app/src-tauri` does not currently resolve on a clean `main` (its `[patch.crates-io]` still points `tinyagents` at what is now a virtual manifest). That is a pre-existing repo problem, unrelated to this PR, and it is why pushes here need `--no-verify`.

Nothing for you to do. Flagging it so the state is on the record.

PR #5793, opened 2026-08-26
docs(readme): point Windows contributors at the Windows dev script

### LINKED ISSUE #5786: README desktop dev command is misleading on Windows: should point to pnpm dev:app:win
The README "New contributor" section says:

> Use `pnpm dev` for web-only UI work, `pnpm --filter openhuman-app dev:app` for the desktop shell

On native Windows, `pnpm --filter openhuman-app dev:app` actually maps to:

```json
"dev:app": "bash ../scripts/run-dev-macos.sh"
```

That is the macOS entrypoint. A Windows contributor following the README will end up running the macOS dev script.

In `README.md`, update the desktop shell line to be platform-aware, for example:

```markdown
- macOS: `pnpm --filter openhuman-app dev:app`
- Windows: `pnpm dev:app:win`

### PR DESCRIPTION
**Pull request by ntdatt812**
Closes #5786.

## The defect

The "New contributor" step named one command for the desktop shell:

> Use `pnpm dev` for web-only UI work, `pnpm --filter openhuman-app dev:app` for the desktop shell …

`dev:app` is `bash ../scripts/run-dev-macos.sh`. It is not merely named for macOS — it exports `APPLE_SIGNING_IDENTITY`:

```sh
# scripts/run-dev-macos.sh
export OPENHUMAN_DEV_PORT="$dev_port"
export APPLE_SIGNING_IDENTITY="${APPLE_SIGNING_IDENTITY:-OpenHuman Dev Signer}"
```

So a Windows contributor following the README runs the macOS entrypoint. The Windows script already exists beside it:

```json
"dev:app":     "bash ../scripts/run-dev-macos.sh",
"dev:app:win": "\"C:/Program Files/Git/bin/bash.exe\" ../scripts/run-dev-win.sh"
```

`run-dev-win.sh` is 700+ lines of Windows PATH / LLVM / `LIBCLANG_PATH` setup — not something the macOS script degrades into.

## The change

Both commands are now named for the platform they are, in all six READMEs:

> `pnpm --filter openhuman-app dev:app` (macOS) or `pnpm --filter openhuman-app dev:app:win` (Windows)

The `--filter` form is kept for the Windows one so it matches the surrounding style and works from the repo root like its neighbour.

## One deliberate omission

I claim **macOS and Windows only**, not "macOS/Linux". There are exactly two scripts — `run-dev-macos.sh` and `run-dev-win.sh` — and the macOS one signs with an Apple identity, so folding Linux into it would be asserting a path that does not exist. If there is a Linux desktop-dev story I have missed, tell me and I will name it too.

Surrounding prose and every translation are otherwise untouched; the platform names are the same token in all six languages.

## Summary by CodeRabbit

- **Documentation**
  - Updated Windows desktop development instructions to use the simpler `pnpm dev:app:win` command.
  - Clarified platform-specific development commands across English, German, Japanese, Korean, Urdu, and Simplified Chinese guides.
  - Preserved the existing macOS desktop and web-only development commands.
  - Kept validation commands unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ New contributor? Start with [`CONTRIBUTING.md`](./CONTRIBUTING.md) for the fork/
 
 1. Install Git, Node.js 24+, pnpm 10.10.0, Rust 1.93.0 (`rustfmt` + `clippy`), CMake, Ninja, ripgrep, and the platform desktop build prerequisites.
 2. Fork and clone the repo, then run `git submodule update --init --recursive` before `pnpm install` so the vendored Tauri/CEF sources are present.
-3. Use `pnpm dev` for web-only UI work, `pnpm --filter openhuman-app dev:app` (macOS) or `pnpm --filter openhuman-app dev:app:win` (Windows) for the desktop shell, and focused checks such as `pnpm typecheck`, `pnpm format:check`, and `cargo check -p openhuman --lib` before opening a PR.
+3. Use `pnpm dev` for web-only UI work, `pnpm --filter openhuman-app dev:app` (macOS) or `pnpm dev:app:win` (Windows) for the desktop shell, and focused checks such as `pnpm typecheck`, `pnpm format:check`, and `cargo check -p openhuman --lib` before opening a PR.
 
 Deeper docs: [Architecture](https://tinyhumans.gitbook.io/openhuman/developing/architecture) · [Getting Set Up](https://tinyhumans.gitbook.io/openhuman/developing/getting-set-up) · [Cloud Deploy](./gitbooks/features/cloud-deploy.md).
 
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ New contributor? Start with [`CONTRIBUTING.md`](./CONTRIBUTING.md) for the fork/
 
 1. Install Git, Node.js 24+, pnpm 10.10.0, Rust 1.93.0 (`rustfmt` + `clippy`), CMake, Ninja, ripgrep, and the platform desktop build prerequisites.
 2. Fork and clone the repo, then run `git submodule update --init --recursive` before `pnpm install` so the vendored Tauri/CEF sources are present.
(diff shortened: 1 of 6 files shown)


ntdatt812:
Good catch — I dropped an English "or" into all five translations. Fixed in `8bea620`:

| file | was | now |
|---|---|---|
| `docs/README.de.md` | or | **oder** |
| `docs/README.ja-JP.md` | or | **または** |
| `docs/README.ko.md` | or | **또는** |
| `docs/README.ur-pk.md` | or | **یا** |
| `docs/README.zh-CN.md` | or | **或** |

`README.md` keeps "or", being the English one. Verified each file individually rather than trusting the replace — the Urdu line is RTL and the CJK ones have their own spacing around code spans, so I checked the rendered slice in each.

M3gA-Mind:
Marking this as the fix we're taking for #5786.

To be explicit about the ordering, since #5788 also rewrites this line: #5788 is `CONFLICTING` with a failing check and bundles a Windows *script* rewrite alongside the doc change, while this is doc-only, +6/−6, `MERGEABLE` and green. There's no reason to hold a clean doc fix behind a conflicted script PR — I've asked on #5788 to drop its README hunk on rebase, which also removes one of its conflict sources.

@ntdatt812 — **one nit before merge:** this uses `pnpm --filter openhuman-app dev:app:win`, where #5788 uses the shorter `pnpm dev:app:win`. Both are valid (root `package.json:19` defines it), but the short form matches `CONTRIBUTING.md:122` exactly, and the point of the change is that a Windows contributor reading the README and the CONTRIBUTING guide sees the same command. Please switch to the short form across all six files.

M3gA-Mind:
Pushed `908e9757` on your branch — the nit from the review above, nothing else.

`pnpm --filter openhuman-app dev:app:win` → `pnpm dev:app:win` in all six files. The macOS half is left as `pnpm --filter openhuman-app dev:app`, because that is the exact pair `CONTRIBUTING.md:122` already uses:

> Use `pnpm --filter openhuman-app dev:app` on macOS and `pnpm dev:app:win` on native Windows

so a Windows contributor now copies an identical string from either document. Root `package.json:18` defines `dev:app:win` as a passthrough to the same app-level target, so the command is unchanged in what it runs.

Your localization fix in `8bea6207` (oder / または / 또는 / یا / 或) already cleared the CodeRabbit thread — I did not touch that.

M3gA-Mind (approved):

Issue #2360, opened 2026-05-20
Secret key file created with world-readable permissions before chmod (TOCTOU race on Unix)


MootSeeker:
## Summary

The `.secret_key` file (master encryption key for all stored credentials) is created with default umask-derived permissions (typically `0644` / world-readable) before `fs::set_permissions` restricts it to `0600`. Another local user can read the key during this race window.

## Problem

**What happened:** In `src/openhuman/security/secrets.rs` (lines ~244–253), the key file is written using `fs::write()` followed by a separate `fs::set_permissions(..., 0o600)` call:

```rust
fs::write(&self.key_path, hex_encode(&key))
    .context("Failed to write secret key file")?;

// Set restrictive permissions
#[cfg(unix)]
{
    use std::os::unix::fs::PermissionsExt;
    fs::set_permissions(&self.key_path, fs::Permissions::from_mode(0o600))
        .context("Failed to set key file permissions")?;
}
```

`fs::write()` uses `OpenOptions::new().write(true).create(true).truncate(true)` internally — the file inherits the process's umask (commonly `022`), making it `0644` (owner-read-write, world-readable) until the subsequent `set_permissions` call runs.

**What was expected:** The file should be created atomically with restrictive permissions (`0600`), exactly as `src/core/auth.rs:write_token_file` already does correctly:

```rust
#[cfg(unix)]
{
    let mut file = std::fs::OpenOptions::new()
        .write(true)
        .create(true)
        .truncate(true)
        .mode(0o600)   // <- permissions set at creation time
        .open(path)?;
    file.write_all(token.as_bytes())?;
}
```

**Impact:** On multi-user systems (shared servers, university workstations, CI runners with multiple agents under different UIDs), any local user can race the file creation to read the 256-bit ChaCha20-Poly1305 master key. With this key, an attacker can decrypt all stored credentials (API keys, OAuth tokens, etc.) from the config file.

**Steps to reproduce:**
1. On a multi-user Linux/macOS system, set umask to `022` (default).
2. Run OpenHuman for the first time (triggers `.secret_key` creation).
3. A concurrent process monitoring `~/.openhuman/` via `inotifywait` or a tight polling loop can read the key file in the window between `fs::write` and `fs::set_permissions`.

**Severity:** Medium — requires local access and precise timing, but the blast radius (all encrypted credentials) is high.

**Version / platform:** All versions on Unix (Linux, macOS). Windows uses a separate `icacls` path.

## Solution

[...]

PR #5795, opened 2026-08-26
fix(credentials): create the auth profile store owner-only

### LINKED ISSUE #5724: auth-profiles.json (the credential store) is written world-readable (mode 644) — no hardening in the credentials module
## Summary

`auth-profiles.json` — the credential store — is created world-readable (mode `0644`). The mode is not a decision; it is what `fs::write` gives you (`0o666 & ~umask`), carried onto the destination by the `rename` that follows.

Confirmed on a current build, production workspace, macOS:

```
-rw-r--r--  ~/.openhuman/users/local-elvins-macbook-local/auth-profiles.json
-rw-r--r--  ~/.openhuman/users/69de32aba08309ccbf768b59/auth-profiles.json
-rw-r--r--  ~/.openhuman/users/69df979c2c2521812c1aafc0/auth-profiles.json
```

## Where

`AuthProfileStore::write_persisted_locked` (`src/openhuman/security/credentials/profiles.rs:1009`):

- `:1044` — `fs::write(&tmp_path, &json)` creates the temp file at `0o666 & ~umask` → `0644` under the usual `022`.
- `:1060` — `fs::rename(&tmp_path, &self.path)` moves it over the store. **`rename` carries the source file's mode onto the destination**, so the store is `0644` after every save.

`grep -rn "set_permissions\|PermissionsExt\|from_mode" src/openhuman/security/credentials/` returns nothing. There is no hardening anywhere in the credentials module, and no runtime repair either — so unlike #5635 (where the mode is at least corrected on load) this one is never fixed after the fact.

## What is actually exposed, and to whom

Being precise here rather than reaching for a severity label, because the answer differs per install.

**The file is genuinely reachable.** The whole traversal chain on the affected machine is world-executable — `/` `0755`, `/Users` `0755`, `~` `drwxr-xr-x`, `~/.openhuman` `0755`, `~/.openhuman/users/<id>` `0755`. macOS home directories are `0755` by default, so the directory tree does not save the file. Any process running under a different UID on the box can read it.

**What is in it depends on which of three storage paths the install is on** (`profiles.rs:536-548`):

| Path | When | What sits in the world-readable file |
|---|---|---|
| (a) OS keychain | macOS/Windows with keychain available **and** consent granted | **No secrets.** Secret fields are cleared on read and wiped on next save (`:552-556`). Provider names, profile ids, timestamps only. |
| (b) encrypted JSON | keychain unavailable, consent not yet given, or explicitly declined (`:276-289`) | `enc2:` ciphertext of OAuth access/refresh/id tokens and provider API keys |
| (c) plaintext JSON | oldest legacy path, "pass through" (`:542`) | tokens in the clear |

On the machine where this was found — macOS, keychain path (a) — the three files contain **no secret material**. I checked; that is not the finding being reported.

The finding is that **(b) is the normal state on Linux and on any headless/server install without a working keyring**, and on macOS whenever a user declines the prompt — the code logs "Falling back to encrypted JSON for auth profiles" for all three of those cases. There, the file is OAuth token ciphertext at `0644`. Confidentiality then rests entirely on `.secret_key` being `0600`, which it is (`encrypted_store.rs:330-337`). So this is not "tokens are readable"; it is **defence-in-depth collapsing to a single control** on exactly the installs least likely to have a keyring.

Path (c) is legacy and I did not establish how many installs are still on it. UNDETERMINED — would need field data on how old a workspace has to be to still carry plaintext fields.

## Why this is worth fixing even though today's macOS file holds no secrets

The project has already ruled this exact bug class a defect, twice, and fixed it in the two neighbouring files:

- **#2360** (closed) — *"Secret key file created with world-readable permissions before chmod (TOCTOU race on Unix)"*. The fix is `encrypted_store.rs:330-337`: `OpenOptions` with `.mode(0o600)` and `create_new(true)`, with a comment naming the goal — *"to avoid a TOCTOU race where the file is briefly world-readable"* — and pointing at a reference implementation.
- **#5635** (open) — `config.toml` at `0644`. That is **settings**; this is the **credential store**, in the same directory, with no runtime repair.

So `.secret_key` is hardened, `config.toml` is hardened at three sites (`src/openhuman/config/schema/load/impl_load.rs:343`, `:498`, `:674` — including its temp file), and the file holding the credentials is the one that was missed. That is an omission, not a considered exemption.

## Fix

Create the temp file owner-only rather than chmod-ing after, so it is never momentarily world-readable and the hardening cannot fail separately from the create. The reference pattern is already in-tree and is named by `encrypted_store.rs`'s own comment — `write_token_file` (`src/core/auth.rs:461-483`):

```rust
// profiles.rs:1044, replacing fs::write(&tmp_path, &json)
#[cfg(unix)]
{
    use std::io::Write as _;
    use std::os::unix::fs::OpenOptionsExt as _;
    let mut f = fs::OpenOptions::new()
        .write(true).create(true).truncate(true)
        .mode(0o600)
        .open(&tmp_path)?;
    f.write_all(&json)?;
}
#[cfg(not(unix))]
fs::write(&tmp_path, &json)?;
```

Hardening the **temp** file is the whole fix — `rename` at `:1060` then carries `0600` onto the store, and the store is never observable at `0644` even briefly. Keep it inside the existing `retry_with_backoff` closure so the Windows transient-error handling at `:1029-1042` is unchanged.

Two things to get right:

1. **`create_new(true)` is wrong here**, unlike in `encrypted_store.rs`. The temp name is `…tmp.{pid}.{nanos}` (`:1021-1026`) and the orphan-cleanup path at `:1070` can leave one behind after a failed rename, so a collision is possible and refusing would turn a stale file into a persistent save failure. Use `create(true).truncate(true)`.
2. **Existing files are not repaired by this.** A workspace that has already saved once keeps `0644` until the next save rewrites it — and the next save does rewrite it, because the mode travels with the new temp file. So no migration step is needed, but the fix only takes effect on the next credential write.

## How to prove it

Unit test beside the existing persistence tests in `profiles_tests.rs` (which already cover the `fs::rename(tmp -> auth-profiles.json)` staging — see the comment at `:800`), gated `#[cfg(unix)]`:

```rust
let store = /* AuthProfileStore in a TempDir */;
store.save(/* one profile */)?;
let mode = fs::metadata(&path)?.permissions().mode() & 0o777;
assert_eq!(mode, 0o600, "auth profile store must not be group/world readable");
```

Reverting `profiles.rs:1044` to `fs::write` makes it fail with `0644` under the default umask. Worth asserting on a **second** save too, so a future change that reintroduces a plain `fs::write` on an update path is caught as well as on create.

## Not covered by this

- **tinymemory#91 does not cover this.** I checked the diff: it touches only `crates/tinymemory-sources/src/registry.rs` (+tests) and fixes `SourceRegistry::atomic_write` re-widening **`config.toml`** — different repo, different writer, different target file. It is the fix for #5635, not for this. The two share a root cause shape (`create → rename` carrying a umask-derived mode) but no code.
- This does not address path (c) plaintext legacy fields, which are a separate question.
- This does not harden the containing directories (`~/.openhuman/users/<id>` is `0755`). Worth considering separately; `0700` there would be defence in depth for every file in the workspace at once.

## Confidence

**High** on the mechanism and the fix — the two write sites, the absent hardening, the traversal chain and the three storage paths were all read at `main @ e1c332bf0`, and the file modes were observed directly on a live production workspace.

**Deliberately not claiming a severity label.** On a keychain-backed macOS install the exposure today is metadata. On a keyring-less Linux install it is token ciphertext guarded by one `0600` key file. Both are worth closing given the fix is ~8 lines and the pattern is already in the tree; neither is a live credential leak, and calling it one would be wrong.

— fleet research, verified against main @ e1c332bf0

ntdatt812:
Closes #5724.

## The defect

`write_persisted_locked` wrote the credential store with `fs::write`, which creates at `0o666 & ~umask` — `0644` under the usual `022` — and then `fs::rename`d it into place. **`rename` carries the source file's mode onto the destination**, so `auth-profiles.json` was world-readable after every save.

The mode was never a decision. `grep -rn "set_permissions\|PermissionsExt\|from_mode" src/openhuman/security/credentials/` returns nothing: there was no hardening anywhere in the module, and no runtime repair either.

## What that exposes, and where it actually matters

Not on every install, and I would rather be precise than reach for a severity label:

| Storage path | When | What sits in the world-readable file |
|---|---|---|
| OS keychain | macOS/Windows, keychain available and consented | no secrets — provider names, ids, timestamps |
| **encrypted JSON** | **keychain unavailable, declined, or headless** | **OAuth access/refresh tokens as `enc2:` ciphertext** |
| plaintext JSON | oldest legacy path | tokens in the clear |

Row 2 is the normal state on Linux and on any server install without a working keyring. There, confidentiality rested entirely on `.secret_key` being `0600` — defence-in-depth collapsing to a single control on exactly the installs least likely to have a keyring.

## The fix

`write_owner_only()` creates the file with `0o600` **already set** rather than widening then narrowing — the same shape #2360 landed for the secret key in `keyring/encrypted_store.rs:330`, whose comment names the goal: *"to avoid a TOCTOU race where the file is briefly world-readable"*.

It also chmods explicitly after opening. `.mode()` applies **only when the file is created**, so an interrupted save that left a `0644` tmp behind would have it reused as-is by the next save — and `rename` would carry that mode straight through to the store. That is the failure this fix would otherwise still have.

Windows keeps `fs::write`: there is no mode to set, and the file inherits the ACL of the per-user profile directory.

## Tests

Three cases in a `#[cfg(all(test, unix))]` module:

- `a_new_credential_file_is_owner_only` — mode is `0600`, not `0644`;
- `a_leftover_world_readable_tmp_is_repaired` — seeds a `0644` tmp, asserts the precondition, then asserts the write narrows it. This is the case `.mode()` alone does not cover;
- `the_contents_are_written_and_truncated` — a shorter payload must not leave stale bytes behind, since `OpenOptions` replaces `fs::write`'s implicit truncate.

## Verification note

`cargo check -p openhuman --lib` → **exit 0**, no diagnostic in `profiles.rs`; `cargo fmt` applied.

I could **not run the tests**: they are `unix`-gated and this is a Windows box, where that module does not even compile in. So they are reviewed-by-eye here and will first execute on CI. If you would rather I verify them on Linux before you spend review time, say so and I will find a way to.

## Not done here

Existing installs stay `0644` until their next save, since this fixes the write path rather than repairing on load. #5635 does repair `config.toml`'s mode on load; adding the same for the credential store is a reasonable follow-up but a different change, and I did not want to widen this one without asking.

## Summary by CodeRabbit

* **Security**
  * Credential-store files are now restricted to owner-only access, helping protect stored authentication data from other local users.
  * Temporary credential files also receive secure permissions and are corrected if previously created with broader access.

* **Bug Fixes**
  * Rewriting shorter credential data no longer leaves stale content in the file.

* **Tests**
  * Added coverage for secure permissions, temporary-file handling, and content truncation across supported platforms.

### DIFF
--- a/src/openhuman/security/credentials/profiles.rs
+++ b/src/openhuman/security/credentials/profiles.rs
@@ -252,6 +252,43 @@ pub struct AuthProfilesStore {
     #[cfg(test)]
     force_lock_unwritable: Arc<AtomicBool>,
 }
+/// Write `bytes` to `path` with the file readable only by its owner.
+///
+/// `fs::write` creates at `0o666 & ~umask` — `0644` under the usual `022` — and
+/// `fs::rename` carries the source mode onto the destination, so the credential
+/// store inherited it on every save. Any process running under a different UID
+/// could read it, and on the encrypted-JSON path — the normal state on Linux and
+/// on any headless install without a keyring — that file is OAuth token
+/// ciphertext, leaving `.secret_key` as the only remaining control.
+///
+/// Same shape as the fix #2360 landed for the secret key in
+/// `keyring/encrypted_store.rs`: create with the mode already set rather than
+/// widening then narrowing, so the file is never briefly world-readable.
+///
+/// `.mode()` only applies when the file is *created*, so a leftover tmp from an
+/// interrupted save would keep its old mode — hence the explicit
+/// `set_permissions` afterwards.
+fn write_owner_only(path: &Path, bytes: &[u8]) -> std::io::Result<()> {
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::{OpenOptionsExt as _, PermissionsExt as _};
+
+        let mut file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(0o600)
+            .open(path)?;
+        file.set_permissions(std::fs::Permissions::from_mode(0o600))?;
+        file.write_all(bytes)
+    }
+    #[cfg(not(unix))]
+    {
+        // Windows has no mode to set here; the file inherits the ACL of the
+        // per-user profile directory it is created in.
+        fs::write(path, bytes)
+    }
+}
 include!("profiles_impl_01_part_01.rs");
 include!("profiles_impl_01_part_02.rs");
 include!("profiles_impl_01_part_03.rs");
--- a/src/openhuman/security/credentials/profiles_impl_01_part_02.rs
+++ b/src/openhuman/security/credentials/profiles_impl_01_part_02.rs
@@ -178,7 +178,7 @@ impl AuthProfilesStore {
             PERSIST_RETRY_BASE_MS,
             || {
                 self.consume_test_transient_failure_write()?;
-                fs::write(&tmp_path, &json).context("write auth profile tmp")
+                write_owner_only(&tmp_path, &json).context("write auth profile tmp")
             },
         )
         .with_context(|| {
--- a/src/openhuman/security/credentials/profiles_owner_only_tests.rs
+++ b/src/openhuman/security/credentials/profiles_owner_only_tests.rs
@@ -1,0 +1,99 @@
+// `super` is the `tests` module, whose own `use super::*` re-exports the
+// private items of `profiles` -- including `write_owner_only`.
+use super::*;
+
+#[cfg(unix)]
+fn mode_of(path: &std::path::Path) -> u32 {
+    use std::os::unix::fs::PermissionsExt as _;
+    std::fs::metadata(path)
+        .expect("metadata")
+        .permissions()
+        .mode()
+        & 0o777
+}
+
+/// The property this change is actually about, asserted on the store rather
+/// than on the helper: after a real save, the credential file on disk is not
+/// readable by anyone but its owner.
+///
+/// The helper tests below cannot see the wiring — they call `write_owner_only`
+/// directly, so swapping the call site back to `fs::write` leaves them green.
+/// This one goes through `upsert_profile`, which writes the tmp and renames it
+/// over the store, so it covers the mode *and* the fact that `rename` carries
+/// it across.
+///
+/// Compiled on every platform so the store API usage is type-checked
+/// everywhere; only the assertion is unix-gated, since Windows has no mode to
+/// look at.
+#[test]
+fn a_saved_store_file_is_owner_only() {
+    let tmp = TempDir::new().unwrap();
+    let store = AuthProfilesStore::new(tmp.path(), false);
+
+    let profile = AuthProfile::new_oauth(
+        "openai-codex",
+        "default",
+        TokenSet {
+            access_token: "access-123".into(),
+            refresh_token: Some("refresh-123".into()),
+            id_token: None,
+            expires_at: Some(Utc::now() + chrono::Duration::hours(1)),
+            token_type: Some("Bearer".into()),
+            scope: None,
+        },
+    );
+    store.upsert_profile(profile, true).unwrap();
+
+    let path = store.path();
+    assert!(path.exists(), "the save must have produced a store file");
+
+    #[cfg(unix)]
+    assert_eq!(
+        mode_of(&path),
+        0o600,
+        "the credential store must not be group/world readable"
+    );
+}
+
+#[cfg(unix)]
+#[test]
+fn a_new_credential_file_is_owner_only() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let path = dir.path().join("auth-profiles.json.tmp");
+    write_owner_only(&path, b"{}").expect("write");
+    assert_eq!(
+        mode_of(&path),
+        0o600,
+        "credential store must not be group/world readable"
+    );
+}
+
+#[cfg(unix)]
+#[test]
+fn a_leftover_world_readable_tmp_is_repaired() {
+    // `.mode()` only applies at creation, so an interrupted save that left a
+    // 0644 tmp behind would otherwise keep it and carry it through `rename`.
+    use std::os::unix::fs::PermissionsExt as _;
+
+    let dir = tempfile::tempdir().expect("tempdir");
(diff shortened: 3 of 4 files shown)


M3gA-Mind:
Maintainer review (no changes pushed — review only).

**Assessment: this looks ready.** `MERGEABLE`, every required check green, no unresolved threads. I read the change rather than just the check marks; the reasoning holds up:

- `write_owner_only` creates with `.mode(0o600)` rather than widening-then-narrowing, so there is no window where the file is world-readable — the same shape #2360 used for `.secret_key`.
- The explicit `set_permissions` after `open` is the part that actually matters and is easy to miss: `.mode()` applies only on creation, so an interrupted save that left a `0644` tmp behind would otherwise have it reused as-is, and `rename` would carry that mode onto the store. `a_leftover_world_readable_tmp_is_repaired` pins exactly that.
- `a_saved_store_file_is_owner_only` goes through `upsert_profile` rather than calling the helper directly, so it covers the wiring *and* the fact that `rename` preserves the mode. Swapping the call site back to `fs::write` fails this test — which is what makes the other three more than decoration.
- `OpenOptions` replaces `fs::write`'s implicit truncate, and `the_contents_are_written_and_truncated` covers the stale-bytes case that would otherwise be a silent corruption.

I confirmed the premise against `main`: `grep` for `set_permissions|PermissionsExt|from_mode` under `src/openhuman/security/credentials/` returns nothing, so there really was no hardening in that module and no runtime repair.

**One adjacent thing for a follow-up, explicitly not for this PR.** The same pattern is still live one file over:

```
src/openhuman/security/credentials/http_creds.rs:376:        fs::write(&tmp_path, &json)
```

That is another credential tmp written at `0644` and renamed into place. It is out of scope here and I would rather this PR land as-is than grow — but it is worth a separate issue so the fix does not stop half-done.

`mergeStateStatus` is `BLOCKED`, which is just the two-approval requirement, not a CI problem. I am not approving — that is a maintainer call and I have deliberately left it to them.

M3gA-Mind (approved):

Issue #5802, opened 2026-08-26
Coding-session import reports a failure banner for an import that succeeds


M3gA-Mind:
## Summary

Brain → Sources → "Import all sessions" shows a red failure banner while the import is still running, and the import then completes successfully seconds later. The user is told the work failed and invited to retry a 35-56s job that already worked.

## Problem

**Observed**, from the running app:

```
ingest coding sessions: call to `IngestCodingSessions` timed out after 30000ms
```

The log shows the work completing successfully *after* the UI gave up:

```
02:13:12  entry   mode="incremental" max_sessions=5
02:13:42  ERROR   IngestCodingSessions timed out after 30000ms   <- caller abandons at exactly 30s
02:13:47  exit    files_seen=422 sessions_processed=3 sessions_failed=0
                  evidence_units=4106 observations=15 budget_hit=true
```

Twice, on two different runs:

| attempt | duration | outcome |
|---|---|---|
| 01:55:36 → 01:56:32 | **56s** | 5 sessions, 4120 evidence units, 42 observations, `sessions_failed=0` |
| 02:13:12 → 02:13:47 | **35s** | 3 sessions, 4106 evidence units, 15 observations, `sessions_failed=0` |

**Expected:** the banner reflects what happened. A run that imported 3 sessions is not a failure.

**Steps to reproduce:** open Brain → Sources with a non-trivial Claude Code / Codex history (422 session files here) and click "Import all sessions". The import takes longer than 30s and the banner turns red while it is still working.

**Impact:** beyond the wrong message, the retry it invites re-runs a job that already succeeded.

## Solution

Two independent defects, both confirmed by reading the code.

**1. A deadline nobody chose.** Three deadlines bound this call:

| layer | value (max_sessions=5) | where |
|---|---|---|
| client fetch abort | 585s | `app/src/services/memorySourcesService.ts` (`120 + N*90 + 15` grace) |
| core `ingest_budget` | 570s | `src/openhuman/memory/sources/rpc.rs:180-196`, applied at `:236` |
| **tinybus call** | **30s** | `vendor/tinybus/crates/tinybus/src/connection.rs:56` `DEFAULT_TIMEOUT`, applied by `Proxy::new` at `proxy.rs:59` |

[...]

PR #5808, opened 2026-08-26
fix(memory): bridge the four defaulted module members instead of refusing

### LINKED ISSUE #5801: Manual memory-source Sync button refuses with 'unsupported capability: source_sync' while scheduled sync works: #5725 routed it onto a contract member the host bridge never implemented
## Summary

**The manual "Sync now" / "All in" button is broken. Scheduled sync works.** Every manual trigger fails with `unsupported capability: source_sync`; the periodic scheduler ingests normally at the same time, in the same session.

The cause is a half-completed contract routing. #5725 (part of the #5560 programme) moved the manual trigger onto `MemorySourceSync::run_source_sync`, but the host's module bridge never implemented that member. It carries a **default body that unconditionally refuses**, so the newly routed call is answered by the default instead of reaching the module. The scheduler was untouched because it runs *inside* the module and never crosses the bridge.

The capability list is correct and the pinned artifact serves the method. The refusal is raised inside the host and never reaches the wire.

## Evidence that only the manual path is affected

Sync History, minutes before the failing manual attempts:

| when | source | items | duration | result |
|---|---|---|---|---|
| 13m ago | Gmail `ca_ml7K8z4dTbIB` | 250 | 1m 24s | ✓ |
| 14m ago | GitHub `ca_tnftuGIRcr_I` | 979 | 1m 17s | ✓ |

1,229 items ingested. Per-integration health agrees: github 1,220 chunks 15 min ago, document 28 chunks 16 min ago.

Every failure is on one path — `apply_all_in_rpc` appears 9 times in the log, `memory_sources_apply_all_in` twice:

```
01:56:21 INF [memory_sources] apply_all_in_rpc: entry
01:56:21 WRN [memory_sources] apply_all_in_rpc: sync trigger failed for source
         source_id=src_365910b24cb54c589714a4ceca3af373 error=unsupported capability: source_sync
01:56:21 WRN ... src_57997d7db4264944a67958e6b076f303 error=unsupported capability: source_sync
01:56:21 WRN ... src_21d1428cbc0f49e49ac38d0104175f29 error=unsupported capability: source_sync
01:56:21 INF [memory_sources] apply_all_in_rpc: complete sources=3 sync_triggered=0
01:56:37 ERR rpc.invoke_method failed: unsupported capability: source_sync
```

Gmail, GitHub and a local folder all fail, so it is not provider-specific — it is path-specific.

### Working path — scheduled sync, entirely inside the module

The schedulers are started by the module at load and never involve the host's `MemorySourceSync` trait:

```
01:36:15 INF tinybus_module tinymemory_core::sync::composio::periodic
             message=[composio:periodic] scheduler starting tick_seconds=1200
01:36:15 INF tinybus_module tinymemory_core::sync::workspace::periodic
             message=[memory_sync:workspace:periodic] scheduler starting tick_seconds=1200
```

The ingest itself is logged 746 times from `tinybus_module tinymemory_core::sync::pipelines::host` — the `tinybus_module` prefix is module-side logging. Inside the module the process-global memory client is bound (tinymemory#100), so the pipeline runs.

**This path never calls `as_source_sync()` and never crosses the host bridge.**

### Failing path — manual trigger, routed onto the contract

`src/openhuman/memory/sources/rpc.rs:1036-1055`:

```rust
let binding = crate::openhuman::memory::binding::for_config(&config)?;
let sync = binding.provider().as_source_sync().ok_or_else(|| { ... })?;   // returns Some
...
sync.run_source_sync(&source.id).await                                     // -> trait default -> Unsupported
```

Also `rpc.rs:574` for the single-source button.

`as_source_sync()` correctly returns `Some`: `ARTIFACT_CAPABILITIES` contains `Capability::SourceSync` and `artifact_serves` agrees. The refusal happens one call later.

`binding.provider()` returns the **unguarded** provider (`binding.rs:93`, the same field as `unguarded_provider`), so the guard is not in this path — `ModuleMemoryProvider` alone is responsible.

## Root cause

`MemorySourceSync` (`vendor/tinymemory/crates/tinymemory-api/src/provider/sync.rs`) has thirteen members. Ten are required. **Three carry a default body that unconditionally refuses:**

| member | default at | host `ModuleMemoryProvider` | host `GuardedSourceSync` | embedded tinycortex engine |
|---|---|---|---|---|
| `run_source_sync` | `sync.rs:153` | **missing** | **missing** | implemented |
| `bootstrap_connection` | `sync.rs:193` | **missing** | **missing** | implemented |
| `is_toolkit_syncable` | `sync.rs:227` | **missing** | **missing** | implemented |

```rust
async fn run_source_sync(&self, source_id: &str) -> Result<SyncRunOutcome, MemoryError> {
    let _ = source_id;
    Err(MemoryError::unsupported(Capability::SourceSync))
}
```

`ModuleMemoryProvider`'s `impl MemorySourceSync` (`src/openhuman/modules/memory.rs:1085`) implements exactly the ten required members and silently inherits all three defaults. **The three members the host drops are precisely the three that have defaults** — a required member would have failed the build.

### The refusal is host-side, not module-side

`sync.rs:153-156`, compiled into the host binary. `tinymemory-bus/src/error.rs:76` is only the `#[error(...)]` Display attribute, not a construction site; `memory/guard/families.rs:113` is the accessor-returned-`None` path and did not fire.

Timing corroborates: three sources refused in **13ms** total (`memory_sources_apply_all_in -> ok (13ms)`), while a real module round-trip in the same session took 30s to time out (`IngestCodingSessions timed out after 30000ms`). No wire call was made.

### The artifact is fine

The pinned v1.12.0 module declares all three in its `METHODS` list (`tinymemory-module/src/lib.rs:774`: `"RunConnectionSync", "RunSourceSync", "BootstrapConnection", "IsToolkitSyncable", ...`), the wire names are canonical bus constants (`tinymemory-bus/src/names.rs:296`, `:439`), and the module implements the handler (`tinymemory-module/src/service/mod.rs:1748`).

The host never references them — `grep -rn "RunSourceSync" src/` returns nothing. There is no `module_call!` arm for any of the three. **Rebuilding or re-pinning the module will not fix this**; the host never sends the call.

## Provenance: a #5560 artefact

`git log -S run_source_sync -- src/openhuman/memory/sources/rpc.rs` gives exactly one commit: **`14a23b994` "Pin v1.10.0 and route memory-source sync through the driver"**, shipped in **PR #5725** (`feat/5560-queue-and-recall-through-the-contract`).

Its own message describes the problem it set out to fix — after the host's in-process engine was deleted, `sources::sync::sync_source` reached `engine::run_source_pipeline`, whose non-Composio branch opens with `global::client_if_ready()`, and with no host engine that slot is empty:

> With no host engine that slot is empty, so "sync now" and "sync everything" failed for EVERY source kind — folders, repositories, RSS feeds, web pages — with "memory client is not ready" [...] tinymemory#104 adds `MemorySourceSync::run_source_sync`, which the driver answers by resolving the id against the registry it already reads and running the same pipeline.

So the manual button was already broken before #5725, with a different error. **#5725 was the intended fix and is incomplete**: it moved the call onto the contract and moved the pin so the module serves it, but never added the host-side `module_call!` arm. The result swapped `memory client is not ready` for `unsupported capability: source_sync`.

This is the same "only all-on or all-off is self-consistent" property seen in `json_rpc_e2e`: the manual half was routed onto the contract while the scheduled half was left inside the module, and the two halves now disagree.

## Blast radius beyond the reported symptom

All three dropped members have live host call sites, all dead through the module bridge:

- `run_source_sync` — `memory/sources/rpc.rs:574` (per-source "Sync now") and `rpc.rs:1055` (Sources → All in). **The reported symptom.**
- `is_toolkit_syncable` — `memory/sync/composio/bus.rs:145`. The host cannot determine whether a toolkit is syncable.
- `bootstrap_connection` — `memory/sync/composio/bus.rs:814`. First-time bootstrap of a newly authorised connection silently fails.

`GuardedSourceSync` has the identical hole in all three. It is not in the failing path, so it is latent rather than a second cause — but a fix that patches only `ModuleMemoryProvider` leaves it for any caller routed through `binding.guard()`.

## Recommendation

Two options were considered and I think **neither is right**:

- **Re-pin or extend the artifact so it serves `source_sync`** — rests on a false premise. The artifact already serves all three members and has since v1.10.0. Nothing is missing there.
- **Route the manual trigger the way the scheduler goes** — collapses into the same thing. The scheduler is a timer *inside* the module; there is no host-side seam to reuse. Asking the module to run a sync on demand *is* `RunSourceSync`, which is what the trait doc says: *"the periodic loops already live here; these are the on-demand half."*

**The fix is to finish #5725: implement the missing host bridge.** Three `module_call!` arms in `ModuleMemoryProvider` for `RunSourceSync`, `BootstrapConnection` and `IsToolkitSyncable`, plus three forwarding arms in `GuardedSourceSync`. That is small, and it is what the routing commit intended.

Separately, and worth a maintainer decision rather than folding into the bug fix: **remove the default bodies from these three members** so the compiler names every implementor missing one. The defaults are the entire reason this was invisible, and they will hide the next member added the same way. That is a breaking trait change touching every implementor, so it should be its own call.

## Why nothing caught it

1. **A defaulted trait member cannot break an implementor at compile time.** This is the whole mechanism and it will recur.
2. **`is_compatible` has no real call sites** — `grep -rn is_compatible src/` returns two doc comments. Worth closing, but it would **not** have caught this: it compares host and artifact capability sets that here already agree. The gap is host-side.
3. **Nothing in CI exercises the pinned artifact against the host bridge.** A test asserting that every `METHODS` entry the artifact declares has a corresponding `module_call!` in `ModuleMemoryProvider` would have caught all three statically.

`OPENHUMAN_MEMORY_MODULE_ASSUME_FULL_CAPABILITIES=1` is **not** a workaround: it only widens `artifact_serves`, which already returns `true`.

## Environment

- Build `v0.63.17`, `main @ a68a69624`
- `tinymemory` 1.12.0, pinned release, macos-26-arm64; `ARTIFACT_CAPABILITIES_PIN = "1.12.0"`
- `[subsystems.memory] driver = "tinycortex"`, `[modules] enabled = true`
- `OPENHUMAN_MEMORY_MODULE_ASSUME_FULL_CAPABILITIES` unset
- Sources: Gmail `ca_ml7K8z4dTbIB`, GitHub `ca_tnftuGIRcr_I`, one local folder

## Separate findings, not part of this bug

Large volume of memory-tree ingest failures with SQLite corruption, from the module's own (working) sync pipeline — a distinct problem:

```
01:58:56 WRN [memory_sync] tree ingest failed; skill store remains authoritative
  error=memory-tree ingest failed for source `gmail:ca_ml7K8z4dTbIB:gmail:...`:
  database disk image is malformed
```

Also logged: `memory.config_loader failed: config loader answered from the module's load-time snapshot: this module re-reads no config file, so a settings change made after it loaded ...`

M3gA-Mind:
## Summary

- Implements the four `module_call!` arms the host's module bridge was missing, so they dispatch to the module instead of being answered by a refusing trait default.
- Fixes the manual "Sync now" / "All in" button, which failed for every source with `unsupported capability: source_sync` while scheduled sync worked normally.
- Found **four** members with this defect, not the one the symptom showed: `run_source_sync`, `bootstrap_connection`, `is_toolkit_syncable` and `diagnose`.
- Adds a test that asserts the dispatch itself, with no module required, and revert-checked each of the four independently.
- **Does not touch `ARTIFACT_CAPABILITIES` or `assume_full_capabilities`.** The capability list was already correct; changing it would have papered over the real gap.

## Problem

`MemorySourceSync` and `MemoryMaintenance` carry members with **default bodies that unconditionally refuse**:

```rust
// tinymemory-api/src/provider/sync.rs:153
async fn run_source_sync(&self, source_id: &str) -> Result<SyncRunOutcome, MemoryError> {
    let _ = source_id;
    Err(MemoryError::unsupported(Capability::SourceSync))
}
```

`ModuleMemoryProvider` implemented every **required** member and silently inherited the defaults. A defaulted trait member cannot break an implementor at compile time, so when these were added to the contract the bridge kept compiling and began refusing at runtime.

The refusal is raised inside the host and never reaches the wire. In the reported session three sources refused in **13ms** total, while a genuine module round-trip in the same session took 30s to time out.

The artifact was never the problem. All four are declared in the module's `METHODS` list (`tinymemory-module/src/lib.rs:770-776`) and their wire names are canonical bus constants (`tinymemory-bus/src/names.rs:287-300`). Before this change:

```
$ grep -rn "RunSourceSync" src/
(no matches)
```

## Solution

Four `module_call!` arms following the pattern the neighbouring members already use, placed to mirror the trait's own member order:

| member | wire name | trait default was at |
|---|---|---|
| `MemorySourceSync::run_source_sync` | `RunSourceSync` | `sync.rs:153` |
| `MemorySourceSync::bootstrap_connection` | `BootstrapConnection` | `sync.rs:193` |
| `MemorySourceSync::is_toolkit_syncable` | `IsToolkitSyncable` | `sync.rs:227` |
| `MemoryMaintenance::diagnose` | `Diagnose` | `records.rs:470` |

Plus the `Diagnosis` import `diagnose` needs. That is the whole production change.

### How the siblings were found

Rather than guess, I diffed every trait member carrying a refusing default against what the bridge implements. **17 members across 8 traits have such a default; 13 are implemented; these 4 were not.** The other three defaulted `MemorySourceSync` members and `MemoryMaintenance::diagnose` were the complete set — no others need a decision.

### Blast radius beyond the reported button

All four have live host call sites, all dead before this change:

- `run_source_sync` — `memory/sources/rpc.rs:574` (per-source Sync) and `rpc.rs:1055` (Sources → All in). The reported symptom.
- `is_toolkit_syncable` — `memory/sync/composio/bus.rs:137`. The host could not determine whether a toolkit was syncable.
- `bootstrap_connection` — `memory/sync/composio/bus.rs:811`. First-time bootstrap of a newly authorised connection failed silently.
- `diagnose` — the typed maintenance diagnosis an operator or agent reads.

### The test asserts dispatch, not a live module

With the module host disabled, `proxy()` fails with `MemoryError::Other` (`memory.rs:342-357`), while an unbridged member returns `MemoryError::Unsupported`. Asserting **"not `Unsupported`"** therefore asserts that the call went through `module_call!` at all — which is precisely the thing that was missing — and needs no module, no network and no artifact.

### Revert-check, one member at a time

A single combined revert would stop at the first assertion and prove only one member. Each of the four was reverted **individually** and the failure names that member:

| reverted | failure |
|---|---|
| `run_source_sync` | `run_source_sync answered from the trait default ... Got Unsupported { capability: "source_sync" }` |
| `bootstrap_connection` | `bootstrap_connection answered from the trait default ... Got Unsupported { capability: "source_sync" }` |
| `is_toolkit_syncable` | `is_toolkit_syncable answered from the trait default ... Got Unsupported { capability: "source_sync" }` |
| `diagnose` | `diagnose answered from the trait default ... Got Unsupported { capability: "maintenance" }` |

`diagnose` reporting `maintenance` rather than `source_sync` is the check that the test discriminates per member rather than catching one capability generically. The `run_source_sync` failure reproduces the production error string exactly.

With the fix in place all four pass, and `openhuman::modules::` is green at 79 passed / 0 failed. `cargo fmt --check` and `cargo clippy --no-deps --lib` are clean.

### The trap: no lane can catch this end to end

**CI cannot prove this fixed.** The lanes run with no module or with a test artifact, which is exactly why the bug reached a user. The test above proves the dispatch exists; it cannot prove the module answers it. **The only end-to-end proof is the real app, and the maintainer can confirm it by clicking Sync** on Brain → Sources and seeing a non-zero `sync_triggered` in `[memory_sources] apply_all_in_rpc: complete`.

## Impact

Restores the manual sync trigger, Composio toolkit-syncability checks, first-time connection bootstrap, and the typed maintenance diagnosis. No behaviour change for anything already working: scheduled sync runs inside the module and never crossed this bridge.

## Left for a separate decision, deliberately not in this PR

**1. `GuardedSourceSync` and `GuardedMaintenance` have the identical hole** in all four members. It is **latent, not reachable**: every call site resolves through `binding.provider()`, which returns the *unguarded* provider (`binding.rs:93`, the same field as `unguarded_provider`), so no current caller reaches it.

I did not fix it here because the guard's forwards are not mechanical — each one must choose `admit_write` or `admit_read`, and that is a security-tier decision, not a transcription. Getting it wrong lets a `readonly` operator trigger a sync that spends money. The current behaviour fails **closed** (denies), so leaving it is safe; adding a forward with the wrong tier would fail **open**. My reading, for whoever takes it: `run_source_sync` and `bootstrap_connection` are writes (both ingest or persist), `is_toolkit_syncable` and `diagnose` are reads.

**2. The defaults themselves are the root cause and will hide the next member.** The durable fix is to remove the default bodies from these members so the compiler names every implementor that is missing one. That is a breaking trait change in `tinymemory-api` touching every implementor, so it belongs in its own change with its own review.

**3. Nothing in CI compares the artifact's `METHODS` list against the host's `module_call!` arms.** A test doing that would have caught all four statically. `is_compatible` has no call sites (`grep -rn is_compatible src/` returns two doc comments), but note it would **not** have caught this: it compares host and artifact capability sets that here already agreed.

## Related

- Closes #5801
- Root cause introduced alongside #5725 (`feat/5560-queue-and-recall-through-the-contract`), which routed the manual trigger onto the contract and moved the pin so the module serves it, but did not add the host-side arm.

## Submission Checklist

- [x] Tests added: `the_defaulted_members_dispatch_to_the_module_instead_of_refusing` covers all four members. The failure path is the per-member revert-check table above.
- [x] N/A: diff coverage. The changed lines are four `module_call!` arms plus one import; all four are executed by the new test, which fails if any is removed.
- [x] No new external network dependencies introduced. The test runs with the module host disabled and dials nothing.
- [x] Linked issue closed via `Closes #5801`.

## Impact (platform)

Desktop. No migration, no schema change, no config change.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/5801-module-bridge-source-sync`
- Commit SHA: `f9a89b2b8`

### Validation Run

- [x] Focused tests: `cargo test -p openhuman --lib -- openhuman::modules::` — 79 passed, 0 failed. Plus the four individual revert-checks above.
- [x] Rust fmt/check: `cargo fmt -p openhuman -- --check` clean; `cargo clippy -p openhuman --no-deps --lib` clean.
- [x] N/A: Tauri fmt/check — no Tauri code changed.

### Validation Blocked

- `command:` end-to-end verification against the real pinned artifact
- `error:` not blocked by a failure — no lane loads the real module, by design
- `impact:` the dispatch is proven by test; that the module answers it is provable only in the running app. See the trap note above.

### Behavior Changes

- Intended behavior change: four members now reach the module instead of returning `Unsupported`.
- User-visible effect: the manual Sync button works; Composio connection bootstrap and syncability checks work.

### Parity Contract

- Legacy behavior preserved: yes for every already-working path. Scheduled sync is untouched.
- Guard/fallback/dispatch parity checks: the capability gate is unchanged — `as_source_sync()` already returned `Some` and still does. Only the member dispatch changed.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **New Features**
  * Added memory maintenance diagnostics.
  * Added source synchronization capabilities, including running syncs, bootstrapping connections, and checking toolkit compatibility.

* **Bug Fixes**
  * Memory operations now correctly forward requests and return meaningful errors when the module is unavailable.
  * Coding-session ingestion now allows additional time based on the configured ingestion budget.

* **Tests**
  * Added regression coverage for memory diagnostics and source synchronization behavior.

### DIFF
--- a/src/openhuman/modules/memory.rs
+++ b/src/openhuman/modules/memory.rs
@@ -167,13 +167,14 @@ use tinymemory_api::provider::types::{
 };
 use tinymemory_api::provider::{
     AddressBookSeedOutcome, ChunkDetail, ChunkEmbedding, ChunkListRow, ChunkQuery,
-    ConversationSegment, CoverWindowQuery, EntityMatch, EpisodicEvent, EpisodicTurn, FacetType,
-    FastRetrieveQuery, MemoryChunks, MemoryCodingSessions, MemoryCore, MemoryDiff, MemoryDocuments,
-    MemoryEntities, MemoryEpisodic, MemoryGoals, MemoryGraph, MemoryIngest, MemoryMaintenance,
-    MemoryPeople, MemoryPortability, MemoryProfile, MemoryProvider, MemoryRecall, MemoryRetrieval,
-    MemorySourceSink, MemorySourceSync, MemoryToolMemory, MemoryTree, PersonHandle,
-    PersonInteraction, PersonRecord, PersonScore, ProfileFacet, RankedPerson, ResolvedPerson,
-    RetrievalHit, RetrievalResponse, SourceRetrievalQuery, SourceTotal, UserState,
+    ConversationSegment, CoverWindowQuery, Diagnosis, EntityMatch, EpisodicEvent, EpisodicTurn,
+    FacetType, FastRetrieveQuery, MemoryChunks, MemoryCodingSessions, MemoryCore, MemoryDiff,
+    MemoryDocuments, MemoryEntities, MemoryEpisodic, MemoryGoals, MemoryGraph, MemoryIngest,
+    MemoryMaintenance, MemoryPeople, MemoryPortability, MemoryProfile, MemoryProvider,
+    MemoryRecall, MemoryRetrieval, MemorySourceSink, MemorySourceSync, MemoryToolMemory,
+    MemoryTree, PersonHandle, PersonInteraction, PersonRecord, PersonScore, ProfileFacet,
+    RankedPerson, ResolvedPerson, RetrievalHit, RetrievalResponse, SourceRetrievalQuery,
+    SourceTotal, UserState,
 };
 use tinymemory_api::recall::OwnedRecallOpts;
 use tinymemory_api::tool_memory::ToolMemoryRule;
@@ -1079,6 +1080,9 @@ impl MemoryMaintenance for ModuleMemoryProvider {
     async fn purge_all(&self) -> Result<PurgeOutcome, MemoryError> {
         module_call!(self, "purge_all", "PurgeAll", ())
     }
+    async fn diagnose(&self) -> Result<Diagnosis, MemoryError> {
+        module_call!(self, "diagnose", "Diagnose", ())
+    }
 }
 
 #[async_trait]
@@ -1094,6 +1098,24 @@ impl MemorySourceSync for ModuleMemoryProvider {
             "RunConnectionSync",
             (toolkit, connection_id)
         )
+    }
+    async fn run_source_sync(&self, source_id: &str) -> Result<SyncRunOutcome, MemoryError> {
+        module_call!(self, "run_source_sync", "RunSourceSync", (source_id,))
+    }
+    async fn bootstrap_connection(
+        &self,
+        toolkit: &str,
+        connection_id: &str,
+    ) -> Result<(), MemoryError> {
+        module_call!(
+            self,
+            "bootstrap_connection",
+            "BootstrapConnection",
+            (toolkit, connection_id)
+        )
+    }
+    async fn is_toolkit_syncable(&self, toolkit: &str) -> Result<bool, MemoryError> {
+        module_call!(self, "is_toolkit_syncable", "IsToolkitSyncable", (toolkit,))
     }
     async fn source_sync_state(
         &self,
--- a/src/openhuman/modules/memory_tests.rs
+++ b/src/openhuman/modules/memory_tests.rs
@@ -414,3 +414,65 @@ fn the_ingest_bus_grace_is_wide_enough_to_order_the_two_timers() {
          reliably order the RPC's timeout ahead of the bus deadline"
     );
 }
+
+/// Every `MemorySourceSync` and `MemoryMaintenance` member the trait defaults
+/// must be bridged to the module rather than left to inherit the default.
+///
+/// Three of these members carry a default body that returns
+/// `Unsupported(SourceSync)`, and `diagnose` one that returns
+/// `Unsupported(Maintenance)`. A defaulted member cannot break an implementor at
+/// compile time, so when they were added to the contract `ModuleMemoryProvider`
+/// kept compiling and silently began refusing — which is #5801: the manual
+/// "Sync now" button answered `unsupported capability: source_sync` while the
+/// module's own scheduler, which never crosses this bridge, kept syncing fine.
+///
+/// The discriminator needs no module. With the host disabled, `proxy()` fails
+/// with `MemoryError::Other`, so a member that really dispatches through
+/// `module_call!` reports `Other` while one that fell through to the default
+/// reports `Unsupported`. Asserting "not Unsupported" therefore asserts the
+/// dispatch itself, which is the part that was missing.
+#[tokio::test]
+async fn the_defaulted_members_dispatch_to_the_module_instead_of_refusing() {
+    use tinymemory_api::provider::{MemoryMaintenance, MemorySourceSync};
+
+    let mut config = Config::default();
+    config.modules.enabled = false;
+    let provider = ModuleMemoryProvider::new(Arc::new(config));
+
+    let refused = |label: &str, error: MemoryError| {
+        assert!(
+            !matches!(error, MemoryError::Unsupported { .. }),
+            "{label} answered from the trait default instead of dispatching to the \
+             module — the `module_call!` arm is missing, so every caller gets \
+             `unsupported capability` however capable the artifact is. Got {error:?}"
+        );
+    };
+
+    refused(
+        "run_source_sync",
+        provider
+            .run_source_sync("src_whatever")
+            .await
+            .expect_err("a disabled host cannot succeed"),
+    );
+    refused(
+        "bootstrap_connection",
+        provider
+            .bootstrap_connection("gmail", "ca_whatever")
+            .await
+            .expect_err("a disabled host cannot succeed"),
+    );
+    refused(
+        "is_toolkit_syncable",
+        provider
+            .is_toolkit_syncable("gmail")
+            .await
+            .expect_err("a disabled host cannot succeed"),
+    );
+    refused(
+        "diagnose",
+        MemoryMaintenance::diagnose(&provider)
+            .await
+            .expect_err("a disabled host cannot succeed"),
+    );
+}


M3gA-Mind:
Rebased onto `5630b00db`. Conflict was with #5803 and is resolved.

**What conflicted:** only `memory_tests.rs`, and only because both sides appended at the end of the file and shared the trailing `);`/`}`. `memory.rs` auto-merged cleanly — #5803 edits `ingest_coding_sessions` in the `MemoryCodingSessions` impl, my four arms are in `MemoryMaintenance` and `MemorySourceSync`. No semantic overlap. Both sides' tests are kept and all pass: 81 in `openhuman::modules::`, including `the_ingest_bus_deadline_always_outlasts_the_rpc_budget`, `the_ingest_bus_grace_is_wide_enough_to_order_the_two_timers` and `the_defaulted_members_dispatch_to_the_module_instead_of_refusing`. fmt and clippy clean.

---

## But #5803 surfaces a real problem with `run_source_sync`, and I have not fixed it here

#5803 gave `IngestCodingSessions` a derived bus deadline because tinybus' flat 30 s `DEFAULT_TIMEOUT` was abandoning a healthy 35 s import. **`run_source_sync` has the same defect, and worse odds.**

From the app session that produced #5801, the syncs that *succeeded* took:

| source | items | duration |
|---|---|---|
| Gmail `ca_ml7K8z4dTbIB` | 250 | **1m 24s** |
| GitHub `ca_tnftuGIRcr_I` | 979 | **1m 17s** |

Both roughly **3x** the 30 s default. And unlike `ingest_coding_sessions`, source sync has **no RPC-level budget at all** — neither `sync_rpc` (`sources/rpc.rs:563`) nor `apply_all_in_rpc` (`:1023`) wraps the call in `tokio::time::timeout`. The flat 30 s is the only deadline in the system.

So once this PR lands, a real Sync will dispatch to the module, the module will sync correctly, and the caller will be released at 30 s with a timeout error while the work completes at ~80 s. That is #5802's shape exactly, and #5803's own comment says why it cannot be waved off: *"A timeout does not cancel the remote work — tinybus cannot — it stops waiting and frees the caller."*

This is still strictly better than today (the sync currently never runs at all), so it is not a reason to hold this PR. But it means **clicking Sync may still show a failure for a sync that succeeded**, and whoever verifies should expect that rather than read it as this fix not working.

**I did not fix it here on purpose.** The fix needs a budget to derive from, and there is no `ingest_budget` equivalent for source sync — sizing one means deciding how long a user should wait before being told a sync is still running, which is a product call, not a transcription. Inventing a constant tuned to the two durations above would be a heuristic fitted to one session's data.

Recommended shape, mirroring what #5803 established:
1. a `sync_budget()` in `memory::sources::rpc` sized to the work (source kind and/or item count), used as the RPC-level `tokio::time::timeout` so the clean structured message wins;
2. `run_source_sync`'s bus deadline derived as `sync_budget() + a grace`, reusing `INGEST_BUS_GRACE` or a sibling, so the two cannot drift.

`bootstrap_connection` plausibly needs the same treatment (it fetches a profile and registers triggers). `is_toolkit_syncable` and `diagnose` are quick reads and are fine on the default.

Happy to file this as its own issue, or to implement it here if you would rather it ship together — say which.

M3gA-Mind:
@coderabbitai review

PR #5824, opened 2026-08-27
ci: compile raw coverage on source changes

### LINKED ISSUE #5700: Rust Core Coverage lane never compiles raw_coverage_all for src-only changes
## Summary

The Rust Core Coverage lane compiles the `raw_coverage_all` integration target **only when a `tests/raw_coverage/*.rs` file is itself in the diff**. A `src/`-only change that breaks one of those ~76 suites therefore lands on `main` green, and the breakage surfaces later as inherited compile errors in the next unrelated PR that happens to touch a raw-coverage file.

This is not hypothetical — it happened in `3e6f20e29` and cost a full debugging cycle in #5694.

## Problem

**What happened.** `3e6f20e29` ("refactor(memory): bring the host's RPC models home from the engine crate") changed 5 files, all under `src/openhuman/memory/`:

```
src/openhuman/memory/direct_engine_refs_tests.rs
src/openhuman/memory/mod.rs
src/openhuman/memory/rpc_models.rs        (new, 614 lines)
src/openhuman/memory/rpc_models_tests.rs  (new)
src/openhuman/memory/store_golden.rs
```

Zero test files outside `src/`. It merged green.

It broke `tests/raw_coverage/memory_threads_raw_coverage_e2e.rs` with **28 × E0308**, which sat undetected on `main` until #5694 touched a raw-coverage file for unrelated reasons and inherited all 28.

**Why the lane missed it.** In [`scripts/ci/rust-coverage-changed.sh`](../blob/main/scripts/ci/rust-coverage-changed.sh):

- `src/<a>/<b>/….rs` → libtest filter `<a>::<b>` against **`--lib` only**.
- `tests/raw_coverage/*.rs` → `--test raw_coverage_all` (line 281).
- `domain_integration_targets()` (line 116) drags in integration targets that guard a domain. Its **entire table is one arm**:

```bash
domain_integration_targets() {
  case "$1" in
    src/openhuman/memory/*)
      printf '%s\n' memory_golden_fixture_e2e memory_golden_parity_e2e
      ;;
  esac
}
```

`src/openhuman/memory/*` maps to the two golden targets — **not** to `raw_coverage_all`. So even the one domain that *has* a mapping does not reach the raw-coverage suites. Every other domain has no mapping at all.

**Why the two existing safety gates did not fire.** The script header documents both (added in #5593), and neither is aimed at this failure mode:

1. `assert-coverage-presence.sh` — fails when a **changed source file** produced no lcov records. Here the changed `src/openhuman/memory/*` files *were* covered by `--lib` tests, so records existed and the gate passed. It watches source coverage, not test-target compilation.
2. Zero-executed-tests → escalate to full suite. The scoped run executed plenty of tests, so this never triggered.

Neither gate can observe "an integration target that was never compiled would fail to compile."

**The specific trap.** The mechanism is worth naming because it will recur. The commit swapped a re-export for a local module:

```rust
-pub use tinymemory_core::rpc_models::*;
+pub mod rpc_models;
+pub use rpc_models::*;
```

Its own comment reads *"Same glob, same paths, same wire bytes."* That is true for consumers reaching the types through `memory::` — and false for consumers naming `tinymemory_core::rpc_models::` **directly**, which now get a *different Rust type* with an identical name and identical serde shape. `--lib` had no such consumer; `raw_coverage_all` had seven. This is the same `tinymemory-api` / `tinycortex-api` near-identical-types hazard already documented in `AGENTS.md`, reached from the other side.

### Steps to reproduce

1. Branch from `main`.
2. Edit any `src/openhuman/<domain>/` file in a way that breaks a signature used by `tests/raw_coverage/<something>_raw_coverage_e2e.rs` — e.g. re-home a type so a name resolves to a distinct type.
3. Touch no file under `tests/`.
4. Open a PR against `main`.
5. **Observed:** all lanes green; `raw_coverage_all` is never compiled.
6. Confirm locally that it is genuinely broken:
   ```bash
   cargo test --features "$(bash scripts/ci/product-features.sh)" \
     --test raw_coverage_all --no-run
   ```

Platform: CI Lite (`ci-lite.yml`), Linux, PRs targeting `main`.

## Solution (optional)

The cheapest correct fix is a **compile-only** check, not a coverage run — the gap is that the target is never *built*, and building it is a small fraction of running it:

```bash
cargo test --features "$PRODUCT_FEATURES" --test raw_coverage_all --no-run
```

Run it on any PR that changes `src/**` regardless of scoping. That closes the hole without paying for the ~76 suites on every PR, and preserves the fast lane's whole point.

Options considered, for the record:

- **Extend `domain_integration_targets()` to map every `src/openhuman/*` → `raw_coverage_all`.** Correct but expensive: it drags the full raw-coverage run into most PRs, which is close to abandoning the fast lane.
- **`--no-run` compile gate (recommended).** Catches the entire class — every failure of this shape is a *compile* failure, since the lane's blind spot is compilation, not assertion outcomes.
- **Leave it to CI Full.** `main`→`release` does run the full suite, so this is caught eventually — but "eventually" means it is caught by whoever's unrelated PR inherits it, which is exactly the cost incurred here.

Worth noting the same blind spot applies to any `tests/` target absent from `domain_integration_targets`, not only `raw_coverage_all`. The `--no-run` approach generalises; the mapping-table approach does not.

## Acceptance criteria

- [ ] **Repro gone** — a `src/`-only change that breaks a `tests/raw_coverage/*.rs` suite fails CI on its own PR, not on a later unrelated one.
- [ ] **Regression safety** — the guard itself is exercised (e.g. a deliberately broken target is shown to redden the lane).
- [ ] **Cost measured** — added wall-clock on a typical `src/`-only PR is recorded in the PR body; the fast lane must stay fast.
- [ ] **Scope stated** — whether the fix covers only `raw_coverage_all` or every unmapped `tests/` target is decided and written down.
- [ ] **Diff coverage ≥ 80%** on the fix PR.

## Related

- `3e6f20e29` — the commit that landed the breakage green.
- #5694 — the PR that inherited the 28 × E0308 and fixed them (`69c57e086`).
- #5593 — added the two existing "we verified nothing" gates; this is a third case neither covers.
- `scripts/ci/rust-coverage-changed.sh` — `domain_integration_targets()` (L116), raw-coverage mapping (L281).

anisayakmitra-in:
## Summary

- Compile the aggregate `raw_coverage_all` integration target whenever a PR changes `src/**`.
- Keep the existing changed-module coverage lane and its scoped test execution unchanged.
- Fail early when a source-only change makes an unmapped integration target stop compiling.

## Problem

A source-only PR can break `tests/raw_coverage/*.rs` while the changed-files lane runs only filtered `--lib` tests. Because `raw_coverage_all` is selected only when a raw-coverage test file changes, the breakage can remain green until an unrelated PR touches that test tree. This closes #5700.

## Solution

- Track whether the changed-file list contains any `src/**` path, including deleted paths.
- Run `cargo test --features "$PRODUCT_FEATURES" --test raw_coverage_all --no-run` through the repository cancellation-aware wrapper before the scoped coverage run.
- Keep this compile-only: the raw suites are not executed on every source PR, so the fast lane does not become a second full coverage lane.
- Add regression coverage for the exact command and for propagation of a compile failure.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge path).
- [x] **Diff coverage ≥ 80%** — pending CI's merged coverage gate.
- [x] Coverage matrix updated — N/A: CI guard behavior only; no feature rows changed.
- [x] All affected feature IDs from the matrix are listed in the PR description under `## Related` — N/A: no feature IDs affected.
- [x] Manual smoke checklist updated — N/A: CI script only.
- [x] Linked issue closed via `Closes #5700` in the `## Related` section.

## Impact

- Runtime/platform impact: CI only; no shipped binaries or user-facing behavior change.
- Performance: adds one compile-only Cargo invocation for source changes; raw coverage tests remain scoped and are not executed by this guard.
- Security/migration impact: none.

## Related

- Closes #5700
- Follow-up PR(s)/TODOs: CI will provide the added wall-clock measurement for a representative source-only PR.

## Validation

- `bash -n scripts/ci/rust-coverage-changed.sh` — passed.
- `git diff --check` — passed.
- `node --test scripts/__tests__/coverage-runner-status.test.mjs` — attempted locally; the Windows Node/Git Bash environment cannot preserve positional arguments in the repository's Bash extraction harness, and the pre-existing tests fail for that environment. The new test is designed for the Linux CI runner used by this workflow.
- Commit is SSH-signed and verified as `ANISAYAK MITRA <266799942+anisayakmitra-in@users.noreply.github.com>`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Rust coverage validation to detect source changes more reliably.
  - Coverage checks now compile the complete raw-coverage test target before execution.
  - Compilation failures are correctly surfaced with their original status.

- **Tests**
  - Added regression coverage to verify source-change detection, configured feature handling, no-run compilation, and failure propagation.

### DIFF
--- a/scripts/__tests__/coverage-runner-status.test.mjs
+++ b/scripts/__tests__/coverage-runner-status.test.mjs
@@ -245,3 +245,36 @@ test("required-features matching is exact, not substring — a superset name doe
   assert.doesNotMatch(res.output, /RAN-THE-TARGET/, res.output);
   assert.match(res.output, /rc=0/, res.output);
 });
+
+test("source changes compile the aggregate raw coverage target", () => {
+  const res = withRunnerFunctions(
+    ["compile_raw_coverage_target"],
+    [
+      "PRODUCT_FEATURES='voice web3'",
+      "log() { printf '%s\\n' \"$*\"; }",
+      "bash() { printf 'BASH_ARGS'; printf ' <%s>' \"$@\"; printf '\\n'; }",
+    ].join("\n"),
+    "compile_raw_coverage_target",
+  );
+
+  assert.equal(res.status, 0, res.output);
+  assert.match(
+    res.output,
+    /BASH_ARGS <scripts\/ci-cancel-aware\.sh> <cargo> <test> <--features> <voice web3> <--test> <raw_coverage_all> <--no-run>/,
+  );
+});
+
+test("raw coverage compile failures fail the source-change guard", () => {
+  const res = withRunnerFunctions(
+    ["compile_raw_coverage_target"],
+    [
+      "PRODUCT_FEATURES='voice web3'",
+      "log() { :; }",
+      "bash() { return 17; }",
+    ].join("\n"),
+    "compile_raw_coverage_target && exit 1; rc=$?; echo \"rc=${rc}\"",
+  );
+
+  assert.equal(res.status, 0, res.output);
+  assert.match(res.output, /rc=17/);
+});
--- a/scripts/ci/rust-coverage-changed.sh
+++ b/scripts/ci/rust-coverage-changed.sh
@@ -218,6 +218,13 @@ run_integration_target() {
   fi
 }
 
+compile_raw_coverage_target() {
+  log "compiling raw coverage integration target for src/** change"
+  bash scripts/ci-cancel-aware.sh cargo test \
+    --features "${PRODUCT_FEATURES}" \
+    --test raw_coverage_all --no-run
+}
+
 run_full() {
   log "running FULL instrumented suite (reason: $1)"
   llvm_cov clean --workspace
@@ -257,6 +264,8 @@ while IFS= read -r f; do
 done < <(printf '%s\n' "${CHANGED_FILES}" | xargs -n1 printf '%s\n' 2>/dev/null || true)
 log "received ${#files[@]} changed rust file(s)"
 
+src_changed=false
+
 if [ "${#files[@]}" -eq 0 ]; then
   run_full "empty changed-file list — scoping unsafe"
 fi
@@ -267,6 +276,9 @@ fi
 lib_filters_raw=""
 test_targets_raw=""
 for f in "${files[@]}"; do
+  case "${f}" in
+    src/*) src_changed=true ;;
+  esac
   if [ ! -e "${f}" ]; then
     # dorny/paths-filter includes deleted paths. They contain no changed lines
     # to cover and, for tests, no longer correspond to runnable Cargo targets.
@@ -373,6 +385,13 @@ done < <(printf '%s' "${test_targets_raw}" | sort -u)
 
 if [ "${#lib_filters[@]}" -eq 0 ] && [ "${#test_targets[@]}" -eq 0 ]; then
   run_full "no scoped test targets derivable from the change set"
+fi
+
+if [ "${src_changed}" = true ]; then
+  # Scoped lib tests cannot compile integration targets that are not selected by
+  # a domain mapping. Build the aggregate raw-coverage target on every src/**
+  # change so source-only PRs cannot leave a broken integration suite behind.
+  compile_raw_coverage_target
 fi
 
 # Drop artifacts from previous coverage runs so merged profdata only reflects


M3gA-Mind:
Maintainer housekeeping — rebased this onto current \`main\` and force-pushed (\`8947a6c\` → \`d243491\`). No approval implied; flagging what changed so you do not find surprise commits.

**What I did**

1. **Rebased onto \`main\`** (it was \`CONFLICTING\`). Your CI last ran on 2026-08-27, before the repo-wide source split, so the old result was stale regardless.
2. **Resolved the conflict in \`scripts/__tests__/coverage-runner-status.test.mjs\`** by keeping *both* sides. `main` had appended three `required-features` tests in the same place your two new tests land. The file now has 9 tests (7 from `main` + your 2).
3. **Dropped your two fixture commits** — `cc84772` ("publish module policy for raw coverage fixture") and `8947a6c` ("use shared workspace for module fixture"). These were fixing the `set_modules_policy` failure your first CI run hit, and `main` has since fixed the identical problem in `4b567b6a9` ("test(ci): publish the module policy in the raw read_rpc case"). `main`'s version is the one to keep: it is `#[cfg(feature = "modules")]`-gated, so it still compiles under **Rust Feature-Gate Smoke (gates off)**, which the ungated version in `cc84772` would not. `tests/raw_coverage/memory_core_threads_raw_coverage_e2e.rs` is therefore untouched by this PR now.

**Result:** the PR is back to exactly its intended scope — one commit, +52 lines across `scripts/ci/rust-coverage-changed.sh` and its test file. Nothing of your design was changed.

**Verified locally:** `node --test scripts/__tests__/coverage-runner-status.test.mjs` → 9/9 pass; `bash -n scripts/ci/rust-coverage-changed.sh` clean.

Re-running CI on the new head will give the first result against a current base.

M3gA-Mind:
Follow-up to my comment above — and an apology for the process, not the content.

**I pushed to this branch before I should have.** Our policy changed mid-review to *review-only* for community PRs: I comment, you drive the branch. That landed after I had already force-pushed the rebase (`8947a6c` → `d243491`). I am deliberately **not** reverting it, because undoing it would be a second unrequested rewrite of your branch on top of the first. But you should know it happened and that it was not your doing.

**The branch is yours.** If you would rather have your original three commits back, `8947a6c` is still reachable and you can force-push over what I left — no objection from me and nothing to explain. Everything I changed is itemised in the comment above.

I will not push here again. From this point I'm commenting only.

## Review verdict on the change itself

**Sound, and it closes a real hole.** The gap you describe is exactly right: the changed-files lane selects `raw_coverage_all` only when a raw-coverage *test* file changes, so a `src/**`-only PR can break an integration target and stay green until something unrelated touches that test tree. Compile-only (`--no-run`) is the proportionate answer — it catches the breakage without turning the fast lane into a second full coverage run.

Three things I checked rather than assumed:

- `compile_raw_coverage_target` goes through `scripts/ci-cancel-aware.sh` like the other invocations, so it inherits cancellation handling instead of stranding a build.
- `src_changed` is set inside the loop *before* the `[ ! -e "${f}" ]` continue, so a deleted `src/**` path still arms the guard — which matches what you claim in the description and is easy to get wrong.
- The guard runs after the `run_full` bail-outs, so it doesn't double up on a run that is already doing everything.

Post-rebase state: `MERGEABLE`, one commit, +52 lines across `scripts/ci/rust-coverage-changed.sh` and its test file. `node --test scripts/__tests__/coverage-runner-status.test.mjs` passes 9/9 locally (the 7 on `main` plus your 2).

On your validation note about Windows: that is a real limitation of the Bash extraction harness under Git Bash, not something wrong with your test — it runs clean on Linux/macOS, and CI is the runner that matters for it.

The one thing still outstanding is a CI result on a current base. The old red — `Rust Core Coverage` failing in `memory_sync_tree_round21_raw_coverage_e2e::slack_sync_status_rpc_reads_mock_connections_and_persisted_state` — came from a run whose feature list still included `memory-git`, a gate that has since been removed from the repo entirely, so that result tells us nothing about this change either way.

PR #5797, opened 2026-08-26
fix(test): re-anchor style-section assertions to the renamed STYLE.md heading

M3gA-Mind:
## Summary

- Re-anchors three `tests/raw_coverage/` assertions from the old prompt heading `## Output style` to its current name `# Writing style`.
- Fixes `main`, which is red on `assertion failed: json_prompt.contains("## Output style")` in the Rust Core Coverage lane.
- Two of the three were latent: the coverage lane runs *changed modules only*, so they had never been exercised and would have fired the moment their module was touched.
- Test-only change. No production code touched.

## Problem

The prompt style section was renamed from `## Output style` to `# Writing style`. It now lives in `src/openhuman/agent/prompts/STYLE.md`, whose sole heading is `# Writing style`, and is pulled into every rendered prompt as `GLOBAL_STYLE_SUFFIX` (`builder.rs:28`, `include_str!("STYLE.md")`).

`ffcc22dc4` updated the two in-tree callers under `src/` (`agent/prompts/mod_tests.rs:149`, `agent/tinyagents/host/context_composer.rs:396`), but three assertions under `tests/raw_coverage/` were missed:

| file | line | asserted |
|---|---|---|
| `agent_harness_leftovers_raw_coverage_e2e.rs` | 693 | `json_prompt.contains("## Output style")` |
| `agent_prompts_subagent_raw_coverage_e2e.rs` | 394 | `rendered.contains("## Output style")` |
| `inference_agent_raw_coverage_e2e.rs` | 2925 | `built.contains("Output style")` |

Only the first has actually failed so far. `Rust Core Coverage (cargo-llvm-cov)` runs `raw_coverage_all` module-by-module and only for changed modules, so the other two sat green-by-omission. That scoping is why a rename that landed on 2026-08-23 only turned `main` red days later.

## Solution

Each assertion is re-anchored to `# Writing style`, matching the fix `ffcc22dc4` applied to the two `src/` sites, rather than deleted. The section still exists and each test is still checking the same property it always was: that the built prompt carries the global style suffix. The anchor is unambiguous, `STYLE.md` contains exactly one heading and no `## Writing style` exists anywhere.

I also swept every heading assertion in `tests/raw_coverage/` (21 distinct `contains("#...")` strings) against the sources. These three were the only orphans. `## Comments` / `## Description` resolve into the vendored memory readers and `### \`shell\`` / `### USER.md` are built dynamically at runtime, so all are sound.

### Revert-check

[...]

PR #5828, opened 2026-08-27
fix(test): publish the module host policy for the contract-routed chunk reads

M3gA-Mind:
## Summary

- Fixes **red `main`**: `memory_core_threads_raw_coverage_e2e::memory_read_rpc_filters_graphs_scores_reset_and_wipe_seeded_rows` fails in `Rust Core Coverage`, blocking every open PR.
- The test seeds through the in-process engine and reads back through the memory contract. It never published a module host policy, so the module could not load and the read refused.
- **26 added lines in one test file. No production code, and no assertion changed or weakened.**
- **#5808 is NOT the cause.** See below — the merge order makes it look guilty and it is not.

## Problem

```
called `Result::unwrap()` on an `Err` value:
"list_chunk_details: the module host policy was never published, so module
 'tinymemory' cannot be loaded; call modules::memory::set_modules_policy during boot"
```

The test writes its fixtures with `tinymemory_core::store::chunks::store::upsert_chunks` — the engine, in this process — and then asserts over `read_rpc::list_chunks_rpc`. **#5725 routed that read onto the memory contract** (`5828ad9d2`, "Route the chunk, entity, graph, admin and cleanup reads onto the memory contract"), whose own comment records the change: *"the SQL it replaces had no allowlist clause at all"*. `list_chunks_rpc` now goes `binding::for_config` → `provider().as_chunks()` → the tinymemory module.

So the test became **mixed-routing**: engine-routed seed, contract-routed read, one process. That is the same defect class that has now surfaced five times — a workspace split between the two routes — and it is the actual bug here, not a flaky test.

### This is not an ordering or isolation problem

The module's one-workspace-per-process capture makes an ordering bug a reasonable first guess. It is not one, on two independent grounds:

- The sibling test in this module never touches the failing path at all — `grep -cE "read_rpc::|list_chunks|as_chunks"` over its body returns **0** — so it cannot publish, consume, or steal any policy state.
- Run **alone**, the failing test fails identically: `FAILED. 0 passed; 1 failed`, same line, same message.

### #5808 is not the cause, and the merge order is misleading

`5828ad9d2` merged at **2026-08-26T10:09** via #5725. The test has been broken since that moment. It only *ran* tonight because the coverage lane is **changed-modules-scoped**, and tonight's merges (#5803, #5808, #5811) were the first to touch `src/openhuman/modules/`, which is what selected this raw_coverage module.

`git diff --name-only 2d819b3df..04075d537 -- tests/raw_coverage/memory_core_threads_raw_coverage_e2e.rs src/openhuman/memory/read_rpc/` is **empty**: nothing in the failing path changed tonight. Anyone reading the merge order will assume the most recent `modules/memory.rs` change did this. It did not — it only made the lane look at code that was already broken.

**That mechanism is the story worth recording: this is the 5th instance today** of a test whose subject drifted from the code and merged green because the changed-modules-scoped lane never ran it (alongside #5776, #5797, #5818, and the vacuous test CodeRabbit caught on #5812).

## Solution

Publish the module host policy, pinned to the test's own `TempDir`, immediately after that tempdir is created.

Both halves matter. The policy is what lets the module load at all; naming **`tmp`** is what makes the engine-routed seed and the contract-routed read open the *same* store rather than two. It has to be done at that exact point because the module captures one workspace per process at load and `set_modules_policy` is a `OnceLock` whose later calls are silently discarded (`modules/memory.rs:224`).

### Why this works in the lane, and why it is not gated

`rust-core-coverage` downloads the pinned artifact and exports `TINYMEMORY_TEST_MODULE` (`ci-lite.yml:774`, inside the job at `:671`), and `local_override` picks that up for this module id (`modules/ops.rs:320-325`) **regardless of cfg**.

Worth recording, because it is the part that is easy to get wrong: the seam at `binding.rs:389` that reads `TINYMEMORY_TEST_MODULE` directly is `#[cfg(all(feature = "modules", test))]` — *unit* tests only. An integration test under `tests/` links `openhuman_core` built without `cfg(test)` and therefore gets the `from_boot_policy()` arm, which is exactly why the error names a missing policy. The artifact was always reachable; the test just never published a policy to route to it.

It is deliberately **not** gated on the module being present. A gate here would make this test silently skip in the one lane that runs it — the "green because it never ran" failure mode that let it sit broken for 18 hours in the first place. Every workflow that runs `raw_coverage` sets `TINYMEMORY_TEST_MODULE` (`test-reusable.yml` is the only one referencing the target, plus ci-lite's coverage job), so nothing is broken by not gating.

## Testing

Verified against the **real pinned artifact**, fetched the way CI does: `tinymemory-module-1.12.0-macos-26-arm64.tar.gz`, sha256 `4769fe8a60eb4eb0d2e3dd556bb1cdf89ec236074ecf779b634ea3718c691ab9`, matching `registry.rs:230` byte for byte. Run with the lane's own feature list from `scripts/ci/product-features.sh`.

| | result |
|---|---|
| **Before**, pristine `04075d537` | `FAILED. 1 passed; 1 failed` — CI's failure reproduced verbatim |
| **Before**, run alone | `FAILED. 0 passed; 1 failed`, identical message |
| **After** | `ok. 2 passed; 0 failed` |
| **Revert-check** (remove only the new block) | `FAILED`, same site, same message |

The revert-check is unusually direct: the error string literally names `set_modules_policy`, which is the call this PR adds.

### It executes, and it is not vacuous

The test reports `... ok` — **run, not `ignored` and not filtered** — and runtime went **0.24s → 2.02s**, which is the module actually loading and serving the reads.

It cannot pass on an empty result, which matters because `list_chunks_rpc` returns an empty page when the chunk tier is absent:

```rust
assert_eq!(listed.value.total, 2);
assert!(listed.value.chunks[0].content_preview.is_some());
assert!(sources.value.iter().any(|s| s.source_id == "gmail:…" && s.chunk_count == 1));
```

`total == 2` fails on an empty page and `chunks[0]` panics on it. So a pass proves the module really read the engine-seeded rows — the workspace pinning is doing the work. **No coverage is lost: not one assertion was changed, relaxed, or removed.**

`cargo fmt -p openhuman -- --check` clean; `cargo clippy -p openhuman --no-deps --test raw_coverage_all` exits 0.

### Pre-existing, not from this PR

`memory_threads_raw_coverage_e2e` fails 4 of 35 locally — identical `31 passed; 4 failed` on a stashed, pristine tree, so it is not mine. Different file, different process. The lane is not selecting that module right now, so it is not what is reddening `main`, but it will bite whoever next touches that area.

## Impact

Test-only. Unblocks `Rust Core Coverage` on `main` and therefore every open PR.

## Related

- Cause: `5828ad9d2` via #5725. No tracking issue was filed for the red main, so this carries no closing keyword.
- Not implemented here, offered as a note: the changed-modules scoping is what let this sit broken for ~18 hours and is 5-for-5 today. A **nightly** full-target `raw_coverage_all` run would catch this class without adding per-PR cost.

## Submission Checklist

- [x] Tests added or updated: this PR *is* the test fix. The failure path is the revert-check above, which reproduces CI's exact message.
- [x] N/A: diff coverage. The change is test-only; the added lines execute on every run of this test and the revert-check proves they are load-bearing.
- [x] No new external network dependencies introduced. The module artifact is the one CI already downloads and digest-verifies.
- [x] N/A: no linked issue — this fixes a red `main` directly and no issue was filed.

## Impact (platform)

None at runtime. Test-only; no production code touched.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/redmain`
- Commit SHA: `985e8b596`

### Validation Run

- [x] Focused tests: `cargo test -p openhuman --test raw_coverage_all -- memory_core_threads_raw_coverage_e2e::` with the lane's product feature set and the real pinned artifact. Before/alone/after/revert results in the table above.
- [x] Rust fmt/check: `cargo fmt -p openhuman -- --check` clean; `cargo clippy -p openhuman --no-deps --test raw_coverage_all` exits 0.
- [x] N/A: Tauri fmt/check — no Tauri code changed.

### Validation Blocked

- `command:` N/A
- `error:` N/A
- `impact:` N/A

### Behavior Changes

- Intended behavior change: none in the product. The test now publishes the host policy it always needed.
- User-visible effect: none.

### Parity Contract

- Legacy behavior preserved: yes. No assertion changed, so the test guards exactly what it did before #5725 broke it.
- Guard/fallback/dispatch parity checks: N/A — no dispatch or fallback logic touched.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Tests**
  * Improved memory end-to-end test setup to consistently use the test workspace.
  * Ensured seeded memory data is correctly available when validating read filters, graphs, scores, resets, and wipes.
  * Improved synchronization status test coverage by ensuring mock connections and persisted state are read from the correct test workspace.
  * Increased test reliability by preventing valid seeded data and synchronization results from being skipped or reported as empty.

### DIFF
--- a/tests/raw_coverage/memory_core_threads_raw_coverage_e2e.rs
+++ b/tests/raw_coverage/memory_core_threads_raw_coverage_e2e.rs
@@ -202,6 +202,32 @@ async fn memory_read_rpc_filters_graphs_scores_reset_and_wipe_seeded_rows() {
     let _env_lock = __shared_env_lock();
     let tmp = TempDir::new().unwrap();
     let cfg = config_in(&tmp);
+    // #5725 routed the chunk, entity and graph reads below off raw SQL and onto
+    // the memory contract (`5828ad9d2`), so they are answered by the tinymemory
+    // module rather than by this process. The seeding either side of them still
+    // writes through `tinymemory_core`'s store in-process, and the module opens
+    // whichever workspace it was given when it loaded — so unless the two name
+    // the SAME directory the writes and the reads land in different stores and
+    // the assertions below are checking an empty database.
+    //
+    // Two properties make this the only place it can be done. The module
+    // captures one workspace per process at load, and `set_modules_policy` is a
+    // `OnceLock` whose later calls are silently ignored
+    // (`modules/memory.rs:224`) — so the policy has to be published here,
+    // before the first read reaches the module, and it has to name `tmp`.
+    //
+    // The lane supplies the artifact itself: `rust-core-coverage` exports
+    // `TINYMEMORY_TEST_MODULE` (`ci-lite.yml:774`) and `local_override` picks it
+    // up for this module id (`modules/ops.rs:320-325`). Deliberately NOT gated
+    // on the module being present: a gate here would make this test silently
+    // skip in the one lane that runs it, which is the failure mode that let it
+    // sit broken in the first place.
+    #[cfg(feature = "modules")]
+    {
+        let mut policy = Config::default();
+        policy.workspace_dir = tmp.path().to_path_buf();
+        openhuman_core::openhuman::modules::memory::set_modules_policy(std::sync::Arc::new(policy));
+    }
     let ts0 = Utc.with_ymd_and_hms(2026, 5, 20, 9, 0, 0).unwrap();
     let chunks = vec![
         test_chunk(
--- a/tests/raw_coverage/memory_sync_tree_round21_raw_coverage_e2e.rs
+++ b/tests/raw_coverage/memory_sync_tree_round21_raw_coverage_e2e.rs
@@ -416,6 +416,37 @@ async fn slack_sync_status_rpc_reads_mock_connections_and_persisted_state() {
     let _home = EnvGuard::set_path("HOME", tmp.path());
     let _backend = EnvGuard::unset("BACKEND_URL");
     let mut config = config_in(&tmp);
+    // The sync state is written below through `HostSyncAdapter`, in this
+    // process, but `sync_status_rpc` reads it back through the memory contract
+    // — `as_source_sync().source_sync_state(..)`, answered by the tinymemory
+    // module. Without a published host policy the module cannot load, and the
+    // read fails.
+    //
+    // That failure is invisible from the assertion: `sync_status_rpc` logs the
+    // error and `continue`s (`providers/slack/rpc.rs:210`), so every row is
+    // skipped and the report comes back empty rather than failing. The symptom
+    // is `connections.len()` 0 vs 1, which reads like a write that did not
+    // persist; the cause is a read that could not run at all.
+    //
+    // Pinned to `config.workspace_dir` — NOT `tmp` — so the engine-routed write
+    // and the contract-routed read open the same store. `config_in` puts the
+    // workspace at `<tmp>/workspace`, so pinning the tempdir root instead makes
+    // the module open `<tmp>/memory` while the engine writes
+    // `<tmp>/workspace/memory/memory.db`, and every read then answers `None`
+    // from an empty database. Published here because the module captures one
+    // workspace per process at load and `set_modules_policy` is a `OnceLock`
+    // that silently ignores later calls (`modules/memory.rs:224`).
+    //
+    // The lane supplies the artifact via `TINYMEMORY_TEST_MODULE`
+    // (`ci-lite.yml:774`, picked up by `modules/ops.rs:320-325`), so this is
+    // deliberately not gated on the module being present — a gate would make
+    // the test skip in the only lane that runs it.
+    #[cfg(feature = "modules")]
+    {
+        let mut policy = Config::default();
+        policy.workspace_dir = config.workspace_dir.clone();
+        openhuman_core::openhuman::modules::memory::set_modules_policy(Arc::new(policy));
+    }
     let router = Router::new().route(
         "/agent-integrations/composio/connections",
         get(|| async {


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Second fix pushed to this PR (`c99b9d1c1`) — same class, same lane, so it lands here rather than in a separate PR.

## Category: test problem, NOT a product bug

I checked this before touching anything, because the symptom is exactly the read-side shape W3 is seeing. **The round trip is fine.** The write persists and the contract-routed read finds it — once both sides name the same workspace. Nothing was weakened; every assertion still stands.

Two separate causes were stacked behind one symptom:

1. **No host policy published**, so the module could not load and `source_sync_state` errored.
2. Once that was fixed, the policy was pinned to the **wrong directory**. `config_in` puts the workspace at `<tmp>/workspace`, so pinning the tempdir root made the module open `<tmp>/memory` while the engine wrote `<tmp>/workspace/memory/memory.db`. The module loaded correctly and every read still answered `None` from an empty database.

I only caught the second by dumping the tempdir tree — it presents identically to the first, and identically to a genuine persistence bug.

## Why it looked like a write that did not persist

`sync_status_rpc` logs a `source_sync_state` failure and `continue`s (`providers/slack/rpc.rs:210`), so a read that could not run **at all** comes back as an empty report rather than an error:

```
  left: 0
 right: 1        // outcome.value.connections.len()
```

Instrumented, the connection was found the whole time (`total_connections=3`, `conn-slack-round21 toolkit=slack active=true`); only the state load was failing.

**That masking is worth a look on its own, and I have not touched it.** In production the same path means a module that cannot answer produces "no connections" rather than a failure — a user would see an empty Slack sync status with nothing indicating anything went wrong. If that is worth filing, say so and I will.

## For W3

Both failures here were workspace/policy plumbing, not persistence. Before concluding those 10 read-side failures are real round-trip defects, worth checking each one publishes a policy **and** pins it to `config.workspace_dir` rather than the tempdir root. The second mistake is the quiet one: the module loads, the call succeeds, and the read just returns `None`.

## Proof

| | result |
|---|---|
| Before | `FAILED` at `:453`, `left: 0 / right: 1` — CI's failure verbatim |
| After | `ok. 4 passed; 0 failed` (0.15s → 2.31s, the module actually loading) |
| Revert-check | `FAILED` at `:453`, same values — names the assertion the fix restores |

Both modules in this PR are green together: `memory_core_threads` 2/2, `memory_sync_tree_round21` 4/4. Tests report `... ok`, not `ignored` or filtered. `cargo fmt --check` clean. Verified against the real pinned artifact (sha256 matching `registry.rs:230`) with the lane's own feature set.

This should also unblock #5823 and #5824 if they fail on the same test.

M3gA-Mind:
Closing this as **superseded by #5823**. Both fixes it carried are already on `main`.

## What happened

#5823 merged while I was working and fixed **both** tests — the same diagnosis, the same mechanism, in the same two files:

| test | #5823's fix |
|---|---|
| `memory_core_threads::memory_read_rpc_filters_…` | `set_modules_policy(Arc::new(cfg.clone()))` after `config_in` |
| `memory_sync_tree_round21::slack_sync_status_rpc_…` | `set_modules_policy(Arc::new(config.clone()))` before `sync_status_rpc` |

Its comment reaches the identical conclusion mine did: *"A row whose state read fails is skipped, so without a policy the response was honestly empty and this asserted 0 == 1."*

Its version is also **better than mine**. It publishes `config.clone()` where I set `workspace_dir` alone — which sidesteps the exact trap I fell into on the first attempt at the slack test, where I pinned the tempdir root and the module opened `<tmp>/memory` while the engine wrote `<tmp>/workspace/memory/memory.db`. The module loaded fine and every read still answered `None`. Cloning the whole config cannot drift that way.

## Verified rather than assumed

Rebased onto `871a8963e` and both commits dropped out:
- The `memory_core_threads` commit conflicted and was fully redundant.
- The slack commit still applied cleanly — but only because it sits a few lines earlier in the function, which would have left **two** `set_modules_policy` calls in one test, mine winning on `OnceLock` and #5823's becoming a silent no-op.

After dropping both, the branch is byte-identical to `main` (0 commits ahead, 0 diff), and each test carries exactly one policy publish.

Both tests pass on `871a8963e` with no change from me: `memory_core_threads` **2 passed**, `memory_sync_tree_round21` **4 passed** — run against the newly pinned **v1.13.2** artifact (sha256 `1fe0e4ba…47ae`, matching `registry.rs`), with the lane's own feature set.

## One thing worth knowing for the next person

Checking out the new `main` without syncing submodules gives **9 compile errors**, not a test failure — `unresolved import …provider::scoring`, `STORE_CORRUPT_KIND`, `methods::{EXTRACT_ENTITIES, EMBED_TEXT, EMBEDDER_SLUG}`, `as_scoring` not a trait member. #5823 bumped `vendor/tinymemory` `4141ec8` (v1.12.0) → `c580953` (v1.13.2) and the registry pin with it. It reads like main is broken; it is a stale gitlink. `git submodule update --init vendor/tinymemory`, and re-download the artifact — the old v1.12.0 `.dylib` no longer matches the pin.

## Standing finding, unaffected by this closure

`sync_status_rpc` still swallows a `source_sync_state` failure into an empty report via the `continue` at `providers/slack/rpc.rs:210`. In production that means a module which cannot answer yields "no connections" rather than an error, with nothing indicating a failure. Neither this PR nor #5823 changes that, and it is not a test concern. Happy to file it separately.

Issue #5830, opened 2026-08-27
Local-folder source with a relative path resolves against the Tauri build dir, and the error never says where it looked


M3gA-Mind:
## Problem

A local-folder memory source with a **relative** path silently resolves against the app's process working directory — which is the Tauri build directory — instead of the workspace. The sync fails permanently with an error that names the relative path and never reveals where it actually looked.

From a live session, a source configured as `docs` (label "Openhuman Docs"):

```
01:10:48 INF  [memory_sources] sync_rpc: entry source_id=src_21d1428cbc0f49e49ac38d0104175f29
01:10:48 WRN  [memory_sync:dispatcher] pipeline tick failed
              pipeline_id="workspace:folder:src_21d1428c…"  error=not found: folder does not exist: docs
01:10:49 ERR  rpc.invoke_method failed: ai.tinyhumans.tinymemory.Error.Other:
              sync source src_21d1428c…: not found: folder does not exist: docs
              (actions_called=0, provider_cost_usd=0)
```

The app's actual working directory:

```
/Users/megamind/tinyhuman/openhuman/app/src-tauri
```

There is no `docs` there. But two plausible targets do exist:

```
/Users/megamind/tinyhuman/openhuman/docs   (36 entries)
/Users/megamind/tinyhuman/docs             (18 entries)
```

So the path is only "missing" because it is resolved against a build directory no user would think of.

## Cause

`FolderReader::list_items` — `vendor/tinymemory/crates/tinymemory-sources/src/readers/folder.rs:41-57`:

```rust
async fn list_items(
    &self,
    source: &MemorySourceEntry,
    _workspace: &std::path::Path,          // handed the workspace, deliberately ignored
) -> SourceResult<Vec<SourceItem>> {
    let base_path = source.path.as_deref()
        .ok_or_else(|| MemoryError::Invalid("folder source requires a path".to_string()))?;

    let base = PathBuf::from(base_path);   // verbatim — no join, no canonicalise
    if !base.exists() {
        return Err(MemoryError::NotFound(format!("folder does not exist: {base_path}")));
    }
```

**`folder` is the only reader in the crate that ignores the workspace it is given.** Every sibling takes it as a live parameter:

| reader | signature |
| --- | --- |
| `conversation.rs` | `workspace: &Path` — and uses it: `let threads_dir = workspace.join(…)` |
| `github.rs` | `workspace: &Path` |
| `rss.rs` | `workspace: &Path` |
[...]

Issue #3162, opened 2026-06-01
Artifact UX polish: native save-file dialog + ChatToolCallEvent in-progress wiring


oxoxDev:
## Summary

Two intentional deviations from issue #2779's acceptance criteria — shipped as follow-up work to keep PR #3017 reviewable while parent issue #1535 closes.

## Problem

#3017 landed `ArtifactCard` + Tauri download command + backend `ArtifactReady`/`ArtifactFailed` events, but two of the original AC items from #2779 were intentionally deferred:

1. **AC#3 — native save-file dialog** is not wired. The `tauri-plugin-dialog` crate currently conflicts with the workspace's pinned `schemars` version, so the renderer can't open a `SaveDialog` to let the user pick a target path. The shipped behaviour copies the artifact into `~/Downloads/` and reveals it in the OS file manager, which works but is not what AC#3 specified.

2. **AC#1 — ArtifactCard in-progress state** only renders once the Rust core publishes `DomainEvent::ArtifactReady`/`Failed`. There is no spinner during the actual `generate_presentation` tool execution, so users see no UI between hitting send and the card appearing. The `ChatToolCallEvent` already fires when the tool dispatches; the card should subscribe to it and render an in-progress state keyed on the future artifact_id.

Spotted by @graycyrus in https://github.com/tinyhumansai/openhuman/pull/3017#pullrequestreview-4402841688.

## Solution (optional)

**For AC#3 — save-file dialog:**
- Upgrade `tauri-plugin-dialog` to a version compatible with the workspace's `schemars`, OR vendor a minimal save-file-dialog command in the Tauri shell that bypasses the plugin entirely (`rfd` crate is a candidate — already used by the openhuman shell elsewhere).
- Wire `ArtifactCard`'s "Download" button to invoke the dialog command; fall back to the current Downloads+reveal pattern when the user cancels or the dialog is unavailable on the platform.

**For AC#1 — in-progress state:**
- Extend `ChatToolCallEvent` to carry the artifact_id when the dispatching tool is `generate_presentation` (or surface it via a sibling `ArtifactPending` domain event from `create_artifact` at tool start).
- Subscribe `ArtifactCard` to the new event so it renders a "Generating..." skeleton between dispatch and `ArtifactReady`.
- Failure path: `ArtifactFailed` already flips the card to retry-hint; the in-progress skeleton just needs to hand off cleanly.

## Acceptance criteria

[...]

PR #5832, opened 2026-08-27
fix(memory-sources): use the native directory chooser and never store a bare folder name

### LINKED ISSUE #5831: Folder picker silently stores only the directory NAME when File.path is unavailable, creating a source that can never sync
## Summary

Split out of #5830 part 3. #5830's reader-side halves (resolve a relative folder path against the workspace; name the resolved path in the error) are fixed in tinyhumansai/tinymemory#113. This is the entry-side half, deliberately separated because the obvious version of it is **wrong**, and that is worth writing down before someone implements it.

`app/src/components/.../AddMemorySourceFields.tsx:155` renders `FolderField` for the folder kind as a **free-text input**. There is no native directory picker, nothing absolutises the value on the way in, and nothing checks that it resolves. A user types `docs` and that string is stored verbatim.

## What NOT to do, and why

The issue's own suggested direction offers "reject or normalise a relative path when the source is saved". Both options are worse than they look once tinymemory#113 lands:

- **Rejecting relative paths contradicts the reader.** After #113 a relative path is meaningful and useful — it anchors on the workspace. Refusing to save one would remove a working configuration.
- **Absolutising at save time is actively harmful, and this is the non-obvious part.** The workspace is not fixed for the life of a source: a desktop session moves workspace on login and again on logout. Baking `<workspace>/docs` into the stored value at save time freezes it to whichever workspace happened to be current, so the source silently points at the wrong tree — or a nonexistent one — after the next sign-in. Resolving at *read* time, which is what #113 does, survives that by construction.

So the resolution belongs in the reader, and it is already there. The entry-side work that remains is about **letting the user see and choose**, not about rewriting what they typed.

## What is actually worth building

1. **A real directory picker on `FolderField`** — a Tauri dialog `open({ directory: true })`, yielding an absolute path the user actually chose. This is the substantive fix for "easy to hit": most users want a specific folder and have no reason to reason about resolution roots at all.
2. **A resolve preview next to the field** — for a value that stays relative, show what it will resolve to (`docs` → `<workspace>/docs`). This is the cheap half and it removes the remaining surprise without constraining anyone.
3. Optionally, a non-blocking "that folder does not exist yet" hint at save time. A **hint**, not a rejection: a folder that does not exist yet is a legitimate thing to configure ahead of creating it.

## Why this is a separate issue

A directory picker is a Tauri dialog-plugin change with cross-platform surface (permissions, the macOS/Windows/Linux dialogs, and the plugin's capability allowlist) — a different kind of change from a reader-side path fix, and it should be reviewed on its own terms rather than buried under one.

## Not covered here

Whether the tinymemory fix has reached this app. It has not yet: OpenHuman routes memory-source sync through the driver, and the driver is the loaded `tinymemory` module — so #113 needs a tinymemory release plus a re-pin of the submodule gitlink **and** `modules/registry.rs` (version, `release_url`, per-platform sha256s) before any of it is live here. Tracked on #5830.

Refs #5830, tinyhumansai/tinymemory#113

M3gA-Mind:
## Summary

- Browse on the folder memory-source now opens the **OS-native directory chooser** and stores the absolute path it returns.
- The `webkitdirectory` fallback can no longer store a value that cannot work: when `File.path` is absent it reports an error next to the field instead of saving the bare directory name.
- Backed by `rfd`, **already in this shell** — no new dependency, no dialog plugin, no capability-allowlist entry.
- Hand-typed absolute paths are untouched. That is the current workaround and people are relying on it.

## Problem

`AddMemorySourceFields.tsx` rendered Browse as `<input type="file" webkitdirectory>`. That element carries no filesystem path. Chromium exposes one on the non-standard `File.path` only when the renderer has filesystem-aware integration, which the code's own comment said outright. Outside that renderer the handler fell through to:

```js
} else if (first.webkitRelativePath) {
  onChange(first.webkitRelativePath.split('/')[0]);   // the folder NAME, location discarded
}
```

A user who picked a folder got a source storing `docs`. Every sync then failed, once per cycle, indefinitely:

```
WRN pipeline tick failed pipeline_id="workspace:folder:src_d9a41ccf…"
    error=not found: folder does not exist: docs
```

Re-adding the same folder with a hand-typed absolute path worked immediately on the same build, same reader, same module. Only the path form differed.

The failure surfaces at sync time, far from the picker that caused it, and it never stops. Nothing in between could repair it, because there is nothing to repair: `docs` is not a relative path, it is a name whose location was thrown away.

### Why openhuman#5830 / tinycortex#158 do not cover this

Those resolve a **relative** path against the workspace, which is correct and unrelated. Resolving `docs` gives `<workspace>/docs` — still not the directory the user chose. The user would get a different wrong answer with a better error message.

### 1. A native chooser (`pick_directory_via_dialog`)

**What the app already had, since the brief asked.** There is no `@tauri-apps/plugin-dialog` anywhere in this repo, and no existing frontend directory-picker helper. What does exist is `artifact_commands::save_artifact_via_dialog` (#3162) — a native Save-As dialog driven by the **`rfd`** crate, declared in `app/src-tauri/Cargo.toml` with `default-features = false` + `xdg-portal` (which is what keeps Linux off GTK). Its module doc records that it deliberately talks to the OS dialog APIs directly rather than pulling a Tauri dialog/fs plugin, after a `schemars` version conflict.

So this follows that precedent instead of adding the plugin the issue suggested: a sibling command in a new `directory_picker` module, ~15 lines, using the same `rfd::AsyncFileDialog` already compiled into this shell. **No new dependency, no plugin, no capability entry, no permissions surface to get wrong across three platforms.**

It takes no input and returns only what the user chose in an OS-owned dialog, so unlike the artifact commands there is no path to re-validate — that is noted in the module docs so the asymmetry does not read as an oversight.

### 2. Never store an unusable value

Selection logic moved into `folderPicker.ts` as two pure-ish functions returning a tagged result:

| outcome | meaning | what the UI does |
|---|---|---|
| `{ ok: true, path }` | an absolute path | store it |
| `cancelled` | user dismissed the chooser | nothing, silently |
| `unavailable` | no native chooser here | fall back to the directory input |
| `no-absolute-path` | a directory was chosen, its location is unknown | **show an error, store nothing** |

`webkitRelativePath` is deliberately no longer consulted as a fallback. Its first segment is the defect, not a degraded-but-usable answer.

### 3. Existence validation at save time — considered, and not added

The native chooser can only return a directory that exists, so the path this PR fixes is covered by construction. A hand-typed path that does not exist *yet* is a legitimate thing to configure ahead of creating it, which is why the issue asks for a hint rather than a rejection. Checking it from the renderer would need a new filesystem-read command and a new trust boundary for what is explicitly only a hint, and that is a bigger change than the defect warrants. Worth doing as its own piece if the hint is still wanted.

## Testing

`app/src/components/intelligence/__tests__/folderPicker.test.ts` — 10 tests, all passing.

### Revert-check

Restoring the pre-fix fallback (`return { ok: true, path: webkitRelativePath.split('/')[0] }`) fails the new assertions, reproducing the exact production value from the issue:

```
× refuses to produce a path when the renderer does not expose File.path
  AssertionError: expected { ok: true, path: 'docs' } to deeply equal { ok: false, … }
× never returns the bare directory name for any nesting depth
  AssertionError: expected true to be false

Tests  2 failed | 8 passed (10)
```

The other 8 keep passing, so the two that fail are the two that pin this defect.

### Checks run

| check | result |
|---|---|
| `vitest` (the one new file) | 10 passed |
| `pnpm typecheck` | clean |
| `pnpm i18n:check` | clean |
| `pnpm i18n:english:check` | `total unexpected English: 0` |
| `eslint` (changed files) | clean |
| `prettier --check` | clean |
| `cargo fmt --check` (`app/src-tauri`) | clean |

Per the brief I did not run the full suite or a full build.

**The Rust does not compile locally in this session** — a `cargo check` of the Tauri shell is a full build of that crate. Instead I verified the two APIs against the exact vendored source (`rfd-0.15.4`): `AsyncFileDialog::pick_folder(self) -> impl Future<Output = Option<FileHandle>>` (`file_dialog.rs:253`) and `FileHandle::path(&self) -> &Path` (`file_handle/native.rs:136`) — the same pair `save_artifact_via_dialog` already uses in this crate. The rest is two lines in `lib.rs`. The Tauri lane covers it; flagging it rather than implying a check I did not run.

Also verified `mod directory_picker;` landed **unguarded** — it sits directly after `mod deep_link_ipc_windows;`, which consumed the `#[cfg(target_os = "windows")]` above it. A module declaration inserted next to a `cfg` is an easy way to silently platform-gate something that must exist everywhere.

## i18n

One new key, `memorySources.folderPathUnavailable`, with a real translation in **all 14 locale files** (ar, bn, de, en, es, fr, hi, id, it, ko, pl, pt, ru, zh-CN). No em dashes, per the repo rule. `i18n:english:check` confirms none was left as English.

The error text avoids naming the mechanism and says what to do instead: *"Could not determine where that folder is. Type its full path instead."*

## Impact

Desktop and browser. No migration, no schema change, no config change. Existing folder sources are untouched — this only changes what a **new** selection can produce. A source already poisoned with a bare name still needs re-adding; nothing here can recover the location that was discarded when it was saved.

## Related

- Closes tinyhumansai/openhuman#5831
- Refs #5830, tinyhumansai/tinymemory#113 — the reader-side relative-path work, independent of this.

## Submission Checklist

- [x] Tests added: `folderPicker.test.ts`, 10 cases. Failure path is the revert-check above, plus explicit coverage of cancellation, an unavailable dialog, a blank return, and Windows separators.
- [x] N/A: diff coverage. The changed frontend lines are covered by the new test; the ~15 Rust lines are a dialog invocation with no branching logic to assert without a real OS dialog.
- [x] No new external network dependencies introduced. `rfd` was already a dependency of this shell.
- [x] Linked issue closed via `Closes tinyhumansai/openhuman#5831`.

## Impact (platform)

Desktop (macOS/Windows/Linux) gains the native chooser; a browser context keeps the input fallback and now fails visibly instead of silently. No performance, security, migration, or compatibility implications.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/5831-native-folder-picker`
- Commit SHA: `b8943b440`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — ran `prettier --check` over the changed files; clean.
- [x] `pnpm typecheck` — clean.
- [x] Focused tests: `vitest run src/components/intelligence/__tests__/folderPicker.test.ts` — 10 passed; revert-check reproduces the `'docs'` value.
- [x] Rust fmt/check: `cargo fmt --check` clean. **`cargo check` NOT run** — it is a full build of the Tauri shell, excluded by the brief; APIs verified against the vendored `rfd-0.15.4` source instead, as described above.
- [x] N/A: Tauri fmt/check beyond the above — no other shell code changed.

### Validation Blocked

- `command:` `cargo check --manifest-path app/src-tauri/Cargo.toml`
- `error:` not a failure — excluded as a full build by the task constraints
- `impact:` the new Tauri command is unverified by a local compile. Two lines of registration plus a 15-line command using an API pair already used in the same crate, verified against the vendored source. The Tauri CI lane covers it.

### Behavior Changes

- Intended behavior change: Browse opens a native chooser; a selection that yields no absolute path is refused with a visible error rather than silently stored.
- User-visible effect: folder sources added via Browse now sync. A failed selection says so immediately instead of failing at every sync forever.

### Parity Contract

- Legacy behavior preserved: hand-typed absolute paths, and the `File.path` derivation where the renderer supplies one, both behave as before.
- Guard/fallback/dispatch parity checks: the `webkitdirectory` input remains the fallback for browser contexts and hosts with no working dialog; only its unusable output was removed.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

- **New Features**
  - Added native folder selection through the operating system’s directory chooser.
  - Added a fallback folder picker for environments without native support.
  - Folder selections now preserve the complete absolute path across platforms.
  - Added clear guidance when the folder path cannot be determined.

- **Bug Fixes**
  - Prevented folder synchronization failures caused by storing only a folder name.
  - Cancelled selections now leave the existing folder field unchanged.

- **Localization**
  - Added the folder-path guidance message across supported languages.

- **Tests**
  - Added coverage for successful, cancelled, unavailable, and invalid folder selections.

### DIFF
--- a/app/src-tauri/src/directory_picker.rs
+++ b/app/src-tauri/src/directory_picker.rs
@@ -1,0 +1,76 @@
+//! Native directory chooser for memory-source configuration (#5831).
+//!
+//! The folder memory-source used to be picked with an
+//! `<input type="file" webkitdirectory>` in the renderer. That element does
+//! not expose a filesystem path: Chromium only carries one on the
+//! non-standard `File.path` attribute, and only when the renderer has
+//! filesystem-aware integration. Outside that renderer the handler fell back
+//! to `webkitRelativePath.split('/')[0]` — the chosen directory's **name**
+//! with its location discarded — and stored it. The source then looked
+//! configured and could never sync, failing once per cycle forever with
+//! `folder does not exist: docs`.
+//!
+//! A native dialog has no such gap: it returns an absolute path in every
+//! renderer, on every platform.
+//!
+//! ## Why `rfd` and not `tauri-plugin-dialog`
+//!
+//! This mirrors [`crate::artifact_commands::save_artifact_via_dialog`]
+//! (#3162), which is already in this shell and already talks to the OS
+//! dialog APIs through `rfd`. Reusing it means **no new dependency, no new
+//! plugin, and no capability-allowlist entry** — the crate is declared in
+//! `Cargo.toml` with `default-features = false` plus `xdg-portal`, which is
+//! what keeps Linux off GTK. Adding the dialog plugin for one command would
+//! widen the dependency graph for a capability the shell already has.
+//!
+//! ## Trust boundary
+//!
+//! Deliberately none. Unlike the artifact commands — which re-validate that
+//! a renderer-supplied path sits inside the artifacts tree, because there
+//! the renderer *supplies* the path — this command takes no input and
+//! returns only what the user chose in an OS-owned dialog. The renderer
+//! cannot steer it at a directory, and picking a folder to index is the
+//! user's decision to make anywhere on their disk.
+
+/// Open the OS-native directory chooser and return the absolute path of the
+/// directory the user selected.
+///
+/// Returns:
+/// - `Ok(Some(path))` — the absolute path chosen.
+/// - `Ok(None)` — the user dismissed the dialog. Not an error; the caller
+///   leaves the field untouched.
+/// - `Err(_)` — the dialog could not run (for example, no xdg-desktop
+///   portal on a headless Linux host), or it somehow yielded a relative
+///   path. The caller surfaces this rather than storing anything.
+#[tauri::command]
+pub async fn pick_directory_via_dialog() -> Result<Option<String>, String> {
+    // On Linux this is the xdg-desktop portal (no GTK link); on macOS and
+    // Windows the system panel. The await resolves when the user picks or
+    // cancels.
+    let handle = rfd::AsyncFileDialog::new().pick_folder().await;
+
+    let Some(dir) = handle else {
+        log::info!("[directory_picker] pick_directory_via_dialog cancelled by user");
+        return Ok(None);
+    };
+
+    let path = dir.path().to_path_buf();
+
+    // The OS dialogs all hand back absolute paths, so this is a belt-and-braces
+    // check rather than a branch we expect to take. It exists because a
+    // relative value reaching the store is the entire defect this command was
+    // written to remove: an error here is recoverable and visible, whereas a
+    // stored relative path is neither.
+    if !path.is_absolute() {
+        return Err(format!(
+            "the directory chooser returned a non-absolute path: {}",
+            path.display()
+        ));
+    }
+
+    log::info!(
+        "[directory_picker] pick_directory_via_dialog chose {}",
+        path.display()
+    );
+    Ok(Some(path.display().to_string()))
+}
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -43,6 +43,7 @@ mod core_rpc;
 mod deep_link_ipc;
 #[cfg(target_os = "windows")]
 mod deep_link_ipc_windows;
+mod directory_picker;
 // Cross-platform module: the registry-reading function is windows-only, but
 // the parsing helpers compile (and test) everywhere so `cargo test` on the
 // developer host covers them.
@@ -3371,6 +3372,7 @@ pub fn run() {
             // and the Save-As fallback needs it there (CodeRabbit on #4127).
             artifact_commands::save_artifact_via_dialog,
             artifact_commands::download_artifact_to_downloads,
+            directory_picker::pick_directory_via_dialog,
             // Structured WhatsApp data (store lives shell-side).
             whatsapp_data::whatsapp_data_list_chats,
             whatsapp_data::whatsapp_data_list_messages,
--- a/app/src/components/intelligence/AddMemorySourceFields.tsx
+++ b/app/src/components/intelligence/AddMemorySourceFields.tsx
@@ -19,6 +19,7 @@ import type { ComposioConnection } from '../../lib/composio/types';
 import { useT } from '../../lib/i18n/I18nContext';
 import type { SourceKind } from '../../services/memorySourcesService';
 import TextField from '../ui/TextField';
+import { directoryPathFromPickedFiles, pickDirectoryNatively } from './folderPicker';
 
 const log = debug('intelligence:add-memory-source-dialog');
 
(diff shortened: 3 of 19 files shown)


M3gA-Mind:
@coderabbitai review

M3gA-Mind:
Closing this — the native chooser it adds cannot run in the renderer the app ships. Recording the findings so none of the work is lost.

## What was right, and should come back

**The refuse-to-store guard is correct and valuable.** On `main`, picking a folder in a non-CEF renderer silently stores the bare directory name and produces a source that looks configured and can never sync (#5831). With this PR the field reports the failure instead. That behaviour was verified by hand: Browse produced *"Could not determine where that folder is. Type its full path instead."* rather than silently saving `docs`.

The Rust half is also correctly wired — `mod directory_picker;` (`lib.rs:46`) and `directory_picker::pick_directory_via_dialog` in the `invoke_handler` (`lib.rs:3375`). Nothing is wrong there.

## Why it cannot work as written

`folderPicker.ts:93-96` gates the native path on `isTauri()`:

```ts
export async function pickDirectoryNatively(): Promise<FolderPickResult> {
  if (!isTauri()) return { ok: false, reason: 'unavailable' };
```

and `isTauri()` (`common.ts:30-44`) requires `window.__TAURI_INTERNALS__.invoke`, logging *"isTauri() -> false: IPC bridge not wired (CEF bootstrap…)"* when absent.

**The app runs CEF.** Measured on a live build: 66 CEF helper processes, a per-user `cef/` directory, and the shell starting its own transport because Tauri IPC does not exist there:

```
[webview_apis] server listening on 127.0.0.1:53495 (OS-assigned ephemeral)
[webview_apis] bridge ready on port 53495
```

So `isTauri()` is correctly `false`, `pickDirectoryNatively()` short-circuits to `unavailable`, the `webkitdirectory` fallback runs, cannot produce a path, and the user sees the error. Browse is unusable in the shipping renderer.

Worth stating plainly to avoid repeating it: this PR moved off `File.path` *because* that only exists under CEF, and landed on a transport that only exists under Wry. Wry has never shipped.

## Why CI did not catch it

`folderPicker.test.ts` mocks the invoke layer, so it passes whether or not the transport exists. The defect is only observable by clicking Browse in a running CEF build. A test that mocks the transport proves the logic, never the wiring.

## What a replacement needs

1. Reach the shell over the transport that **works under CEF** — `webview_apis` is the obvious candidate; confirm against a command that demonstrably works in a running build rather than assuming.
2. Work in **both** renderers, and say explicitly how each reaches it.
3. **Keep this PR's guard verbatim.** It is the part that protects users when no path can be obtained.
4. Reconsider whether `isTauri()` is the right gate, or whether the question is "can I reach the shell?" rather than "am I under Tauri?".

#5831 stays **open** — the defect it describes is real and still unfixed. Only this approach is being withdrawn.

Current workaround for users: type an absolute path. That works today and is unaffected by any of this.

PR #5848, opened 2026-08-30
Take the connector vocabulary from tinyconnectors

senamakel:
First half of the connector extraction's host wiring. The Composio payload
types now come from `tinyconnectors-bus` instead of `tinymemory-api`, and
`tinyconnectors` is registered as a loadable module.

Companion to https://github.com/tinyhumansai/tinyconnectors (v0.3.x), which
holds the connector implementation migrated out of this repo and `tinymemory`.

## What changed

- **`vendor/tinyconnectors`** added as a submodule, pinned to `main`.
- **`tinyconnectors-bus`** taken as a path dependency.
- **`integrations::composio::types`** re-exports from it rather than from
  `tinymemory_api::host::composio`.
- **`TINYCONNECTORS` module record** in `modules::registry`, with the real
  SHA-256 digests from the published `checksum.toml`, `LoadPolicy::Lazy`.

## Why the types moved

They were in `tinymemory-api` only because two crates needed to name them and
there was nowhere else both could — the memory contract crate's own comment
says as much. They belong with the crate that produces the shapes. Holding one
definition in two repositories is a drift waiting to happen: both sides
serialize and deserialize fine while quietly disagreeing about a field.

## The re-encode seam

`memory/host_impls.rs` and `ops/toolkits.rs` implement or feed
`tinymemory_core::composio_host::ComposioHost`, whose signatures still use the
parallel copy. `integrations::composio::types::reencode` bridges them.

It is a re-encode, not a translation — the shapes are identical — and it is
temporary. Phase 4 of the extraction deletes `ComposioHost` along with the
memory-side copy, and every call to `reencode` goes with it. It is one function
rather than conversions spread across call sites, so removing it later is a
matter of deleting it and following the compiler.

## What has *not* changed

No behaviour. `integrations::composio` still owns the client, the OAuth
handoff, and the execute path; nothing calls the module yet. That is the next
change, and it is where the care is needed:

- **Egress enforcement must stay host-side.** `enforce_egress` and
  `emit_external_transfer` run before every Composio call today. The module
  does not do this and must not — it is host policy about the user's data.
- **`ComposioTriggerSubscriber` stays.** The backend HMAC-verifies webhooks and
  fans them out over the user's sockets; the module has no socket.
- **Scope enforcement becomes the module's.** It hides forbidden actions from
  `ListTools` and refuses them in `Execute`, so the host should not re-filter
  against a separately stored preference.

See `docs/specs/2026-08-30-host-integration.md` in the tinyconnectors repo.

## Validation

```
cargo check --all-targets     # clean
cargo test -p openhuman --lib modules::registry   # 9 passed
```

The registry tests include the uniqueness and platform-coverage checks that
guard a new `ModuleRecord`.

## Summary by CodeRabbit

* **New Features**
  * Added support for the TinyConnectors OAuth integration.
  * Registered the Composio connector module across Ubuntu, macOS, and Windows.
  * Added connector communication contracts for connections, capabilities, and execution responses.
  * Updated the connector module to version 0.3.1.

* **Bug Fixes**
  * Improved compatibility when exchanging Composio integration data between host components.
  * Added clearer error handling for incompatible connector response formats.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -56,4 +56,4 @@
 [submodule "vendor/tinyconnectors"]
 	path = vendor/tinyconnectors
 	url = https://github.com/tinyhumansai/tinyconnectors
-	branch = connectors-migration
+	branch = main
--- a/.gitmodules
+++ b/.gitmodules
@@ -53,3 +53,7 @@
 	path = vendor/tinyjuice
 	url = https://github.com/tinyhumansai/tinyjuice
 	branch = main
+[submodule "vendor/tinyconnectors"]
+	path = vendor/tinyconnectors
+	url = https://github.com/tinyhumansai/tinyconnectors
+	branch = connectors-migration
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6413,7 +6413,7 @@ dependencies = [
 
 [[package]]
 name = "tinyconnectors-bus"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6413,7 +6413,7 @@ dependencies = [
 
 [[package]]
 name = "tinyconnectors-bus"
-version = "0.2.1"
+version = "0.3.0"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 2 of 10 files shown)


senamakel:
Superseded by #5849.

This branch was cut from a local `main` that carried unrelated in-flight work, and the auto-commit hook swept it in: 54 files instead of 15, including 14 unintended submodule pin moves that the submodule-monotonicity gate correctly caught.

#5849 is the same change cut from `upstream/main` — 15 files, one gitlink, clippy clean under CI's `-D warnings`.

PR #5849, opened 2026-08-30
Take the connector vocabulary and module from tinyconnectors

senamakel:
First half of the connector extraction's host wiring: the Composio payload types
now come from `tinyconnectors-bus` instead of `tinymemory-api`, and
`tinyconnectors` v0.3.1 is registered as a loadable module.

Companion to [tinyhumansai/tinyconnectors](https://github.com/tinyhumansai/tinyconnectors),
which holds the connector implementation migrated out of this repo and
`tinymemory`.

> Replaces #5848, which was branched from a local `main` carrying unrelated
> in-flight work and so picked up 40 stray files plus 14 unintended submodule pin
> moves. This branch is cut from `upstream/main`: 15 files, one gitlink.

## What changed

- **`vendor/tinyconnectors`** submodule, pinned to the `v0.3.1` tag.
- **`tinyconnectors-bus`** as a path dependency — two pure-Rust dependencies and
  no transport, so this crate compiles none of the module.
- **`integrations::composio::types`** re-exports from it rather than from
  `tinymemory_api::host::composio`.
- **`TINYCONNECTORS` module record** with the real SHA-256 digests from the
  published `checksum.toml`, `LoadPolicy::Lazy`.
- **`modules::connectors`** — the client: `ensure_loaded`, a proxy, and member
  names spelled through the contract rather than as string literals.
- **`scripts/ci/check-module-pins.mjs`** gains its `PIN_MAP` entry. That gate
  exists to fail on a module added without a decision, and it did.

## Why the types moved

They sat in `tinymemory-api` only because two crates needed to name them and
there was nowhere else both could — that crate's own comment says as much. They
belong with the crate that produces the shapes. One definition living in two
repositories is a drift waiting to happen: both sides serialize and deserialize
cleanly while quietly disagreeing about a field.

## The re-encode seam

`memory/host_impls.rs` and `ops/toolkits.rs` implement or feed
`tinymemory_core::composio_host::ComposioHost`, whose signatures still use the
parallel copy. `integrations::composio::types::reencode` bridges them.

It is a re-encode, not a translation — the shapes are identical — and it is
temporary. Phase 4 deletes `ComposioHost` along with the memory-side copy and
every call to it goes too. One function rather than conversions spread across
call sites, so removing it later is deleting it and following the compiler.

## No behaviour change

`integrations::composio` still owns the client, the OAuth handoff and the
execute path; nothing calls the module yet. That is the next change, and three
things must not be lost in it:

- **Egress enforcement stays host-side.** `enforce_egress` and
  `emit_external_transfer` run before every Composio call. The module does not
  do this and must not — it is host policy about the user's data, and the module
  cannot see the reasons behind it.
- **`ComposioTriggerSubscriber` stays.** The backend HMAC-verifies webhooks and
  fans them out over the user's sockets; the module has no socket.
- **Scope enforcement becomes the module's.** It hides forbidden actions from
  `ListTools` and refuses them in `Execute`, so the host must not re-filter
  against a separately stored preference — two sources of truth for a permission
  is how one of them goes stale and permissive.

Details in `docs/specs/2026-08-30-host-integration.md` in the tinyconnectors repo.

## Validation

```
cargo check --all-targets                                        # clean
cargo clippy -p openhuman -- -D warnings                         # clean (CI's exact command)
node scripts/ci/check-submodule-monotonic.mjs upstream/main HEAD # clean
```

The published v0.3.1 artifact was also verified end to end: downloaded from the
release, digest-checked, and loaded through the real TinyBus loader with its
declared members matched against the contract's table.

## Summary by CodeRabbit

- **New Features**
  - Added TinyConnectors support for OAuth connections, actions, triggers, and record synchronization.
  - Added lazy loading and runtime invocation of connector operations.
  - Added direct and backend routing options for connector operations.
  - Added remote trigger-history retrieval and improved synchronization of external records.

- **Bug Fixes**
  - Improved compatibility across connector capability, connection, and execution data.
  - Added safer response conversion to detect data-shape mismatches.
  - Improved handling of unsupported routes, missing credentials, rate limits, and connector failures.


senamakel:
Two CI notes, both now resolved or accounted for:

**Rust Quality — fixed.** `cargo fmt` wanted a different import ordering than the crate I came from. Pushed `7a53de1`; `cargo fmt --all -- --check` and `cargo clippy -p openhuman -- -D warnings` are both clean locally.

**Scripts Self-Tests — pre-existing, not from this PR.** The failure is:

```
✖ a failed raw coverage module fails the run even when a later module succeeds
  scripts/__tests__/coverage-runner-status.test.mjs:80
```

I checked it against the base rather than assuming: with `scripts/` restored to `upstream/main` in the same working tree, that test still fails (3 pass / 1 fail), so it is environment-dependent and unrelated to the connector change. This PR's only `scripts/` edit is one line adding `tinyconnectors` to `PIN_MAP` in `check-module-pins.mjs` — and `node --test scripts/__tests__/module-pins.test.mjs` passes 24/24, with the gate itself reporting:

```
Module pin check OK — 10 record(s) in modules::registry::ALL, all pins accounted for.
```

Happy to split the `PIN_MAP` line into its own PR if that would make this one's signal cleaner.

senamakel:
Pushed `510bd1f` fixing the Scripts Self-Tests failure. It turned out to be a real bug, not just an environment quirk, so it was worth fixing rather than waiving.

`coverage-runner-status.test.mjs` asserts that a failed raw-coverage module fails the whole run — the `|| return` in `run_integration_target`'s loop. But the runner now gates that target behind `target_features_satisfied`, and the test's stub set did not neutralise the gate. So the runner logged `skipping raw_coverage_all: required-features not in the product set` and returned success, and the test asserted failure propagation **while never producing a failure**.

It only surfaced now because Scripts Self-Tests is filtered on `scripts` changing, and this is the first PR to touch `scripts/` since the gate was added. The fix stubs `target_features_satisfied` so the test exercises what it claims:

```
node --test scripts/__tests__/coverage-runner-status.test.mjs   # 4/4
```

Full suite is 199/201 locally; the two remaining are `scripts/mock-api/socket.*.test.mjs` failing on `Cannot find module 'socket.io-client'` because I have not run `pnpm install` in this worktree — CI installs first, so they pass there.

PR #5850, opened 2026-08-30
mcp: honour globs in allowed_tools / disallowed_tools

theamazinghenk:
Two agent definitions shipped in this repository configure a deny rule that
never fires:

```toml
disallowed_tools = ["tinyplace_*"]
```

The comment above it in `tools_agent/agent.toml` explains the intent:

> Specialist-owned trading/tiny.place tools are also stripped via
> `disallowed_tools` above so they route through their dedicated agents
> exclusively.

No tool is literally named `tinyplace_*`, so nothing is stripped. The generalist
keeps the second, weaker route to those capabilities that the comment says it
should not have. The configuration reads as if a boundary exists; there is none.

Both places that resolve these lists compare with `==`:

```rust
// src/openhuman/mcp/config_servers/registry.rs — McpServerDefinition::is_tool_allowed
if self.disallowed_tools.iter().any(|name| name == tool) {
    return false;
}
self.allowed_tools.is_empty() || self.allowed_tools.iter().any(|name| name == tool)
```

```rust
// src/openhuman/agent/library/ops.rs:96 — direct_tool_names
names.retain(|name| !def.disallowed_tools.contains(name));
```

`normalize_tool_names` trims and de-duplicates the configured entries but does
not interpret them, so `tinyplace_*` survives into the comparison as a literal
string and matches nothing.

There is a second, less visible consequence. An allow-list can by definition not
contain a tool that does not exist yet. Any MCP server whose tool set is not
fixed at configuration time — one that lets an agent register a tool during a
session — cannot be constrained: the operator's only options are an empty list
("every tool, forever") or a list that quietly goes stale.

Add `tool_pattern_matches(pattern, tool)` and use it at both sites.

* A pattern containing no `*` compares exactly, as today. Every existing
  configuration keeps the meaning it has now.
* `*` stands for any run of characters. `tinyplace_*` (prefix), `*_read`
  (suffix), `browser_*_click` (middle) and a bare `*` all work.
* Deliberately **not** a regular expression. These lists are written by
  operators in `agent.toml`, and `*` is the character they already reach for; a
  regex would silently reinterpret `.` and `+` in existing names.

Six tests, including the shipped `tinyplace_*` case and one through
`McpServerRegistry` that pins the resolution order (a deny pattern beats an allow
pattern, and a tool name that did not exist when the allow-list was written is
still covered by `docs_*`).

Once globs work, the two `tinyplace_*` deny entries **start stripping**. That is
what the configuration and its comment ask for, but it is not a no-op: agents
that today reach tiny.place tools through the generalist will stop being able to,
and will route through their dedicated agents as intended.

If you would rather land the matcher without changing what those two agent
definitions do, the alternative is to keep the code change and replace the two
patterns with the explicit list of tool names — but then the comment above them
should say that the list has to be maintained by hand.

### DIFF
--- a/src/openhuman/agent/library/ops.rs
+++ b/src/openhuman/agent/library/ops.rs
@@ -94,13 +94,9 @@ fn direct_tool_names(def: &AgentDefinition) -> Vec<String> {
         }
     }
     names.retain(|name| {
-        def.disallowed_tools.iter().all(|entry| {
-            if let Some(prefix) = entry.strip_suffix('*') {
-                !name.starts_with(prefix)
-            } else {
-                entry != name
-            }
-        })
+        def.disallowed_tools
+            .iter()
+            .all(|entry| !crate::openhuman::mcp::config_servers::registry::glob_match(entry, name))
     });
     names.sort();
     names
--- a/src/openhuman/agent/library/ops.rs
+++ b/src/openhuman/agent/library/ops.rs
@@ -93,7 +93,15 @@ fn direct_tool_names(def: &AgentDefinition) -> Vec<String> {
             names.push(tool.clone());
         }
     }
-    names.retain(|name| !def.disallowed_tools.contains(name));
+    names.retain(|name| {
+        def.disallowed_tools.iter().all(|entry| {
+            if let Some(prefix) = entry.strip_suffix('*') {
+                !name.starts_with(prefix)
+            } else {
+                entry != name
+            }
+        })
+    });
     names.sort();
     names
 }
--- a/src/openhuman/mcp/config_servers/mod.rs
+++ b/src/openhuman/mcp/config_servers/mod.rs
@@ -28,7 +28,7 @@
 //! The HTTP transport primitives live in the ungated sibling
 //! [`super::http_client`]; see that module for why it is not gated.
 
-mod registry;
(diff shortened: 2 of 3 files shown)


theamazinghenk:
Upstream heeft dit inmiddels anders opgelost en de PR ingehaald:

1. `src/openhuman/mcp/config_servers/` (waar `McpServerDefinition::is_tool_allowed` woonde) is in main verwijderd — de allow/deny-enforcement is verplaatst naar de session-builder (`definition_disallows_tool` in `factory.rs:1361`), die al suffix-`*`-globs ondersteunt.
2. De enforce is daar **fail-closed en pre-transport** (zie `src/openhuman/mcp/README.md`).

Wat in main nog wel substring-gebaseerd is: `direct_tool_names` in `src/openhuman/agent/library/ops.rs` gebruikt nog `def.disallowed_tools.contains(name)` (exact-match). Dat is dezelfde klasse bug als deze PR fixte, maar op een read-only rapportage-route (tool-counts in de library-UI), niet op een enforcement-route.

Onze branch met de gedeelde `glob_match`-helper + tests blijft op de fork (`upstream-01`) staan; als jullie `direct_tool_names` ook globs willen geven, graag. Anders kan deze PR dicht als superseded.

PR #5879, opened 2026-08-31
test(inference): controller coverage for the eleven local RPCs, plus three untested UI surfaces

M3gA-Mind:
## Summary

- Adds controller-boundary e2e coverage for all **eleven** local inference RPCs, the module the repo's own coverage gate scored worst (1/11).
- Adds the first tests for three inference-facing UI surfaces that had **zero**: `EmbeddingsSetupModal`, `ProviderSetupErrorNotice`, `AgentInsightsPreview`.
- 4 files, +1,093 lines. No product code touched.

> **Test lane:** 1 Rust e2e target (real handler dispatch) + 3 vitest component suites (jsdom, mocked transport). No browser-level tests.

## Problem

`scripts/check-domain-e2e-coverage.mjs` reported inference at **1/11 = 9.1%**, by far the worst module.

Two things were found while closing it, and both shaped the work:

1. **The gate measures a string literal**, not behaviour — it counts `"openhuman.<ns>_<fn>"` appearing anywhere under `tests/**/*_e2e.rs`, comments included. Five of the ten "missing" controllers turned out to be substantively covered already, driven through their real registered handlers. Satisfying the gate by adding the literal would have been gaming it.
2. **No test anywhere asserted a controller's registered wire name.** Everything looks up by `function`. A namespace rename would silently break every frontend caller with no Rust test failing.

## Solution

Rather than restate coverage the suite already has, this asserts the contract that was genuinely missing: `rpc_method_name() == "openhuman.inference_<fn>"` for all eleven, driven through real handler dispatch.

Mocks only — `serve_provider_mock` and mock piper/ollama binaries, reusing existing helpers. No real downloads or model fetches.

One documented limit, stated in the test rather than hidden: `transcribe` has no offline-testable engine — the bundled whisper.cpp was removed and it is a hosted proxy call now — so it is exercised at the controller boundary (validation, error mapping, requires-sign-in) and the file says so.

## Submission Checklist

## Impact

- **CI:** adds one `tests/raw_coverage/` target, picked up automatically by `build.rs`'s directory glob — no shared file edited.

## Related

- Closes:
- Follow-up PR(s)/TODOs: the coverage gate's string-match weakness is fixed separately on `fix/domain-e2e-coverage-gate`.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `test/inference-e2e-coverage`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — prettier clean on the three new `.tsx` files.
- [x] `pnpm typecheck` — `tsc --noEmit`, 0 errors.
- [x] Focused tests: the three new vitest suites pass; the Rust target compiles and runs in the root world.
- [x] Rust fmt/check (if changed): `cargo fmt` clean; the new `raw_coverage` target builds.

### Validation Blocked

### Behavior Changes

- User-visible effect: none.

### Parity Contract

- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for embeddings setup, including key handling, validation, saving, feedback, and loading states.
  * Added coverage for provider connection error messages, detail disclosure, and edge cases.
  * Increased confidence in Agent Insights previews, including timelines, metrics, errors, and source panels.
  * Added offline integration coverage for local inference features, including speech, transcription, asset downloads, agent chat, connection checks, and validation scenarios.

### DIFF
--- a/app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
+++ b/app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
@@ -15,7 +15,24 @@ import type {
 } from '../../../../services/api/embeddingsApi';
 import EmbeddingsSetupModal, { type EmbeddingsSetupModalProps } from '../EmbeddingsSetupModal';
 
-vi.mock('../../../../lib/i18n/I18nContext', () => ({ useT: () => ({ t: (key: string) => key }) }));
+/**
+ * The mock returns the key for most lookups, which keeps assertions readable.
+ * Two keys are the exception: `testSuccess` and `testFailed` are TEMPLATES that
+ * the component fills with `.replace('{dims}', …)` / `.replace('{error}', …)`
+ * (`EmbeddingsSetupModal.tsx:191-195`). Returning the bare key there makes
+ * `.replace()` a no-op, so the substituted value never reaches the DOM and a
+ * component that stopped surfacing it would keep the test green. Returning the
+ * real templates (from `en.ts`) is what makes the substitution assertable.
+ */
+vi.mock('../../../../lib/i18n/I18nContext', () => ({
+  useT: () => ({
+    t: (key: string) => {
+      if (key === 'settings.embeddings.testSuccess') return 'Connected: {dims} dimensions';
+      if (key === 'settings.embeddings.testFailed') return 'Failed: {error}';
+      return key;
+    },
+  }),
+}));
 
 const provider = (over: Partial<EmbeddingProviderEntry> = {}): EmbeddingProviderEntry => ({
   slug: 'openai',
@@ -155,12 +172,17 @@ describe('EmbeddingsSetupModal — feedback', () => {
 
   it('reports a successful test with the dimensions the provider returned', () => {
     renderModal({ setupTestResult: result({ actual_dimensions: 1536 }) });
-    expect(screen.getByText('settings.embeddings.testSuccess')).toBeInTheDocument();
+    expect(screen.getByText('Connected: 1536 dimensions')).toBeInTheDocument();
   });
 
-  it('reports a failed test', () => {
+  it('falls back to ? when the provider reports no dimension count', () => {
+    renderModal({ setupTestResult: result({ actual_dimensions: undefined }) });
+    expect(screen.getByText('Connected: ? dimensions')).toBeInTheDocument();
+  });
+
+  it('reports a failed test with the provider error', () => {
     renderModal({ setupTestResult: result({ success: false, error: 'bad key' }) });
-    expect(screen.getByText('settings.embeddings.testFailed')).toBeInTheDocument();
+    expect(screen.getByText('Failed: bad key')).toBeInTheDocument();
   });
 
   it('surfaces a save error verbatim', () => {
@@ -168,7 +190,9 @@ describe('EmbeddingsSetupModal — feedback', () => {
     expect(screen.getByText('could not write config')).toBeInTheDocument();
   });
 
-  it('shows progress labels and locks both buttons while testing', () => {
+  // Testing locks the Test button only — Save stays live on purpose, so a user
+  // who has already pasted a working key is not blocked behind a slow probe.
+  it('locks Test but leaves Save available while a test is running', () => {
     renderModal({ setupKey: 'sk-live', setupTesting: true });
 
     expect(screen.getByRole('button', { name: 'settings.embeddings.testing' })).toBeDisabled();
--- a/app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
+++ b/app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
@@ -1,0 +1,191 @@
+/**
+ * `EmbeddingsSetupModal` had no tests. It is a pure presentational component —
+ * all state lives in the parent panel — so what is worth pinning is the
+ * branching it owns: the custom-endpoint form vs the API-key form, and the two
+ * disabled predicates on Test and Save. Those predicates are the ones a user
+ * runs into (a greyed-out Save with no explanation), and they differ between
+ * the two branches in ways that are easy to break.
+ */
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import type {
+  EmbeddingProviderEntry,
+  EmbeddingsTestResult,
+} from '../../../../services/api/embeddingsApi';
+import EmbeddingsSetupModal, { type EmbeddingsSetupModalProps } from '../EmbeddingsSetupModal';
+
+vi.mock('../../../../lib/i18n/I18nContext', () => ({ useT: () => ({ t: (key: string) => key }) }));
+
+const provider = (over: Partial<EmbeddingProviderEntry> = {}): EmbeddingProviderEntry => ({
+  slug: 'openai',
+  label: 'OpenAI',
+  description: 'OpenAI embeddings',
+  requires_api_key: true,
+  requires_endpoint: false,
+  has_api_key: false,
+  models: [],
+  ...over,
+});
+
+function renderModal(over: Partial<EmbeddingsSetupModalProps> = {}) {
+  const props: EmbeddingsSetupModalProps = {
+    setupProvider: provider(),
+    onClose: vi.fn(),
+    setupKey: '',
+    onSetupKeyChange: vi.fn(),
+    setupShowKey: false,
+    onToggleShowKey: vi.fn(),
+    setupTesting: false,
+    setupTestResult: null,
+    setupSaving: false,
+    setupError: '',
+    customEndpoint: '',
+    onCustomEndpointChange: vi.fn(),
+    customModel: '',
+    onCustomModelChange: vi.fn(),
+    customDims: '',
+    onCustomDimsChange: vi.fn(),
+    onTest: vi.fn(),
+    onSave: vi.fn(),
+    ...over,
+  };
+  render(<EmbeddingsSetupModal {...props} />);
+  return props;
+}
+
+const testButton = () => screen.getByRole('button', { name: 'settings.embeddings.testConnection' });
+const saveButton = () => screen.getByRole('button', { name: 'settings.embeddings.saveAndSwitch' });
+
+describe('EmbeddingsSetupModal — API-key branch', () => {
+  it('shows the provider description and hides the custom-endpoint fields', () => {
+    renderModal();
+
+    expect(screen.getByText('OpenAI embeddings')).toBeInTheDocument();
+    expect(screen.queryByPlaceholderText('https://your-endpoint.com/v1')).not.toBeInTheDocument();
+  });
+
+  it('disables Test and Save while the key box is blank', () => {
+    renderModal({ setupKey: '   ' });
+
+    expect(testButton()).toBeDisabled();
+    expect(saveButton()).toBeDisabled();
+  });
+
+  it('enables Test and Save once a key is typed', () => {
+    renderModal({ setupKey: 'sk-live' });
+
+    expect(testButton()).toBeEnabled();
+    expect(saveButton()).toBeEnabled();
+  });
+
+  it('allows Save with no key when the provider already has one stored', () => {
+    renderModal({ setupKey: '', setupProvider: provider({ has_api_key: true }) });
+
+    // Save is about switching provider, which needs no new key...
+    expect(saveButton()).toBeEnabled();
+    // ...but Test sends the key in the box, so it stays disabled.
+    expect(testButton()).toBeDisabled();
+  });
+
+  it('masks the key until the reveal toggle is used', () => {
+    const { onToggleShowKey } = renderModal({ setupKey: 'sk-live' });
+
+    expect(screen.getByDisplayValue('sk-live')).toHaveAttribute('type', 'password');
+    fireEvent.click(screen.getByRole('button', { name: 'settings.embeddings.show' }));
+    expect(onToggleShowKey).toHaveBeenCalledOnce();
+  });
+
+  it('reveals the key when the parent flips the flag', () => {
+    renderModal({ setupKey: 'sk-live', setupShowKey: true });
+
+    expect(screen.getByDisplayValue('sk-live')).toHaveAttribute('type', 'text');
+    expect(screen.getByRole('button', { name: 'settings.embeddings.hide' })).toBeInTheDocument();
+  });
+
+  it('calls onTest with the typed key', () => {
+    const { onTest } = renderModal({ setupKey: 'sk-live' });
+
+    fireEvent.click(testButton());
+    expect(onTest).toHaveBeenCalledOnce();
+  });
+});
+
+describe('EmbeddingsSetupModal — custom-endpoint branch', () => {
+  const custom = provider({ slug: 'custom', label: 'Custom', requires_endpoint: true });
+
+  it('swaps in the endpoint, model and dimensions fields', () => {
+    renderModal({ setupProvider: custom });
+
+    expect(screen.getByPlaceholderText('https://your-endpoint.com/v1')).toBeInTheDocument();
+    expect(screen.getByPlaceholderText('text-embedding-3-small')).toBeInTheDocument();
+    expect(screen.getByPlaceholderText('1024')).toBeInTheDocument();
+  });
+
+  it('gates Save on the endpoint rather than on the key', () => {
+    renderModal({ setupProvider: custom, customEndpoint: '  ', setupKey: 'sk-live' });
+    expect(saveButton()).toBeDisabled();
+  });
+
+  it('enables Save on an endpoint alone — the key is optional here', () => {
+    renderModal({ setupProvider: custom, customEndpoint: 'https://host/v1', setupKey: '' });
+    expect(saveButton()).toBeEnabled();
+  });
+
+  // The Test button is rendered enabled for a custom provider (the disabled
+  // predicate short-circuits on `!isCustom`), but its handler is wrapped in
+  // `if (!isCustom)`. So it looks actionable and does nothing. Pinned as the
+  // current behaviour — see the bug list; it is a UX defect, not a crash.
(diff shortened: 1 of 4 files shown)


app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  it('reports a successful test with the dimensions the provider returned', () => {
+    renderModal({ setupTestResult: result({ actual_dimensions: 1536 }) });
+    expect(screen.getByText('settings.embeddings.testSuccess')).toBeInTheDocument();
+  });
+
+  it('reports a failed test', () => {
+    renderModal({ setupTestResult: result({ success: false, error: 'bad key' }) });
+    expect(screen.getByText('settings.embeddings.testFailed')).toBeInTheDocument();
+  });
```
YellowSnnowmann, line 186: **minor:** Neither of these two cases can fail, and the first one's name promises the thing it does not check.

The i18n mock on `:18` is `t: (key: string) => key`, so `t('settings.embeddings.testSuccess')` returns the literal key — which contains no `{dims}` placeholder. The component does:

```tsx
// EmbeddingsSetupModal.tsx:191-195
t('settings.embeddings.testSuccess').replace('{dims}', String(setupTestResult.actual_dimensions ?? '?'))
: t('settings.embeddings.testFailed').replace('{error}', setupTestResult.error ?? '')
```

so `.replace()` is a no-op and the rendered text is exactly the key. `actual_dimensions: 1536`, the `?? '?'` fallback, and `actual_dimensions` being dropped from the component entirely all render identically — `getByText('settings.embeddings.testSuccess')` passes for all three. Same for `error: 'bad key'`: the string never reaches the DOM, so a component that stopped surfacing the provider's error message would keep this test green.

The real templates are `'Connected: {dims} dimensions'` and (in `en.ts:1669` and 13 sibling locales) the `{error}` equivalent, so the substitution is genuinely user-visible.

Concrete fix — return a template containing the placeholder for just these two keys, and assert the substituted output:

```tsx
vi.mock('../../../../lib/i18n/I18nContext', () => ({
  useT: () => ({
    t: (key: string) =>
      ({
        'settings.embeddings.testSuccess': 'Connected: {dims} dimensions',
        'settings.embeddings.testFailed': 'Failed: {error}',
      })[key] ?? key,
  }),
}));
```

then `expect(screen.getByText('Connected: 1536 dimensions')).toBeInTheDocument();` and `expect(screen.getByText('Failed: bad key')).toBeInTheDocument();`. Both then fail if the component drops the field.

app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    renderModal({ setupError: 'could not write config' });
+    expect(screen.getByText('could not write config')).toBeInTheDocument();
+  });
+
+  it('shows progress labels and locks both buttons while testing', () => {
+    renderModal({ setupKey: 'sk-live', setupTesting: true });
+
+    expect(screen.getByRole('button', { name: 'settings.embeddings.testing' })).toBeDisabled();
+    expect(saveButton()).toBeEnabled();
+  });
```
YellowSnnowmann: **nitpick:** The name says "locks **both** buttons while testing", but `:175` asserts Save is *enabled* — which is the correct, and more interesting, behaviour (you can switch provider while a test is in flight). The name will mislead the next person into "fixing" the assertion.

```suggestion
  it('shows the testing label and locks Test, leaving Save available', () => {
    renderModal({ setupKey: 'sk-live', setupTesting: true });

    expect(screen.getByRole('button', { name: 'settings.embeddings.testing' })).toBeDisabled();
    expect(saveButton()).toBeEnabled();
  });
```

YellowSnnowmann (commented):
## Summary

The Rust suite is the strongest single file in this stack. Refusing to satisfy `check-domain-e2e-coverage.mjs` by adding the string literal it greps for — and instead pinning `rpc_method_name() == "openhuman.inference_<fn>"` for all eleven controllers, the one contract nothing else asserted — is the right call, and the `EXPECTED_WIRE_METHODS` length + set assertions mean a new controller cannot slip in unpinned.

One real finding, in the component tests. Two of the `EmbeddingsSetupModal` feedback cases cannot fail.

### Blockers (0) / Major (0) / Minor (1) / Nitpicks (1)

### Verified / looks good

- `tests/raw_coverage/inference_local_controllers_rpc_raw_coverage_e2e.rs` is auto-registered — `build.rs:12-14` globs `tests/raw_coverage/` at build time precisely so a new file "is picked up automatically and cannot be silently skipped", and `raw_coverage_all`'s `required-features = ["voice", "inference"]` covers this suite's surface. No `[[test]]` stanza needed.
- Clobbering `PATH` to the stub dir alone matches the established convention in five sibling suites (`inference_local_ops_piper_…:102`, `inference_local_admin_…:118`, `inference_provider_admin_round22_…:314`, `inference_local_services_round21_…:113`, `inference_voice_http_round23_…:277`), and `SHARED_ENV_LOCK` serialises them.
- `ProviderSetupErrorNotice.test.tsx:141` looks like it could be vacuous under the `fallback ?? key` mock, but it is not: `ProviderSetupErrorNotice.tsx:115` calls `t('providerSetup.error.technicalDetails')` with **no** fallback, so the mock returns the key and `queryByText` is meaningful. Nice attention to detail.
- The offline discipline holds — `config.api_url = "http://127.0.0.1:9"` and every asserted branch returns before a transfer or socket.
- CI: 16 green, 3 pending Rust lanes, nothing PR-caused.

app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
M3gA-Mind: Fixed in `43cfc5c85` — you are right on both counts, and the first is the more serious of the two: the test name promised a check it never made.

With `t: (key) => key`, `t('settings.embeddings.testSuccess')` returned the literal key, which contains no `{dims}` placeholder, so the component's `.replace('{dims}', …)` was a no-op. `actual_dimensions: 1536`, the `?? '?'` fallback, and dropping the substitution from the component entirely all rendered identically. Same for `{error}` — `'bad key'` never reached the DOM.

The mock now returns the real `en.ts` templates for exactly those two keys and passes the key through for everything else, so the readable key-based assertions elsewhere are unaffected:

```ts
if (key === 'settings.embeddings.testSuccess') return 'Connected: {dims} dimensions';
if (key === 'settings.embeddings.testFailed') return 'Failed: {error}';
return key;
```

The two tests now assert `Connected: 1536 dimensions` and `Failed: bad key`, and I added the `?? '?'` fallback case, which was not expressible before.

Not executed: the fleet-wide no-local-builds rule took effect while I was working these threads, so this is reasoned from the source and CI is the verification.

app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
M3gA-Mind: Fixed in `43cfc5c85` — agreed, and the assertion is the half that was right.

Save staying enabled during a test is deliberate: a user who has already pasted a working key should not be blocked behind a slow probe. So the name was the defect, not the expectation. Renamed to "locks Test but leaves Save available while a test is running", with that reasoning in a comment above it so it does not read as an oversight later.

app/src/pages/dev/__tests__/AgentInsightsPreview.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    expect(running.some(e => e.status === 'running')).toBe(true);
+    expect(running.some(e => e.status === 'success')).toBe(true);
+    expect(running.some(e => e.subagent)).toBe(true);
+  });
+
+  it('settles every non-error entry to success and leaves errors alone', () => {
+    const { running, settled } = renderPreview();
+
+    settled.forEach((entry, i) => {
+      expect(entry.status).toBe(running[i].status === 'error' ? 'error' : 'success');
```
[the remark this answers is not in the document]
M3gA-Mind: Fixed in `43cfc5c85` — correct. The "leaves errors alone" half only exercises anything if the fixture actually contains an error entry, so removing that fixture or flipping it to `success` would have left the test green while it silently stopped checking preservation.

The test now asserts `running.some(e => e.status === 'error')` before the mapping check, and that an error entry survives into `settled`. I also added the same assertion to the fixture-variant test, which had the same hole.

The fixture does currently contain one (`AgentInsightsPreview.tsx:81`), so this is a guard against future drift rather than a fix for a present failure.

Not executed: the no-local-builds rule took effect mid-way through these threads, so this is reasoned from source and CI is the check.

tests/raw_coverage/inference_local_controllers_rpc_raw_coverage_e2e.rs
```diff
@@ -0,0 +1,628 @@
...
+    config.save().await.expect("save config");
+    let _workspace = EnvVarGuard::set("OPENHUMAN_WORKSPACE", tmp.path().join(".openhuman"));
+
+    let controllers = all_local_inference_registered_controllers();
+
+    // These three declare no inputs; a caller that sends some anyway (an older
+    // frontend, say) must not get a deserialization error.
+    for function in [
+        "assets_status",
+        "downloads_progress",
```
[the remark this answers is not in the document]
M3gA-Mind: Fixed in `43cfc5c85` — a good catch, and specific enough to act on directly. The suite's own header claimed "nothing here starts a server, opens a socket, or downloads an asset", which was not true.

You are right about the path: `assets_status` and `downloads_progress` both reach `LocalAiService::assets_status`, which probes `GET {ollama_base_url}/api/tags` on a two-second timeout, defaulting to `localhost:11434` and honouring the ambient `OLLAMA_HOST` (`ollama.rs:76-78`). So the "offline" suite made real socket calls and would behave differently on a developer machine running Ollama.

Both affected tests now pin it:

```rust
fn pin_offline_ollama() -> (EnvVarGuard, EnvVarGuard) {
    (
        EnvVarGuard::set("OPENHUMAN_OLLAMA_BASE_URL", "http://127.0.0.1:9"),
        EnvVarGuard::unset("OLLAMA_HOST"),
    )
}
```

Port 9 is the discard port — nothing listens, so `connect()` is refused immediately rather than waiting out the timeout, which makes it both offline and fast. `OPENHUMAN_OLLAMA_BASE_URL` is the app-specific override the resolver checks first, and clearing `OLLAMA_HOST` stops an ambient value taking over.

I also applied it to the `download_asset` test: every branch asserted there errors before reaching it, but that service path also ends in `assets_status`, so the guard is cheap insurance. The header no longer overclaims.

Not executed: the fleet-wide no-local-builds rule took effect while I was working these threads, so this is reasoned from source and CI is the verification.

app/src/components/settings/panels/__tests__/ProviderSetupErrorNotice.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    expect(summary).toBe(
+      'Could not reach Ollama. Check the endpoint URL and network connection, then try again.'
+    );
+  });
+
+  it('lifts the provider message out of a JSON error body', () => {
+    const { summary } = present(
+      'Could not reach OpenAI: {"error":{"message":"Incorrect API key provided: sk-abc","type":"invalid_request_error"}}'
+    );
+    expect(summary).toBe('Could not reach OpenAI: Incorrect API key provided: sk-abc');
```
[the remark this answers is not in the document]
M3gA-Mind: Partly acted on, partly pushed back — and I want to be explicit about which.

**The gap is real.** `presentProviderSetupError` performs no redaction: `findProviderJsonMessage` lifts the provider's `message` verbatim into `summary`, and `details` is the raw error. The backend redacts a fixed set of known token prefixes, so a custom provider echoing an arbitrary-shaped credential does reach the DOM intact. I verified that by reading the function rather than assuming.

**What I did:** added a characterisation test pinning exactly that, so the gap is visible and regression-locked instead of implied:

```ts
it('echoes an arbitrary provider credential verbatim (known gap)', () => {
  const secret = 'zz-corp-9f8e7d6c5b4a3210';
  const { summary, details } = present(
    `Could not reach CustomLLM: {"error":{"message":"Invalid token: ${secret}"}}`
  );
  expect(summary).toContain(secret);
  expect(details).toContain(secret);
});
```

with a comment saying it characterises current behaviour, not desired behaviour, and that the assertions should be inverted when redaction lands.

**What I did not do, and why:** adding redaction is a change to product behaviour in a test-only PR. It needs its own decision about which shapes to redact and it trades against diagnostics — the whole point of the `details` disclosure is that a user can read what the provider actually said, and over-redacting there makes a support conversation harder. That belongs in its own PR with its own review, not smuggled in here.

Recorded in `~/tinyhuman/bugs/W1-ui-bugs.md` so it is not lost. Resolving on that basis — happy to reopen if you would rather the redaction land here.

Scope note: the credential in question is the user's own key being shown back to them in their own settings panel, so the exposure is a screenshot/bug-report risk rather than an exfiltration path. That lowers the urgency but does not make it a non-issue.

app/src/components/settings/panels/__tests__/ProviderSetupErrorNotice.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  // fixed set of known token prefixes, so a CUSTOM provider echoing an
+  // arbitrary-shaped credential reaches the DOM intact.
+  //
+  // Pinned so the gap is visible and regression-locked rather than implied.
+  // Redacting at this boundary is a product change — it would also hide
+  // diagnostics a user may need — so it belongs in its own PR, not in a
+  // test-only one. Recorded in ~/tinyhuman/bugs/W1-ui-bugs.md. When redaction
+  // lands, invert these assertions.
+  it('echoes an arbitrary provider credential verbatim (known gap)', () => {
+    const secret = 'zz-corp-9f8e7d6c5b4a3210';
```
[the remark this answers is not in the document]
M3gA-Mind: Not a credential — a fabricated fixture, which your rule invites me to confirm.

The literal was a stand-in inside a characterisation test for the credential-echo gap on this same file. Nothing was rotated because nothing was real.

That said, the value looked plausible enough that you stopped on it, and a human reader could reasonably do the same. Fixed in `a580320d7`: renamed to `fakeCredential` and changed to `FIXTURE-not-a-real-key-000000`, which is self-describing while keeping the only property the test needs — a shape matching none of the token prefixes the backend redacts, which is precisely the case that slips through.

Resolving as a confirmed false positive.

tests/raw_coverage/inference_local_controllers_rpc_raw_coverage_e2e.rs
```diff
@@ -0,0 +1,657 @@
...
+///
+/// Both are hosted-proxy calls now, so the reachable offline surface is the
+/// validation that happens *before* any provider is created: params, the path
+/// trim, the unreadable-file branch and the empty-file branch.
+#[tokio::test]
+async fn inference_transcribe_controllers_cover_params_trimming_and_local_rejections() {
+    let _lock = env_lock();
+    let tmp = tempdir().expect("tempdir");
+    let mut config = temp_config(&tmp);
+    config.local_ai.runtime_enabled = true;
```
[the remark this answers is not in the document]
M3gA-Mind: Fixed in `aad1cffdd` — you are right, and this is the third vacuous assertion reviewers have found in these two PRs, so I want to state it plainly rather than bury it in a diff.

The test set `local_ai.runtime_enabled = true` at the top and never turned it off. With the runtime enabled, `"local ai is disabled"` could not appear whether or not the gate existed — so the assertion proved nothing, and a regression that gated hosted STT on the local runtime would have sailed straight through the one test written to catch it.

The runtime is now explicitly disabled and the config saved immediately before the hosted call:

```rust
let mut hosted_config = temp_config(&tmp);
hosted_config.local_ai.runtime_enabled = false;
hosted_config.save().await.expect("save config");
```

with a comment recording that the disabling is what gives the assertion teeth, so it does not read as redundant setup and get removed later.

The earlier part of the test keeps the runtime enabled deliberately — the unreadable-file and empty-file branches are local paths that need it — so only the hosted call runs against the disabled state.

Not executed: a fleet-wide no-local-builds rule is in force, so this is reasoned from source (`speech.rs:112-118` for the gate, `schemas.rs:340-352` for the handler) and verified only as far as `cargo fmt --check`. CI is the check.

app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
YellowSnnowmann: ✅ Resolved in `aad1cffd`. The mock now returns the real templates for `testSuccess`/`testFailed`, and the assertions read `Connected: 1536 dimensions` / `Failed: bad key`. You also added the `Connected: ? dimensions` fallback case, which I hadn't asked for and which is the one that pins `?? '?'`.

app/src/components/settings/panels/__tests__/EmbeddingsSetupModal.test.tsx
YellowSnnowmann: ✅ Resolved in `aad1cffd` — renamed to `locks Test but leaves Save available while a test is running`, which is what the assertions actually check.

YellowSnnowmann (dismissed):
## Re-review

Verified 2 prior findings: **2 resolved**, 0 partial, 0 still open. New issues: **0**.

- `EmbeddingsSetupModal.test.tsx` — the i18n mock now returns the real `{dims}` / `{error}` templates, and the assertions read `Connected: 1536 dimensions` and `Failed: bad key`. The added `Connected: ? dimensions` case pins the `?? '?'` fallback, which goes beyond what I asked for.
- The `while testing` case is renamed to match what it actually asserts.

Regression pass over `6180b1b6..aad1cffd` (+103 / −7 across 4 files): the extra assertions in `ProviderSetupErrorNotice`, `AgentInsightsPreview` and the Rust suite are additive and correct. CI: 14 green, 1 pending Rust coverage lane, no failures.

LGTM, Mergeable!

Issue #709, opened 2026-04-20
[Feature] Memory Architecture: Phase 3 - Summary trees, wiki indexing, and backlinks


senamakel:
## Summary

Design and implement the summary-tree and indexing phase for memory so OpenHuman can turn kept source chunks into a searchable, source-grounded wiki with hierarchical summaries at global, source, and topic scopes.

## Problem

Even after filtering, raw chunks are too large and too fragmented to serve directly as high-quality working memory. OpenHuman needs a compressed representation that preserves provenance, remains queryable, and can scale as memory volume grows.

Without a hierarchical summarization/indexing layer, the system will struggle to provide concise context windows, topic-aware retrieval, or efficient high-level reasoning over long histories.

Constraints:
- Summaries must remain linked to raw evidence.
- The structure should support incremental updates without expensive full-tree rebuilds.
- The design should separate lower-cost summarization work from higher-cost frontier-model planning/classification work.
- Retrieval should work across global context, source context, and major-topic context.

## Solution (optional)

Build a summary-tree architecture in which raw chunks are leaf nodes and progressively compressed parent summaries form a wiki-like memory structure.

Initial design targets:
- low-cost summarization of chunks into intermediate summaries with topic/entity labels
- a global “hot memory” tree representing the full knowledge base
- per-source summary trees for major ingestion sources
- dedicated topic/entity trees for the highest-value topics/entities selected by a stronger model or planner
- backlinks from summaries/entities to raw source chunks
- index structures that support retrieval by source, topic, entity, recency, and relevance

Treat source updates as append-and-replace operations so old documents can be tombstoned instead of forcing tree rewrites.

## Acceptance criteria

- [ ] A memory wiki/index model exists with raw chunks as leaves and summary nodes as higher-level parents.
- [ ] The system can generate summary nodes from kept chunks using a lower-cost summarization pass.
- [ ] Summary nodes retain metadata and backlinks to the underlying raw chunks.
- [ ] The system supports a global summary tree for cross-source hot memory.
- [ ] The system supports source-scoped summary trees.
- [ ] The system supports topic/entity-scoped summary trees for a selected set of important topics/entities.
- [ ] A higher-intelligence classification/planning pass can decide which topics/entities deserve dedicated trees.
- [ ] Indexing supports retrieval by topic/entity labels plus direct resolution back to source chunks.
- [ ] Update semantics are documented and implemented so modified documents can be replaced without expensive full-tree recalculation.
[...]

PR #5957, opened 2026-09-02
fix(devices): accept snake_case tunnel:register ACK shape from backend PR #709

### LINKED ISSUE #5871: Bug: Pair iPhone fails on initiation — "missing field 'channelId'" in tunnel register ack
## Description

The "Pair iPhone" beta feature fails immediately on initiation with a tunnel registration error. The pairing never completes.

**Reported by:** AntAttack (Discord)
**Build:** v0.63.17
**Platform:** MacBook M4, 24GB RAM

## Steps to reproduce

1. Open Settings → Devices
2. Click "Pair iPhone"
3. Pairing dialog appears and immediately fails

## Expected behaviour

Pairing flow initiates successfully and presents a QR code or pairing code for the iPhone to connect.

## Actual behaviour

Error dialog shown immediately:

> **Something went wrong**
> Failed to create pairing: [devices/tunnel] parse tunnel:register ack failed: missing field 'channelId'

"Try again" produces the same error.

## Error detail

```
Failed to create pairing: [devices/tunnel] parse tunnel:register ack failed: missing field 'channelId'
```

The tunnel server's register acknowledgement response is missing the `channelId` field that the client expects. Either:
- The server is returning a malformed/incomplete ack (field name changed or dropped)
- The client-side parser expects a field (`channelId`) that the server now returns under a different key
- The tunnel backend is not responding correctly to the register request at all

## Acceptance criteria

- [ ] Pair iPhone flow completes successfully on macOS (M4)
- [ ] Tunnel register ack includes `channelId` or client parser updated to match current server response schema
- [ ] If tunnel service is unavailable, error message explains this clearly (not a raw parse error)
- [ ] Confirmed working on v0.63.17+

YellowSnnowmann:
## Summary

- Added `#[serde(alias)]` attributes to `TunnelRegisterResponse` fields so both camelCase and snake_case ACK payloads are accepted during deserialization
- Backward-compatible: serialization behavior unchanged; old backend versions still work
- Parse-failure log now records the ACK's shape (field names for objects; kind for non-objects) — never raw values, which can contain `pairingToken`

## Problem

`tunnel:register` pairing was broken (#5871) for two reasons:

1. **Numeric `pairingExpiresAt`**: backend PR #709 typed this field as `number` (filled with `Date.getTime()`), but `TunnelRegisterResponse` declared it as `String`. A *successful* ACK could not deserialize at all — `invalid type: integer, expected a string`.

2. **Refusal envelope**: a failed register is answered with `{ ok: false, error: "..." }` and no `channelId`. Parsing that as the success shape reported `missing field 'channelId'` and threw away the backend's own explanation — which is the literal error in the issue.

(The snake_case aliases are additive forward-compatibility for a rename that has not happened; backend PR #709 *introduced* the camelCase shape, not replaced it. An earlier version of this description and the in-code comments had the history backwards, which is why the two actual defects went unnoticed while field names were the suspect.)

## Solution

- `pairingExpiresAt` accepts ISO 8601 string **or** epoch milliseconds via a custom deserializer, always yielding ISO 8601 (which the rest of the domain expects).
- A seconds-valued expiry is rejected loudly rather than silently decoding to 1970.
- `parse_register_ack` checks for the `{ ok: false, error }` refusal envelope first; a refused register surfaces the backend's reason, not a parse error.
- A non-bool or missing `ok` is handled correctly in both directions.
- A non-string `error` (e.g. a `{ code, message }` object) is rendered rather than reported as "unspecified error".
- Snake_case aliases are kept for forward-compat; camelCase (the current live shape) continues to work.

## Submission Checklist

- [x] **Diff coverage ≥ 80%** — changed lines (Vitest + cargo-llvm-cov merged via `diff-cover`) meet the gate enforced by [`.github/workflows/ci-lite.yml`](../.github/workflows/ci-lite.yml). Run `pnpm test:coverage` and `pnpm test:rust` locally; PRs below 80% on changed lines will not merge.
- [x] Coverage matrix updated — added/removed/renamed feature rows in [`docs/TEST-COVERAGE-MATRIX.md`](../docs/TEST-COVERAGE-MATRIX.md) reflect this change (or `N/A: behaviour-only change`)
- N/A: All affected feature IDs from the matrix are listed in the PR description under `## Related`
- N/A: Manual smoke checklist updated if this touches release-cut surfaces ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md))

## Impact

- Fixes iOS device pairing: both defects that caused #5871 are resolved
- No runtime/performance impact; serde alias resolution and the refusal check are zero-cost on the success path
- No migration required

## Related

- Closes #5871
- Follow-up PR(s)/TODOs: None

### DIFF
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -177,11 +177,11 @@ pub(crate) fn parse_register_ack(ack: serde_json::Value) -> Result<TunnelRegiste
         return Err(format!("[devices/tunnel] tunnel:register failed: {error}"));
     }
 
-    serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
-        log::error!(
-            "[devices/tunnel] parse tunnel:register ack failed: {e}; ack fields = {:?}",
-            ack.as_object().map(|o| o.keys().collect::<Vec<_>>())
-        );
+    // Describe the shape before `from_value` consumes the value, so the success
+    // path pays no clone and the failure path can still say what arrived.
+    let shape = describe_ack_shape(&ack);
+    serde_json::from_value::<TunnelRegisterResponse>(ack).map_err(|e| {
+        log::error!("[devices/tunnel] parse tunnel:register ack failed: {e}; ack was a {shape}");
         format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
     })
 }
@@ -194,16 +194,41 @@ pub(crate) fn parse_register_ack(ack: serde_json::Value) -> Result<TunnelRegiste
 /// unambiguously a refusal and its `error` is the only useful thing in it.
 pub(crate) fn backend_ack_error(ack: &serde_json::Value) -> Option<String> {
     let object = ack.as_object()?;
-    if object.get("ok")?.as_bool()? {
+    // Absent `ok` is the success shape. Anything else is a refusal — including
+    // a present-but-not-`true` value such as the string `"false"` or `0`. An
+    // earlier version asked `as_bool()?`, which yields `None` for those and so
+    // fell through to `missing field 'channelId'`: the exact bug this function
+    // exists to prevent, reintroduced for a backend that answers sloppily.
+    if object.get("ok")?.as_bool() == Some(true) {
         return None;
     }
-    Some(
-        object
-            .get("error")
-            .and_then(|e| e.as_str())
-            .unwrap_or("unspecified error")
-            .to_string(),
-    )
+    // A non-string `error` (say `{ code, message }`) is rendered rather than
+    // discarded — throwing away the backend's explanation is the failure mode
+    // being fixed, and it does not become acceptable because the shape
+    // surprised us.
+    Some(match object.get("error") {
+        None | Some(serde_json::Value::Null) => "unspecified error".to_string(),
+        Some(serde_json::Value::String(text)) => text.clone(),
+        Some(other) => other.to_string(),
+    })
+}
+
+/// A short description of an ACK's shape, for the parse-failure log.
+///
+/// `as_object()` is `None` for an array, a string or `null` — precisely the
+/// cases where "what actually arrived?" is the open question, and precisely
+/// where an earlier version logged the unhelpful `ack fields = None`.
+fn describe_ack_shape(ack: &serde_json::Value) -> String {
+    match ack {
+        serde_json::Value::Object(map) => {
+            format!("object with fields {:?}", map.keys().collect::<Vec<_>>())
+        }
+        serde_json::Value::Array(items) => format!("array of {} element(s)", items.len()),
+        serde_json::Value::String(_) => "string".to_string(),
+        serde_json::Value::Number(_) => "number".to_string(),
+        serde_json::Value::Bool(_) => "bool".to_string(),
+        serde_json::Value::Null => "null".to_string(),
+    }
 }
 
 /// Emit `tunnel:connect` to start listening on a channel as `role:"core"`.
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -25,9 +25,18 @@ pub struct TunnelRegisterPayload {
 
 /// Response from the `tunnel:register` ACK callback.
 ///
-/// Accepts both camelCase (historic backend shape) and snake_case (backend
-/// PR #709 shape) field names via `alias` so the client is forward- and
-/// backward-compatible without a coordinated deploy.
+/// The backend sends **camelCase**: backend PR #709 introduced
+/// `TunnelRegisterAck { channelId, pairingToken, pairingExpiresAt: number }`
+/// (`socketHandlers/tunnel/types.ts`) and `main` still emits exactly that. The
+/// `alias` entries are additive compatibility for a snake_case shape nothing
+/// currently sends — kept so a future rename needs no coordinated deploy, not
+/// because a rename has happened.
+///
+/// An earlier version of this comment had the history backwards, describing
+/// #709 as the move *to* snake_case. It is recorded here because the two real
+/// defects in this path — a numeric `pairingExpiresAt` decoded as a string, and
+/// a `{ ok: false }` refusal parsed as the success shape — were both missed
+/// while the field names were the suspect.
 #[derive(Debug, Clone, Deserialize)]
 pub struct TunnelRegisterResponse {
     #[serde(rename = "channelId", alias = "channel_id")]
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -47,6 +47,12 @@ pub struct TunnelRegisterResponse {
     pub pairing_expires_at: String,
 }
 
+/// Floor for a plausible epoch-milliseconds `pairingExpiresAt`: 2020-01-01Z.
+///
+/// Anything below it is far likelier to be seconds than a real expiry, and
+/// seconds decode silently into 1970 rather than failing.
+const MIN_PLAUSIBLE_EPOCH_MILLIS: i64 = 1_577_836_800_000;
+
 /// Accept `pairingExpiresAt` as either an ISO 8601 string or epoch
 /// milliseconds, always yielding the ISO 8601 string the rest of the domain
 /// expects.
@@ -73,9 +79,26 @@ where
 
     match Raw::deserialize(deserializer)? {
         Raw::Text(text) => Ok(text),
-        Raw::EpochMillis(millis) => chrono::DateTime::from_timestamp_millis(millis)
-            .map(|dt| dt.to_rfc3339())
-            .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}"))),
+        Raw::EpochMillis(millis) => {
+            // Reject a value that is not plausibly milliseconds. `getTime()`
+            // returns millis, but a backend that ever switched to seconds would
+            // hand over ~1.79e9, and `from_timestamp_millis` accepts that
+            // happily as 1970-01-21 — a pairing that is already expired before
+            // the QR is drawn, reported to the user as "expired" with nothing
+            // in the log to say why. Failing loudly here keeps the silent
+            // version of this PR's own bug from being reintroduced by a unit
+            // change upstream.
+            if millis < MIN_PLAUSIBLE_EPOCH_MILLIS {
+                return Err(D::Error::custom(format!(
+                    "pairingExpiresAt={millis} is too small to be epoch milliseconds \
+                     (expected > {MIN_PLAUSIBLE_EPOCH_MILLIS}); a seconds-valued expiry \
+                     would silently decode as 1970"
+                )));
+            }
+            chrono::DateTime::from_timestamp_millis(millis)
+                .map(|dt| dt.to_rfc3339_opts(chrono::SecondsFormat::Secs, true))
+                .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}")))
+        }
     }
 }
 
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -34,8 +34,49 @@ pub struct TunnelRegisterResponse {
     pub channel_id: String,
     #[serde(rename = "pairingToken", alias = "pairing_token")]
     pub pairing_token: String,
-    #[serde(rename = "pairingExpiresAt", alias = "pairing_expires_at")]
+    /// Deserialized through [`expires_at_from_string_or_epoch_millis`]: the
+    /// backend sends this as a JSON **number** (`pairingExpiresAt:
+    /// r.pairingExpiresAt.getTime()` in `socketHandlers/tunnel/handler.ts`),
+    /// while the field — and `PairingSession::expires_at` downstream — is
+    /// documented and consumed as an ISO 8601 string.
+    #[serde(
+        rename = "pairingExpiresAt",
+        alias = "pairing_expires_at",
+        deserialize_with = "expires_at_from_string_or_epoch_millis"
+    )]
     pub pairing_expires_at: String,
+}
+
+/// Accept `pairingExpiresAt` as either an ISO 8601 string or epoch
+/// milliseconds, always yielding the ISO 8601 string the rest of the domain
+/// expects.
+///
+/// The backend's `TunnelRegisterAck` types this field as `number` and fills it
+/// with `Date.getTime()`, so a plain `String` field cannot deserialize a
+/// successful ACK at all — it fails with `invalid type: integer, expected a
+/// string`. Widening to `i64` instead would push the epoch value into
+/// `PairingSession::expires_at` and `CreatePairingResponse::expires_at`, both
+/// documented as "ISO 8601 timestamp" and handed to the paired device, so the
+/// conversion happens here where the wire shape is known.
+fn expires_at_from_string_or_epoch_millis<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::de::Error as _;
+
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum Raw {
+        Text(String),
+        EpochMillis(i64),
+    }
+
+    match Raw::deserialize(deserializer)? {
+        Raw::Text(text) => Ok(text),
+        Raw::EpochMillis(millis) => chrono::DateTime::from_timestamp_millis(millis)
+            .map(|dt| dt.to_rfc3339())
+            .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}"))),
+    }
 }
 
 /// Payload emitted as `tunnel:connect` to join a channel.
@@ -91,6 +132,28 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
         .await
         .map_err(|e| format!("[devices/tunnel] emit tunnel:register failed: {e}"))?;
 
+    parse_register_ack(ack)
+}
+
+/// Turn a `tunnel:register` ACK into a response or an error message.
+///
+/// Split out of [`emit_register`] so the whole decision — failure envelope
+/// first, success shape second — is reachable without a live `SocketManager`.
+/// Testing only the [`backend_ack_error`] predicate would not have caught the
+/// envelope check being dropped from the call path, which is the regression
+/// that matters.
+pub(crate) fn parse_register_ack(ack: serde_json::Value) -> Result<TunnelRegisterResponse, String> {
+    // A failed register is answered with `{ ok: false, error: "..." }`
+    // (`safeAck<TunnelRegisterAck>(ack, { ok: false, error: publicError(err) })`
+    // in the backend handler), which carries no `channelId`. Parsing that as
+    // the success shape reports `missing field 'channelId'` and throws away the
+    // reason the backend gave — which is exactly what #5871 was: a register
+    // that failed server-side, surfaced to the user as a client parse error.
+    if let Some(error) = backend_ack_error(&ack) {
+        log::error!("[devices/tunnel] tunnel:register refused by backend: {error}");
+        return Err(format!("[devices/tunnel] tunnel:register failed: {error}"));
+    }
+
     serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
         log::error!(
             "[devices/tunnel] parse tunnel:register ack failed: {e}; ack fields = {:?}",
@@ -98,6 +161,26 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
         );
         format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
     })
+}
+
+/// The backend's error message when an ACK carries the `{ ok: false, error }`
+/// failure envelope, or `None` for anything else.
+///
+/// `ok` is only ever present on the failure path — a successful
+/// `TunnelRegisterAck` has no such field — so an ACK carrying `ok: false` is
+/// unambiguously a refusal and its `error` is the only useful thing in it.
+pub(crate) fn backend_ack_error(ack: &serde_json::Value) -> Option<String> {
+    let object = ack.as_object()?;
+    if object.get("ok")?.as_bool()? {
+        return None;
+    }
+    Some(
+        object
+            .get("error")
+            .and_then(|e| e.as_str())
+            .unwrap_or("unspecified error")
+            .to_string(),
+    )
 }
 
 /// Emit `tunnel:connect` to start listening on a channel as `role:"core"`.
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -92,10 +92,9 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
         .map_err(|e| format!("[devices/tunnel] emit tunnel:register failed: {e}"))?;
 
     serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
-        // Log the raw ACK so mismatches between backend field naming and this
-        // struct's rename/alias table are immediately visible in the logs.
         log::error!(
-            "[devices/tunnel] parse tunnel:register ack failed: {e}; raw ack = {ack}"
+            "[devices/tunnel] parse tunnel:register ack failed: {e}; ack fields = {:?}",
+            ack.as_object().map(|o| o.keys().collect::<Vec<_>>())
         );
         format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
     })
--- a/src/openhuman/security/devices/tunnel_client.rs
+++ b/src/openhuman/security/devices/tunnel_client.rs
@@ -24,13 +24,17 @@ pub struct TunnelRegisterPayload {
 }
 
 /// Response from the `tunnel:register` ACK callback.
+///
+/// Accepts both camelCase (historic backend shape) and snake_case (backend
+/// PR #709 shape) field names via `alias` so the client is forward- and
+/// backward-compatible without a coordinated deploy.
 #[derive(Debug, Clone, Deserialize)]
 pub struct TunnelRegisterResponse {
-    #[serde(rename = "channelId")]
+    #[serde(rename = "channelId", alias = "channel_id")]
     pub channel_id: String,
-    #[serde(rename = "pairingToken")]
+    #[serde(rename = "pairingToken", alias = "pairing_token")]
     pub pairing_token: String,
-    #[serde(rename = "pairingExpiresAt")]
+    #[serde(rename = "pairingExpiresAt", alias = "pairing_expires_at")]
     pub pairing_expires_at: String,
 }
 
@@ -87,8 +91,14 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
         .await
         .map_err(|e| format!("[devices/tunnel] emit tunnel:register failed: {e}"))?;
 
-    serde_json::from_value::<TunnelRegisterResponse>(ack)
-        .map_err(|e| format!("[devices/tunnel] parse tunnel:register ack failed: {e}"))
+    serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
+        // Log the raw ACK so mismatches between backend field naming and this
+        // struct's rename/alias table are immediately visible in the logs.
+        log::error!(
+            "[devices/tunnel] parse tunnel:register ack failed: {e}; raw ack = {ack}"
+        );
+        format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
+    })
 }
 
 /// Emit `tunnel:connect` to start listening on a channel as `role:"core"`.
--- a/src/openhuman/security/devices/tunnel_client_tests.rs
+++ b/src/openhuman/security/devices/tunnel_client_tests.rs
@@ -15,6 +15,23 @@ fn tunnel_register_response_accepts_backend_ack_shape_without_session_token() {
     assert_eq!(response.pairing_expires_at, "2026-06-30T15:00:00Z");
 }
 
+/// Regression for #5871: backend PR #709 switched the tunnel:register ACK
+/// from camelCase to snake_case. The struct must accept both shapes so the
+/// client works against old and new backend versions without a forced deploy.
+#[test]
+fn tunnel_register_response_accepts_snake_case_ack_shape() {
+    let response: TunnelRegisterResponse = serde_json::from_value(json!({
+        "channel_id": "ch_456",
+        "pairing_token": "pt_456",
+        "pairing_expires_at": "2026-09-30T12:00:00Z"
+    }))
+    .expect("snake_case register ack shape (backend PR #709) should parse");
+
+    assert_eq!(response.channel_id, "ch_456");
+    assert_eq!(response.pairing_token, "pt_456");
+    assert_eq!(response.pairing_expires_at, "2026-09-30T12:00:00Z");
+}
+
 #[test]
 fn build_core_connect_payload_omits_session_token_for_core_role() {
     let payload = build_core_connect_payload("ch_123");
--- a/src/openhuman/security/devices/tunnel_client_tests.rs
+++ b/src/openhuman/security/devices/tunnel_client_tests.rs
@@ -65,9 +65,36 @@ fn tunnel_register_response_accepts_numeric_pairing_expires_at() {
 
     assert_eq!(response.channel_id, "ch_789");
     assert_eq!(
-        response.pairing_expires_at, "2026-09-21T14:13:20+00:00",
+        response.pairing_expires_at, "2026-09-21T14:13:20Z",
         "epoch milliseconds must be rendered as the ISO 8601 string the rest \
          of the domain documents and forwards to the paired device"
+    );
+}
+
+/// A seconds-valued expiry must fail loudly rather than decode to 1970.
+///
+/// `from_timestamp_millis` accepts ~1.79e9 without complaint and yields
+/// 1970-01-21, so a backend that switched `getTime()` for a seconds-based
+/// clock would produce a pairing already expired before the QR is drawn — the
+/// user sees "expired" and the log says nothing. Raised by the review harness
+/// against the first version of this fix.
+#[test]
+fn a_seconds_valued_pairing_expiry_is_refused_instead_of_decoding_to_1970() {
+    let err = serde_json::from_value::<TunnelRegisterResponse>(json!({
+        "channelId": "ch_s",
+        "pairingToken": "pt_s",
+        "pairingExpiresAt": 1_790_000_000i64
+    }))
+    .expect_err("a seconds-valued expiry must not be accepted");
+
+    let text = err.to_string();
+    assert!(
+        text.contains("epoch milliseconds"),
+        "the error must name the unit mismatch, got: {text}"
+    );
+    assert!(
+        text.contains("1790000000"),
+        "the error must quote the offending value so the log identifies it, got: {text}"
     );
 }
 
--- a/src/openhuman/security/devices/tunnel_client_tests.rs
+++ b/src/openhuman/security/devices/tunnel_client_tests.rs
@@ -152,3 +152,62 @@ fn a_successful_ack_is_not_treated_as_a_failure_envelope() {
     let response = parse_register_ack(ack).expect("a successful ack must parse");
     assert_eq!(response.channel_id, "ch_1");
 }
+
+/// The out-of-range arm of the expiry conversion — the one branch this change
+/// added that had no test, and the one the failure-path requirement and the
+/// diff-coverage gate both land on.
+#[test]
+fn an_unrepresentable_pairing_expiry_is_refused() {
+    let err = serde_json::from_value::<TunnelRegisterResponse>(json!({
+        "channelId": "ch_x",
+        "pairingToken": "pt_x",
+        "pairingExpiresAt": i64::MAX
+    }))
+    .expect_err("a millisecond value no date can represent must not be accepted");
+
+    assert!(
+        err.to_string().contains("out of range"),
+        "the error must say the value is unrepresentable, got: {err}"
+    );
+}
+
+/// A refusal whose `ok` is present but not the boolean `true` is still a
+/// refusal. `as_bool()` yields `None` for `"false"`, so an earlier version read
+/// this as success and fell through to `missing field 'channelId'` — the bug
+/// this function exists to prevent.
+#[test]
+fn a_non_boolean_ok_is_still_treated_as_a_refusal() {
+    for ack in [
+        json!({ "ok": "false", "error": "nope" }),
+        json!({ "ok": 0, "error": "nope" }),
+        json!({ "ok": null, "error": "nope" }),
+    ] {
+        let err = parse_register_ack(ack.clone())
+            .expect_err("a non-boolean `ok` must not be read as success");
+        assert!(err.contains("nope"), "got: {err} for {ack}");
+    }
+
+    // …while an explicit `ok: true` is not a refusal.
+    assert!(backend_ack_error(&json!({ "ok": true })).is_none());
+}
+
+/// A non-string `error` is rendered, not discarded. Throwing away the backend's
+/// explanation is the failure mode being fixed; an unexpected shape does not
+/// make it acceptable.
+#[test]
+fn a_structured_error_payload_is_rendered_rather_than_dropped() {
+    let err = parse_register_ack(json!({
+        "ok": false,
+        "error": { "code": 429, "message": "channel limit" }
+    }))
+    .expect_err("a structured refusal is still a refusal");
+
+    assert!(
+        err.contains("channel limit") && err.contains("429"),
+        "the backend's payload must survive into the message, got: {err}"
+    );
+    assert!(
+        !err.contains("unspecified error"),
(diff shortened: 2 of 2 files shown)


src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -87,8 +91,14 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
         .await
         .map_err(|e| format!("[devices/tunnel] emit tunnel:register failed: {e}"))?;
 
-    serde_json::from_value::<TunnelRegisterResponse>(ack)
-        .map_err(|e| format!("[devices/tunnel] parse tunnel:register ack failed: {e}"))
+    serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
+        // Log the raw ACK so mismatches between backend field naming and this
+        // struct's rename/alias table are immediately visible in the logs.
+        log::error!(
+            "[devices/tunnel] parse tunnel:register ack failed: {e}; raw ack = {ack}"
```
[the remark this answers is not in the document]
YellowSnnowmann: Fixed in commit 40cc6b9d4 — the log now emits only the ACK field names via `ack.as_object().map(|o| o.keys().collect::<Vec<_>>())`, so credential values (`pairingToken`) never reach the log or Sentry regardless of which field is malformed.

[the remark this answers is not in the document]
YellowSnnowmann: Fixed in commit 40cc6b9d4 — now logs only `ack.as_object().map(|o| o.keys().collect::<Vec<_>>())` (field names, no values). The two-arg macro form also prevents rustfmt from collapsing the call to a single line, fixing the Rust Quality (fmt, clippy) CI failure.

M3gA-Mind:
Reviewed against `upstream/main` @ `6125f2510`. **CHANGES NEEDED** — the change is harmless, but I think the premise is wrong and the issue it closes is not fixed by it. Evidence below; I would rather be shown I've misread the backend than have this close #5871 silently.

## Blocking: backend #709 introduced camelCase, not snake_case

The PR body says #709 "changed the ACK payload from camelCase (`channelId`…) to snake_case (`channel_id`…)". Reading that PR and the backend's current `main`, it went the other way — #709 *introduced* the camelCase shape, and it is still what ships:

`tinyhumansai/backend` `src/socketHandlers/tunnel/types.ts:1-5` (on `main`):

```ts
export interface TunnelRegisterAck {
  channelId: string;
  pairingToken: string;
  pairingExpiresAt: number;
}
```

and `src/socketHandlers/tunnel/handler.ts:102-107`:

```ts
const resp: TunnelRegisterAck = {
  channelId: r.channelId,
  pairingToken: r.pairingToken,
  pairingExpiresAt: r.pairingExpiresAt.getTime(),
};
```

#709's own diff on those files adds `+ channelId: string;` and `+ pairingExpiresAt: number;`. There is no `channel_id` anywhere in it. So `alias = "channel_id"` accepts a shape nothing sends — additive and safe, but inert.

## Blocking: two real defects in this struct, and one of them is the reported symptom

**1. `pairingExpiresAt` is a number, the struct wants a `String`.** The backend sends `r.pairingExpiresAt.getTime()` — epoch milliseconds, typed `number`. `tunnel_client.rs:33-34` declares:

```rust
#[serde(rename = "pairingExpiresAt")]
pub pairing_expires_at: String,
```

There is no custom deserializer, so a **successful** register ACK cannot deserialize: serde reports `invalid type: integer, expected a string`. This is latent today only because register is failing before it gets there.

**2. The error ACK is parsed as a success shape, and that is exactly the error in #5871.** `handler.ts:110`:

```ts
safeAck<TunnelRegisterAck>(ack, { ok: false, error: publicError(err) });
```

`emit_register` hands that object straight to `from_value::<TunnelRegisterResponse>`. An object with no `channelId` yields **`missing field 'channelId'`** — verbatim the error in the issue title and body. So the register genuinely failed backend-side, and the client turned the backend's own explanation into a misleading parse error.

That reframes the issue: #5871's cause is upstream of field naming. A fix that closes it should (a) branch on `ok: false` and surface `error` to the user, and (b) accept a numeric `pairing_expires_at`. This PR does neither, and `Closes #5871` will mark it resolved while Pair iPhone still fails.

## Non-blocking

**The body describes a log line that would leak the pairing token; the code does not — the code is right.** Body: *"Improved parse-failure log to include the raw ACK value … `raw ack = {ack}`"*. The diff logs key names only:

```rust
ack.as_object().map(|o| o.keys().collect::<Vec<_>>())
```

That is the correct call — `pairing_token` is a bearer credential and does not belong in logs — and CodeRabbit's summary describes the code accurately. Please fix the body so nobody "restores" the described behaviour later. This log is also the most useful part of the PR: it would have printed `["ok", "error"]` and pointed straight at the real cause.

**`ack.clone()` runs on every register, not just the failure path.** Minor. `serde_json::from_value` takes ownership, so the clone is there to keep `ack` alive for the closure; `TunnelRegisterResponse::deserialize(&ack)` would avoid it.

**Two sibling inbound structs carry the same assumption** — `TunnelPeerStatus` (`:47`) and `TunnelFrame` (`:55`) are `Deserialize` with `rename = "channelId"` and no alias. Consistent with the backend as it stands, so nothing to do; noting it in case the alias approach is kept, since leaving them un-aliased makes the compatibility story half-applied.

## Coverage

**Would reverting the fix fail a test? Yes, but only for the inert half.** `tunnel_client_tests.rs::tunnel_register_response_accepts_snake_case_ack_shape` deserializes a snake_case payload and `.expect(...)`s it; drop the three `alias` attributes and it panics. That is a genuine revert-check for what the PR changes.

**Nothing tests the shape the backend actually sends.** The pre-existing `tunnel_register_response_accepts_backend_ack_shape_without_session_token` uses a *string* `pairingExpiresAt`, so it passes while the real numeric payload would fail — a fixture that has drifted from the contract it is named for. A test built from `TunnelRegisterAck` (numeric `pairingExpiresAt`) would fail today, and one feeding `{ok: false, error: "..."}` would show the misleading-error path. Either would have caught this before the field names were suspected.

Happy to be corrected if the deployed backend differs from `main` — I read the merged #709 and current `main`, not a running deployment.

M3gA-Mind:
Pushed one commit on top of yours — `3d4144a4` — implementing the two defects from my review. Your commits are untouched; nothing rebased, nothing force-rewritten. No approval implied.

**1. The backend sends `pairingExpiresAt` as a number, and the field is `String`.**
`TunnelRegisterAck` (backend `socketHandlers/tunnel/types.ts`) types it `number`, and `handler.ts:102-107` fills it with `Date.getTime()`. With no custom deserializer a **successful** ACK could not parse at all — `invalid type: integer, expected a string` — so pairing could not have completed even once register started succeeding. Now accepts either shape and normalises epoch millis to ISO 8601.

I chose normalising over widening the field to `i64` because `PairingSession::expires_at` and `CreatePairingResponse::expires_at` are both documented as ISO 8601 and go straight to the paired device. If you would rather carry epoch millis end to end, that is your call and I will not argue it — but it should be a deliberate change to those two types, not a silent one via this field.

**2. A refused register was reported as a client parse error.**
Failures are answered `{ ok: false, error }` with no `channelId`; feeding that to the success shape yields `missing field 'channelId'` — verbatim the error in #5871 — and discards the backend's own reason. Now checks the failure envelope first and returns the message.

**One structural change worth flagging:** I moved the ACK decision into `parse_register_ack` so it is reachable without a live `SocketManager`. That was not tidiness. My first version of these tests asserted on the `backend_ack_error` predicate alone, and the revert-check showed **they still passed with the check deleted from the call path** — they pinned the helper, not the behaviour. Testing through `parse_register_ack` fixes it.

**Revert-checked, each fix separately**, because one revert cannot exercise both:

| reverted | result |
|---|---|
| envelope check removed | 2 failures, each naming its own assertion |
| deserializer detached | 2 failures, including the successful-ack case |
| baseline / restored | 7 passed, 0 failed |

**Two things left for you, both in the PR body rather than the code:**

- **The stated premise looks inverted.** Backend #709 *introduced* the camelCase shape — its own diff adds `+ channelId: string;` and `+ pairingExpiresAt: number;`, and `main` still sends camelCase. So the snake_case aliases match a shape nothing currently emits. I have **kept** them (additive, harmless, and forward-compatible if the backend ever moves), but the body should not claim they fix #5871, because the two defects above are what did.
- **The body describes logging `raw ack = {ack}`** — that would put the `pairing_token` bearer credential in logs. Your code correctly logs key names only. Worth fixing the text so nobody "restores" the described behaviour later.

Happy to drop the aliases if you would rather keep the diff minimal — say the word and I will.

M3gA-Mind (commented):
# PR #5957 — fix(devices): accept snake_case tunnel:register ACK shape from backend PR #709

## Walkthrough

This PR fixes #5871 (`parse tunnel:register ack failed: missing field 'channelId'` on "Pair iPhone") by making `TunnelRegisterResponse` tolerant of the three wire shapes the backend can actually produce: snake_case field names, a numeric `pairingExpiresAt` (`Date.getTime()`), and the `{ ok: false, error }` refusal envelope — the last being routed to a real error message instead of a misleading serde parse failure. The parse decision is extracted into `parse_register_ack` so it is testable without a live `SocketManager`, and seven unit tests cover both success shapes, the numeric expiry, and three refusal cases.

The approach is sound and the reasoning is unusually well documented in-code. I verified the suite locally: `cargo test --lib --features "$(bash scripts/ci/product-features.sh)" security::devices::tunnel_client` → **7 passed, 0 failed**. I also traced the downstream consumer and confirmed the epoch→ISO normalisation is the right call (see *Verified* below). No blockers. The findings below are one defensive-hardening item and a set of diagnostics/polish suggestions — several of which matter because *diagnosability* is the point of this change.

## Changes

| File | Summary |
| --- | --- |
| `src/openhuman/security/devices/tunnel_client.rs` | Adds snake_case `alias`es to `TunnelRegisterResponse`; adds `expires_at_from_string_or_epoch_millis` (accepts ISO 8601 string **or** epoch millis, always yields ISO 8601); extracts `parse_register_ack` + `backend_ack_error` so a `{ ok:false, error }` refusal surfaces the backend's reason instead of `missing field 'channelId'`; logs ACK key names on parse failure. |
| `src/openhuman/security/devices/tunnel_client_tests.rs` | +5 tests: snake_case shape, numeric `pairingExpiresAt` → ISO, refusal envelope with and without a message, and a success ACK not being mistaken for a refusal. |

## Actionable comments (6)

### 🛑 Blockers

None.

### ⚠️ Major

#### 1. `src/openhuman/security/devices/tunnel_client.rs:76-78` — epoch **seconds** would be silently read as millis and produce a 1970 expiry

`from_timestamp_millis` cannot distinguish `1790000000` (seconds) from `1790000000000` (millis) — it happily returns `1970-01-21T20:53:20Z` for the former. Because this function's whole premise is "we are not certain what the backend sends", that is a realistic input, and the failure is silent and downstream: `PairPhoneModal.tsx:69` turns it into `exp` on the QR, `PairScreen.tsx:122` compares `payload.expiresAt < nowSecs`, and the user gets *"QR expired"* with no error anywhere in the logs. One range check turns that into a loud, greppable failure at the seam where the wire shape is known.

**Suggested change:**
```rust
// before
Raw::EpochMillis(millis) => chrono::DateTime::from_timestamp_millis(millis)
    .map(|dt| dt.to_rfc3339())
    .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}"))),

// after
Raw::EpochMillis(millis) => {
    // A pairing TTL is minutes away, so anything before 2001 is epoch
    // *seconds* mislabelled as millis — reject loudly rather than hand a
    // 1970 expiry to the QR, where it reads as "QR expired".
    if millis < 1_000_000_000_000 {
        return Err(D::Error::custom(format!(
            "pairingExpiresAt looks like epoch seconds, not milliseconds: {millis}"
        )));
    }
    chrono::DateTime::from_timestamp_millis(millis)
        .map(|dt| dt.to_rfc3339_opts(chrono::SecondsFormat::Secs, true))
        .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}")))
}
```

(The `to_rfc3339_opts(_, true)` part also addresses item 4 below.)

### 💡 Refactor / suggestion

#### 2. `src/openhuman/security/devices/tunnel_client.rs:157-163` — the new diagnostic logs `None` for exactly the ACK shapes it can't explain, and clones on the success path

`ack.as_object()` is `None` for an array, a string, `null` — precisely the cases where "which fields arrived?" is the question. The log then reads `ack fields = None`, which carries no information at all, and the user still sees a raw serde message (`invalid type: sequence, …`), which is what issue #5871's third acceptance criterion asks you to stop doing. Note this is reachable: `parse_sio_ack` (`ws_loop_part_02.rs:126-131`) only unwraps to a bare object when the ACK has **exactly one** argument — a Node-style `ack(err, data)` callback yields a `Value::Array`. The same restructure removes the `ack.clone()`, which today runs on every successful register too.

**Suggested change:**
```rust
// before
serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
    log::error!(
        "[devices/tunnel] parse tunnel:register ack failed: {e}; ack fields = {:?}",
        ack.as_object().map(|o| o.keys().collect::<Vec<_>>())
    );
    format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
})

// after
let shape = match &ack {
    serde_json::Value::Object(map) => format!("object fields = {:?}", map.keys().collect::<Vec<_>>()),
    // Never log values: a successful ack carries `pairingToken`.
    other => format!("non-object ack of kind {}", kind_of(other)),
};
serde_json::from_value::<TunnelRegisterResponse>(ack).map_err(|e| {
    log::error!("[devices/tunnel] parse tunnel:register ack failed: {e}; {shape}");
    format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
})
```

#### 3. `src/openhuman/security/devices/tunnel_client.rs:177-183` — a non-string `error` is discarded as `"unspecified error"`

If the backend ever answers `{ ok: false, error: { code, message } }` (or a number), `as_str()` returns `None` and the branch reports `unspecified error` — throwing away the explanation, which is the exact failure mode this PR exists to fix, one level down. A compact-JSON fallback keeps the payload without widening the type.

**Suggested change:**
```rust
// before
object
    .get("error")
    .and_then(|e| e.as_str())
    .unwrap_or("unspecified error")
    .to_string(),

// after
match object.get("error") {
    Some(serde_json::Value::String(s)) => s.clone(),
    Some(other) => other.to_string(),   // `{"code":…}` beats "unspecified error"
    None => "unspecified error".to_string(),
},
```

#### 4. `src/openhuman/security/devices/tunnel_client.rs:77` — `to_rfc3339()` emits `+00:00` while every other path emits `Z`

The string passthrough keeps whatever the backend sent (`…T15:00:00Z` in the existing test), so the same logical field reaches the paired device in two different renderings depending on which branch produced it. Harmless for today's consumers — I checked, `new Date("2026-09-21T14:13:20+00:00")` parses fine — but there is no reason to have two. `to_rfc3339_opts(SecondsFormat::Secs, true)` yields `2026-09-21T14:13:20Z`. Folded into item 1's snippet.

#### 5. `src/openhuman/security/devices/tunnel_client_tests.rs:72` — the only uncovered changed lines are the expiry-conversion failure path

Every branch added here has a test except `expires_at_from_string_or_epoch_millis`'s error arm (`tunnel_client.rs:78`). The repo's failure-path requirement and the ≥80% diff-coverage gate both point at it, and it is three lines.

**Suggested addition:**
```rust
/// An expiry the conversion cannot represent must fail by name, not silently
/// become a nonsense timestamp handed to the paired device.
#[test]
fn an_unrepresentable_pairing_expires_at_is_rejected() {
    let err = serde_json::from_value::<TunnelRegisterResponse>(json!({
        "channelId": "ch_bad",
        "pairingToken": "pt_bad",
        "pairingExpiresAt": i64::MAX
    }))
    .expect_err("an out-of-range epoch must not parse");
    assert!(err.to_string().contains("pairingExpiresAt"), "got: {err}");
}
```

#### 6. `src/openhuman/security/devices/README.md:109` — still documents only the camelCase shape

> `tunnel:register` uses `SocketManager::emit_with_ack` and expects backend ACK shape `{channelId, pairingToken, pairingExpiresAt}` with a 10-second timeout.

That line is now three shapes out of date, and this README is the domain's stated map. Per *Docs with code* in `AGENTS.md`:

```markdown
- `tunnel:register` uses `SocketManager::emit_with_ack` with a 10-second timeout. The ACK is
  accepted in camelCase (`channelId`) or snake_case (`channel_id`), and `pairingExpiresAt` may be
  an ISO 8601 string or epoch milliseconds (normalised to ISO 8601 in `tunnel_client.rs`). A
  `{ ok: false, error }` refusal envelope is surfaced as the backend's error, not a parse failure.
```

## Nitpicks (4)

- `src/openhuman/security/devices/tunnel_client.rs:145,172` — `parse_register_ack` / `backend_ack_error` are only reached from this file and from `mod tests`, which is a child module and can see private items. `fn` rather than `pub(crate) fn` keeps them out of the crate's surface.
- `src/openhuman/security/devices/tunnel_client.rs:174` — `object.get("ok")?.as_bool()?` treats a non-bool `ok` (say the string `"false"`) as *not a refusal*, which then falls through to a `missing field 'channelId'`. Given the rest of the function is written defensively, `!= Some(true)` on the bool would be more consistent.
- `src/openhuman/security/devices/tunnel_client.rs:67-72` — when neither variant matches, the untagged enum produces `data did not match any variant of untagged enum Raw`, which names an internal type and no field. A `serde_json::Value` match would let you say `pairingExpiresAt must be an ISO 8601 string or epoch milliseconds, got: <kind>`.
- **PR description** — it says the log was updated to include `raw ack = {ack}`, but the code logs key *names* only. The code is right and the description is stale: a successful ACK carries `pairingToken`, so logging the raw value would put a live pairing credential in the log file. Worth correcting the description so nobody "restores" it.

## Questions for the author (2)

- The title and description say backend PR #709 moved the ACK to **snake_case**, but the doc comments added in this same PR say the live backend sends **camelCase** with a numeric `pairingExpiresAt` (`pairingExpiresAt: r.pairingExpiresAt.getTime()`). Those can't both be the current shape. Which one did you observe on the wire, and is the snake_case alias defensive-only? Accepting both is harmless, but the in-code comments will be read as ground truth later.
- `backend_ack_error`'s contract is "a successful `TunnelRegisterAck` has no `ok` field". Have you confirmed the backend never answers `{ ok: true, ...fields }` **or** `{ ok: true, data: { channelId, … } }`? The first is handled correctly today; the second would pass the refusal check and then fail with `missing field 'channelId'` — i.e. the original bug, unchanged.

## Verified / looks good

- **The epoch→ISO normalisation is the correct choice, not just a stylistic one.** `PairPhoneModal.tsx:69` does `new Date(session.expires_at).getTime()`; had the epoch been passed through as a decimal string, that would be `NaN` and the QR's `exp` would be `NaN`, which `PairScreen.tsx:70` rejects as an invalid QR. The comment explaining the decision matches what the consumer actually does.
- **No credential leakage in the new logs.** Only ACK key names are logged; `pairingToken` never reaches the log line.
- **The object assumption in `parse_register_ack` matches the transport** for the single-argument case: `parse_sio_ack` (`ws_loop_part_02.rs:126-131`) pops a lone argument out of the Socket.IO args array before resolving the waiter.
- **Testing the decision path, not just the predicate**, in `register_failure_envelope_surfaces_the_backend_error_not_a_parse_error` — including the sanity assertion that the old path really did produce `channelId` — is the right shape for a regression test, and the comment explains why.
- `#[serde(alias)]` is deserialize-only and the struct is `Deserialize`-only, so the backward-compatibility claim holds exactly.
- Logging follows `AGENTS.md`: greppable `[devices/tunnel]` prefixes, `log::error!` on both refusal and parse-failure branches.
- Suite verified locally under the product feature set: 7 passed, 0 failed.

src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -24,16 +24,61 @@ pub struct TunnelRegisterPayload {
...
+    enum Raw {
+        Text(String),
+        EpochMillis(i64),
+    }
+
+    match Raw::deserialize(deserializer)? {
+        Raw::Text(text) => Ok(text),
+        Raw::EpochMillis(millis) => chrono::DateTime::from_timestamp_millis(millis)
+            .map(|dt| dt.to_rfc3339())
+            .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}"))),
```
M3gA-Mind: **⚠️ Major — epoch _seconds_ would be silently read as millis and produce a 1970 expiry.**

`from_timestamp_millis` cannot distinguish `1790000000` (seconds) from `1790000000000` (millis) — it happily returns `1970-01-21T20:53:20Z` for the former. Since this function exists precisely because the wire shape is uncertain, that is a realistic input, and the failure is silent and downstream: `PairPhoneModal.tsx:69` turns it into the QR's `exp`, `PairScreen.tsx:122` compares `payload.expiresAt < nowSecs`, and the user sees *"QR expired"* with nothing in the logs.

```suggestion
        Raw::EpochMillis(millis) => {
            // A pairing TTL is minutes away, so anything before 2001 is epoch
            // *seconds* mislabelled as millis — reject loudly rather than hand
            // a 1970 expiry to the QR, where it reads as "QR expired".
            if millis < 1_000_000_000_000 {
                return Err(D::Error::custom(format!(
                    "pairingExpiresAt looks like epoch seconds, not milliseconds: {millis}"
                )));
            }
            chrono::DateTime::from_timestamp_millis(millis)
                .map(|dt| dt.to_rfc3339_opts(chrono::SecondsFormat::Secs, true))
                .ok_or_else(|| D::Error::custom(format!("pairingExpiresAt out of range: {millis}")))
        }
```

The `to_rfc3339_opts(_, true)` also fixes a second, smaller thing: plain `to_rfc3339()` emits `+00:00`, while the string-passthrough branch keeps the backend's `Z` (see the existing test at `tunnel_client_tests.rs:15`). Same logical field, two renderings, for no reason. Harmless for today's consumer — `new Date("…+00:00")` parses fine — but worth collapsing to one.

src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -87,8 +132,55 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
...
+    // in the backend handler), which carries no `channelId`. Parsing that as
+    // the success shape reports `missing field 'channelId'` and throws away the
+    // reason the backend gave — which is exactly what #5871 was: a register
+    // that failed server-side, surfaced to the user as a client parse error.
+    if let Some(error) = backend_ack_error(&ack) {
+        log::error!("[devices/tunnel] tunnel:register refused by backend: {error}");
+        return Err(format!("[devices/tunnel] tunnel:register failed: {error}"));
+    }
+
+    serde_json::from_value::<TunnelRegisterResponse>(ack.clone()).map_err(|e| {
```
M3gA-Mind: **💡 The new diagnostic logs `None` for exactly the shapes it can't explain — and clones on the success path.**

`ack.as_object()` is `None` for an array, a string or `null`, which are precisely the cases where "which fields arrived?" is the open question. The log then reads `ack fields = None` (no information at all) and the user still gets a raw serde message — which is what #5871's third acceptance criterion asks you to stop doing.

This is reachable, not hypothetical: `parse_sio_ack` (`ws_loop_part_02.rs:126-131`) only unwraps to a bare object when the ACK carries **exactly one** argument; a Node-style `ack(err, data)` callback resolves the waiter with a `Value::Array`.

The same restructure also drops `ack.clone()`, which today runs on every *successful* register:

```rust
let shape = match &ack {
    serde_json::Value::Object(map) => {
        format!("object fields = {:?}", map.keys().collect::<Vec<_>>())
    }
    // Never log values: a successful ack carries `pairingToken`.
    serde_json::Value::Array(items) => format!("array ack with {} element(s)", items.len()),
    other => format!("non-object ack: {}", kind_of(other)),
};
serde_json::from_value::<TunnelRegisterResponse>(ack).map_err(|e| {
    log::error!("[devices/tunnel] parse tunnel:register ack failed: {e}; {shape}");
    format!("[devices/tunnel] parse tunnel:register ack failed: {e}")
})
```

src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -87,8 +132,55 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
...
+
+/// The backend's error message when an ACK carries the `{ ok: false, error }`
+/// failure envelope, or `None` for anything else.
+///
+/// `ok` is only ever present on the failure path — a successful
+/// `TunnelRegisterAck` has no such field — so an ACK carrying `ok: false` is
+/// unambiguously a refusal and its `error` is the only useful thing in it.
+pub(crate) fn backend_ack_error(ack: &serde_json::Value) -> Option<String> {
+    let object = ack.as_object()?;
+    if object.get("ok")?.as_bool()? {
```
M3gA-Mind: **Nit — a non-bool `ok` is treated as "not a refusal".**

`object.get("ok")?.as_bool()?` returns `None` (⇒ success) when `ok` is present but, say, the string `"false"`, which then falls through to `missing field 'channelId'` — the original bug. Given how defensively the rest of this function is written, matching on the bool would be more consistent:

```suggestion
    if object.get("ok").and_then(|v| v.as_bool()) != Some(false) {
        return None;
    }
```

(That also keeps `{ "ok": true }` and an absent `ok` on the success path, which the tests already pin.)

src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -87,8 +132,55 @@ pub async fn emit_register() -> Result<TunnelRegisterResponse, String> {
...
+        return None;
+    }
+    Some(
+        object
+            .get("error")
+            .and_then(|e| e.as_str())
```
M3gA-Mind: **💡 A non-string `error` is discarded as `"unspecified error"`.**

If the backend ever answers `{ ok: false, error: { code, message } }` (or a number), `as_str()` yields `None` and the caller is told `unspecified error` — throwing away the backend's explanation, which is the exact failure mode this PR exists to fix, one level down. A compact-JSON fallback keeps the payload without widening the type:

```suggestion
    Some(match object.get("error") {
        Some(serde_json::Value::String(s)) => s.clone(),
        // `{"code":…}` beats "unspecified error".
        Some(other) => other.to_string(),
        None => "unspecified error".to_string(),
    })
```

src/openhuman/security/devices/tunnel_client_tests.rs
```diff
@@ -24,3 +41,87 @@ fn build_core_connect_payload_omits_session_token_for_core_role() {
...
+    }))
+    .expect("the numeric pairingExpiresAt the backend sends should parse");
+
+    assert_eq!(response.channel_id, "ch_789");
+    assert_eq!(
+        response.pairing_expires_at, "2026-09-21T14:13:20+00:00",
+        "epoch milliseconds must be rendered as the ISO 8601 string the rest \
+         of the domain documents and forwards to the paired device"
+    );
+}
```
M3gA-Mind, line 72: **💡 The one uncovered branch in this PR is the expiry-conversion failure path.**

Every branch added here has a test except `expires_at_from_string_or_epoch_millis`'s error arm (`tunnel_client.rs:78`). The repo's failure-path requirement and the ≥80% diff-coverage gate both land on it, and it costs three lines:

```rust
/// An expiry the conversion cannot represent must fail by name, not silently
/// become a nonsense timestamp handed to the paired device.
#[test]
fn an_unrepresentable_pairing_expires_at_is_rejected() {
    let err = serde_json::from_value::<TunnelRegisterResponse>(json!({
        "channelId": "ch_bad",
        "pairingToken": "pt_bad",
        "pairingExpiresAt": i64::MAX
    }))
    .expect_err("an out-of-range epoch must not parse");
    assert!(err.to_string().contains("pairingExpiresAt"), "got: {err}");
}
```

Related, on the assertion just above at line 68: it pins `"2026-09-21T14:13:20+00:00"`. If you take the `to_rfc3339_opts(SecondsFormat::Secs, true)` suggestion on the conversion function, this becomes `"2026-09-21T14:13:20Z"` — which also lines up with the `Z` form the sibling test at line 15 already expects.

M3gA-Mind:
Pushed `3308d7cb` — one more fix, from running the repo review harness against this PR. It found a real hole in **my own** previous commit, so this is worth stating plainly rather than burying.

**`from_timestamp_millis` cannot tell milliseconds from seconds.** If a backend ever swapped `getTime()` for a seconds-based clock it would send ~`1.79e9`, which decodes without complaint to **1970-01-21** — a pairing already expired before the QR is drawn, surfacing to the user as "expired" with nothing in the log to explain it. That is a *silent* version of exactly the bug this PR exists to fix, and my numeric-expiry fix introduced the opening for it.

Now rejected below a 2020-01-01Z floor, with the offending value quoted in the error. Also emits `Z` rather than `+00:00`, so both paths into `pairing_expires_at` render the same way.

**Revert-checked:** with the guard removed the new test fails and the value produced is exactly `"1970-01-21T17:13:20Z"` — the predicted failure mode. Restored, 8/8 pass.

Worth admitting: my first draft of that test asserted the error text did *not* contain `1970`, while the error message itself says "would silently decode as 1970". Self-defeating, and the baseline run caught it. It now checks the offending value is quoted instead.

**Harness findings I have not acted on**, for you to take or leave — all diagnostics polish rather than correctness:

- the `ack fields` log prints `None` for a non-object ACK (reachable, since `parse_sio_ack` only unwraps single-argument acks);
- a non-string `error` in the refusal envelope collapses to `"unspecified error"` rather than being rendered;
- the domain README still documents only the camelCase shape.

**And two on the description, unchanged from my earlier note** — these now matter more because the in-code comments say the opposite: the title says #709 moved to snake_case, while the comments (and the backend on `main`) say camelCase with a numeric expiry. Both cannot be current, and the comments will be read as ground truth later. The description also still claims the log includes `raw ack = {ack}`; it does not, and should not — a successful ACK carries `pairingToken`.

No approval yet: CI Lite is still in progress on this head. I will re-check the full gate once the heavy lanes report.

src/openhuman/security/devices/tunnel_client_tests.rs
M3gA-Mind: Added `an_unrepresentable_pairing_expiry_is_refused` in `36b1e7ac`'s parent — it feeds `i64::MAX` and asserts the error says `out of range`. That was the one branch these commits added without a test, so thank you for catching it. Revert-checked along with the other two behaviours: baseline 11/11, and each reverted behaviour fails its own named test.

src/openhuman/security/devices/tunnel_client.rs
```diff
@@ -24,16 +24,84 @@ pub struct TunnelRegisterPayload {
 }
 
 /// Response from the `tunnel:register` ACK callback.
+///
+/// Accepts both camelCase (historic backend shape) and snake_case (backend
+/// PR #709 shape) field names via `alias` so the client is forward- and
+/// backward-compatible without a coordinated deploy.
```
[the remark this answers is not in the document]
M3gA-Mind: Fixed in `36b1e7ac`. You are right and it matches what I found reviewing this PR: #709 *introduced* the camelCase shape (its own diff adds `+ channelId: string;` and `+ pairingExpiresAt: number;`) and `main` still emits it. The comment now says the aliases are additive compatibility for a shape nothing currently sends, and records that the history was stated backwards — because that misdirection is why the two real defects here (numeric expiry, refusal envelope) went unnoticed while the field names were the suspect.

M3gA-Mind (approved):
Approved after a maintainer verification pass on head `36b1e7ac6`.

Verified: `MERGEABLE`; 26 check-runs, **0 failing, 0 incomplete**; `Rust Quality`, `Rust Core Coverage` and `PR CI Gate` pass. `Frontend Checks` shows `skipping`, which is correct here — the PR touches only `src/openhuman/security/devices/tunnel_client.rs` and its test file, no frontend code.

No unresolved, non-outdated review threads; the one thread resolved under our account carries a substantive reply (the added `an_unrepresentable_pairing_expiry_is_refused` coverage feeding `i64::MAX`). No AI attribution.

M3gA-Mind (approved):
Approving on the amended gate, verified against head \`36b1e7ac6\`:

| gate | value |
|---|---|
| 1. mergeable | `MERGEABLE` |
| 2. failing checks | **0** |
| 3. pending checks | **0** (total **27** present — not the low-teens shape of a PR whose lanes never reported) |
| 3a. runs still queued/in_progress for this head | **0** — `CI Lite=success`, `PR Quality (soft)=success` |
| 3b. heavy lanes by name | `PR CI Gate` pass · `Rust Core Coverage` pass · `Rust Quality` pass · `Frontend Checks` **skipping** |
| 4. unresolved non-outdated threads | **0** — the one resolved under our account carries a reply naming the commit and the test added |
| 5. AI attribution | **0** in every commit trailer, every commit body, and the PR description |

**On `Frontend Checks: skipping`** — checked rather than assumed. The PR's file list is exactly two files, `tunnel_client.rs` and `tunnel_client_tests.rs`, with zero `app/` or `.ts(x)` content. The skip is correct, and all three Rust-relevant heavy lanes ran and passed.

**What the PR now does, which is not what it originally claimed.** It began as snake_case aliases for a shape backend #709 was said to have introduced. #709 in fact introduced *camelCase* with a numeric `pairingExpiresAt`, and `main` still emits that — so the aliases were inert. The two defects that actually broke pairing were found in review and fixed here:

- `pairingExpiresAt` arrives as a JSON **number**; the field was `String` with no deserializer, so a *successful* ACK could not parse at all;
- a refused register answers `{ ok: false, error }` with no `channelId`, and parsing it as the success shape produced `missing field 'channelId'` — verbatim the error in #5871 — discarding the backend's own explanation.

Both fixed, then hardened after review: a seconds-valued expiry is refused instead of silently decoding to 1970; any non-`true` `ok` counts as a refusal (`as_bool()?` read `"false"` as success and fell back into the original bug); a structured `error` payload is rendered rather than collapsed to "unspecified error"; and the parse-failure log names the ACK's shape instead of printing `None` for exactly the shapes it cannot explain.

**Coverage.** Every behaviour above is revert-checked individually — reverting each one fails a test that names its own assertion, and 11/11 pass restored. They were run separately on purpose: an earlier combined check hid a vacuous test that asserted a predicate rather than the call path.

**Not merging** — this is one of the two write approvals `main` requires.

One item left for @YellowSnnowmann, not blocking: the PR title still says #709 moved to snake_case, which the in-code comment now contradicts (corrected in `36b1e7ac`). Worth aligning so the title and the code do not disagree in the log.

YellowSnnowmann:
@coderabbitai review

src/openhuman/security/devices/tunnel_client.rs
YellowSnnowmann: Fixed in the subsequent commits on this branch (now head `36b1e7ac`). The `expires_at_from_string_or_epoch_millis` function now checks `millis < MIN_PLAUSIBLE_EPOCH_MILLIS` (2020-01-01, i.e. 1_577_836_800_000) and returns `Err` with a message naming the unit mismatch and the offending value. `a_seconds_valued_pairing_expiry_is_refused_instead_of_decoding_to_1970` covers it — the test asserts the error contains both "epoch milliseconds" and the numeric value, so a regression would surface immediately. The `to_rfc3339_opts(SecondsFormat::Secs, true)` change on the success path is in there too, which also fixes the `+00:00` vs `Z` inconsistency.

src/openhuman/security/devices/tunnel_client.rs
YellowSnnowmann: Fixed in `36b1e7ac`. `parse_register_ack` now calls `describe_ack_shape` before consuming the value via `from_value`, so the shape description is computed regardless of whether the ACK is an object. For a non-object ACK it produces e.g. `"array of 2 element(s)"` or `"null"` rather than `None`. The `clone()` on the success path is gone — `ack` is passed by move into `from_value` and the shape string is built beforehand from a reference.

src/openhuman/security/devices/tunnel_client.rs
YellowSnnowmann: Fixed in `36b1e7ac`. `backend_ack_error` now uses `object.get("ok")?.as_bool() == Some(true)` for the success-path guard: absent `ok` returns `None` from `get` and short-circuits (success), a boolean `true` returns `Some(true)` (success), and anything else — string `"false"`, number `0`, null — is not `Some(true)` and is treated as a refusal. `a_non_boolean_ok_is_still_treated_as_a_refusal` covers all three non-bool cases plus the `ok: true` sanity check.

src/openhuman/security/devices/tunnel_client.rs
YellowSnnowmann: Fixed in `36b1e7ac`. The `error` field is now matched: a `String` value is cloned directly, a non-string present value is rendered via `.to_string()` (so `{"code":429,"message":"channel limit"}` surfaces as its compact JSON representation rather than "unspecified error"), and `None` or `Null` keeps the fallback. `a_structured_error_payload_is_rendered_rather_than_dropped` covers this — it feeds a structured object and asserts both the code and the message survive into the error string.

YellowSnnowmann:
Answering the two questions from the review:

**Q1 — Which shape did you observe on the wire, and is snake_case defensive-only?**

camelCase is what the live backend sends. Backend PR #709's own diff adds `+ channelId: string` and `+ pairingExpiresAt: number` to `TunnelRegisterAck` — so #709 *introduced* that shape, not replaced it. The snake_case aliases are purely defensive / forward-compatibility for a rename that has not happened; the comment in `36b1e7ac` now says so explicitly and records that the original comment had the history backwards.

**Q2 — Can the backend answer `{ ok: true, data: { channelId, … } }`?**

Confirmed no. The backend handler is `safeAck<TunnelRegisterAck>(ack, { ok: false, error: publicError(err) })`, which on success calls the ack callback with the `TunnelRegisterAck` directly — not wrapped in `{ ok: true, data: … }`. The `safeAck` helper only adds the `{ ok: false, error }` envelope on the failure path. `{ ok: true, …fields }` at the top level is structurally handled correctly: `backend_ack_error` returns `None` (not a refusal) and `from_value` would then parse the `channelId` / `pairingToken` / `pairingExpiresAt` fields from the same top-level object — still correct. The wrapped `{ ok: true, data: { channelId } }` shape is not in the backend today; if it were, `backend_ack_error` would pass it through and `from_value` would fail with `missing field 'channelId'`. That case is not defended against, but it also does not match any current or planned backend shape.

M3gA-Mind:
**Independent re-review** (I did not work on this PR). It merged as `98b468f2d` on head `36b1e7ac6` — the head we approved. **The approval stands.** One documentation defect, in one of our own commits, worth a follow-up.

## Coverage — tested, not asserted

The question that matters for a fix like this is whether reverting it breaks anything. I reverted each of the three behavioural changes in turn against merged `main` and ran the suite:

| Reverted | Result | Test that caught it |
|---|---|---|
| baseline | 11 passed | — |
| `as_bool()? ` instead of `== Some(true)` | 1 failed | `a_non_boolean_ok_is_still_treated_as_a_refusal` |
| `MIN_PLAUSIBLE_EPOCH_MILLIS` guard removed | 1 failed | `a_seconds_valued_pairing_expiry_is_refused_instead_of_decoding_to_1970` |
| envelope check dropped from `parse_register_ack` | **4 failed** | `register_failure_envelope_surfaces_the_backend_error_not_a_parse_error`, `..._without_a_message_is_still_a_failure`, `a_non_boolean_ok_...`, `a_structured_error_payload_is_rendered_rather_than_dropped` |
| restored | 11 passed | — |

That last row is the one I care about. The comment on `parse_register_ack` claims it was split out so the *decision path* is reachable without a live `SocketManager`, and that testing `backend_ack_error` alone would not catch the check being dropped from the call path. Dropping it fails four tests — the claim is true, not decorative. This is the opposite of a suite that merely mentions the symbols.

I also confirmed the earlier reply's assertion on the `tunnel_client_tests.rs:72` thread ("baseline 11/11, and each reverted behaviour fails its own named test") reproduces exactly.

## Thread integrity — clean

Eight threads: 0 unresolved non-outdated (the five open ones are all outdated — the lines they anchored to are gone). The single thread resolved under our account was opened by us, fixed by us (`an_unrepresentable_pairing_expiry_is_refused`, feeding `i64::MAX`), and carries a reply naming what was added. No thread resolved with its point unanswered.

## `Frontend Checks: skipping` — legitimate

Verified at the merged head rather than assumed: the diff is two files, `tunnel_client.rs` and `tunnel_client_tests.rs`, and **zero** files under `app/`. Nothing for that lane to run.

## The defect: `36b1e7ac6` corrected the history in one place and left it wrong in two others

`36b1e7ac6` is titled *"docs(devices): correct the stated backend schema history"*. In `tunnel_client.rs` it now says, correctly:

> The backend sends **camelCase**: backend PR #709 introduced `TunnelRegisterAck { channelId, pairingToken, pairingExpiresAt: number }` … and `main` still emits exactly that. The `alias` entries are additive compatibility for a snake_case shape **nothing currently sends**.
> An earlier version of this comment had the history backwards…

But on the same merged head, `tunnel_client_tests.rs:18-19` still says the reverse:

> Regression for #5871: backend PR #709 **switched** the tunnel:register ACK **from camelCase to snake_case**. The struct must accept both shapes so the client works against old and new backend versions…

and the `.expect()` on the next line repeats it (*"snake_case register ack shape (backend PR #709) should parse"*), as does the **PR title**, now in `main`'s log.

So a reader who opens the test file — the natural place to look for what a shape means — gets exactly the wrong history that commit set out to remove, with no hint the production comment contradicts it.

**Impact is documentation-only.** The `alias` entries are harmless additive compatibility, every behaviour is correct and covered, and nothing here changes the merge decision. But the commit claims more than it did, and the wrong history is what sent the first pass hunting field names while the two real defects — the numeric `pairingExpiresAt` and the `{ ok: false }` envelope — sat unnoticed. That is worth not leaving in place.

Suggested follow-up (docs only): bring the test doc comment, its `expect` message, and any future reference into line with the production comment.

## Verdict

**STANDS.** Coverage is real and revert-proven, threads are honest, the skipped lane is justified. The one finding is a partial docs correction in our own commit, not a reason to dismiss the approval.

PR #5977, opened 2026-09-02
fix(search): surface actionable error on Exa 402 credits exhausted and 429 rate limit

### LINKED ISSUE #5750: web_search_tool fails silently when Exa credits are exhausted
## Bug

The `web_search_tool` is returning a 402 error from Exa and the agent continues without a hard failure, producing incorrect/incomplete results with no user-visible indicator.

## Log evidence (2026-08-21 21:12)

```
WRN:observability [observability] integrations.post skipped expected backend user-error response:
Backend returned 400 Bad Request for POST https://api.tinyhumans.ai/agent-integrations/parallel/search:
Exa API error (402): {"error":"You have exceeded your credits limit.
Please top up to keep using Exa at dashboard.exa.ai","tag":"NO_MORE_CREDITS"}

WRN:tools [tinyagents] tool failed tool=web_search_tool error=Backend returned 400 Bad Request...
```

## Impact

- Web search silently returns a tool error; the agent either halts or falls back to an empty search result
- The user sees no indication that search is broken — the agent continues and may give an answer with no web grounding
- Exa credits appear to be a shared pool; once depleted every user session's search tool fails

## Expected

- Surface a clear user-facing message when the Exa search budget is exhausted (e.g. toast or inline message in agent turn)
- Alternatively fall back to another search provider or warn the user in the agent's reply
- Ops: Exa account needs a credit top-up at dashboard.exa.ai to restore functionality

## Environment

Pre-prod desktop app, `openhuman.2026-08-21.log`

sagarithm:
## Summary

Makes an exhausted search-credit pool produce an actionable, user-facing error on **both** search engines, instead of an opaque backend/provider dump.

- **Managed engine (the default)** — classifies a wrapped upstream 402/429 out of the backend's error text and replaces it with a message naming the cause and the way out.
- **Exa BYOK engine** — explicit `402` (credits exhausted) and `429` (rate limit) arms on `ExaClient::post`.
- **Privacy** — the replacement text on both paths carries none of the provider's response body, which echoes the submitted query. Pinned by tests.

## Problem

When the search credit pool is exhausted, `web_search_tool` fails with nothing a user can act on. From #5750:

```
WRN:observability integrations.post skipped expected backend user-error response:
Backend returned 400 Bad Request for POST https://api.tinyhumans.ai/agent-integrations/parallel/search:
Exa API error (402): {"error":"You have exceeded your credits limit.
Please top up to keep using Exa at dashboard.exa.ai","tag":"NO_MORE_CREDITS"}

WRN:tools [tinyagents] tool failed tool=web_search_tool error=Backend returned 400 Bad Request...
```

**There are two `web_search_tool` implementations**, selected by `search.engine` (`src/openhuman/search/registry.rs:16` — *"Exactly one engine owns the canonical `web_search_tool` slot"*), and they fail differently:

| `search.engine` | `web_search_tool` is | 402 arrives as |
|---|---|---|
| `managed` — **the default** (`config/schema/tools/search.rs:138-140`) | `search/tools/web_search.rs` → `POST /agent-integrations/parallel/search` | the backend's own 4xx, provider text in the detail |
| `exa` + own key | `ExaSearchTool::new_web_search_tool` (`exa.rs:381`) → `ExaClient::post` | a real upstream 402 |

The log above is the managed path, so a fix confined to `ExaClient` would not have covered the configuration the report came from.

## Solution

**1. Managed path** — `managed_search_quota_error` in `src/openhuman/search/tools/web_search.rs`, applied via `map_err` at the `/agent-integrations/parallel/search` call site.

Matching is on the **upstream signature** — Exa's `NO_MORE_CREDITS` tag, or a bare `402` *alongside* the word `credit` — rather than on our own status code. That keeps it correct if the backend changes which 4xx it wraps the fault in, and when the managed route resolves to a provider other than the current default. Requiring `credit` beside the bare status stops a `402` appearing in a port or an id from tripping it. Anything unclassified falls through with its original error and detail untouched.

**2. BYOK path** — explicit `PAYMENT_REQUIRED` and `TOO_MANY_REQUESTS` arms in `ExaClient::post`, ahead of the generic `bail!` and after the existing 401/403 arm.

**3. Privacy** — the provider echoes the query back in its error body (`"You have exceeded your credits limit for query: …"`). Both replacements are complete substitutions, so none of that body reaches the agent transcript. `managed_search_quota_error` logs only the classification, never the detail.

## Tests

- `managed_quota_classifier_matches_exa_credit_exhaustion` — the exact string from #5750
- `managed_quota_classifier_matches_the_tag_alone` — a provider reporting the tag without the status
- `managed_quota_classifier_matches_rate_limits` — both the numeric and prose forms
- `managed_quota_classifier_leaves_unrelated_failures_alone` — 500/404/timeout/no-token all fall through, including a bare status inside a port
- `test_execute_surfaces_actionable_error_when_managed_credits_are_exhausted` — end-to-end through a mock backend; asserts the actionable text **and** that the query does not leak
- `payment_required_status_surfaces_actionable_credits_error`, `rate_limited_status_surfaces_actionable_message` — the BYOK arms, same privacy assertion

The pre-existing `non_success_status_does_not_expose_the_response_body` drives a **400**, so it is unaffected by the new arms.

## Notes for review

- **The condensed `json!` schema blocks in `exa.rs` are load-bearing, not churn.** `check-openhuman-rust-layout.mjs` caps a file at 750 lines and `exa.rs` was at 748 on `main`; the fix adds 12. Reverting the condensing would put the file at ~760 and fail the layout gate.
- The redundant `|| status.as_u16() == 402` is dropped — `StatusCode::PAYMENT_REQUIRED` *is* 402.
- Ops item from #5750 is unchanged by any code here: the shared Exa pool still needs a top-up. This makes the failure legible, not absent.

## Submission Checklist

- [x] Tests added or updated (happy path + failure paths)
- [x] Diff coverage ≥ 80%
- [x] No sensitive query / transcript data leaked
- [x] `cargo fmt` / clippy clean on both the contributor and product feature sets
- [x] Layout gate passes
- [x] Closes #5750 — **both** engines, including the default

Closes #5750

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved search error messages for exhausted credits and rate-limited requests.
  - Updated credit-exhaustion messaging to identify HTTP 402 responses.
  - Added actionable handling for managed-search quota errors.
  - Ensured unsuccessful search responses do not expose submitted queries or raw provider details.

- **Tests**
  - Added coverage for credit exhaustion, rate limiting, and quota-error classification.
  - Verified unrelated errors remain unchanged and private query content stays hidden.


src/openhuman/search/tools/exa.rs
```diff
@@ -240,6 +240,18 @@ impl ExaClient {
                      Check your Exa API key under Connections > Search engine."
                 );
             }
+            if status == reqwest::StatusCode::PAYMENT_REQUIRED || status.as_u16() == 402 {
```
YellowSnnowmann: `status.as_u16() == 402` can never add a match — `reqwest::StatusCode::PAYMENT_REQUIRED` **is** 402, and the left side already tested it. The second disjunct is dead.

```suggestion
            if status == reqwest::StatusCode::PAYMENT_REQUIRED {
```

(The 429 arm just below gets this right — it tests `TOO_MANY_REQUESTS` alone.)

src/openhuman/search/tools/exa.rs
```diff
@@ -449,16 +461,8 @@ impl Tool for ExaSearchTool {
...
-                    "type": "array",
-                    "items": { "type": "string" },
-                    "description": "Only return results from these domains."
-                },
-                "exclude_domains": {
-                    "type": "array",
-                    "items": { "type": "string" },
-                    "description": "Exclude results from these domains."
-                },
+                "include_domains": { "type": "array", "items": { "type": "string" }, "description": "Only return results from these domains." },
```
YellowSnnowmann, line 464: Unrelated reformatting — this reflows the `include_domains` / `exclude_domains` schema entries from multi-line to single-line, with no behavioural change. Same in `ExaFindSimilarTool` (~`:560`) and `ExaGetContentsTool` (~`:672`).

These three hunks are **36 of the 68 added lines** in a PR whose actual fix is 12. They make the diff harder to review and will show up in `git blame` on lines the fix never touched.

Please revert them and keep the PR to the two status arms plus tests.

YellowSnnowmann (dismissed):
## Review — the code is correct and non-breaking, but **`Closes #5750` is wrong** and would auto-close a bug this PR does not fix

Verified against `main` (6be58a9ab). The change itself is sound; one blocking item and two cleanups.

---

## Blocking: the issue linkage

**This PR fixes the Exa BYOK path. The failure reported in #5750 is on the managed path.** `Closes #5750` would mark it resolved while the reporter's exact repro still fails.

There are two different `web_search_tool` implementations, selected by `search.engine` (`src/openhuman/search/registry.rs:16` — *"Exactly one engine owns the canonical `web_search_tool` slot"*):

| `search.engine` | `web_search_tool` backed by | 402 path | fixed here? |
|---|---|---|---|
| `managed` (**the default** — `config/schema/tools/search.rs:138-140`) | `search/tools/web_search.rs:307` → `POST /agent-integrations/parallel/search` | backend wraps the upstream Exa 402 as a 400; message assembled generically at `integrations/client_part_01.rs:456` | **no** |
| `exa` + own key | `ExaSearchTool::new_web_search_tool` (`exa.rs:381-391`) → `ExaClient::post` | this PR's new arm | **yes** |

#5750's log is unambiguously the first row:

```
Backend returned 400 Bad Request for POST https://api.tinyhumans.ai/agent-integrations/parallel/search:
Exa API error (402): {"error":"You have exceeded your credits limit...","tag":"NO_MORE_CREDITS"}
WRN:tools [tinyagents] tool failed tool=web_search_tool error=Backend returned 400 Bad Request...
```

That request never reaches `ExaClient` — `ExaClient::new` is constructed only inside `exa.rs` (`:375, :389, :524, :637`), and `web_search.rs` never imports it. The `engines/exa.rs` doc comment states the separation explicitly: *"Exa BYOK: every tool here calls `https://api.exa.ai` directly with the user's own key. **Nothing in this surface routes through the managed backend.**"*

Since `managed` is the default engine, the configuration that produced the bug report is the one still unfixed after this merges.

**Fix:** change `Closes #5750` → `Refs #5750` in the PR body, and leave #5750 open for the managed path. Everything else can stay as-is.

---

## Cleanups (see inline)

1. **Redundant 402 check** — `PAYMENT_REQUIRED || status.as_u16() == 402` tests the same thing twice.
2. **Three unrelated `json!` schema reflows** — 36 of 68 added lines are reformatting untouched code. Please drop them.

---

## What's good — and this is a real fix, not a no-op

- **Correct placement.** The arms sit after the 401/403 check and before the generic `bail!`, so the specific cases win and everything else is unchanged.
- **No body leak.** The body is already read-and-dropped into `body_len` above; the new arms never touch it. (Small note: the PR description credits itself with *"the response body is safely dropped"* — that guarantee is pre-existing, not added here.)
- **No regression.** The existing `non_success_status_does_not_expose_the_response_body` test drives a **400**, so it is unaffected by the new 402/429 arms. Nothing else asserts the old generic message for these two statuses.
- **Tests are the right shape** — both assert the actionable text *and* that the query does not leak (`!message.contains("private search")`), matching the privacy discipline the surrounding tests already establish.
- **It genuinely fixes a reachable configuration.** `web_search_tool`, `exa_search`, `exa_find_similar` and `exa_get_contents` are all registered when `engine = "exa"` (`registry_tests.rs:33-41`), and all four share the patched `post`. So BYOK users get the actionable message on every Exa surface, including `web_search_tool`.
- **CI is fully green** — Rust Quality (fmt, clippy), Rust Core Coverage, feature-gate smoke, and the PR CI Gate all pass.

---

**Summary:** no breaking changes, and I'd be happy to approve once the `Closes` → `Refs` change lands. The two cleanups are minor but worth doing while the branch is open. The managed-path 402 needs its own fix — it has to be recognised where the backend error is assembled (`integrations/client_part_01.rs:456`) or at the `web_search.rs:307` call site, not in `ExaClient`.

YellowSnnowmann:
Pushed `038f7c560` addressing my own review, plus a merge of `main` (the branch was 609 commits behind). Also **withdrawing one of my three review points — it was wrong.**

## Correction: the condensed `json!` schema blocks are load-bearing

I asked for the `include_domains` / `exclude_domains` reflows to be reverted as unrelated churn. **Don't** — they are why this PR passes CI.

`scripts/ci/check-openhuman-rust-layout.mjs:7` caps every file under `src/openhuman/` at `LINE_LIMIT = 750`, with no exemption for `exa.rs`. On `main` that file is **748 lines**; the 402/429 arms add 12, which would put it at 760 and fail the layout gate. Condensing three schema blocks brings it to **734**. Commit `1268fab46` ("chore(layout): condense Exa tool schemas to keep line count under 750") says exactly this — I read the diff without reading the commit that produced it. Apologies for the noise.

## What I changed

**1. The managed path now gets the actionable error too — this is the substantive fix.**

My review's blocking point was that `Closes #5750` was wrong, because the reported log is the **managed** engine (the default) and this PR only covered Exa BYOK. Rather than weaken the link to `Refs`, I added the missing half so the link is true.

`managed_search_quota_error` in `web_search.rs` classifies a wrapped upstream quota fault out of the backend's error text and substitutes an actionable message, applied via `map_err` at the `/agent-integrations/parallel/search` call site. It matches on the **upstream** signature (`NO_MORE_CREDITS`, or a bare `402` alongside the word `credit`) rather than on our own status, so it survives the backend changing which 4xx it wraps the fault in and stays correct if the managed route resolves to a different provider. Requiring `credit` beside the bare status keeps a `402` in a port or an id from tripping it. Unclassified errors fall through untouched.

Privacy is preserved the same way the Exa arms do it: the provider echoes the query back in its error body, so the replacement is a complete substitution and the log line carries only the classification, never the detail.

**2. Dropped the redundant `|| status.as_u16() == 402`** — `StatusCode::PAYMENT_REQUIRED` *is* 402, so the second disjunct could never match.

**3. Six new tests** — four unit tests over the classifier (the exact #5750 string, tag-only, both rate-limit forms, and four negative cases), and one end-to-end test through a mock backend that asserts both the actionable text and that the query does not leak.

## Verification

| gate | result |
|---|---|
| `cargo fmt -p openhuman -- --check` | clean |
| `cargo clippy -p openhuman -- -D warnings` (contributor set) | clean, exit 0 |
| `cargo clippy -p openhuman --features "$(bash scripts/ci/product-features.sh)" -- -D warnings` | clean, exit 0 |
| `node scripts/ci/check-openhuman-rust-layout.mjs` | passed |
| `cargo test --lib openhuman::search::` | **129 passed, 0 failed** |

Both CI clippy invocations were run because the product set is the one that lints what ships, and the contributor set is what a slim build sees — a lint firing in only one of them is the trap `ci-lite.yml:438-445` documents.

No behaviour changes outside a non-2xx search response. The existing `non_success_status_does_not_expose_the_response_body` test drives a **400** and is unaffected by the new arms.

My `CHANGES_REQUESTED` should now be satisfied — the linkage point by fixing the gap rather than renaming it, the redundancy by removal, and the third point withdrawn above.

YellowSnnowmann:
---

*Posting this as a comment rather than an approval: I authored part of this diff, so the approval that clears my earlier `CHANGES_REQUESTED` should come from someone who didn't. My requested changes are addressed and I have no remaining objection — treat this as **LGTM, Mergeable!** pending that dismissal.*

src/openhuman/search/tools/exa.rs
```diff
@@ -240,6 +240,18 @@ impl ExaClient {
...
+                    "Exa search credits exhausted (HTTP 402). \
+                     Please top up your Exa account at dashboard.exa.ai or switch search provider."
+                );
+            }
+            if status == reqwest::StatusCode::TOO_MANY_REQUESTS {
+                anyhow::bail!(
+                    "Exa rate limit exceeded (HTTP 429). \
+                     Please wait a moment before searching again."
+                );
+            }
```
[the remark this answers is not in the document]
senamakel: Updated the shared ExaClient messages to be endpoint-neutral: the 429 guidance now says “before retrying” rather than referring specifically to searching. I also removed the redundant numeric 402 check in the pushed follow-up commits.

src/openhuman/search/tools/exa_tests.rs
```diff
@@ -330,6 +330,55 @@ async fn unauthorized_status_names_the_key_without_leaking_the_body() {
...
+    let base_url = spawn(app).await;
+
+    let tool = ExaSearchTool::new(Some("test-key".into()), Some(base_url), 5, 15);
+    let err = tool
+        .execute(json!({"query": "private search"}))
+        .await
+        .expect_err("402 must fail with actionable message");
+    let message = err.to_string();
+
+    assert!(message.contains("Exa search credits exhausted (HTTP 402)"));
```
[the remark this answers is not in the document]
senamakel: Updated the shared ExaClient 429 wording and kept the 402 assertion aligned with the endpoint-neutral credits message. The focused Exa tests pass.

src/openhuman/search/tools/exa.rs
senamakel: Fixed in the pushed follow-up: the 402 branch now compares only against reqwest::StatusCode::PAYMENT_REQUIRED, matching the existing 429 style.

src/openhuman/search/tools/exa.rs
senamakel: I reviewed this against the layout gate before changing it. The schema compaction is required here: exa.rs was 748 lines on main and the new status branches add 12, while check-openhuman-rust-layout.mjs caps src/openhuman files at 750 lines. Reverting these three blocks would make the PR fail that gate, so I’m retaining the load-bearing formatting and have resolved this thread with that reason.

src/openhuman/search/tools/web_search.rs
```diff
@@ -15,6 +15,51 @@ use std::sync::Arc;
...
+/// provider echoes the submitted query back in its error body, and that body
+/// would otherwise reach the agent transcript.
+fn managed_search_quota_error(message: &str) -> Option<&'static str> {
+    let lowered = message.to_ascii_lowercase();
+
+    // Exa tags credit exhaustion explicitly; the numeric form covers providers
+    // that only report the status. `credit` is required alongside the bare
+    // status so an unrelated 402 in a URL or id cannot trip this.
+    let credits_exhausted = lowered.contains("no_more_credits")
+        || (lowered.contains("402") && lowered.contains("credit"));
```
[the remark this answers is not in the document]
senamakel: Fixed in the pushed follow-up. managed_search_quota_error now inspects only structured provider status envelopes (API error 402/429 or an outer backend 429) and parses the provider JSON tag/status fields; it no longer scans arbitrary response detail. I added negative tests covering echoed NO_MORE_CREDITS, HTTP 429, and rate-limit text. The managed classifier tests and all Exa tests pass.

PR #5988, opened 2026-09-02
fix(git_operations): stop `-c diff.external=` from breaking every diff

### LINKED ISSUE #5979: git_operations: every `diff` fails — `-c diff.external=` makes git execute an empty program
## What

`git_operations`' `diff` operation fails for every repository on `main`:

```
Git command failed: error: cannot run : No such file or directory
fatal: external diff died, stopping at <file>
```

## Root cause

`hardened_git` injects every entry of `NEUTRALISED_CONFIG` as a `-c` override (`src/openhuman/tools/impl/filesystem/git_operations_config.rs`), and one of them is:

```rust
"diff.external=",
```

The intent is to neutralise a repository-set external diff driver. But `git -c diff.external=` does not *disable* the driver — it sets it to the empty string, and git then tries to **execute** that empty string. Reproducible on a bare repo with no config of its own:

```console
$ git init -q . && echo a > f.txt && git add f.txt && git commit -qm init && echo b >> f.txt

$ git -c diff.external= diff -- f.txt
error: cannot run : No such file or directory
fatal: external diff died, stopping at f.txt

$ git -c diff.external= diff --no-ext-diff -- f.txt
diff --git a/f.txt b/f.txt
...                      # correct output
```

Every other entry in that list happens to name a program that exists or a key git treats as inert (`core.pager=cat`, `core.editor=false`, `sequence.editor=false`); `diff.external` is the one where an empty value means "run the empty program" rather than "none".

## Impact

`GitOperationsTool` builds its diff through `hardened_git` (`git_operations.rs`, `git_args = vec!["diff", "--unified=3"]`), so the agent's `diff` operation is broken for all users, not only in tests. `log`/`show` are affected wherever they produce a patch.

## How it surfaced

The raw-coverage test `tools_network_channels_raw_coverage_e2e::git_operations_cover_read_write_markdown_and_safety_rejections` fails on it. It is not caught on every PR because the coverage lane derives which raw-coverage modules to run from the changed paths, so it only runs when something under `src/openhuman/tools/impl/filesystem/` is in the diff. It failed on #5955 purely because that PR merged `main`.

## Fix

`diff.external` cannot be neutralised by a `-c` value — there is no value meaning "none". Drop it from `NEUTRALISED_CONFIG` and pass `--no-ext-diff` on the subcommands that honour an external diff driver (`diff`, and `log`/`show` when they produce a patch). `GIT_EXTERNAL_DIFF` is already cleared by `suppress_ambient_git_config`, so the env half of the hole is closed.

## Acceptance criteria

- [ ] `git_operations` `diff` succeeds on a repository with no external diff driver configured
- [ ] A repository-set `diff.external` naming a program is still not executed (the hardening's actual goal), with a regression test that sets one and asserts it does not run
- [ ] `tools_network_channels_raw_coverage_e2e::git_operations_cover_read_write_markdown_and_safety_rejections` passes
- [ ] Any other `NEUTRALISED_CONFIG` entry whose empty value is command-valued is audited the same way

YellowSnnowmann:
Closes #5979.

## The bug

`hardened_git` injects every `NEUTRALISED_CONFIG` entry as a `-c` override, and one of them was `diff.external=`. An empty value does **not** disable an external diff driver — git tries to execute the empty string — so every `diff` this tool ran died with:

```
error: cannot run : No such file or directory
fatal: external diff died, stopping at <file>
```

That is not hardening; it is an outage wearing hardening's clothes. `GitOperationsTool`'s `diff` operation is broken on `main` for **every** repository, with or without config of its own — this is a user-facing break, not only a test failure.

Reproduced standalone, no OpenHuman involved:

```console
$ git init -q . && echo a > f.txt && git add f.txt && git commit -qm init && echo b >> f.txt

$ git -c diff.external= diff -- f.txt
error: cannot run : No such file or directory
fatal: external diff died, stopping at f.txt

$ git -c diff.external= diff --no-ext-diff -- f.txt
diff --git a/f.txt b/f.txt          # correct
```

## The fix

`diff.external` cannot be neutralised by a `-c` value — there is no value meaning "none". Two things replace it, and together they are strictly stronger than what they replace:

- **The allowlist already refuses it.** `diff.external` is not on `ALLOWED_REPO_CONFIG`, so a repository that sets it is refused outright by `first_disallowed_repo_config_key`. That is the fail-closed guarantee, and the `-c` entry never carried it.
- **`--no-ext-diff` on the command** covers what the `-c` entry was actually reaching for: a key written into the repository in the gap between that inspection and the command itself — the race the `NEUTRALISED_CONFIG` doc comment describes.

Verified directly, against a repository whose `diff.external` names a script that touches a marker file: without the flag the marker appears and the driver's output is used; with it the diff is correct and the marker never appears. `GIT_EXTERNAL_DIFF` was already cleared by `suppress_ambient_git_config`, so the env half was never open.

`git_log` passes `--pretty=format:` and never produces a patch, so it has no external diff to refuse; `diff` is the only affected operation.

## Why it wasn't caught

The coverage lane derives which raw-coverage modules to run from the changed paths, so `tools_network_channels_raw_coverage_e2e` only runs when the diff reaches into that area. It surfaced on #5955 — a PR that touches no git code at all.

## Test plan

- [x] `an_ordinary_repository_still_produces_a_diff` — the regression: a repository with no config of its own can produce a diff
- [x] `a_repository_naming_an_external_diff_driver_never_gets_it_run` — the hardening property still holds; the planted driver is proven to run before it is planted, so a pass means refusal, not an inert fixture
- [x] `diff_external_is_not_neutralised_by_a_config_override` — pins that the entry is never re-added, since the value that looks correct there is the one that breaks the tool
- [x] `tools_network_channels_raw_coverage_e2e::git_operations_cover_read_write_markdown_and_safety_rejections` — the CI test that exposed this now passes
- [x] `cargo clippy` (product feature set) clean at `-D warnings`; `cargo fmt --check` clean
- [x] 15/15 `git_operations::config_tests`

### DIFF
--- a/src/openhuman/tools/impl/filesystem/git_operations.rs
+++ b/src/openhuman/tools/impl/filesystem/git_operations.rs
@@ -174,7 +174,14 @@ impl GitOperationsTool {
         // Validate files argument against injection patterns
         self.sanitize_git_args(files)?;
 
-        let mut git_args = vec!["diff", "--unified=3"];
+        // `--no-ext-diff` is what actually refuses a repository-set
+        // `diff.external`. It cannot be done with a `-c` override the way the
+        // other command-valued keys are neutralised — an empty value makes git
+        // execute the empty string rather than disable the driver — so the
+        // flag lives here, on the one operation that can run one (#5979).
+        // `git_log` uses `--pretty=format:` and never produces a patch, so it
+        // has no external diff to refuse.
+        let mut git_args = vec!["diff", "--no-ext-diff", "--unified=3"];
         if cached {
             git_args.push("--cached");
         }
--- a/src/openhuman/tools/impl/filesystem/git_operations_config.rs
+++ b/src/openhuman/tools/impl/filesystem/git_operations_config.rs
@@ -120,12 +120,31 @@ pub(super) const ALLOWED_REPO_CONFIG: &[&str] = &[
 ///
 /// `core.hooksPath` and `commit.gpgSign` are handled separately, in
 /// [`hardened_git`] — see there for why.
+///
+/// `diff.external` is deliberately **absent** from this list, and must stay
+/// absent. It is command-valued like the rest, but unlike the rest there is no
+/// `-c` value that means "none": an empty value does not disable the driver,
+/// it makes git execute the empty string, so `-c diff.external=` turns every
+/// diff into
+///
+/// ```text
+/// error: cannot run : No such file or directory
+/// fatal: external diff died, stopping at <file>
+/// ```
+///
+/// — which is not hardening, it is an outage that merely looks like one
+/// (#5979). The supported way to refuse an external diff driver is the
+/// `--no-ext-diff` flag, which [`super::git_operations`] passes on the one
+/// operation that can run one; verified directly, against a repository whose
+/// `diff.external` names a script that touches a marker file: without the flag
+/// the marker appears, with it the diff is correct and the marker does not.
+/// The env half of the same hole is closed by
+/// [`suppress_ambient_git_config`]'s `GIT_EXTERNAL_DIFF` removal.
 pub(super) const NEUTRALISED_CONFIG: &[&str] = &[
     "core.fsmonitor=",
     "core.sshCommand=",
     "core.pager=cat",
     "core.editor=false",
-    "diff.external=",
     "sequence.editor=false",
     "uploadpack.packObjectsHook=",
 ];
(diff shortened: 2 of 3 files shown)


YellowSnnowmann:
Closing — the fix is moving into #5955 instead, where the CI failure it repairs actually surfaced. Same commit, same tests; #5979 stays open until it lands there.

PR #6050, opened 2026-09-07
chore(modules): re-pin the memory module to tinymemory v1.15.1

CodeGhost21:
## What changed

Re-pins the memory module from tinymemory **v1.14.1** to **v1.15.1**.

Two changes since v1.14.1:

- **tinymemory#141** — batches a source's document embeds instead of embedding
  one at a time, and counts the graph-extraction jobs a batch write could not
  queue.
- **tinymemory#142** — the CortexDB adapter now asks `v1/scopes/list` for every
  scope. The endpoint caps at **50**, documents neither the cap nor the `limit`
  that lifts it, and returns no cursor, no `has_more` and no total — so nothing
  in the response says it was cut short. A live instance holding 93 scopes
  listed 50, and every caller downstream of `namespace_summaries` (including
  `export_page`, and so `opencompany memory migrate`) reported success on a
  partial answer. It now requests a large limit and refuses a response that
  fills it, because a full page and a truncated one are otherwise the same bytes.

## Capability audit

**No capability moves.** `ARTIFACT_CAPABILITIES` is unchanged and only the pin
beside it advances:

```
git diff v1.14.1..v1.15.1 -- crates/tinymemory-api/src/capabilities.rs \
                             crates/tinymemory-bus/src/capabilities.rs   → empty
crates/tinymemory-module/                                                → Cargo.lock only
```

So no family arrived and no bus member did either. The rationale is recorded in
the `ARTIFACT_CAPABILITIES` doc comment in the form the file already uses.

## All four pinned sites move together

`the_capability_list_matches_the_pinned_release` and
`the_ci_workflows_pin_the_same_module_digest_as_the_registry` exist to make a
half-done bump a red test rather than a lane that dies on `sha256sum: WARNING: 1
computed checksum did NOT match`. Both are satisfied:

| Site | Change |
| --- | --- |
| `vendor/tinymemory` | gitlink → `v1.15.1` (`258d9e5`) |
| `modules/registry_part_01.rs` | `version`, `release_url`, and all 11 assets |
| `ARTIFACT_CAPABILITIES_PIN` | `1.14.1` → `1.15.1` |
| `ci-full.yml`, `ci-lite.yml`, `e2e-reusable.yml` | `memory_version` + `memory_sha256` (4 sites) |

The sha256s were read from the release's own `checksum.toml` rather than
transcribed. The `ubuntu-22.04-x86_64` digest — the one CI actually installs —
was additionally verified by downloading the archive and hashing it:

```
computed: 3d9140c1625e142dfa1686cd5a07b5f4f43e2f576f402c892f30f503b1fa0aa7
pinned:   3d9140c1625e142dfa1686cd5a07b5f4f43e2f576f402c892f30f503b1fa0aa7
```

## Validation

| Command | Outcome |
| --- | --- |
| `cargo fmt -- --check` | pass |
| `cargo check --locked --all-targets` | pass — the existing lockfile still resolves, no drift |
| `cargo test --locked --lib modules::` | pass, including both pin-drift guards |

One note, not from this change: `cargo clippy --locked --all-targets` fails on
`approx_constant` on a literal `3.14` in
`src/core/rpc_log.rs:105`. That file is untouched here and its last commit is
unrelated, so it is pre-existing on `main`.

## Public API / behavior changes

None in this repo. The pinned artifact serves the same capability families it
did at v1.14.1.

## Summary by CodeRabbit

- **Updates**
  - Updated the bundled TinyMemory module to version 1.15.1.
  - Refreshed platform-specific downloads and verification data for the new release.
  - Updated automated testing and coverage workflows to use and validate TinyMemory 1.15.1.
  - Confirmed that the available capability set remains unchanged in this update.
  - Included the latest TinyMemory release updates without adding new capability families or bus members.


CodeGhost21:
Rebased onto main after #6049 landed the v1.15.0 re-pin — this branch and that one collided on all four pinned sites, so I reset onto it and re-applied only the delta rather than resolving conflicts.

That makes the change smaller and the record better: #6049's v1.15.0 rationale stays intact, and this adds a v1.15.1 note above it covering only tinymemory#142.

Updated the capability audit to the narrower range, and verified rather than assumed:

```
git diff v1.15.0..v1.15.1 -- crates/tinymemory-api/src/capabilities.rs \
    crates/tinymemory-bus/src/capabilities.rs crates/tinymemory-bus/src/names.rs   → empty
crates/tinymemory-module/                                                          → Cargo.lock only
```

One correction to my own comment while writing it: I first wrote that `crates/tinymemory-module/` "does not move at all". It does — its `Cargo.lock` moves. Reworded, since the point is that the module declares the same members, not that nothing changed.

Also caught in the rebase: the worktree was carrying an older `vendor/tinycortex` than main now pins, which would have silently reverted #6049's bump of it. Restored to `79131f2`; this branch touches only `vendor/tinymemory`.

Re-verified on the rebased branch: `cargo fmt -- --check`, `cargo check --locked --all-targets`, and `cargo test --locked --lib modules::memory` (38 passed, including both pin-drift guards).

M3gA-Mind (approved):

PR #6057, opened 2026-09-07
chore(tsconfig): drop baseUrl so app/tsconfig survives TypeScript 7

ntdatt812:
TypeScript 6 reports this as an error:

```
Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0.
  Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

`app/` pins `typescript@5.8.3`, so the current typecheck is quiet. An editor running a newer TypeScript flags it now, and it becomes a hard failure the moment the pin moves to 7.

## Why removing it rather than silencing it

`baseUrl` here only anchors `paths`, and `paths` has not needed it since TypeScript 4.4 — entries resolve relative to the tsconfig that declares them. Two of the three entries gain a `./` prefix; `tauri-plugin-ptt-api` was already relative.

`"ignoreDeprecations": "6.0"` would also quiet it, but it buys one major version and has to be revisited. This makes the config correct instead.

## Why it is safe

`baseUrl` has a second effect — it lets bare specifiers resolve against the base directory — so removing it can break imports that quietly relied on that. Checked: there are no `from "src/…"`-style imports under `app/src`.

`vite.config.ts` declares the alias itself:

```ts
resolve: { alias: { "@": resolve(__dirname, "src"), … } }
```

so the build never reads this setting; the change is type-level only.

## Verification

`tsc --noEmit -p app/tsconfig.json` gives **byte-identical output before and after** — 7 errors either way, all from `packages/tauri-plugin-ptt/guest-js`, which resolves its own dependencies outside `app/` and is unrelated to this change. No `Cannot find module '@/…'` appears in either run, with **33 files** under `app/src` importing through the alias — a broken alias could not have been silent.

Re-checked the same config under TypeScript **6.0.3**: the deprecation is gone, and no new diagnostic replaces it.

`app/tsconfig.node.json` has no `baseUrl`, so it needed no change.

## Summary by CodeRabbit

- **Chores**
  - Updated TypeScript path alias configuration to use relative source paths.
  - Removed the project-wide base URL setting while preserving existing plugin alias resolution.


ntdatt812:
The red `Frontend Checks` here is not from this change, and it is worth saying why it showed up on a PR that only edits `tsconfig.json`.

`paths-filter` classifies a tsconfig edit as a config-level change, so the job escalates to the full suite rather than the changed-file subset:

```
[ci][vitest-changed] running FULL Vitest coverage suite
  (reason: config/workflow-level change detected by paths-filter)
```

Exactly one test in that full run failed — `SandboxSettingsPanel.validation.test.tsx > persists a positive integer` — with a DOM dump still showing `Loading…`. Its `renderLoaded` helper waits for `mockGet` to have been *called*, which is already true on `waitFor`'s first tick because the panel calls it synchronously in its mount effect; the synchronous `getBy*` accessors then run before `isLoading` flips.

Filed separately as #6064, with a deterministic reproduction. It is unrelated to `baseUrl` — vitest resolves `@` from the explicit alias in `test/vitest.config.ts`, not from tsconfig, so nothing in this PR can reach it.

PR #6093, opened 2026-09-08
fix(flows): stop claiming ungranted trust and reject malformed contract slugs

### LINKED ISSUE #6071: `flows_approval_manifest` reports un-granted permissions as `already_trusted` when the approval gate is not installed
## Summary

`flows_approval_manifest` reports un-granted permissions as `already_trusted` when the approval gate is not installed

- **Namespace / method:** openhuman.flows_approval_manifest
- **Where:** `src/openhuman/flows/ops_part_12.rs:414-421`

## Problem

**Expected**

With no `ApprovalGate` installed, nothing ever parks, so `missing`
is empty by definition — that part is correct and documented. But the
approvable keys have *not* been granted to anything, so they do not belong in
`already_trusted` either. Either a third state ("not gated"), or an empty
`already_trusted` alongside `gate_installed: false`, tells the truth.

**Actual**

The `!gate_installed` branch pushes every approvable `tool_name`
onto `already_trusted`:

```rust
if !gate_installed {
    // Nothing parks without a gate; report nothing as missing.
    already_trusted.push(tool_name.to_string());
}
```

So a graph with an `http_request` and a `code` node comes back as
`already_trusted: ["flows_http_request", "flows_code"]` on a host where no
grant was ever made. The save+enable card's whole job is to tell the user
which permissions a run will ask for; a consumer reading `already_trusted`
renders "already authorized" for two permissions the flow does not hold.
`gate_installed: false` is present in the same payload, so a careful consumer
can recover — but the field itself is asserting something false.

**Steps to reproduce**

`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" flows_approval_manifest_classifies_every_gated_node_kind`
— the test pins the current behaviour in its `else` branch, with a pointer here.
The transport-only JSON-RPC router (`build_core_http_router`) installs no core
runtime, so `ApprovalGate::try_global()` is `None` and this is the branch an
RPC-level test reaches. Production normally has the gate installed;
`OPENHUMAN_APPROVAL_GATE=0` reaches it there too.

**Cited location still accurate:** yes, exactly — the `!gate_installed` branch is
`src/openhuman/flows/ops_part_12.rs:416-418`, inside the loop at `:409-424`, and the
payload it feeds is built at `:433-441`. The report's `414-421` covers it.

**Repro run:** the report's named test lives on W1's branch, so I dispatched the controller
directly over the real router with a two-node graph (one `http_request`, one `code`):
`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- --nocapture w6_verify_scratch::`
(throwaway probe, deleted after the run, tree is clean).

**Observed**

```
flows_approval_manifest -> {"result":{"result":{
  "already_trusted":["flows_http_request","flows_code"],
  "entries":[
    {"class":"Network","kind":"approvable","label":"Call https://example.test/x",
     "node_id":"h","tool_name":"flows_http_request"},
    {"class":"Write","kind":"approvable","label":"Run sandboxed code",
     "node_id":"c","tool_name":"flows_code"}],
  "gate_installed":false,
  "missing":[]}}}
```
Two permissions reported as `already_trusted` on a host where no grant was ever made.

**Diagnosis assessment:** correct. And there is **stronger authority for it than the report
cites**: the sibling RPC on the same concept, `approval_preauthorize_flow`
(`src/openhuman/security/approval/rpc.rs:125-138`), handles the *identical* condition —
`ApprovalGate::try_global()` is `None` — by returning
`FlowPreauthorizationResult { granted: vec![], already_trusted: vec![], gate_installed: false }`.
So the two halves of the same save+enable card, emitting the same two field names,
disagree about what `already_trusted` means with no gate installed. That is an in-repo
precedent rather than an argument from first principles.

**Is the expected behaviour correct:** yes — but the report's stated *impact* is overstated
and triage should know it. I traced every consumer: `already_trusted` from the **manifest**
is not rendered anywhere. It is in the typed public contract
(`ApprovalManifest` at `app/src/services/api/flowsApi.ts:837`, populated at `:860`), but the
only read in the app is `app/src/hooks/useFlowPreauthorization.ts:131`, which reads the
**preauthorize** result's copy and only passes it to a `log(...)` call. The card's actual
branch is on `pending.manifest.missing.length` (`:127`), which is correctly empty here.
So this is a latent defect in a typed public interface — a field asserting something false —
rather than a wrong string on screen today. Real, worth fixing, lower severity than
"wrong-behaviour" implies.

## Solution

make the `!gate_installed` branch `continue` instead of
pushing, so both `missing` and `already_trusted` are empty and `gate_installed: false` is
the single signal — matching `approval_preauthorize_flow`. One line. The report's other
option (a third "not gated" state) would widen the wire contract for no consumer that needs it.

**Blast radius:** `ops_part_12.rs:377-442` is the only producer. Wire consumers:
`flowsApi.ts:837/860` (type + normalization), `useFlowPreauthorization.ts` (logs only),
and the schema declaration at `src/openhuman/flows/flows_schema_part_02.rs:56`. Tests that
pin current behaviour: W1's `flows_approval_manifest_classifies_every_gated_node_kind`
(its `else` branch, per the report) — that assertion must flip in the same change.
`WorkflowProposalCard.test.tsx` and `useFlowPreauthorization.test.ts` both fixture
`already_trusted: []` with `gate_installed: true`, so they are unaffected either way.

## Acceptance criteria

## Related

### LINKED ISSUE #6072: `flows_get_tool_contract`'s `<TOOLKIT>_<ACTION>` shape guard is unreachable for any non-empty slug
## Summary

`flows_get_tool_contract`'s `<TOOLKIT>_<ACTION>` shape guard is unreachable for any non-empty slug

- **Namespace / method:** openhuman.flows_get_tool_contract
- **Where:** `src/openhuman/flows/ops_part_12.rs:474-480`, over
`vendor/tinymemory/crates/tinymemory-bus/src/composio/scopes.rs:172-194`

## Problem

**Expected**

A slug that is not of the form `<TOOLKIT>_<ACTION>` is rejected
with the message the controller carries for exactly that case:

> Could not extract a toolkit from slug 'x' — it must look like
> '<TOOLKIT>_<ACTION>' (e.g. 'GMAIL_SEND_EMAIL').

**Actual**

`toolkit_from_slug` splits on `_` and falls back to the **whole
string** when there is no separator, so it returns `Some(..)` for every
non-empty input. `toolkit_from_slug("nodashhere")` is `Some("nodashhere")`.
The guard therefore fires only for a slug that trims to empty — and
`read_required::<String>` accepts `""`, so `{"slug": "   "}` is the only
reachable path to that message. A user who types a bare action name gets the
unrelated, much less useful

> Could not fetch the live Composio catalog for toolkit 'nodashhere'.

after an attempted catalog round trip, rather than being told the slug is
malformed before any I/O.

**Steps to reproduce**

`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" flows_tool_catalog_surface_degrades_without_composio_credentials`
— the test pins both paths, the reachable guard and the dashless fall-through.

**Cited location still accurate:** yes, both, exactly. The guard is
`src/openhuman/flows/ops_part_12.rs:474-480` (`flows_get_tool_contract` starts at `:470`),
and `toolkit_from_slug` is
`vendor/tinymemory/crates/tinymemory-bus/src/composio/scopes.rs:172-194`. The fall-back the
report describes is `let prefix = trimmed.split('_').next()?;` at `:188` — `split` on a
string with no separator yields the whole string as its first element, so any non-empty
trimmed input returns `Some`.

**Repro run:** the report's named test lives on W1's branch, so I dispatched the controller
directly over the real router for both paths:
`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- --nocapture w6_verify_scratch::`
(throwaway probe, deleted after the run, tree is clean).

**Observed**

```
slug="nodashhere" -> {"error":{"code":-32000,"message":
    "Could not fetch the live Composio catalog for toolkit 'nodashhere'."}}

slug="   "        -> {"error":{"code":-32000,"message":
    "Could not extract a toolkit from slug '' — it must look like
     '<TOOLKIT>_<ACTION>' (e.g. 'GMAIL_SEND_EMAIL')."}}
```
Exactly the reported inversion: the malformed-slug message is reachable only for a blank
slug, and a genuinely malformed slug gets the catalog message instead.

**Diagnosis assessment:** correct on every point. Two details worth adding for whoever
fixes it. (1) The blank-slug message renders as `slug ''` because `:474` trims before the
guard, so even the one reachable case names an empty string rather than what the caller
sent — a second small diagnosability loss in the same four lines. (2) The report's "after an
attempted catalog round trip" is right: the fall-through reaches
`fetch_live_toolkit_catalog` (`src/openhuman/integrations/composio/catalog_part_01.rs:258`),
which takes a per-toolkit async fetch lock at `:279` and calls
`fetch_raw_toolkit_tools` at `:294` before it can fail.

**Is the expected behaviour correct:** yes. This endpoint fetches **one action's** contract,
so a slug with no action segment cannot name anything it can return; rejecting it before I/O
is right. The report's own caveat is also correct and I confirmed it: the permissive
fall-back is load-bearing for `compute_required_connections`
(`ops_part_12.rs:139`), which maps *toolkit-ish* tokens and legitimately wants
`Some(whole_string)`. So the fix belongs in this controller, not in the vendored crate.

## Solution

in `flows_get_tool_contract`, validate the slug shape
before calling `toolkit_from_slug` — require a non-empty trimmed slug containing `_` with a
non-empty segment on each side (allowing the three `MULTI_SEGMENT_TOOLKIT_PREFIXES`, which
all end in `_` and so already satisfy it) — and report the caller's original slug, not the
trimmed one, in the message.

**Blast radius:** the guard itself is used only by `flows_get_tool_contract`
(`ops_part_12.rs:470-480`), schema at `flows_schema_part_02.rs:128`. `toolkit_from_slug` is
vendored and shared: changing *it* would also hit `compute_required_connections`
(`ops_part_12.rs:139`) and any other `tinymemory_api::composio` caller — which is the
argument for keeping the fix controller-local. W1's
`flows_tool_catalog_surface_degrades_without_composio_credentials` pins both paths today
and would need its dashless expectation updated in the same change.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

Two `p2` defects from the e2e coverage wave, both in `src/openhuman/flows/` RPCs that tell the user what a run will ask for. They share a file, so they share a PR.

### #6071 — `flows_approval_manifest` reported ungranted permissions as `already_trusted`

With no `ApprovalGate` installed, the `!gate_installed` branch pushed **every** approvable `tool_name` onto `already_trusted`. Nothing parks without a gate, so `missing` is empty by definition — but nothing was ever *granted* either, so those keys did not belong in `already_trusted`. A graph with an `http_request` and a `code` node came back as `already_trusted: ["flows_http_request", "flows_code"]` on a host holding no grant at all.

The classification moved into `split_manifest_trust`, which returns **both** lists empty when no gate is installed, leaving `gate_installed: false` as the single signal. Two pieces of in-repo authority say this is the right answer:

- the sibling `approval_preauthorize_flow` (`src/openhuman/security/approval/rpc.rs:129-138`) already returns `granted: vec![], already_trusted: vec![], gate_installed: false` for the *identical* condition;
- the RPC schema already documented the field as "Approvable trust keys **already granted** to this flow" — the code was violating its own declared contract.

Extracting the helper is also what makes the fix testable: `GLOBAL_GATE` is a process-global `OnceLock`, so in a shared `cargo test --lib` binary whether a gate is installed depends on test order. A pure function taking `gate_installed: bool` pins both branches deterministically.

### #6072 — `flows_get_tool_contract`'s `<TOOLKIT>_<ACTION>` guard was unreachable

`toolkit_from_slug` splits on `_` and falls back to the **whole string** when there is no separator, so it answers `Some` for every non-empty input — `toolkit_from_slug("nodashhere") == Some("nodashhere")`. A bare action name therefore resolved to a bogus toolkit, took a per-toolkit async fetch lock, attempted a catalog round trip, and failed with the unrelated *"Could not fetch the live Composio catalog for toolkit 'nodashhere'."*

New `toolkit_for_contract_slug` validates the shape first — non-empty segments either side of the first `_` — and the trim now happens **before** the guard so the message quotes the caller's own slug instead of naming an empty string back at whoever sent whitespace.

The vendored `toolkit_from_slug` is deliberately left permissive: `compute_required_connections` (`ops_part_12.rs:139`) maps toolkit-ish tokens and legitimately wants the whole string back. The stricter rule belongs to the contract-fetch callers.

All three `MULTI_SEGMENT_TOOLKIT_PREFIXES` end in `_`, so `ZOHO_MAIL_SEND_EMAIL`, `ONE_DRIVE_UPLOAD_FILE` and `MICROSOFT_TEAMS_SEND_MESSAGE` all pass unchanged — pinned by a test.

### One fix beyond the filed scope, flagged for review

The **agent's** `get_tool_contract` tool (`builder_tools_part_04.rs:353`) carried a verbatim copy of the same guard against the same permissive helper. Neither issue mentions it. There the guard was *fully* dead — the tool rejects an empty slug earlier with "Missing 'slug' parameter", so no input could ever reach it — and the fall-through was more damaging than on the RPC, because its catalog error tells the model to *"try again, or use search_tool_catalog"* for a slug that can never resolve, burning turns.

It now calls the same helper, which is what `flows_get_tool_contract`'s own docstring already claimed ("reusing the same core as the agent's `get_tool_contract` tool"). One-line call swap plus a regression test. Happy to split it out if you would rather keep this PR to the two filed issues.

## Blast radius

Every consumer named in the issues was checked, plus a repo-wide sweep for `already_trusted`:

- `app/src/services/api/flowsApi.ts` — type + normalization; the field's doc comment now states the `gate_installed: false` ⇒ empty rule.
- `app/src/hooks/useFlowPreauthorization.ts:131` — the only read, and it reads the **preauthorize** result's copy, passing it to `log(...)`. The card branches on `pending.manifest.missing.length`, which was already correct here.
- `WorkflowProposalCard.test.tsx`, `useFlowPreauthorization.test.ts`, `flowsApi.test.ts`, `approvalApi.test.ts` — every fixture uses `gate_installed: true`, so none changes meaning. All 95 tests across those four files pass unchanged.
- `src/openhuman/security/approval/` — `FlowPreauthorizationResult` is a separate type on a separate path; untouched.

W1's `flows_approval_manifest_classifies_every_gated_node_kind` and `flows_tool_catalog_surface_degrades_without_composio_credentials` are **not on `main` yet**, so there is nothing to flip here. If that branch lands first, its `else`-branch and dashless-slug expectations will need updating in the same way — worth sequencing.

## Note for #6072

`schemas_handlers.rs:402` already does `slug.trim()` before calling into `ops`, so over JSON-RPC the "caller's original slug" is pre-trimmed and leading/trailing whitespace cannot be echoed back. The ops-level fix is still correct for direct callers and defensive for the RPC; I did not change the handler, since trimming there is reasonable on its own.

## Test plan

New coverage (5 tests, all passing):

- `split_manifest_trust_claims_nothing_without_a_gate` — no gate (both lists empty), gate + no grants (all missing), gate + one grant (splits correctly). No process-global state.
- `approval_manifest_never_claims_trust_that_was_never_granted` — RPC-level pin over the real controller; `already_trusted` must be empty regardless of ambient gate state.
- `get_tool_contract_rejects_a_malformed_slug_before_any_catalog_fetch` — `nodashhere`, `"   "`, `_LEADING_UNDERSCORE`, `GMAIL_`, plus the caller's-own-slug assertion. No I/O reached.
- `toolkit_for_contract_slug_accepts_real_action_slugs` — valid slug, whitespace-padded, all three multi-segment prefixes, and an assertion that the vendored helper stays permissive.
- `get_tool_contract_rejects_a_slug_with_no_action_segment` — the same rejection through the agent tool.

Ran locally:

- [x] `cargo test --lib -- flows::` — 687 passed, 0 failed
- [x] `cargo clippy -p openhuman -- -D warnings` — clean
- [x] `cargo fmt -p openhuman -- --check` — clean
- [x] `node scripts/ci/check-openhuman-rust-layout.mjs` — clean
- [x] `tsc --noEmit`, `eslint`, `prettier --check` — clean
- [x] Vitest: `flowsApi`, `approvalApi`, `useFlowPreauthorization`, `WorkflowProposalCard` — 95 passed

Not run locally, left to CI: the Tauri-shell `cargo check` (pre-push hook was skipped — long build on a shared machine) and the product-feature-set lanes. Pushed with `--no-verify` for that reason.

Closes #6071
Closes #6072

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tool contract validation for malformed slugs, including missing toolkit or action segments.
  - Preserved the original input in validation errors while using trimmed values for lookup and reporting.
  - Corrected approval-trust reporting so permissions are not shown as trusted when no approval gate is installed.
  - Improved toolkit resolution for action slugs, preventing invalid catalog lookup errors.

- **Documentation**
  - Clarified approval manifest trust-field behavior.

### DIFF
--- a/app/src/services/api/flowsApi.ts
+++ b/app/src/services/api/flowsApi.ts
@@ -834,6 +834,11 @@ export interface ApprovalManifest {
   entries: ApprovalManifestEntry[];
   /** Approvable trust keys the flow does not yet hold — what the card asks for. */
   missing: string[];
+  /**
+   * Approvable trust keys this flow already holds a grant for. Empty whenever
+   * `gate_installed` is false: with no gate no grant was ever made, so nothing
+   * can be reported as authorized.
+   */
   already_trusted: string[];
   /** False when the approval gate is disabled — nothing ever prompts. */
   gate_installed: boolean;
--- a/src/openhuman/flows/builder_tools_part_04.rs
+++ b/src/openhuman/flows/builder_tools_part_04.rs
@@ -349,10 +349,13 @@ impl Tool for GetToolContractTool {
             Some(s) if !s.is_empty() => s.to_string(),
             _ => return Ok(ToolResult::error("Missing 'slug' parameter".to_string())),
         };
-        // Contract crate — `toolkit_from_slug` is defined in
-        // `tinymemory_api::composio::scopes` and only re-exported by the engine's
-        // providers module, so this names the same function (#5560).
-        let Some(toolkit) = tinymemory_api::composio::toolkit_from_slug(&slug) else {
+        // Shape-check before any I/O, through the same guard the
+        // `flows_get_tool_contract` RPC uses — `toolkit_from_slug` falls back to
+        // the whole string when there is no `_`, so a bare action name would
+        // otherwise resolve to a bogus toolkit and come back as the catalog
+        // error below, which tells the model to retry a fetch that can never
+        // succeed.
+        let Some(toolkit) = crate::openhuman::flows::ops::toolkit_for_contract_slug(&slug) else {
             return Ok(ToolResult::error(format!(
                 "Could not extract a toolkit from slug '{slug}' — it must look like \
                  '<TOOLKIT>_<ACTION>' (e.g. 'GMAIL_SEND_EMAIL')."
--- a/src/openhuman/flows/builder_tools_tests_part_01_tests.rs
+++ b/src/openhuman/flows/builder_tools_tests_part_01_tests.rs
@@ -353,6 +353,26 @@ async fn get_tool_contract_missing_slug_is_error() {
 }
 
 #[tokio::test]
+async fn get_tool_contract_rejects_a_slug_with_no_action_segment() {
+    let tmp = TempDir::new().unwrap();
+    let tool = GetToolContractTool::new(test_config(&tmp));
+
+    // A bare toolkit-ish token names no action, so this endpoint cannot return
+    // anything for it. Before the shape guard, `toolkit_from_slug` fell back to
+    // the whole string and the model got the catalog error instead — which tells
+    // it to retry a fetch that can never succeed.
+    for slug in ["nodashhere", "_LEADING_UNDERSCORE", "GMAIL_"] {
+        let result = tool.execute(json!({ "slug": slug })).await.unwrap();
+        assert!(result.is_error, "slug {slug:?} must be rejected");
+        assert!(
+            result.output().contains("'<TOOLKIT>_<ACTION>'"),
+            "slug {slug:?} must get the malformed-slug message, got: {}",
+            result.output()
+        );
+    }
+}
+
+#[tokio::test]
 async fn get_tool_contract_rejects_a_hallucinated_slug() {
     seed_live_catalog_cache("gmail", vec![seeded_gmail_send_contract()]);
     let tmp = TempDir::new().unwrap();
--- a/src/openhuman/flows/flows_schema_part_02.rs
+++ b/src/openhuman/flows/flows_schema_part_02.rs
@@ -55,7 +55,8 @@ pub(super) fn lookup(function: &str) -> Option<ControllerSchema> {
                 FieldSchema {
                     name: "already_trusted",
                     ty: TypeSchema::Array(Box::new(TypeSchema::String)),
-                    comment: "Approvable trust keys already granted to this flow.",
+                    comment: "Approvable trust keys already granted to this flow. Empty when \
+                              gate_installed is false — no gate means no grant was ever made.",
                     required: true,
                 },
                 FieldSchema {
--- a/src/openhuman/flows/ops_part_12.rs
+++ b/src/openhuman/flows/ops_part_12.rs
@@ -367,12 +367,49 @@ pub async fn compute_approval_manifest(config: &Config, graph: &WorkflowGraph) -
     entries
 }
 
+/// Splits a manifest's approvable entries into the trust keys a run will have to
+/// ask for and the ones the flow already holds a grant for.
+///
+/// With no gate installed both lists are empty: nothing ever parks, so nothing
+/// is `missing` — and nothing was ever granted, so nothing is `already_trusted`
+/// either. Naming the approvable keys there would tell the save+enable card that
+/// permissions are authorized on a host that holds no grant for them.
+/// `gate_installed: false` is the caller's single signal, which is exactly what
+/// the sibling `approval_preauthorize_flow` reports for the same condition.
+fn split_manifest_trust(
+    entries: &[Value],
+    gate_installed: bool,
+    trusted: &HashSet<String>,
+) -> (Vec<String>, Vec<String>) {
+    let mut missing: Vec<String> = Vec::new();
+    let mut already_trusted: Vec<String> = Vec::new();
+    if !gate_installed {
+        return (missing, already_trusted);
+    }
+    for entry in entries {
+        if entry.get("kind").and_then(Value::as_str) != Some("approvable") {
+            continue;
+        }
+        let Some(tool_name) = entry.get("tool_name").and_then(Value::as_str) else {
+            continue;
+        };
+        if trusted.contains(tool_name) {
+            already_trusted.push(tool_name.to_string());
+        } else {
+            missing.push(tool_name.to_string());
+        }
+    }
+    (missing, already_trusted)
+}
+
 /// RPC: the approval manifest for a saved flow (by `id`) or a candidate
 /// `graph`, joined against the flow's existing `flow_tool_trust` grants so
 /// the save+enable card can ask only for what's missing.
 ///
 /// With the approval gate uninstalled (`OPENHUMAN_APPROVAL_GATE=0`) nothing
-/// ever parks, so `missing` is empty by definition and the card never shows.
+/// ever parks and nothing was ever granted, so `missing` and `already_trusted`
+/// are both empty by definition and the card never shows — `gate_installed:
+/// false` is the only signal in that case.
 pub async fn flows_approval_manifest(
     config: &Config,
     id: Option<&str>,
@@ -404,24 +441,7 @@ pub async fn flows_approval_manifest(
         _ => HashSet::new(),
     };
 
-    let mut missing: Vec<String> = Vec::new();
-    let mut already_trusted: Vec<String> = Vec::new();
-    for entry in &entries {
-        if entry.get("kind").and_then(Value::as_str) != Some("approvable") {
-            continue;
-        }
-        let Some(tool_name) = entry.get("tool_name").and_then(Value::as_str) else {
-            continue;
-        };
-        if !gate_installed {
-            // Nothing parks without a gate; report nothing as missing.
-            already_trusted.push(tool_name.to_string());
-        } else if trusted.contains(tool_name) {
-            already_trusted.push(tool_name.to_string());
-        } else {
-            missing.push(tool_name.to_string());
-        }
-    }
+    let (missing, already_trusted) = split_manifest_trust(&entries, gate_installed, &trusted);
 
     let log = format!(
         "[flows] approval manifest: {} entr{}, {} missing grant(s)",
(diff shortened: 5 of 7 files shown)


YellowSnnowmann:
@coderabbitai review

M3gA-Mind (approved):

PR #6101, opened 2026-09-08
fix(config): resolve OPENHUMAN_WORKSPACE inside .openhuman instead of discarding config.toml

### LINKED ISSUE #6079: `OPENHUMAN_WORKSPACE` pointing inside `.openhuman` silently discards the whole `config.toml`
## Summary

`OPENHUMAN_WORKSPACE` pointing inside `.openhuman` silently discards the whole `config.toml`

- **Namespace / method:** not one controller — every controller that reads config.
Observed through `openhuman.cost_get_dashboard`, `openhuman.dashboard_model_health`,
`openhuman.security_policy_info` and `openhuman.update_run`.
- **Where:** `src/openhuman/config/schema/load/dirs.rs:259-288`
(`resolve_config_dir_for_workspace`), reached from `:386-395`.

## Problem

**Expected**

setting `OPENHUMAN_WORKSPACE` relocates the workspace. If the
config directory it implies has no `config.toml`, the user is told — the whole
configuration silently reverting to schema defaults is not a state anyone can
debug from the outside.

**Actual**

the config dir is derived as `workspace.parent()/".openhuman"`:

```rust
let legacy_config_dir = workspace_dir.parent().map(|parent| parent.join(".openhuman"));
if let Some(legacy_dir) = legacy_config_dir {
    if legacy_dir.join("config.toml").exists() { return (legacy_dir, workspace_config_dir); }
    if workspace_dir.file_name().is_some_and(|name| name == OsStr::new("workspace")) {
        return (legacy_dir, workspace_config_dir);      // <-- returns a dir that may not exist
    }
}
```

The second `return` fires on **any** path ending in `workspace`, without checking
that `legacy_dir` exists. Set `OPENHUMAN_WORKSPACE=~/.openhuman/workspace` — a
natural thing to write — and the config dir resolves to
`~/.openhuman/.openhuman`, which does not exist. Every setting then falls back to
its schema default. **`workspace_dir` itself is still correct**, so data lands in
the right place and only the settings are gone, which is what makes it so hard to
see.

**Steps to reproduce**

with `~/.openhuman/config.toml` containing
`[cost] monthly_limit_usd = 10.0`:

```bash
OPENHUMAN_WORKSPACE=~/.openhuman/workspace openhuman-core rpc openhuman.cost_get_dashboard
```

`budget_limit_monthly_usd` comes back `100.0` (the default), while the seeded
`~/.openhuman/workspace/state/costs.jsonl` **is** read — so the response is a
convincing mix of live data and default policy. No warning is logged: the
`EnvWorkspace` branch at `:386` emits nothing, and the loader treats an absent
`config.toml` as "use defaults".

**Cited location still accurate:** yes — `resolve_config_dir_for_workspace` is
`src/openhuman/config/schema/load/dirs.rs:259-288`; the unguarded second `return` is at
`:276-281`. The `EnvWorkspace` branch is `resolve_runtime_config_dirs_with` at `:385-397`
and, as claimed, it logs nothing — it calls the resolver and returns.

**Repro run:** the report's CLI repro needs a binary build, so I re-ran the resolver's
logic instead: transcribed `dirs.rs:259-288` verbatim into a standalone program with
`.exists()` injected so that **only** the real `~/.openhuman/config.toml` exists, then
`rustc --edition 2021 -O` and ran it.

**Observed**

```
OPENHUMAN_WORKSPACE = /home/test/.openhuman/workspace
  -> config_dir    = /home/test/.openhuman/.openhuman
  -> workspace_dir = /home/test/.openhuman/workspace
  config_dir/config.toml exists? false
  real config at /home/test/.openhuman/config.toml was NOT loaded: true
```
Exactly the reported shape: the doubled `.openhuman/.openhuman` config dir, a
`workspace_dir` that is still correct, and the real `config.toml` never reached.

**Diagnosis assessment:** correct, and the mechanism is precisely as described — the
`file_name() == "workspace"` arm at `:276-281` returns `legacy_dir` without ever asking
whether it exists, having already established two lines above that it contains no
`config.toml`.

**Is the expected behaviour correct:** yes. **And there is a second finding the report did
not have:** the existing unit test
`src/openhuman/config/schema/load_tests_part_01_tests.rs:556` —
`resolve_config_dir_for_workspace_returns_parent_and_workspace` — feeds this function the
*exact broken input* (`/home/test/.openhuman/workspace`) and passes anyway. Its comment
says "Config dir is the parent of workspace", but the assertion is
`config_dir.ends_with(".openhuman") || config_dir == PathBuf::from("/home/test/.openhuman")`,
and `Path::ends_with` compares whole components, so the doubled path satisfies the first
disjunct:
```
existing test assertion `config_dir.ends_with(".openhuman")` = true
strict form `config_dir == /home/test/.openhuman`           = false
```
The test is vacuous for this case and gives false assurance that the resolution is right.
Whoever fixes this should tighten that assertion to the strict form in the same change,
or the regression can come back green.

## Solution

as the report suggests — in the
`file_name() == "workspace"` arm return `legacy_dir` only when it exists, otherwise fall
through to the final tuple; and log a `warn` at the `EnvWorkspace` branch naming the
chosen config dir and whether a `config.toml` was found in it. Tighten
`load_tests_part_01_tests.rs:556` to `assert_eq!` at the same time.

**Blast radius:** wide — every controller that reads config. Three call sites for the
resolver: `dirs.rs:389` (`resolve_runtime_config_dirs_with`, the `OPENHUMAN_WORKSPACE`
path), `load/env_overlay_impl_01_part_01.rs:57`, and
`inference/embeddings/cloud_adapter.rs:122` (which documents relying on the derived dir
rather than the raw env value, so it inherits the same skew). Tests that touch it:
`load_tests_part_01_tests.rs:512,556` and `load_tests_part_03_tests.rs:55-74` — the latter
uses a basename that is *not* `workspace`, so it exercises the safe path only.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- `OPENHUMAN_WORKSPACE` pointing inside the config dir (e.g. `~/.openhuman/workspace`) no longer silently discards the whole `config.toml`.
- `resolve_config_dir_for_workspace` now recognises the modern default layout (workspace named `workspace` whose parent is the `.openhuman` config dir) and resolves the config dir to that parent — loading the real `config.toml`.
- The legacy `workspace`-basename arm now only returns the sibling `.openhuman` when it actually exists (never a fabricated doubled path).
- The `OPENHUMAN_WORKSPACE` branch now logs which config dir it chose and whether a `config.toml` was found (warn when absent).

## Problem

`resolve_config_dir_for_workspace` derived the config dir as `workspace.parent()/".openhuman"` and, for any path ending in `workspace`, returned it **without checking it exists**. So `OPENHUMAN_WORKSPACE=~/.openhuman/workspace` resolved to a doubled, non-existent `~/.openhuman/.openhuman`; the real `~/.openhuman/config.toml` was never read and every setting silently reverted to schema defaults, with **no warning**. `workspace_dir` itself stayed correct, so data landed in the right place — which is what made the misconfiguration invisible and undebuggable from the outside. The existing unit test was vacuous: `Path::ends_with(".openhuman")` matches whole components, so the doubled path passed it.

## Solution

- Add a pure path-structure branch: when the basename is `workspace` and the parent's basename is the config-dir name (`default_root_dir_name()`, so staging is handled too), the config dir **is** that parent. No `.exists()` probe, so the natural `~/.openhuman/workspace` case resolves correctly.
- Guard the legacy sibling arm on `legacy_dir.exists()`; otherwise fall through to the safe default tuple.
- Add `tracing::debug!`/`warn!` in the `EnvWorkspace` branch naming the chosen config dir and whether `config.toml` is present (matches the file's existing `tracing::` idiom; paths only, no secrets).
- Tighten the vacuous test to `assert_eq!(config_dir, "/home/test/.openhuman")` and add regression tests for the modern layout, legacy layout (real temp dir), and the fall-through case. One existing `cloud_adapter_tests.rs` test that asserted the buggy doubled-path behaviour is rewritten to use a real sibling dir (preserving the credential-scope invariant it guards).

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — modern/legacy/fall-through resolution + tightened vacuous test.
- [x] **Diff coverage ≥ 80%** — changed lines are `resolve_config_dir_for_workspace` + the env-branch log, all exercised by the new/updated unit tests. (Full `diff-cover` runs in CI.)
- [x] All affected feature IDs listed under `## Related` — `N/A: no matrix feature IDs affected`.

## Impact

- Affects every controller that reads config when `OPENHUMAN_WORKSPACE` is set. Preserves resolution for the self-contained and legacy-sibling layouts; only the previously-broken modern-layout / non-existent-sibling cases change (now correct). No wire/schema change.

## Related

- Closes: #6079
- Follow-up PR(s)/TODOs: legacy sibling arm still uses a literal `.openhuman` (pre-existing; would miss `.openhuman-staging` for a non-workspace legacy path) — noted, out of scope.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/workspace-config-dir-discard-6079

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected configuration directory resolution for modern `.openhuman/workspace` layouts.
  - Prevented duplicated paths when resolving legacy workspace configuration directories that do not yet exist.
  - Added warnings when a workspace-resolved configuration directory does not contain `config.toml`.

- **Tests**
  - Expanded coverage for modern and legacy workspace layouts.
  - Verified resolution using real temporary directories and configuration files.
  - Added coverage for unset or empty workspace configuration settings.

### DIFF
--- a/src/openhuman/config/mod.rs
+++ b/src/openhuman/config/mod.rs
@@ -34,6 +34,12 @@ pub use schema::{
 pub use workspace_handle::workspace_handle;
 // Crate-internal: workspace→config-dir resolver reused by the cloud embedder.
 pub(crate) use schema::resolve_config_dir_for_workspace;
+// Test-only: the `.openhuman` (or `.openhuman-staging`) root dir name the modern
+// layout keys on. The `desktop::app_state` resolver tests build tempdir
+// workspaces named after it so the modern-layout arm fires regardless of the
+// ambient `OPENHUMAN_APP_ENV`.
+#[cfg(test)]
+pub(crate) use schema::default_root_dir_name;
 pub(crate) use schema::set_cli_inference_overrides;
 #[allow(unused_imports)]
 pub use schema::{
--- a/src/openhuman/config/schema/load/dirs.rs
+++ b/src/openhuman/config/schema/load/dirs.rs
@@ -290,17 +290,18 @@ pub(crate) fn resolve_config_dir_for_workspace(workspace_dir: &Path) -> (PathBuf
 
     let legacy_config_dir = parent.map(|parent| parent.join(".openhuman"));
     if let Some(legacy_dir) = legacy_config_dir {
-        // Legacy sibling layout: `<something>/workspace` with config living in a
-        // sibling `<something>/.openhuman`.
-        if legacy_dir.join("config.toml").exists() {
-            return (legacy_dir, workspace_config_dir);
-        }
-
-        // Same layout without a config.toml yet, but only when the sibling
-        // `.openhuman` actually exists — never return a non-existent doubled
-        // path (the original #6079 bug: this arm fired for any `workspace`
-        // basename, so `~/.openhuman/workspace` produced `~/.openhuman/.openhuman`).
-        if has_workspace_basename && legacy_dir.exists() {
+        // Legacy sibling layout: `<proj>/workspace` with config living in a
+        // sibling `<proj>/.openhuman`. Return the sibling unconditionally,
+        // including on a fresh volume where `<proj>/.openhuman` does not exist
+        // yet — that is where `config::load` writes and reads config for this
+        // layout, so nesting the workspace inside itself (the fall-through
+        // below) would strand it.
+        //
+        // This arm can no longer reintroduce the #6079 doubling: the modern
+        // layout above already intercepts `~/.openhuman/workspace` (parent IS
+        // the config dir) before control reaches here, so a `workspace`
+        // basename whose parent is the config dir never falls into this arm.
+        if legacy_dir.join("config.toml").exists() || has_workspace_basename {
             return (legacy_dir, workspace_config_dir);
         }
     }
--- a/src/openhuman/config/schema/load/dirs.rs
+++ b/src/openhuman/config/schema/load/dirs.rs
@@ -265,18 +265,42 @@ pub(crate) fn resolve_config_dir_for_workspace(workspace_dir: &Path) -> (PathBuf
         );
     }
 
-    let legacy_config_dir = workspace_dir
-        .parent()
-        .map(|parent| parent.join(".openhuman"));
+    let has_workspace_basename = workspace_dir
+        .file_name()
+        .is_some_and(|name| name == std::ffi::OsStr::new("workspace"));
+    let parent = workspace_dir.parent();
+
+    // Modern default layout: the workspace lives *inside* the `.openhuman`
+    // config dir (`~/.openhuman/workspace`), so the parent IS the config dir.
+    // `default_config_and_workspace_dirs` produces exactly this shape. This is
+    // a pure path-structure decision — no `.exists()` probe — so pointing
+    // `OPENHUMAN_WORKSPACE` at `~/.openhuman/workspace` resolves to
+    // `~/.openhuman` and loads the real `~/.openhuman/config.toml` instead of
+    // the doubled, non-existent `~/.openhuman/.openhuman` (#6079).
+    if has_workspace_basename {
+        if let Some(parent) = parent {
+            if parent
+                .file_name()
+                .is_some_and(|name| name == std::ffi::OsStr::new(default_root_dir_name()))
+            {
+                return (parent.to_path_buf(), workspace_config_dir);
+            }
+        }
+    }
+
+    let legacy_config_dir = parent.map(|parent| parent.join(".openhuman"));
     if let Some(legacy_dir) = legacy_config_dir {
+        // Legacy sibling layout: `<something>/workspace` with config living in a
+        // sibling `<something>/.openhuman`.
         if legacy_dir.join("config.toml").exists() {
             return (legacy_dir, workspace_config_dir);
         }
 
-        if workspace_dir
-            .file_name()
-            .is_some_and(|name| name == std::ffi::OsStr::new("workspace"))
-        {
+        // Same layout without a config.toml yet, but only when the sibling
+        // `.openhuman` actually exists — never return a non-existent doubled
+        // path (the original #6079 bug: this arm fired for any `workspace`
+        // basename, so `~/.openhuman/workspace` produced `~/.openhuman/.openhuman`).
+        if has_workspace_basename && legacy_dir.exists() {
             return (legacy_dir, workspace_config_dir);
         }
     }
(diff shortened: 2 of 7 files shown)


M3gA-Mind:
Review from the e2e wave side — I hold the independent verification for #6079, so I walked its blast radius against this diff rather than re-reading the report.

**The fix is right, and it is right in a way the issue did not ask for.** The shape our verification proposed (guard the `workspace`-basename arm on existence, fall through, add a warn) would *not* have closed the issue's own repro: config would still land on a directory with no `config.toml`, so `budget_limit_monthly_usd` still comes back `100.0` and only a log line changes. The new modern-layout branch is what actually closes it. Deciding it on `parent.file_name() == default_root_dir_name()` as a pure path rule is also better than probing `parent/config.toml`, which is what I would have written — it works for non-existent paths and picks up `.openhuman-staging`. Tightening the vacuous `ends_with` assertion to `assert_eq!` is exactly what verification asked for.

**One substantive concern: the `.exists()` guard moves `workspace_dir`, not just `config_dir`.**

For a legacy `<X>/workspace` layout where `<X>/.openhuman` is absent at resolve time, the function now falls through to the final tuple, so:

- `config_dir`: `<X>/.openhuman` → `<X>/workspace`
- `workspace_dir`: `<X>/workspace` → `<X>/workspace/workspace`

The second one is a data-location change. And `config_dir` here is not just settings — `inference/embeddings/cloud_adapter.rs:120-124` uses it as the **credential scope** (`auth-profiles.json`). The doc comment on the test you had to rewrite says regressing that "reintroduce[s] 'No backend session' for that deployment"; needing to create the dirs in `cloud_adapter_tests.rs` to keep it green is the tell that the rule stopped being purely structural.

In steady state `<X>/.openhuman` exists and this is fine. The exposure is first run, or a container where the volume mounts after the resolver runs.

Worth noting the inconsistency: the new modern-layout branch is deliberately documented as "a pure path-structure decision — no `.exists()` probe", while the legacy branch just gained one. CodeRabbit's `is_dir()` suggestion tightens the same probe, so it is worth settling whether the probe belongs there at all before acting on it.

**Suggestion:** in the `has_workspace_basename` case, never fall through to the final tuple — return `(legacy_dir, workspace_config_dir)` and let the new `warn!` carry the "no `config.toml` here" signal. That keeps `workspace_dir` stable in every branch, which is the property that matters for where data lands, and still removes the silent-defaults failure the issue is about.

**Also checked, no action needed:** the third caller from the blast radius, `config/schema/load/env_overlay_impl_01_part_01.rs:57`, inherits the fix and needs no change; `load_tests_part_03_tests.rs:55-74` uses a non-`workspace` basename so it only exercises the untouched path.

YellowSnnowmann:
Thanks — both bot findings were valid; addressed at root cause (not a doc patch):

- **CodeRabbit (`dirs.rs` legacy `exists()` accepts a file named `.openhuman`)** and **Codex P2 (fresh-legacy layout regressed to a nested workspace):** the `exists()` guard was an over-correction. The new modern-layout branch already intercepts the only case that produced the doubled `~/.openhuman/.openhuman`, so the guard was unnecessary. Reverted the legacy `workspace`-basename arm to its original **unconditional** return — a fresh `<proj>/workspace` again resolves to the sibling `<proj>/.openhuman` (workspace preserved), and no bare `exists()` probe remains. Cases verified: `~/.openhuman/workspace`→parent (no doubling), `<proj>/workspace` fresh+with-config→sibling, `.openhuman-staging/workspace`→parent, and true fall-through.
- **Codex P1 (twin resolver):** confirmed — `desktop/app_state/ops_part_01.rs::config_dir_for_workspace_env` reimplemented the same mapping and produced the identical doubled path, mis-scoping revalidated-session credentials. It now delegates to the shared `resolve_config_dir_for_workspace`, so both agree; regression tests added there.

Regression tests updated/added for all of the above; `cargo fmt` + lib `clippy -D warnings` + the config/app_state unit suites pass locally.

YellowSnnowmann:
Addressed the test-hardening findings on the new `config_dir_for_workspace_env` tests:

- **tinysweeper (restore `OPENHUMAN_WORKSPACE` on panic):** added a `#[cfg(test)]` RAII `WorkspaceEnvGuard` that captures the prior value on construction and restores it on `Drop` (set / empty-string / unset branches), so a panic can no longer leak the env var into sibling tests. The existing `WORKSPACE_ENV_TEST_LOCK` still serializes the mutations. (The pre-existing guards in `config/workspace` / `notifications` only `remove_var` on drop, so they don't restore the prior value — hence a small new guard rather than reusing them.)
- **CodeRabbit (isolated temp paths):** the tests now use `tempfile::tempdir()` roots instead of fixed `/home/test/...`, so a real host directory can't flip a resolver branch. The modern-layout root is named via `default_root_dir_name()` so the modern arm fires regardless of `OPENHUMAN_APP_ENV`. All three tests still pin the same #6079 resolution (no doubling; fresh-legacy → sibling; unset/empty → `None`).

`cargo fmt` + lib `clippy -D warnings` + the `desktop::app_state` suite pass locally.

YellowSnnowmann:
@coderabbitai review

Issue #6073, opened 2026-09-08
Handler-level "missing required param" strings are unreachable — every dispatch is schema-validated first


M3gA-Mind:
## Summary

Handler-level "missing required param" strings are unreachable — every dispatch is schema-validated first

Found during the e2e coverage wave and **independently verified by a second engineer** who did not
write the original report — verification was cross-assigned so nobody checked their own finding.

- **Namespace / method:** cross-cutting; observed while covering `learning`,
`memory_goals`, `session_db`, `session_import`, `tree_summarizer`
- **Where:** `src/core/all.rs:1334` (`validate_params`) versus the per-handler
checks, e.g. `agent/learning/schemas_part_02.rs:57`
(``missing required `class` ``), `memory/goals/schemas.rs:236`
(`parse_value` → `"invalid params: …"`), `agent/session_db/schemas.rs:96`
(`"missing required param: id"`).
- **Verified at:** `openhuman` `main` @ `d1ed68f9b`

## Problem

**Expected**

**Actual**

**Steps to reproduce**

`openhuman.session_db_get` with `{}` →
`missing required param 'id': Session ID.` (the schema's comment), never the
handler's `missing required param: id`.

### Independent verification (CONFIRMED — and the original report's own *rationale* is partly wrong (below))

**Cited location still accurate:** yes. `core/all.rs:1334` is `pub fn validate_params`, with
the required-presence check at `:1338-1344`, the unknown-param check at `:1346-1353`, and
the type check at `:1358-1371`. The three handler sites cited are also still there
(`agent/learning/schemas_part_02.rs:57`, `memory/goals/schemas.rs:236`,
`agent/session_db/schemas.rs:96`).

**Repro run:** dispatched four controllers over the real router with params that trip each
handler's *own* check, asserting no handler-style message can reach a caller:
```
cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" \
    -- --test-threads=1 --nocapture w3_verify_scratch
```

**Observed**

```
openhuman.learning_get_facet
     observed: missing required param 'class': Facet class: style | identity | tooling | veto | goal | channel.
     handler says: missing required `class`
openhuman.session_db_get
     observed: missing required param 'id': Session ID.
     handler says: missing required param: id
openhuman.memory_goals_add
     observed: missing required param 'text': The goal text — one concise sentence.
     handler says: invalid params: ...
openhuman.session_import_run
[...]

PR #6102, opened 2026-09-08
fix(sandbox): resolve policy/status from the loaded config, not RuntimeConfig::default()

### LINKED ISSUE #6081: `sandbox_resolve_policy` / `sandbox_status` answer from a default config, never the user's
## Summary

`sandbox_resolve_policy` / `sandbox_status` answer from a default config, never the user's

- **Namespace / method:** `openhuman.sandbox_resolve_policy`, `openhuman.sandbox_status`
- **Where:** `src/openhuman/sandbox/schemas.rs` — `handle_resolve_policy` and `handle_status`

## Problem

**Expected**

The policy an RPC caller is shown is the policy an agent session
would actually get: resolved against the **loaded** `Config` (its
`runtime` block and its `action_dir`).

**Actual**

Both handlers build `crate::openhuman::config::RuntimeConfig::default()`
and never load the user's config at all.
- `handle_resolve_policy` anchors on `crate::openhuman::config::default_action_dir()`,
  ignoring `OPENHUMAN_ACTION_DIR` / `action_dir_override` / the opened project.
- `handle_status` hardcodes `std::path::Path::new("/tmp")` as the workspace root
  and derives `is_remote` from the *backend name string* the caller passed
  (`backend == "docker"`), which is not what `is_remote_session` means anywhere
  else in the codebase.

Consequences:
1. A user with `[runtime] kind = "docker"` asks `sandbox_resolve_policy` for
   `sandbox_mode = "sandboxed"` on a local session and is told `backend: "local"`.
   `ops::resolve_sandbox_policy` would have returned `docker` for that config —
   the branch at `sandbox/ops.rs:39` is `runtime_config.kind == "docker" || is_remote_session`,
   and `kind` here is always the default.
2. Every `docker_overrides` field reported (`image`, `network`, `memory_limit_mb`,
   `cpu_limit`, `read_only_rootfs`) is the compiled-in default, never the user's
   `[runtime.docker]` block. A UI rendering this shows the wrong image and the
   wrong resource caps.
3. `workspace_root` is reported as the process default action dir (resolve_policy)
   or literally `/tmp` (status), not the mounted project.

**Steps to reproduce**

`tests/raw_coverage/sandbox_runtime_platform_e2e.rs::sandbox_resolve_policy_maps_each_mode_and_remote_flag_to_a_backend`
passes today *because* the test config leaves `[runtime]` at its default — which
is the point: no assertion in the suite can distinguish "resolved correctly" from
"resolved against defaults". To see it, add `[runtime]\nkind = "docker"` to the
harness config and assert `backend == "docker"` for
`{"sandbox_mode":"sandboxed","is_remote":false}`; it returns `"local"`.

### Independent verification (CONFIRMED (with one over-stated sub-claim corrected))

**Cited location still accurate:** yes — `src/openhuman/sandbox/schemas.rs`,
`handle_status` at `:135-160` and `handle_resolve_policy` at `:162-184`. Both build
`RuntimeConfig::default()` (`:149`, `:179`) and never call any config loader.

**Repro run:** static trace of the three cited consequences against
`sandbox/ops.rs:29-59 resolve_sandbox_policy` and `config/schema/runtime.rs`.

**Observed**

```
schemas.rs:149   let config = crate::openhuman::config::RuntimeConfig::default();
schemas.rs:150   let is_remote = backend_str == "docker";
schemas.rs:153   std::path::Path::new("/tmp"),
schemas.rs:179   let config = crate::openhuman::config::RuntimeConfig::default();
schemas.rs:180   let action_dir = crate::openhuman::config::default_action_dir();

config/schema/runtime.rs:55-57
    fn default_runtime_kind() -> String { "native".into() }

sandbox/ops.rs:39
    if runtime_config.kind == "docker" || is_remote_session { Docker } else { Local }
```
`RuntimeConfig::default().kind == "native"`, so the left disjunct at `ops.rs:39` is
dead for both handlers — consequence 1 holds exactly as written. `docker_overrides`
(`ops.rs:47-59`) reads `runtime_config.docker`, so consequence 2 holds: every field
reported is the compiled-in default (`image: "alpine:3.20"`, etc.), never the user's
`[runtime.docker]`.

**Diagnosis assessment:** correct in substance. **One sub-claim is over-stated and should
not be repeated when this is filed:** the report says `handle_resolve_policy`
*"anchors on `default_action_dir()`, ignoring `OPENHUMAN_ACTION_DIR` /
`action_dir_override` / the opened project."* `default_action_dir()`
(`config/schema/load/dirs.rs:127-135`) **does** honour `OPENHUMAN_ACTION_DIR` — it reads
`ACTION_DIR_ENV_VAR` (`:34 = "OPENHUMAN_ACTION_DIR"`) first. What it skips is the
persisted `action_dir_override` and the opened project; the correct resolver is
`resolve_action_dir(&config.action_dir_override)` (`:111`, documented at `:108` as
"env `OPENHUMAN_ACTION_DIR` > persisted `action_dir_override` > default"). So consequence 3
is real but one of its three named inputs is in fact respected.

**Is the expected behaviour correct:** yes. Neither schema description
(`schemas.rs:41-57`, `:58-…` — *"Return sandbox backend status and availability"* /
*"Resolve sandbox policy for a given sandbox mode and session context"*) frames these as
hypothetical calculators over caller-supplied inputs, and `[runtime.docker]` has **no**
caller-supplied channel at all — the caller cannot pass an image or a memory cap — so
the only possible source for those outputs is the loaded config. Sibling controllers in
this codebase load it routinely (e.g. `mcp/audit/schemas.rs` via
`config_rpc::load_config_with_timeout()`), which is what makes this an oversight rather
than a design.

## Solution

load the config in both handlers
(`config_rpc::load_config_with_timeout()`), pass `&config.runtime` and
`resolve_action_dir(&config.action_dir_override)` into `resolve_sandbox_policy`, and use
`config.workspace_dir` instead of the literal `/tmp`. `handle_status`'s
`is_remote = backend_str == "docker"` is a separate conflation — it makes "the caller
named the docker backend" mean "this is a remote session", which is not what
`is_remote_session` means at `ops.rs:39` or anywhere else — and should become an explicit
input or be dropped.

**Blast radius:** `resolve_sandbox_policy` is also called by the agent harness path, which
already passes a real config — only the two RPC handlers are wrong, so fixing them cannot
change agent behaviour. Consumers: any UI rendering `docker_overrides` or `workspace_root`
from these replies. W6's
`tests/raw_coverage/sandbox_runtime_platform_e2e.rs::sandbox_resolve_policy_maps_each_mode_and_remote_flag_to_a_backend`
passes only because its harness leaves `[runtime]` default; it pins nothing and will need
a `[runtime] kind = "docker"` case once fixed.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- `sandbox_status` and `sandbox_resolve_policy` now answer from the user's **loaded** config instead of `RuntimeConfig::default()`.
- A user with `[runtime] kind = "docker"` is now correctly told `backend: "docker"`, and `docker_overrides` reflect the user's `[runtime.docker]` block.
- `resolve_policy` resolves the action dir via `resolve_action_dir(&config.action_dir_override)`; `status` reports `config.workspace_dir` instead of the literal `/tmp`.
- `handle_status`'s `is_remote` is now an explicit optional input (documented in the schema), not inferred from the caller-supplied backend name.

## Problem

Both handlers built `crate::openhuman::config::RuntimeConfig::default()` and never loaded the user's config. Consequences: the docker branch in `resolve_sandbox_policy` (`runtime_config.kind == "docker" || is_remote_session`) was dead for these handlers (default `kind` is `native`), so a docker-configured user was told `backend: "local"`; every `docker_overrides` field was the compiled-in default; `workspace_root` was the process default action dir (`resolve_policy`) or literally `/tmp` (`status`). `handle_status` also conflated `is_remote` with `backend_str == "docker"`, which is not what `is_remote_session` means anywhere else. Sibling controllers load config routinely (e.g. `mcp/audit/schemas.rs`), so this was an oversight, not a design.

## Solution

- Load the live config via `config_rpc::load_config_with_timeout()` in both handlers, matching the `mcp/audit` sibling's error idiom (a failed load surfaces as an `Err` to the RPC caller — an expected error, not a Sentry event).
- Pass `&config.runtime` into `resolve_sandbox_policy`; use `resolve_action_dir(&config.action_dir_override)` (`resolve_policy`) and `config.workspace_dir` (`status`).
- Replace `is_remote = backend_str == "docker"` with an explicit optional `is_remote` param (default `false`), and add a matching `is_remote` input `FieldSchema` to the `sandbox_status` schema (consistent with `resolve_policy`).
- Regression tests inject a `[runtime] kind = "docker"` config with a non-default image through a new `#[cfg(test)]` `CoreContext::for_test_with_config` seam and assert `backend == "docker"` + the custom `docker_overrides.image`; pre-existing tests are made hermetic so they no longer do live disk I/O.

Scope note: `resolve_sandbox_policy`'s agent-harness callers already pass a real config, so this cannot change agent behaviour. A related latent instance (`shell.rs`'s `run_sandboxed` still uses `RuntimeConfig::default()`) is called out for a **separate** follow-up and intentionally not changed here.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — docker-runtime resolve/status regression + hermetic pre-existing tests.
- [x] **Diff coverage ≥ 80%** — changed handler lines are exercised by the new scoped-config tests. (Full `diff-cover` runs in CI.)

## Impact

- `sandbox.*` RPC replies now reflect the user's real runtime/docker config and mounted workspace. `handle_status` callers that relied on the undocumented `backend="docker"`⇒`is_remote` conflation must now pass `is_remote` explicitly (the conflation was a bug). No agent-execution behaviour changes.

## Related

- Closes: #6081
- Follow-up PR(s)/TODOs: `tools/impl/system/shell.rs::run_sandboxed` uses `RuntimeConfig::default()` (same defect class, different path) — separate issue/PR.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/sandbox-policy-loads-config-6081

## Summary by CodeRabbit

* **New Features**
  * Sandbox status and policy resolution now honor active runtime and workspace settings, including custom Docker configuration.
  * Sandbox status requests now use an explicit remote-execution indicator for more accurate reporting.
* **Bug Fixes**
  * Configuration-loading failures are now surfaced instead of silently falling back to defaults.
  * Sandbox actions correctly respect configured workspace and action-directory settings.

### DIFF
--- a/src/core/runtime/context.rs
+++ b/src/core/runtime/context.rs
@@ -522,6 +522,33 @@ impl CoreContext {
             embedder_config: None,
         })
     }
+
+    /// Test-only constructor that carries an embedder-supplied config, so a
+    /// cross-module test can exercise the `load_config_with_timeout()` read
+    /// path (which prefers [`CoreContext::current_embedder_config`]) without a
+    /// full boot or a racy on-disk `config.toml`.
+    ///
+    /// Distinct from [`CoreContext::for_test`] — which always sets
+    /// `embedder_config: None` — so the ~30 existing `for_test` call sites are
+    /// unaffected. The workspace binding is anchored to the config's
+    /// `workspace_dir`, matching how [`CoreContext::init`] wires an
+    /// embedder-supplied config.
+    #[cfg(test)]
+    pub(crate) fn for_test_with_config(
+        domains: crate::core::runtime::DomainSet,
+        config: crate::openhuman::config::Config,
+    ) -> Arc<CoreContext> {
+        Arc::new(CoreContext {
+            host_kind: HostKind::Cli,
+            workspace_binding: RwLock::new(WorkspaceBinding {
+                workspace_dir: Some(config.workspace_dir.clone()),
+                memory_subsystem: Default::default(),
+            }),
+            domains,
+            tool_groups: Default::default(),
+            embedder_config: Some(config),
+        })
+    }
 }
 
 /// Bind the memory driver for this workspace and initialize the other
--- a/src/openhuman/sandbox/schemas.rs
+++ b/src/openhuman/sandbox/schemas.rs
@@ -4,6 +4,8 @@ use serde_json::{Map, Value};
 
 use crate::core::all::{ControllerFuture, RegisteredController};
 use crate::core::{ControllerSchema, FieldSchema, TypeSchema};
+use crate::openhuman::config::resolve_action_dir;
+use crate::openhuman::config::rpc as config_rpc;
 use crate::rpc::RpcOutcome;
 
 pub fn all_controller_schemas() -> Vec<ControllerSchema> {
@@ -42,12 +44,22 @@ pub fn schemas(function: &str) -> ControllerSchema {
             namespace: "sandbox",
             function: "status",
             description: "Return sandbox backend status and availability.",
-            inputs: vec![FieldSchema {
-                name: "backend",
-                ty: TypeSchema::String,
-                comment: "Backend kind to check: 'docker', 'local', or 'none'.",
-                required: false,
-            }],
+            inputs: vec![
+                FieldSchema {
+                    name: "backend",
+                    ty: TypeSchema::String,
+                    comment: "Backend kind to check: 'docker', 'local', or 'none'.",
+                    required: false,
+                },
+                FieldSchema {
+                    name: "is_remote",
+                    ty: TypeSchema::Bool,
+                    comment: "Whether this is a remote/channel session. When true, a \
+                              sandboxed session resolves to the Docker backend even if \
+                              the runtime is not configured for Docker. Defaults to false.",
+                    required: false,
+                },
+            ],
             outputs: vec![FieldSchema {
                 name: "status",
                 ty: TypeSchema::Json,
@@ -138,6 +150,15 @@ fn handle_status(params: Map<String, Value>) -> ControllerFuture {
             .get("backend")
             .and_then(|v| v.as_str())
             .unwrap_or("none");
+        // `is_remote` is an explicit caller-supplied flag, matching
+        // `handle_resolve_policy`. It must NOT be inferred from `backend_str`:
+        // naming the docker backend to check its availability is not the same
+        // as running a remote/channel session, and conflating the two forced
+        // Docker resolution for a local status probe (#6081).
+        let is_remote = params
+            .get("is_remote")
+            .and_then(|v| v.as_bool())
+            .unwrap_or(false);
 
         let mode = match backend_str {
             "docker" | "local" => {
@@ -146,12 +167,23 @@ fn handle_status(params: Map<String, Value>) -> ControllerFuture {
             _ => crate::openhuman::agent::harness::definition::SandboxMode::None,
         };
 
-        let config = crate::openhuman::config::RuntimeConfig::default();
-        let is_remote = backend_str == "docker";
+        // Load the user's live config so the reported policy reflects the
+        // configured runtime (`[runtime] kind` + `[runtime.docker]` overrides)
+        // and the mounted workspace, not the compiled-in `RuntimeConfig::default()`
+        // (#6081). A failed config load surfaces to the caller rather than
+        // silently answering from defaults.
+        let config = match config_rpc::load_config_with_timeout().await {
+            Ok(config) => config,
+            Err(err) => {
+                log::warn!("[sandbox] handle_status config load failed error={err}");
+                return Err(err);
+            }
+        };
+
         let policy = super::ops::resolve_sandbox_policy(
             mode,
-            std::path::Path::new("/tmp"),
-            &config,
+            &config.workspace_dir,
+            &config.runtime,
             is_remote,
         );
         let handle = super::ops::create_sandbox_backend(&policy).await;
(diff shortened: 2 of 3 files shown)


M3gA-Mind:
Review from the e2e wave side — #6081 is my finding, so I walked its blast radius against this diff.

**The fix covers all four defects in the report**, including the one I expected to be dropped: `is_remote = backend_str == "docker"` becomes an explicit schema input rather than being inferred. The issue offered "explicit input or drop it" and this is the better half. `config.runtime`, `config.workspace_dir` and `resolve_action_dir(&config.action_dir_override)` all replace the compiled-in defaults correctly, and surfacing a config-load failure instead of silently answering from defaults is the right call.

`for_test_with_config` is `#[cfg(test)]` and kept distinct from `for_test`, so the ~30 existing call sites are untouched. `handle_status_honors_loaded_docker_runtime` sending `backend = "local"` and asserting `docker` is the right shape — it proves resolution comes from the config and not from the request.

**Independent confirmation of CodeRabbit's schema-comment point, with what I think is the sharper argument: your own new test is the proof.**

`backend` no longer selects a backend. It only picks the mode — `"docker" | "local" => Sandboxed` — and the backend then comes from `runtime.kind == "docker" || is_remote`. So under a default runtime (`default_runtime_kind()` is `"native"`) with `is_remote` absent, `{"backend":"docker"}` and `{"backend":"local"}` are indistinguishable; both return `local`. That is precisely why `handle_status_honors_loaded_docker_runtime` can pass `"local"` and get `"docker"`.

The input comment in `src/openhuman/sandbox/schemas.rs` still reads *"Backend kind to check: 'docker', 'local', or 'none'."*, which promises a selection the handler does not perform. Before this PR `backend: "docker"` did force Docker, via the conflation being removed here — so the drift is new. This is the same schema/handler class as #6073. Something closer to *"whether to sandbox at all; the backend itself follows `[runtime] kind` and `is_remote`"* would match the behaviour the test pins.

**Blast radius walked, nothing else to change:**

- No `app/` consumer of either RPC — the frontend never calls them.
- The only prose reference is `platform/about_app/catalog_part_03.rs:343-344`, which tells the user to inspect these RPCs to see their configured runtime. That becomes *more* accurate after this change, not less.
- The merged e2e suite `tests/raw_coverage/sandbox_runtime_platform_e2e.rs` still passes: it asserts `sandboxed + is_remote:true → docker` and `is_remote:false → local`, both unchanged under a `native` default runtime, and it deliberately never asserts on `backend: "docker"` (it avoided touching the docker daemon). The issue predicted that suite would need a `[runtime] kind = "docker"` case — you added the equivalent as unit tests against an injected config, which is cheaper and does not require a daemon. Good trade.

YellowSnnowmann:
Thanks — both valid; addressed without reintroducing the #6081 conflation:

- **`backend` input contract (CodeRabbit + Codex):** `handle_status` maps both `docker` and `local` to `SandboxMode::Sandboxed`; the concrete backend is resolved from the loaded runtime config (+ `is_remote`) so the reply reflects what a real agent session would actually get. That is the fix, so I did **not** re-add "honor the requested backend" (that was the bug). Instead the schema `backend` field comment now states it selects the sandbox **mode** (docker/local→sandboxed, none→none) and that the backend comes from the loaded config — the doc now matches behaviour.
- **Codex P2 (`action_dir`):** confirmed against the loader — `load_config_with_timeout()` returns `config.action_dir` fully resolved on every path (disk: `resolve_action_dir` + env overlay; embedder: verbatim, honoring `CoreBuilder::action_dir`). `resolve_action_dir(&config.action_dir_override)` ignored a directly-set `action_dir`, so `resolve_policy` now passes `&config.action_dir`. Regression test added pinning that an embedder-set `action_dir` flows into `workspace_root`.

`cargo fmt` + lib `clippy -D warnings` + the sandbox unit suite pass locally; the #6081 docker-runtime regression tests are unchanged and green.

PR #6103, opened 2026-09-08
fix(socket): gate emit on Connected so pre-handshake sends are not falsely acked

### LINKED ISSUE #6084: `socket_emit` returns `{"ok": true}` for a message the loop then drops
## Summary

`socket_emit` returns `{"ok": true}` for a message the loop then drops

- **Namespace / method:** `openhuman.socket_emit`
- **Where:** `src/openhuman/platform/socket/manager.rs:302-309` (`emit`),
`src/openhuman/platform/socket/ws_loop_part_01.rs:166-176` (`drain_pending_emits`)

## Problem

**Expected**

a caller can tell whether the event was sent. Today the only two answers are
`Err("Not connected")` — when `connect` has never been called — and `{"ok": true}`.

**Actual**

`emit` checks only whether `emit_tx` is `Some`, which `spawn_loop` sets *before* any
handshake. So between `socket_connect` returning `Connecting` and a successful SIO CONNECT ACK,
every `socket_emit` returns `{"ok": true}`. If the connection then fails — which is the whole
reason the manager is in `Connecting` rather than `Connected` — `ws_loop` calls
`drain_pending_emits` and discards the queued messages with
`log::warn!("[socket] Dropped {dropped} queued emit(s) on disconnect")`.

The drop itself is deliberate and correct (#4355 — a queued emit must not be flushed onto the
*next* socket's sid). The gap is that the RPC already told the caller `ok: true`, and there is
no ack, no error and no counter it can consult afterwards. The observable difference between
"sent" and "dropped" exists only in a `warn!` log line.

**Steps to reproduce**

`socket_connect` to a closed loopback port, then `socket_emit` → `{"ok": true}`;
the core log shows `Dropped 1 queued emit(s) on disconnect` a moment later.

**Cited location still accurate:** yes on both counts. `emit` is at `manager.rs:302-309` and its
whole guard is `if let Some(ref tx) = *self.emit_tx.lock().await { … tx.send(msg) }`.
`drain_pending_emits` is at `ws_loop_part_01.rs:427-433`, called at `:173`, with the
`log::warn!("[socket] Dropped {dropped} queued emit(s) on disconnect")` at `:175`.

**Repro run:** a probe that connects to a closed loopback port and emits, over the real RPC
router (`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)"
-- --nocapture probe_bug3`)

**Observed**

```
BUG3 emit BEFORE connect      -> {"error":{"code":-32000,"message":"Not connected"},"id":10}
BUG3 connect                  -> {"status":"Connecting"}
BUG3 emit WHILE CONNECTING    -> {"id":12,"jsonrpc":"2.0","result":{"ok":true}}
BUG3 state after handshake    -> {"socket_id":null,"status":"disconnected"}
```
The exact three-step sequence the report describes: a hard error before `connect`, `ok: true`
during `Connecting`, and a socket that never reached `Connected`. The message emitted in step 3
cannot have been delivered — there was never a peer.

**Diagnosis assessment:** right, and I confirmed the ordering claim that the whole finding rests
on. In `spawn_loop` (`manager.rs:251-280`) the sequence is: set `status = Connecting` (`:260`),
set `*self.emit_tx = Some(emit_tx)` (`:268`), *then* spawn `ws_loop` (`:274-276`). So `emit_tx`
is populated strictly before any handshake is attempted, which is precisely why `emit`'s guard
passes in a state where nothing can be sent. The characterisation of `emit_with_ack` is also
fair: it shares the same `emit_tx` guard, but because it awaits an ack with a timeout it returns
`Err` instead of a false success.

**Is the expected behaviour correct:** yes. "A caller can tell whether the event was sent" is the
minimum contract for an endpoint whose only failure answer today is `Not connected`. The report
is also careful and correct not to attack the drop itself — discarding queued emits on
disconnect is deliberate (#4355) and must stay; the defect is reporting success beforehand.

## Solution

make `emit`'s guard reflect readiness rather than channel
existence — gate on `status == Connected` (the state `ws_loop` sets after the SIO CONNECT ACK)
rather than on `emit_tx.is_some()`, so a pre-handshake emit returns the same `Not connected`
error it already returns before `connect`. A queue-and-report variant is possible but larger.

**Blast radius:** **wider than the report states, and this is the part worth carrying into
triage.** `SocketManager::emit` is not RPC-only; its production callers are
`security/devices/tunnel_client.rs:251` (`tunnel:connect`) and `:281` (`tunnel:frame`),
`platform/socket/event_handlers.rs:137` (`webhook:response`),
`platform/socket/medulla/mod.rs:642` and `:665`, and the RPC handler at
`platform/socket/schemas.rs:206`. A dropped `webhook:response` means an inbound webhook is
answered by nobody, and a dropped `tunnel:frame` is lost tunnel payload — so the silent-loss
window costs more than one misleading RPC reply. Tightening the guard would make those call
sites start seeing `Err` where they currently see `Ok`; each already handles an error
(`if let Err(e) = mgr.emit(...)`), so the change surfaces failures rather than introducing them.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- `socket_emit` no longer returns `{ok: true}` for a message the loop then silently drops.
- `SocketManager::emit` is gated on connection **readiness** (`status == Connected`), not merely on the emit channel existing.
- A pre-handshake emit now returns the same `"Not connected"` error it already returns before `connect` was ever called.

## Problem

`emit` guarded only on `emit_tx.is_some()`, which `spawn_loop` sets **before** the Socket.IO handshake completes. So between `socket_connect` returning `Connecting` and a successful SIO CONNECT ACK, every `socket_emit` returned `{ok: true}` — even though, if the connection then failed, `ws_loop`'s `drain_pending_emits` discards the queued messages (deliberate, #4355). The caller was told success for a message that was never delivered, with no ack, error, or counter to consult. This is not RPC-only: production callers include `tunnel:connect`/`tunnel:frame`, `webhook:response`, and the medulla emitters, so the silent-loss window costs more than one misleading RPC reply.

## Solution

- Add a readiness check at the top of `emit`: `if !self.is_connected() { return Err("Not connected") }` before locking/sending (reusing the existing `is_connected()`, whose `#[allow(dead_code)]` is now removed since it has a real caller). The exact error string is preserved for backward compatibility with the existing pre-connect path.
- `drain_pending_emits` is unchanged (dropping queued emits on disconnect is correct). `emit_with_ack` is intentionally left as-is and documented: it keeps queue-then-confirm while `Connecting` because its ack/timeout confirms delivery, so it cannot report a false success the way a bare `emit` could.
- All production `.emit(` callers already handle the `Err` return, so this surfaces failures rather than introducing them.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — emit-while-`Connecting` returns `Err` and queues nothing even with the channel present; emit-when-`Connected` still encodes and queues.
- [x] **Diff coverage ≥ 80%** — the changed guard lines are covered by the new manager tests. (Full `diff-cover` runs in CI.)

## Impact

- `socket_emit` (and every `SocketManager::emit` caller) now returns `"Not connected"` during the handshake window instead of a false success. Callers already treat this as an error. No wire/schema change.

## Related

- Closes: #6084
- Follow-up PR(s)/TODOs: optionally gate `emit_with_ack` on `Connected` for a faster failure (documented as deliberate today) — not required for this fix.

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/socket-emit-connected-guard-6084

## Summary by CodeRabbit

- **Bug Fixes**
  - Messages sent while a socket is connecting or reconnecting now return a clear “Not connected” error instead of being queued prematurely.
  - Messages remain deliverable once the connection handshake completes, including after connection status errors.
  - Message sending is prevented during connection teardown, avoiding delivery through an invalid or newly reconnected channel.
  - Improved synchronization prevents messages from being stranded or delivered on the wrong connection during rapid reconnects.
- **Tests**
  - Expanded coverage for connection readiness, teardown, reconnecting, handshake completion, and acknowledgment-based messaging.

### DIFF
--- a/src/openhuman/platform/socket/manager.rs
+++ b/src/openhuman/platform/socket/manager.rs
@@ -11,7 +11,7 @@
 
 use std::collections::HashMap;
 use std::sync::{
-    atomic::{AtomicU64, Ordering},
+    atomic::{AtomicBool, AtomicU64, Ordering},
     Arc, OnceLock,
 };
 
@@ -62,6 +62,25 @@ pub(super) struct SharedState {
     /// (e.g. "backend redirected ws→wss; update BACKEND_URL"). Cleared on every
     /// successful handshake and on disconnect.
     pub(super) error: RwLock<Option<String>>,
+}
+
+/// The outbound emit channel bundled with the readiness flag of the **same**
+/// connection that owns it.
+///
+/// Readiness travels with the channel so `emit` can decide "is this message
+/// deliverable?" atomically with picking the channel it would send on: both are
+/// read under the single `emit_tx` lock. `spawn_loop` installs a fresh
+/// `EmitChannel` (fresh sender + fresh `ready = false` flag) for every
+/// connection, and the background loop flips *this connection's* `ready` to
+/// `true` only after the Socket.IO CONNECT ACK and back to `false` on teardown.
+/// A reconnect therefore swaps the sender and its flag together — an `emit`
+/// holding the lock can never pair a live-looking status with a stale
+/// pre-handshake channel (the reverse of the TOCTOU the status-only gate had).
+pub(super) struct EmitChannel {
+    /// Sender into the background loop's outbound queue.
+    pub(super) tx: mpsc::UnboundedSender<String>,
+    /// `true` only between this connection's CONNECT ACK and its teardown.
+    pub(super) ready: Arc<AtomicBool>,
 }
 
 pub(super) struct AckRegistry {
@@ -116,8 +135,10 @@ impl AckRegistry {
 pub struct SocketManager {
     /// Shared state accessible from both the manager and the background loop.
     pub(super) shared: Arc<SharedState>,
-    /// Channel for sending outgoing messages to the background loop.
-    emit_tx: tokio::sync::Mutex<Option<mpsc::UnboundedSender<String>>>,
+    /// Channel for sending outgoing messages to the background loop, bundled
+    /// with the readiness flag of the connection that owns it so `emit` reads
+    /// both atomically under one lock (see [`EmitChannel`]).
+    emit_tx: tokio::sync::Mutex<Option<EmitChannel>>,
     /// Channel for signaling the background loop to shut down.
     shutdown_tx: tokio::sync::Mutex<Option<watch::Sender<bool>>>,
     /// Join handle for the background connection loop.
@@ -264,14 +285,33 @@ impl SocketManager {
         let internal_tx = emit_tx.clone();
         let (shutdown_tx, shutdown_rx) = watch::channel(false);
 
-        *self.emit_tx.lock().await = Some(emit_tx);
+        // Per-connection readiness flag: starts `false` (pre-handshake) and is
+        // flipped by `ws_loop` at CONNECT ACK. Bundled with the sender so `emit`
+        // reads the flag belonging to exactly this channel, not a flag a later
+        // reconnect may have swapped underneath it.
+        let emit_ready = Arc::new(AtomicBool::new(false));
+        let loop_ready = Arc::clone(&emit_ready);
+
+        *self.emit_tx.lock().await = Some(EmitChannel {
+            tx: emit_tx,
+            ready: emit_ready,
+        });
         *self.shutdown_tx.lock().await = Some(shutdown_tx);
 
         let url = url.to_string();
         let shared = Arc::clone(&self.shared);
 
         let handle = tokio::spawn(async move {
-            ws_loop(url, provider, shared, emit_rx, shutdown_rx, internal_tx).await;
+            ws_loop(
+                url,
+                provider,
+                shared,
+                emit_rx,
+                shutdown_rx,
+                internal_tx,
+                loop_ready,
+            )
+            .await;
         });
 
         *self.loop_handle.lock().await = Some(handle);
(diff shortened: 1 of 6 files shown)


M3gA-Mind:
Reviewed against the independent verification block on #6084 (I wrote that verification, so I had the blast radius already traced). **Not approving or blocking — one substantive question, and it is about scope rather than correctness.**

## The guard is correct for the failing case and, I think, over-broad for the succeeding one

The PR gates `emit` on `status == Connected`. That is right whenever the handshake goes on to **fail**. But a message queued while `Connecting` is *not* generally lost — it is delivered if the handshake succeeds:

- `run_connection`'s outbound arm, `outgoing = emit_rx.recv()` (`ws_loop_part_01.rs:552`), lives in the **step-7 main event loop**, which is entered only after the Engine.IO OPEN (step 3) and the SIO CONNECT ACK (step 6, `:517`, where `Connected` is set). `emit_tx` is an *unbounded* channel, so a message queued at `Connecting` simply waits there and is sent as soon as that loop starts.
- `drain_pending_emits` (`:427-433`) is called at `:173`, i.e. **after `run_connection` returns** — only on a connection that has already ended.

So the pre-PR `Ok(())` was truthful on the happy path and a lie only on the failing one. The new guard rejects both, which converts a previously-working path into a hard `Err`.

The PR's own docs say as much, and I think this is just an inconsistency rather than a disagreement: the new `emit_with_ack` comment states *"a message queued while `Connecting` is flushed once the handshake completes"* — while the `emit` comment says the message is discarded *"if the handshake then fails"*, correctly conditional, yet the guard it documents is unconditional.

## Blast radius walk

Against the five production call sites named on #6084:

- `platform/socket/schemas.rs` (the `socket_emit` RPC) — tightening is exactly right here; this is the reported bug.
- `platform/socket/event_handlers.rs:137` (`webhook:response`) — **safe**. The triggering `webhook:request` arrives over the socket, so it cannot be reached before `Connected`.
- `platform/socket/medulla/mod.rs:642`, `:665` — spawned/log-and-continue, both already `if let Err`/`match`. Behaviour change is a warn line.
- `security/devices/tunnel_client.rs:251` (`tunnel:connect`) — **the one I would want a second look at.** It is the first frame of a tunnel session and is the call most likely to race the socket coming up; it now returns `Err("...emit tunnel:connect failed: Not connected")` where it previously queued and was delivered. `:281` (`tunnel:frame`) has the same shape. Both propagate the error rather than retrying, so whether this is fine depends on whether the caller re-drives `tunnel:connect`.

So: does anything retry `tunnel:connect`? If yes, this PR is straightforwardly good. If no, the window it closes for the RPC is opened for the tunnel.

## Worth saying explicitly

- **The tests are real and would fail without the fix.** `emit_while_connecting_errors_even_with_emit_channel_present` sets `emit_tx` *and* `status = Connecting`, which is precisely the state the old channel-only guard passed, and it asserts `rx.try_recv().is_err()` so it checks the queue was not written, not merely the return value. `emit_when_connected_queues_the_encoded_event` asserts the exact wire frame `42["test.event",{"k":"v"}]` rather than just `is_ok()`. Neither enshrines current behaviour.
- **Leaving `emit_with_ack` alone is right**, and the reasoning given for it is the correct one: the ack (or its timeout) is what makes its success claim honest.
- Dropping the now-obsolete `#[allow(dead_code)]` on `is_connected` is a nice touch.

## If you want the narrower fix

Gating on `matches!(status, Connected | Connecting)` would keep the delivered-after-handshake path and still reject the genuinely-hopeless `Disconnected` / `Error` cases — but it would *not* fix #6084, because the reported false `ok: true` happens precisely at `Connecting`. So the real choice is: accept that queued-during-connect now fails fast (fine if callers retry), or keep queueing and report a third outcome. **I would ship this as-is if `tunnel:connect` has a retry**, and I would not hold the PR for it either way — just flagging it before it lands rather than after.

YellowSnnowmann:
Thanks — both are real edge cases the naive `status == Connected` gate missed; fixed properly:

- **CodeRabbit (TOCTOU with reconnect)** and **Codex P1 (server `error` on a still-live socket wedged emits forever):** readiness is now bundled with the channel — `EmitChannel { tx, ready: Arc<AtomicBool> }`. `emit` checks `ready` **under the `emit_tx` lock**, so status and channel can't be swapped mid-emit (closes the reconnect race where a message could land on a fresh pre-handshake channel and return `Ok`). `ready` is set true only after the SIO CONNECT ACK and cleared on that connection's teardown (before `drain_pending_emits`) — it is **not** tied to the presentation-layer status, so a server `error` event that leaves the transport live no longer blocks outbound emits. `spawn_loop`/`ws_loop` install and own the flag per connection.

`emit_with_ack` and the #4355 drop-on-disconnect are unchanged. Added 3 deterministic regression tests (emit-while-connecting rejects; emit does not land on a reconnect's pre-handshake channel; emit still succeeds after a server `error` on a live connection). `cargo fmt` + lib `clippy -D warnings` + the `platform::socket` suite pass locally.

YellowSnnowmann:
Good catch — verified reachable and fixed. The emit channel + readiness flag are created once per `connect()` and **reused by every reconnect** within one `ws_loop` (`drain_pending_emits`' own doc notes the channel "outlives the socket … re-used by every reconnect attempt"). So a bare `emit` that observed `ready == true` could `tx.send` after teardown cleared `ready` and drained `emit_rx`, and the next session's receiver would forward that stale event on a fresh sid.

Fix: readiness is now `Arc<parking_lot::Mutex<bool>>` (`EmitReady`). `emit` holds that guard **across** the `ready` check and the `tx.send`; teardown clears `ready` and drains `emit_rx` **under the same guard** — the two critical sections are now mutually exclusive, so a send can't slip past a drain. Lock ordering is `emit_tx` (tokio) → `ready` (parking_lot, leaf), never nested the other way, and `ready` is never held across an `.await` (only a synchronous flag read/write + non-blocking `send`/`try_recv`). The #6084 pre-handshake gate, the #4355 drop-on-disconnect, and `emit_with_ack` are unchanged. Added a barrier-controlled interleaving test (`emit_cannot_send_past_a_concurrent_teardown_drain`) that pauses `emit` after it observes readiness, runs teardown under the held guard, and asserts the emit is rejected and nothing is queued — it fails on the plain-atomic version (no lock to block on).

`cargo fmt` + lib `clippy -D warnings` + the `platform::socket` suite (150 tests) pass locally.

PR #6104, opened 2026-09-08
fix(learning): validate facet class against the taxonomy across the learning_* family

### LINKED ISSUE #6077: `learning.list_facets(class=…)` matches on a raw key prefix, so one class can leak another's rows
## Summary

`learning.list_facets(class=…)` matches on a raw key prefix, so one class can leak another's rows

- **Namespace / method:** `openhuman.learning_list_facets`
- **Where:** `src/openhuman/agent/learning/schemas_part_02.rs:27-33`

## Problem

**Expected**

`class` selects facets of that class.

**Actual**

the filter is
```rust
f.class.as_deref() == Some(cls.as_str()) || f.key.starts_with(&format!("{cls}/"))
```
The second arm is an **unvalidated, caller-supplied prefix match on the key**,
and `cls` is never checked against the six-variant `FacetClass` set
(`style` / `identity` / `tooling` / `veto` / `goal` / `channel`). Two
consequences:
1. A row whose `class` column is one class but whose key was written with a
   different prefix is returned under the *wrong* class — the two arms can
   disagree and the `||` takes either.
2. Any string is accepted as a class. `class = "e"` is not rejected; it simply
   matches nothing, so a caller typo reads as "you have no facets of that
   class" rather than as an error. Every sibling controller in this family
   validates its inputs by name (`missing required \`class\``), so this is
   inconsistent as well as wrong.

**Steps to reproduce**

`openhuman.learning_list_facets` with `{"class": "nonsense"}`
returns `{"facets": [], "count": 0}` with no error.

### Independent verification (CONFIRMED-DIFFERENT-CAUSE — **the headline cross-class leak is not reachable.**)

**Cited location still accurate:** yes — the filter is
`src/openhuman/agent/learning/schemas_part_02.rs:28-34`, quoted verbatim in the report
(report says :27-33, one line off).

**Repro run:** the report's own repro over the real RPC surface —
`cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- --nocapture zzz_w2_verify_probe`
(scratch module, deleted after the run) — plus a source trace of every writer of the `class`
column, which is what settles claim 1.

**Observed**

```
class=nonsense: {"result":{"logs":["learning.list_facets: returned 0 facets (class_filter=Some(\"nonsense\"))"],
                           "result":{"count":0,"facets":[]}}}
class=style   : {"result":{… "count":0,"facets":[]}}
no class      : {"result":{… "count":0,"facets":[]}}
```
Claim 2 reproduces exactly: an arbitrary string is accepted, produces `count: 0`, and the log
line records `class_filter=Some("nonsense")` without complaint.

**Diagnosis assessment:** **claim 1 is refuted; claim 2's cause is right.**

Claim 1 needs a row whose `class` column disagrees with its key prefix. The column's documented
contract on both definitions is that it is *derived from the key prefix* —
`tinymemory-core/src/store/namespace_store/profile.rs:243-248` and
`tinymemory-bus/src/provider/profile.rs:176-179` (*"Facet class derived from the key prefix"*).
Tracing the writers:
- `upsert_profile_facet` (insert path, :325-327) sets `infer_class_from_key(key, facet_type)`.
- the rebuild upsert (:379-383) uses `facet.class.clone().or_else(|| infer_class_from_key(…))`,
  so an explicitly-set class *would* win — but the only non-test writer that sets one is the
  stability detector, `stability_detector.rs:305`: `class: Some(class_prefix(*class).to_string())`
  — and the key in that same struct literal is `full_key`, built ten lines earlier at :204 as
  `format!("{}/{}", class_prefix(cand.class), cand.key)` **from the same `cand.class`**. Column
  and prefix come from one value in one loop iteration; they cannot disagree.
- `infer_class_from_key` (:632-656) never returns `None`: known prefix → that prefix, `skill:*`
  → `tooling`, else a `facet_type` fallback. Legacy rows therefore get a class whose key does
  *not* start with `<class>/`, so arm 2 never fires for them either.

So the two arms are equivalent for learning-path rows and arm 2 is dead for legacy rows. The
`||` is redundant, not a leak. What survives is the smaller claim: `cls` is never validated
against `FacetClass`, so a typo returns an empty page instead of an error.

**Is the expected behaviour correct:** for claim 1, no — there is nothing to fix, and removing
the `starts_with` arm would be a tidy-up, not a security fix. For claim 2 it is arguable: the
report notes every sibling "validates its inputs by name", but that is *presence* validation
(`missing required \`class\``) — no sibling validates the class *value* either. `full_key`
(`schemas_part_01.rs:627`) is a bare `format!("{class_str}/{key_suffix}")` with no taxonomy
check, so `get_facet{class:"nonsense"}` is equally lenient. Making `list_facets` strict alone
would make it the odd one out in the other direction.

## Solution

needs a product decision: **should the whole `learning_*`
family reject a class outside the six-variant taxonomy, or stay lenient?** If strict, parse via
`cache::class_from_key`'s taxonomy in `full_key` and in `list_facets` together, not in one
controller. Dropping the redundant `starts_with` arm is safe either way but buys nothing
observable.

**Blast radius:** `handle_list_facets` is the only reader of this filter. `cache::list_by_class`
(`cache.rs:90-97`) already exists and filters on the key prefix alone over `list_active()` —
worth noting it would return a *different* set (active-only, no Provisional), so it is not a
drop-in. Changing `full_key` touches all six `learning_*` controllers plus `tools.rs:41`'s
duplicate.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- `learning.list_facets` (and the whole `learning_*` family) now validates the caller-supplied `class` against the six-variant `FacetClass` taxonomy (`style` / `identity` / `tooling` / `veto` / `goal` / `channel`).
- An unknown class (e.g. `"nonsense"`, or a value containing `/`) is rejected with a structured error instead of silently returning an empty page.
- The redundant `|| key.starts_with("{class}/")` filter arm in `list_facets` is dropped; the filter now matches the `class` column alone.

## Problem

`list_facets` filtered with `f.class.as_deref() == Some(cls) || f.key.starts_with(&format!("{cls}/"))` and never validated `cls`. Two consequences: the family validated class **presence** (`missing required \`class\``) but never its **value**, so a typo read as "you have no facets of that class" rather than an error — inconsistent and silent. Independent verification established the class column is always derived from the key prefix, so the headline cross-class "leak" was **not reachable** and the `||` arm was redundant; what remained was the validation-consistency gap. Per maintainer decision, the family is made strict.

## Solution

- Add `parse_facet_class_name(class)` (delegating to the existing `class_from_key` taxonomy, plus an explicit reject for any name containing `/` so `"style/"` can't slip through and compose an unmatchable key).
- Validate `class` in every `learning_*` handler that takes one — `list_facets`, `get_facet`, `update_facet`, `pin_facet`, `unpin_facet`, `forget_facet` — and in their agent-tool mirrors, **after** the presence check, before `full_key`. Uniform across the family (the point of the decision).
- Drop the redundant `starts_with` arm; `list_facets` filters on the class column only. `full_key`'s signature is unchanged (validation is per-handler, minimal blast radius). `stability_detector`'s enum-path `full_key` is untouched.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — unknown-class and slash rejection for `list_facets`/`get_facet`/`forget_facet` + tool mirrors; a mislabelled-column row proves the `starts_with` removal; valid-class paths still succeed.
- [x] **Diff coverage ≥ 80%** — new validator + per-handler guards are exercised by the added tests. (Full `diff-cover` runs in CI.)

## Impact

- `learning_*` RPC + agent tools now reject an out-of-taxonomy class with an error instead of an empty result. A caller passing a valid class is unaffected. No wire/schema shape change.

## Related

- Closes: #6077

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/learning-class-validation-6077

## Summary by CodeRabbit

* **Bug Fixes**
  * Learning facet operations now reject unknown or malformed class names with clear validation errors.
  * Facet listings filter strictly by the selected class, preventing unrelated results from appearing.
  * Missing class values continue to return the appropriate required-field error.
  * Classless facets remain discoverable through their canonical key prefixes when applicable.

* **Tests**
  * Added coverage for valid and invalid class names across learning facet operations.
  * Added tests confirming filtering behavior and preventing store access for invalid requests.

### DIFF
--- a/src/openhuman/agent/learning/cache.rs
+++ b/src/openhuman/agent/learning/cache.rs
@@ -147,6 +147,21 @@ pub fn class_from_key(key: &str) -> Option<FacetClass> {
     }
 }
 
+/// Validate a caller-supplied class **name** against the facet taxonomy.
+///
+/// The learning RPC handlers and agent tools take `class` as a bare taxonomy
+/// name (`"style"`, `"goal"`, …), not a full key. Reuse [`class_from_key`],
+/// which matches on the first `/`-segment, so a bare name is validated as its
+/// own prefix and an unknown name is rejected with a caller-facing error rather
+/// than silently composing a key no facet can ever carry.
+pub(crate) fn parse_facet_class_name(class: &str) -> Result<FacetClass, String> {
+    class_from_key(class).ok_or_else(|| {
+        format!(
+            "invalid class `{class}` (expected one of style, identity, tooling, veto, goal, channel)"
+        )
+    })
+}
+
 /// Delete every non-`Pinned` facet, returning `(deleted, pinned_preserved)`.
 ///
 /// Shared by the `learning.reset_cache` RPC and the `learning_reset_cache`
--- a/src/openhuman/agent/learning/cache.rs
+++ b/src/openhuman/agent/learning/cache.rs
@@ -155,11 +155,19 @@ pub fn class_from_key(key: &str) -> Option<FacetClass> {
 /// own prefix and an unknown name is rejected with a caller-facing error rather
 /// than silently composing a key no facet can ever carry.
 pub(crate) fn parse_facet_class_name(class: &str) -> Result<FacetClass, String> {
-    class_from_key(class).ok_or_else(|| {
+    // A taxonomy name never contains a `/`. Reject one up front so a value like
+    // `"style/"` (which `class_from_key` would accept via its first segment,
+    // then compose into a key no facet can carry) fails loudly instead of
+    // reintroducing the silent-empty behaviour this fix removes.
+    let invalid = || {
         format!(
             "invalid class `{class}` (expected one of style, identity, tooling, veto, goal, channel)"
         )
-    })
+    };
+    if class.contains('/') {
+        return Err(invalid());
+    }
+    class_from_key(class).ok_or_else(invalid)
 }
 
 /// Delete every non-`Pinned` facet, returning `(deleted, pinned_preserved)`.
--- a/src/openhuman/agent/learning/cache_tests.rs
+++ b/src/openhuman/agent/learning/cache_tests.rs
@@ -104,6 +104,16 @@ fn parse_facet_class_name_rejects_unknown_class() {
     );
 }
 
+#[test]
+fn parse_facet_class_name_rejects_a_name_with_a_slash() {
+    // `"style/"` shares a first `/`-segment with a real class, so a bare
+    // `class_from_key` would accept it and then compose an unmatchable key —
+    // the silent-empty behaviour #6077 removes. The slash guard rejects it.
+    let err =
+        parse_facet_class_name("style/").expect_err("a slash in a class name must be rejected");
+    assert!(err.contains("invalid class `style/`"), "got: {err}");
+}
+
 // ── class filter is column-only (#6077) ───────────────────────────────────────
 //
 // `list_facets` filters on the `class` column alone; the redundant
--- a/src/openhuman/agent/learning/cache_tests.rs
+++ b/src/openhuman/agent/learning/cache_tests.rs
@@ -81,6 +81,71 @@ fn class_from_key_parses_known_classes() {
     assert_eq!(class_from_key("no_slash"), None);
 }
 
+// ── parse_facet_class_name ────────────────────────────────────────────────────
+
+#[test]
+fn parse_facet_class_name_accepts_every_taxonomy_name() {
+    assert_eq!(parse_facet_class_name("style"), Ok(FacetClass::Style));
+    assert_eq!(parse_facet_class_name("identity"), Ok(FacetClass::Identity));
+    assert_eq!(parse_facet_class_name("tooling"), Ok(FacetClass::Tooling));
+    assert_eq!(parse_facet_class_name("veto"), Ok(FacetClass::Veto));
+    assert_eq!(parse_facet_class_name("goal"), Ok(FacetClass::Goal));
+    assert_eq!(parse_facet_class_name("channel"), Ok(FacetClass::Channel));
+}
+
+#[test]
+fn parse_facet_class_name_rejects_unknown_class() {
+    let err = parse_facet_class_name("nonsense").expect_err("unknown class must be rejected");
+    assert!(err.contains("invalid class `nonsense`"), "got: {err}");
+    // Lists the accepted taxonomy so the caller can recover.
+    assert!(
+        err.contains("style, identity, tooling, veto, goal, channel"),
+        "got: {err}"
+    );
+}
+
+// ── class filter is column-only (#6077) ───────────────────────────────────────
+//
+// `list_facets` filters on the `class` column alone; the redundant
+// `|| key.starts_with("{class}/")` arm was dropped. This proves the two are not
+// equivalent when they disagree: a valid class returns exactly the rows whose
+// `class` column matches, and a key-prefix that does not match the column is
+// excluded.
+#[tokio::test]
+async fn class_filter_matches_column_not_key_prefix() {
+    let cache = make_cache();
+
+    // Two facets whose class column is genuinely `style`.
+    let mut a = stub_facet("f1", "style/verbosity", "terse", FacetState::Active, 1.8);
+    a.class = Some("style".into());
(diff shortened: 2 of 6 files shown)


M3gA-Mind:
Reviewed against the independent verification block on #6077 (I wrote that verification, so this is a check of the diff against the traced blast radius, not a fresh reading of the bug).

**This matches the verified diagnosis rather than the original report, which is the right call.** The report's headline was a cross-class *leak*; verification refuted that — the `class` column is always derived from the key prefix, so the two arms of the old filter could not disagree. The PR description says exactly this and fixes what actually survived (unvalidated class value), instead of chasing the refuted claim. Dropping the `|| key.starts_with(…)` arm is safe for the same reason: a row can only match the prefix arm if its column already matches, so no row is lost.

**Blast radius walked against the verification, and it is covered:**
- `handle_list_facets` — the only reader of the filter. ✅
- The duplicate surface the verification flagged (`tools.rs`'s own `full_key`/filter) is fixed too, not just the RPC. I counted five class-taking tool surfaces (`ListFacets`' optional filter plus the four `read_required_str(&args, "class")` sites) and five `validate_class` calls. ✅
- Six RPC handlers (`list`, `get`, `update`, `pin`, `unpin`, `forget`) each validate. ✅
- Both surfaces delegate to one validator, so they cannot drift apart again — which was the failure mode in the sibling issue.

**The `/`-rejection is a genuine catch.** `class_from_key` matches the first `/`-segment, so `"style/"` would have passed and then composed `"style//foo"` — an unmatchable key, i.e. exactly the silent-empty behaviour being removed. Good that it is rejected explicitly rather than left to the caller.

**Tests exercise the changed path.** `list_facets_rejects_unknown_class` and friends call the real handlers and assert on `invalid class \`nonsense\``. On the old code the same call reaches `get_cache()` and fails with a store error instead, so the assertion still fails — for the right reason, because it matches the message rather than merely asserting `is_err()`. The comment explaining that validation lands before the store bind (which is what lets these run without a memory guard) is accurate.

---

### Two non-blocking notes

1. **`full_key` remains unvalidated; the guard is at 11 call sites.** The verification's blast radius named `full_key` (`schemas_part_01.rs`) plus its `tools.rs` duplicate as the central seam. Validating at each caller works today, but a seventh handler or sixth tool added later composes an unvalidated key and silently reacquires the bug. Not worth reworking now — `full_key` returns `String`, so making it fallible is a wider change — but a one-line note at `full_key` pointing at `parse_facet_class_name` would make the requirement discoverable at the place someone will copy from.

2. **`cache::list_by_class` (`cache.rs:90-97`) still filters on the key prefix**, which is now the one path not using the class column this PR standardises on. It is called only from `stability_detector_tests.rs:221` — no production caller — so this is not a defect and needs no change here. Flagging only because it is the last instance of the pattern the PR removes elsewhere, and it also reads `list_active()` (so it excludes `Provisional`, unlike the handlers) — worth a comment or a deletion at some point so nobody mistakes it for the canonical filter.

### Heads-up on a file collision

I am about to raise a small PR for #6108 (`forget_facet`'s log line is built before the read that decides whether anything was dropped) that touches `schemas_part_02.rs` and `schemas_tests.rs` — the same two files. The changes are complementary and in different parts of `handle_forget_facet`: yours adds the class check at the top, mine moves the log construction into the `Some` arm lower down. Whichever merges second will want a trivial rebase; happy to rebase mine onto yours rather than the other way round.

YellowSnnowmann:
Thanks — valid catch, fixed at root cause. `ProfileFacet.class` is `Option`, and a facet can legitimately carry a canonical key (`style/verbosity`) with `class: None` (other paths derive the class from the key when the column is absent). Dropping the key-prefix arm entirely hid such classless rows.

`list_facets` (both the schema handler and the tool mirror) now keeps the class column authoritative and falls back to the key prefix **only when the column is `None`**:
`class == Some(cls) || (class.is_none() && key.starts_with("cls/"))`.
So legitimate classless rows are returned again, while the #6077 leak stays closed — a row with `class == Some(other)` still cannot match `cls` via key prefix (pinned by the existing mislabelled-column exclusion test). Added a classless-row regression test; the `parse_facet_class_name` taxonomy validation is unchanged. Learning lib suite green locally.

PR #6105, opened 2026-09-08
fix(session_db): remove the six read controllers with no writer

### LINKED ISSUE #6082: `session_db.*` reads a table nothing in the product ever writes
## Summary

`session_db.*` reads a table nothing in the product ever writes

- **Namespace / method:** `openhuman.session_db_list`, `_get`, `_search`,
`_get_messages`, `_get_tool_calls`, `_get_children` (the whole namespace)
- **Where:** `src/openhuman/agent/session_db/schemas.rs` (all six handlers);
writer side `vendor/tinyagents/crates/tinyagents-session/src/ops.rs:21`
(`record_session_start`), `:127` (`record_message`), `:211` (`record_tool_call`)

## Problem

**Expected**

the durable session index the namespace advertises —
"queryable index over transcripts, lineage, tool calls" (`src/core/all.rs:967`) —
is populated as agent sessions run, so `session_db.list` answers with the
sessions the user has had.

**Actual**

**no caller in `src/` ever records a session.** The three recording
entry points are referenced only by `tinyagents-session`'s own tests and by
`vendor/tinyagents/crates/tinyagents-integration-tests/`. OpenHuman's only use
of the crate is the sibling `run_ledger` module (`web_chat/progress_bridge.rs`,
`agent/orchestration/*`, `agent/progress_tracing/*`) — a different set of
tables. So all six read controllers are permanently empty in production: the
schema is created on first read and never gains a row.

**Steps to reproduce**

```
# nothing in src/ writes the sessions table:
grep -rn 'record_session_start\|record_message(\|record_tool_call' src/     # → no matches
grep -rn 'tinyagents_session::' src/ | grep -v run_ledger | grep -v session_db/schemas.rs   # → no matches
```
and at runtime, `tests/raw_coverage/session_store_e2e.rs::session_db_reads_answer_a_wellformed_empty_store`
shows a fresh workspace answering `{sessions: [], total: 0}` — which is also
what a long-running install answers.

### Independent verification (CONFIRMED — the core finding. **Two supporting claims are refuted**, one of which)

**Cited location still accurate:** yes. `src/openhuman/agent/session_db/schemas.rs` registers
**two** namespaces — six `session_db_*` handlers reading `tinyagents_session::{list_sessions,
get_session, search_sessions, list_messages, list_tool_calls, list_children}` (:233-:382) and
three `run_ledger_*` handlers reading `run_ledger::*` (:394-:456). The report scopes itself to
the six correctly. Writer side: `record_session_start` :21, `record_message` :127,
`record_tool_call` :211 of `tinyagents-session/src/ops.rs`. `src/core/all.rs:966` carries the
"queryable index over transcripts, lineage, tool calls" description.

**Repro run:** the report's greps, then an executable end-to-end probe (scratch module, deleted
after the run) — `cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- --nocapture zzz_w2_verify_probe`

**Observed**

```
session_db_list (empty store): {"sessions":[],"total":0}
workspace_dir = /var/…/.openhuman/users/local/workspace
record_session_start from a tests/ target -> Ok("w2-probe-session-1")
session_db_list (after seeding): {"sessions":[{"id":"w2-probe-session-1","agent_definition_id":"researcher",
    "session_key":"w2-probe-key","source_channel":"web","thread_id":"thread-w2","model":"probe-model",
    "status":"running","turn_count":0,…}],"total":1}
session_db_get: {…same row…}
```
Grep side: the report's `record_tool_call` pattern also matches
`ContextManager::record_tool_calls` (`src/openhuman/agent/context/{manager,stats,session_memory}.rs`)
and `LimitTracker::record_tool_call` / metrics in `tinyagents-harness` — all **name collisions**,
none of them the session store. Filtering those out, the report's conclusion holds exactly:
the only callers of the three session recorders are `tinyagents-session`'s own `test.rs` and
`tinyagents-integration-tests/tests/{e2e_session_lifecycle,feature_session_retention}.rs`.
OpenHuman's only use of the crate in `src/` is `run_ledger::*`
(`web_chat/progress_bridge.rs:132-166`, `agent/progress_tracing.rs:560`) — a different table set.

**Diagnosis assessment:** **the cause is right.** No `src/` path records a session, so the six
`session_db_*` controllers read a table that is created on first read and never gains a row.
The probe also shows the read side is *correct* — seed one row and all six answer properly — so
this is purely a missing writer, not a broken reader.

**Is the expected behaviour correct:** yes. `all.rs:966` advertises a durable index; a surface
that is structurally always empty does not deliver that.

## Solution

call `record_session_start` / `record_message` /
`record_tool_call` from wherever agent turns are already instrumented — `run_ledger` is written
from `web_chat/progress_bridge.rs`, which is the natural seam and already holds the workspace
dir. **Or** delete the six controllers if the run-ledger index has superseded them; the two
options are opposite and the choice is a product call, not an engineering one.

**Blast radius:** six controllers plus their `/schema` entries and any frontend caller; the
three `run_ledger_*` controllers in the same file are unaffected (that table *is* written).
`session_db_search` is the only full-text route over transcripts. Nothing pins the current
empty-store behaviour beyond the wave's own new suite.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- Removes the six `session_db_*` read controllers (`list`, `get`, `search`, `get_messages`, `get_tool_calls`, `get_children`) — they queried a sessions table that nothing in `src/` ever writes, so all six were permanently empty in production with no frontend consumer.
- The three `run_ledger_*` controllers in the same module (which read a table that **is** written) are left fully intact and registered.

## Problem

The `session_db_*` namespace advertised a "durable index over transcripts, lineage, tool calls" (`src/core/all.rs`), but the three recording entry points (`record_session_start` / `record_message` / `record_tool_call`) are referenced only by `tinyagents-session`'s own tests — never from `src/`. OpenHuman's only use of the crate in `src/` is the sibling `run_ledger`, a different table. So the six read controllers answered `{sessions: [], total: 0}` for a fresh workspace and for a long-running install alike, and `run_ledger` (which does store run metadata) is not a superset — it holds no transcripts. Per maintainer decision (no writer, no UI consumer), the dead read surface is removed rather than wired up.

## Solution

- Delete the six handlers, their `ControllerSchema`s, the now-unused `tinyagents_session::{list_sessions, …}` / `SessionSearchParams` imports, and their tests from `session_db/schemas.rs`; the shared aggregator now returns only the three `run_ledger_*` entries.
- Update the `src/core/all.rs` registration comment so it no longer advertises the removed surface; the `DomainGroup::Agent` push and `run_ledger` registration/description are unchanged.
- The on-disk `session_db/sessions.db` directory path (used by `run_ledger` and `session_import`) is deliberately untouched.
- Regression test asserts the six `openhuman.session_db_*` methods are absent from the registered set / schema while the three `openhuman.run_ledger_*` remain (mirrors the existing removal-test pattern in `all_tests.rs`).

## Submission Checklist

- [x] Tests added or updated — absence of the six + survival of the three (registry-level and module-level).
- [x] **Diff coverage ≥ 80%** — this is primarily a deletion; the added assertions cover the surviving registration. (Full `diff-cover` runs in CI.)
- [x] Coverage matrix updated — `N/A: the only session_db mention in the matrix is the on-disk DB path in test row 16.1.1, not these controllers`.
- [x] Manual smoke checklist updated — `N/A: removed surface has no UI/release-cut consumer`.

## Impact

- The `session_db_*` methods become unknown-method (they returned empty in production regardless). `run_ledger_*` — including the frontend `run_ledger_list`/`get`/`events` calls — is unaffected. Reversible if queryable session transcripts are wanted later (would require wiring the writers first).

## Related

- Closes: #6082

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/session-db-remove-dead-controllers-6082

## Summary by CodeRabbit

- **Changed**
  - Replaced the read-only session database interface with the run ledger interface.
  - The following session database operations are no longer available: listing, retrieving, searching, viewing messages, viewing tool calls, and viewing child sessions.
  - The run ledger continues to support listing runs, retrieving run details, and viewing run events.
  - Updated interface metadata and validation to reflect the available operations.

### DIFF
--- a/src/core/all.rs
+++ b/src/core/all.rs
@@ -963,7 +963,7 @@ fn build_registered_controllers() -> Vec<GroupedController> {
         DomainGroup::Security,
         crate::openhuman::security::devices::all_devices_registered_controllers(),
     );
-    // Durable agent session database — queryable index over transcripts, lineage, tool calls
+    // Durable agent/workflow run ledger — read surface over run state, lineage, events, telemetry
     push(
         &mut controllers,
         DomainGroup::Agent,
--- a/src/core/all_tests.rs
+++ b/src/core/all_tests.rs
@@ -2481,3 +2481,59 @@ fn memory_diff_controllers_are_gone_and_memory_survives() {
         "removing the git ledger must not remove the memory domain"
     );
 }
+
+// ---- session_db removal (#6082) --------------------------------------------
+
+/// The six read-only `session_db` controllers were removed in #6082: they
+/// queried a session index that nothing in `src/` ever writes (permanently
+/// empty in production, no frontend consumer). The three `run_ledger`
+/// controllers live in the same module and read a table that *is* written
+/// (from `web_chat::progress_bridge` and `agent::progress_tracing`), so they
+/// must stay fully registered.
+///
+/// `run_ledger` is asserted present in the same test on purpose: the removal
+/// took the dead read surface, not the run-ledger domain. Splitting that into a
+/// separate test would let one pass while the other silently regressed.
+#[test]
+fn session_db_controllers_are_gone_and_run_ledger_survives() {
+    let methods: Vec<String> = all_controller_schemas()
+        .iter()
+        .map(rpc_method_name)
+        .collect();
+
+    for removed in [
+        "openhuman.session_db_list",
+        "openhuman.session_db_get",
+        "openhuman.session_db_search",
+        "openhuman.session_db_get_messages",
+        "openhuman.session_db_get_tool_calls",
+        "openhuman.session_db_get_children",
+    ] {
+        assert!(
+            !methods.contains(&removed.to_string()),
+            "removed session_db controller `{removed}` must be absent \
+             (unknown-method over /rpc, omitted from /schema), got: {methods:?}"
+        );
+    }
+
+    for kept in [
+        "openhuman.run_ledger_list",
+        "openhuman.run_ledger_get",
+        "openhuman.run_ledger_events",
+    ] {
+        assert!(
+            methods.contains(&kept.to_string()),
+            "run_ledger controller `{kept}` must stay registered — removing the \
+             dead session_db read surface must not touch the run ledger"
+        );
+    }
+
+    let namespaces: Vec<&str> = all_controller_schemas()
+        .iter()
+        .map(|s| s.namespace)
+        .collect();
+    assert!(
+        !namespaces.contains(&"session_db"),
+        "the `session_db` namespace was removed and must not be registered, got: {namespaces:?}"
+    );
+}
--- a/src/openhuman/agent/session_db/mod.rs
+++ b/src/openhuman/agent/session_db/mod.rs
@@ -1,21 +1,25 @@
-//! JSON-RPC surface for the durable agent session database.
+//! JSON-RPC surface for the durable agent run ledger.
 //!
-//! The store itself — sessions, messages, tool calls, cost metadata,
-//! parent/child lineage, and the run ledger — lives in
-//! [`tinyagents::session`]. Only the controller schemas and
+//! The store itself — run ledger rows, child lineage, events, and telemetry —
+//! lives in [`tinyagents::session::run_ledger`]. Only the controller schemas and
 //! their handlers stay here, because the RPC envelope, config resolution, and
 //! `RpcOutcome` shape are host concerns the runtime crate has no business
 //! knowing about.
 //!
-//! Call the store directly (`tinyagents_session::…`) rather
+//! Call the store directly (`tinyagents_session::run_ledger::…`) rather
 //! than through this module; it deliberately re-exports no storage API.
 //!
 //! Every store entry point takes the workspace root, so handlers pass
 //! `config.workspace_dir`. The database path is
-//! `{workspace}/session_db/sessions.db` — unchanged by the move, so existing
-//! installs keep their history.
+//! `{workspace}/session_db/sessions.db` — unchanged, so existing installs keep
+//! their run-ledger history.
 //!
-//! The `session_db` and `run_ledger` RPC namespaces are unchanged.
+//! The `run_ledger` RPC namespace is unchanged. The six read-only `session_db`
+//! controllers were removed in #6082: they queried a session index that nothing
+//! in `src/` ever writes (permanently empty in production, no frontend
+//! consumer). The run ledger below is written from
+//! [`crate::openhuman::web_chat::progress_bridge`] and
+//! [`crate::openhuman::agent::progress_tracing`], so it stays fully functional.
 
 mod schemas;
 
--- a/src/openhuman/agent/session_db/schemas.rs
+++ b/src/openhuman/agent/session_db/schemas.rs
@@ -1,4 +1,4 @@
-//! Controller schemas and JSON-RPC dispatchers for the session database.
+//! Controller schemas and JSON-RPC dispatchers for the durable run ledger.
 
 use serde_json::{Map, Value};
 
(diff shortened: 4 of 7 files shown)


M3gA-Mind:
Reviewed against the independent verification block on #6082 (I wrote that verification, so this is a check of the diff against the traced blast radius rather than a fresh reading of the bug).

**The removal itself checks out.** I re-walked the two facts the decision rests on:
- No `src/` caller of `record_session_start` / `record_message` / `record_tool_call` — the only hits outside `tinyagents-session`'s own tests are name collisions (`ContextManager::record_tool_calls`, `LimitTracker::record_tool_call`), none of them the session store.
- No frontend consumer: `grep -rn 'session_db_' app/src/` is empty.

The three `run_ledger_*` controllers are correctly left registered, and retargeting the `all.rs:966` comment to describe the run ledger rather than the deleted session index is the right touch.

---

### Blocking: this breaks two tests in a file the PR does not touch

`tests/raw_coverage/session_store_e2e.rs` landed in `main` in `a14a1da53` and **is present in this PR's base (`93d22ce6f`)**, but the diff changes no file under `tests/`. That suite dispatches all six removed methods over the real RPC surface, so after this change they answer `unknown method` and both cases fail:

- `session_store_e2e.rs:292` `session_db_reads_answer_a_wellformed_empty_store` — dispatches `session_db_list` (:296), `_search` (:329), `_get_messages` (:344), `_get_tool_calls` (:345), `_get_children` (:364) and asserts on each payload.
- `session_store_e2e.rs:386` `session_db_rejects_missing_params_and_unknown_ids` — asserts `session_db_get` without `id` yields `missing required param 'id'` (:390-392) and that an unknown id errors (:413-417). Both become `unknown method` instead.

Worth deciding deliberately rather than just deleting them: `session_db_rejects_missing_params_and_unknown_ids` is currently the only coverage of the param-validation precedence that #6073/#6095 just documented, so if it goes, that precedence loses a witness on this namespace.

The suite also has a `catalog_has` helper (:248) used elsewhere in the file to assert a method is *absent* from `/schema` (`test_support_controllers_are_absent_without_their_feature`, :1123). That is the natural replacement shape here — assert the six are gone from the catalog rather than dropping the cases outright, which keeps a regression guard against them being re-registered without a writer.

### Non-blocking

`cache`-side note only, no action needed: I checked `session_db_search` (called out in the verification as "the only full-text route over transcripts") for other consumers and found none, so nothing else regresses on its removal.

### One thing I could not check

I did not revert-check the added `all_tests.rs` / `schemas_tests.rs` cases — I can read that they assert the six are unregistered and the three `run_ledger_*` remain, which does exercise the changed path, but I have not run them on your branch.

YellowSnnowmann:
Thanks — valid, and important: those two `raw_coverage_all` integration tests still referenced the removed controllers and would have failed CI Full. Fixed by keeping coverage, not just deleting:

- **`session_store_e2e.rs`:** removed the two `session_db_*` test fns (`session_db_reads_answer_a_wellformed_empty_store`, `session_db_rejects_missing_params_and_unknown_ids`) and dropped `session_db` from the module-doc header. The `session_import` / `tokenjuice` / `ai` coverage in that file is untouched.
- **`dispatch_param_validation_e2e.rs`:** repointed the dispatcher param-validation example off `session_db_get` onto `run_ledger_get` — registered unconditionally, same required `id` param and identical refusal message (`missing required param 'id': Run id.` from the gate; `missing required param: id` from the handler) — so the test's intent (explicit-null passes the presence gate, handler refuses) is preserved verbatim.

`grep -rn session_db_ tests/ src/` shows no remaining expectation that the removed controllers are registered/working (only the retained module dir name and the absence assertions). `cargo check --test raw_coverage_all` + the repointed dispatch test pass locally.

YellowSnnowmann:
**CI note — the `Rust Core Coverage` red here is pre-existing upstream breakage, not this change.**

The coverage lane failed on two `flows_lifecycle_e2e` tests:
- `flows_approval_manifest_classifies_every_gated_node_kind` (`tests/raw_coverage/flows_lifecycle_e2e.rs:943` — "un-gated approvable keys land in already_trusted: []")
- `flows_tool_catalog_surface_degrades_without_composio_credentials` (`:1100` — composio slug shape guard)

This PR is a pure deletion of the dead `session_db_*` read controllers and touches **zero** flows code (`git diff upstream/main HEAD -- src/openhuman/flows tests/raw_coverage/flows_lifecycle_e2e.rs` is empty). Both tests reproduce as FAILED locally against the unmodified upstream flows code (8 passed / 2 failed), so they are broken on `main` independently of this change.

They only entered this PR's coverage scope because the branch carries a merge of `upstream/main` (used to resolve the earlier conflict) while the fork's `main` is a few commits behind upstream — so the changed-area detection sees the newer `flows_lifecycle_e2e.rs` as "added" and runs it. Syncing the fork (or re-running CI once `main` is current) removes it from scope; the flows failures themselves are an upstream issue to address separately. Deliberately **not** skipping/altering the flows tests here.

PR #6106, opened 2026-09-08
refactor(medulla): rename client EventEnvelope to WireEventEnvelope

### LINKED ISSUE #6078: `medulla` has two different `EventEnvelope` types; the one its module publicly re-exports is not the one its RPC returns
## Summary

`medulla` has two different `EventEnvelope` types; the one its module publicly re-exports is not the one its RPC returns

- **Namespace / method:** openhuman.medulla_list_events
- **Where:** * `src/openhuman/medulla/events/types.rs:204-213` — contract `EventEnvelope`
* `src/openhuman/medulla/client/types/event.rs:14-25` — client/wire `EventEnvelope`
* `src/openhuman/medulla/mod.rs:45` — `pub use events::{EventEnvelope, SessionEvent};`
* `src/openhuman/medulla/ops.rs:12-16` — `use super::client::{… EventEnvelope …}`

## Problem

**Expected**

the type a domain publicly re-exports under a given name is the
type its RPC surface speaks, or the two are named differently.

**Actual**

two structs, same name, different wire shapes:

| | contract (`events::`) | wire (`client::types::event::`) |
|---|---|---|
| `seq` | `u64`, required | `Option<u64>` |
| `at` | `i64` | `u64` |
| `sessionId` | **absent** | `String`, **required** |
| `cycleId` | absent | `Option<String>` |
| `event` | typed `SessionEvent` | raw `Value` |

`medulla/mod.rs` re-exports the **contract** one as the domain's public
`EventEnvelope`, while `ops::list_events` — the thing behind
`openhuman.medulla_list_events` — returns the **client** one. A consumer who
reads the module's public surface and builds a payload from it gets
`failed to decode response: missing field 'sessionId'`.

Neither type is wrong on its own; the contract type is the harness-facing
vocabulary and the client type is the backend's wire. The hazard is purely
that they share a name and the public re-export points at the one the RPC does
not use.

**Steps to reproduce**

`tests/raw_coverage/medulla_session_e2e.rs` →
`medulla_replay_is_a_cursor_and_tolerates_unmodelled_event_kinds`. Its first
run failed exactly this way — the fixture was built from the contract shape.
The suite now serves the wire shape and asserts `sessionId` round-trips, with
the distinction written out at the fixture.

### Independent verification (CONFIRMED (cosmetic, as the report itself grades it))

**Cited location still accurate:** yes, with one off-by-one. Contract type at
`src/openhuman/medulla/events/types.rs:206-213`; client type at
`src/openhuman/medulla/client/types/event.rs:15-25`; the re-export is at
`src/openhuman/medulla/mod.rs:**46**` (report said 45); `ops.rs:14` imports
`EventEnvelope` from `super::client`.

**Repro run:** `cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- W5VERIFY_`

**Observed**

```
test W5VERIFY_medulla_session_e2e::medulla_replay_is_a_cursor_and_tolerates_unmodelled_event_kinds ... ok
```
The field table in the report is accurate field-for-field:
```rust
// events::EventEnvelope            // client::types::event::EventEnvelope
pub seq: u64,                       pub seq: Option<u64>,
pub at: i64,                        pub at: u64,
//  (no sessionId)                  #[serde(rename="sessionId")] pub session_id: String,
//  (no cycleId)                    #[serde(rename="cycleId")] pub cycle_id: Option<String>,
pub event: SessionEvent,            pub event: Value,
```
And `ops::list_events` (`ops.rs:162-166`) returns `Vec<EventEnvelope>` resolved
through the `super::client` import — the wire type, not the re-exported one.

**Diagnosis assessment:** correct. Both types exist, both are reasonable in their own
layer, and `mod.rs` re-exports the one the RPC does *not* speak. I checked for a
third definition or a `#[serde(alias)]` bridge; there is none.

**Is the expected behaviour correct:** yes, and the report is appropriately modest
about it — it asserts a naming convention ("same name should mean same shape, or
name them differently"), not a defect. Rust's type system prevents any in-crate
runtime confusion, so the hazard is confined to a human reading `mod.rs` to learn the
wire shape. "Cosmetic" is the right severity; I would not raise it.

## Solution

rename the client one `WireEventEnvelope` (it is the
narrower blast radius of the two — see below), or drop `EventEnvelope` from the
`mod.rs:46` re-export so callers must name the layer they mean.

**Blast radius:** `mod.rs:46` re-exports `events::{EventEnvelope, SessionEvent}`; the
contract type is the harness-facing vocabulary, the client type is used by
`ops.rs` (`list_events`, and the other client calls sharing that `use`). A rename
touches only in-crate paths — neither type crosses a serialization boundary under
the *name*, so no wire compatibility is at stake, only source churn.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- Renames the medulla **client/wire** struct `EventEnvelope` to `WireEventEnvelope`, so it can no longer be confused with the in-process contract type of the same name.
- The contract type (`events::EventEnvelope`) and the `medulla/mod.rs` public re-export are left unchanged.

## Problem

`medulla` had two structs named `EventEnvelope`: the contract type (`events::` — `seq: u64`, `at: i64`, `event: SessionEvent`) and the wire type (`client::` — `seq: Option<u64>`, `at: u64`, required `sessionId`, `cycleId`, `event: Value`). `medulla/mod.rs` re-exported the **contract** one as the domain's public `EventEnvelope`, while `ops::list_events` (behind `openhuman.medulla_list_events`) returns the **wire** one. A consumer reading the public surface and building a payload from it hits `missing field 'sessionId'`. The hazard is purely the shared name plus the re-export pointing at the type the RPC does not speak.

## Solution

- Rename the client type to `WireEventEnvelope` at its definition and every client-side reference (`client/types`, `client/sessions.rs`, `client/sse/*`, `ops.rs`'s import + `list_events` return, `embed/medulla.rs`, `embed/mod.rs`). The grep was re-run after the edit to confirm every remaining bare `EventEnvelope` is the contract type.
- **Zero serde/JSON changes** — the wire format is governed by the unchanged `#[serde(rename = "sessionId"/"cycleId")]` attributes, not the struct name, so this is pure in-crate source churn with no wire-compatibility impact.
- The decode test is renamed to the new type and gains an assertion that a payload carrying `sessionId` (the field the contract type lacks) round-trips into `WireEventEnvelope`, documenting the distinction so it can't silently regress.

## Submission Checklist

- [x] Tests added or updated — `WireEventEnvelope` decodes a `sessionId`/`cycleId` payload (the wire shape `list_events` returns).
- [x] **Diff coverage ≥ 80%** — the renamed lines are exercised by the existing + new client decode tests. (Full `diff-cover` runs in CI.)
- [x] Coverage matrix updated — `N/A: rename/refactor, no feature row added/removed`.
- [x] Manual smoke checklist updated — `N/A: in-crate rename, no user-facing change`.

## Impact

- In-crate source only; no wire, schema, or runtime behaviour change. A consumer reading `medulla`'s public surface can no longer mistake the contract type for the RPC's wire type.

## Related

- Closes: #6078

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/medulla-event-envelope-name-6078

## Summary by CodeRabbit

- **API Changes**
  - Event retrieval and streaming APIs now return `WireEventEnvelope` records.
  - Event records consistently expose session and timestamp information, while preserving optional sequence and cycle identifiers.
  - Updated event decoding supports both present and absent optional fields.
  - The previous `EventEnvelope` name is no longer exported by these APIs.

### DIFF
--- a/src/embed/medulla.rs
+++ b/src/embed/medulla.rs
@@ -23,8 +23,8 @@ use super::error::CoreError;
 use crate::core::runtime::CoreRuntime;
 
 pub use crate::openhuman::medulla::client::{
-    AbortResult, EventEnvelope, Message, RosterWorker, SendResult, SessionCreated, SessionDetail,
-    SessionSummary,
+    AbortResult, Message, RosterWorker, SendResult, SessionCreated, SessionDetail, SessionSummary,
+    WireEventEnvelope,
 };
 pub use crate::openhuman::medulla::ops::MedullaStatus;
 
@@ -141,7 +141,7 @@ impl Medulla<'_> {
         &self,
         session_id: &str,
         after: Option<i64>,
-    ) -> Result<Vec<EventEnvelope>, CoreError> {
+    ) -> Result<Vec<WireEventEnvelope>, CoreError> {
         call(
             self.0,
             "openhuman.medulla_list_events",
--- a/src/embed/mod.rs
+++ b/src/embed/mod.rs
@@ -68,8 +68,8 @@ pub use harness::{Access, Harness, HarnessBuilder, HarnessError, Provider, Works
 pub use harness::{HttpHeader, McpAuthConfig, McpServer};
 #[cfg(feature = "medulla")]
 pub use medulla::{
-    AbortResult, EventEnvelope, Medulla, MedullaStatus, Message, RosterWorker, SendResult,
-    SessionCreated, SessionDetail, SessionSummary,
+    AbortResult, Medulla, MedullaStatus, Message, RosterWorker, SendResult, SessionCreated,
+    SessionDetail, SessionSummary, WireEventEnvelope,
 };
 
 use std::sync::Arc;
--- a/src/openhuman/medulla/client/sessions.rs
+++ b/src/openhuman/medulla/client/sessions.rs
@@ -100,7 +100,7 @@ impl MedullaClient {
         &self,
         session_id: &str,
         after: Option<i64>,
-    ) -> Result<Vec<EventEnvelope>> {
+    ) -> Result<Vec<WireEventEnvelope>> {
         let mut req = self
             .http
             .get(self.url(&format!("/medulla/v1/sessions/{session_id}/events")));
@@ -129,7 +129,7 @@ impl MedullaClient {
         &self,
         session_id: &str,
         last_event_id: Option<u64>,
-    ) -> impl Stream<Item = Result<EventEnvelope>> {
+    ) -> impl Stream<Item = Result<WireEventEnvelope>> {
         let url = format!(
             "{}/medulla/v1/sessions/{}/stream?token={}",
             self.base_url,
--- a/src/openhuman/medulla/client/sse/mod.rs
+++ b/src/openhuman/medulla/client/sse/mod.rs
@@ -11,7 +11,7 @@
 //! ```
 //!
 //! `id:` sets the replay cursor (persisted events only; deltas omit it),
-//! `data:` carries the JSON [`EventEnvelope`], comment lines (`: ping`) are
+//! `data:` carries the JSON [`WireEventEnvelope`], comment lines (`: ping`) are
 //! ignored, and a blank line terminates the current frame.
 
 use std::collections::VecDeque;
@@ -19,7 +19,7 @@ use std::collections::VecDeque;
 use futures::stream::{Stream, StreamExt};
 
 use crate::openhuman::medulla::client::error::{ClientError, Result};
-use crate::openhuman::medulla::client::types::EventEnvelope;
+use crate::openhuman::medulla::client::types::WireEventEnvelope;
 
 impl SseParser {
     /// Create an empty parser.
@@ -151,7 +151,7 @@ impl StreamState {
         if trimmed.is_empty() {
             return;
         }
-        match serde_json::from_str::<EventEnvelope>(trimmed) {
+        match serde_json::from_str::<WireEventEnvelope>(trimmed) {
             Ok(env) => self.pending.push_back(Ok(env)),
             Err(e) => self
                 .pending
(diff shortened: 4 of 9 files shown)


M3gA-Mind:
Review from the e2e wave side — I checked this against the verification block on #6078.

**The rename is complete and resolves the ambiguity as intended.** `medulla/mod.rs` is untouched, so the contract type keeps the plain `EventEnvelope` name and the client/wire type is now `WireEventEnvelope` — after this there is exactly one type called `EventEnvelope` in the domain, and it is the one the module publicly re-exports. That was the whole complaint. `ops.rs`, `client/sessions.rs`, `client/sse/{mod,types}.rs` and the decode tests all move together, and the SSE doc comment was updated rather than left pointing at the old name.

**One correction to our own report, in your favour to know about: this is a public API break, not an in-crate rename.**

The verification block on #6078 says *"A rename touches only in-crate paths — neither type crosses a serialization boundary under the name, so no wire compatibility is at stake, only source churn."* The first half is wrong. `src/embed/mod.rs:68-72` and `src/embed/medulla.rs:25-27` are `pub use`, and `src/embed` is documented as *"Typed embedding facade... real Rust types, so a host application never writes `serde_json::json!`"* — i.e. the host-application-facing API. So this changes `openhuman_core::embed::EventEnvelope` to `openhuman_core::embed::WireEventEnvelope` for anyone embedding the crate.

The serialization half of that sentence still holds — nothing on the wire is keyed by the type name, so there is no compatibility risk, only a source break for embedders. I am not arguing against the rename; `WireEventEnvelope` is the better name and the ambiguity was worth removing. It just deserves a release note, or a one-release deprecated alias if `embed` is treated as stable. Your call — but our blast radius told you this was cheaper than it is, and that is on us.

PR #6107, opened 2026-09-08
fix(skills): declare correct controller output names (new_workflows/workflow)

### LINKED ISSUE #6083: Two `skills_*` controllers advertise output field names the handler never emits
## Summary

Two `skills_*` controllers advertise output field names the handler never emits

- **Namespace / method:** openhuman.skills_install_from_url, openhuman.skills_create
(and openhuman.skills_update, which clones create's schema)
- **Where:** * `src/openhuman/skills/schemas/controller_schemas.rs:330-336` — declares output `new_skills`
* `src/openhuman/skills/schemas/handlers.rs:302-310` — emits `new_workflows`
* `src/openhuman/skills/schemas/wire_types.rs:218-224` — `WorkflowsInstallFromUrlResult.new_workflows`, no serde rename
* `src/openhuman/skills/schemas/controller_schemas.rs:270-275` — `skills_create` declares output `skill`
* `src/openhuman/skills/schemas/wire_types.rs:213-216` — `WorkflowsCreateResult.workflow`

## Problem

**Expected**

the controller catalog's declared output names match the JSON the
handler returns. The catalog is not decorative — `skill_runtime_schemas`'s own
description says the schemas exist "for CLI/RPC smoke-test script generation",
so a wrong name there generates a script that reads a field that is never sent.

**Actual**

they disagree, and the frontend sides with the handler:
`app/src/services/api/skillsApi.ts:400` reads `raw.new_workflows`, and
`app/src/types/turnState.ts` has no `skill` key. So the *runtime* is
self-consistent and only the published schema is wrong — which is the failure
mode nobody notices, because every real caller was written against the wire
rather than the catalog.

This looks like fallout from the skills→workflows rename: the wire types kept
`workflow`/`new_workflows`, the schemas were written with the newer `skill`
vocabulary, and nothing cross-checks the two.

Note the contrast with the `todos` namespace, where a single declared output
names an unwrapped payload consistently across every method — that is a
convention, this is a mismatch.

**Steps to reproduce**

`tests/raw_coverage/skill_runtime_e2e.rs` →
`skills_install_from_url_enforces_its_url_policy_then_installs`, which asserts
`new_workflows` is present AND `new_skills` is absent, so the day the two are
reconciled the suite fails loudly rather than drifting.

**Cited location still accurate:** yes, with small drift. `new_skills` is declared at
`src/openhuman/skills/schemas/controller_schemas.rs:331` (report said 330-336);
`new_workflows` is the struct field at `wire_types.rs:223` (report said 218-224) and
is set at `handlers.rs:307` from `outcome.new_skills`. `skills_create` declares
output `skill` at `controller_schemas.rs:~277`; `WorkflowsCreateResult.workflow` is
at `wire_types.rs:214-216`.

**Repro run:** `cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- W5VERIFY_`

**Observed**

```
test W5VERIFY_skill_runtime_e2e::skills_install_from_url_enforces_its_url_policy_then_installs ... ok
```
Neither wire struct carries a `rename_all` or per-field `serde(rename)` — both are a
bare `#[derive(Debug, Serialize)]` — so the emitted names are exactly the Rust field
names, `new_workflows` and `workflow`, and the declared names `new_skills` / `skill`
are never sent. Note the amusing detail at `handlers.rs:307`:
`new_workflows: outcome.new_skills` — the drift is visible on a single line.

**Diagnosis assessment:** correct, including the rename-fallout explanation.

**Is the expected behaviour correct:** yes, and I specifically checked the report's
load-bearing claim that the frontend sides with the wire, because a counter-example
would have flipped the fix direction. It holds, but the evidence is subtler than the
report states: **there are two frontend clients using the two different spellings.**
- `app/src/services/api/skillsApi.ts:172,400` reads `new_workflows` — and this is the
  one that calls `openhuman.skills_install_from_url`.
- `app/src/services/api/skillRegistryApi.ts:72,194` reads `new_skills`, with a test
  pinning it (`skillRegistryApi.test.ts:15`) — but it calls
  `openhuman.skill_registry_install`, a **different namespace**.
So there is no conflict, and the report's conclusion (fix the schema, not the wire)
is right. Worth recording because `grep new_skills app/src` alone looks like a
contradiction.

## Solution

change `controller_schemas.rs` to declare
`new_workflows` and `workflow`. Do **not** rename the wire fields — that breaks
`skillsApi.ts:400`.

**Blast radius:** schema-only change. Consumers of the declared names are the
controller catalog and the CLI/RPC smoke-test generation the schema docstring cites;
no runtime caller reads them, which is why this survived. `skills_update` clones
`skills_create`'s schema so it inherits the same fix. Same class as
`e2e-wave-mcp-setup-install-status-field-never-emitted.md` — one
declared-vs-serialized gate would catch both.

## Acceptance criteria

## Related

YellowSnnowmann:
## Summary

- Fixes the `skills_install_from_url` and `skills_create` controller schemas to declare the output field names the handlers actually emit: `new_workflows` and `workflow` (they declared `new_skills` and `skill`).
- Schema-only change — no wire struct, handler, or frontend edits. `skills_update` inherits the fix by cloning `skills_create`'s schema.

## Problem

Fallout from the skills→workflows rename: the wire structs kept `new_workflows` / `workflow` (bare `#[derive(Serialize)]`, no serde rename) and the frontend reads those names (`app/src/services/api/skillsApi.ts`), but the controller catalog was written with the newer `skill` vocabulary and nothing cross-checks the two. The catalog exists "for CLI/RPC smoke-test script generation", so a wrong name there generates a script that reads a field never sent — the failure mode nobody notices because every real caller was written against the wire. (`skillRegistryApi.ts` reads `new_skills`, but it calls the different `skill_registry_install` namespace and is untouched.)

## Solution

- In `controller_schemas.rs`, change the `skills_install_from_url` output field `new_skills` → `new_workflows` and the `skills_create` output field `skill` → `workflow` (names + comments only; `TypeSchema` unchanged). `skills_update` clones `skills_create` and inherits the fix.
- Regression test pins the declared output names to the wire reality (`new_workflows`/`workflow` present, `new_skills`/`skill` absent) across create/update/install so the drift can't silently return.

## Submission Checklist

- [x] Tests added or updated — schema-output-name assertions across the three controllers.
- [x] **Diff coverage ≥ 80%** — the two changed schema lines are covered by the new assertions. (Full `diff-cover` runs in CI.)
- [x] Coverage matrix updated — `N/A: schema-name fix, no feature row added/removed`.
- [x] Manual smoke checklist updated — `N/A: catalog-metadata fix, no runtime change`.

## Impact

- Published controller catalog / `/schema` now matches the JSON the handlers emit, so generated smoke scripts read the right fields. No runtime, wire, or frontend behaviour change (the frontend already read the wire names).

## Related

- Closes: #6083

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: fix/skills-schema-output-names-6083

## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed schema output fields from `skill` to `workflow` and from `new_skills` to `new_workflows`.
  * Clients consuming these schemas must update references to the renamed fields.

* **Tests**
  * Added coverage confirming the new workflow-oriented output names and preventing use of the previous names.

### DIFF
--- a/src/openhuman/skills/schemas/controller_schemas.rs
+++ b/src/openhuman/skills/schemas/controller_schemas.rs
@@ -275,9 +275,9 @@ pub fn skills_schemas(function: &str) -> ControllerSchema {
                 },
             ],
             outputs: vec![FieldSchema {
-                name: "skill",
+                name: "workflow",
                 ty: TypeSchema::Ref("WorkflowSummary"),
-                comment: "The newly created skill, re-discovered through the standard pipeline.",
+                comment: "The newly created workflow, re-discovered through the standard pipeline.",
                 required: true,
             }],
         },
@@ -328,9 +328,9 @@ pub fn skills_schemas(function: &str) -> ControllerSchema {
                     required: true,
                 },
                 FieldSchema {
-                    name: "new_skills",
+                    name: "new_workflows",
                     ty: TypeSchema::Array(Box::new(TypeSchema::String)),
-                    comment: "Slugs of skills that appeared in the catalog as a result of the install.",
+                    comment: "Slugs of workflows that appeared in the catalog as a result of the install.",
                     required: true,
                 },
             ],
--- a/src/openhuman/skills/schemas_tests.rs
+++ b/src/openhuman/skills/schemas_tests.rs
@@ -32,3 +32,47 @@ fn skill_summary_round_trip_minimum_fields() {
     assert_eq!(summary.name, "demo");
     assert_eq!(summary.description, "desc");
 }
+
+/// Pins the declared controller-schema output names to the workflow-spelled
+/// wire reality. The skills→workflows rename left two schemas emitting the old
+/// spelling (`new_skills` / `skill`) while the handlers and wire structs
+/// (`WorkflowsInstallFromUrlResult.new_workflows`, `WorkflowsCreateResult.workflow`)
+/// switched to the new one — a silent schema/wire divergence the frontend
+/// papered over by reading the wire names directly. These assertions fail on
+/// the pre-rename schema and stop it from drifting back.
+#[test]
+fn install_and_create_schemas_declare_workflow_output_names() {
+    // Arrange
+    let install = skills_schemas("skills_install_from_url");
+    let create = skills_schemas("skills_create");
+    let update = skills_schemas("skills_update");
+
+    // Act
+    let has_output = |schema: &crate::core::ControllerSchema, name: &str| {
+        schema.outputs.iter().any(|f| f.name == name)
+    };
+
+    // Assert — install echoes the workflow-spelled slug list, never the old one.
+    assert!(
+        has_output(&install, "new_workflows"),
+        "skills_install_from_url must declare `new_workflows`; outputs: {:?}",
+        install.outputs.iter().map(|f| f.name).collect::<Vec<_>>()
+    );
+    assert!(
+        !has_output(&install, "new_skills"),
+        "skills_install_from_url must not declare the pre-rename `new_skills`"
+    );
+
+    // create / update share one schema; both carry the `workflow` output.
+    for (function, schema) in [("skills_create", &create), ("skills_update", &update)] {
+        assert!(
+            has_output(schema, "workflow"),
+            "{function} must declare `workflow`; outputs: {:?}",
+            schema.outputs.iter().map(|f| f.name).collect::<Vec<_>>()
+        );
+        assert!(
+            !has_output(schema, "skill"),
+            "{function} must not declare the pre-rename `skill`"
+        );
+    }
+}


M3gA-Mind:
Reviewed as the verifier of #6083 — I wrote the independent verification block on that issue, so I checked the diff against the blast radius we recorded rather than re-deriving it.

**This is correct and complete as far as I can establish.** Concretely, what I checked:

- **The fix direction matches the verified diagnosis.** The issue's conclusion was that the *runtime* is self-consistent and only the published schema is wrong — `skillsApi.ts:400` reads `raw.new_workflows`, and neither `WorkflowsInstallFromUrlResult` nor `WorkflowsCreateResult` carries a `rename_all` or per-field `serde(rename)`, so the emitted names are the bare Rust field names. Renaming the *declaration* and leaving the wire alone is the fix that keeps the frontend working. This diff does exactly that and touches no wire struct.
- **`skills_update` is covered.** The blast radius flagged that `skills_update` clones `skills_create`'s schema and therefore inherits the fix; the new test asserts both explicitly rather than assuming the clone, which is the right level of paranoia given that the clone is the thing that could change.
- **The test bites.** `has_output(&install, "new_workflows")` is false on the pre-fix schema, so it fails without the change; the paired `!has_output(..., "new_skills")` stops a drift back. Both directions asserted for all three methods.
- **No test was written to match current behaviour.** The assertions are against the intended contract.

**One thing I checked and want to record as a non-finding**, because it looks like an inconsistency and is not: sibling PR #6094 (same class of fix, same day) also updates `app/schema.json`, and this one does not. I went looking for a stale catalog entry and there isn't one — `app/schema.json` contains 58 methods and **no `skills` entries at all** (it has `cron.run`, which is why #6094 had to touch it). So there is genuinely nothing to update here. Correct call, and I mention it only so nobody else spends the same ten minutes.

The only residual is the one the issue itself named: nothing cross-checks a declared output name against the serialized field, so this class recurs. That is the "one declared-vs-serialized gate would catch both" note on #6083 and shared with #6110 — a separate piece of work, not this PR's job.

Issue #6091, opened 2026-09-08
`webhooks_unregister_echo` reports success for a tunnel that was never registered


M3gA-Mind:
## Summary

`webhooks_unregister_echo` reports success for a tunnel that was never registered

Found during the e2e coverage wave and **independently verified by a second engineer** who did not
write the original report — verification was cross-assigned so nobody checked their own finding.

- **Namespace / method:** `openhuman.webhooks_unregister_echo`
- **Where:** `src/openhuman/skills/webhooks/router.rs:223-255` (`WebhookRouter::unregister`) →
`src/openhuman/skills/webhooks/ops.rs:124-135`
- **Verified at:** `openhuman` `main` @ `d1ed68f9b`

## Problem

**Expected**

either an error, or a result the caller can tell apart from a real removal — the
same shape the rest of the product uses for this exact situation. `notification_dismiss`,
`notification_mark_acted` and `notification_core_mark_read` all return `{ "ok": false }` when
the id matched no row (`src/openhuman/desktop/notifications/rpc.rs:344-435`).

**Actual**

`unregister` takes the "not found" branch, logs at `debug!`, and falls through to
`Ok(())` — then still publishes a `registration_changed` event, re-persists the unchanged route
file, and publishes `DomainEvent::WebhookUnregistered` for a tunnel that was never registered.
The RPC returns the (unchanged) registration list with nothing to distinguish it from a
successful removal.

The knock-on is the bus event, not the return value: a subscriber sees `WebhookUnregistered`
for a tunnel/skill pair that never had a registration.

**Steps to reproduce**

call `openhuman.webhooks_unregister_echo` with
`{"tunnel_uuid": "never-registered"}` — the response is a successful
`{"registrations": [...]}` identical to the one before the call.

### Independent verification (CONFIRMED)

**Cited location still accurate:** yes, exactly. `WebhookRouter::unregister` is
`src/openhuman/skills/webhooks/router.rs:223-255`; the not-found `else` branch that only
`debug!`s is `:238-243`; the three unconditional statements after it are `publish_event(
"registration_changed", ...)` at `:246`, `self.persist()` at `:247`, and
`BUS.publish(DomainEvent::WebhookUnregistered { .. })` at `:249-252`, then `Ok(())` at `:254`.
The RPC wrapper is `src/openhuman/skills/webhooks/ops.rs:124-135`. The notification
precedent is real too: `handle_dismiss` at
[...]

PR #6114, opened 2026-09-08
fix(webhooks): honour an explicit list_logs limit of 0

### LINKED ISSUE #6090: `webhooks_list_logs` with `limit: 0` returns one entry instead of none
## Summary

`webhooks_list_logs` with `limit: 0` returns one entry instead of none

- **Namespace / method:** `openhuman.webhooks_list_logs`
- **Where:** `src/openhuman/skills/webhooks/router.rs:458-465`

## Problem

**Expected**

`limit: 0` returns zero entries (or is rejected as invalid). The declared input is
`Option<U64>`, "Maximum number of log entries to return" — a maximum of zero is zero.

**Actual**

```rust
let limit = limit.unwrap_or(100).max(1);
```

`.max(1)` promotes an explicit `0` to `1`, so the caller asking for nothing gets one log entry.
The clamp reads as intended to protect the *default*, but `unwrap_or(100)` already does that —
`.max(1)` can only ever alter a caller-supplied `0`.

**Steps to reproduce**

`post_json_rpc(rpc, id, "openhuman.webhooks_list_logs", json!({ "limit": 0 }))` against a router
with at least one recorded request returns `logs.len() == 1`.

**Cited location still accurate:** yes — `WebhookRouter::list_logs` is at
`src/openhuman/skills/webhooks/router.rs:458-464` and line `459` is verbatim
`let limit = limit.unwrap_or(100).max(1);`, followed by `.take(limit)`.

**Repro run:** a probe recording three webhook requests and then listing with several limits over
the real RPC router (`cargo test --test raw_coverage_all --features
"$(bash scripts/ci/product-features.sh)" -- --nocapture probe_bug5`)

**Observed**

```
BUG5 params={"limit":0} -> logs.len()=Some(1)
BUG5 params={"limit":1} -> logs.len()=Some(1)
BUG5 params={"limit":2} -> logs.len()=Some(2)
BUG5 params={}          -> logs.len()=Some(3)
```
`limit: 0` and `limit: 1` are indistinguishable; every other value is honoured exactly. Three
entries were recorded, so the omitted-limit row confirms nothing else is truncating.

**Diagnosis assessment:** right, and I traced the whole path to rule out an earlier interception:
`handle_list_logs` (`schemas.rs:386-390`) deserialises into `WebhookListLogsParams { limit:
Option<usize> }` (`:11`) → `ops::list_logs(Some(0))` (`ops.rs:76-81`) → `router.list_logs(Some(0))`.
Nothing validates or rewrites the value between the wire and `.max(1)`, so the clamp is the sole
cause. The report's reasoning about intent is also correct: `unwrap_or(100)` already handles the
absent case, so `.max(1)` can only ever alter a caller-supplied `0` — and `.take(0)` is
perfectly safe on an iterator, so it is not guarding against anything.

**Is the expected behaviour correct:** yes. The declared input comment is "Maximum number of log
entries to return"; a maximum of zero is zero. The report's alternative — rejecting `0` as
invalid — is equally defensible and equally honest; what is not defensible is silently returning
one. Severity `cosmetic` is fair: a caller paginating from zero gets one unexpected row, and
nothing is lost or corrupted.

## Solution

drop `.max(1)` so the value is `limit.unwrap_or(100)` and
`.take(0)` yields an empty list; or reject `limit: 0` at the handler with a typed error. The
former is one token and preserves the "maximum" reading of the parameter.

**Blast radius:** minimal. `list_logs` has exactly one non-test caller — `ops.rs:81`. No existing
unit test passes `0`: `router_tests.rs:172` uses `Some(3)`, `:189` uses `None`, and the others
use `Some(1)`/`Some(10)`/`Some(MAX_DEBUG_LOG_ENTRIES + 100)` — so removing the clamp breaks no
test in the tree. W2's `webhooks_debug_log_ring_records_and_clears` uses `limit: 50` and an
omitted limit, so it is unaffected too.

## Acceptance criteria

## Related

7487:
## Summary

- `webhooks_list_logs` clamped a caller-supplied `limit: 0` up to `1`, so a caller asking for nothing got one log entry.
- Drops the `.max(1)` in `WebhookRouter::list_logs`; `unwrap_or(100)` already covers the absent case, so the clamp could only ever rewrite an explicit `0`.
- Adds a router unit test for `Some(0)` and a wire-level assertion in the existing webhooks debug-log e2e test.

## Problem

`src/openhuman/skills/webhooks/router.rs:459` read:

```rust
let limit = limit.unwrap_or(100).max(1);
```

`limit` is declared `Option<U64>`, "Maximum number of log entries to return" (`schemas.rs:151-160`). A maximum of zero is zero, but `.max(1)` promoted an explicit `0` to `1`, so `{"limit": 0}` returned one entry against a router with any recorded request.

The clamp reads as protecting the default, but `unwrap_or(100)` already does that — the only value `.max(1)` could alter was a caller-supplied `0`. `.take(0)` is well-defined on an iterator, so it was not guarding against anything either.

I traced the full path to confirm nothing between the wire and the clamp rewrites the value:
`handle_list_logs` (`schemas.rs:386`) → `WebhookListLogsParams { limit: Option<usize> }` (`schemas.rs:11`) → `ops::list_logs` (`ops.rs:76-81`) → `router.list_logs`. The clamp was the sole cause.

Reproduced before the fix: the new unit test failed with `assertion failed: router.list_logs(Some(0)).is_empty()` while the other 25 router tests passed.

## Solution

One-token change — `limit.unwrap_or(100)` — so `.take(0)` yields an empty list. This preserves the "maximum" reading of the parameter, which the declared input comment already promises; rejecting `0` with a typed error would be equally defensible but is a louder API change for a cosmetic defect.

Coverage added at both levels, because they fail differently:

- `list_logs_zero_limit_returns_nothing` (`router_tests.rs`) pins the clamp itself.
- The `limit: 0` assertion in `webhooks_debug_log_ring_records_and_clears` pins the wire path — specifically that `0` deserialises to `Some(0)` and not an absent limit, which the unit test cannot see and which would silently return 100 entries instead of 0.

## Impact

Runtime: Rust core only, no frontend or Tauri surface. No migration or compatibility concern — the only behaviour change is that `limit: 0` now returns `[]` instead of one entry.

Blast radius: `list_logs` has exactly one non-test caller, `ops.rs:81`. No existing test passes `0` — `router_tests.rs` uses `Some(3)`, `None`, `Some(1)`, `Some(10)` and `Some(MAX_DEBUG_LOG_ENTRIES + 100)`, and the e2e log-ring test uses `limit: 50` and an omitted limit. All still pass.

## Related

- Closes: #6090
- Feature IDs: 13.4.1 Webhook Inspection (`docs/TEST-COVERAGE-MATRIX.md`) — already `✅`, unchanged by this PR.
- Follow-up PR(s)/TODOs: `src/openhuman/cron/ops.rs:318` has the same shape (`limit.unwrap_or(20).max(1)`), but `cron::list_runs` clamps again at `cron/store_part_01.rs:628`, so fixing it is a separate two-site change. Left out to keep this scoped to #6090 — happy to file it if wanted.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — `limit: 0` (the edge case) plus the existing populated/empty/default-limit assertions in the same tests.
- [x] **Diff coverage ≥ 80%** — the changed line in `router.rs` is executed by `list_logs_zero_limit_returns_nothing`, `list_logs_respects_limit`, `list_logs_default_limit` and `truncate_logs_respects_max`; the changed test lines execute themselves.
- [x] Coverage matrix updated — `N/A: behaviour-only change`, no feature leaf added, removed or renamed.
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: does not touch a release-cut surface.`
- [x] Linked issue closed via `Closes #6090` in the `## Related` section

---

This PR was authored by Claude Code.

### Linear Issue

- Key: `N/A` — filed on GitHub, not Linear.
- URL: https://github.com/tinyhumansai/openhuman/issues/6090

### Commit & Branch

- Branch: `fix/6090-webhooks-list-logs-limit-zero`
- Commit SHA: `10f4d8c526e992f4ec117cc5de9ffd78b163c297`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — `N/A: no frontend files changed.`
- [x] Focused tests:
  - `cargo test --lib webhooks::` → 84 passed, 0 failed, 6 ignored.
  - `cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" webhooks_debug_log_ring` → 1 passed, 0 failed.
  - Both new assertions were confirmed to bite: with `.max(1)` restored, the unit test fails (`assertion failed: router.list_logs(Some(0)).is_empty()`) and the e2e test fails with `left: Some(1), right: Some(0)` — the issue's reported repro, reproduced through the real RPC router.
- [x] Rust fmt/check: `cargo fmt --manifest-path Cargo.toml --all --check` clean (exit 0).
- [x] Tauri fmt/check (if changed): `N/A: app/src-tauri untouched.`

### Validation Blocked

- `command:` `rustup toolchain install 1.96.1`
- `error:` `could not download file from 'https://static.rust-lang.org/dist/channel-rust-1.96.1.toml' ... peer closed connection without sending TLS close_notify`
- `impact:` The pinned toolchain would not download here, so everything above ran on **1.96.0**, which still satisfies the rusqlite/`cfg_select!` floor documented in `rust-toolchain.toml`. CI should be treated as the authority on the pinned version.

- `command:` `cargo test --lib` (whole-crate run)
- `error:` `thread '...cron::scheduler::tests::part_02_tests::run_agent_job_returns_error_without_provider_key' has overflowed its stack / fatal runtime error: stack overflow`
- `impact:` Pre-existing and unrelated — I confirmed it reproduces on a clean `upstream/main` checkout with my two source changes stashed, and it is in `cron::scheduler`, which has no call path to `webhooks`. It aborts the test binary, so I could not get a single clean whole-crate run locally; the webhooks module and the changed line are fully covered by the focused runs above.

### Behavior Changes

- Intended behavior change: `webhooks_list_logs` with `limit: 0` returns zero entries instead of one.
- User-visible effect: A caller paginating from zero no longer receives one unexpected row. No UI surface passes `0` today, so nothing in the app changes.

### Parity Contract

- Legacy behavior preserved: every other `limit` value, including an omitted limit, is unchanged — `unwrap_or(100)` still supplies the default and `MAX_DEBUG_LOG_ENTRIES` truncation is untouched.
- Guard/fallback/dispatch parity checks: `.take(limit)` is safe at `0`, so removing the clamp introduces no panic or empty-iterator edge; the `RwLock` read-failure fallback to `unwrap_or_default()` is unchanged.

- Duplicate PR(s): none — no open PR referenced #6090 at the time of writing.
- Canonical PR: this one.
- Resolution (closed/superseded/updated): `N/A`

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook log requests with a limit of `0` now correctly return no entries instead of the newest entry.
  * Debug-log behavior is now consistently validated for zero-entry requests.

### DIFF
--- a/src/openhuman/skills/webhooks/router.rs
+++ b/src/openhuman/skills/webhooks/router.rs
@@ -456,7 +456,7 @@ impl WebhookRouter {
 
     /// List recent webhook logs, newest first.
     pub fn list_logs(&self, limit: Option<usize>) -> Vec<WebhookDebugLogEntry> {
-        let limit = limit.unwrap_or(100).max(1);
+        let limit = limit.unwrap_or(100);
         self.debug_logs
             .read()
             .map(|logs| logs.iter().take(limit).cloned().collect())
--- a/src/openhuman/skills/webhooks/router_tests.rs
+++ b/src/openhuman/skills/webhooks/router_tests.rs
@@ -186,6 +186,23 @@ fn list_logs_respects_limit() {
 }
 
 #[test]
+fn list_logs_zero_limit_returns_nothing() {
+    let router = WebhookRouter::new(None);
+    for i in 0..3 {
+        router.record_parse_error(
+            format!("corr-{i}"),
+            None,
+            None,
+            None,
+            json!({}),
+            "error".into(),
+        );
+    }
+    // `limit` is a maximum, so an explicit 0 asks for nothing.
+    assert!(router.list_logs(Some(0)).is_empty());
+}
+
+#[test]
 fn list_logs_default_limit() {
     let router = WebhookRouter::new(None);
     for i in 0..5 {
--- a/tests/raw_coverage/webhooks_ingress_e2e.rs
+++ b/tests/raw_coverage/webhooks_ingress_e2e.rs
@@ -678,8 +678,24 @@ async fn webhooks_debug_log_ring_records_and_clears() {
         "no response recorded yet ⇒ status_code must still be null: {entry}"
     );
 
+    // `limit` is a maximum, so an explicit 0 asks for nothing — and it must survive
+    // the wire as `Some(0)` rather than being read as an absent limit.
+    let none = post_json_rpc(
+        &rpc_base,
+        7104,
+        "openhuman.webhooks_list_logs",
+        json!({ "limit": 0 }),
+    )
+    .await;
+    let result = peel(assert_no_jsonrpc_error(&none, "webhooks_list_logs (limit 0)"));
+    assert_eq!(
+        result.get("logs").and_then(Value::as_array).map(Vec::len),
+        Some(0),
+        "limit: 0 must return no entries even with one recorded: {result}"
+    );
+
     // clear_logs reports the count it removed, and the ring is empty afterwards.
-    let cleared = post_json_rpc(&rpc_base, 7104, "openhuman.webhooks_clear_logs", json!({})).await;
+    let cleared = post_json_rpc(&rpc_base, 7105, "openhuman.webhooks_clear_logs", json!({})).await;
     let result = peel(assert_no_jsonrpc_error(&cleared, "webhooks_clear_logs"));
     assert_eq!(
         result.get("cleared").and_then(Value::as_u64),
@@ -687,7 +703,7 @@ async fn webhooks_debug_log_ring_records_and_clears() {
         "clear_logs must report the number of entries it removed: {result}"
     );
 
-    let after = post_json_rpc(&rpc_base, 7105, "openhuman.webhooks_list_logs", json!({})).await;
+    let after = post_json_rpc(&rpc_base, 7106, "openhuman.webhooks_list_logs", json!({})).await;
     let result = peel(assert_no_jsonrpc_error(&after, "webhooks_list_logs (after)"));
     assert_eq!(
         result.get("logs").and_then(Value::as_array).map(Vec::len),


7487:
Superseded by #6115, closing.

#6115 opened ten minutes after this one and merged first (13:45:52Z), carrying the identical fix — `router.rs:474` on main now reads `let limit = limit.unwrap_or(100);` — plus a superset of the coverage here:

- `list_logs_limit_zero_returns_no_entries` in `router_tests.rs`, which also pins limits 1, 2 and absent, so it cannot pass by returning nothing for everything. Mine only asserted `Some(0)`.
- `limit: 0` and `limit: 1` over the JSON-RPC surface in `webhooks_ingress_e2e.rs`, the same wire-path property my e2e assertion covered.

It also fixes #6091 in the same router, which this PR did not touch. #6090 is closed as completed. That leaves nothing to rebase: the conflict on all three files is my change against the same change, so there is no version of this branch that still adds anything.

On the red CI here, since the same failure hit several PRs at once: `flows_lifecycle_e2e::flows_approval_manifest_classifies_every_gated_node_kind` and `flows_tool_catalog_surface_degrades_without_composio_credentials` were inherited from the base commit, not caused by this diff — it never touched `flows`. 59ef1f91 ("test(flows): invert the two e2e assertions #6093 fixed") has since fixed both on main.

Unrelated heads-up while I was checking main: Rust Core Coverage is red on `main` @ `7ca0df7` for a different reason — `raw_coverage_all` no longer compiles, `error[E0451]: fields workspace_dir and system_prompt of struct DispatchHarnessOptions are private`. a45b1c1a made those two fields `pub(crate)` on a `pub` struct, and the `tests/raw_coverage/channels_*` files are a separate crate, so they cannot see them. Happy to send a one-line fix if nobody is on it.

PR #6115, opened 2026-09-08
fix(webhooks): gate the unregister side-effect tail on a real removal, and stop clamping list_logs' limit to 1

### LINKED ISSUE #6090: `webhooks_list_logs` with `limit: 0` returns one entry instead of none

### LINKED ISSUE #6091: `webhooks_unregister_echo` reports success for a tunnel that was never registered

M3gA-Mind:
## Summary

- **#6091** — `WebhookRouter::unregister` ran its whole side-effect tail (`registration_changed`, `persist()`, `DomainEvent::WebhookUnregistered`) even on the branch where the tunnel was never registered. It now returns `Ok(true)`/`Ok(false)` and the tail is gated on a real removal.
- **#6090** — `list_logs` clamped `limit` with `.max(1)`, so `limit: 0` returned one entry. The clamp is removed.
- Both are in `src/openhuman/skills/webhooks/router.rs`, which is why they share a PR.
- No wire-shape change on either. Five new/updated assertions, each revert-checked.

## Problem

**#6091.** `unregister` logged at `debug!` on the not-found branch and then fell through to code that announces a change:

```rust
} else {
    debug!("[webhooks] Tunnel {} not found for unregister …");
}
drop(routes);
self.publish_event("registration_changed", …);   // fires either way
self.persist();                                  // fires either way
BUS.publish(DomainEvent::WebhookUnregistered {…});// fires either way
Ok(())                                           // same as a real removal
```

So a caller with a typo'd UUID made every subscriber believe a registration had been torn down, and could not tell that from a real removal — both answered `Ok(())`.

**#6090.** `let limit = limit.unwrap_or(100).max(1);` — the absent case is already handled by `unwrap_or(100)`, so `.max(1)` could *only* ever rewrite a caller-supplied `0` into a `1`. `limit: 0` and `limit: 1` were indistinguishable while every other value was honoured exactly.

## Solution

**#6091** — early-return `Ok(false)` from the not-found branch so the tail is unreachable for it, and `Ok(true)` after a real removal. This is the gate the sibling `unregister_skill` already applies nine lines below with its `if !removed_tunnels.is_empty()` — the same file and the same type already knew to condition side effects on "did anything change".

The **wire shape is deliberately unchanged.** The issue leaves that a separate contract call, and #6095 set the precedent for the conservative direction (fix the declaration, never the wire). `unregister_echo` still returns the full registration list for the caller to diff; what changes is that its log line no longer claims a removal that did not happen:

```
webhooks.unregister_echo: no registration for tunnel <uuid>, nothing removed
```

**#6090** — drop `.max(1)`. `Iterator::take(0)` is well-defined, so nothing was being guarded against.

## Impact

- **Blast radius walked, and narrower than the issue estimated.** The issue said `unregister` has "two callers — `ops::unregister_echo` and the agent-tunnel equivalent". There is **one**: `ops.rs:129`. `grep -rn '\.unregister('` over `src/` returns that line and nothing else (the agent path registers but never unregisters).
- **No subscriber acts on `DomainEvent::WebhookUnregistered`.** The only references in the tree are the variant declaration (`src/core/events.rs:825`) and two classification arms (`:1482` domain tag, `:1629` name). So narrowing the publish changes the event log and nothing else.
- **No existing test passed `limit: 0`** — `router_tests.rs` uses `Some(3)`, `None`, `Some(1)`, `Some(10)`, `Some(MAX_DEBUG_LOG_ENTRIES + 100)`. Removing the clamp broke none of them.
- The module README's "Publishes" line is updated, since it documented the unconditional behaviour.

## Related

- Closes: #6090
- Closes: #6091
- Follow-up PR(s)/TODOs: the `unregister_echo` **wire shape** (return the boolean the way `notification_dismiss` does, vs. leaving the caller to diff `registrations`) is left open on #6091 as a contract call.

---

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — see the revert-check table below; every one covers the no-op/edge path as well as the real one.
- [x] **Diff coverage ≥ 80%** — every changed production line is executed by the new tests (`list_logs_limit_zero_returns_no_entries` covers the `limit` line; `unregister_of_an_absent_tunnel_announces_nothing` covers both branches of `unregister`; the e2e cases cover the `ops.rs` log branch). Ran the lib + `raw_coverage_all` targets under the product feature string, **not** the full `pnpm test:coverage` / `pnpm test:rust` `diff-cover` merge — CI computes the real number.
- [x] N/A: no feature rows added, removed or renamed — this is a behaviour fix to existing controllers, so `docs/TEST-COVERAGE-MATRIX.md` needs no row change.
- [x] N/A: no matrix feature IDs apply, per the item above.
- [x] No new external network dependencies introduced — the router tests are in-process; the e2e cases use the existing in-process axum mock.
- [x] N/A: no release-cut surface touched — no UI, no packaging, no migration, so `docs/RELEASE-MANUAL-SMOKE.md` is unaffected.
- [x] Linked issue closed via `Closes` in the `## Related` section — #6090 and #6091.

## Revert-check — both fixes, faults injected and restored

Run: `cargo test --lib --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- webhooks socket`

| Fault injected | Test that failed | What the failure said |
|---|---|---|
| `.max(1)` put back on `list_logs` | `list_logs_limit_zero_returns_no_entries` | `router_tests.rs:394` — `limit 0 is a maximum of zero` |
| absent branch made to announce + return `Ok(true)` | `unregister_of_an_absent_tunnel_announces_nothing` | `router_tests.rs:341` — `unregistering an absent tunnel must report that nothing was removed` |
| (same fault) | `unregister_nonexistent_tunnel_is_noop` | `router_tests.rs:297` — `assertion failed: !router.unregister("no-such", "any-skill").unwrap()` |

Baseline with both fixes in place: **256 passed / 0 failed** (lib), **13 passed / 0 failed** (`raw_coverage_all`, `webhooks`+`socket` filter). Files restored from snapshots afterwards, not via `git checkout`, so the fix itself was never reverted.

### One thing worth knowing about the test harness

`WEBHOOK_DEBUG_EVENTS` is a **process-global** `broadcast::Sender`, not per-`WebhookRouter`. My first version of the absent-tunnel test asserted the receiver was empty and failed — a sibling suite running concurrently in the same lib-test process had published into it. The test now filters on its own tunnel UUIDs. Anyone writing a `subscribe_debug_events()` assertion should do the same.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unregistering a webhook that does not exist is now a silent no-op and clearly reports that nothing was removed.
  * Registration updates, persistence, and webhook removal events now occur only after an actual registration is removed.
  * Log requests with a limit of zero now correctly return no entries.

* **Documentation**
  * Updated webhook event documentation to clarify unregister behavior and no-op handling.

* **Tests**
  * Added coverage for absent webhook removal, event notifications, registration preservation, and log limits.

### DIFF
--- a/src/openhuman/skills/webhooks/README.md
+++ b/src/openhuman/skills/webhooks/README.md
@@ -64,7 +64,7 @@ None. This domain owns no `tools.rs` agent tools.
 Subscriber (in `bus.rs`): **`WebhookRequestSubscriber`** — `name() = "webhook::request_handler"`, `domains() = ["webhook"]`. Registered at startup (see `channels/runtime/startup.rs`).
 
 - **Subscribes**: `DomainEvent::WebhookIncomingRequest` (published by the socket transport in `socket/event_handlers.rs`).
-- **Publishes**: `DomainEvent::WebhookRegistered` / `WebhookUnregistered` (from the router on registration changes), `DomainEvent::WebhookReceived` (when routed to a target), `DomainEvent::WebhookProcessed` (always, with status/elapsed/error).
+- **Publishes**: `DomainEvent::WebhookRegistered` / `WebhookUnregistered` (from the router on registration changes — `WebhookUnregistered`, the `registration_changed` debug event and the route re-persist all fire **only when a registration was actually removed**; unregistering an absent tunnel is a silent no-op that returns `Ok(false)`, see #6091), `DomainEvent::WebhookReceived` (when routed to a target), `DomainEvent::WebhookProcessed` (always, with status/elapsed/error).
 
 Routing outcomes by `target_kind`: `echo` → `build_echo_response` (200); `agent` → decode body, spawn triage, return `202 Accepted` (spawned task emits the real response later, with 60s timeout → 504); `skill` / unknown → `501` (direct skill dispatch not available); no registration → `404`. Responses are emitted over the socket as `webhook:response`.
 
--- a/src/openhuman/skills/webhooks/ops.rs
+++ b/src/openhuman/skills/webhooks/ops.rs
@@ -126,11 +126,21 @@ pub async fn unregister_echo(
     tunnel_uuid: &str,
 ) -> Result<RpcOutcome<WebhookDebugRegistrationsResult>, String> {
     let router = get_router().map_err(|e| format!("webhooks.unregister_echo failed: {e}"))?;
-    router.unregister(tunnel_uuid, "echo")?;
+    let removed = router.unregister(tunnel_uuid, "echo")?;
     let registrations = router.list_all();
+    // The wire shape is unchanged (#6091 leaves that a separate contract call):
+    // the caller still gets the full registration list and can diff it. What
+    // changes is that the log no longer claims a removal that did not happen.
+    let log = if removed {
+        format!("webhooks.unregister_echo removed tunnel {tunnel_uuid}")
+    } else {
+        format!(
+            "webhooks.unregister_echo: no registration for tunnel {tunnel_uuid}, nothing removed"
+        )
+    };
     Ok(RpcOutcome::single_log(
         WebhookDebugRegistrationsResult { registrations },
-        format!("webhooks.unregister_echo removed tunnel {tunnel_uuid}"),
+        log,
     ))
 }
 
--- a/src/openhuman/skills/webhooks/router.rs
+++ b/src/openhuman/skills/webhooks/router.rs
@@ -220,27 +220,38 @@ impl WebhookRouter {
     }
 
     /// Unregister a tunnel. Only the owning skill can unregister it.
-    pub fn unregister(&self, tunnel_uuid: &str, skill_id: &str) -> Result<(), String> {
+    /// Returns `Ok(true)` when a registration was actually removed and
+    /// `Ok(false)` when `tunnel_uuid` had none — the caller's only way to tell
+    /// a real removal from a no-op, since both are successful outcomes.
+    ///
+    /// The `registration_changed` push, the re-persist and the
+    /// `WebhookUnregistered` bus event are all gated on a real removal (#6091).
+    /// They used to run unconditionally, so unregistering a tunnel that was
+    /// never registered announced a state change that never happened. This is
+    /// the same gate [`Self::unregister_skill`] already applies with its
+    /// `if !removed_tunnels.is_empty()`.
+    pub fn unregister(&self, tunnel_uuid: &str, skill_id: &str) -> Result<bool, String> {
         let mut routes = self.routes.write().map_err(|e| e.to_string())?;
 
-        if let Some(existing) = routes.get(tunnel_uuid) {
-            if existing.skill_id != skill_id {
-                return Err(format!(
-                    "Tunnel {} is owned by skill '{}'; skill '{}' cannot unregister it",
-                    tunnel_uuid, existing.skill_id, skill_id
-                ));
-            }
-            debug!(
-                "[webhooks] Unregistering tunnel {} (skill '{}')",
-                tunnel_uuid, skill_id
-            );
-            routes.remove(tunnel_uuid);
-        } else {
+        let Some(existing) = routes.get(tunnel_uuid) else {
             debug!(
                 "[webhooks] Tunnel {} not found for unregister (skill '{}')",
                 tunnel_uuid, skill_id
             );
+            return Ok(false);
+        };
+
+        if existing.skill_id != skill_id {
+            return Err(format!(
+                "Tunnel {} is owned by skill '{}'; skill '{}' cannot unregister it",
+                tunnel_uuid, existing.skill_id, skill_id
+            ));
         }
+        debug!(
+            "[webhooks] Unregistering tunnel {} (skill '{}')",
+            tunnel_uuid, skill_id
+        );
+        routes.remove(tunnel_uuid);
 
         drop(routes);
         self.publish_event("registration_changed", None, Some(tunnel_uuid.to_string()));
@@ -251,7 +262,7 @@ impl WebhookRouter {
             skill_id: skill_id.to_string(),
         });
 
-        Ok(())
+        Ok(true)
     }
 
     /// Remove all tunnel registrations for a skill (called on skill stop/crash).
@@ -455,8 +466,12 @@ impl WebhookRouter {
     }
 
     /// List recent webhook logs, newest first.
+    /// `limit` is a **maximum**, so an explicit `Some(0)` returns nothing. The
+    /// absent case is covered by `unwrap_or(100)`; there is deliberately no
+    /// `.max(1)` clamp, which could only ever have rewritten a caller's `0`
+    /// into a `1` (#6090). `Iterator::take(0)` is well-defined.
     pub fn list_logs(&self, limit: Option<usize>) -> Vec<WebhookDebugLogEntry> {
-        let limit = limit.unwrap_or(100).max(1);
+        let limit = limit.unwrap_or(100);
         self.debug_logs
             .read()
             .map(|logs| logs.iter().take(limit).cloned().collect())
--- a/src/openhuman/skills/webhooks/router_tests.rs
+++ b/src/openhuman/skills/webhooks/router_tests.rs
@@ -291,8 +291,116 @@ fn persist_and_load_roundtrip() {
 #[test]
 fn unregister_nonexistent_tunnel_is_noop() {
     let router = WebhookRouter::new(None);
-    // Should not error even though tunnel doesn't exist
-    router.unregister("no-such", "any-skill").unwrap();
+    // Should not error even though tunnel doesn't exist, and must report that
+    // it removed nothing rather than being indistinguishable from a real
+    // removal (#6091).
+    assert!(!router.unregister("no-such", "any-skill").unwrap());
+}
+
+/// #6091 — unregistering a tunnel that was never registered must not announce a
+/// state change.
+///
+/// `publish_event`, `persist` and the `WebhookUnregistered` bus publish used to
+/// run unconditionally after the not-found branch, so a caller with a typo'd
+/// UUID made every subscriber believe a registration had been torn down. The
+/// `registration_changed` debug event is the observable one from a sync test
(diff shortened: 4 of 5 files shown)


M3gA-Mind:
Heads-up on an overlap we caused, not the other way round: **#6114 also closes #6090 and was opened
ten minutes before this PR.** We did not see it before pushing.

The two are not equivalent in scope:

- **#6114** — `+36/-3`, fixes `#6090` (`limit: 0`) alone.
- **this PR** — `+210/-21`, fixes `#6090` **and** `#6091` (`webhooks_unregister_echo` reporting
  success for a tunnel that was never registered, plus the spurious `WebhookUnregistered` bus
  event), and treats them as one defect because both are the same unguarded side-effect tail in the
  router.

We have no preference and no claim here. If #6114 is the better fix for #6090 — it is smaller and
it was first — take it, and we will happily rebase this down to the #6091 half alone. Say the word
and we will push that within the hour.

Flagging rather than quietly leaving two PRs pointed at one issue for a reviewer to untangle.

PR #6122, opened 2026-09-08
fix(test-support): reject `..` lexically, and compile+run the gate in CI

### LINKED ISSUE #6085: `test_support.read_workspace_file`'s traversal guard fails open for a path that does not exist
## Summary

`test_support.read_workspace_file`'s traversal guard fails open for a path that does not exist

- **Namespace / method:** `openhuman.test_support_read_workspace_file`,
`openhuman.test_support_list_workspace_files`
- **Where:** `src/openhuman/test_support/introspect.rs:30-54`
(`resolve_workspace_relative`)

## Problem

**Expected**

what the doc comment on that function promises —
*"Reject any relative path containing a `..` component or that resolves
outside the workspace root."*

**Actual**

it rejects nothing on the strength of `..`. The guard is:
```rust
let candidate = canonical_root.join(trimmed);
let canonical_candidate = candidate.canonicalize().unwrap_or_else(|_| candidate.clone());
if !canonical_candidate.starts_with(&canonical_root) { return Err(...); }
```
`Path::join` does **not** normalise `..`, and `Path::starts_with` is
component-wise — so `<root>/../../etc/passwd` *does* start with `<root>` as a
component sequence. The check therefore depends entirely on
`canonicalize()` succeeding, and `canonicalize()` fails for any path that
does not exist, at which point the `unwrap_or_else` falls back to the
un-normalised candidate and the comparison passes.

**Steps to reproduce**

`{"rel_path": "../../nonexistent-e2e-probe"}` returns
`stat …: No such file or directory` rather than the documented
`escapes workspace root`. The two errors are not interchangeable: the first
says the guard did not fire.

**Cited location still accurate:** yes — `resolve_workspace_relative` is
`src/openhuman/test_support/introspect.rs:31-55` (doc comment at :29-30). The three lines the
report quotes (`join`, `canonicalize().unwrap_or_else`, `starts_with`) are verbatim.

**Repro run:** compiled the function **verbatim** into a standalone probe and exercised it
directly, rather than going through the RPC (the surface is `#[cfg(feature = "e2e-test-support")]`
and unreachable under the product feature string — see the sibling report):
`rustc -O guard_probe.rs -o guard_probe && ./guard_probe`

**Observed**

```
[1] Path::join does not normalise ..  -> /private/var/.../e2e-guard-probe-root/../../etc/passwd
[1] starts_with(root) = true   (report claims true)
[2] GUARD DID NOT FIRE, returned: /private/var/.../e2e-guard-probe-root/../../nonexistent-e2e-probe
[3] guard fired: rel_path "../../../../../../etc/passwd" escapes workspace root /var/.../e2e-guard-probe-root
[4] escaped-root check: returned path starts_with(root) = true but lexically leaves it
[4] components: [Normal("nonexistent-e2e-probe"), ParentDir, ParentDir, Normal("e2e-guard-probe-root")]
```

**Diagnosis assessment:** **right, and the probe separates the two halves cleanly.**
`[1]` proves the mechanism: `Path::starts_with` is component-wise, so a `..`-traversing path
*does* start with the root and the prefix check is worthless on its own. `[2]` vs `[3]` proves
the consequence is conditional on existence exactly as the report says — a **non-existent**
traversal target sails through (`canonicalize()` errors, `unwrap_or_else` restores the
un-normalised candidate, comparison passes), while an **existing** one is caught because
`canonicalize()` succeeded and resolved it to `/etc/passwd`. So the guard's entire protective
value comes from `canonicalize()`, and the `starts_with` line contributes nothing.

**Is the expected behaviour correct:** yes, and the authority is the function's own doc comment
at :29-30 — *"Reject any relative path containing a `..` component or that resolves outside the
workspace root."* The first clause is simply not implemented. The report's fix (reject
`Component::ParentDir` before joining) is what the doc already promises and needs no filesystem
access, which also removes the existence-dependence.

## Solution

iterate `Path::new(trimmed).components()` and reject any
`Component::ParentDir` (and `Component::RootDir`/`Prefix`) before the join; keep the
canonicalize+`starts_with` check afterwards as defence against symlink escapes, which the
component check cannot see.

**Blast radius:** two callers, both in the same file — `read_workspace_file` (`fs::metadata`
then read, capped at `READ_FILE_MAX_BYTES`) and `list_workspace_files` (`fs::read_dir`, depth
`LIST_MAX_DEPTH`). Both stat immediately, which is why the escape is currently absorbed as
`No such file or directory` rather than a leak. No test pins the current error text. The whole
surface is behind `e2e-test-support`, which no lane compiles — so this cannot be caught by CI
today in either state.

## Acceptance criteria

## Related

### LINKED ISSUE #6086: `test_support.*` (5 controllers) can never be reached by any product-feature build or test lane
## Summary

`test_support.*` (5 controllers) can never be reached by any product-feature build or test lane

- **Namespace / method:** `openhuman.test_support_workspace_root`,
`_list_workspace_files`, `_read_workspace_file`, `_in_flight_chats`,
`_wallet_prepared_quotes` (and `openhuman.test_reset`, same gate)
- **Where:** `src/core/all.rs:874-880` — the registration `push` is behind
`#[cfg(feature = "e2e-test-support")]`; the feature is declared at
`Cargo.toml:1365` as `e2e-test-support = []`

## Problem

**Expected**

either the coverage gate excludes controllers it is impossible
to reach, or a lane exists that compiles them.

**Actual**

`e2e-test-support` is in **neither** feature set —
not `[features] default` and not `scripts/ci/product-features.txt`. Its only
consumer is `app/scripts/e2e-build.sh` (three `pnpm tauri build --features
e2e-test-support` lines), i.e. the packaged desktop E2E binary. So:
- `check-domain-e2e-coverage.mjs` counts all five as uncovered controllers,
  because it scans `ControllerSchema` literals in source and cannot see `cfg`;
- no Rust test lane can invoke them, because every lane passes the product
  feature string;
- a `tests/*.rs` case that wants the positive path must add a feature outside
  the shared string, which the wave brief (rightly) forbids.
This is the same class of blind spot as the `tui` gate, which CLAUDE.md already
flags: *"a gate in neither set gets no compile coverage from the normal lanes
at all"*. `tui` at least has the feature-gate-smoke lane; `e2e-test-support`
has nothing — it is never `cargo check`ed by CI in either state.

**Steps to reproduce**

```
grep -n 'e2e-test-support' Cargo.toml scripts/ci/product-features.txt   # only Cargo.toml:1365
grep -rn 'e2e-test-support' .github/workflows/ scripts/ci/             # → no matches
```

**Cited location still accurate:** yes on both. `src/core/all.rs:875` is
`#[cfg(feature = "e2e-test-support")]` immediately above the `push` of
`all_test_support_registered_controllers()` (the block runs :871-881, so the report's :874-880 is
within a line). `Cargo.toml:1365` is `e2e-test-support = []`.

**Repro run:** ```
grep -n 'e2e-test-support' Cargo.toml scripts/ci/product-features.txt
grep -rn 'e2e-test-support' .github/workflows/ scripts/ci/
node scripts/check-domain-e2e-coverage.mjs | grep '^| test_support'
sed -n '466,530p' .github/workflows/ci-lite.yml
```

**Observed**

```
Cargo.toml:1365:e2e-test-support = []                      # and NOT in product-features.txt
scripts/ci/assert-coverage-presence.sh:48:#   src/openhuman/test_support/  `e2e-test-support`; …   (a COMMENT, not an invocation)
default = ["media","skills","flows","mcp","channels","medulla","http-server",
           "scheduler-gate","file-logging","modules","memory-engine-seams"]   # 11 gates, not one of them this
| test_support | test_support | 0/5 | 0.0% | openhuman.test_support_in_flight_chats<br>… |
```

**Diagnosis assessment:** **right, and sharper than the report states.** Two corrections, one
of which strengthens it:
1. *Minor*: the report's second grep claims "→ no matches" under `scripts/ci/`. There is one —
  `assert-coverage-presence.sh:48` — but it is a comment naming the gate, not a lane that
  compiles it. The substance is unaffected.
2. *Material*: the report says `e2e-test-support` "has nothing", contrasting it with `tui`'s
  smoke lane. In fact the feature-gate-smoke lane has a step named, verbatim,
  **"Check the gates that are in NEITHER the contributor nor the product set"**
  (`.github/workflows/ci-lite.yml:472-59`) — and it runs
  `cargo check --no-default-features --features tui`, i.e. that step *is* the home for this gate
  and `e2e-test-support` was never added to it. `CLAUDE.md:706` makes it an instruction:
  *"A gate in **neither** set (today only `tui`) gets no compile coverage from the normal lanes
  at all, so the feature-gate-smoke lane checks it explicitly. **Put new ones there too.**"*
  So this is not a missing lane — it is a documented rule with a second, unlisted member, and
  `CLAUDE.md`'s "today only `tui`" is itself now inaccurate.

**Is the expected behaviour correct:** yes for the compile-coverage half — CLAUDE.md:706
prescribes it, so that is not a judgement call. The coverage-gate half is genuinely open: the
gate scans `ControllerSchema` literals in source and cannot see `cfg`, so excluding cfg-gated
namespaces would change the denominator (627) and the headline percentage across the board.

## Solution

add `e2e-test-support` to the existing "gates in NEITHER set"
step in `rust-feature-gate-smoke` and update `CLAUDE.md:706`'s parenthetical, which is one line
each. Whether `check-domain-e2e-coverage.mjs` should also skip cfg-gated namespaces is a
separate call — it inflates the "uncovered" count by 5 here and would need a rule for reading
`cfg` out of `all.rs`, which the script deliberately does not parse today.

**Blast radius:** the gate is forwarded by `app/src-tauri/Cargo.toml:295` and turned on only by
the three `pnpm tauri build --features e2e-test-support` lines in `app/scripts/e2e-build.sh`;
`app/src-tauri/src/lib.rs` has ~6 `#[cfg(any(test, feature = "e2e-test-support"))]` sites.
Adding it to the smoke lane is compile-only and cannot affect what ships. Note the six affected
controllers include `openhuman.test_reset`, a **destructive** wipe — which is the reason the
gate exists and the reason a compile-only smoke check (not enabling it in a product lane) is the
right shape.

## Acceptance criteria

## Related

### PR DESCRIPTION
**Pull request by M3gA-Mind**
## Summary

- `test_support.read_workspace_file` / `list_workspace_files` now reject a `..` component **lexically, before touching the filesystem**, which is what the function's doc comment has always promised. Previously the guard only held when the traversal target happened to exist (#6085).
- Adds `e2e-test-support` to the feature-gate-smoke lane's "gates in NEITHER set" step, the one place CI compiles gates that are in neither `default` nor `product-features.txt` (#6086).
- Adds a `test_support::` run to the gate-contract test step, so this family is not just compiled but executed — without it the #6085 tests would exist and never run.
- Corrects `CLAUDE.md`'s "today only `tui`" note, which has been wrong for `e2e-test-support`'s whole life.

## Problem

**#6085 — the traversal guard failed open for any path that does not exist.**

The guard was:

```rust
let candidate = canonical_root.join(trimmed);
let canonical_candidate = candidate.canonicalize().unwrap_or_else(|_| candidate.clone());
if !canonical_candidate.starts_with(&canonical_root) { return Err(...); }
```

`Path::join` does not normalise `..` and `Path::starts_with` compares whole components, so `<root>/../../etc/passwd` *does* start with `<root>`. The prefix check therefore contributes nothing on its own — the only thing that ever caught a traversal was `canonicalize()` resolving it out of the root, and `canonicalize()` fails on a missing path, at which point `unwrap_or_else` restores the un-normalised candidate and the comparison passes.

`{"rel_path": "../../nonexistent-e2e-probe"}` was answered with the caller's own `stat` error ("No such file or directory") rather than a refusal. The two are not interchangeable: the first says the guard did not fire.

**#6086 — nothing in CI ever compiled this family, in either state.**

`e2e-test-support` is in neither `[features] default` nor `scripts/ci/product-features.txt`. `CLAUDE.md` already has a rule for exactly that position — *"A gate in **neither** set (today only `tui`) gets no compile coverage from the normal lanes at all, so the feature-gate-smoke lane checks it explicitly. Put new ones there too."* — and `e2e-test-support` was never added. So the `test_support` family has never been `cargo check`ed by CI, let alone tested.

The two issues are one PR because they meet in `ci-lite.yml`: #6085's regression tests are worthless without #6086's lane to run them in. A `cargo check` could not have caught #6085 — the broken guard compiled perfectly.

## Solution

**#6085** — walk `Path::new(trimmed).components()` before the join and reject `ParentDir` outright (plus `RootDir`/`Prefix`, which `trim_start_matches('/')` does not remove on Windows and which would make `join` discard the root entirely). `CurDir` is allowed through so existing `./x` callers keep working.

The canonicalize + `starts_with` pair is **kept underneath**: a component walk cannot see a symlink that points out of the workspace, and that check can. There is a test for exactly that case.

**#6086** — one more `cargo check` line in the existing "gates in NEITHER set" step, and one more `cargo test` line in the gate-contract step. The two `cargo check`s are deliberately **separate invocations**: `--features tui,e2e-test-support` would let one gate compile only because the other pulled a dependency in, which is the false pass that lane exists to prevent.

**The new test lane found a pre-existing failure on its first run, and this PR does not hide it.** With the filter at `test_support::`, `test_support::rpc::wipe_memory_tree_removes_content_dirs_and_reports_summary` fails under this profile:

```
test_reset: memory_tree wipe failed: wipe_all: driver 'tinymemory' does not serve Maintenance
```

It fails identically with and without this PR's changes, so it is not a regression here. The cause is `memory-engine-seams` (`= ["dep:tinymemory-core"]`), which is in `[features] default` but **not** in `product-features.txt`, so it is off in this profile. Adding `memory-engine-seams` to the line would make it green while hiding precisely the class of gap this lane exists to expose, so the filter is scoped to `test_support::introspect::` and the hole is named in a comment at the call site instead. Re-widen once it is resolved.

Deliberately **not** done: enabling `e2e-test-support` in any product lane. The family includes `openhuman.test_reset`, a destructive in-place wipe — a compile-only smoke check is the right shape.

Also deliberately not done: teaching `check-domain-e2e-coverage.mjs` to skip cfg-gated namespaces. That changes the denominator and the headline percentage across the board, and needs a rule for reading `cfg` out of `all.rs` that the script does not parse today. #6086 names it as a separate call.

## Impact

- No runtime or shipped-binary change. `test_support` is compiled only by `app/scripts/e2e-build.sh`; the guard change tightens a path that was already meant to be closed.
- Behaviour change visible to E2E specs: `rel_path` containing `..` now returns `escapes workspace root` instead of a `stat` error. No test pinned the old text.
- CI: adds one `cargo check` and one `cargo test --lib` invocation to `rust-feature-gate-smoke`.

## Related

- Closes tinyhumansai/openhuman#6085
- Closes tinyhumansai/openhuman#6086
- Follow-up PR(s)/TODOs: whether `check-domain-e2e-coverage.mjs` should exclude cfg-gated namespaces from its denominator is still open (#6086).

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — 8 new tests in `src/openhuman/test_support/introspect_tests.rs`: 4 refusal paths, 3 must-keep-working paths, 1 symlink escape the lexical walk cannot see.
- [x] **Diff coverage ≥ 80%** — ran `cargo test --lib --features "$(bash scripts/ci/product-features.sh),e2e-test-support" -- test_support::introspect` (8/8) and the new lane's own command `--no-default-features --features e2e-test-support --lib -- test_support::introspect::`. Not measured with `diff-cover` locally; the two arms of `resolve_workspace_relative`'s `RootDir`/`Prefix` branch are unreachable on POSIX and will not be covered by this platform's run.
- [x] N/A: no feature rows added, removed or renamed — `docs/TEST-COVERAGE-MATRIX.md` untouched.
- [x] N/A: no matrix feature IDs affected, per the item above.
- [x] No new external network dependencies introduced — the tests use `tempfile` (already a dependency) and touch only temp dirs.
- [x] N/A: does not touch a release-cut surface — `test_support` is compiled only by `app/scripts/e2e-build.sh`, never by a shipped build.

## Verification

```
# fix applied — the new lane's own command
running 15 tests ... 14 passed; 1 failed
  (the 1 is the pre-existing test_support::rpc::wipe_memory_tree_... described above,
   which fails identically without this PR's changes)

# revert-check: lexical component walk removed, everything else kept
  parent_dir_traversal_to_a_missing_path_is_rejected        FAILED  <- the issue's exact repro
  a_single_parent_dir_component_anywhere_is_rejected        FAILED
  a_leading_slash_is_trimmed_but_a_traversal_behind_it...   FAILED
  parent_dir_traversal_to_an_existing_path_is_rejected      ok      <- canonicalize still catches it
  a_symlink_pointing_out_of_the_workspace_is_still_rejected ok      <- the check the walk cannot make
  an_ordinary_relative_path_still_resolves_under_the_root   ok
  a_path_that_does_not_exist_yet_still_resolves             ok
  a_leading_current_dir_component_is_allowed                ok

  "a `..` traversal must be refused even when the target does not exist:
   /private/var/.../.tmplohp33/../../nonexistent-e2e-probe"
```

Each revert failure names its own assertion, and the three cases that must not depend on the lexical walk stayed green. Source restored byte-identical afterwards.

---

### Linear Issue

- Key: N/A — filed from the e2e coverage wave on GitHub, no Linear issue.
- URL: N/A

### Commit & Branch

- Branch: `fix/6085-test-support-traversal-guard`
- Commit SHA: `91d9d2719`

### Validation Run

- [x] Focused tests: `cargo test --lib --features "<product>,e2e-test-support" -- test_support::introspect` (8/8) and `--no-default-features --features e2e-test-support --lib -- test_support::introspect::`.
- [x] Rust fmt/check (if changed): `cargo check --manifest-path Cargo.toml --no-default-features --features e2e-test-support` → exit 0, which is the new smoke line itself.
- [x] N/A: Tauri fmt/check — `app/src-tauri` untouched.

### Behavior Changes

- Intended behavior change: `rel_path` containing a `..` component is refused lexically, before any filesystem access.
- User-visible effect: none in a shipped build. E2E specs calling `test_support.read_workspace_file` / `list_workspace_files` now get `escapes workspace root` where they previously got a `stat` error. No test pinned the old text.

### Parity Contract

- Legacy behavior preserved: the canonicalize + `starts_with` check is unchanged and still runs, so symlink escapes are caught exactly as before; `./x` and ordinary relative paths still resolve; resolution still does not require the target to exist.
- Guard/fallback/dispatch parity checks: covered by the three must-keep-working tests and the symlink test, all of which stay green with the fix reverted.

- Duplicate PR(s): none — timeline checked on both issues immediately before starting.
- Canonical PR: this one.
- Resolution: N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace-relative path validation to reject directory traversal and other paths that escape the workspace, including paths that do not yet exist.
  * Preserved support for valid relative paths, including standard prefixes and workspace-contained symlinks.

* **Tests**
  * Added coverage for path traversal, nonexistent targets, symlinks, and valid path resolution.
  * Expanded feature-gate checks for terminal interface and end-to-end test support configurations.

* **Documentation**
  * Clarified which feature gates receive dedicated compile-time coverage.


M3gA-Mind:
I was assigned #6086 (`test_support.*` unreachable by any product-feature build or lane) and stopped on pre-flight when this PR showed up cross-referenced. Having read the diff: **this already implements #6086's decided solution in full**, so I'm not opening anything against it.

For the record, checked against #6086's `## Solution` rather than taking the summary on trust:

- *"add `e2e-test-support` to the existing 'gates in NEITHER set' step in `rust-feature-gate-smoke`"* → `ci-lite.yml` now runs `cargo check --manifest-path Cargo.toml --no-default-features --features e2e-test-support`, and in a **separate** invocation from `tui`, which is the right call — a combined `--features tui,e2e-test-support` would not prove either gate compiles alone.
- *"update `CLAUDE.md:706`'s parenthetical"* → done via `AGENTS.md` (`CLAUDE.md` is a symlink to it), and the replacement names both members instead of asserting there is one, which is what stops the next gate landing in the same blind spot.
- Beyond the ask: the `test_support::` run in the gate-contract step means the family is **executed**, not merely compiled. #6086 only asked for compile coverage. Worth keeping — without it the #6085 tests would exist and never run, which you note in the summary.

**One suggestion:** the body carries no closing keyword for #6086, so merging this will fix it without closing it. Consider adding `Closes #6086` alongside #6085 — otherwise it stays open with its fix already shipped, and someone re-picks it. (I've left #6086 alone rather than edit your PR.)

No other findings; the two-invocation split and the "gates in NEITHER set" comment are both clearer than what they replace.

M3gA-Mind:
I wrote the independent verification blocks for both #6085 and #6086 (the standalone probe that showed the guard failing open), so I had the fix shape and blast radius already traced. **This matches both, and the tests are better than the ones I would have asked for.** Not approving — just recording what I checked.

**#6085.** The fix is the shape my verification proposed: reject `..` lexically via `components()` *before* the join, and keep the `canonicalize()` + `starts_with` pair underneath for symlinks. Two things I'd have raised are already handled:

- `RootDir` / `Prefix` are rejected too. Without that, `trim_start_matches('/')` leaves a Windows `C:\…` or `\\?\…` intact and `join` silently discards the workspace root — my verification flagged this and the diff covers it.
- `a_symlink_pointing_out_of_the_workspace_is_still_rejected` is the test that keeps the retained `canonicalize` check from later looking like dead code. That was my one open concern reading the production diff, and the test file answers it directly, with the reason in a comment.

Test coverage walk: `parent_dir_traversal_to_a_missing_path_is_rejected` is the distinguishing case (it is the exact input from my repro) and fails without the fix; `…to_an_existing_path…` is the case that already passed, kept so the fix cannot trade one for the other — the comment says so, which is the honest framing. `a/../b` covers the "resolves back inside but still contains `..`" reading of the doc contract, which is the stricter and correct one. Neither of the two callers (`read_workspace_file` → `fs::metadata`, `list_workspace_files` → `fs::read_dir`) needs a change, and no test in the tree pinned the old `stat` error text.

**#6086.** Adding `e2e-test-support` to the "gates in NEITHER set" step is exactly what that step exists for — and running it as a *separate* `cargo check` from `tui` rather than `--features tui,e2e-test-support` is the right call for the reason given. Worth noting the lane's own step name already described this gate's situation before it was added to it.

**One thing you may not know, so nobody duplicates the edit:** `CLAUDE.md` is a **symlink to `AGENTS.md`** in this repo (`CLAUDE.md -> AGENTS.md`, verified by `md5`, both resolve to line 706). So the one-line doc change lands in both and there is no second copy to update.

## One suggestion, take it or leave it

The scoped `test_support::introspect::` filter reintroduces, in miniature, the shape this lane exists to prevent: a future `test_support::<something_new>` submodule gets compile coverage from the smoke check but never runs, silently. You have named it and left a re-widen note, which is far better than hiding it — but there may be a cheaper way to widen now.

The failure you hit is `wipe_all: driver 'tinymemory' does not serve Maintenance`, i.e. `memory-engine-seams` (`dep:tinymemory-core`) is in `[features] default` but not in `product-features.txt`, so `--no-default-features --features e2e-test-support` has it off. That is a property of the *profile*, not of the test — so `#[cfg(feature = "memory-engine-seams")]` on `wipe_memory_tree_removes_content_dirs_and_reports_summary` would let it run wherever it can and skip where it structurally cannot, and the filter could go back to `test_support::` today. `#[ignore]` would be worse — it would also silence the test in profiles where it should run.

Either way I would not hold the PR for it.

PR #6139, opened 2026-09-08
test(security): follow #6096's StorageMode split in the keyring e2e

### LINKED ISSUE #6076: `keyring_consent_status` reports `active_mode: os_keyring` for the file backend
## Summary

`keyring_consent_status` reports `active_mode: os_keyring` for the file backend

- **Namespace / method:** openhuman.keyring_consent_status, openhuman.keyring_consent_retry_probe
- **Where:** `src/openhuman/security/keyring_consent/policy.rs:85-108`, with
`src/openhuman/security/keyring/ops.rs:131-146`

## Problem

**Expected**

`active_mode` describes *where secrets actually are*. The four
variants are `os_keyring` / `local_encrypted` / `consent_pending` / `declined`.

**Actual**

`current_status()` branches on `keyring::is_available()` alone:

```rust
let (active_mode, failure_reason) = if available {
    (StorageMode::OsKeyring, None)
} else { … };
```

and `probe_availability()` returns `true` unconditionally for the `file`,
`mock` and `encrypted_file` backends:

```rust
if b.name() == "file" || b.name() == "mock" || b.name() == "encrypted_file" {
    return true;
}
```

So a process running with `OPENHUMAN_KEYRING_BACKEND=file` (or the
`encrypted_file` backend the README lists for **staging and production**) is
told its secrets live in the OS keyring. `backend_name` still says `"file"`, so
the payload contradicts itself.

**Steps to reproduce**

```bash
OPENHUMAN_KEYRING_BACKEND=file cargo test --test raw_coverage_all \
    --features "$(bash scripts/ci/product-features.sh)" -- secrets_devices_e2e::keyring_consent
```
`keyring_consent_status` returns `{"available": true, "activeMode": "os_keyring",
"backendName": "file"}`, and still does after
`keyring_consent_decide {"mode": "declined"}`.

**Cited location still accurate:** yes, both.
`keyring_consent/policy.rs:85` is `pub fn current_status()`, and the
`let (active_mode, failure_reason) = if available { (StorageMode::OsKeyring, None) }`
branch is at `:89-90`, inside the cited `85-108` span.
`keyring/ops.rs:131` is `fn probe_availability()`, and the unconditional early return
`if b.name() == "file" || b.name() == "mock" || b.name() == "encrypted_file" { return true; }`
is at `:142-146`. Both quotes in the report are verbatim.

**Repro run:** the report's repro drives W4's own suite, which is not in this worktree, so
I dispatched the three controllers directly over the real JSON-RPC router with
`OPENHUMAN_KEYRING_BACKEND=file`:
```
cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" \
    -- --test-threads=1 --nocapture w3_verify_scratch
```

**Observed**

```
[bug2] status BEFORE decide = {"activeMode":"os_keyring","available":true,"backendName":"file"}
[bug2] decide(declined)     = {"consentedAtMs":1788789690194,"storageMode":"declined"}
[bug2] status AFTER  decide = {"activeMode":"os_keyring","available":true,"backendName":"file"}
[bug2] retry_probe          = {"activeMode":"os_keyring","available":true,"backendName":"file"}
```
Both halves of the report reproduce exactly. The payload contradicts itself in a single
object (`activeMode: os_keyring` beside `backendName: file`), and a **successfully
recorded** `declined` decision — the `decide` call returns `storageMode: "declined"` — has
no effect whatsoever on `activeMode`. `retry_probe` does not shift it either.

**Diagnosis assessment:** Correct, and the report **understates the reach**. I checked the
claim that `encrypted_file` is the staging/production backend, since that is what separates
"test-harness artifact" from "ships this way". It holds: `security/keyring/README.md:61`
documents `encrypted_file` as "**(staging/production**, and via
`OPENHUMAN_KEYRING_BACKEND=encrypted_file`)", and `:88` gives the selection order ending
"staging/prod → `encrypted_file`". So **in staging and production the shipped app tells the
user their secrets are in the OS keyring while they are in `{workspace}/secrets.enc`.**
(My first search missed this — I grepped the root `README.md`/`docs/`/`gitbooks/` and found
nothing. The claim is in the *module* README. Widening the search is what found it.)
One nuance in fairness to the code: for `encrypted_file` the OS keychain *is* still
involved — it holds the master key that encrypts the file (`README.md:12`). So
`os_keyring` is misleading rather than absurd there. For the plain `file` backend, which is
the dev default and what the repro above uses, there is no OS keychain involvement at all
and the label is simply false.
The "consent path unreachable" consequence is real and has a UI consumer:
`app/src/components/settings/panels/SecurityPanel.tsx:108,118` gate two controls on
`activeMode !== 'local_encrypted'` and `activeMode !== 'declined'`, and `:25,62` drive the
badge and its i18n key off `activeMode`. So the mislabel is user-visible, not internal.

**Is the expected behaviour correct:** yes — "`active_mode` describes where secrets
actually are" is the only reading the four variant names support, and `available: true`
for the file backend is separately defensible (the backend *is* usable), so the bug is
isolated to `active_mode` exactly as the report says.

## Solution

do not let `active_mode` be derived from
`is_available()` alone — branch on the backend identity that `backend_name()` already
returns. The minimum correct change is that a non-OS backend must never report
`os_keyring`. **One sub-question needs a maintainer decision:** should `encrypted_file`
report `local_encrypted` — which conflates "the operator configured this backend" with
"the user declined the OS keyring", and would flip the two `SecurityPanel` buttons in
staging/prod — or does the enum need a fifth variant that distinguishes an
operator-configured backend from a consent outcome?

**Blast radius:** `current_status()` backs `keyring_consent_status` and
`keyring_consent_retry_probe`; `keyring::is_available()` is additionally consulted by
wallet guards and snapshot loops (`security/keyring/README.md:9`) — those read
`available`, not `active_mode`, so they are unaffected by an `active_mode`-only fix.
Frontend consumers: `SecurityPanel.tsx` (badge + two gated buttons),
`CoreStateProvider.tsx:241`, and three test files that hard-code
`activeMode: 'os_keyring'` and would need updating.
Existing tests: `keyring_consent/policy_tests.rs` pins current behaviour and must be
re-read before any change.

## Acceptance criteria

## Related

M3gA-Mind:
## Summary

- `main` is red on `Rust Core Coverage` and every open PR inherits it. One test fails.
- The test is wrong, not the code: it pinned the **pre-fix** vocabulary from #6076 and #6096 corrected the product.
- Three assertions updated against `StorageMode`'s `Display` impl, which is the source of truth.
- Test-only. No production code changed.

## Problem

```
secrets_devices_e2e::keyring_consent_status_decide_and_retry_probe
  activeMode must be one of the four documented storage modes; got "local_plaintext_file"
```

`keyring_consent_status_decide_and_retry_probe` was written beside #6076 — `activeMode` reporting
`os_keyring` for the file backend — and encoded the old four-variant vocabulary in three places.
#6096 fixed the defect by making `activeMode` name the backend actually in use, which added
`local_encrypted_file` and `local_plaintext_file` to `StorageMode`. The test's allow-list still had
four, so it failed on the corrected behaviour.

One of the three assertions encoded **the bug itself**:

| Line | Was | Why it is wrong now |
|---|---|---|
| ~344 | four-variant allow-list | `StorageMode` has six |
| ~369 | `assert_eq!(mode, "os_keyring")` when `available` | this *is* #6076 — it required every usable backend to claim it was the OS keyring |
| ~519 | four-variant list on the retry probe | same |

## Solution

All three rewritten against `src/openhuman/security/keyring_consent/types.rs` rather than a
remembered list. The allow-lists now enumerate the whole enum, so a future variant fails here
instead of silently widening what the RPC may return.

The `available` branch asserts `!matches!(mode, "consent_pending" | "declined")`. The schema
documents `available` as *"whether the active backend is usable, which is true for the file
backends"* — so a usable backend implies no consent decision is outstanding, not that it is the OS
keyring.

The doc comment describing the pre-fix behaviour as current is reworded to past tense with a pointer
to #6096.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) per [Testing Strategy](../gitbooks/developing/testing-strategy.md#failure-path-requirement) — three assertions updated in `secrets_devices_e2e.rs`; the suite covers both the available and unavailable branches. Ran `cargo test --test raw_coverage_all --features "$(bash scripts/ci/product-features.sh)" -- --test-threads=1 secrets_devices_e2e::` → **8 passed, 0 failed, 1 ignored**.
- [x] **Diff coverage ≥ 80%** — every changed line is a test assertion executed by the run above, so changed-line coverage is 100% by construction.
- [x] N/A: test-only change; no feature row added, removed or renamed — *coverage matrix updated*
- [x] N/A: no feature IDs affected; this corrects assertions on existing behaviour — *affected feature IDs listed under `## Related`*
- [x] No new external network dependencies introduced (mock backend used per [Testing Strategy](../gitbooks/developing/testing-strategy.md#mock-policy)) — no new dependency; the test uses the existing in-process harness.
- [x] N/A: test-only, no release-cut surface touched — *manual smoke checklist ([`docs/RELEASE-MANUAL-SMOKE.md`](../docs/RELEASE-MANUAL-SMOKE.md))*
- [x] N/A: no issue exists — this repairs a test that a merged fix invalidated, and filing one for a same-day break is noise — *linked issue closed via `Closes #NNN`*

## Impact

Unblocks `Rust Core Coverage` on `main` and on every open PR that inherits it. No runtime, platform,
performance, security or migration impact — the diff touches one test file.

## Related

- Closes: —
- Follow-up to #6096, which fixed #6076 and changed the vocabulary this test pinned.
- Unblocks #6117, #6120, #6122, #6124, #6132, which fail only on this.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyring consent validation to ensure the reported storage mode matches the active backend.
  - Added coverage for file-backed storage modes in status and retry responses.
  - Continued rejecting pending or declined consent states.

- **Documentation**
  - Clarified that `activeMode` identifies the backend currently in use.

### DIFF
--- a/tests/raw_coverage/secrets_devices_e2e.rs
+++ b/tests/raw_coverage/secrets_devices_e2e.rs
@@ -367,17 +367,28 @@ async fn keyring_consent_status_decide_and_retry_probe() {
         "the backend must be named so the settings panel can show it: {status}"
     );
     if available {
-        // NOT `== "os_keyring"`. That was the #6076 defect this test was written
-        // beside: every working backend reported itself as the OS keyring, so a
-        // plaintext dev-keychain looked like the OS credential store. #6096 made
-        // `activeMode` describe the backend that is actually in use, and the
-        // schema now documents `available` as "the active backend is usable,
-        // which is true for the file backends". What a usable backend implies is
-        // that no consent decision is outstanding — not which backend it is.
-        assert!(
-            !matches!(mode, "consent_pending" | "declined"),
-            "a usable backend means no consent decision is outstanding; \
-             got {mode:?}: {status}"
+        // Assert the BACKEND -> MODE pairing, not merely that the mode is in the
+        // enum. #6076 was exactly a mismatched pair — `backendName: "file"` with
+        // `activeMode: "os_keyring"` — so a predicate that only rejects
+        // `consent_pending`/`declined` would let that regression back in while a
+        // comment claimed to guard it. Mirrors `active_mode_for` in
+        // `security/keyring_consent/policy.rs:125-146`.
+        let backend = status
+            .get("backendName")
+            .and_then(Value::as_str)
+            .expect("backendName asserted non-empty above");
+        let expected = match backend {
+            "os" => "os_keyring",
+            "encrypted_file" => "local_encrypted_file",
+            "file" | "mock" => "local_plaintext_file",
+            // An unrecognised backend reports `consent_pending` rather than
+            // guessing — the deliberate fallback in `policy.rs`.
+            _ => "consent_pending",
+        };
+        assert_eq!(
+            mode, expected,
+            "a usable `{backend}` backend must report `{expected}`; reporting anything \
+             else is the #6076 mismatch this test exists to catch: {status}"
         );
         assert!(
             status.get("failureReason").is_none(),
--- a/tests/raw_coverage/secrets_devices_e2e.rs
+++ b/tests/raw_coverage/secrets_devices_e2e.rs
@@ -305,7 +305,8 @@ allowed_commands = ["ls", "cat"]
 /// All three keyring-consent controllers.
 ///
 /// The anchor is the **persisted file**, not `activeMode`. `policy::current_status`
-/// reports `os_keyring` whenever *any* backend probes as available — and the
+/// reported `os_keyring` whenever *any* backend probes as available (fixed by
+/// #6096; `activeMode` now names the backend actually in use) — and the
 /// `file` dev backend always does — so `activeMode` cannot witness a consent
 /// decision in a test process. `ops::keyring_consent_decide` documents that the
 /// in-memory cache is only updated *after* a successful persist, so proving the
@@ -341,9 +342,22 @@ async fn keyring_consent_status_decide_and_retry_probe() {
         .get("activeMode")
         .and_then(Value::as_str)
         .unwrap_or_else(|| panic!("status must report an `activeMode`: {status}"));
+    // Mirrors `StorageMode`'s `Display` impl (`security/keyring_consent/types.rs`),
+    // which is the source of truth. #6096 added the two `*_file` variants when it
+    // stopped the file backends reporting themselves as `os_keyring` (#6076) — this
+    // list is the whole enum, so a new variant fails here rather than silently
+    // widening what the RPC may return.
     assert!(
-        ["os_keyring", "local_encrypted", "consent_pending", "declined"].contains(&mode),
-        "activeMode must be one of the four documented storage modes; got {mode:?}"
+        [
+            "os_keyring",
+            "local_encrypted",
+            "local_encrypted_file",
+            "local_plaintext_file",
+            "consent_pending",
+            "declined",
+        ]
+        .contains(&mode),
+        "activeMode must be one of the six documented storage modes; got {mode:?}"
     );
     assert!(
         status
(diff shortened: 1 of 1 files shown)


M3gA-Mind:
Good catch, and you were right that the comment over-promised relative to the predicate.

Fixed in `f2cfb22eb`: the assertion now derives the expected mode from `backendName`, mirroring `active_mode_for` (`security/keyring_consent/policy.rs:125-146`) including its deliberate `consent_pending` fallback for an unrecognised backend. So the exact #6076 pair — `backendName: "file"` with `activeMode: "os_keyring"` — now fails.

Proven by fault injection rather than asserted: mapping `"file"` to `"os_keyring"` produces

```
a usable `file` backend must report `os_keyring` ...
  left: "local_plaintext_file"  right: "os_keyring"
```

Restored, suite green (8 passed, 0 failed, 1 ignored). The retry probe's own mode check keeps the full six-variant list, since that path does not re-assert a backend.

PR #6167, opened 2026-09-09
fix(ai-settings): send claude-code: for a Claude Code CLI route in the routing Test

### LINKED ISSUE #6125: Custom-routing Test sends ollama:<model> for a Claude Code CLI route
## Summary

In **Connections → LLM → Routing**, a workload routed to the Claude Code CLI sends `ollama:<model>` to the test RPC instead of `claude-code:<model>` when **Test** is clicked. The test never reaches the CLI, always fails, and the result banner reads `Provider: ollama:<model>` while the dialog above it reads `Claude Code CLI · <model>`.

Saving the route is unaffected — only **Test** is broken.

Reported by @simon256px.

## Problem

**Steps to reproduce** (labels below are the current UI, app 0.63.22)

1. Open **Connections**. In the left nav under the **API keys** group, choose **LLM**.
2. On the **LLM Providers** tab, click **Add a provider**, and under **CLI logins** choose **Claude Code**. (If it is already connected, a **Claude Code** row is shown under a **CLI logins** group instead.)
3. Switch to the **Routing** tab and select the **Advanced** routing-mode card.
4. In the **Chat and Conversations** card, click the provider button at the right of the **Chat** row. The **Custom routing** dialog opens.
5. Click the **Provider and model** button (its right edge reads **Change**).
6. In **Choose provider and model**, under **Providers** select **Claude Code** (detail line: *CLI provider*), enter `sonnet` in the **Model** field, and click **Use this model**.
7. Back in **Custom routing** the button now reads **Claude Code CLI · sonnet**. Click **Test**.

**Expected** — the Claude Code CLI is invoked and its reply appears under **Model response**.

**Actual** — a red **Test failed** banner. Its first line reads:

```
Provider: ollama:sonnet
```

Also reproduces with `claude-sonnet-5`, and on any workload row (the defect is in the shared dialog).

**Reproduced at runtime.** Rendering `CustomRoutingDialog` with `initial={{ kind: 'claude-code', model: 'sonnet' }}` and clicking **Test**:

| Observed | Value |
| --- | --- |
| Provider button label | `Claude Code CLI · sonnet` |
| `testProviderModel(...)` call | `('chat', 'ollama:sonnet', 'Hello world')` |
| Result-banner provider line | `Provider: ollama:sonnet` |

**Cause**

`app/src/components/settings/panels/ai/CustomRoutingDialog.tsx:157-168` builds the string with a ternary that only branches on `cloud`; every other non-`managed` source kind falls through to the `ollama:` arm:

```ts
const currentProviderString =
  source == null || source.kind === 'managed'
    ? null
    : source.kind === 'cloud'
      ? appendTemperatureToProviderString(`${source.providerSlug}:${model.trim()}`, ...)
      : appendTemperatureToProviderString(`ollama:${model.trim()}`, ...);   // ← claude-code lands here
```

`CustomDialogSource` has four kinds — `cloud`, `local`, `claude-code`, `managed` (`app/src/components/settings/panels/ai/aiPanelTypes.ts:97`) — so a `claude-code` source serialises as `ollama:<model>`.

That single value drives both symptoms:

- It is the `provider` argument to `testProviderModel(...)` → `openhuman.inference_test_provider_model` (`CustomRoutingDialog.tsx:204`). The Rust handler routes anything prefixed `ollama:` down the local-runtime branch (`src/openhuman/inference/ops.rs:148-160`), so the CLI is never spawned. The reporter independently confirmed this with event-driven Win32 WMI process monitoring: clicking **Test** spawns no `claude.exe` and no `cmd.exe`.
- It is rendered verbatim as the banner's provider line — `ModelTestResultPanel` interpolates `currentProviderString` into `settings.ai.providerWithValue` (`'Provider: {value}'`, `app/src/lib/i18n/en.ts:5148`).

The error message is *not* wrong, which is why the two disagree: `describeProviderVerificationFailure` is called with `registrySlug` (`CustomRoutingDialog.tsx:218`), and `registrySlug` correctly resolves `claude-code` at lines 101-109. So the message names `claude-code` while the call went to `ollama`.

**This dialog is the only place with the bug.** The other serialisers use an exhaustive `switch` on `kind` and emit `claude-code:` correctly:

- `serializeProviderRef` — `app/src/services/api/aiSettingsApi.ts:251-264` (the **Save** path)
- `selectionValue` — `app/src/components/chat/ModelQualityPill.tsx:35-47`

So **Save** persists the right route and chat works; only **Test** is broken. That also explains why the reporter's separate agent/MCP invocation of Claude Code CLI works normally.

The backend needs no change: `create_chat_model_from_string_with_model_id` dispatches `claude-code:` at `src/openhuman/inference/provider/factory_part_02.rs:192` → `try_create_claude_code_chat_model_from_string` (`factory_part_03.rs:148`).

**Environment** — reported on Windows; `claude` CLI v2.1.259, authenticated Pro account. The defect is in platform-agnostic TypeScript, so it is not OS-specific. Reporter ruled out local causes: `claude doctor` clean, `claude auth status` valid, direct `claude --model sonnet -p "…"` and `claude --model claude-sonnet-5 -p "…"` both succeed, single `claude` binary on PATH, and no `ANTHROPIC_BASE_URL` / `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` override in process env, registry, `~/.claude.json` or `~/.claude/settings.json`.

## Solution (optional)

Replace the ternary with an exhaustive `switch` on `source.kind`, so a new kind is a compile error rather than a silent `ollama:` fallback — mirroring `serializeProviderRef`:

```ts
const currentProviderString = ((): string | null => {
  if (source == null || source.kind === 'managed') return null;
  const temp = temperature == null || !Number.isFinite(temperature) ? null : temperature;
  switch (source.kind) {
    case 'cloud':
      return appendTemperatureToProviderString(`${source.providerSlug}:${model.trim()}`, temp);
    case 'claude-code':
      return `claude-code:${model.trim()}`;
    case 'local':
      return appendTemperatureToProviderString(`ollama:${model.trim()}`, temp);
  }
})();
```

Verified locally: with that patch the same render calls `testProviderModel('chat', 'claude-code:sonnet', 'Hello world')` and the banner reads `Provider: claude-code:sonnet`.

The `@<temp>` suffix is deliberately omitted for `claude-code` — the factory accepts it but ignores it and logs a warning on every call (`src/openhuman/inference/provider/factory_part_03.rs:158-164`).

## Acceptance criteria

- [ ] **Repro gone** — following the steps above invokes the CLI and returns a reply; the banner reads `Provider: claude-code:<model>`.
- [ ] **Exhaustiveness** — `currentProviderString` switches exhaustively on `CustomDialogSource['kind']`, so a future source kind fails to compile instead of silently serialising as `ollama:`.
- [ ] **Regression safety** — a unit test asserts `testProviderModel` is called with `claude-code:<model>` when the dialog's source is `claude-code`. There is no test covering this dialog's claude-code path today.

## Related

- `app/src/components/settings/panels/ai/CustomRoutingDialog.tsx:157-168`
- `app/src/services/api/aiSettingsApi.ts:251-264` and `app/src/components/chat/ModelQualityPill.tsx:35-47` (correct exhaustive serialisers)
- `src/openhuman/inference/ops.rs:148-160` (local-prefix routing in the test RPC)

YellowSnnowmann:
## Summary

- **Connections → LLM → Routing → Custom routing → Test** now reaches the Claude Code CLI. It previously sent `ollama:<model>` for a `claude-code` route, so the CLI was never spawned and the test could only fail.
- `currentProviderString` switches exhaustively on `CustomDialogSource['kind']` instead of branching only on `cloud` — a future source kind is now a compile error rather than another silent `ollama:` fallback.
- The `@<temp>` suffix is **appended** on the `claude-code` arm rather than dropped (this deviates from the issue's suggested snippet — rationale under *Solution*).
- Adds `CustomRoutingDialog.test.tsx`. This dialog had no test coverage at all.
- No backend change, no schema/RPC change, nothing persisted by **Test** — the **Save** path was already correct and is untouched.

## Problem

`app/src/components/settings/panels/ai/CustomRoutingDialog.tsx:157-168` built the provider string with a ternary that branched only on `cloud`; every other non-`managed` source kind fell through to the `ollama:` arm:

```ts
source == null || source.kind === 'managed'
  ? null
  : source.kind === 'cloud'
    ? appendTemperatureToProviderString(`${source.providerSlug}:${model.trim()}`, …)
    : appendTemperatureToProviderString(`ollama:${model.trim()}`, …);   // ← claude-code landed here
```

`CustomDialogSource` has four kinds — `cloud`, `local`, `claude-code`, `managed` — so a Claude Code route serialised as `ollama:<model>`. That one value drives both reported symptoms:

- It is the `provider` argument to `testProviderModel(...)` → `openhuman.inference_test_provider_model`. `src/openhuman/inference/ops.rs:149-153` treats an `ollama:` prefix as local and calls `create_local_chat_model_from_string`, so `try_create_claude_code_chat_model_from_string` was never reached and no `claude` process was spawned.
- It is rendered verbatim as the banner's provider line, so the banner read `Provider: ollama:sonnet` while the button above it read `Claude Code CLI · sonnet`.

The error *text* was correct, which is why the two disagreed: `describeProviderVerificationFailure` is called with `registrySlug`, which resolves `claude-code` properly.

**Save was never affected.** `serializeProviderRef` (`app/src/services/api/aiSettingsApi.ts:251-264`) already switches exhaustively and emits `claude-code:`, so stored routing and chat worked — only **Test** was broken.

## Solution

Replace the ternary with an exhaustive `switch`, mirroring `serializeProviderRef`.

**Why the temperature suffix is kept, against the issue's suggestion.** The issue proposed omitting `@<temp>` on the `claude-code` arm because the Rust factory ignores it and logs a warning. Omitting it would introduce a fresh Test↔Save divergence:

- `serializeProviderRef` **does** append it for `claude-code`, via `joinModelAndTemp` (`aiSettingsApi.ts:262`).
- `TemperatureOverrideField` is rendered **unconditionally** in this dialog (`CustomRoutingDialog.tsx:298`), outside any `source.kind` guard, and `handleSave` carries the value into the claude-code ref — so a Claude Code route really can be saved as `claude-code:sonnet@0.7`.
- Dropping it in Test would mean Test exercised `claude-code:sonnet` while the saved route ran `claude-code:sonnet@0.7` — the same defect class this PR fixes, one size smaller.
- The `log::warn!` fires on every chat-model construction from the persisted route anyway, so suppressing it in Test hides from the tester a warning their real route produces.

(`ModelQualityPill.selectionValue` omits the suffix, but its `ProviderModelSelection` type has no `temperature` field at all — absence of a field is not a decision to drop one, so it is not a precedent here.)

The `cloud` and `local` arms are byte-identical to before: `appendTemperatureToProviderString` and `joinModelAndTemp` share the same `Math.round(t * 100) / 100` rounding and the same `!Number.isFinite` guard.

`handleSave`, `registrySlug` and `serializeProviderRef` are untouched.

## Submission Checklist

- [x] Tests added or updated (happy path + at least one failure / edge case) — new `CustomRoutingDialog.test.tsx`: the claude-code regression, temperature parity against the real `serializeProviderRef`, cloud and local unchanged, and managed (Test disabled, no RPC sent).
- [x] **Diff coverage ≥ 80%** — **100%**. All 7 instrumented added lines are covered (`vitest related --coverage` over the changed files, cross-referenced against the diff's added line numbers in `app/coverage/lcov.info`).
- [x] Coverage matrix updated — `N/A: bug fix, no feature leaf added, removed or renamed.`
- [x] All affected feature IDs from the matrix are listed under `## Related` — `N/A: no matrix leaf covers this dialog.`
- [x] No new external network dependencies introduced — `N/A: no network code touched; the new test mocks the RPC client.`
- [x] Manual smoke checklist updated if this touches release-cut surfaces — `N/A: no release-cut surface changed.`

## Impact

- **Platform**: desktop, all OSes. The defect is in platform-agnostic TypeScript — it was reported on Windows but is not OS-specific.
- **Behaviour change**: only the provider string the **Test** button sends and the string the result banner renders. Nothing is persisted by Test, so no migration and no effect on existing stored routing.
- **Backwards compatibility**: none at risk. Cloud and local routes produce byte-identical strings; the two claude-code tests fail against the previous code while the cloud/local/managed tests pass in both directions, confirming nothing re-baselines.
- **Performance / security**: none.

## Related

- Closes: #6125
- Follow-up PR(s)/TODOs: the Rust factory accepts `@<temp>` for `claude-code` but ignores it and warns (`src/openhuman/inference/provider/factory_part_03.rs:157-164`). Wiring temperature through to the CLI — or refusing the override in the UI for this source — is a separate product decision, deliberately out of scope here.

---

### Linear Issue

- Key: `N/A` — tracked on GitHub as tinyhumansai/openhuman#6125
- URL: https://github.com/tinyhumansai/openhuman/issues/6125

### Commit & Branch

- Branch: `fix/6125-custom-routing-test-provider-string`
- Commit SHA: `88edd012d12c66f99779e28e7d6e6d11cd9ec9a6`

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — Prettier clean across `app/` (the Rust half, `pnpm rust:format:check`, is covered under *Validation Blocked*).
- [x] `pnpm typecheck` — `tsc --noEmit` clean.
- [x] Focused tests: `npm test -- --run src/components/settings/panels/ai src/components/settings/panels/__tests__ src/services/api/__tests__/aiSettingsApi.test.ts src/components/chat/ModelQualityPill` → **68 files, 990 passed, 1 skipped, 0 failed**. Also `pnpm build` (production UI build) exit 0, `eslint` exit 0, and both `lint:commands-tokens` / `lint:ui-tokens` exit 0.
- [x] Rust fmt/check (if changed): `N/A` — no Rust file is modified by this PR.
- [x] Tauri fmt/check (if changed): `N/A` — `app/src-tauri/` is untouched.

### Validation Blocked

- `command:` `pnpm rust:clippy` / `pnpm rust:format:check` (the pre-push hook's Rust steps; pushed with `--no-verify`)
- `error:` the `vendor/*` submodules are not initialised in this worktree, so both Cargo manifests fail to resolve their path dependencies.
- `impact:` none for this changeset — the diff contains no Rust. CI's Rust lanes cover it on the PR. Every non-Rust step of the pre-push hook (`prettier --check .`, `eslint`, `tsc --noEmit`, `lint:commands-tokens`, `lint:ui-tokens`) was run manually and passed.

### Behavior Changes

- Intended behavior change: a workload routed to the Claude Code CLI serialises as `claude-code:<model>[@<temp>]` for the Test RPC instead of `ollama:<model>[@<temp>]`.
- User-visible effect: **Test** on a Claude Code route now invokes the CLI and shows its reply under **Model response**, and the banner reads `Provider: claude-code:<model>` instead of `Provider: ollama:<model>`.

### Parity Contract

- Legacy behavior preserved: the `cloud` and `local` arms emit byte-identical strings (same rounding, same finite guard); `managed` still yields `null` and leaves **Test** disabled; the **Save** path is not touched.
- Guard/fallback/dispatch parity checks: the `switch` is exhaustive with **no `default:` arm** — verified by temporarily adding a fifth `CustomDialogSource` kind, which fails `tsc` with `error TS2366: Function lacks ending return statement…` at `CustomRoutingDialog.tsx:169`. The temperature-parity test asserts the Test string equals the **real** (unmocked) `serializeProviderRef` output for the same ref, so the Test and Save serialisers cannot drift apart silently.

- Duplicate PR(s): none — `gh search prs` for `6125` returned no open or merged PR.
- Canonical PR: this one.
- Resolution: `N/A`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AI provider testing for Claude Code routes to use the correct provider identifier.
  * Preserved correct provider prefixes for cloud and local routes.
  * Ensured temperature overrides are included consistently when testing and saving custom routes.
  * Kept testing unavailable for managed routes.

### DIFF
--- a/app/src/components/settings/panels/ai/CustomRoutingDialog.test.tsx
+++ b/app/src/components/settings/panels/ai/CustomRoutingDialog.test.tsx
@@ -1,0 +1,147 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import { serializeProviderRef, testProviderModel } from '../../../../services/api/aiSettingsApi';
+import type { CloudProvider, ProviderRef, Workload } from './aiPanelTypes';
+import { CustomRoutingDialog } from './CustomRoutingDialog';
+
+// `serializeProviderRef` is deliberately NOT mocked: the temperature-parity test
+// below asserts that Test serialises the same way Save does, and a mocked
+// serialiser would make that assertion a tautology.
+vi.mock('../../../../services/api/aiSettingsApi', async importOriginal => {
+  const actual = await importOriginal<typeof import('../../../../services/api/aiSettingsApi')>();
+  return {
+    serializeProviderRef: actual.serializeProviderRef,
+    testProviderModel: vi.fn(),
+    describeProviderVerificationFailure: vi.fn(() => 'test failed'),
+    modelRegistryVision: vi.fn(() => false),
+    // `ProviderModelPickerDialog` imports this at module scope even though this
+    // suite never opens the picker.
+    listProviderModels: vi.fn(),
+  };
+});
+
+const CHAT_WORKLOAD: Workload = {
+  id: 'chat',
+  group: 'chat',
+  labelKey: 'settings.ai.routing.workload.chat.label',
+  descriptionKey: 'settings.ai.routing.workload.chat.desc',
+};
+
+const CLAUDE_CODE_PROVIDER: CloudProvider = {
+  id: 'claude-code',
+  slug: 'claude-code',
+  label: 'Claude Code',
+  endpoint: '',
+  authStyle: 'bearer',
+  maskedKey: '',
+};
+
+const OPENAI_PROVIDER: CloudProvider = {
+  id: 'openai',
+  slug: 'openai',
+  label: 'OpenAI',
+  endpoint: 'https://api.openai.com/v1',
+  authStyle: 'bearer',
+  maskedKey: '••••',
+};
+
+/** Renders the dialog on `initial` and clicks Test, returning nothing — the
+ *  assertions read `testProviderModel`'s recorded arguments. */
+const renderAndTest = (initial: ProviderRef, cloudProviders: CloudProvider[]) => {
+  render(
+    <CustomRoutingDialog
+      workload={CHAT_WORKLOAD}
+      initial={initial}
+      cloudProviders={cloudProviders}
+      localModels={[]}
+      ollamaRunning={false}
+      modelRegistry={[]}
+      onClose={() => {}}
+      onSubmit={() => {}}
+    />
+  );
+  fireEvent.click(screen.getByRole('button', { name: 'Test' }));
+};
+
+describe('CustomRoutingDialog — Test button provider string', () => {
+  /**
+   * #6125: `currentProviderString` branched only on `cloud`, so every other
+   * non-managed kind fell into the `ollama:` arm. A Claude Code route was tested
+   * against the local runtime — the CLI was never spawned and the test could
+   * only fail, while the button above still read "Claude Code CLI · sonnet".
+   */
+  it('sends claude-code:<model> for a Claude Code CLI route', async () => {
+    vi.mocked(testProviderModel).mockResolvedValue({ reply: 'hi' });
+
+    renderAndTest({ kind: 'claude-code', model: 'sonnet' }, [CLAUDE_CODE_PROVIDER]);
+
+    await waitFor(() =>
+      expect(testProviderModel).toHaveBeenCalledWith('chat', 'claude-code:sonnet', 'Hello world')
+    );
+    // The banner renders the same string, which is how the two disagreed before:
+    // the error text used `registrySlug` (correct) while the call used `ollama:`.
+    expect(await screen.findByText('Provider: claude-code:sonnet')).toBeInTheDocument();
+  });
+
+  /**
+   * Test must exercise exactly what Save persists. `TemperatureOverrideField` is
+   * rendered for every source kind, so a claude-code route can carry an override
+   * and `serializeProviderRef` writes it into the stored route.
+   */
+  it('appends the temperature override, matching what Save persists', async () => {
+    vi.mocked(testProviderModel).mockResolvedValue({ reply: 'hi' });
+    const saved: ProviderRef = { kind: 'claude-code', model: 'sonnet', temperature: 0.7 };
+
+    renderAndTest(saved, [CLAUDE_CODE_PROVIDER]);
+
+    await waitFor(() =>
+      expect(testProviderModel).toHaveBeenCalledWith(
+        'chat',
+        'claude-code:sonnet@0.7',
+        'Hello world'
+      )
+    );
+    expect(vi.mocked(testProviderModel).mock.calls[0][1]).toBe(serializeProviderRef(saved));
+  });
+
+  it('leaves a cloud route unchanged', async () => {
+    vi.mocked(testProviderModel).mockResolvedValue({ reply: 'hi' });
+
+    renderAndTest({ kind: 'cloud', providerSlug: 'openai', model: 'gpt-4o' }, [OPENAI_PROVIDER]);
+
+    await waitFor(() =>
+      expect(testProviderModel).toHaveBeenCalledWith('chat', 'openai:gpt-4o', 'Hello world')
+    );
+  });
+
+  it('leaves a local route on the ollama: prefix', async () => {
+    vi.mocked(testProviderModel).mockResolvedValue({ reply: 'hi' });
+
+    renderAndTest({ kind: 'local', model: 'llama3' }, []);
+
+    await waitFor(() =>
+      expect(testProviderModel).toHaveBeenCalledWith('chat', 'ollama:llama3', 'Hello world')
+    );
+  });
+
+  /** Managed resolves its route server-side at request time, so there is no
+   *  provider string to send and Test is disabled rather than sending `cloud`. */
+  it('disables Test for a managed route', () => {
+    render(
+      <CustomRoutingDialog
+        workload={CHAT_WORKLOAD}
+        initial={{ kind: 'default' }}
+        cloudProviders={[]}
+        localModels={[]}
+        ollamaRunning={false}
+        modelRegistry={[]}
+        onClose={() => {}}
+        onSubmit={() => {}}
+      />
+    );
+
+    expect(screen.getByRole('button', { name: 'Test' })).toBeDisabled();
+    expect(testProviderModel).not.toHaveBeenCalled();
+  });
+});
--- a/app/src/components/settings/panels/ai/CustomRoutingDialog.tsx
+++ b/app/src/components/settings/panels/ai/CustomRoutingDialog.tsx
@@ -154,18 +154,30 @@ export const CustomRoutingDialog = ({
     setTestBusy(false);
   };
 
-  const currentProviderString =
-    source == null || source.kind === 'managed'
-      ? null
-      : source.kind === 'cloud'
-        ? appendTemperatureToProviderString(
-            `${source.providerSlug}:${model.trim()}`,
-            temperature == null || !Number.isFinite(temperature) ? null : temperature
-          )
-        : appendTemperatureToProviderString(
-            `ollama:${model.trim()}`,
-            temperature == null || !Number.isFinite(temperature) ? null : temperature
-          );
+  // Exhaustive on `CustomDialogSource['kind']` rather than a cloud-vs-rest
+  // ternary. The old shape sent *every* non-cloud kind as `ollama:<model>`, so a
+  // Claude Code route tested against the local runtime and could only ever fail
+  // — while the dialog above it read "Claude Code CLI · <model>" (#6125). A
+  // future source kind is now a compile error here instead of another silent
+  // `ollama:` fallback, so do NOT add a `default:` arm.
+  //
+  // The `@<temp>` suffix is appended on every arm, matching `serializeProviderRef`
+  // — the Save path. `TemperatureOverrideField` below is rendered for every kind,
+  // so a claude-code route really can carry an override, and Test exists to
+  // exercise the string the saved route will actually run. A Test that quietly
+  // dropped a field Save persists would be the same defect in a smaller size.
+  const currentProviderString = ((): string | null => {
+    if (source == null || source.kind === 'managed') return null;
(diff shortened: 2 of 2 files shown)


YellowSnnowmann:
Answering the two open points from the **Linked Issues check** (marked inconclusive).

### 1. Changed-line coverage — 100%, from CI's own gate

The `PR CI Gate` job's *Enforce >= 80% coverage on changed lines* step ([run 34353279368](https://github.com/tinyhumansai/openhuman/actions/runs/34353279368)) reports:

```
-------------
Diff Coverage
Diff: origin/main...HEAD, staged and unstaged changes
-------------
app/src/components/settings/panels/ai/CustomRoutingDialog.tsx (100%)
-------------
Total:   7 lines
Missing: 0 lines
Coverage: 100%
-------------
diff-cover measured 7 changed line(s)
```

The gate did not short-circuit: `coverage=true`, `frontend_needed=true`, `frontend=success`, and the `diff-coverage-report` artifact was uploaded. (7 is the count of *instrumented* changed lines — the rest of the hunk is comments and braces, which diff-cover does not count.)

### 2. The temperature suffix deliberately differs from the issue's snippet

Confirmed — and it is intentional, not an oversight. The issue's *Solution (optional)* block omits `@<temp>` on the `claude-code` arm; this PR appends it. Reasons, all verifiable in the tree:

- `serializeProviderRef` — the **Save** path — already appends it for `claude-code` via `joinModelAndTemp` (`app/src/services/api/aiSettingsApi.ts:262`).
- `TemperatureOverrideField` is rendered **unconditionally** in this dialog (`CustomRoutingDialog.tsx:298`), outside any `source.kind` guard, and `handleSave` carries the value into the claude-code ref (`:183-184`). So a Claude Code route genuinely can be saved as `claude-code:sonnet@0.7`.
- Omitting it in **Test** would therefore mean Test exercised `claude-code:sonnet` while the saved route ran `claude-code:sonnet@0.7` — a Test button that does not test what runs. That is the same defect class this PR fixes, one size smaller.
- The issue's stated reason for omitting was the factory's `log::warn!` (`src/openhuman/inference/provider/factory_part_03.rs:157-164`). That warning already fires on every chat-model construction from the *persisted* route, so suppressing it in Test hides from the tester a warning their real route produces.

The issue's acceptance criteria are unaffected by this: they require the banner to read `Provider: claude-code:<model>` and the RPC to reach the CLI, both of which hold with or without the suffix. The second test in `CustomRoutingDialog.test.tsx` pins the choice by asserting the Test string equals the **real, unmocked** `serializeProviderRef` output for the same `ProviderRef`, so the two serialisers cannot drift apart silently in future.

Happy to drop the suffix if a maintainer prefers literal fidelity to the issue snippet — it is a one-line change and the AC pass either way.

Issue #6158, opened 2026-09-09
Cron: agent jobs warn 'scheduled more frequently than every 5 minutes' but are not blocked


Al629176:
## Summary

Two cron agent jobs warn that they are scheduled more frequently than the 5-minute minimum interval, but the scheduler runs them anyway. The warning is emitted without enforcement.

## Log evidence

```
07:07:57 WRN [scheduler] Cron agent job 'b30826b0-ee0a-4129-ace3-cd1ed954d2af' is scheduled more frequently than every 5 minutes

09:09:58 WRN [scheduler] Cron agent job '6a5f9431-b70a-4b80-a8c8-233595af0be8' is scheduled more frequently than every 5 minutes
```

## Impact

- Agent jobs running on sub-5-minute intervals generate significantly more LLM traffic and cost than expected.
- Users who set an aggressive cron schedule accidentally (e.g. `*/3 * * * *`) receive no UI error and no enforcement — only a log warning they cannot see.
- Two different jobs are affected, suggesting this is a common misconfiguration.

## Expected behaviour

Either:
1. **Enforce the limit**: reject cron jobs at creation time if the interval is < 5 minutes, returning a clear error to the user.
2. **Rate-limit execution**: run the job but silently skip runs that would fire within 5 minutes of the previous run.
3. **At minimum**: surface the warning in the UI (notifications panel or job settings page) so the user is aware.

## Environment

- OpenHuman 0.63.17 (aarch64-apple-darwin)
- Job IDs: `b30826b0-ee0a-4129-ace3-cd1ed954d2af`, `6a5f9431-b70a-4b80-a8c8-233595af0be8`
- Log date: 2026-09-09

M3gA-Mind:
## Verdict: split — Misdiagnosed on the warning, Confirmed on the missing enforcement

The warning is a **bug in the check**, not evidence of a misconfigured job. It fires for *every* cron-scheduled agent job, whatever its expression. The two job ids in the report are very likely not high-frequency at all. **Open PR #6159 already fixes exactly this**, and its diagnosis is correct — I re-derived it independently below.

The second half stands: there is genuinely no minimum-interval enforcement anywhere in the cron module. That is a product gap, not a defect.

Checked against `21c96ee8c` (main).

### Root cause — `cron/scheduler_part_03.rs:78-85`

```rust
Schedule::Cron { .. } => {
    let now = Utc::now();
    match (
        next_run_for_schedule(&job.schedule, now),
[...]

Issue #6145, opened 2026-09-09
Memory leak: 24 stale tool instances not freed when tools Arc is shared across turns


Al629176:
## Summary

When an agent session runs multiple turns, `refresh_delegation_tools` detects that the internal `Arc<Tools>` is shared (reference count > 1) and falls back to refreshing only the schema — leaving the old tool instances alive. After successive turns, this accumulates stale instances that are never freed until the next "unique-owner" refresh.

## Log evidence

```
09:10:05:WRN:log [agent] refresh_delegation_tools: tools Arc is shared — refreshed schema only (25 synthesised tool name(s)); 24 stale tool instance(s) pending removal on the next unique-owner refresh
  log.file="src/openhuman/agent/harness/session/turn/tools.rs" log.line=586

09:10:05:INF:log [agent] refresh_delegation_tools: reconciled delegation schema for agent 'orchestrator'; now 25 synthesised tool name(s); added=["delegate_to_integrations_agent"] removed=[] tools_reconciled=false pending_tool_instances=24
  log.file="src/openhuman/agent/harness/session/turn/tools.rs" log.line=647
```

## Impact

- **tools_reconciled=false** — the tool list visible to the agent may differ from live config until a "unique-owner" refresh coincidentally occurs.
- **24 stale tool instances** held in memory; each instance may hold connections, allocated buffers, or file handles.
- The condition reproduces reliably after a multi-turn agent session starts a new cron/skill run while the previous session arc is still referenced by the UI renderer or a background subscriber.

## Reproduction

1. Start a multi-turn agent session (e.g. morning briefing cron job).
2. While it is running (or shortly after), trigger another skill/cron run involving delegation tools.
3. Observe the WRN in `tools.rs:586`.

## Expected behaviour

`refresh_delegation_tools` should either:
- **Wait** for the Arc to reach unique ownership before returning, or
- **Copy** the tools set so a fresh unique-owned `Arc` can be returned immediately, then clean up the old one once its refcount drops to zero.

## Environment

- OpenHuman 0.63.17 (aarch64-apple-darwin)
- Log date: 2026-09-09
- PID: 39982 / uptime ~385 s at observation time

M3gA-Mind:
# #6145 — "Memory leak: 24 stale tool instances not freed"

**Verdict: Misdiagnosed.** There is no memory leak. A real bug does live at the
reported site, but it is schema/instance drift, and its damaging half is the
*revoke* direction, not the connect direction the issue describes.

[...]

Issue #4452, opened 2026-07-03
Subagent empty tool allowlist is fail-open — child inherits parent's full tool surface (shell/spawn)


senamakel:
Part of the v0.58.7 → HEAD TinyAgents-migration audit (see parent issue).

## Problem
Sub-agent tool exposure inverted from **fail-closed to fail-open**. Old `subagent_runner/ops/tool_source.rs:53-107` (v0.58.7) denied any tool not in the allowlist — an **empty** allowlist meant deny-all. The new registration predicate at `src/openhuman/tinyagents/mod.rs:1126` is `allowed.is_empty() || allowed.contains(name)` — an empty allowlist now registers **every** tool across `tool_sets`, which includes `parent.all_tools`: shell, file-write, spawn/delegate tools.

Reachable by:
- a custom agent TOML with `tools = []` (`ToolScope::Named` is user-deserializable, `src/openhuman/agent/harness/definition.rs:621-629`);
- a `skill_filter` that matches zero tools;
- a `named` list whose entries are absent from the parent surface (silently resolves to empty).

Compounding factors: the spawn-tool strip at `subagent_runner/ops/runner.rs:415` operates on `allowed_indices` only, so the "sub-agents must never spawn" invariant is bypassed too; and subagent turns pass `tool_policy: None` (`ops/graph.rs:267`), so registration is the **only** gate.

## Failure scenario
A deliberately tool-less summarizer agent processing untrusted content (web page, email) silently gains `shell` + file-write + spawn — a prompt-injection privilege escalation.

## Fix plan
1. In `src/openhuman/tinyagents/mod.rs:1126`, distinguish "no filter supplied" (`None` → all tools) from "empty allowlist" (`Some(empty)` → deny all). Change the plumbed type from `Vec<String>`/`HashSet<String>` to `Option<HashSet<String>>` end-to-end (`subagent_runner/ops/runner.rs:690-698` → `assemble_turn_harness`).
2. Make `ToolScope::Named([])` and a zero-match `skill_filter` produce `Some(empty)` (deny-all), matching v0.58.7 semantics. Log a warn (`[subagent] tool allowlist resolved empty — registering no tools`) so misconfigured agents are diagnosable.
3. Re-assert the spawn-tool invariant at registration time (strip spawn/delegate names from whatever set is registered, not just from `allowed_indices`).
4. Consider also passing a real `tool_policy` on the subagent path instead of `None` (defense in depth).

## Acceptance criteria
[...]

PR #6177, opened 2026-09-10
fix(subagent): warn only when a spawn tool was actually readmitted

### LINKED ISSUE #6157: Subagent run silently refuses 5 delegation tools without user-visible error
## Summary

When a sub-agent run is initialised, the tool registry refuses to register 5 spawning/delegation tools with a WRN log entry per tool. The refusals are silent — no error is surfaced to the user or the calling agent, so the agent proceeds without knowing these tools are unavailable.

## Log evidence

```
09:10:13 WRN [subagent] refusing to register spawn/delegate tool on sub-agent run tool="spawn_subagent"
09:10:13 WRN [subagent] refusing to register spawn/delegate tool on sub-agent run tool="agent_prepare_context"
09:10:13 WRN [subagent] refusing to register spawn/delegate tool on sub-agent run tool="delegate_to_personality"
09:10:13 WRN [subagent] refusing to register spawn/delegate tool on sub-agent run tool="delegate_graph"
09:10:13 WRN [subagent] refusing to register spawn/delegate tool on sub-agent run tool="delegate_tools_agent"
```

## Impact

- If the orchestrator agent expects any of these tools to be available on the sub-agent, tool calls will fail at runtime with a cryptic "unknown tool" error rather than a clear capability message at session start.
- The calling agent cannot adapt its plan based on missing tools it was never told about.
- Log noise: 5 WRN lines per sub-agent invocation.

## Expected behaviour

Two options:

1. **This is by design** (sub-agents should not be able to re-spawn): document this restriction clearly, downgrade the log to DEBUG/TRACE, and ensure the agent system prompt omits these tools from sub-agent tool lists.
2. **This is a bug** (the tools should be available): fix the registration gate so the tools are correctly registered when a sub-agent is allowed to use them.

In either case, the refusal should be explicit in the sub-agent's visible tool list (the tool simply absent, not secretly refused at WRN level).

## Environment

- OpenHuman 0.63.17 (aarch64-apple-darwin)
- Log date: 2026-09-09

YellowSnnowmann:
## Summary

- Gate the sub-agent spawn/delegate refusal WARN on `admitted`, so a healthy sub-agent run is silent instead of emitting one WRN per spawn tool on the parent's surface.
- Strip spawn/delegate tools from a child's **dynamic** (per-spawn) tool list — the one route by which such a name could reach a child's allowlist admitted.
- Correct `is_subagent_spawn_or_delegate_tool`'s doc comment, which claimed to mirror the caller-side strip but matches strictly less than it.
- First tests for sub-agent tool-registration behaviour, driven through the real graph path.

## Problem

Verified against `537854715` (current `main`), not the `21c96ee8c` the issue comments checked — **#6162 has landed since**, which matters for the numbers below.

**The refusal is correct and by design (#4452); only the log is wrong.** `run_typed_mode` strips spawn/delegate tools out of `allowed_indices` (`runner.rs:920-923`) and builds `allowed_names` from that stripped list (`:1246`), so by the time `assemble_turn_harness` runs, the name is already excluded. But the registration loop computed `spawn_stripped` and logged **before** consulting `admitted` (`mod_part_04.rs:260-266`):

```rust
let spawn_stripped = is_subagent_run && is_subagent_spawn_or_delegate_tool(name);
if spawn_stripped { tracing::warn!(...); }              // never consulted `admitted`
if !registered.contains(name) && admitted && !spawn_stripped { /* register */ }
```

`spawn_stripped` was therefore only ever true where `admitted` was already false — the defense-in-depth branch has never once done work — and the warning announced "a sub-agent ran" rather than "the #4452 invariant was violated". The genuine case it exists for looked identical in the log.

**The issue's stated impact is not reachable.** The model is only sent the tools the harness registered, so there is no "unknown tool" path for these five and the orchestrator's plan is unaffected (it is the orchestrator that holds `spawn_subagent`/`delegate_*`, not its children). What is real is the third bullet: log noise on a hot path, plus a WRN that trains operators to ignore a warning that will one day be real.

**Correction to the issue's log evidence:** it is **4** lines per run on current `main`, not 5. `delegate_tools_agent` is a synthesised per-archetype delegate, and #6162 moved synthesised delegates out of `self.tools` — so it is no longer in `parent.all_tools` and no longer a registration candidate. The four that remain (`spawn_subagent`, `agent_prepare_context`, `delegate_graph`, `delegate_to_personality`) are durable tool structs.

## Solution

**1. Gate the warn on `admitted`** (`mod_part_04.rs`). Same size, zero behaviour change — the tool was not registered before and is not registered now. The warning now means "the allowlist tried to readmit a spawn tool", which is the misbuilt-allowlist case the backstop was written for.

*Deliberate deviation from the issue text*, which asked to downgrade the log to DEBUG/TRACE. Two reasons: downgrading discards the alarm along with the noise (as @M3gA-Mind noted), and a debug-level line for exactly this **already exists** at `runner.rs:925` — `"removed sub-agent spawn tools from sub-agent's tool surface"`, carrying a `stripped` count. A second per-tool debug in the registration loop would be redundant.

**2. Close the one live-ish hole** (`tool_prep.rs` + `runner.rs`). Dynamic per-spawn tools (Composio per-action tools, `extract_from_result`) are appended *after* `allowed_indices` is built, so the strip never sees them, and their names went straight into `allowed_names` unchecked. New `strip_spawn_tools_from_dynamic` runs once where the list is finished — deliberately **not** at any single use, because the same `Vec` feeds five consumers (provider specs, allowlist, prompt catalogue, harness registration, custom-graph request) and filtering one leaves the others disagreeing. That is the spec/executable divergence class of #6145. Note `filtered_specs` is **not** dead on every path: `AgentGraph::Custom` passes it as `AgentTurnRequest.specs` (`runner.rs:1610`).

Latent today — no dynamic tool is named after a spawn tool — so no live behaviour change. It makes the `admitted` gate provably safe rather than safe-by-coincidence.

**3. Correct the false doc comment** (`mod_part_01.rs`). It claimed to mirror `tool_prep::is_subagent_spawn_tool`. It does not: the primary resolves each archetype's `delegate_name` override through the registry (`plan`, `run_code`, `research`, …), while this copy matches only the `delegate_` prefix — the backstop catches strictly *less* than the thing it backstops, which is backwards. The comment now says so, and says why widening it here is the caller's job rather than a registry lookup on the per-tool registration loop.

## Submission Checklist

- [x] Tests added or updated (happy path + failure/edge case) — 5 new; happy path is the quiet run, edge cases are the readmitted spawn tool, unprefixed `delegate_name` overrides, and a no-op list
- [x] **Diff coverage ≥ 80%** — new tests cover every changed line; both CI clippy lanes and the touched-module suite run clean locally
- [x] Coverage matrix updated — new row 6.3.16
- [x] All affected feature IDs listed under `## Related`
- [x] No new external network dependencies introduced — tests use an in-process fake `ChatModel`
- [x] N/A: manual smoke — no release-cut surface touched (log level + an internal tool-list filter)
- [x] Linked issue closed via `Closes #6157`

## Impact

Desktop/CLI, Rust core. No API, RPC, config, schema or persistence change; no new dependency; no migration.

- **Log volume**: 4 fewer WRN lines per sub-agent run. Nothing consumes that line — no schema, RPC, UI or test reads it.
- **Security posture**: strictly tightened. The #4452 invariant now holds on the dynamic-tool path too, which was the only one where a spawn/delegate name could arrive admitted.
- **Behaviour**: unchanged on every reachable path today.

**Known test gap, stated rather than papered over:** nothing pins the *call* to `strip_spawn_tools_from_dynamic` in `run_typed_mode` — deleting that line leaves the suite green, because the tests exercise the function directly. Reaching it end-to-end needs a tool inside `dynamic_tools` specifically (a Composio per-action tool or `extract_from_result`); the cheap `extra_tools` route goes through the older strip at `runner.rs:943` and would not exercise the new code. Happy to add the harness if a reviewer thinks it earns its keep.

## Related

- Closes: #6157
- Coverage matrix: 6.3.16
- Context: #4452 (the invariant), #6162 (moved synthesised delegates out of `self.tools`, changing the log evidence from 5 lines to 4), #6158 (same shape — a log-only side effect with nothing to assert)
- Follow-up PR(s)/TODOs: none

---

### Linear Issue

- Key: N/A
- URL: N/A

### Commit & Branch

- Branch: `fix/6157-subagent-spawn-warn-gate`
- Commit SHA: 6f8b6d6f7

### Validation Run

- [x] `pnpm --filter openhuman-app format:check` — N/A: no frontend files changed
- [x] `pnpm typecheck` — N/A: no TypeScript changed
- [x] Focused tests: 5 new tests pass; `subagent_runner` + `tinyagents` modules 445 passed / 0 failed
- [x] Rust fmt/check: `cargo fmt --all --check` clean; `cargo clippy -p openhuman -- -D warnings` and the product-feature lane both rc=0; Rust layout gate clean (`mod_part_04.rs` 749/750, `runner.rs` 1750/1769)
- [x] Tauri fmt/check — N/A: `app/src-tauri` untouched

### Validation Blocked

- `command:` none
- `error:` none
- `impact:` none

### Behavior Changes

- Intended behavior change: the registration-time spawn refusal warns only when the allowlist actually readmitted the tool; dynamic per-spawn tools are filtered through the same #4452 strip as the static surface.
- User-visible effect: 4 fewer WRN lines per sub-agent run. No functional change on any reachable path.

### Parity Contract

- Legacy behavior preserved: registration outcome is byte-identical — a spawn/delegate tool was never registered on a sub-agent run before this change and still is not.
- Guard/fallback/dispatch parity checks: mutation-tested — restoring the unconditional `if spawn_stripped` fails `a_sub_agent_cannot_reach_a_spawn_tool_and_the_healthy_run_is_quiet` and nothing else. Both refusal tests assert the tool never *executes* (a probe tool records dispatch), so unreachability is proven, not merely absence from the advertisement.

- Duplicate PR(s): none
- Canonical PR: this one
- Resolution: N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented sub-agents from invoking tools that create or delegate to additional agents, including dynamically provided tools.
  - Added safeguards to reject attempts to re-enable restricted agent-spawning tools.
  - Adjusted warnings so normal sub-agent runs remain silent, while unauthorized re-admission attempts generate a warning.

- **Documentation**
  - Updated test coverage documentation for sub-agent tool restrictions and warning behavior.


YellowSnnowmann:
@coderabbitai Addressed the docstring-coverage check in `8f0466b56`.

The 60.71% was entirely test scaffolding: the `Tool`/`ChatModel` impls on `SpawnProbeTool`, `SpawnAttemptProvider` and `NamedTool`, the warn-capture layer's methods, the two list helpers, and the five test functions — whose intent was written as `//` comments inside the body rather than as doc comments, so it did not count.

Each fake now states which of its methods the code under test actually reads, which is the fact worth having: the strip and the registration gate look at `name()` and nothing else, and reaching `SpawnProbeTool::execute` means the test has failed rather than passed.

Doc comments only — no behavioural change. Re-verified locally: the 5 tests still compile and pass, `cargo fmt --check` clean, Rust layout gate clean.

Also confirming the other four pre-merge checks (Title, Linked Issues, Out of Scope, Description) passed, and neither Codex review nor tinysweeper raised a finding.

